Issue #7430, opened 2017-06-23
Reader: Followed Sites can have an unmarked gap in content


aerych:
It seems there is a situation where the followed sites feed can have an unmarked gap in content.  As observed in a user's feed, the user was following 100s of blogs.  When returning to the followed sites stream after several hours new content would sync but there would be a gap of several hours between the new posts and the previously cached content and no gap button was displayed. 

I was able to reproduce this in the simulator by fetching the list of posts...then returning to the simulator after three or so hours.  

The causes is currently unclear... it might be due to an issue in the app code... it might be due to something with the api.

designsimply:
This came up in another mobile request with the following description:

<blockquote>For example, I’ll see a post in my Reader that says it was posted 8 hours ago and the next post says it was published 3 hours ago…a 5 hour gap. But when I open up Safari or Chrome on my iPhone and log into WordPress and go to the reader, there is no such gap. All the “missing” posts between 8 hours ago and 3 hours ago are there.</blockquote>

<sup>(internal references: p4a5px-2mT-p2, 1852243-zen)</sup>

designsimply:
@aerych can you take a look at this when you have a chance? It was recently fixed for WPAndroid via https://github.com/wordpress-mobile/WordPress-Android/issues/5491 🎉 

<sub>(internal reference: noting that p77Llu-74g-p2 may have some relevant discussion around pagination & sync)</sub>

aerych:
Potentially fixed with #11553 but testing this is wonky so I want to spend a little more time with it to be sure.

thehenrybyrd:
User came back to ask about this in 2028689-zen. It looks like https://github.com/wordpress-mobile/WordPress-iOS/pull/11553 will go out with 12.4. @aerych what's our confidence level with the fix?

designsimply:
@thehenrybyrd there is not a guarantee this issue will fix the case noted by the user in 2028689-zen but since there was a recent attempt to fix the problem that shipped with 12.4 it would be good to let that user know what was updated, that the update was shipped on 2019-05-20, and that if they're still experiencing problems then we may want to create a new issue with steps to reproduce that are as detailed as we can come up with based on their past reports.

designsimply:
@thehenrybyrd can you check back in with that user who reported seeing gaps in content for posts in the Reader? We are wanting to find out if the potential fix in #11553 that shipped in 12.4 was helpful for them.

pmciano:
[...]

PR #11553, opened 2019-04-25
Issues/7430 missing posts

aerych:
Refs #7430 
This has been a tricky glitch to sort out. This time round, while reviewing the code it occurred to me that if a site owner were to change the publish date of a previously published post it would be possible to recreate the behavior reported by our users due to the way overlaps were detected.  

When an overlap is detected the app assumes that all available content is present in the app and no gap marker is inserted to indicate there might be missing posts. 

Here's the scenario:
- A user follows the Coffee tag and syncs its posts, including one from Juan's Coffee Blog.
- Later, Juan publishes some more posts, but decides to change the date of an earlier post (one that was synced by our user) so it shows up at the top of his site.  (Poor Juan hasn't heard of sticky posts.)
- Our faithful users returns to the app some time later and syncs the Coffee tag again. 
- The newly fetched posts include the one with the changed date. 
- When the app checks to see if any of the newly fetched posts have are already synced, its going to flag the one with the date changed.   We actually don't want this. Since the post is changing its position in the list we want to treat it like new content, not an overlap. 

This PR improves the way overlaps are detected to check that dates match between what was synced and what was cached. If dates match then there is an overlap.  If dates do not match we do not treat it as an overlap. 

This should fix the problem our users are reporting.  (Unless there's something else going on.)

To test:
Testing is a little tricky. 
1. Edit ReaderPostService and set `ReaderPostServiceNumberToSync` to a low number like `5`. 
2. Edit ReaderStreamViewController and set `refreshInterval` to something like `10`. 
3. Comment out `syncHelper.syncMoreContent()` on line 1402 of ReaderStreamViewController to disable the loadMore mechanism otherwise it will conflict with a blog with only a handful of posts.
These will make it easier to test. 
4. Create a test blog and follow it. 
5. Add 5 test posts via a desktop browser
6. In the app go to Reader > Following > Manage > Your site and view its posts.
7. Tap the back button to leave the stream. 
8. Return to the blog and create ten more test posts.  Basically you want enough to create a gap
9. Edit one of the first posts you published so its publish date is the most recently published post.
10. Go back to the app and tap to view the posts for the site again. 
11. Confirm you see a gap marker. 

This seems to work for testing. Something odd I noticed was one of the posts was missing from the results, but it eventually showed up.  I'm tempted to chalk this up to caching but I'd be curious to see if it turns up in other tests. 

Needs Review: @frosty could I trouble you with this one?


aerych:
Thanks for testing @frosty! 

>  a) the early post with updated date still appeared down the bottom of the list (although it's in the correct place on a subsequent refresh), and b) I had a duplicate post 14 either side of the gap:

Interesting!  We definitely should not see duplicate posts. If I were to guess I'd say something is amiss with the merge logic that might account for both errant posts. I'll see what I can find. 

> The original 1 - 5 posts are now missing.

This is actually expected (and I say this assuming you commented out `syncHelper.syncMoreContent`).  When we fill a gap, we discard all content below the gap as a strategy to not have multiple gaps to fill. Scrolling down would normally trigger the load more behavior but if its commented out you wouldn't see anything after the batched fetched to fill the gap.

aerych:
@frosty it turns out there was a bit of logic that was preventing a posts sort rank from being updated (sort rank for non-search posts is the posts create date).  This should fix the issue with the post not moving. 
I wasn't able to reproduce the behavior you saw with the duplicate 14 post.  I'm not sure if that is because this change also fixed that issue or if there is some other subtle glitch having to do with the number of posts returned.  I'm tempted to think its because the latest fix corrects what might have been a counting issue with the original post not updating its sort location.  If you can still reproduce let me know. I might need to pick your brain for more detailed testing instructions to get it to repro.

Reader for another peek!

frosty:
@aerych Code changes look good to me, but I'm afraid I'm AFK for the rest of this week and won't have chance to re-review. Would you be able to get somebody else to take a look?

aerych:
@frosty No problem!  Thanks for taking a peek. We can hand off to another.

danielebogo:
@aerych I tested it again and I can confirm everything works! I saw the gap marker and the post I modified (changing also the date) moved as first post!
:shipit:

aerych:
Thanks @frosty and @danielebogo!

PR #8267, opened 2017-12-05
Swift App Delegate 3: Third Party SDK configuration

frosty:
Continuing from #8258, this PR moves methods for configuring the various third party SDKs we use over to Swift. Ported over:

* Reachability
* Hockey
* Analytics
* Crashlytics
* App Rating Utility

I also removed some unnecessary StoreKit code, and fixed a logic issue with analytics where it didn't reflect the way the editor and NUX are presented these days.

As with the previous PRs for the Great Swiftening, this PR is primarily concerned with just getting the code translated to Swift as-is – we can make improvements later.

To test:

* Check the Swift code against the obj-c code
* Ensure the app builds correctly for both Debug and Release builds (as there's some conditional code in there) and that tests pass

PR #11556, opened 2019-04-25
Swift App Delegate 4: Tracks and Keychain

frosty:
Continuing from #8267, this PR ports over the tracks and keychain methods from the app delegate into Swift. I'm trying to keep things as close as possible to the current implementation to make it easy to review. We can always go back and clean up the code afterwards.

I'm not sure we can easily test the keychain migration code, and I don't even think it's probably getting triggered any more, as it was originally written 6 years ago to fix a keychain storage issue which we since rectified. We could just remove it?

Also, per-file log levels seem difficult / not really supported in CocoaLumberjack for Swift. But everything in here is now `Info` level, so I think it shouldn't matter.

**To test:**

* Review the code
* Do a smoke test to ensure logging is still happening

@astralbodies would you mind taking a look?


frosty:
@astralbodies Thanks for the review! I removed `fixKeychainAccess`. Still okay for me to merge?

astralbodies:
@frosty Go for it!

Issue #8916, opened 2018-03-16
After creating a new post draft, it doesn't appear in the draft list


etoledom:
### Expected behavior
A new draft post appears in the draft list.

### Actual behavior
When creating a new draft and going back to the draft list, the newly created draft doesn't appear in the list until I refresh the list.

### Steps to reproduce the behavior
1. Go to the posts list
2. Display the drafts list
3. Open the editor to create a new Post
4. Save it as a Draft
5. Close the editor to go back to the Draft list
6. The list will be as it was before, without displaying the new draft

![vid2](https://user-images.githubusercontent.com/9772967/37521037-03737fe4-28fe-11e8-854f-5bf23018d181.gif)

frosty:
Thanks for opening this. This is likely to change fairly soon, as we make tweaks for async – we'll ensure that posts show up in the list during the upload process and reflect their current status :)

frosty:
This is fixed in one situation, and still valid in another:

* Following the steps shown above (tapping ... > Save as Draft, then leaving the editor), the issue still persists and the post doesn't immediately appear in the list.
* If you instead tap the X in the editor and choose "Save Draft" from that menu, the draft appears in the posts list as soon as the editor disappears.

PR #11558, opened 2019-04-25
Draft missing from post list

elibud:
**Fixes** #8916 

**To test:**

1. Go to the posts list
2. Display the drafts list
3. Open the editor to create a new Post
4. Save it as a Draft from the top right three dots menu
5. Close the editor from the cross on the top left and to go back to the Draft list
6. The newly created post should be there

Also test: publishing a new post from the draft list, from the published list, scheduling posts, etc. Basically we need to run through all the publishing scenarios to make sure nothing is broken.

Notes:

There were two issues at play here:
 - The `shouldRemovePostOnDismiss` property was being calculated during the ViewController initialization, which was incorrect because the post status could change later on.
- When closing the editor from the top left cross, and assumption that there were no changes was made, and this is not correct since the post could've been previously saved via the right three dots menu. Basically when we close the editor we can't assume no changes were made unless the post was created and deleted/discarded.

I've made the minimum amount of changes that fix the bug, however there's a lot of dead code on `PostEditor+Publish`, and logic around PostPost that should be cleaned up in a subsequent PR.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -232,8 +232,9 @@ extension PostEditor where Self: UIViewController {
 
         post = originalPost
         post.deleteRevision()
+        let shouldRemovePostOnDismiss = post.hasNeverAttemptedToUpload() && !post.isLocalRevision
 
-        if post.hasNeverAttemptedToUpload() && !post.isLocalRevision {
+        if shouldRemovePostOnDismiss {
             post.remove()
             postDeleted = true
         } else if shouldCreateDummyRevision {
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -232,7 +232,7 @@ extension PostEditor where Self: UIViewController {
         post = originalPost
         post.deleteRevision()
 
-        if shouldRemovePostOnDismiss {
+        if post.hasNeverAttemptedToUpload() && !post.isLocalRevision {
             post.remove()
         } else if shouldCreateDummyRevision {
             post.createRevision()
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -192,14 +192,13 @@ extension PostEditor where Self: UIViewController {
             showPostHasChangesAlert()
         } else {
             editorSession.end(outcome: .cancel)
-            discardChangesAndUpdateGUI()
+            discardUnsavedChangesAndUpdateGUI()
         }
     }
 
-    func discardChangesAndUpdateGUI() {
+    func discardUnsavedChangesAndUpdateGUI() {
         discardChanges()
-
-        dismissOrPopView(didSave: false)
+        dismissOrPopView()
     }
 
     func discardChanges() {
@@ -283,7 +282,7 @@ extension PostEditor where Self: UIViewController {
         // Button: Discard
         alertController.addDestructiveActionWithTitle(discardTitle) { _ in
             self.editorSession.end(outcome: .discard)
-            self.discardChangesAndUpdateGUI()
+            self.discardUnsavedChangesAndUpdateGUI()
         }
 
         alertController.popoverPresentationController?.barButtonItem = navigationBarManager.closeBarButtonItem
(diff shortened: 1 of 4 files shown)


elibud:
Thanks @pinarol and I agree, we definitely want more eyes on it. I'll wait until either @diegoreymendez or @koke take it for a spin.

diegoreymendez:
Fwiw, I think this code is complex enough that no-one can really give a good answer.  I'm wrapping something up but will give it a look as soon as possible.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -374,13 +373,13 @@ extension PostEditor where Self: UIViewController {
...
 
-    func dismissOrPopView(didSave: Bool, shouldShowPostEpilogue: Bool = true) {
+    func dismissOrPopView() {
         stopEditing()
 
         WPAppAnalytics.track(.editorClosed, withProperties: [WPAppAnalyticsKeyEditorSource: analyticsEditorSource], with: post)
 
         if let onClose = onClose {
-            onClose(didSave, shouldShowPostEpilogue)
+            onClose(true, false)
```
koke: I'm resisting the urge to suggest further refactors, since this is so hard to follow, but if I had to pick just one, this method could really benefit from some argument labels.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -232,7 +231,7 @@ extension PostEditor where Self: UIViewController {
         post = originalPost
         post.deleteRevision()
 
-        if shouldRemovePostOnDismiss {
+        if post.hasNeverAttemptedToUpload() && !post.isLocalRevision {
```
koke: I think `shouldRemovePostOnDismiss` was more descriptive of what was being checked, perhaps we can keep it as a computed property?

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
elibud: Yes, I felt the same way while writing the code, there are a ton of improvements that could be made here, but on this PR I'm just trying to fix the bug.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
elibud: I'm reluctant to have a property for something that's calculated only once and is a one liner &&. I also want to avoid having to lose the context of the method when you go check what `shouldRemovePostOnDismiss` really means.
But if you feel strongly about it I'll change it :D

elibud:
@koke Yes, changing previous behavior was the only way to fix this one. Basically you can't really know if changes were saved or not when you close the editor from the top left cross, any value for the `changesSaved` parameter will be incorrect on some cases. If you always send `false` as we were doing it would be incorrect for any post saved from the three dot menu.
Good find on the testing, I'm checking to see if I can figure out a solution.

elibud:
Ready for another look, now instead of assuming changes were always saved, I'm assuming they were always saved unless the post was deleted.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
diegoreymendez: I agree with both (so helpful, I know!).

One middle ground option (that I often use when I want to provide more clarity) could be something along the lines of:

```swift
let shouldRemovePostOnDismiss =
        post.hasNeverAttemptedToUpload()
        && !post.isLocalRevision

if shouldRemovePostOnDismiss {
```

It doesn't really change the logic, but makes the code slightly more readable.

elibud: I like it, changes made!

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -192,19 +192,20 @@ extension PostEditor where Self: UIViewController {
...
+            discardUnsavedChangesAndUpdateGUI()
         }
     }
 
-    func discardChangesAndUpdateGUI() {
-        discardChanges()
-
-        dismissOrPopView(didSave: false)
+    func discardUnsavedChangesAndUpdateGUI() {
+        let postDeleted = discardChanges()
```
koke: I wasn't supper happy about this, since it's not obvious that a method named `discardChanges` would return whether a post was deleted or not, and then `didSave == !postDeleted` seems like a non-obvious assumption. However I kept following the code and the whole thing is quite confusing, so I'll 🙈 for now and hope that having autosaves eventually allows us to simplify all this logic.

elibud: I didn't like that either, but yes, the whole thing is a mess, we need to rewrite it.

diegoreymendez: I agree.  I think it would be nice to try to think of ways in which we can incrementally improve this.

Issue #8844, opened 2018-03-07
Launch timeout on iPhone 6 Plus


koke:
We've had some reports of the app crashing right after launch, and inspecting the user's crash reports it looks like it's timing out and being killed:

```
Exception Type:  EXC_CRASH (SIGKILL)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: Namespace SPRINGBOARD, Code 0x8badf00d
Termination Description: SPRINGBOARD, scene-create watchdog transgression: org.wordpress exhausted CPU time allowance of 4.31 seconds |  | Elapsed total CPU time (seconds): 23.710 (user 23.710, system 0.000), 100% CPU | Elapsed application CPU time (seconds): 5.314, 22% CPU | 
```

The stack traces are all different but somewhere inside the launch sequence.

gemmagarner:
Reported in 1030062-zen (iPhone 6s, iOS 10.2).

gemmagarner:
Reported in 1040635-zen (iPhone 7, iOS 11.2)

gemmagarner:
1040782-zen (iPhone 6, iOS 11.1)

koke:
Tagging for groundskeeper rotation. I know it's hard to work on this since we can't reproduce consistently, but let's try something:
- Investigate if we can get metrics of the amount of timeouts on launch (I'm not sure Fabric's crash count is including this).
- Profile the startup sequence and make sure only what is essential is blocking the app launch.
- If there are no clear solutions, let's document what's happening on launch and why it can't be deferred.

brucea8c:
I checked in with the user in 1040635-zen and she reported that the problem is no longer happening.

designsimply:
Because the original issue was filed over a year ago with no new reports for ~1y and at least one report saying the problem is no longer happening, I'm asking that a Groundskeeper please investigate following @koke's notes and report what you find. Please close the issue if it is no longer actionable based on the findings.

jklausa:
I tried to reproduce it for few hours today but came up empty-handed — I think I'm not well-versed enough in Core Data to handle this.

@astralbodies — would you be able to take a quick peek somewhere this week to maybe point me in the right direction? I _think_ the code is doing all the right things when it comes to handling concurrency, but I feel like I'm in over my head.

jklausa:
[...]

PR #11559, opened 2019-04-25
Podfile: Link all non shared pods as static frameworks to reduce pre-main launch time

jtreanor:
## Description

Currently, all of our CocoaPods dependencies are loaded as dynamic frameworks. Consequently, the app currently loads 44 non-system dynamic frameworks on launch. It is well documented that loading a large number of dynamic frameworks can slow down the time it takes to launch an app (see https://github.com/artsy/eigen/issues/586).

This moves most of the pods we use to be loaded as static frameworks (supported since [CocoaPods 1.5](http://blog.cocoapods.org/CocoaPods-1.5.0/)). It does not make all pods static as that would cause duplicate symbols with `WordPressComStatsiOS`. It reduces the number of non-system dynamic frameworks loaded on launch to 16. I suspect some of the pods could be removed from `WordPressComStatsiOS` to reduce this more.

In my testing, the difference is much more noticeable on older devices. I have tested on an iPhone XS and an iPhone 6. The first launch after install or boot is the slowest. Frameworks can be cached on subsequent launches.

Possibly related to https://github.com/wordpress-mobile/WordPress-iOS/issues/8844.

## Benchmarks

The following was measured with `DYLD_PRINT_STATISTICS` on an iPhone 6 running iOS 12.0.

### Before

```
Total pre-main time: 1.8 seconds (100.0%)
         dylib loading time: 1.4 seconds (78.5%)
        rebase/binding time:  89.79 milliseconds (4.8%)
            ObjC setup time:  98.66 milliseconds (5.2%)
           initializer time: 211.11 milliseconds (11.3%)
           slowest intializers :
             libSystem.B.dylib :   8.72 milliseconds (0.4%)
                         React : 299.66 milliseconds (16.0%)
```

### After

```
Total pre-main time: 1.2 seconds (100.0%)
         dylib loading time: 825.39 milliseconds (66.8%)
        rebase/binding time:  79.56 milliseconds (6.4%)
            ObjC setup time: 167.34 milliseconds (13.5%)
           initializer time: 163.08 milliseconds (13.1%)
           slowest intializers :
             libSystem.B.dylib :   7.68 milliseconds (0.6%)
                     WordPress : 251.09 milliseconds (20.3%)
```

## Testing

A smoke test of the app should confirm that this change is working correctly. Since pods are loaded statically the most likely thing to go wrong would be incorrect loading of resources (e.g. images missing from UI).

Update release notes:


Podfile
```diff
@@ -273,3 +273,28 @@ target 'WordPressScreenshotGeneration' do
...
+          dynamic << pod
+          next
+        end
+        static << pod
+        def pod.static_framework?;
+          true
+        end
+    end
+    puts "Installing #{static.count} pods as static frameworks"
+    puts "Installing #{dynamic.count} pods as dynamic frameworks"
```
frosty: Nice touch! :)

PR #11560, opened 2019-04-25
Remove unused View-related classes

shiki:
I have found that these classes are no longer referenced in the code or in xib/storyboard files. 

## Testing

1. Validate that tests pass.
2. Please test for possible regressions. 

## Release Notes


WordPress/Classes/ViewRelated/Cells/WPReusableTableViewCells.swift
```diff
@@ -76,16 +76,6 @@ class WPTableViewCellValue1: WPReusableTableViewCell {
     }
 }
 
-class WPTableViewCellValue2: WPReusableTableViewCell {
```
frosty: I'd be tempted to leave this one here – the idea I think is that that there's a WPTableViewCell style available to match each of the standard table view cell styles (and it's only a short snippet).

shiki: Makes sense. Restored in 8bc8597c9b0bab4f886a9a066d2462d2b210c459.

PR #11561, opened 2019-04-25
[Stats Refresh] Insights: fix getting most popular day

ScoutHarris:
Fixes #n/a

When accessing Insights, if the raw `mostPopularWeekday` was 0, subtracting 1 from it would go 💥  when used as the index for `standaloneWeekdaySymbols`. This adds a check for that, and sets the index accordingly.

To test:
- On a site where the most popular day is the start of the week, go to Stats > Insights.
- Verify it doesn't crash.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsViewModel.swift
@@ -204,7 +204,7 @@ private extension SiteStatsInsightsViewModel {
         calendar.locale = Locale.autoupdatingCurrent
 
         // Back up mostPopularWeekday by 1 to get correct index for standaloneWeekdaySymbols.
-        mostPopularWeekday = mostPopularWeekday == 0 ? 6 : mostPopularWeekday - 1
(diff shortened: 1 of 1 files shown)


ScoutHarris:
Hey @SergioEstevao . Since you have a site that triggered this, would you mind testing this? Thank you!

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsViewModel.swift
```diff
@@ -193,29 +193,28 @@ private extension SiteStatsInsightsViewModel {
...
+            else {
                 return []
         }
 
         var calendar = Calendar.init(identifier: .gregorian)
         calendar.locale = Locale.autoupdatingCurrent
 
-        let dayString = calendar.standaloneWeekdaySymbols[mostPopularWeekday - 1]
+        // Back up mostPopularWeekday by 1 to get correct index for standaloneWeekdaySymbols.
+        mostPopularWeekday = mostPopularWeekday == 0 ? 6 : mostPopularWeekday - 1
```
SergioEstevao: Instead of hardcoding 6 here, what about of using 'calendar.standaloneWeekdaySymbols.count -1'

ScoutHarris: Sure. And done!

PR #11562, opened 2019-04-25
Added accessibility label to add image in new user epilogue

### LINKED ISSUE #11489: New Account VoiceOver: add image button label is not accessible
Steps to reproduce:

1. Log out of the WPiOS app.
1. Follow the prompts until you reach the New Account screen.
1. Tap the "Add Image" button to find that doesn't have a label.

Result: only "button" is announced when you tap on the "Add Image" button on the New Account screen. ([5s](https://cloudup.com/cV5xWztv8Rx))

voiceover-add-image-button-is-not-announced

stevebaranski:
Fixes #11489.

To test:
- Replicate the steps in the issue with VoiceOver, and verify that the accessible description is read.

Update release notes:

@danielebogo or @etoledom , would one of you be willing to take a look at this?

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/EpilogueUserInfoCell.swift
+++ b/WordPress/Classes/ViewRelated/NUX/EpilogueUserInfoCell.swift
@@ -18,9 +18,10 @@ extension EpilogueUserInfoCellViewControllerProvider where Self: UIViewControlle
 class EpilogueUserInfoCell: UITableViewCell {
 
     @IBOutlet var activityIndicator: UIActivityIndicatorView!
-    @IBOutlet var gravatarView: UIImageView!
-    @IBOutlet var gravatarAddIcon: UIImageView!
     @IBOutlet var gravatarActivityIndicator: UIActivityIndicatorView!
+    @IBOutlet var gravatarAddIcon: UIImageView!
+    @IBOutlet var gravatarButton: UIButton!
+    @IBOutlet var gravatarView: UIImageView!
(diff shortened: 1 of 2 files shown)


etoledom:
Thank you @danielebogo for testing it!

My grain of salt: I think it would be good to add a hint stating what this image is for. By the shape of the image you can tell that it will be used as an avatar for the site, but a VoiceOver user won't know that.

Or instead of a simple "Add image", it could be a more descriptive, something like "Add account image" (plus the hint).

danielebogo:
@etoledom that's a good point!

stevebaranski:
@danielebogo thanks for the review!

> My grain of salt: I think it would be good to add a hint stating what this image is for. By the shape of the image you can tell that it will be used as an avatar for the site, but a VoiceOver user won't know that.
> 
> Or instead of a simple "Add image", it could be a more descriptive, something like "Add account image" (plus the hint).

@etoledom that's valid feedback, thank you. 

I have pushed an additional commit that revises the description, and adds a hint as you suggested. I took special care to honor the distinctions between the two. Can you let me know if this is more inline with your expectations?

etoledom:
That looks great! Thank you @stevebaranski for updating it 👍 

:shipit:

designsimply:
LGTM! Dropping by with a video for reference because I re-tested this today using WP Internal 12.8.0.201907011 on iPhone 6S iOS 12.31. ([39s](https://cloudup.com/cnfOnDotUHS))

Issue #11467, opened 2019-04-12
Use Domain Credit


danielebogo:
All _WordPress.com plans_ include a custom domain name. This is given to the users as a _domain credit_, which they can redeem against any valid domain.

The image shows the new flow. The Domain Selection and the Register Domain exist within the Site Creation flow and [Automated Transfer flow](https://github.com/wordpress-mobile/WordPress-iOS/issues/9630). After completing the flow the user should be returned to the site dashboard.
![domain_credits_flow-1](https://user-images.githubusercontent.com/912252/56037253-ba76d080-5d26-11e9-800d-355f22c3b16f.png)

cc: @SylvesterWilmott

jaclync:
subtasks:
- [ ] Create feature flag `domainCredit`
- [ ] Add new subsection UI
- [ ] Add congrats screen UI
- [ ] Add logic that determines whether domain credit can be redeemed + testing
- [ ] Integrate the whole domain registration flow + testing
- [ ] Analytics

PR #11572, opened 2019-04-26
Domain credit success UI

jaclync:
Refs. #11467 - subtask - add success UI

* Set register domain row `showsSelectionState` to false, because its action presents a modal.
* Added `attributedSubtitleConfiguration` parameter in `NoResultsViewController` constructor/configuration to allow caller to further configure subtitle attributed text after default styling.
  * Passed `nil` value in Objective-C callers.
  * This addition is to allow different text style of domain string on success screen.
* Created `DomainCreditRedemptionSuccessViewController` that displays congrats messaging after domain credit redemption, with delegate to handle continue button action. It reuses UI of `NoResultsViewController` by using it as a child view controller.
* On site dashboard, on tapping "Register Domain" row, showed the new `DomainCreditRedemptionSuccessViewController`. After continue button is pressed on success screen, the screen is dismissed and a snackbar message with more details about domain registration is shown on site dashboard.

To test:
* Turn on `domainCredit` feature flag in `FeatureFlag` enum. You can also change domain name on success screen in `showDomainCreditRedemption`
* Go to a site/blog dashboard
* Tap on "Register Domain" row

Example expected screenshots

device | success screen | after dismissing success screen
--- | --- | ---
iPhone SE | Simulator Screen Shot - iPhone SE - 2019-04-26 at 21 19 43 | Simulator Screen Shot - iPhone SE - 2019-04-26 at 21 20 18
iPad Air 2 | Simulator Screen Shot - iPad Air 2 - 2019-04-26 at 21 45 14 | Simulator Screen Shot - iPad Air 2 - 2019-04-26 at 21 45 18

Update release notes:


WordPress/Classes/ViewRelated/Blog/DomainCredit/DomainCreditRedemptionSuccessViewController.swift
```diff
@@ -0,0 +1,69 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        let attributedSubtitleConfiguration: NoResultsViewController.AttributedSubtitleConfiguration = {
+            [weak self] attributedText in
+            guard let `self` = self else {
+                fatalError()
```
danielebogo: You can use _self_ with no backticks. We use this approach in many places.
I was thinking, since `attributedSubtitleConfiguration` is an optional argument, can the `AttributedSubtitleConfiguration` return an optional? In this way you don't need to re-bind self but just use the weak reference. What do you think?

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
```diff
@@ -41,6 +41,11 @@ import Reachability
     private var accessorySubview: UIView?
     private var hideImage = false
 
+    /// Allows caller to customize subtitle attributed text after default styling.
+    typealias AttributedSubtitleConfiguration = (_ attributedText: NSAttributedString) -> NSAttributedString
+    /// Called after default styling of attributed subtitle, if non nil.
```
danielebogo: See the comment above 😊

SylvesterWilmott:
The screenshots look good @jaclync , thanks!

One thing I noticed is that the footer text on the register domain row is (i assume) sitting adjacent to the section heading on "Next Steps". We're using a typical spacing of 20pt in instances like this.

WordPress/Classes/ViewRelated/Blog/DomainCredit/DomainCreditRedemptionSuccessViewController.swift
jaclync: > You can use self with no backticks. We use this approach in many places.

thanks for pointing it out, now I know - looks like being able to write ``guard let `self` = self else { ... }`` is even a compiler bug according to [Swift evolution](https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md#relying-on-a-compiler-bug)

> I was thinking, since attributedSubtitleConfiguration is an optional argument, can the AttributedSubtitleConfiguration return an optional? In this way you don't need to re-bind self but just use the weak reference.

just double checking, did you mean `typealias AttributedSubtitleConfiguration = (_ attributedText: NSAttributedString) -> NSAttributedString?`? then going to replace it with `guard let domain = self?.domain else { return nil }`. I also like that this way the configuration block can specify not to alter the attributed subtitle by returning `nil`

jaclync:
> One thing I noticed is that the footer text on the register domain row is (i assume) sitting adjacent to the section heading on "Next Steps". We're using a typical spacing of 20pt in instances like this.

@SylvesterWilmott confirming on the spacing: if the next section has title (like QuickStart in the screenshot), is that there should be extra 20pt spacing between the end of Domain Credit footer and the beginning of QuickStart section (or any section with title)? since this PR is on success UI, I just added this issue to the [subtasks](https://github.com/wordpress-mobile/WordPress-iOS/issues/11467#issuecomment-485809805) and will get back to update the spacing

WordPress/Classes/ViewRelated/Blog/DomainCredit/DomainCreditRedemptionSuccessViewController.swift
```diff
@@ -0,0 +1,69 @@
...
+            return self?.applyDomainStyle(to: attributedText, domain: domain)
+        }
+        let controller = NoResultsViewController.controllerWith(title: NSLocalizedString("Congratulations", comment: "Title on domain credit redemption success screen"),
+                                                                buttonTitle: NSLocalizedString("Continue", comment: "Action title to dismiss domain credit redemption success screen"),
+                                                                subtitle: nil,
+                                                                attributedSubtitle: generateDomainDetailsAttributedString(domain: domain),
+                                                                attributedSubtitleConfiguration: attributedSubtitleConfiguration,
+                                                                image: "wp-illustration-domain-credit-success",
+                                                                subtitleImage: nil,
+                                                                accessoryView: nil)
```
ScoutHarris: These `nil` parameters when calling `NoResultsViewController.controllerWith` aren't necessary since they all have default values of `nil`.

WordPress/Classes/ViewRelated/Blog/DomainCredit/DomainCreditRedemptionSuccessViewController.swift
```diff
@@ -0,0 +1,69 @@
+import UIKit
+
+@objc protocol DomainCreditRedemptionSuccessViewControllerDelegate {
```
ScoutHarris: I'm wondering why the protocol is `@objc`. Since `continueButtonPressed` doesn't appear to be optional, would this be more Swifty? 
`protocol DomainCreditRedemptionSuccessViewControllerDelegate: class`

ScoutHarris:
Hey @jaclync . I left a couple minor comments, but otherwise it looks good. Well done!

WordPress/Classes/ViewRelated/Blog/DomainCredit/DomainCreditRedemptionSuccessViewController.swift
jaclync: yup not needed! just removed

WordPress/Classes/ViewRelated/Blog/DomainCredit/DomainCreditRedemptionSuccessViewController.swift
jaclync: Yes, looks more Swify without `@objc`. just updated!

jaclync:
@ScoutHarris thanks for the Swifty suggestions, just updated for another pass!

SylvesterWilmott:
>if the next section has title (like QuickStart in the screenshot), is that there should be extra 20pt spacing between the end of Domain Credit footer and the beginning of QuickStart section (or any section with title)?

@jaclync That's correct, if there's a title below then add 20pt spacing. However, feel free to add screenshots here of both states (with and without quick start enabled) just to confirm.

jaclync:
@SylvesterWilmott here's a side by side comparison:

next section without title | next section with title
--- | ---
Simulator Screen Shot - iPhone SE - 2019-04-30 at 18 05 49 | Simulator Screen Shot - iPhone SE - 2019-04-30 at 18 13 33

SylvesterWilmott:
@jaclync Thanks for the screenshots, we should only ad the 20pt space is there is a section with a title below it

PR #11573, opened 2019-04-26
Notifications: Mark as unread

frosty:
This PR adds the ability to mark notifications as unread. Swipe left-to-right on a read notification to mark unread.

mark-unread

**To test:**

* Build and run and go to your Notifications
* Swipe left-to-right on a read notification and check it becomes unread
* Ensure you can still swipe on unread notifications to mark them read
* Open the same WPCom account in a browser and check the state is reflected there. Note that you may need to reload the page to see your changes on the app take effect.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -691,6 +691,14 @@ private extension NotificationsViewController {
 
         NotificationSyncMediator()?.markAsRead(note)
     }
+
+    func markAsUnread(note: Notification) {
+        guard note.read else {
+            return
+        }
+
+        NotificationSyncMediator()?.markAsUnread(note)
+    }
 }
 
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1011,15 +1019,19 @@ extension NotificationsViewController: WPTableViewHandlerDelegate {
 //
 private extension NotificationsViewController {
     func leadingButtons(note: Notification) -> [MGSwipeButton] {
-        guard !note.read else {
-            return []
-        }
+        let isRead = note.read
+
```
danielebogo: What do you think if `Notification` can have a readonly property named _isRead_?In that way we can use `notice.isRead` to read the value without updating the data model.

frosty: I think that could make sense, although I only created this local variable because I thought it was slightly more readable in this case. I'll leave it for this PR, but it's something we could consider if we need it elsewhere :) Thanks!

PR #11575, opened 2019-04-26
Reader: Enforce max image size to prevent OOM errors due to gigantic images

aerych:
Yesterday I was trying to view a post in the reader and the app was being shut down and spring board restarted on my iPhone X.  Investigating, I discovered the images in the post were quite large, over 10,000x8,000 pixels which was this being scaled up to over 30,000x25,000 (x3 for the iPhone X retina display).  Memory pressure from trying to process the image was apparently yielding an OOM error.
Screen Shot 2019-04-26 at 9 07 38 AM
To resolve the issue, this patch limits the maximum size for an image shown in the reader. 

To test:
Use this post to test: paKIRI-1c-p2
- Confirm the bug on a device by viewing the post in the reader.  Liking the post then viewing your My Likes feed is an easy way.  You can also check the post in the simulator.  The app probably won't crash but the post will not scroll smoothly while the images are being loaded.
- Run the patch and confirm that: 
1. The app does not crash or get ejected from memory. 
2. Scrolling is smooth. 

@etoledom could I trouble you with this one? 

Update release notes:


aerych:
Thanks @etoledom !!

PR #11578, opened 2019-04-26
Improved accessibility of the steps in a Quick Start tour

stevebaranski:
Fixes #11492. 

The biggest change here was the addition of a new field on `QuickStartTour`. With that, I took the liberty to create hints in accordance with [our guidelines](https://github.com/wordpress-mobile/WordPress-iOS/wiki/VoiceOver-guideline), and set them appropriately.

To test:
- Checkout the branch, verify it builds, and that tests pass.
- Replicate the steps in the issue with VoiceOver, and verify that the desired accessibility behavior is observed.

Update release notes:

- [X] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #11579, opened 2019-04-26
Improved accessibility of Quick Start checklist header

### LINKED ISSUE #11493: Quick Start v2 VoiceOver: "Customize Your Site" completed cell state is not accessible
Steps to reproduce:

1. If you already have a site with Quick Start setup, go to that site, double tap on "Customize Your Site", and skip to step 7.
1. Log out of the WPiOS app.
1. Follow the prompts until you reach "Want a little help getting started?"
1. Double tap the "Yes, Help Me" button.
1. Try interacting with the "Completed (x)" cell on the "Customize Your Site" screen.

Result: the “Complete (x)” cell is not accessible and the completed cells state is not communicated by VoiceOver. ([7s](https://cloudup.com/cAdp2RYcvB5))

complete-cell-on-the-customize-your-site-screen

stevebaranski:
Fixes #11493. 

The `QuickStartChecklistHeader` is somewhat dynamic, so I devoted a bit of attention to updating the `accessibilityLabel` & `accessibilityHint` in conjunction with the view state.

_NB: It's probably best to review this in conjunction with #11578, as the same screen (and release notes entry) are impacted._

To test:
- Replicate the steps in the issue with VoiceOver, and verify that the desired accessibility behavior is observed.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
@@ -1,7 +1,6 @@
 import Gridicons
 
 class QuickStartChecklistHeader: UIView {
-
     var collapseListener: ((Bool) -> Void)?
     var collapse: Bool = false {
         didSet {
@@ -88,7 +87,7 @@ extension QuickStartChecklistHeader: Accessible {
 
         // From an accessibility perspective, this view is essentially monolithic, so we configure it accordingly
         isAccessibilityElement = true
-        accessibilityTraits = [ .header, .button ]
+        accessibilityTraits = [.header, .button]
 
         updateCollapseHeaderAccessibility()
     }
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
@@ -16,29 +16,13 @@ class QuickStartChecklistHeader: UIView {
                 self?.bottomStroke.alpha = CGFloat(alpha)
                 self?.chevronView.transform = CGAffineTransform(rotationAngle: rotate)
             })
-
-            let hintToken: String
-            if collapse {
-                hintToken = NSLocalizedString("Collapses", comment: "This token is used with ~ to format the accessibility hint to reflect the expanded and collapsed state of the list.")
-            } else {
-                hintToken = NSLocalizedString("Expands", comment: "This token is used with ~ to format the accessibility hint to reflect the expanded and collapsed state of the list.")
-            }
-
-            let accessibilityFormat = NSLocalizedString("%@ the list of completed tasks.", comment: "Accessibility hint for the list of completed tasks presented during Quick Start.")
-            let localizedAccessibilityHint = String(format: accessibilityFormat, hintToken)
-
-            accessibilityHint = localizedAccessibilityHint
+            updateCollapseHeaderAccessibility()
         }
     }
     var count: Int = 0 {
         didSet {
             titleLabel.text = String(format: Constant.title, count)
-
-            // Admittedly pluralization support could be improved here. See also: `paaHJt-c3-p2`
-            let accessibilityFormat = NSLocalizedString("%i completed tasks, toggling expands & contracts the list of these tasks", comment: "Accessibility description for the list of completed tasks presented during Quick Start.")
-            let localizedAccessibilityDescription = String(format: accessibilityFormat, arguments: [count])
-
-            accessibilityLabel = localizedAccessibilityDescription
+            updateCollapseHeaderAccessibility()
         }
     }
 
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
```diff
@@ -15,11 +16,29 @@ class QuickStartChecklistHeader: UIView {
...
             })
+
+            let hintToken: String
+            if collapse {
+                hintToken = NSLocalizedString("Collapses", comment: "This token is used with ~ to format the accessibility hint to reflect the expanded and collapsed state of the list.")
+            } else {
+                hintToken = NSLocalizedString("Expands", comment: "This token is used with ~ to format the accessibility hint to reflect the expanded and collapsed state of the list.")
+            }
+
+            let accessibilityFormat = NSLocalizedString("%@ the list of completed tasks.", comment: "Accessibility hint for the list of completed tasks presented during Quick Start.")
```
etoledom: In this case I believe that is better to localize the whole sentence. In other languages the order of words might be different, and the "Expands"/"Collapses" translation won't necessarily be always at the beginning.

WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
```diff
@@ -15,11 +16,29 @@ class QuickStartChecklistHeader: UIView {
...
+
+            accessibilityHint = localizedAccessibilityHint
         }
     }
     var count: Int = 0 {
         didSet {
             titleLabel.text = String(format: Constant.title, count)
+
+            // Admittedly pluralization support could be improved here. See also: `paaHJt-c3-p2`
+            let accessibilityFormat = NSLocalizedString("%i completed tasks, toggling expands & contracts the list of these tasks", comment: "Accessibility description for the list of completed tasks presented during Quick Start.")
```
etoledom: Since this control holds a state, I think it would be good to state its current state in the accessibility label.
Something like: "Expanded. 1 Completed task. Button. (hint) expans & collapse the list of completed tasks"

etoledom: It also might be helpful to let translators now that `%i` is a number (the count of completed tasks).

stevebaranski:
@etoledom thanks for the review, and for providing such helpful feedback. The app is better as a result. 😄 

I have pushed an additional commit that addresses your concerns. Because there are two different stimuli for the control (i.e., collapse & expand, as well as task completion), I created a new method that updates both labels in conjunction. It's now called when those values change, and via `awakeFromNib()`.

When you have a moment, would you be willing to take another look?

WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
```diff
@@ -1,6 +1,7 @@
 import Gridicons
 
 class QuickStartChecklistHeader: UIView {
+
```
nheagy: why the space here?

WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
```diff
@@ -60,8 +63,54 @@ class QuickStartChecklistHeader: UIView {
...
+        bottomStroke.isAccessibilityElement = false
+        contentView.isAccessibilityElement = false
+        titleLabel.isAccessibilityElement = false
+        chevronView.isAccessibilityElement = false
+
+        // Neither the top stroke nor the button (overlay) are outlets, so we configured them in the nib
+
+        // From an accessibility perspective, this view is essentially monolithic, so we configure it accordingly
+        isAccessibilityElement = true
+        accessibilityTraits = [ .header, .button ]
```
nheagy: We're not terribly consistent, but both our reference style guides have so spaces inside array list brackets: `[.header, .button]`

WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
```diff
@@ -60,8 +63,54 @@ class QuickStartChecklistHeader: UIView {
...
+
+        updateCollapseHeaderAccessibility()
+    }
+
+    func updateCollapseHeaderAccessibility() {
```
nheagy: I like the flow of this method.

WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
stevebaranski: I'll tidy this up.

WordPress/Classes/ViewRelated/Blog/QuickStartChecklistHeader.swift
stevebaranski: This was unintentional - I'll tidy this up.

stevebaranski:
Thanks for the input @nheagy - I pushed an additional commit that addresses those concerns.

@etoledom with @nheagy's approval and your reaction to my change comment, I'm going to infer that this is okay to merge. If I'm mistaken, please reach out, and I can address any concerns in a subsequent PR.

PR #11580, opened 2019-04-26
Improvements to signup epilogue accessibility

### LINKED ISSUE #11490: New Account VoiceOver: TextField cells could read out the current value when the label is selected
Steps to reproduce:

1. Log out of the WPiOS app.
1. Follow the prompts until you reach the "New Account" screen.
1. Tap each of the text field labels and then fill them out.

Result: when the label is selected, the current value is not read. ([26s](https://cloudup.com/cuJ4wNNHBQ7))

new-account-text-fields

stevebaranski:
Fixes #11490, via the following changes:

- Applied "extract method" refactoring* to the existing configuration of cell accessory view & text content type.
- Although there is a low likelihood of cell reuse on this table, implemented `prepareForReuse()` to reset existing cell configuration to "known, good" state.
- Disabled `isAccessibilityElement` of the cell's label, as its value is now read implicitly.
- Set `accessibilityTraits` of `username` cell configuration to reflect that it's a button. This is consistent with other parts of the app (e.g., `MyProfileViewController`) where the value is _not_ edited inline.
- Modified the cell's `accessibilityLabel` to read back contents of the label & the current text field value, as applicable (i.e., for `displayName` & `username` types, but _not_ the `password`).

In the future, it might be worthwhile to further specialize `SignupEpilogueCell` into separate cells for each particular type.

_*It can be argued that this is not really refactoring without tests to validate the behavior before & after the change._

To test:
- Replicate the steps in the issue with VoiceOver, and verify that the desired accessibility behavior is observed.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
+++ b/WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
@@ -31,6 +31,34 @@ class SignupEpilogueCell: UITableViewCell {
 
         accessoryType = .none
         cellField.textContentType = nil
+        isAccessibilityElement = false
+    }
+
+    override var accessibilityLabel: String? {
+        get {
+            let emptyValue = NSLocalizedString("Empty", comment: "Accessibility value presented in the signup epilogue for an empty value.")
+            let secureTextValue = NSLocalizedString("Secure text", comment: "Accessibility value presented in the signup epilogue for a password value.")
+
+            let labelValue = cellLabel.text ?? emptyValue
+
+            let fieldValue: String
+            if let cellText = cellField.text, !cellText.isEmpty {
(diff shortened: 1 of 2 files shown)


stevebaranski:
@etoledom thanks for the thorough review. I have pushed a few commits that I think tidy this up.

1. I reverted the bulk of the accessibility changes in the original PR.
1. Addressed the "old/new value updating" phenomena you observed. I failed to update the `accessibilityLabel` upon updates to the text field. The revisions simply override the `accessibilityLabel` of the cell.
1. Because the username field still transitions to another screen for editing (ala _My Profile_), we still need to conditionally update the a11y attributes in that particular case.

I believe the editing UX regression has been addressed with this changes. Can you double-check that on your end, please? Thank you!

PR #11581, opened 2019-04-26
Share Extension: Fix for posts published with local paths

### LINKED ISSUE #9118: Images sometimes keep local path when using the share extension
### Expected behavior

When using the share extension to publish a new blog post, the URL/path for any images should be updated to reference the media library.

### Actual behavior

We've had some reports of the path for images not being correctly updated and, instead, using the local path. For example:

``

The image is correctly uploaded to the media library but the path is not correctly updated within the post.

1. Open the Photos app.
2. Tap "Select" in top right corner.
3. Choose 2 images.
4. Tap share/action button.
5. Tap WordPress icon.
6. Fill in title/body.
7. Hit next and complete publish.
8. Look at the HTML for any images in your new post.

##### Reported in 274774-h. Tested on iPhone X, iOS 11.3, WPiOS 9.6.1

Neither @bummytime or myself have been able to replicate but making a note here for further investigation.

nheagy:
Fixes #9118

Sometimes the images in a post created with the share extension would keep their local paths once published. This PR should fix those problems.

This was a fun one to track down: It seems that when a media's upload is interupted the server may alter the file's name, such as turning "localfile.jpg" into "localfile-1.jpg". I assume this is related to having to reattempt the upload but I wasn't able to be sure of that. Regardless I was able to fix the issue with that assumption by making the filename comparison a bit more forgiving.

### To test:
Perform the text before and after this fix to experience the issue, and then ensure it has been addressed:
1. Build to device
2. Share two images with the WordPress share extension.
3. Immediately after tapping "Publish" turn off wifi (and LTE if you have it) on the device to interupt media upload
4. Wait a bit, then turn wifi back off.
5. Compare results:
   - Before the fix, the post should have broken media [like this](https://natesnotesandnovels.wordpress.com/2019/04/26/169/)
   - After the fix, embedded images should [appear correctly](https://natesnotesandnovels.wordpress.com/2019/04/26/201/)

Update release notes:


nheagy:
Thanks @bummytime! Seems like my solution may still be too fragile 😬 I'll take another look.

jkmassel:

nheagy:
Closing in favour of #11702

PR #11582, opened 2019-04-27
Issues/improve reader comments performance

aerych:
For hack week I wanted to try to improve the performance of the reader's comments view controller.  Scrolling the view seemed a little janky at times and when I profiled the controller the Time Profiler showed > 25% of time was spent building the attributed strings we show in each comment cell.  This seemed like a good candidate for improvement and led to this  PR. 

In this PR we do some refactoring to allow us to cache the attributed strings we show in the reader comments view controller rather than creating them on demand every time a cell is configured.   It was necessary to first implement a small refactor of the WPRichContentView to move logic specific to margin handling into the ReaderPostDetailController where its actually used.  Once done, ReaderCommentsViewController and ReaderCommentCell could be refactored to implement a new caching strategy.  Of note is the use of NSCache for storing attributed strings both for thread safety and for its memory handling behavior.

With these changes the Time Profiler is showing me time spent is down to ~15%. 

To test:
If you like, find a reader post with lots of comments and first examine its performance by profiling with the Time Profiler instrument. 
Switch to this patch.
Confirm reader comments load as expected. 
Scrolling should be smoother than it was previously tho this may be subjective and vary in its amount based on the comments being viewed. 
Compare its performance in the Time Profiler again and confirm that less time is spent building attributed strings. 

@stevebaranski Could I trouble you for a review of this one? 

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
@@ -38,6 +38,8 @@ class ReaderCommentCell: UITableViewCell {
 
     @objc var comment: Comment?
 
+    @objc var attributedString: NSAttributedString?
+
     @objc var showReply: Bool {
         get {
             if let comment = comment, let post = comment.post as? ReaderPost {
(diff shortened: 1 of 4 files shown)


aerych:
I agree Time Profiler can be tricky to get consistent numbers.  To mitigate variables I tried always using the same comment and attempting to scroll exactly the same amount.   There was still variance but the final numbers were consistently lower than what they had before.  At least for me 🤷‍♂. 

Thanks @stevebaranski!!!

PR #11584, opened 2019-04-29
Swift App Delegate 6: That's All, Folks!

frosty:
Following on from #11557, this PR completes the transition to an all-Swift app delegate. A few things to note:

* The aim for these PRs was to switch to Swift with minimal structural code changes, primarily to assist with review and testing. We can improve and refactor the code in the future, but if you have any smallish suggestions to roll into this PR then please let me know :)
* During the other piecemeal PRs I pulled out parts of the app delegate to an `WordPressAppDelegate+Swift` file. In this PR I've merged these extensions back into the `WordPressAppDelegate`, since the whole thing is now in swift.
* I know there will probably be a few additional tweaks to make to account for recent app delegate changes during hack week, such as the switch to Sentry. I'll have to wait until #11557 is merged until I can target `develop`.

**To test:**

* Review the code
* Build and run. Try and smoke test some of the major flows of the app, exercising things that are set up in the app delegate such as 3d touch shortcuts, reachability, notices, and logging
* Check all existing tests still pass

@astralbodies I think my testing setup is fine (as you recommended) – tests pass, etc, but could you double check that it looks ok?


jkmassel:

frosty:
@jkmassel Not a problem bumping it – it's going to take some time to get this mergeable due to changes to the app delegate since the PR was opened.

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -37,20 +317,20 @@ extension WordPressAppDelegate {
         })
     }
 
-    @objc func configureCrashLogging() {
+    func configureCrashLogging() {
```
jkmassel: @frosty:

This method a little out of date with develop – now we just call `WPCrashLogging.start()` inside of it. I've encapsulated the send-or-don't send logic in that class so that it's easier to implement in other projects.

Now that we're not doing the Obj-C => Swift thing, IMHO we could just eliminate the `configureCrashLogging` method and just call `WPCrashLogging.start()` directly?

frosty:
Thanks both for your reviews! Both good points, @stevebaranski – I'm going to respectfully leave them as-is for this PR, but I agree that grouping the objc methods may make sense as we continue to transition things.

@jkmassel I've updated the crash logging as you requested. Are we good to merge?

PR #11586, opened 2019-04-29
Notifications: Sharing and Reminders

frosty:
This PR adds two 'experimental' notifications features, that probably need some design review before they're ready to ship to everybody. Although as somebody whose daily work revolves a lot around notifications, I feel they might be useful to ship in internal builds straight away to gather feedback. 

Both features are accessible via a new `...` "More" item, visible if you swipe left-to-right on a notification:

Screenshot 2019-04-29 at 15 43 58

This is one thing I'd like feedback on – I'm not sure it's the best UI, but I don't know what would be better. Perhaps a long press? 

Select the `...` item to see options for each of the features:

Screenshot 2019-04-29 at 15 46 04

---

### Share

Share launches the system share sheet, sharing the notification's URL and subject. This is useful for sending a link to others, or storing in your todo app of choice:

note-shares

--- 

### Reminders

The other option, Reminders, presents a list of time periods after which you'll be notified again about the notification. I used the same periods that Slack uses: 20 minutes, 1 hour, 3 hours, tomorrow (9am), next week (Monday, 9am).

note-reminds

The notifications are iOS local notifications, so everything is specific to the device – you won't get a notification on a different device, or on the web. If you tap the system notification, you'll be taken to the detail view for the notification you've been reminded about.

When you have reminders scheduled, you can access them via a reminders icon in the top of the Notifications screen which only appears when you have some scheduled. This presents a list of all scheduled reminders. You can swipe them to cancel, or cancel them all with a single tap.

reminders-1

---

Let me know what you think! Is there anything that needs changing? Could we ship this internally for now?

**To test:**

* Build and run
* I added a few tests, so check those pass
* Try sharing some notifications
* Try setting reminders for some notifications and check they're triggered. If you want to make them trigger sooner than normal, you can alter `dateComponents(from date: Date)` in `NotificationReminderPeriod` in `NotificationRemindersHelper.swift`.

I haven't updated release notes yet, as this is still a draft and the functionality may change.

Update release notes:


iamthomasbishop:
This is _so_ great, I know I will find a lot of use for Reminders in particular. Even if this ends up having pretty narrow-impact, I think it's worth building/testing/monitoring internally :) With that said, I have a couple thoughts, take them with a grain of salt :)

### Swipe Actions

* I think adding the `•••` in the swipe actions works fine. However, we might want to move the `•••` action to last in the list of swipe actions, and perhaps change the `•••` icon to a `More` text label.

* Per your question re: long-press, I do think long-press should show _all_ possible cell actions in an action sheet, not just the ones within the `•••` menu.

### Share Flow

I think this flow makes sense, nothing standing out in terms of issues. I don't follow this use case very often, so others might have more feedback.

### Reminder Flow

The flow is pretty straight-forward for the most part. A couple things stand out:

* I'd expect to be able to take these actions from the detail view as well – maybe that's okay to leave out of scope for a quick v1?

* I would make the label on the `cancel reminder` cell swipe action read simply `cancel`. The action itself makes sense.

* I would change the `cancel` action on the Notification Reminders navbar to `Close`, as it's not really taking an action other than closing the view.

Additionally, but maybe out of scope here:

* We might want to consider adding the `Copy Link to Comment` action that the user can take from the comment detail.

* We might want to attach the same "remind me" action to the Reminders index, so I can "re-up" a reminder multiple times.

* I suggest we consider adding an accent dot or some sort of indicator to the `clock` icon when any number of my reminders are up or have passed, at least until I've seen them in my Reminders list. If this adds a bunch of work, then we consider this out of scope.

WordPress/Classes/Utility/NotificationRemindersHelper.swift
```diff
@@ -0,0 +1,276 @@
...
+
+    // MARK: - Managing Reminders
+
+    /// Logs the specified reminder ID, used when fetching and filtering
+    /// pending reminders. A reminder ID must be tracked using this
+    /// method for it to be returned by `getPendingReminders`.
+    ///
+    /// - parameter id: The identifier of a UNNotificationRequest
+    ///
+    func trackReminder(withId id: String) {
```
yaelirub: Is this function expected to be used outside of this helper class and getPendingReminders or can it be private?

WordPress/Classes/Utility/NotificationRemindersHelper.swift
```diff
@@ -0,0 +1,276 @@
...
+            return date.addingComponents(components)?.timeAndDateComponents
+        case .in1hour:
+            components.hour = 1
+            return date.addingComponents(components)?.timeAndDateComponents
+        case .in3hours:
+            components.hour = 3
+            return date.addingComponents(components)?.timeAndDateComponents
+        case .tomorrow:
+            components.day = 1
+            guard let tomorrow = date.addingComponents(components),
```
yaelirub: Do you think it will be useful to move `tomorrow` to the private Date extension below like: 
` var tomorrow: Date? {
        var components = DateComponents()
        components.day = 1
        return self.addingComponents(components)
    }`

Also maybe it would be useful to add those to a Date extension that other classes can use?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationRemindersViewController.swift
```diff
@@ -0,0 +1,153 @@
...
+            displayNoResults()
+        }
+    }
+
+    private func row(for request: UNNotificationRequest) -> ReminderRow {
+        let title = helper.reminderTitle(for: request)
+
+        let value: String
+        if let date = helper.reminderTriggerDate(for: request) {
+            value = formatter.string(from: date)
```
yaelirub: You can use `date.shortStringWithTime()` instead and so you don't even need to define the formatter

yaelirub:
Hey @frosty, I tested the reminders and sharing and it seems to work as expected. 
I got the reminders as notifications after 20 minutes.

IMG_571805A7454C-2

* I noticed that I could add a reminder to the same notification multiple time:
IMG_1D7B5DD94E97-1
and it made me think that maybe we should identify when a notification is already pending (the id is in the reminderIDs)and if it is, show an option to remove from reminders when bringing up the "reminder/share/cancel" action sheet? 

* I previously implemented an option to copy a comment url by long pressing on the date stamp label. If we do roll out this feature, I wonder if we should update the copy comment url to behave similarly (bringing up the share action sheet instead of copy to paste board)

* >I would change the cancel action on the Notification Reminders navbar to Close, as it's not really taking an action other than closing the view.
I agree with @iamthomasbishop .

WordPress/Classes/Services/NotificationSyncMediator.swift
```diff
@@ -159,15 +159,32 @@ class NotificationSyncMediator {
...
-        remote.updateReadStatus(noteID, read: true) { error in
+        remote.updateReadStatus(noteID, read: read) { error in
             if let error = error {
-                DDLogError("Error marking note as read: \(error)")
-                // Ideally, we'd want to revert to the unread status if this
+                let readState = read ? "read" : "unread"
+                DDLogError("Error marking note as \(readState): \(error)")
+                // Ideally, we'd want to revert to the previous status if this
                 // fails, but if the note is visible, the UI layer will keep
-                // trying to mark this as read and fail.
```
stevebaranski: Nice. Thanks for updating the docs.

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -331,6 +334,16 @@ class NotificationsViewController: UITableViewController, UIViewControllerRestor
             self.selectRow(for: note)
         }
     }
+
+
+    fileprivate func configureViewControllerPopover(_ controller: UIViewController, forNote note: Notification) {
```
stevebaranski: Nice touch for iPad! 
I believe this method could be `private`.

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1047,26 +1098,101 @@ private extension NotificationsViewController {
...
+        let controller = UIActivityViewController(activityItems: activityItems, applicationActivities: nil)
+        configureViewControllerPopover(controller, forNote: note)
+        present(controller, animated: true, completion: nil)
     }
 }
 
 
+// MARK: - Reminders
+//
+private extension NotificationsViewController {
```
stevebaranski: I admire your courage - `NotificationsViewController` is not for the faint of heart. 😄

ghost (commented):
Hi @frosty - thanks for proposing these changes. I think this feature could be quite useful!

I left a few comments on `p77Llu-c1p#13313-p2` for your consideration. The only other thing I'll add is this - should the swipable actions be gated by [`authorizationStatus`](https://developer.apple.com/documentation/usernotifications/unnotificationsettings/1648391-authorizationstatus)? Presumably this would also influence the visibility of the Reminders icon.

frosty:
Going to close this one out until I get chance to revisit it!

Issue #11189, opened 2019-03-05
[Stats Refresh] Transition to Detailed Statistics for a Selected Post (Post Stats)


stevebaranski:
As a user, I want to be able to view detailed statistics for a selected post (i.e., "Post Stats").

Post Stats can be accessed by either of the following actions:

1. Tapping _View More_ from _Insights : Latest Post Summary_
1. Selecting a given row from the _DWMY : Posts and Pages_ card

The Post Stats screen should include the following elements:

- Overview (including post / page title)
- Chart container, which presents total Views for a given post. When a given bar is tapped, it will also depict the change in an absolute & percentage basis from the preceding period.
- Historical data (i.e., "Months and Years"), as noted in https://github.com/wordpress-mobile/WordPress-iOS/projects/13#card-14080008
- Average views, as noted in https://github.com/wordpress-mobile/WordPress-iOS/projects/13#card-14080055
- Recent weeks, as noted in https://github.com/wordpress-mobile/WordPress-iOS/projects/13#card-14080356

These elements may be further decomposed into separate issues.

stevebaranski:
Closing this as Done based on conversation with @ScoutHarris.

PR #11589, opened 2019-04-30
[Stats Refresh] Post Stats: add Avg Views Per Day

ScoutHarris:
Ref #11189

This adds the 'Avg Views Per Day' card to the Post Stats view. 
- It shows average views for the last 6 years. 
- If there are more than 6 years, `View more` will be added.
  - NOTE: `View more` doesn't do anything just yet. Coming soon.
- Expanding a year shows the average views for each month in the year.

To test:
- Tip: en.blog > Home Page has views for several years.
- Access Post Stats via:
  - Insights > Latest Post Summary > View more
  - Period > Posts and Pages > row selection
  - Period > Posts and Pages > View more > row selection
- Verify 'Avg Views Per Day' appears, and each year expands to show all months in the year.

views_init

views_partial_year

views_full_year

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsViewModel.swift
@@ -53,6 +53,7 @@ class PostStatsViewModel: Observable {
         static let views = NSLocalizedString("Views", comment: "Label for number of views.")
         static let period = NSLocalizedString("Period", comment: "Label for date periods.")
         static let monthsAndYears = NSLocalizedString("Months and Years", comment: "Post Stats months and years header.")
+        static let averageViewsPerDay = NSLocalizedString("Avg. Views Per Day", comment: "Post Stats average views per day header.")
     }
 
     // MARK: - Init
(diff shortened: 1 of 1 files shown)


ScoutHarris:
@SylvesterWilmott - iPad views:

---
ipad_avg_init

---
ipad_avg_expanded

---
ipad_landscape

WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsViewModel.swift
```diff
@@ -130,39 +132,44 @@ private extension PostStatsViewModel {
         return tableRows
     }
 
-    func yearsTableRows() -> [ImmuTableRow] {
+    func yearsTableRows(forAverages: Bool = false) -> [ImmuTableRow] {
```
stevebaranski: Do you anticipate overloading this method further? If so, it might be an opportunity to add an `enum` here. If not, this approach also works.

ScoutHarris: > Do you anticipate overloading this method further?

Nope. This is the last card for Post Stats. Thanks!

PR #11594, opened 2019-05-01
Removing "Tap to retry"

### LINKED ISSUE #11432: Offline Support: Row showing actionable alert.
## Description:

1. Go to a post's settings screen.
2. Go offline.
3. Tap on "Tags".

You'll see that there's a row below the input field with the text "Couldn't load tags. Tap to retry".

As the message states, you can tap on that row to try to reload the contents.

This screen should probably both auto-refresh when a connection is detected, and offer a pull-down to refresh option.  This requires design approval though.

## Screenshots:

rowOfflineMessage

yaelirub:
Fixes #11432 

Expected: When unable to load tags the bottom label will show "Couldn't load tags."
To test:
1. Be offline 
2. Go to post from the post list
3. Go to post settings
4. Go to tags
5. see that the label shows  "Couldn't load tags."

BEFORE:
tags_label

AFTER:
tags_notice

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -12,6 +12,7 @@
 * Better URL validation when logging in with a self hosted site.
 * Account Settings Primary Site now shows the site URL if the site has no name.
 * Implemented incremental improvements to accessibility experience across the app.
+* Updated error message when tag loading failed.
 
 12.3
 -----
--- a/WordPress/Classes/Utility/WPError+Swift.swift
+++ b/WordPress/Classes/Utility/WPError+Swift.swift
@@ -4,6 +4,7 @@ import WordPressFlux
 
 extension WPError {
     private static let noticeTag: Notice.Tag = "WPError.Networking"
+    private static let noticeErrorTag: Notice.Tag = "WPError.anyError"
 
     /// Show a Notice with the message taken from the given `error`
     ///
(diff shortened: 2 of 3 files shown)


diegoreymendez:
@yaelirub - The way in which we show this message seems a bit non-standard to me.  Any thoughts on having this come up using the snack bar?

yaelirub:
@diegoreymendez , I implemented and it seems like the notice is not showing when the keyboard is displayed and the tags text view becomes first responder when the view appears.
@shiki , is there a known issue with notices displaying with keyboard

shiki:
@yaelirub It should be sufficiently handled by this:

https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift#L96

I wonder if it's a race condition. 🤔

shiki:
@yaelirub Would you mind sharing how you used `Notice` in here so I can take a look?

yaelirub:
@shiki , I'm using `WPError.showNetworkingNotice(title: title, error: error)` to show the notice. 
When I commented out the textview.becomeFirstResponder I do see the notice, but I think you're right and there might be a race condition there

shiki:
Thank you, @yaelirub, I created a separate issue for it. #11620.

shiki:
@diegoreymendez @yaelirub If you both think we should proceed with showing a Notice, I sent a PR for the Notice issue here: #11632.

jkmassel:

yaelirub:
@diegoreymendez, updated

yaelirub:
Thank you @shiki ! 🙏 
Updated.

Issue #11595, opened 2019-05-01
App crashes on login with password


pinarol:
### Expected behavior
No crash during login

### Actual behavior
App crashes with this message after entering password and tapping next.
```

2019-05-01 16:50:43.253635+0300 WordPress[49800:40116925] -[__SwiftValue encodeWithCoder:]: unrecognized selector sent to instance 0x600001e03f60
2019-05-01 16:50:43.253965+0300 WordPress[49800:40116925] *** -[NSKeyedArchiver dealloc]: warning: NSKeyedArchiver deallocated without having had -finishEncoding called on it.
2019-05-01 16:50:43.261098+0300 WordPress[49800:40116925] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__SwiftValue encodeWithCoder:]: unrecognized selector sent to instance 0x600001e03f60'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000116d7b6fb __exceptionPreprocess + 331
	1   libobjc.A.dylib                     0x00000001161aaac5 objc_exception_throw + 48
	2   CoreFoundation                      0x0000000116d99ab4 -[NSObject(NSObject) doesNotRecognizeSelector:] + 132
	3   CoreFoundation                      0x0000000116d80443 ___forwarding___ + 1443
	4   CoreFoundation                      0x0000000116d82238 _CF_forwarding_prep_0 + 120
	5   Foundation                          0x000000010e45b3d0 _encodeObject + 1230
	6   Foundation                          0x000000010e45a48b +[NSKeyedArchiver archivedDataWithRootObject:] + 156
	7   UIKitCore                           0x000000011e9770dc -[_UIAppearanceRecorder _recordInvocation:withClassName:containerClassNames:traitCollection:selectorString:forRemoteProcess:] + 3117
	8   UIKitCore                           0x000000011e970e22 __54+[_UIAppearance _recordersExcludingSource:withWindow:]_block_invoke + 907
	9   CoreFoundation                      0x0000000116deb9c2 -[__NSDictionaryM enumerateKeysAndObjectsWithOptions:usingBlock:] + 226
	10  UIKitCore                           0x000000011e970a85 +[_UIAppearance _recordersExcludingSource:withWindow:] + 131
	11  UIKitCore                           0x000000011f13a384 UIViewServiceCurrentAppearanceSerializedRepresentations + 87
[...]

PR #11596, opened 2019-05-01
Fix crash on login with password

pinarol:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/11595

To test:
Open login screen, enter your mail address, tap next
Choose enter your password instead
Enter your password and tap next
Application shouldn't crash

Update release notes:


pinarol:
Thank you @shiki !

Issue #10379, opened 2018-11-01
[Stats Refresh] Insights: show loading view


ScoutHarris:
On the Stats Insight view, show a loading view while the data is loading.

Currently, the Insights view is shown with no data:

![insights_loading](https://user-images.githubusercontent.com/1816888/47866208-b764d500-ddc3-11e8-9d0e-7eaccaa38cbb.png)

We should show the typical loading view. For example (from Activities):

![loading](https://user-images.githubusercontent.com/1816888/47866222-c186d380-ddc3-11e8-83b5-925c9d187be1.png)

PR #11597, opened 2019-05-01
[Stats Refresh] Insights: Loading View

danielebogo:
Refs. #10379

This PR adds a loading view for the Stats Insights section.

stats-insights-loading-view

## NoResultsViewHost
Because we often duplicate the code to add the NoResultsViewController as child, I thought to create a helper using a protocol called `NoResultsViewHost` (thanks @frosty for the name suggestion). Its extension constrained to a UIViewController contains the logic to configure, display and hide the noResultsViewController object. Also using associated objects allows us to implement the protocol without any other particular implementation.

## Error Handling
I will push another PR for the error handling. This because it requires some change on the Store.

## To test:
• From _My Sites_ select a site and open its dashboard
• Select _Stats_ to open the Stats section. Be sure the tab _Insights_ is selected.
• You should see the loading view
• The loading view should disappear as soon the data is loaded

Update release notes:


danielebogo:
Thanks @ScoutHarris !

SylvesterWilmott:
Looks good @danielebogo

PR #11603, opened 2019-05-01
[Stats Refresh] Post Stats: add date selector

ScoutHarris:
Ref #11189

This adds the date selector to the Post Stats view.
- When the date is changed, the Views count, difference, and percentage update on the Overview card.
- Back button goes back by one day.
- Forward button goes forward by one day.

Behavior inherited from the date selector (i.e. this is not new, just reiterating the behavior):
- Forward button is disabled if the next day is beyond the current day.
- Back button is disabled if the next day is beyond the number of days shown on the Overview chart, which is 13.

To test:

- Access Post Stats via:
  - Insights > Latest Post Summary > View more
  - Period > Posts and Pages > row selection
  - Period > Posts and Pages > View more > row selection
- Change the date back and forward.
- Verify it performs as expected, and the overview data updates.

post_stats

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsViewModel.swift
@@ -10,6 +10,7 @@ class PostStatsViewModel: Observable {
 
     let changeDispatcher = Dispatcher<Void>()
 
+    private var selectedDate = Date()
     private var postID: Int?
     private var postTitle: String?
     private var postURL: URL?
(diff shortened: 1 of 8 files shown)


ScoutHarris:
@SylvesterWilmott - iPad views:

---
ipad_port

---
ipad_land

WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsViewModel.swift
```diff
@@ -232,4 +240,32 @@ private extension PostStatsViewModel {
         return weekDateFormatter.string(from: day)
     }
 
+    // MARK: - Overview Helpers
+
+    func dayDataFrom(_ daysData: [StatsPostViews]) -> (viewCount: Int, difference: Int, percentage: Int) {
```
stevebaranski: Named tuple parameters: 👍

SylvesterWilmott:
Look good @ScoutHarris 

But are we missing the highlighted period on the chart? I also noticed the date selector is on May 1 and the chart ends at Feb 28 which should not be possible or is this just a mock?

PR #11610, opened 2019-05-02
Append locale as query parameter to TOS URL

### LINKED ISSUE #9824: TOS page linked in the app in not localised
The TOS page we link from the App (Me -> App Settings -> WordPress for iOS -> Terms of Service) is not localised. 
Since translated versions of the page exist for some languages, we should use those where possible.

ctarda:
Fix #9824 

As mentioned in [this comment](https://github.com/wordpress-mobile/WordPress-iOS/issues/9824#issuecomment-487056318), this PR passes the locale identifier as a parameter in the TOS URL. 

To test:
- Check out the branch
- Run the app on a device (or simulator) with the system set to a language other than English.
- Navigate to Me > App Settings > About WordPress for iOS > tap Terms of Service

@etoledom Since you reviewed the original code that is updated now, would you mind taking a look?
cc / @aforcier


jkmassel:
Merged this to get it into 12.4.

PR #11611, opened 2019-05-02
[Stats Refresh] No Results view when the Insights Overview fails

### LINKED ISSUE #10379: [Stats Refresh] Insights: show loading view

danielebogo:
Fixes #10379 

This PR adds the no results screen when the whole overview fails.

## Store
I followed the logic the `StatsInsightsStore` uses when the api service is fetching a particular overview section. I use different `Bool` value types, each for every overview section. Then the viewModel and the view controller will check only the `fetchingOverviewHasFailed` property.
Also `func tableViewModel() -> ImmuTable` will return always an empty `ImmuTable` if `fetchingOverviewHasFailed` will be true.

## To test:
• From My Sites select a site and open its dashboard
• Activate your _Network Link Conditioner_ or any other tool you want to use. The important thing is the response should be always **100% Loss**
• Select Stats to open the Stats section. Be sure the tab Insights is selected.
• You should see the loading view (if the tool you are using allows the service to try a connection) and then the no results screen with the _Retry_ button
• Disable your _Network Link Conditioner_
• Tap on Retry button and leave the service load the overview.
• Enable again your _Network Link Conditioner_ and pull down to refresh
• You should see the again the no results screen.

**Important:** **DO NOT SWITCH OFF** your network because the reachability will automatically display the no internet connection screen.

Update release notes:


danielebogo:
@SylvesterWilmott 

IMG_2995

IMG_2996

SylvesterWilmott:
Looks good!

ScoutHarris:
Hey @danielebogo .

As I mentioned in Slack -

@jklausa had made it so stats will show cached data if offline (https://github.com/wordpress-mobile/WordPress-iOS/pull/11295).

We should do two things:
1. Reinstate that logic (i.e. don’t show the connection error if offline - show cached data).
2. If fetching Insights fails, we should check for cached data. If there is some, use it. If not, _then_ show the error view.

I'll approve this PR, but we really should fix using the cache data prior to release.

ScoutHarris:
Hey @danielebogo .

Again from our Slack conversation - things that don't seem quite right.

Prereq: Let Insights load at least once so there is cached data.

1. Disconnect your network. The NRV shows with the no connection error. The cached data should show.
2. Again with disconnected network, still on Insights, pull to refresh. The error NRV shows with the Refresh button. To me, if we have cached data, we should show it.

danielebogo:
Hey @ScoutHarris ! I added a little improvement. Now the NRV with the retry button appears only if the store doesn't have any cached data, otherwise we will see the cached data displayed.

ScoutHarris:
Hey @danielebogo . Thank you for making those changes!

The only other issue is see is this:
1. Let Insights load.
2. Go back to the dashboard.
3. Turn off network.
4. Select Stats.
5. The connection error NRV shows instead of the cached data.

If you'd rather address that separately, I can approve this PR. Just let me know.

ScoutHarris:
Once again recapping a Slack conversation, @danielebogo & I agreed to commit this PR, and he'll address the [last issue](https://github.com/wordpress-mobile/WordPress-iOS/pull/11611#issuecomment-489208543) next week.

danielebogo:
Thanks @ScoutHarris !

Issue #10532, opened 2018-11-23
Domain Registration: Validation errors = endless spinner


rachelmcr:
While registering a new domain, I triggered a validation error by entering unacceptable characters in the postal code field. However, a spinner then appeared that I could not dismiss without navigating away from the screen and back again. This spinner appears with any validation error in the registration form.

I was able to type in the field even while the spinner was on the screen, but I still couldn’t do anything else (e.g. fix the validation error and then try to register the domain again).

![registerdomainspinner](https://user-images.githubusercontent.com/8658164/48938236-82522b00-ef08-11e8-8e49-ab839be4de9e.png)

rachelmcr:
Testing feedback from @thehenrybyrd:

> When triggering a validation error, I get a persistent spinner, and can type and navigate, but can’t fix the validation error, and the “register domain” button doesn’t re-load.

> In addition, after clicking “register domain” (with correct information), I got that same persistent spinner, and was able to navigate and type while registration was ongoing.

jklausa:
This was fixed in #10503, closing :)

PR #11613, opened 2019-05-02
Integrate domain registration flow to domain credit redemption

jaclync:
Refs. #11467 - subtask - Integrate the whole domain registration flow

Code changes:

* Integrated domain registration flow from domain credit redemption entry point.
* In `RegisterDomainDetailsViewController`, moved domain registration success callback from before view controller dismissal to dismissal's completion block. This change was made so that after domain registration UI is dismissed, new UI can be presented (otherwise presenting view controller is not visible in view hierarchy when success block is called).

To test:

- Prerequisites: have a site/blog with paid WordPress.com plan (e.g. Blogger plan) and its domain credit still available
- Go to a site dashboard where the site meets conditions in prerequisites, domain credit section should be shown
- Tap on "Register Domain" row
- Pick a `*.blog` domain (`.blog` has high purchase volume so that cancellation rate from testing has less effect on production domain service)
  - Note that the same domain cannot be purchased by the same user after cancellation, so make sure this testing domain is not what you might want to own with the WP.com account :)
- Fill out registration details (these will be saved for the next time)
- Submit domain registration, and wait for registration service to complete
  - One issue I noticed and fixed in this PR (same with Business account automated transfer): the spinner appears after submitting the form, and then disappears shortly while registration is still in progress (network spinner on).
  - An edge case crash fixed in this PR in https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/2de5bc062ca3ca600a007a5cb5c1884f40cc13be: while registration service is in progress, tapping "< Back" quickly and dismissing the domain registration modal used to crash due to retained `self`.
- Around the time an email on domain registration success is received, the domain registration modal should be dismissed and a congratulations screen is shown with the purchased domain
- After dismissing the congratulations screen, a snackbar should appear notifying further instructions in email. At the same time, domain credit UI should not be shown on site dashboard
- **[Important]** Cancel domain purchase shortly after testing in Account Profile --> Manage Purchases ([official FAQ](https://en.support.wordpress.com/manage-purchases/#FAQ-Cancelling))

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
@@ -18,10 +18,15 @@ extension BlogDetailsViewController {
     }
 
     @objc func showDomainCreditRedemption() {
-        // TODO-#11467 - subtask - integration with domain registration.
-        // Temporarily shows success screen before domain registration is integrated.
-        let domain = "lifeoftea.com"
-        presentDomainCreditRedemptionSuccess(domain: domain)
+        guard let site = JetpackSiteRef(blog: blog) else {
+            return
+        }
+        let controller = RegisterDomainSuggestionsViewController
+            .instance(site: site, domainPurchasedCallback: { [weak self] domain in
+                self?.presentDomainCreditRedemptionSuccess(domain: domain)
+            })
+        let navigationController = UINavigationController(rootViewController: controller)
+        present(navigationController, animated: true)
     }
 
     private func presentDomainCreditRedemptionSuccess(domain: String) {
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
@@ -26,7 +26,7 @@ extension BlogDetailsViewController {
             )
             let alertController = UIAlertController(
                 title: NSLocalizedString("Unable to register domain", comment: "Alert title when `JetpackSiteRef` cannot be initialized from a blog during domain credit redemption."),
-                message: NSLocalizedString("Something went wrong. Please try again.", comment: "Alert message when `JetpackSiteRef` cannot be initialized from a blog during domain credit redemption."),
+                message: NSLocalizedString("Something went wrong, please try again.", comment: "Alert message when `JetpackSiteRef` cannot be initialized from a blog during domain credit redemption."),
                 preferredStyle: .alert
             )
             alertController.addCancelActionWithTitle(cancelActionTitle, handler: nil)
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
@@ -19,6 +19,18 @@ extension BlogDetailsViewController {
 
     @objc func showDomainCreditRedemption() {
         guard let site = JetpackSiteRef(blog: blog) else {
+            DDLogError("Error: couldn't initialize `JetpackSiteRef` from blog with ID: \(blog.dotComID?.intValue ?? 0)")
+            let cancelActionTitle = NSLocalizedString(
+                "OK",
+                comment: "Title of an OK button. Pressing the button acknowledges and dismisses a prompt."
+            )
+            let alertController = UIAlertController(
+                title: NSLocalizedString("Unable to register domain", comment: "Alert title when `JetpackSiteRef` cannot be initialized from a blog during domain credit redemption."),
+                message: NSLocalizedString("Something went wrong. Please try again.", comment: "Alert message when `JetpackSiteRef` cannot be initialized from a blog during domain credit redemption."),
+                preferredStyle: .alert
+            )
+            alertController.addCancelActionWithTitle(cancelActionTitle, handler: nil)
+            present(alertController, animated: true, completion: nil)
             return
         }
         let controller = RegisterDomainSuggestionsViewController
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
@@ -1098,11 +1098,16 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 
     if (row.showsSelectionState) {
         self.restorableSelectedIndexPath = indexPath;
-    } else if (![self splitViewControllerIsHorizontallyCompact]) {
-        // Reselect the previous row
-        [tableView selectRowAtIndexPath:self.restorableSelectedIndexPath
-                               animated:YES
-                         scrollPosition:UITableViewScrollPositionNone];
+    } else {
+        if ([self splitViewControllerIsHorizontallyCompact]) {
+            // Deselect current row when not in split view layout
+            [tableView deselectRowAtIndexPath:indexPath animated:YES];
+        } else {
+            // Reselect the previous row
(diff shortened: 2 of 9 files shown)


WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
```diff
@@ -18,10 +18,15 @@ extension BlogDetailsViewController {
     }
 
     @objc func showDomainCreditRedemption() {
-        // TODO-#11467 - subtask - integration with domain registration.
-        // Temporarily shows success screen before domain registration is integrated.
-        let domain = "lifeoftea.com"
-        presentDomainCreditRedemptionSuccess(domain: domain)
+        guard let site = JetpackSiteRef(blog: blog) else {
+            return
```
jaclync: Is it worth logging an error here, or this case should not happen and `fatalError` is more suitable?

jaclync:
Regarding the spinner behavior I noticed during testing
> the spinner appears right after submitting the form, but then disappears shortly while registration is still in progress (network spinner on) so user is left on the registration form without actions nor activity indicator until registration completes and form is dismissed

I did more testing on domain registration via domain credit flow, and the spinner is hidden shortly after submitting registration form by https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Domains/Register/RegisterDomainDetails/ViewModel/RegisterDomainDetailsViewModel.swift#L488 - my guess is that this was to fix this issue https://github.com/wordpress-mobile/WordPress-iOS/issues/10532 in case of remote validation errors, and I think it could be moved to where response error occurs https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Domains/Register/RegisterDomainDetails/ViewModel/RegisterDomainDetailsViewModel.swift#L494 ? I just tried it with some bad phone number and the spinner is dismissed on remote validation errors, and stays on throughout registration

---
Another edge case I played around was to tap on navigation bar "Back" button all the way to dismiss domain registration flow while domain registration is still in progress. Then when registration completes, a crash happened on https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Domains/Register/RegisterDomainDetails/ViewController/RegisterDomainDetailsViewController.swift#L61 where `self` was already deallocated. Maybe while registration is in progress, navigation actions could be disabled? Or is it better for user to have a way out? (Then maybe `weak self` would help with the crash)

pinarol:
> I did more testing on domain registration via domain credit flow, and the spinner is hidden shortly after submitting registration

hey @jaclync 👋 I think this looks good overall but that spinner issue looks like sth we need to address.    As you already know currently the spinner is hidden even though there's some service call going on and I can hit the register button multiple times. 
One thing I can recommend is to use unittests on debugging this, otherwise it'll be really hard for you to register domain everytime you want to test this. You can refer to RegisterDomainDetailsViewModelTests, there are already some unittests there including some service calls, you can see there how service calls are mocked. That way you can try different success/fail combinations of service calls involved in this particular scenario and then maybe count loading events and assert the result. wdyt?

WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController+DomainCredit.swift
danielebogo: @pinarol thanks for reviewing it. I like more the second solution. Displaying an alert and log the error might be the best solution avoiding to add another case to handle the Domain Credit Cell logic.
```
I couldn't think of a valid case where JetpackSiteRef returns nil and at the same time canRedeemDomainCredit returns true.
```
I agree with this. But if it might happen then I prefer log the error and display the alert.

jaclync: Just updated with temporary copy in https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/a54f173335b8e7f938b1306e7101690e2a69e167, will update after the copy is confirmed.

In the same commit, also added code to deselect currently selected row so that "Register domain" row doesn't stay selected after error alert is dismissed. Deselecting currently selected row was not needed before for row with action that presents a modal, because `tableView.reloadData()` is called in `viewWillAppear` that is triggered on fullscreen modal dismissal. Dismissing an alert does not trigger `viewWillAppear`.

jaclync:
Hello @pinarol 👋 thanks for the pointers, the view model unit tests are super helpful! I updated the tests, service mock and interface in https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/5381db427f68e7761bd2979d46b0498458b188da and changed the loading state logic of view model in https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/8cd7190e9b961953fbcb84269f0707c4c3dc8136 (one test failed before this commit, as I expected from previous manual testing with breakpoints).

In https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/5381db427f68e7761bd2979d46b0498458b188da, I created an empty protocol `CartResponseProtocol` so that I could mock the original `CartResponse` in tests since its initializer and stored properties are all internal to `WordPressKit` framework. I tried not changing `WordPressKit` framework, and had to convert `CartResponseProtocol` back to `CartResponse` when calling a method in `WordPressKit`. Let me know if you have other suggestions!

jaclync:
Thanks to `RegisterDomainDetailsServiceProxyMock`, now I can test the UI flow without hitting remote domain registration endpoints by adding `WordPress` target membership to `RegisterDomainDetailsServiceProxyMock` and replacing existing remote service in `RegisterDomainDetailsViewModel` locally. While the registration is in progress, the register action button is disabled but I was still able to tap "< Back" and "Help" button (as shown in screenshot below) that could result in weird states if I didn't come back to the registration page when registration completes. Would it help with these edge cases if the background user interactions are not allowed, like by setting `SVProgressHUD` mask to [`.clear` or `.black`](https://github.com/SVProgressHUD/SVProgressHUD/blob/master/SVProgressHUD/SVProgressHUD.h#L28-L29) as also used in some cases in the app? (`SVProgressHUD.setDefaultMaskType(.black)` and `SVProgressHUD.setDefaultMaskType(.clear)`)

jaclync:
Added https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/2de5bc062ca3ca600a007a5cb5c1884f40cc13be to fix a reproducible crash on an edge case related to background user interactions while registration service is in progress. Steps to reproduce are in the PR description testing section under "Submit domain registration, and wait for registration service to complete"

pinarol:
> Would it help with these edge cases if the background user interactions are not allowed, like by setting SVProgressHUD mask to .clear or .black as also used in some cases in the app? (SVProgressHUD.setDefaultMaskType(.black) and SVProgressHUD.setDefaultMaskType(.clear))

great catch! yes I think we should disable background user interactions to avoid potential issues on this, currently user can change change the form input during the form validation and registration service calls, this is can be prone to weird issues.

jaclync:
yep sounds like a simple workaround to prevent edge cases from background user interactions. just updated with https://github.com/wordpress-mobile/WordPress-iOS/pull/11613/commits/181bb6804bde9f6290fc09423bee7d012d7ed923, tested it with mock service and now I couldn't navigate away from the form nor change the form anymore

danielebogo:
Hey @pinarol @jaclync what do you think if this problem about the HUD can be fixed in another PR?
Since @jaclync reported an issue already [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/11656) we could merge this and work on it in a separate one.
What do you think?

pinarol:
ok didn't know the deprecation of `showWithMaskType` so if `11/14 of the use cases are setting default to clear` then we probably don't have to worry about it. And knowing that we have a separate issue to check this later I think everything sounds cool.

Alright then @danielebogo @jaclync let's move on and :shipit: 🎉 

Thanks again @jaclync for all your efforts!

danielebogo:
Thanks @pinarol for reviewing this!

Issue #10846, opened 2019-01-17
Notice View doesn't adapt well to bigger font sizes.


etoledom:
### Expected behavior
Notice View adapts to bigger font sizes, and longer button names.

### Actual behavior
![simulator screen shot - iphone 5s - 2019-01-11 at 10 27 10](https://user-images.githubusercontent.com/9772967/51310585-42f71500-1a47-11e9-8070-1a837617b986.png)

PR #11614, opened 2019-05-02
Adapt NoticeView to bigger font sizes

pinarol:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/10846

**To test:**
Create a site and enter the QuickStart tour and see the message in the related issue ☝️ 

Another option could be mocking:

There's a `Revert fake notice` commit included in this PR, if that commit is reverted in the local environment a notice is seen on blog list page.
 
Update release notes:


WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
```diff
@@ -113,7 +114,6 @@ class NoticeView: UIView {
         labelStackView.alignment = .leading
         labelStackView.axis = .vertical
         labelStackView.spacing = Metrics.labelLineSpacing
-        labelStackView.isBaselineRelativeArrangement = true
```
pinarol: Removed this because it caused overlapping of title and message

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
```diff
@@ -260,7 +298,7 @@ class NoticeView: UIView {
     private enum Metrics {
         static let cornerRadius: CGFloat = 4.0
         static let dualLayoutMargins = UIEdgeInsets(top: 6.0, left: 6.0, bottom: 6.0, right: 6.0)
-        static let labelLineSpacing: CGFloat = 18.0
```
pinarol: needed to update this value after removing: `labelStackView.isBaselineRelativeArrangement = true`

WordPress/Classes/ViewRelated/System/Notices/NoticeStyle.swift
```diff
@@ -53,6 +53,18 @@ public struct QuickStartNoticeStyle: NoticeStyle {
...
 
     public let isDismissable = false
+
+    static func titleFont() -> UIFont {
+        let font = WPStyleGuide.fontForTextStyle(.subheadline, fontWeight: .semibold)
+        if #available(iOS 11.0, *) {
+            // setting adjustsFontForContentSizeCategory=true is not working with fonts with custom fontWeight
+            // so we are using UIFontMetrics to get it work
+            let metrics = UIFontMetrics(forTextStyle: .subheadline)
+            return metrics.scaledFont(for: font)
```
etoledom: It seems that we are scaling an already scaled font.
`font` has the proper pointSize for the "current" dynamic size. Then we ask the system to scale it again.
This strategy works well when the font given to the `UIFontMetrics` has (always) the proper size for the **default** dynamic size configuration (15 in the case of `subheadline`)

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
```diff
@@ -202,21 +215,46 @@ class NoticeView: UIView {
...
+        super.traitCollectionDidChange(previousTraitCollection)
+        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
+            preferredContentSizeDidChange()
+        }
+    }
+
+    private func preferredContentSizeDidChange() {
+        cancelButton.titleLabel?.font = notice.style.cancelButtonFont
+        actionButton.titleLabel?.font = notice.style.actionButtonFont
+        titleLabel.font = notice.style.titleLabelFont
```
etoledom: To set the font when the dynamic size changes works well when font instance also changes.
In this case It seems that the `style` field and the fonts that it holds are all immutable, so this should not affect (given that the fonts are set somewhere else).

The fonts themselves are not responding properly to content size changes though, so it could be beneficial to make `style` return a new font instance each time.

**Note:** It might be worth it to check `WPStyleGuide` fonts system, and see why they don't change dynamically. But that sounds as a small project of its own (for the future).

WordPress/Classes/ViewRelated/System/Notices/NoticeStyle.swift
etoledom: It seems like we are making the same mistake here:

https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/ccc3244543248c29243e682e49e8f75e11378190/WordPressShared/Core/Views/WPStyleGuide%2BDynamicType.swift#L76

😱

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
pinarol: Great catch @etoledom I have updated the style implementations to return new font instances everytime we access, that seems like solving all our problems.

>Note: It might be worth it to check WPStyleGuide fonts system, and see why they don't change dynamically. But that sounds as a small project of its own (for the future).

 I didn't quite understand this, actually WPStyleGuide creates a new font everytime.

WordPress/Classes/ViewRelated/System/Notices/NoticeStyle.swift
pinarol: I am removing this altogether, returning new font instances seems to solve our problems

pinarol:
Thank you @etoledom great catches! Could you give this another look?

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
etoledom: It should not be necessary to reset the font every time content size changes in cases where we just need one of the default size categories without custom font weight. They should adapt automatically given that the label's `adjustsFontForContentSizeCategory` is set to `true`. 

If I recall correctly, this was working properly at some point. Maybe we broke it when we added the maximum font size parameter 🤔

pinarol: Oh I see so you are saying, it should be working without creating a new instance.

WordPress/Classes/ViewRelated/System/Notices/NoticeStyle.swift
```diff
@@ -24,9 +24,10 @@ public protocol NoticeStyle {
 public struct NormalNoticeStyle: NoticeStyle {
     public let attributedMessage: NSAttributedString? = nil
 
-    public let titleLabelFont = UIFont.boldSystemFont(ofSize: 14.0)
-    public let messageLabelFont = UIFont.systemFont(ofSize: 14.0)
-    public let actionButtonFont: UIFont? = UIFont.systemFont(ofSize: 14.0, weight: .medium)
+    // Return new UIFont instance everytime in order to be responsive to accessibility font size changes
+    public var titleLabelFont: UIFont { return UIFont.boldSystemFont(ofSize: 14.0) }
+    public var messageLabelFont: UIFont { return UIFont.systemFont(ofSize: 14.0) }
+    public var actionButtonFont: UIFont? { return UIFont.systemFont(ofSize: 14.0, weight: .medium) }
```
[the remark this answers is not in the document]
pinarol: `TIL`

WordPress/Classes/ViewRelated/System/Notices/NoticeStyle.swift
```diff
@@ -41,10 +42,11 @@ public struct NormalNoticeStyle: NoticeStyle {
...
 
-    public let titleLabelFont = WPStyleGuide.fontForTextStyle(.subheadline, fontWeight: .semibold)
-    public let messageLabelFont = WPStyleGuide.fontForTextStyle(.subheadline)
-    public let actionButtonFont: UIFont? = WPStyleGuide.fontForTextStyle(.headline)
-    public let cancelButtonFont: UIFont? = WPStyleGuide.fontForTextStyle(.body)
+    // Return new UIFont instance everytime in order to be responsive to accessibility font size changes
+    public var titleLabelFont: UIFont { return WPStyleGuide.fontForTextStyle(.subheadline, fontWeight: .semibold) }
+    public var messageLabelFont: UIFont { return WPStyleGuide.fontForTextStyle(.subheadline) }
+    public var actionButtonFont: UIFont? { return WPStyleGuide.fontForTextStyle(.headline) }
+    public var cancelButtonFont: UIFont? { return WPStyleGuide.fontForTextStyle(.body) }
```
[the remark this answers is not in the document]
pinarol: > But dynamic sizes are not working properly now for some reason
That's not good, yes it looks like we'll need to fix WPStyleGuide+DynamicType

PR #11605, opened 2019-05-01
 Transitioned selected post chart to leverage actual data

stevebaranski:
Fixes #11068.

To test:

- Checkout the branch & confirm that existing tests pass.
- View stats for a selected post (1 of 3 ways).
- Observe the chart. There should be no aesthetic differences from #11173, but the chart should now be populated with live data (see below image).

Update release notes:

- [X] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

11068a

Issue #11061, opened 2019-02-15
[Stats Refresh] Insights : Construct Chart for Latest Post Summary (LPS)


stevebaranski:
The _Insights_ view specifies a Chart highlighting performance of a given site's last post, as measured by Views. The chart should allow the user to accomplish the following:

- Show a bar graph of views with available data.
- Adjust chart scale and dates relevant to data displayed

Because this is the first Charts-related work, this issue will include the following:

- Integration of the [`Charts`](https://github.com/danielgindi/Charts) framework within WPiOS via CocoaPods
- Creation of stubbed data based on Views of Site ID : `101549565`
- Definition of `BarChartConvertible` protocol suitable for passing data to a `BarChartView`
- Properly embedding the `BarChartView` in Insights, ensuring that it displays properly on both iPhone & iPad
- Configuring the appearance of the chart.

PR #11616, opened 2019-05-02
Transitioned Latest Post chart to leverage actual data

### LINKED ISSUE #11067: [Stats Refresh] Adapt Latest Post Chart to leverage persisted data
In general, we will be transitioning the [Latest Post chart](https://github.com/wordpress-mobile/WordPress-iOS/issues/11061) from using stubbed data to use real data, vended by the persistence layer @jklausa is working on.

stevebaranski:
Fixes #11067.

To test:
- Navigate to Stats for a given site. Observe the Latest Post Summary now reflects live data (i.e., similar aesthetics to #11061, up to 14 bars, not tappable).
- For the same site as in the preceding step, view stats for the same selected post (1 of 3 ways, as in #11605). Note the bars are still tappable (as in #11605), and the shape of the graph is the same.

Update release notes:

| Latest Post Chart | Selected Post Chart |
| - | - |
| 11067a | 11067b |


WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsViewModel.swift
```diff
@@ -98,6 +110,13 @@ class SiteStatsInsightsViewModel: Observable {
         ActionDispatcher.dispatch(InsightAction.refreshInsights)
     }
 
+    private func fetchStatsForInsightsLatestPost(postID: Int) {
```
ScoutHarris: Hey @stevebaranski . 

1. It doesn't appear the chart is updated when Insights in refreshed. I think just adding this to `refreshInsights()` will do it:
```
        if let postID = insightsStore.getLastPostInsight()?.postID {
            ActionDispatcher.dispatch(PeriodAction.refreshPostStats(postID: postID))
        }
```
2. This is so nitpicky, but if you move `fetchStatsForInsightsLatestPost` into the private extension, you won't need to mark it as `private`.

stevebaranski: Thanks for the feedback, @ScoutHarris. I have applied the changes you requested via an additional commit. I believe this is ready for another look.

ScoutHarris:
Thanks @stevebaranski . GTG! 

:shipit:

PR #11619, opened 2019-05-02
Reader: Disable text editing in reader views.

aerych:
Refs #11401 

While re-testing a different issue I noticed that text in the reader detail and reader comments was editable.  It looks like we missed that storyboards and xibs had flagged the WPRichContentView as not editable when refactoring.  This is a quick patch to disable editing again. 

To test:
Confirm the bug by running `develop` and tapping into a reader comment or detail and typing some text. 
Run this patch and repeat the test.  Confirm you do not see a cursor and can not edit text. 

@diegoreymendez could you take a quick peek at this one?


diegoreymendez:
Very good find @aerych.  On it.

aerych:
Thank you @diegoreymendez and @jkmassel !

PR #11622, opened 2019-05-02
[Stats Refresh] Post Stats: abbreviate Recent Weeks values.

ScoutHarris:
Ref #11189

This abbreviates the Recent Weeks values in Post Stats starting at 10,000.

To test:
- Go to Stats > Period > Posts and Pages.
- Select one with a lot of views.
  - Tip: en.blog Home Page is a good one.
- Verify the numbers are abbreviated.
  - NOTE: I don't have a site where a _day_ has high enough views to abbreviate, but it uses the same logic.

Before:

recent_weeks_no_abbrev

After:
recent_weeks_abbrev

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
@@ -53,7 +53,7 @@ extension Double {
         }
     }
 
-    func formatWithCommas() -> String {
+    private func formatWithCommas() -> String {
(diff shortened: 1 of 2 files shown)


ScoutHarris:
And I just notice Recent Weeks no longer has the column headers. I'll look at that separately.

ScoutHarris:
@stevebaranski 

> I no longer see any callers of formatWithCommas(); perhaps we could make this private?

Ah, good catch. It's been done. Thanks!

PR #11499, opened 2019-04-18
[Stats Refresh] Post Stats: show web view when post title tapped

ScoutHarris:
Fixes #11344 

When on Post Stats, tapping the post title will now open a web view for that post.

To test:
- Access Post Stats via:
  - Insights > Latest Post Summary > View more
  - Period > Posts and Pages > row selection
  - Period > Posts and Pages > View more > row selection
- Verify tapping in the post title area opens a web view for that post.

post_title

PR #11624, opened 2019-05-02
[Stats Refresh] Post Stats: fix Post Title display issues

### LINKED ISSUE #11618: [Stats Refresh] Post Stats: title UI issues
From @SylvesterWilmott :
- The title should be dark grey (same as the text above "Showing stats for:".
- We should use the serif font, `noto-serif`,
- The spacing between the labels should be same same if there is 1 line of title text or two lines.

single_line_title

multi_line_title

ScoutHarris:
Fixes #11618

This addresses the problems noted in #11618 regarding the Post Title displayed in the title card, namely:
- The Post Title is now dark grey
- The font is NotoSerif-Bold
- The spacing between the `Showing stats for` label and the Post Title is now consistent regardless of the post title length.

The problems stemmed from [this](https://github.com/wordpress-mobile/WordPress-iOS/pull/11499) change, where I got...creative, and made the Post Title look like that on Latest Post Summary since it's now clickable.

The fix was simply to remove the `postTitleLabel.attributedText`, thus allowing the original formatting to shine through.

To test:
- Go to Post Stats, and verify the three points noted above.

single_line_title

multi_line_title

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTitleCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTitleCell.swift
@@ -9,15 +9,14 @@ class PostStatsTitleCell: UITableViewCell, NibLoadable {
     @IBOutlet weak var bottomSeparatorLine: UIView!
 
     private typealias Style = WPStyleGuide.Stats
-    private var postTitle: String?
     private var postURL: URL?
     private weak var postStatsDelegate: PostStatsDelegate?
 
     // MARK: - Configure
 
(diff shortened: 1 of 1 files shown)


SylvesterWilmott:
Looks good!

Can you double check the spacing between the labels when there's 1/2 lines. In the screenshots above it looks like there's around a 2pt different in spacing between the two.

ScoutHarris:
Hey @SylvesterWilmott .

> Can you double check the spacing between the labels when there's 1/2 lines. In the screenshots above it looks like there's around a 2pt different in spacing between the two.

The space between the labels is 2pts. However, the one line Post Title label is a bit spacious - 

one_line

two_lines

Not sure why, but I'll create an issue to look at it.

PR #11632, opened 2019-05-03
Fix Notices sometimes shown behind the keyboard

### LINKED ISSUE #11620: Notice is shown behind the keyboard
Found in #11594. 

### Expected behavior

The error Notice will be visible on top of the keyboard.

### Actual behavior

The error Notice is hidden behind the keyboard. 

This can be easily reproduced using the Simulator. 

1. Add this line to the end of `PostTagPickerViewController.tagsFailedLoading`:
    ```swift
    WPError.showNetworkingNotice(title: "Error", error: error as NSError)    
    ```
2. Make sure your simulator is showing the soft keyboard
3. Edit a post from the post list
4. Navigate to the post settings
5. Go offline
6. Click on Tags
7. After a few seconds, hide the soft keyboard (CMD+K).

shiki:
Fixes #11620.

I found that in `NoticePresenter`, we do watch for keyboard notifications and move the **existing** Notice up when it's shown. However, we don't check if there is a keyboard already present when showing a new Notice. This adds that check.

| Before | After |
|--------|-------|
| notice-before gifcask 2019-05-03 13_26_47|  notice-after |

## Testing

This can be tested in, but not limited to, these scenarios: 

### Post List

1. Navigate to Site → Blog Posts
2. Tap on the search box to trigger the keyboard
3. Go offline. 

The offline Notice should be shown (eventually) on top of the keyboard. If not, pull to refresh. Please see _Testing Notes_ below. 

### Tags

1. Add this line to the end of `PostTagPickerViewController.tagsFailedLoading`:
    ```swift
    WPError.showNetworkingNotice(title: "Error", error: error as NSError)    
    ```
2.  Edit a post from the post list
3. Navigate to the post settings
4. Go offline
5. Click on Tags

The Notice should be shown on top of the keyboard. 

### Testing Notes

If you're using the simulator, play around with hiding and showing the keyboard (CMD+K). 

Please note that offline message Notices sometimes take a while to be shown. I noticed that on simulators, our API requests tend to wait for a long time before responding with a timeout. After that, the Notice is shown.

## Release Notes

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
+++ b/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
@@ -115,7 +115,7 @@ class NoticePresenter: NSObject {
             }
 
             UIView.animate(withDuration: durationValue.doubleValue, animations: {
-                currentContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant()
+                currentContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant
                 self.view.layoutIfNeeded()
             })
         }
@@ -130,7 +130,7 @@ class NoticePresenter: NSObject {
             }
 
             UIView.animate(withDuration: durationValue.doubleValue, animations: {
-                currentContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant()
+                currentContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant
                 self.view.layoutIfNeeded()
             })
         }
@@ -312,13 +312,13 @@ class NoticePresenter: NSObject {
             }
 
             noticeContainer.noticeView.alpha = WPAlphaFull
-            noticeContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant()
+            noticeContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant
 
             self.view.layoutIfNeeded()
         }
     }
 
-    private func onscreenNoticeContainerBottomConstraintConstant() -> CGFloat {
+    private var onscreenNoticeContainerBottomConstraintConstant: CGFloat {
         switch self.currentKeyboardPresentation {
         case .present(let keyboardHeight):
             return -keyboardHeight
--- a/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
+++ b/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
@@ -57,7 +57,7 @@ class NoticePresenter: NSObject {
     private var storeReceipt: Receipt?
 
     private var currentNoticePresentation: NoticePresentation?
-    private var currentKeyboardPresentation = KeyboardPresentation(isPresent: false, height: 0)
+    private var currentKeyboardPresentation: KeyboardPresentation = .notPresent
 
     private init(store: NoticeStore) {
         self.store = store
--- a/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
+++ b/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
@@ -37,6 +37,12 @@ class NoticePresenter: NSObject {
         let notice: Notice
         let containerView: NoticeContainerView?
     }
+    private struct KeyboardPresentation {
+        let isPresent: Bool
+        let height: CGFloat
+
+        static let notPresent = KeyboardPresentation(isPresent: false, height: 0)
+    }
 
     private let store: NoticeStore
     private let window: UntouchableWindow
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -294,12 +316,20 @@ class NoticePresenter: NSObject {
...
 
             noticeContainer.noticeView.alpha = WPAlphaFull
-            noticeContainer.bottomConstraint?.constant = -self.window.untouchableViewController.offsetOnscreen
+            noticeContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant()
 
             self.view.layoutIfNeeded()
         }
     }
 
+    private func onscreenNoticeContainerBottomConstraintConstant() -> CGFloat {
```
yaelirub: this method could be a computed property instead but what ever you prefer

yaelirub:
Hey @shiki , I tested and it looks good. The code also LGTM

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -37,6 +37,13 @@ class NoticePresenter: NSObject {
         let notice: Notice
         let containerView: NoticeContainerView?
     }
+    /// Used to determine if the keyboard is currently shown and what its height is.
+    private struct KeyboardPresentation {
```
nheagy: What do you think of an enum with two states:
```swift
case notPresent
case present(height: CGFloat)
```
?

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -96,21 +104,29 @@ class NoticePresenter: NSObject {
     private func listenToKeyboardEvents() {
         NotificationCenter.default.addObserver(forName: UIResponder.keyboardWillShowNotification, object: nil, queue: nil) { [weak self] (notification) in
             guard let self = self,
-                let currentContainer = self.currentNoticePresentation?.containerView,
                 let userInfo = notification.userInfo,
-                let keyboardFrameValue = userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue,
-                let durationValue = userInfo[UIResponder.keyboardAnimationDurationUserInfoKey] as? NSNumber else {
+                let keyboardFrameValue = userInfo[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue else {
```
nheagy: Is there a real case where `UIResponder.keyboardFrameEndUserInfoKey` is present but `UIResponder.keyboardAnimationDurationUserInfoKey` is not? Breaking up the guard makes this code a bit harder to reason about, so I want to be sure its necessary.

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
shiki: That's perfect! I'll add that.

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
shiki: I broke up the `guard` based on **when** the variable is used. In this case, I thought that adding the `durationValue` in here would make it confusing since it's not used until after the second `guard` below. But I guess doing it this way is still confusing so I'll change it so they're together. 😄

nheagy: I could be convinced otherwise if you feel strongly 😉

shiki: I'm not strongly attached to that decision so I'll change it. 🙂

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
shiki: Done in e619cb746e8ce443fc32ad562a9c6276bdbcc8cd.

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
shiki: Done in 90140076c9be37a6cf8f3df0cbfcc232da5af9fb.

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
shiki: Good idea, @yaelirub. Done in 230580a1a2ea0348ca65289c44c6e9b9739ac590.

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -109,10 +107,7 @@ class NoticePresenter: NSObject {
                     return
             }
 
-            self.currentKeyboardPresentation = KeyboardPresentation(
-                isPresent: true,
-                height: keyboardFrameValue.cgRectValue.size.height
-            )
+            self.currentKeyboardPresentation = .present(height: keyboardFrameValue.cgRectValue.size.height)
```
nheagy: Do we need `self.` here?

yaelirub: It's in the closure, so we need the self.

PR #11635, opened 2019-05-03
Improve Sentry tags

jkmassel:
- Simplifies calling crash logging
- Fixes Sentry environment tag
- Simplifies the release name
- Adds the locale as a Sentry tag
- Fixes `logged_in` and `connected_to_dotcom` variables to be correct

**To Test:**

See `2d0c52b8f4f642e5a3e75d53cd163d44-sentry` and note the fixed tags.


jkmassel:
@loremattei / @jtreanor was hoping to get this in before the code freeze, but I'm ok to do a new beta with it. 

Assuming there are no changes from here, this class will be copied over to Woo.

Issue #9282, opened 2018-05-07
Deprecate “simple” previews


koke:
We have a “simple” preview mode in the app that we use when we can’t show the actual server preview. It’s not very helpful now that we have a visual editor, and can be confusing for users.

There are a few steps before we can remove simple previews from the app:

1. [x] [#9285 Stop using simple previews as fallback](https://github.com/wordpress-mobile/WordPress-iOS/issues/9285)
2. [x] [#9286 Automatically save drafts before preview](https://github.com/wordpress-mobile/WordPress-iOS/issues/9286)
3. [x] [#9287 Implement autosaves to preview published posts](https://github.com/wordpress-mobile/WordPress-iOS/issues/9287)
4. [x] #11699 Remove `FakePreviewBuilder` from the app.

----

Original issue asking for feedback:

![simulator screen shot - iphone 8 - 2018-05-07 at 12 29 35](https://user-images.githubusercontent.com/8739/39698094-ac52c0ac-51f3-11e8-9fcf-21547ea96509.png)

There are a few cases where this is shown, and I’m not sure how we should address all of them:

1. When we can show a real preview in theory but it fails (offline, network/server error,…). This one is easier and I’d rather show an error with a retry button than fall back to a simple preview.
2. When the post is a local draft. To show a preview we would need to save the draft first.
3. When the post is published on the server, but we have local changes. There’s an `autosaves` endpoint on WordPress.com that would allow us to save a new revision without publishing, but this won’t work for WordPress.org sites.

There is some upcoming work in autosaves which could help with 2 and 3, and I think we can do 1 easily. But what do we do in the meantime, or for edits to published posts on WordPress.org sites? Do we still want to show this fake preview, or we’d rather not offer a preview and explain why?

folletto:
I think it's a good idea.

1. Yes: let's show a no connection message for this and a button to retry.
2. Yes: I'm not sure why we wouldn't save the draft on the server however. Why is that? Let's just save it, no?
3. Is this mostly due to a lack of API, or other reasons?

koke:
I had to dig through the archives for my notes on a previous attempt (in 2013!), I had some concerns back then: 

> I create a new post, preview it, and decide I don’t really want it. Tapping “Discard draft” should remove it from the server.

I don't think that as a problem anymore. You can even save drafts now without leaving the editor, so tapping preview shouldn't be much different.

[...]

PR #11638, opened 2019-05-04
Auto save published post upon preview and show preview url

### LINKED ISSUE #9287: Implement autosaves to preview published posts
When the user wants to preview a draft, we can just save a revision and preview it (#9286). However, when dealing with published posts saving would mean changing the published content and not just previewing it.

**Warning:** Note that "autosaves" here doesn't refer to the editor periodically saving changes, but core's idea of an [autosaved post revision](https://codex.wordpress.org/Revisions#Autosaves):

> Autosaves are stored as a special type of revision; they do not overwrite the actual post. In fact, whether your power goes out, your browser crashes, or you lose your internet connection, when you go back to edit that post, WP will toss up a warning telling you that it has a backup of your post and a link to restore the backup

I'm not sure if it's technically feasible, but what I know so far:
- There is a `/sites/$site/posts/$post/autosave` endpoint on WordPress.com that will return a preview URL for the changed content.
- Looking at an [old conversation in trac](https://core.trac.wordpress.org/ticket/21397) we might be able to create an autosave on self hosted by setting the `status` to `autosave` in `wp.editPost`, but this has to be tested.

If this is not possible on self hosted sites, I'd first make sure we're getting this on WP-API so we can do it whenever we add support. Also, I'd rather tell users that we can't preview edits to a published post rather than showing a fake preview.

yaelirub:
Fixes #9287  (also refer to #9282 )
[autosave API call PR on WordPressKit](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/155)

The gist:
When asking to preview a post we check if the post has unsaved changes (now also returning true if the post status is the new status: `AbstractPostRemoteStatusSaved`). 
If the post is not draft, we call the save endpoint that checks if the post blog supports WPRestAPI since this endpoint is not available for self hosted sites.
Not WPCom - we return failure with PostServiceErrorDomain error. and show a **noResultsViewController with "preview not available" - needs copy/ design approval (please see gif)**
Is WPCom - we call the autosave endpoint and get the previewURL on successful response.
We pass the previewURL to the preview generator to display.

WPCom site preview published post:

preview1

Self hosted site preview published post:
IMG_4064

To test:
 **Following code review and testing by developers, we have discovered an issue where the preview doesn't show the changes for a Jetpack site.**
A. preview post no updates 

1. go to a published post from your post list.
2. go to preview.
3. see the post preview with the latest content.

B. Preview post with updates
1. go to a published post from your post list
2. edit the post
3. preview the post
4. see "Generating preview..."
5. see the post preview with the latest content.

C. preview after discard
1. Steps in B
2. Exit the editor
2. Tap "Discard" when prompted in the action sheet
3. go back to the post
4. hit preview 
5. see the post preview with the latest (published) content.

* I will update the release notes once approved.


WordPress/Classes/Services/PostService.h
```diff
@@ -92,6 +92,21 @@ extern const NSUInteger PostServiceDefaultNumberToSync;
...
+ 
+ @param post The post or page to be saved
+ @param success A success block.  If the post object exists locally (in CoreData) when the upload
+ succeeds, then this block will also return a pointer to the updated local AbstractPost
+ object.  It's important to note this object may not be the same one as the `post` input
+ parameter, since if the input post was a revision, it will no longer exist once the upload
+ succeeds.
+ @param failure A failure block
+ */
+- (void)savePost:(AbstractPost *)post
```
aerych: Same question about naming that I had in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/156

WordPress/Classes/Services/PostService.m
```diff
@@ -271,6 +270,62 @@ - (void)uploadPost:(AbstractPost *)post
...
+
+    id<PostServiceRemote> remote = [self remoteForBlog:post.blog];
+    RemotePost *remotePost = [self remotePostWithPost:post];
+    NSManagedObjectID *postObjectID = post.objectID;
+    void (^successBlock)(RemotePost *post, NSString *previewURL) = ^(RemotePost *post, NSString *previewURL) {
+        [self.managedObjectContext performBlock:^{
+            AbstractPost *postInContext = (AbstractPost *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
+            if (postInContext) {
+                postInContext.remoteStatus = AbstractPostRemoteStatusSaved;
+                NSPredicate *unattachedMediaPredicate = [NSPredicate predicateWithFormat:@"postID <= 0"];
```
aerych: The unattached media handling is almost identical between the `savePost` and `updatePost` methods, the only difference is the parameters in the success blocks.   I wonder if we could refactor this to its own method to avoid the duplication?

WordPress/Classes/ViewRelated/Post/PostEditor+MoreOptions.swift
```diff
@@ -23,33 +23,68 @@ extension PostEditor where Self: UIViewController {
...
-                SVProgressHUD.dismiss()
-                }, failure: { error in
+    private func savePostBeforePreview(completion: @escaping ((String?, Error?) -> Void)) {
+        let context = ContextManager.sharedInstance().mainContext
+        let postService = PostService(managedObjectContext: context)
+        let draftStatus = NSLocalizedString("Saving...", comment: "Text displayed in HUD while a post is being saved as a draft.")
+        let publishedStatus = NSLocalizedString("Generating Preview...", comment: "Text displayed in HUD while a post is being saved.")
+        SVProgressHUD.setDefaultMaskType(.clear)
+
+        if post.hasUnsavedChanges() {
```
aerych: Nitpicky, but maybe this could be tweaked to avoid the nested conditionals?  Something like
```
if !post.hasUnsavedChanges() {
      completion(nil, nil)
      return
}
```

aerych:
Tested as described and worked for me!  I also tried with a post with the pending review status. No problems.

yaelirub:
@aerych , thank you so much for your review. I updated the PR based on your comments. Ready for another look.

yaelirub:
@megsfulton , moving towards removing "fake previews", we are now showing this screen when the user asks to preview a post from a self hosted site. Please your advice 🙏 
Simulator Screen Shot - iPhone Xs - 2019-05-08 at 21 59 59

aerych:
So... I'm not sure what to make of this.   When I was initially testing a self-hosted Jetpack connected site, I was not seeing a preview that included the changes. 

### Example 1 - Changes missing
autosave 2019-05-09 10_41_32
Note that the site here is a jurassic.ninja test site but its showing the same behavior that I saw on my own test site -- demonstrating that the behavior should be reproducible. 

But as I continue to test I do see that changes are showing up when previewing. 

### Example 2 - Changes present
autosave 2 2019-05-09 11_11_44

Example 2 shows my test site but now previews are working.  

I am quite perplexed.

megsfulton:
>moving towards removing "local previews", we are now showing this screen when the user asks to preview a post from a self hosted site. 


Some copy suggestions here to make it clear why the preview is unavailable: 

Title:
**Preview Unavailable for Self-Hosted Sites** 

Subtitle:
1) The updates to your post need to be published to be viewed.
or
2) To view the changes, publish the updates to your post. 

-

One other thing, in the gif you shared for "WPCom site preview published post:" -- I noticed "Privately Published" in the post list -- Just want to check that the "Privately Published" label there is unrelated to the preview changes in this issue and it's showing because the post you were updating was a private post?

diegoreymendez:
> Note that the site here is a jurassic.ninja test site but its showing the same behavior that I saw on my own test site -- demonstrating that the behavior should be reproducible.

But as I continue to test I do see that changes are showing up when previewing.

@aerych - Not sure if this is relevant but I recall seeing the same behavior using the web editor, where I had to preview twice for the autosave to be properly loaded.

yaelirub:
>Or if we want to be paranoid .jp sites could be treated like self-hosted sites.
I'm testing some more but think that might be what I'd end up doing.

yaelirub:
@megsfulton , thanks so much for your input. I have updated the view as suggested:
Simulator Screen Shot - iPhone Xs - 2019-05-09 at 14 52 38

> One other thing, in the gif you shared for "WPCom site preview published post:" -- I noticed "Privately Published" in the post list -- Just want to check that the "Privately Published" label there is unrelated to the preview changes in this issue and it's showing because the post you were updating was a private post?

The privately published label is unrelated. I set it as a private post.

aerych:
> I recall seeing the same behavior using the web editor, where I had to preview twice for the autosave to be properly loaded.

Potentially relevant.  Was it when previewing a jetpack site via the wpcom rest api by any chance? 

> I'm testing some more but think that might be what I'd end up doing.

I'm torn about this but I also don't have another answer.

diegoreymendez:
> Potentially relevant. Was it when previewing a jetpack site via the wpcom rest api by any chance?

In my case it was one of my WP.com test sites.

My feeling around this is that it's a problem with the autosaves API, or something similar... one possibility could be to record this as a known issue and split it from this PR.

It is my opinion that this PR already offers a much better experience than what we had before, even with this bug.

WordPress/Classes/Models/AbstractPost.h
```diff
@@ -13,6 +13,7 @@ typedef NS_ENUM(NSUInteger, AbstractPostRemoteStatus) {
     AbstractPostRemoteStatusLocal,       // Only local version
     AbstractPostRemoteStatusSync,       // Post uploaded
     AbstractPostRemoteStatusPushingMedia, // Push Media
+    AbstractPostRemoteStatusSaved,       // Post saved
```
diegoreymendez: If this is specifically refering to autosaved posts, can we rename it to something like `AbstractPostRemoteStatusAutoSaved`?

It's a minor difference, but "saving" can a bit vague and be confused with the regular uploading of the post, whereas "autosaving" has a more concrete meaning (at least in my head).

diegoreymendez: Also, if possible, update the comment to be "Post auto-saved" as well.

WordPress/Classes/Services/PostService+UnattachedMedia.swift
```diff
@@ -0,0 +1,22 @@
...
+        mediaToUpdate.forEach { media in
+            media.postID = post.postID
+        }
+
+        let mediaService = MediaService(managedObjectContext: self.managedObjectContext)
+        mediaService.updateMedia(mediaToUpdate, overallSuccess: {
+            ContextManager.sharedInstance().save(self.managedObjectContext)
+            success()
+        }) { error in
+            success()
```
diegoreymendez: Can you provide some background regarding why on failure we are reporting success?

The reason I'm asking is because if we decide we want to mute a specific error, it's usually best to leave that decision for the code that's calling this method, rather than to have this method automatically silence the error.

WordPress/Classes/Services/PostService.h
```diff
@@ -92,6 +92,21 @@ extern const NSUInteger PostServiceDefaultNumberToSync;
...
+ @param success A success block.  If the post object exists locally (in CoreData) when the upload
+ succeeds, then this block will also return a pointer to the updated local AbstractPost
+ object.  It's important to note this object may not be the same one as the `post` input
+ parameter, since if the input post was a revision, it will no longer exist once the upload
+ succeeds.
+ @param failure A failure block
+ */
+- (void)autoSave:(AbstractPost *)post
+         success:(nullable void (^)(AbstractPost *post, NSString *previewURL))success
+           failure:(void (^)(NSError * _Nullable error))failure;
```
diegoreymendez: Nitpicky: there seems to be an indentation issue with this line.

megsfulton:
Looking at this with fresh eyes today, I think this copy might be clearer for that screen:

Preview Unavailable for Published Posts
Update the published post to view your changes.

yaelirub:
Thanks @megsfulton , updated:
Simulator Screen Shot - iPhone Xs - 2019-05-10 at 16 10 12

WordPress/Classes/Services/PostService.h
yaelirub: Great catch!

WordPress/Classes/Services/PostService+UnattachedMedia.swift
yaelirub: I deleted the comment that explained this when I refactored the method:

'// Sergio Estevao (2018-02-27): even if media fails to attach we are answering with success because the post uploaded successful and the only thing that failed was attaching the media to it.'
I updated the callback to `completion` instead of `success` as Eric suggested but realize that you are right and it would be better handled by the caller so I added a failure block, logging the failure and added back the comment.

yaelirub:
@diegoreymendez, updated the code according to your comments. 
>My feeling around this is that it's a problem with the autosaves API, or something similar... one possibility could be to record this as a known issue and split it from this PR.

I can create a ticket for this. Do you agree, @aerych ?

aerych:
> I can create a ticket for this

Sounds reasonable.  :+1:

aerych:
Hmm. I marked approved but maybe I should have dismissed my earlier review instead? 🤔 
Regardless, its 👍 from me with Diego's recommended changes + the conflicts resolved.

PR #11654, opened 2019-05-08
[Stats Loading] Display cached data if there's no connection

danielebogo:
This PR removes the NRV from the `StatsViewController` if there's no connection only if the feature flag is enabled. Cached data will be displayed instead (if it does exist).

## To test:
• Install a fresh version of the App
• Put the phone in airplane mode 
• Select stats. You should see the NRV (no connection)
• Tap back and switch off the airplane mode
• Select Stats and leave the insights load the data
• Tap back and put the phone in airplane mode
• Select Stats and you should see the cached data

Update release notes:


WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
```diff
@@ -236,6 +236,11 @@ import Reachability
     /// Public method to hide/show the image view.
     ///
     @objc func hideImageView(_ hide: Bool = true) {
+        if let isReachable = reachability?.isReachable(),
```
ScoutHarris: Hey @danielebogo . I'm a little concerned about putting this here. `hideImageView` is a public method that is called to essentially force the hide the image. A cursory search shows this is typically done based on device orientation, especially for iPhone landscape where there isn't a lot of space. On the views that expect the image to be hidden, it will instead be shown for no connection, which probably won't look very nice. 

Thoughts?

danielebogo: Ah! Good catch! Yes in that case I have to remove that.
I have a question about this. In `func configureView()`, line 311, can the `imageView.image` be nil by default?

danielebogo: Nevermind @ScoutHarris ! I came up with another solution!

danielebogo: I fixed it in [7ae9a3f](https://github.com/wordpress-mobile/WordPress-iOS/pull/11654/commits/7ae9a3f1fc893fe1a2c021b97b9f110722387b3d).
You were right and that check was out of scope.

Issue #11360, opened 2019-03-28
[Stats Refresh] Investigate Stats parsing performance


jklausa:
During testing, I've found some operations can take painfully long time to finish in the new Stats architecture — e.g. viewing a list of `referres` for a whole year takes _ages_.

Fetching `summary` also takes very long time, as mentioned by @ScoutHarris.

We need to profile this and fix the issues.

jklausa:
At first glance it seems it's not due to parsing at all, but rather UIKit is doing something really bizarre...

Note the runtime in _minutes_!!!

Like it's trying to display millions of row at the same time?
<img width="1412" alt="Screen Shot 2019-03-28 at 04 28 17" src="https://user-images.githubusercontent.com/1594081/55127975-34e20680-5112-11e9-9ec2-7fd657d023cf.png">

jklausa:
(to be fair, the runtime is much less dramatic when not being instrumented... but still.)

jklausa:
So I think I know what is going on here.

@ScoutHarris I think there's some work required on how the details screen are displayed.

Right now it's just a single `UITableViewCell` (`TopTotalsCell`) that has a `UIStackView` that contains all the items in it. This works fine for a small amount of objects, but breaks down rather... badly for lists with HUGE count of items in it. Displaying "Clicks" for the past year for en.blog causes the cell to grow to almost `40000`pt in height. 

Turns out, iOS doesn't handle trying to render bitmaps this large well :D

I think we need to make this use actual `UITableViewCell`s, instead of the "fake" ones in the `UIStackView`.

![](https://user-images.githubusercontent.com/1594081/55128535-37ddf680-5114-11e9-96b5-c6899f857e42.png)

![](https://user-images.githubusercontent.com/1594081/55128543-40363180-5114-11e9-9d3d-a092c0a582d6.png)

ScoutHarris:
Created https://github.com/wordpress-mobile/WordPress-iOS/issues/11657 for the `summary`. I'll address the StackViews with a series of PRs.

ScoutHarris:
All details views have been fixed on referenced PRs.

PR #11662, opened 2019-05-09
[Stats Refresh] Period Search Terms: fix displaying details view

ScoutHarris:
Ref #11360

This changes Period > Search Terms to use `UITableViewCell`s instead of one `UIStackView`.

Previously, all the data was stuffed into the stack view on the `TopTotalsCell` card. Now it only uses the `TopTotalsCell` card to show the "header" via `DetailSubtitlesHeaderRow`. The stack view is left empty. Instead, a `UITableViewCell` is added to the table for each data row via `q`. 

There should be no visual or functional changes, but showing the details view is _a lot_ faster. To note, there is still no loading view, so there may still be a blank view for a couple seconds for longer periods.

To test:
- Go to Stats > Period > Search Terms on a very active site (en.blog Years was particularly painful).
- Select `View more`.
- Verify selecting a row does nothing.

Update release notes:

search_terms_details


ScoutHarris:
@danielebogo this is based off of https://github.com/wordpress-mobile/WordPress-iOS/pull/11660. Please review that one first. Thank you!

PR #11663, opened 2019-05-09
Updating upload post method failure UI to show notice instead of a HUD

### LINKED ISSUE #11421: Offline Support: UI Inconsistency when Updating a Post
When editing a published post, pressing the Update button (top right) stops you with a HUD that it failed saving (“Error occurred during saving”). However, pressing X (top left) and tapping on “Update Post” in the action sheet seems to save the post.

yaelirub:
For the sake of consistency like we do upon failure to upload when the user choose to update after closing the editor.

Fixes #11421

To test:
1. update published post
2. be offline
3. tap "Update" from the top right in the navigation
4. See notice message on failure to upload 

**Before**
updating_ui_before_1

**After**
updating_ui

* Also updated strongSelf to self in this PR to fit  > Swift 4.0 

** Will update Release notes before merging 

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -1,4 +1,5 @@
 import Foundation
+import WordPressFlux
 
 extension PostEditor where Self: UIViewController {
 
@@ -318,10 +319,10 @@ extension PostEditor where Self: UIViewController {
         postEditorStateContext.updated(isBeingPublished: true)
 
         uploadPost() { [weak self] uploadedPost, error in
-            guard let strongSelf = self else {
+            guard let self = self else {
                 return
             }
-            strongSelf.postEditorStateContext.updated(isBeingPublished: false)
+            self.postEditorStateContext.updated(isBeingPublished: false)
             SVProgressHUD.dismiss()
 
             let generator = UINotificationFeedbackGenerator()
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -318,30 +319,31 @@ extension PostEditor where Self: UIViewController {
...
             let generator = UINotificationFeedbackGenerator()
             generator.prepare()
 
             if let error = error {
                 DDLogError("Error publishing post: \(error.localizedDescription)")
-
-                SVProgressHUD.showDismissibleError(withStatus: action.publishingErrorLabel)
+                SVProgressHUD.dismiss()
+                let model = PostNoticeViewModel(post: self.post)
+                ActionDispatcher.dispatch(NoticeAction.post(model.notice))
```
shiki: ⚠️ This `Notice` shows a _Retry_ button which changes the workflow. Should we consider using the previous message with no _Retry_ button instead? 

| Before | After |
|--------|-------|
|  2019-05-10 08 08 31 gifcask 2019-05-10 08_13_11 |     |

yaelirub: Thanks. I think you're right. There's no need for an update to the workflow here with the "retry"
I removed it and now showing informative notice.
update_error_message

yaelirub:
@shiki  thank you for being so thorough here. I really appreciate it.
I updated the PR to clear the notice when exiting the editor. I could add `ActionDispatcher.dispatch(NoticeAction.clearWithTag(uploadFailureNoticeTag))` to each one of the editors on `displayMoreSheet` but I think I'll leave it out of this PR.
It seems like a more generic solution with the notices might be in place here since the same thing happens when we're displaying any other notice (like no connectivity one) 

update_

Issue #9618, opened 2018-06-19
Login: Error says log in with username instead of email


aerych:
In certain (probably rare) circumstances, a user logging into wpcom with their email address might encounter an error prompting them to "log in using your WordPress.com username instead of your email address".  Since we don't allow a username login in the wpcom flow, its not obvious to a user what to do next.  

We should be checking for `email_login_not_allowed` for the type of error and perhaps send the user to the self-hosted flow with the site set to "wordpress.com".

aerych:
Related Android issue https://github.com/wordpress-mobile/WordPress-Android/issues/7922

pinarol:
Here is how I reproduced it maybe it might help on testing. I changed my email address in my WP.com account settings and when try to login with my old email I get this error all the time.

mindgraffiti:
More background in comments p77Llu-b5Q-p2

SiobhyB:
This caused some confusion again in 1585388-z.

aerych:
Candidate fix in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/42

aerych:
Fixed in wordpress-mobile/WordPressAuthenticator-iOS#42

PR #11665, opened 2019-05-09
Testing: Enable UI LoginTests to pass in any language

rachelmcr:
Addresses https://github.com/wordpress-mobile/WordPress-iOS/issues/11710 and https://github.com/wordpress-mobile/WordPress-iOS/issues/11759 for the login tests.

Currently, the UI tests only pass when the device UI is set to English. This adds some new accessibility identifiers to the app and updates the login tests so they pass in any language, and confirms we log in to the expected account/site.

To test:

* Run `rake mocks` to start the mock server.
* Open the simulator and set the device to a non-English language supported by the app (e.g. Spanish or Arabic).
* Run the `LoginTests` test suite in the `WordPressUITests` target.

Update release notes:


yaelirub:
Hey @rachelmcr ,
I ran the LoginTests suite and all the tests but testSelfHostedUsernamePasswordLoginLogout are failing and the UI:
Simulator Screen Shot - iPhone Xs - 2019-05-09 at 17 52 03

* tested this with a simulator set to Hebrew.
Simulator Screen Shot - iPhone Xs - 2019-05-09 at 17 50 27

JavonDavis:
@yaelirub for additional information it's a backend issue that was raised in [this Android PR](https://github.com/wordpress-mobile/WordPress-Android/issues/7922) and [iOS PR](https://github.com/wordpress-mobile/WordPress-iOS/issues/9618)... They're closed because a fix was implemented to redirect around them but I'm not sure that was implemented on iOS?

yaelirub:
Thanks @JavonDavis! The workaround on iOS was to direct the users to the email/password screen when they dismiss the alert and the UI actually shows that. 
While I think it's worth adding a test for that case, I updated the credentials file to a different user. Now the test continue pass the login and seems like they should pass but they still fail for some reason (?!)
login_test2
Also, @rachelmcr , do the tests cover a login with 2FA?

rachelmcr:
Thank you, @JavonDavis and @yaelirub, and sorry for the trouble you had! Once again, I forgot about that security error showing up when other people run these tests.

I'm going to put this PR on hold for the moment. We're planning to mock the API calls for these tests, which will avoid that security error without having to add a workaround for it into the tests.

@yaelirub I think I know why that tests fail for you on the epilogue screen when you run them with another test account. There's a check for the correct account/site info on that screen, but it only works for accounts with a single site; I'll work on a fix for that.

> do the tests cover a login with 2FA?

Not at the moment. Once we have mocking in place, we can consider adding that coverage.

> Also, we're using element(boundBy:) and then comments in a few areas here and I'm just wondering if that's the best approach... 

Thanks for raising that! As we discussed in Slack, we'll need to use those selectors for system UI (system alerts and text edit menus) but I'll look at using localized text for other alerts.

jkmassel:
I'm removing the milestone here, because this task isn't really tied to any specific release :)

rachelmcr:
This is ready for review again. With network mocking now in place (see updated testing steps in the PR description), we can avoid the backend login error some people were seeing when running the login tests.

PR #11668, opened 2019-05-09
[Stats Refresh] Period Videos: fix displaying details view

ScoutHarris:
Ref #11360

This changes Period > Videos to use `UITableViewCell`s instead of one `UIStackView`.

Previously, all the data was stuffed into the stack view on the `TopTotalsCell` card. Now it only uses the `TopTotalsCell` card to show the header via `DetailSubtitlesHeaderRow`. The stack view is left empty. Instead, a `UITableViewCell` is added to the table for each data row via `DetailDataRow`. 

There should be no visual or functional changes, but showing the details view is a lot faster. To note, there is still no loading view, so there may still be a blank view for a couple seconds for longer periods.

To test:
- Go to Stats > Period > Videos on a very active site.
- Select `View more`.
- Verify selecting a row displays the media.

Update release notes:

videos_details


ScoutHarris:
Hey @danielebogo .

> I don't think is related to this PR.

You are correct, that is behavior from `MediaItemViewController`.

PR #11669, opened 2019-05-09
[Stats Refresh] Period Countries: fix displaying details view

ScoutHarris:
Ref #11360

This changes Period > Countries to use `UITableViewCell`s instead of one `UIStackView`.

Previously, all the data was stuffed into the stack view on the `CountriesCell` card. Now it only uses the `CountriesCell` card to show the header via the new `DetailSubtitlesCountriesHeaderRow`. The stack view is left empty. Instead, a `UITableViewCell` is added to the table for each data row via `DetailDataRow`. 

There should be no visual or functional changes, but showing the details view is a lot faster. To note, there is still no loading view, so there may still be a blank view for a couple seconds for longer periods.

To test:
- Go to Stats > Period > Countries on a very active site.
- Select `View more`.
- Verify selecting a row does nothing.

Update release notes:

countries_details

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/CountriesCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/CountriesCell.swift
@@ -63,7 +63,7 @@ private extension CountriesCell {
 
     func setSubtitleVisibility() {
 
-        guard !forDetails else {
+        if forDetails {
             subtitleStackView.isHidden = false
             rowsStackView.isHidden = true
             return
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/CountriesCell.swift
```diff
@@ -58,6 +62,13 @@ private extension CountriesCell {
     }
 
     func setSubtitleVisibility() {
+
+        guard !forDetails else {
```
danielebogo: I think avoid the double negative here is more clear to read.
```
if forDetails {
.....
return
}
```
What do you think?

ScoutHarris: Sure. It does read better. I've changed that one. And since I had done the exact same thing in `TopTotalsCell`, I went ahead & changed that one too.

ScoutHarris:
@danielebogo ready for another pass please.

Issue #11671, opened 2019-05-10
[Stats Refresh] Coordinate navigation between the date bar & the Post Stats & Period Chart


stevebaranski:
This issue harvests some notes from the [Master Issue](#10840) with regard to the charts, as seen in both the Post Stats & Period charts.

Namely those charts should exhibit the following behavior:

-  When a bar is selected, update overview data accordingly.
-  When the view first appears, auto select the most recent day.
- When the date changes via the date chooser, auto select corresponding bar.

This does not apply to the chart on the Insights view for the _Latest Post Summary_, as that chart cannot be tapped to highlight individual metrics.

The initial project scoping for Charts functionality was conducted in early February, and the data bar was introduced in late January (via `p9xubY-uh-p2`). The above behavior has already been added to the Android implementation.

PR #11210, opened 2019-03-07
Preliminary implementation of Period Chart

stevebaranski:
### Description
Fixes #11063, which calls for the introduction of chart styling for period data, with an emphasis on the overlay of Visitors / Views data. 

To test:

- Checkout the branch, confirm that it builds & tests pass.
- For a site, navigate to Stats, and select a given period (i.e., Days, Weeks, Months, Years). 
- Please note that at the moment, the chart is the same regardless of period or dimension (Views, Visitors, Likes, Comments).

Update release notes:

- [X] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

### Exhibits

11063pr

_iPad - Portrait_
11063pr_ipadportrait

_iPad - Landscape_
11063pr_ipadlandscape

_iPhone SE_
11063pr_iphonese

Issue #11143, opened 2019-02-25
[Stats Refresh] Correct number formatting on Chart axes


stevebaranski:
With #11061, we constructed a chart with some stub data to match Zeplin. The formatting here is *not* currently locale-sensitive, nor does it match the formatting applied [elsewhere](https://github.com/wordpress-mobile/WordPress-iOS/blob/75d92eb5687da89fc6dda707734ce15c43410f1a/WordPress/Classes/ViewRelated/Stats/Extensions/Double%2BStats.swift#L5-L43) in Stats.

In `paCL20-3g-p2`, however, we have obtained confirmation that this behavior _should_ differ. So rather than transition to use the existing `Double+Stats`, we will improve the placeholder to account for broader ranges and improve locale support.

It's worth noting that our Woo colleagues have noted similar requirements both [here](https://github.com/woocommerce/woocommerce-ios/issues/584#issuecomment-462808329) and [here](https://github.com/woocommerce/woocommerce-ios/issues/584#issuecomment-462808329).

PR #11672, opened 2019-05-10
Transitioned Period Chart to use live data

### LINKED ISSUE #11069: [Stats Refresh] Adapt DWMY Charts to leverage persisted data
In general, we will be transitioning the [DWMY charts](https://github.com/wordpress-mobile/WordPress-iOS/issues/11063) from using stubbed data to use real data, vended by the persistence layer @jklausa is working on.

stevebaranski:
### Description
Fixes #11069.

To test:

- Navigate to Stats for a given site and select a period in the filter bar (e.g., Days, Weeks, Months, Years).
- Observe that the chart now reflects live data, and updates when you select (by tapping) a dimension like Views, Visitors, Likes, or Comments.
- Rinse & repeat for a couple of different device types & orientations.
- Please keep in mind the caveats below. I've included some screenshots as well, but there should be no real aesthetic difference from #11210.

Update release notes:

### Notes

- The charts in the screenshots below include 10 bars; I think this can be attributed to [this default value](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/StatsServiceRemoteV2.swift#L198). I intend to fix this as part of #11671. We might find that a change to `WordPressKit` is in order, but I'm hoping we can just specify a non-default value [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Stores/StatsPeriodStore.swift#L307).
- The chart does not currently update or really interact in any way with the Date bar. This will also be accounted for via #11671.
- You might wonder why we pass arrays of `BarChartDataConvertible` to the `PeriodChart` in `SiteStatsPeriodViewModel`. This accommodates the filter bar instance for Views, Visitors, Likes & Comments that resides in our `OverviewCell`. That is to say, a `FilterTabBar` selection essentially changes our data source index.
- On the bright side, no changes were made in this PR (or #11605 or #11616)  to `StatsBarChartView`. That suggests that we probably got the abstractions right.
- Peril has expressed a concern about the LOC changed, but a fair bit of the changes in this PR are deletions.

_iPhone XS Max (Portrait):_
11069a

_iPhone XS Max (Landscape):_
11069b


ScoutHarris:
Hey @stevebaranski . 

The chart is looking great! 🎉 

The only thing I noticed is the numbers on the y-axis are duplicated. This is from en.blog, which has a lot of data:

ios_chart

Comparing that with Android:

android_chart

stevebaranski:
Thanks @ScoutHarris - thanks for the review, and for surfacing that issue. I guess you could say the PR review process works! 😂  

That's certainly not what our users deserve to see. I think this _might_ be a regression introduced via #11143, but I will investigate more thoroughly & report back once a fix is worth your time. 😄

stevebaranski:
@ScoutHarris as discussed in `p1557761654009100-slack-nde-stats-refresh`, Zeplin comps depicted axes without decimals, and accounts for the as-built discrepancy.

That said, you rightly noted a case where the lack of a decimal falls down a bit. This is also evident on sites with little traffic (i.e., single-digit counts).

I examined the [Android implementation](https://github.com/wordpress-mobile/WordPress-Android/blob/fa4f422c863a521675782a7f5ef053e8a202b395/WordPress/src/main/java/org/wordpress/android/ui/stats/refresh/lists/sections/viewholders/BarChartViewHolder.kt#L95), and observed that it uses a formatter that is included with the [`MPAndroidChart`](https://github.com/PhilJay/MPAndroidChart/blob/master/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/LargeValueFormatter.java) framework.

The iOS framework (i.e., `Charts`) includes a [similar](https://github.com/danielgindi/Charts/blob/0d33db4da87a549abb8d517d71ac2c36ecd731a5/ChartsDemo-iOS/Swift/Formatters/LargeValueFormatter.swift) formatter, but it does not appear to be part of the pod, nor is it as well-documented.

I pushed an additional commit that uses that formatter, and applies a few minor chart configuration changes to account for different data sets. Hopefully you'll agree that this an improvement. 😄 

11672b-3

cc: @SylvesterWilmott

ScoutHarris:
Hey @stevebaranski . It is indeed better! Thank you.

Unfortunately, there is now one minor visual glitch. There is now basically no spacing between the `Visitors` label and the top of the chart.

yaxis_spacing

stevebaranski:
Good catch @ScoutHarris. I double-checked the dimension in Zeplin, and I think it looks better now with the additional commit.

11672b-4

Hopefully you agree. 😄

stevebaranski:
Thanks again @ScoutHarris!

SylvesterWilmott:
Hey @stevebaranski 

Do we need to show decimals on full numbers too? 

I can see numbers like `993.3` or `860.6`. The decimals should only be applied to shortened numbers like `4.2k`

Can we also limit the number of values on the Y axis to 5 (including the 0)?

stevebaranski:
👋 @SylvesterWilmott

> Do we need to show decimals on full numbers too?

This is a consequence of transitioning to use the value formatter that comes with the Charts framework for all values. Making this change has the following benefits:

- It is more consistent with the Android implementation.
- In my testing, having a decimal was more consistent across a broad range of actual data. For example, in the comment [above](https://github.com/wordpress-mobile/WordPress-iOS/pull/11672#issuecomment-491901057), I noted that the presence of the decimal was useful for sites with low traffic (i.e., smaller counts). That is to say, _not_ having the decimal resulted in multiple axis entries where single digits were repeated.

Once all of the Chart functionality is in place, I'd like to revisit this to see if we can do better - honoring the intent you stated. Is that okay?

> Can we also limit the number of values on the Y axis to 5 (including the 0)?

Absolutely. The current behavior is based on what I saw in Zeplin, but it is a [one line](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Stats/Charts/StatsBarChartView.swift#L249) change.

Issue #11080, opened 2019-02-18
[Stats Refresh] Referrers card: add placeholder icon


ScoutHarris:
When provided, add a placeholder Referrer icon to be displayed until the Referrer icon is loaded. (Ref #10998)

ScoutHarris:
From latest enhancements, listed on #10840:
  - [ ] Referrers: 
    - [ ] When no icon is available then show `gridicons-globe`. 
    - [ ] Replace the `Search Engines` magnifying glass icon with `gridicons-search`.

ScoutHarris:
Both issues addressed in referenced PRs.

PR #11674, opened 2019-05-10
[Stats Refresh] StatsViewModels: Replace default search engine icon

frosty:
Refs #11080.

I took an attempt at replacing the search engine icon with a Gridicon as referenced in the issue above.

I figured the view model would be the best place for it, as that's all about transforming the data into the correct format to view. I also opted to switch on the existing icon URL filename instead of the referrer title, as I thought that would change for different locales.

I don't love that it's duplicated in two places, although the rest of the method is already duplicated. If we were to use this in more places, then maybe extracting it somewhere else would make sense. But if you have any suggestions for improvements, or you'd tackle this a different way, let me know!

**Before**

Screenshot 2019-05-10 at 11 00 43

**After**

Screenshot 2019-05-10 at 10 37 45

**To test:**

* Build and run, navigate to search referrers in Stats, and check it looks as above.


ScoutHarris:
Hey @frosty .

The only issue is the image is a bit too big. 

`StatsTotalRow` bases the image size on what type it is. Since you're using the `icon` parameter, the image size is 24x24. The Referrer icons are meant to be 20x20, which is intrinsic to the `socialIconURL` param. 

The complication is that, before this, all Referrer icons were downloaded, and thus the size is tied to this bit of logic, in `StatsTotalRow.configureIcon`:
```
        if let iconURL = rowData.socialIconURL {
            imageWidthConstraint.constant = Constants.socialImageSize
            downloadImageFrom(iconURL)
        }
```

PR #11677, opened 2019-05-11
Add Condensed Steps to Build Instructions

designsimply:
Condensed Steps are a shortened version of the build instruction steps in a numbered list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Please refer to the sections below for more detailed information.
 1. `cd WordPress-iOS` to enter the working directory.
 1. `rake dependencies` to install all dependencies required to run the project (this may take some time to complete).
 1. `rake xcode` to open the project in Xcode. 
-1. Compile and run the app on a device or an emulator.
+1. Compile and run the app on a device or an simulator.
 
 In order to login to WordPress.com using the app:
 
--- a/README.md
+++ b/README.md
@@ -12,20 +12,22 @@ Please refer to the sections below for more detailed information.
 1. Download and install Xcode.
 1. `git clone git@github.com:wordpress-mobile/WordPress-iOS.git` in the folder of your preference.
 1. `cd WordPress-iOS` to enter the working directory.
-1. `rake dependencies` this may take some time to complete.
-1. `rake xcode` to launch the workspace.
+1. `rake dependencies` to install all dependencies required to run the project (this may take some time to complete).
+1. `rake xcode` to open the project in Xcode. 
 1. Compile and run the app on a device or an emulator.
 
 In order to login to WordPress.com using the app:
 
 1. Create a WordPress.com account at https://wordpress.com/start/user (if you don't already have one).
-1. Create an application at https://developer.wordpress.com/apps/ 
-1. Set "Redirect URLs"= `https://localhost` and "Type" = `Native` and click "Create".
+1. Create an application at https://developer.wordpress.com/apps/.
+1. Set "Redirect URLs"= `https://localhost` and "Type" = `Native` and click "Create" then "Update".
 1. Copy the `Client ID` and `Client Secret` from the OAuth Information. 
 1. `mkdir -p ~/.mobile-secrets/iOS/WPiOS/` to create a place to store app credentials.
 1. `cp ./WordPress/Credentials/wpcom_app_credentials.txt ~/.mobile-secrets/iOS/WPiOS/wpcom_app_credentials` to copy the sample credentials file to your home folder.
 1. Paste `Client ID` and `Client Secret` from the app you created into `WPCOM_APP_ID` and `WPCOM_APP_SECRET` in `~/.mobile-secrets/iOS/WPiOS/wpcom_app_credentials`.
-1. Compile and run the app on a device or an emulator.
+1. Recompile and run the app on a device or an emulator.
+
+You can only log in with the WordPress.com account that you used to create the WordPress application.
 
 ### Download Xcode
 
--- a/README.md
+++ b/README.md
@@ -5,11 +5,9 @@
 
 ## Build Instructions
 
-### Condensed Steps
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,28 @@
...
+1. `rake dependencies` this may take some time to complete.
+1. `rake xcode` to launch the workspace.
+1. Compile and run the app on a device or an emulator.
+
+In order to login to WordPress.com using the app:
+
+1. Create a WordPress.com account at https://wordpress.com/start/user (if you don't already have one).
+1. Create an application at https://developer.wordpress.com/apps/ 
+1. Set "Redirect URLs"= `https://localhost` and "Type" = `Native` and click "Create".
+1. Copy the `Client ID` and `Client Secret` from the OAuth Information.
```
shiki: After finishing the _Create_ step above, I get forwarded to this page. 

app

It's not obvious where the `Client ID` and `Client Secret` should be copied from. Should we expand this or do you think it's understandable enough?

README.md
```diff
@@ -5,6 +5,28 @@
...
+In order to login to WordPress.com using the app:
+
+1. Create a WordPress.com account at https://wordpress.com/start/user (if you don't already have one).
+1. Create an application at https://developer.wordpress.com/apps/ 
+1. Set "Redirect URLs"= `https://localhost` and "Type" = `Native` and click "Create".
+1. Copy the `Client ID` and `Client Secret` from the OAuth Information. 
+1. `mkdir -p ~/.mobile-secrets/iOS/WPiOS/` to create a place to store app credentials.
+1. `cp ./WordPress/Credentials/wpcom_app_credentials.txt ~/.mobile-secrets/iOS/WPiOS/wpcom_app_credentials` to copy the sample credentials file to your home folder.
+1. Paste `Client ID` and `Client Secret` from the app you created into `WPCOM_APP_ID` and `WPCOM_APP_SECRET` in `~/.mobile-secrets/iOS/WPiOS/wpcom_app_credentials`.
+1. Compile and run the app on a device or an emulator.
```
shiki: "Recompile" would probably be clearer.

README.md
```diff
@@ -5,6 +5,28 @@
...
+
```
shiki: ⚠️ After the steps, this space here is probably a good opportunity that "You can only log in with the WordPress.com account that you used to create the WordPress application.". 

To expand, I have 2 accounts, Alpha and Bravo, I used Alpha to create the WP app Client ID and Secrets. When I ran the app, I could not log in as Bravo. I can only log in as Alpha.

README.md
```diff
@@ -5,6 +5,28 @@
...
 
+### Condensed Steps
+
+Please refer to the sections below for more detailed information.
+
+1. Download and install Xcode.
+1. `git clone git@github.com:wordpress-mobile/WordPress-iOS.git` in the folder of your preference.
+1. `cd WordPress-iOS` to enter the working directory.
+1. `rake dependencies` this may take some time to complete.
+1. `rake xcode` to launch the workspace.
```
shiki: What do you think about this? 

```
`rake xcode` to open the project in Xcode. 
```

README.md
```diff
@@ -5,6 +5,28 @@
...
 ## Build Instructions
 
+### Condensed Steps
+
+Please refer to the sections below for more detailed information.
+
+1. Download and install Xcode.
+1. `git clone git@github.com:wordpress-mobile/WordPress-iOS.git` in the folder of your preference.
+1. `cd WordPress-iOS` to enter the working directory.
+1. `rake dependencies` this may take some time to complete.
```
shiki: I wonder if we should explain what `rake dependencies` does. 

```
`rake dependencies` to install all dependencies required to run the project.
```

README.md
```diff
@@ -5,6 +5,28 @@
...
+1. `rake xcode` to launch the workspace.
+1. Compile and run the app on a device or an emulator.
+
+In order to login to WordPress.com using the app:
+
+1. Create a WordPress.com account at https://wordpress.com/start/user (if you don't already have one).
+1. Create an application at https://developer.wordpress.com/apps/
```
shiki: Nitpick: This is the only step that does not end with a period. 🙂

README.md
designsimply: Good suggestion. I will update that.

README.md
designsimply: I'll add a note in the previous step to click "Update".

README.md
designsimply: I thought about whether to say "You can only" or "Now you can"—and I took your suggestion as the final. Thanks!

README.md
designsimply: Love it.

README.md
designsimply: I like it but I also intentionally tried to keep each line VERY short and I still wanted to keep in the "this may take some time to complete" part because it's disconcerting when quite a lot of time goes by and you're not sure whether something was working or not. I'll add both for now, and I think I may move the timing part to lower in the document on the next pass.

README.md
designsimply: Hah. That's because I have the habit of leaving punctuation off of sentences that end with a URL… I will fix.

README.md
shiki: That's true. I was concerned about making the line longer too. I wanted to emphasize that we use `rake dependencies` to install all dependencies for developers who already know about `pod install` and other types of dependency managers. But since we already explain this in the _Third party tools_ section, maybe it's not needed. 🤷‍♂ I leave that up to you. 🙂

designsimply: Sweet! I'm keeping it for now and will likely move it in the next edit.

shiki:
@designsimply I like the _Getting Started_ idea. We could link to it both after the condensed build steps in `README.md` and in `CONTRIBUTING.md`.

PR #11689, opened 2019-05-14
Allow overriding Gutenberg dependencies with a local checkout

koke:
This PR introduces a way to run install Gutenberg from a local checkout without modifications to the Podfile.

To test:

- Running `rake dependencies:pod:install` (or any other variant that runs `pod install`) should work as before
- Running `rake dependencies:pod:install LOCAL_GUTENBERG='/path/to/gutenberg-mobile'` should install Gutenberg and its dependencies from the local copy

Update release notes:

cc @etoledom @pinarol @SergioEstevao


Podfile
```diff
@@ -76,6 +76,9 @@ end
 
 def gutenberg(options)
     options[:git] = 'http://github.com/wordpress-mobile/gutenberg-mobile/'
+    if ENV['LOCAL_GUTENBERG']
+      options = { :path => ENV['LOCAL_GUTENBERG'] }
```
jtreanor: We could use a default path here, since `gutenberg-mobile` is likely to be in the same directory as WPiOS.

```suggestion
      options = { :path => ENV['GUTENBERG_PATH'] || '../gutenberg-mobile' }
```

jtreanor:
This is a great idea! I've tested it and it works as described. I have a few thoughts:

- Running `rake dependencies:pod:install LOCAL_GUTENBERG='/path/to/gutenberg-mobile'` will update `Podfile.lock`. Anyone using this must make sure not the commit those changes as they will fail CI.
- One possible improvement would be to default to Gutenberg being in the same directory as WPiOS. Then dependencies could be installed just with something like `rake dependencies:pod:install LOCAL_GUTENBERG=1`. I've commented above with my suggestion for this. What do you think? I'm fine with leaving it as is also.

Podfile
etoledom: If we use a default local path, won't it _always_ use gutenberg from local?

etoledom:
Works great on my end!
> rake dependencies:pod:install LOCAL_GUTENBERG=1

This sounds nice!
Would it be possible to add the default in the command itself?
Or to make the command to be something like:
`rake dependencies:pod:install:local_gutenberg` and let the script do everything (if there is a default path)

Podfile
jtreanor: Nope, because we still have the `if ENV['LOCAL_GUTENBERG']` check above, but it's treated as a boolean rather than a path.

etoledom: Right! Sounds good :)

koke:
I really like the idea of a default path, but not so much having two different variables for it, so in 88e18cd747 I still kept the `LOCAL_GUTENBERG` variable, but if it looks like a path (it has a `/` in it) it will use the provided path, and otherwise it uses the default (so it works if you put `LOCAL_GUTENBERG=1`, but also `true`, `yes`, or `whatever` 😬). What do you think about this instead?

> to make the command to be something like: rake dependencies:pod:install:local_gutenberg

I thought about doing a special command, but environment variables seemed like a better approach since there are different ways you could do `pod install`, I often do `rake xcode`, but also `rake dependencies` or `bundle exec pod install`. The env variable would work with any of those variations.

> Running rake dependencies:pod:install LOCAL_GUTENBERG='/path/to/gutenberg-mobile' will update Podfile.lock. Anyone using this must make sure not the commit those changes as they will fail CI

True, but this doesn't change from the previous workflow of updating the `Podfile` manually. It'd be interesting to explore pre-commit hooks for this kind of thing. I like how Gutenberg uses [husky](https://github.com/typicode/husky) to enforce hooks for everyone (they get set up on `npm install`), so maybe we can come up with something similar in our `Rakefile`, but it seems out of scope for this change.

Issue #11683, opened 2019-05-13
2019 Color Palette Update Master Issue


nheagy:
## Goals
- Update the app to use the new Classic Bright color palette, just as [Android has](https://github.com/wordpress-mobile/WordPress-Android/pull/9495)

## Tasks
- [x] Update project to iOS 11+
- [x] Add iOS format to the [Automattic color studio](https://github.com/automattic/color-studio) (thanks @porada)
- [x] New color API (#11946)
- [x] Replace colors throughout app
  - [x] Tab bar
  - [x] Navigation bar
  - [x] Table views
  - [x] Existing code colors to new API
  - [x] Login/Signup flow

## For iteration 2 or dark mode
- [ ] Make tab bar translucent
  - Also make sure views go below the tab bar. For example, reader currently doesn't.

## Smaller issues
**Note**: Use the semantic colors whenever possible. That means `primary` for Blue, `neutral` for gray, `text` for text/default text, `textSubtle` when text is `neutral-50`, etc. When unsure, check the values in `extension UIColor` in `UIColor+MurielColor` (or the ObjC file when necessary). 

- [x] Update colors in new blog post list once that is merged #12087
- [x] Background color of custom table views #12007
- [x] Notification selection color #2589
- [x] Ensure media picker doesn't revert navigation bar colors
- [x] Reader buttons are a bit off
- [x] Tab bar unselected state should be neutral-200
- [x] Yellow "private" text should be gray, preceded by a •
- [x] Register domain yellow icon should be yellow-20
- [x] The border below this header should be neutral-5 (#12167)
- [x] The “Comment on…” text should be neutral-50  (#12167)
- [x] The “Weekly hangout” text can be the default text color  (#12167)
- [x] Number of likes button text in Reader comments is still orange (#12167)
- [x] Likes star color in Reader comments view should match reader post list (#12167)
- [x] No results view: lighter than the table view background
---
- [x] On the manage plugins screen, we can use the same green-40 and yellow-40 for Active/Needs Update icons. For the text, we can use green-50 and yellow-50.
- [ ] For the plugins search field, default `text` color in place of the current dark gray; neutral-50 in place of the lighter gray text.
- [ ] Change the default plugin icon from that shade of blue to neutral-50
---
- [ ] backgrounds: Activity Log has a blue background
	- [ ] Themes is also blue and Menus is too dark
	- [ ] media background also wrong
- [ ] Animation box needs new assets from @mattmiklic 
[...]

Issue #9633, opened 2018-06-22
Replace UIWebView with WKWebView


nheagy:
Seems like it's time to fully excise UIWebView from the codebase:

https://twitter.com/bradeeoh/status/1009302764722184192

https://developer.apple.com/documentation/uikit/uiwebview?changes=_6

As the tweet points out, Apple is looking for feedback on any remaining issues with migrating away from UIWebView, so if we find we should [log bugs with apple](http://bugreporter.apple.com).

cc: @kwonye

koke:
I'm not sure if there's more but the thing holding us back that I'm aware of is authentication support.

We've switched to WebKit in a bunch of places already but only for iOS 11+ devices, since that's when WebKit introduced support to look at the cookies (there were some other issues on iOS 10 IIRC). Because of this, fully dropping UIWebView is conditioned to dropping support for iOS 10.

prathap2626:
What is the alternative for UIWebView in iOS 12.0?

stevebaranski:
Greetings @prathap2626! Apple recommends the use of [`WKWebView`](https://developer.apple.com/documentation/webkit/wkwebview).

yangkunlll:
We have a iOS web application that uses remotely served images.

The images are served with a Cache-Control header: 'Cache-Control': 'max-age=31536000'.

Prior to iOS 12, the images were fetched once and cached during the usage of the app. After upgrading to iOS 12, the cache stopped working and images are always fetched from the server. We use UIWebView (we have not yet migrated to WKWebView).

It does not happen in Safari on iOS 12.

I cannot find any clue to why this suddenly no longer working. Does anyone have a clue?

frosty:
This would also be great for Marzipan – at least at the moment, it doesn't include `UIWebView`.

designsimply:
Noting that this is probably worth somebody investigating again so I'm adding it to the next Groundskeeping rotation for a check. "I don’t think the switch should take too long, it’s more about feasibility / investigation maybe."

diegoreymendez:
@designsimply - Removed from Groundskeeping and mover to a mini-project board.

koke:
It’s official now:

> The App Store will no longer accept new apps using UIWebView as of April 2020 and app updates using UIWebView as of December 2020.

https://developer.apple.com/news/?id=12232019b

PR #11691, opened 2019-05-14
Drop support for iOS 10

nheagy:
Refs #11683

This PR removes support for iOS 10.
1. Changed targets' deployment target to iOS 11
2. Removed all `available` checks for iOS 11, and their `else` clauses, etc
3. Fixed as many new warnings as I could easily fix

### Areas to test:
It's impossible to anticipate all areas of the app that might be affected but this change, however the following were areas that had warnings in iOS 11 which I fixed:
- PluginDirectoryViewController
- BlogListViewController
- BlogSelectorViewController
- PageListViewController
- AbstractPostListViewController
- ThemeBrowserViewController

### Remaining warnings:
- `'addressDictionary' was deprecated in iOS 11.0: Use @properties` in CLPlacemark+Formatting.swift. Maybe understandable to @SergioEstevao @jleandroperez @frosty ?
- `'automaticallyAdjustsScrollViewInsets' was deprecated in iOS 11.0: Use UIScrollView's contentInsetAdjustmentBehavior instead` in MediaLibraryViewController.swift. Maybe help from @frosty?


WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
```diff
@@ -349,22 +300,14 @@ class AbstractPostListViewController: UIViewController,
...
 
     fileprivate func configureSearchBackingView() {
         // This mask view is required to cover the area between the top of the search
         // bar and the top of the screen on an iPhone X and on iOS 10.
-        var topAnchor = topLayoutGuide.bottomAnchor
-
-        if #available(iOS 11.0, *) {
-            topAnchor = view.safeAreaLayoutGuide.topAnchor
-        }
+        var topAnchor = view.safeAreaLayoutGuide.topAnchor
```
frosty: I get a warning here recommending using a `let`, as `topAnchor` never changes.

nheagy:
> Regarding the Core Location deprecation... looks like we may need to build an address manually using the properties of CLPlacemark

I'm hoping someone that has actually worked on that code is interested in fixing it. I don't even know where it's used 😜

nheagy:
@frosty this is ready for a second review.

SergioEstevao:
> 'addressDictionary' was deprecated in iOS 11.0: Use @properties in CLPlacemark+Formatting.swift. Maybe understandable to @SergioEstevao @jleandroperez @frosty ?

@nheagy Just pushed a commit that sorts out this warning, I hope you don't mind.

nheagy:
I don't mind at all @SergioEstevao, you're my hero!

@frosty I fixed that constraint problem when viewing articles in the reader

koke:
Nice, this would also unlock #9633 🎉

nheagy:
@frosty @shiki thanks for the reviews! 🎉

> I think we can also remove the version check in

Removed 👍

> This is also present in develop.

If there's not an issue for that yet, you should create it!

PR #11704, opened 2019-05-15
Update README.md

designsimply:
It's complex to compile and run the app on a device because it would require making changes to use a local provisioning profile. Let's take out the mention of compiling to a device because of that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ In order to use these details, you'll need to create a credential file in your b
 
 Then edit the `~/.mobile-secrets/iOS/WPiOS/wpcom_app_credentials` file and change the `WPCOM_APP_ID` and `WPCOM_APP_SECRET` fields to the values you got for your app.
 
-Then you can compile and run the app on an emulator and log in with a WordPress.com account.  Note that authenticating to WordPress.com via Google is not supported in development builds of the app, only in the official release.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,7 +70,7 @@ In order to use these details, you'll need to create a credential file in your b
 
 Then edit the `~/.mobile-secrets/iOS/WPiOS/wpcom_app_credentials` file and change the `WPCOM_APP_ID` and `WPCOM_APP_SECRET` fields to the values you got for your app.
 
-Then you can compile and run the app on a device or an emulator and log in with a WordPress.com account.  Note that authenticating to WordPress.com via Google is not supported in development builds of the app, only in the official release.
+Then you can compile and run the app on an emulator and log in with a WordPress.com account.  Note that authenticating to WordPress.com via Google is not supported in development builds of the app, only in the official release.
```
jtreanor:
```suggestion
Then you can compile and run the app on a simulator and log in with a WordPress.com account.  Note that authenticating to WordPress.com via Google is not supported in development builds of the app, only in the official release.
```

Its a little pedantic but we don't have emulators on iOS. They're simulators 😅

designsimply: Lol. It's easy to mix them up because they're different on each platform! Thanks @jtreanor. 🙂

PR #11372, opened 2019-03-28
Integrated new analytics event for stats chart interaction

stevebaranski:
Fixes #11066.

To test:

- Checkout the branch and confirm that tests pass.
- Exercise the charts by selecting the Post Stats & Period charts, selecting different bars.
- In the console, you should see evidence that `stats_overview_bar_chart_tapped` is logged to the console. If a period is selected, it will appear in the properties with a `granularity` key. Otherwise, the properties dictionary will be empty.
- This should also be visible in _Tracks_ as `wpios_stats_overview_bar_chart_tapped`.

Update release notes:

- [X] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #11706, opened 2019-05-15
Resolved a defect where in limited cases a chart would not refresh with new data

stevebaranski:
Fixes an issue where the Post Stats chart presented stale data. 

As discussed via Slack, this issue is most apparent when navigating from _Posts and Pages_ to _Post Stats_ for a given post. Currently the data on the subsequent screen does update unless a pull-to-refresh gesture is performed. The chart, however, was _not_ refreshed in response to pull-to-refresh.

Now the chart updates via pull-to-refresh along with the other screen content.

To test:
- Checkout the branch and confirm that existing tests pass.
- Replicate the scenario above, and confirm that it updates as expected.

Update release notes:


stevebaranski:
This was an optimization introduced while the stub data was still [in place](https://github.com/wordpress-mobile/WordPress-iOS/pull/11372/).

In hindsight, I should have removed it via either #11616 or #11672.

PR #11707, opened 2019-05-16
[Stats Refresh] Insights Tags & Categories: fix displaying details view

ScoutHarris:
Ref #11360

This changes Insights > Tags and Categories to use `UITableViewCell`s instead of one `UIStackView`.

Previously, all the data was stuffed into the stack view on the `TopTotalsCell` card. Now it only uses the `TopTotalsCell` card to show the header via `DetailSubtitlesHeaderRow`. The stack view is left empty. Instead, a `UITableViewCell` is added to the table for each data row via the new `DetailExpandableDataRow`. 

There should be no visual or functional changes, but showing the details view is faster. To note, there is still no loading view, so there may still be a blank view for a couple seconds for longer periods.

Known Issues:
- When a row is expanded/collapsed, _all_ the rows' arrows update. I don't _think_ this is new as I've seen it before, but it's pretty obvious now. I'll address that separately.
- With all these "replace stack views with table cells" PRs, it's introduce a new problem where the details table isn't actually being refreshed, it's being duplicated, displaying on top of itself. (It's painfully obvious when trying to Debug View Hierarchy in Xcode.) Woops. Since I broke it long before now, and it's a problem with the details table itself, I'll fix that separately as well.

Update release notes:

To test:

---
- Go to Stats > Insights > Tags and Categories on a very active site.
  - Tip: Hogwarts has several expandable tags/categories.
- Select `View more`.
- For a _non-expandable_ row, verify row selection shows a web view.

---
- For an expandable row, row selection shows _all_ child rows.
  - Previously, child rows on Details were limited to 10. This was a bug that has now been addressed.
- When a row is expanded:
  - The child label text color is light grey.
  - Child rows have no separator line between them.
  - There is a full width separator line above the parent row, and below the last child.
- Selecting a child row shows a web view.

tags_cats

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsViewModel.swift
@@ -53,7 +53,7 @@ class SiteStatsInsightsViewModel: Observable {
 
         if insightsStore.fetchingOverviewHasFailed &&
             !insightsStore.containsCachedData {
-            return ImmuTable(sections: [])
+            return ImmuTable.Empty
         }
 
         insightsToShow.forEach { insightType in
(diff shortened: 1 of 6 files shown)


ScoutHarris:
Thanks @stevebaranski !

Issue #10380, opened 2018-11-01
[Stats Refresh] Period stats: show loading view


ScoutHarris:
On the Stats Period data views, show a loading view while the data is loading.

Currently, the Period view is shown with no data:

![period_loading](https://user-images.githubusercontent.com/1816888/47866329-1296c780-ddc4-11e8-8aeb-648388f8b5be.png)

We should show the typical loading view. For example (from Activities):

![loading](https://user-images.githubusercontent.com/1816888/47866222-c186d380-ddc3-11e8-83b5-925c9d187be1.png)

danielebogo:
Done in #11634 and #11711

PR #11711, opened 2019-05-16
[Stats Refresh] Revert cache method

danielebogo:
Refs. #10380

This PR just has a small revert to the previous method to check if there's cached data. This because the new block used to bind the loading view and its implementation don't require any change to that method.

## To test:
• From My Sites select a site and open its dashboard
• Select Stats to open the Stats section.
• Select one of the tab period tab _Days, Weeks, Months, Years_.
• You should see the loading view
• The loading view should disappear as soon the data is loaded
• The loading view should appear only if there's no data cached
• You can also test it tapping on the stats bar or changing the period date

Update release notes:


danielebogo:
Thanks @ScoutHarris !

PR #11714, opened 2019-05-16
[Stats Refresh] Insights Followers details: get max results from API

### LINKED ISSUE #11695: Stats: Insights Followers: Not all Followers are Displayed
### Expected behavior
When clicking "View all" below the Followers section of Stats > Insights, I expect to be able to scroll through all

### Actual behavior
After clicking "View all", I can see the most recent 20 followers, but no more. 

1. Open the mobile app to a site with many followers.
2. Go to My Site > Stats > Insights, find the "Followers" section, and tap "View all".
3. Compare the number of followers you can see here with those in My Site > People > Followers.

##### Tested on iPhone SE, iOS 12.2, WPiOS 12.3

Here's how that looks - the site has way more than 20 followers, but only 20 show up in Stats > Insights > Followers:
statsfollowers

User report in 2026096-zen.
In Calypso, Stats > Insights > Followers > View All redirects to People > Followers, so the same list is seen in both places.

ScoutHarris:
Fixes #11695

Using a limit of 0 on the followers API returns the default 20 results, and not _all_ results as one would expect. This API also returns a maximum of 100, regardless of the requested number. So this sets the `limit` on the API request to get all 100 results.

To test:
- Go to Stats > Insights > Followers > View more.
  - Tip: en.blog has lots of both follower types.
- Verify there are now 100 results displayed for both WordPress.com and Email. (Or, at least verify there's more than 20.)

Update release notes:

### DIFF
--- a/WordPress/Classes/Stores/StatsInsightsStore.swift
+++ b/WordPress/Classes/Stores/StatsInsightsStore.swift
@@ -465,14 +465,18 @@ private extension StatsInsightsStore {
 
         let api = apiService(for: siteID)
 
-        api.getInsight(limit: 0) { (dotComFollowers: StatsDotComFollowersInsight?, error) in
+        // The followers API returns a maximum of 100 results.
+        // Using a limit of 0 returns the default 20 results.
+        // So use limit 100 to get max results.
+
+        api.getInsight(limit: 100) { (dotComFollowers: StatsDotComFollowersInsight?, error) in
             if error != nil {
                 DDLogInfo("Error fetching dotCom Followers: \(String(describing: error?.localizedDescription))")
             }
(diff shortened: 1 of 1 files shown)


WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -490,8 +494,8 @@ private extension StatsInsightsStore {
         state.fetchingAllCommentsInsight = true
 
         // The API doesn't work when we specify `0` here, like most of the other endpoints do, unfortunately...
-        // 1000 was chosen as an arbitraily large number that should be "big enough" for all of our users
-        // but I'm open to increasing it.
+        // 1000 was chosen as an arbitrarily large number that should be "big enough" for all of our users.
```
stevebaranski: 👍

PR #11715, opened 2019-05-16
Use editor published post event, instead of quick published post.

SergioEstevao:
This PR changes the stat sent when the Publish Now button is used on the editor. Instead of using the  `editorQuickPublishedPos`t event we now consolidate all publish actions on the editor with `editorPublishedPost`.

To test:
 - Start the open
 - Create a new post
 - Add some content
 - Save the post has draft
 - Exit editor
 - Open the post again
 - Tap on publish now
 - Check if event sent is `editorPublishedPost`. I checked this by putting a breakpoint on `PostEditor+Publish.swift` on the `publishPost` method

Update release notes:


jkmassel:
Merged this to get it into `12.5`.

PR #11719, opened 2019-05-17
Completed transition to coordinate period bar and chart interaction

### LINKED ISSUE #10671: [Site Creation] VerticalsService (Prompt)
Integration with the `verticals/prompt` endpoint, as part of the implementation of the `VerticalsService`

### LINKED ISSUE #11671: [Stats Refresh] Coordinate navigation between the date bar & the Post Stats & Period Chart

stevebaranski:
### Description
Fixes #11671.

To test:
- Navigate to Stats for a given site, and review the Post Stats & Period Charts presented. 
- Observe that navigating with the Period bar updates the chart, as well as the highlighted bar when applicable.
- Rinse & repeat for a couple of different periods, dimensions, device types & orientations, confirming no regressions in behavior or appearance.

Update release notes:

### Screenshots

_Period Chart (iPhone) :_

11671c_iPhone_periodChart

_Post Stats Chart (iPhone) :_

11671c_iPhone_postStats

_Post Stats Chart (iPad)  :_

11671c_iPad_02

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTableViewController.swift
@@ -52,8 +52,7 @@ class PostStatsTableViewController: UITableViewController, StoryboardLoadable {
             return nil
         }
 
-        let periodCount = SiteStatsTableHeaderView.defaultPeriodCount
-        cell.configure(date: selectedDate, period: .day, delegate: self, expectedPeriodCount: periodCount)
+        cell.configure(date: selectedDate, period: .day, delegate: self)
         viewModel?.statsBarChartViewDelegate = cell
 
         return cell
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTableViewController.swift
@@ -52,7 +52,8 @@ class PostStatsTableViewController: UITableViewController, StoryboardLoadable {
             return nil
         }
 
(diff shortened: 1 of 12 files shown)


WordPress/Classes/ViewRelated/Stats/Shared Views/Post Stats/PostStatsTableViewController.swift
```diff
@@ -52,7 +52,8 @@ class PostStatsTableViewController: UITableViewController, StoryboardLoadable {
             return nil
         }
 
-        cell.configure(date: selectedDate, period: .day, delegate: self)
+        let periodCount = SiteStatsTableHeaderView.defaultPeriodCount
+        cell.configure(date: selectedDate, period: .day, delegate: self, expectedPeriodCount: periodCount)
```
ScoutHarris: Just a thought - if you give `expectedPeriodCount` in `configure`a default value, like maybe `expectedPeriodCount: Int = defaultPeriodCount`, this won't be necessary since it's just setting the count to the default.

stevebaranski: Thanks for the suggestion @ScoutHarris! I have pushed a commit that takes your suggestion into account.

PR #11722, opened 2019-05-17
Removing FakePreviewBuilder

### LINKED ISSUE #11699: Remove FakePreviewBuilder from the app
Now that we are auto saving a post and uploading a draft before preview,
we can take the last step in [Get rid of local previews](https://github.com/wordpress-mobile/WordPress-iOS/issues/9282) and remove the FakePreviewBuilder from the code.

After looking into this, I have a couple of questions that maybe @koke or @diegoreymendez  can help figure out. 

1. Currently we are showing the fake preview in cases where we can't find a url (either a preview url returned from auto save or a permalink on the post):

https://github.com/wordpress-mobile/WordPress-iOS/blob/425915492ad719b81f39e00cfc533dcbacba24bd/WordPress/Classes/ViewRelated/Post/PostPreviewGenerator.swift#L31

https://github.com/wordpress-mobile/WordPress-iOS/blob/425915492ad719b81f39e00cfc533dcbacba24bd/WordPress/Classes/ViewRelated/Post/PostPreviewGenerator.swift#L104

https://github.com/wordpress-mobile/WordPress-iOS/blob/425915492ad719b81f39e00cfc533dcbacba24bd/WordPress/Classes/ViewRelated/Post/PostPreviewGenerator.swift#L113

In this case, should we replace showFakePreview() with ` delegate?.previewFailed...` (which shows the empty view with a defined error message)?

2. When trying to preview while offline, the check for `url = post.permaLink.flatMap(URL.init(string:))` will fail for draft (since the upload call failed). I believe the technical solution proposed [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/11653) will allow the upload post to get called when we regain connectivity and as a result create the permalink which we will have to reload the preview web view with.  Is that correct?

yaelirub:
Now that we are auto saving a post and uploading a draft before preview,
we can take the last step in to get rid of local previews and remove the FakePreviewBuilder from the code. The expected behavior is to show an "error view" when ever a preview request fails instead of showing the fake preview. 

In this PR we are replacing the calls to showFakePreview() with previewFailed delegate method while logging the reason for the failure to Sentry.

This failure is unlikely to happen since we added a check for the existence of the post permalink before generating the preview.

**Before**

preview_offline_3

**after**
preview_offline_2

Fixes #11699 

To test:
1. Be offline
2. Create a new draft
3. preview the draft
4. See the "Empty view"

Update release notes:


WordPress/Classes/Utility/Logging/WPCrashLogging.swift
```diff
@@ -72,6 +72,13 @@ extension WPCrashLogging {
         Client.shared?.appendStacktrace(to: event)
         Client.shared?.send(event: event)
     }
+
+    static func logMessage(_ message: String) {
```
koke: I have some doubts about this, and I wonder if this should go to Tracks instead, but then I'm excited to try Sentry's logging features 😁 

One thing I would suggest is that we don't send a bunch of different messages all related to previews failing, because I imagine it'll be hard to group them together. From what I understand, you can [send any extra properties you want](https://docs.sentry.io/clients/cocoa/advanced/#sending-events) with an event. We could have the same message for all the failed previews, then have a `reason` property in `extra`, that detailed whether permalink was nil, there were no credentials, or if there was any generic web view error.

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -180,7 +180,7 @@ - (void)webView:(UIWebView *)webView didFailLoadWithError:(NSError *)error
 
     [self stopLoading];
 
-    [self.generator previewRequestFailedWithError:error];
+    [self.generator previewRequestFailedWithMessage:error.localizedDescription];
```
koke: I'm concerned about sending this error as-is, and potentially leaking private information. Also because it might get too noisy if it's also generated by normal connectivity issues. If you think there's still value on getting those, I'd log only the error domain and code.

WordPress/Classes/ViewRelated/Post/PostEditor+MoreOptions.swift
```diff
@@ -64,9 +64,7 @@ extension PostEditor where Self: UIViewController {
         }
     }
 
-    private func displayPreviewNotAvialble() {
-        let title = NSLocalizedString("Preview Unavailable for Published Posts", comment: "Title on display preview error" )
-        let subtitle = NSLocalizedString("Update the published post to view your changes.", comment: "subtitle on display preview error" )
+    private func displayPreviewNotAvialble(title: String, subtitle: String? = nil) {
```
koke: I know this was already like that, but it the name says `Avialble` instead of `Available` 😬

yaelirub: oops. Good catch.

WordPress/Classes/Utility/Logging/WPCrashLogging.swift
diegoreymendez: I agree with what's mentioned above.

On top of that, this function is called from `previewRequestFailed(message: String)`.  Isn't this an error message for that flow (instead of an informative one)?

I know there were some cases where this was being brought up in non-error scenarios and I was wondering if those are still valid.

yaelirub: @diegoreymendez,
>Isn't this an error message for that flow (instead of an informative one)?
Not sure I understand what you mean.

yaelirub:
@koke , @diegoreymendez , updated and ready for another look 🙏

yaelirub:
Waiting on https://github.com/Automattic/Automattic-Tracks-iOS/pull/103 to be merged

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -180,7 +180,7 @@ - (void)webView:(UIWebView *)webView didFailLoadWithError:(NSError *)error
 
     [self stopLoading];
 
-    [self.generator previewRequestFailedWithError:error];
+    [self.generator previewRequestFailedWithReason:@"Generic web view error"];
```
koke: I think it would still be valuable to log the error domain/code so we can differentiate expected errors like timeouts from things that we might want to address.

yaelirub:
@koke thank you so much for taking another look at this. 
I updated the generic web error to show the code and domain.

For the HUD blocking the UI on low connectivity, I thought about adding the option to dismiss the HUD on tap but while testing I realized that it might introduce a confusion as to what is happening especially if connectivity is regained.

 I will create a separate issue for this.

yaelirub:
As Waiting on Automattic/Automattic-Tracks-iOS#103 to be merged has been merged but WPiOS has not been updated yet (and it's unknown when it will be) I am merging this PR with the changes in WPCrashLogging.
Will remove this function after the updates to crash logging are merged to develop

Issue #11653, opened 2019-05-08
Offline Support: Implement UploadsManager.


diegoreymendez:
## Dependencies:

This issue depends on https://github.com/wordpress-mobile/WordPress-iOS/issues/11694 since it defines our main uploaders and implements the code to execute them.

## Description:

This issue was spawned from the discussion here: https://github.com/wordpress-mobile/WordPress-iOS/issues/11640

Based on this graph:

<img width="976" alt="57587998-167a7380-74e4-11e9-925f-408b4636f77d" src="https://user-images.githubusercontent.com/1836005/57740311-8aa44b00-768d-11e9-9ca7-585db74e399d.png">

We need to implement a class that can execute operations when any of these triggers are detected:

- Reachability detects we're back online (after being offline).
- When the app is placed in the foreground.

The `ViewController` pull-down-to-refresh can be considered out of scope for now, for simplicity.  We can implement it incrementally after we have an initial solution.

## Details:

This App needs to interact with several classes that implement the `Uploader` protocol, and ask them to start all pending uploads.

This class doesn't need to know about the uploads, and it doesn't need to know how to query our CoreData database.  The uploaders are supposed to take care of that.

diegoreymendez:
Updated the description to match the latest developments.

diegoreymendez:
Reopening and re-closing since this is coming up in my GitHub notifications.

PR #11736, opened 2019-05-21
Fix discard draft loosing all changes

### LINKED ISSUE #11435: Offline Support: Discard Losing all changes
## Description

We need to improve how we deal with local changes, and make sure that saving those changes has a meaningful and consistent meaning.

### Expected Behavior

If the device is offline, saving changes to a post means those changes shouldn't be lost when the user selects "Discard Changes" in a latter editing session.

### Actual Behavior

Local saves while offline don't properly create a saving checkpoint.  Tapping "Discard Changes" in a latter editing session is discarding all local changes.

####  Steps to reproduce

1. Tap post in post list
2. Edit post 
3. be offline
4. Close editor
5. Choose "Update Post" from action sheet 
6. Stay offline. The post will show "Uploading post" .
7. Stay offline. Wait till post shows "Failed to upload"
8. Go back to the post editor
9. Edit post 
10. Close editor and discard changes. 

discardChanges

---

### Update 2019-Nov-29

There is another discussion about the “Discard Changes” option most recently in https://github.com/wordpress-mobile/WordPress-iOS/pull/12947#issuecomment-559808264. An older discussion is from here https://github.com/wordpress-mobile/WordPress-Android/issues/9810#issuecomment-507716708. The gist is we should remove the Discard Changes dialog altogether.

yaelirub:
When a post fails to save, it status was set to `AbstractPostRemoteStatusFailed` which should have
made `post.hasNeverAttemptedToUpload` return true hence return false in `shouldRemovePostOnDismiss` in `discardChanges()`.
https://github.com/wordpress-mobile/WordPress-iOS/blob/d05cb734d50c253e0bf67fc8415dbc90f7089eff/WordPress/Classes/ViewRelated/Post/PostEditor%2BPublish.swift#L239
The problem was that the remoteStatus property was not really being saved. In order to save those changes we are merging the revisions back into the base post.

Fixes #11435

To test:

**upload failed post with no changes**

1.Be offline
2.Create a new draft
3.Save draft (either through options menu or by exiting the editor and choosing "save draft")
4.rerun xcode
5.See the post with "failed upload" status
6.tap on post
7.Exist editor
8.Choose discard
*Expected - post is not be deleted.

**upload failed post with changes:**

1.Follow steps 1-6 above
2.Make changes to post
3.Exit editor
4.Choose discard
Expected - Changes made in step 2 will be discarded and the post is not be deleted.
Before:
discard_offline_2

After:
discard_offline

Update release notes:

 If there are user facing changes, I have added an item to RELEASE-NOTES.txt.

### DIFF
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -236,8 +236,13 @@ const NSUInteger PostServiceDefaultNumberToSync = 40;
     };
     void (^failureBlock)(NSError *error) = ^(NSError *error) {
         [self.managedObjectContext performBlock:^{
-            Post *postInContext = (Post *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
+            AbstractPost *postInContext = (AbstractPost *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
             if (postInContext) {
+                if ([postInContext isRevision]) {
+                    postInContext = postInContext.original;
+                    [postInContext applyRevision];
+                    [postInContext deleteRevision];
+                }
                 postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
                 // If the post was not created on the server yet we convert the post to a local draft post with the current date.
                 if (!postInContext.hasRemote) {


yaelirub:
Thanks, @jklausa! 
>really dislike that we're doing [NSManagedObjectContext performBlock:] inside a network callback

I see your point. Hopefully the[ implementation of uploads manager](https://github.com/wordpress-mobile/WordPress-iOS/issues/11653) will separate the network call from saving locally. At least for uploading

PR #11739, opened 2019-05-21
Site Creation: first responder can now be resigned in the Domain Selection screen.

### LINKED ISSUE #11702: Site Creation: title and subtitle are sometimes transitioned out.
### Description:

The search field from the Domain Selection step never resigns first responder when going back.  This causes the keyboard to come up, and the headers to be transitioned out (if there's text in the search field or a title was set in the previous screen.

### Steps to Reproduce:

There are two ways to reproduce.

#### Path 1:

1. ...Go through the site creation flow
2. On Step 2 of 3 (basic information) enter a site title
3. Press "Next"
4. Go back one step
5. Press "Next"

Notice that the title and subtitle are transitioned out.  This is because there are search results based on the title AND because the keyboard is up since first responder wasn't properly resigned.

#### Path 2:

1. ...Go through the site creation flow
2. On Step 2 of 3 (basic information) leave the title site empty
3. Press "Next"
4. Enter some text in the search field
5. Go back one step
6. Press "Next"

Notice that the title and subtitle are transitioned out.  This is because there are search results based on the search field AND because the keyboard is up since first responder wasn't properly resigned.

diegoreymendez:
### Description:

This PR addresses the issue reported in #11702 that makes it impossible for the search field in the Domain Selection screen to resign first responder.  It also unifies the search logic a bit.

Fixes #11702 

### Important:

There are more issues when going back and re-opening this screen that can cause the layout to look a bit broken.  From my research those issues are not related to the changes in this PR, and I'm planning to address them incrementally after this fix.

#### Path 1:

1. ...Go through the site creation flow
2. Go to step 3 of 3, and tap on the search field.
4. Go back one step
5. Press "Next"

Make sure the search field doesn't have first responder, and make sure the title and subtitle aren't hidden.

### DIFF
--- a/WordPress/Classes/ViewRelated/Site Creation/Shared/TableViewOffsetCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Site Creation/Shared/TableViewOffsetCoordinator.swift
@@ -136,6 +136,7 @@ final class TableViewOffsetCoordinator {
 
     @objc
     private func keyboardWillHide(_ notification: Foundation.Notification) {
+        keyboardContentOffset = 0
         toolbarHasBeenAdjusted = false
         toolbarBottomConstraint?.constant = 0
     }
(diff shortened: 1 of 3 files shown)


diegoreymendez:
Thanks @stevebaranski !

PR #11500, opened 2019-04-18
Share Extension: Improve TextBundle/Pack import

nheagy:
Fixes #11223

*Also may fix these issues:*:
Refs #11350
Fixes #10736 (maybe—the caveat is that it ignores max image size entirely atm)
Fixes #9555

This improves our support for TextBundle/Pack support in two ways:
1. Images in the bundle are now added to the post
2. Markdown support is added (via [Down](http://github.com/iwasrobbed/Down), which wraps CommonMark). This also has the side effect of adding Markdown support for import in general.

### To test:
- First, we should ensure that all existing sharing functionality is unharmed…
- To test text bundle support, launch bear:
  - In the list of notes, swipe left on the one you'd like to share
  - Tap more
  - Tap Export note…
  - Tap Text Bundle
  - Tap WordPress
- Ensure that the text is formatted properly, that any included images are included, and that generally things look mostly correct

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #11740, opened 2019-05-21
Fix broken images in posts created by the share extensions

### LINKED ISSUE #9118: Images sometimes keep local path when using the share extension

nheagy:
Fixes #9118 
Replaces #11581

This fixes the problems with the share extension creating posts with broken images.

## To test:
- Share two images to the share extension, and publish
- Ensure both images are visible in the post
- Share a TextBundle that includes one or two images (from Bear, likely)
- Ensure the images work there, too

## Discussion

My first attempt to solve this, #11581, changed `AppExtensionsService.swift` to solve the possibility that the server had altered the name of the file, returning a filename that wouldn't match to our local file. However, I had failed to notice that `SharedCoreDataStack.swift` has the same issues, and also lacked the accounting for case changes that `AppExtensionsService` already had.

The app extension uploads images in a background session. It returns in one of two places: in the callback in `AppExtensionsService` if the upload is fast, or in `ShareExtensionSessionManager` via `WordPressAppDelegate` in the main app.

I've identified three reasons for the filename to change:
1. The filename changes case. It seems totally consistent for the server to return the filename as lowercase. When #11500 changed all temporary files created by the app extension to be `UUID`s it basically guaranteed that anytime the task completion was handled by `ShareExtensionSessionManager` the post would have broken images.
2. The extension changes. `PNG`s become `JPG` for example.
3. The server adds a counter to the filename. These seems to happen in some cases where the upload fails and is re-attempted. `myimg.jpg` becomes `myimg-1.jpg`, for example.

## Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 * Implemented Domain Credit feature
 * Implemented auto saving a post on preview  
 * The app now launches a bit more quickly.
+* Fixed broken images in posts created by the share extension.
 
 12.4
 -----
--- a/WordPress/WordPressShareExtension/AppExtensionsService.swift
+++ b/WordPress/WordPressShareExtension/AppExtensionsService.swift
@@ -349,9 +349,11 @@ extension AppExtensionsService {
                     if let remoteMediaID = remoteMedia.mediaID?.int64Value,
                         let remoteMediaURLString = remoteMedia.url?.absoluteString,
                         let localFileName = mediaUploadOp.fileName,
-                        let remoteFileName = remoteMedia.file {
+                        let remoteFileName = remoteMedia.file,
+                        let localFileNoExt = URL(string: localFileName)?.deletingPathExtension().lastPathComponent {
 
-                        if localFileName.lowercased().trim() == remoteFileName.lowercased().trim() {
+                        // sometimes the remote file name has been altered, such as `localfile.jpg` becoming `localfile-1.jpg`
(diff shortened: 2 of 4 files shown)


SergioEstevao:
> The server adds a counter to the filename. These seems to happen in some cases where the upload fails and is re-attempted. myimg.jpg becomes myimg-1.jpg, for example.

This happens when you try to update a file to the server that has the same name as an existing file on the server, no?

nheagy:
> This happens when you try to update a file to the server that has the same name as an existing file on the server, no?

That sounds right to me. If that's the case, using a UUID for the filename would have made that impossible. I'll have to see if I can get it to happen, still.

nheagy:
Thanks for the review, @SergioEstevao!

PR #11741, opened 2019-05-21
[Stats Refresh] Period Referrers: fix displaying details view

ScoutHarris:
Ref #11360, #11080

This changes Period > Referrers to use `UITableViewCell`s instead of one `UIStackView`.

Previously, all the data was stuffed into the stack view on the `TopTotalsCell` card. Now it only uses the `TopTotalsCell` card to show the header via `DetailSubtitlesHeaderRow`. The stack view is left empty. Instead, a `UITableViewCell` is added to the table for each data row via:
- `DetailExpandableRow` - for parent rows
- `DetailExpandableChildRow` - for child rows

And, because Referrers `Search Engines` rows can be expanded twice, "grandchild" rows are now supported.

This also addresses two other issues:
From #11080 - When no icon is available then show `gridicons-globe`.
From https://github.com/wordpress-mobile/WordPress-iOS/pull/11674#issuecomment-491386360 - fix `gridicons-search` icon size.

Update release notes:

To test:

---
- Go to Stats > Period > Referrers.
- For any Referrers that don't have an icon, verify the globe icon is displayed.

---
- Select `View more`.
- For any Referrers that don't have an icon, verify the globe icon is displayed.
- For a _non-expandable_ row, verify row selection shows a web view.

---
- For an expandable row, row selection shows all child rows.
- When a row is expanded:
  - The child label text color is light grey.
  - Child rows have no separator line between them.
  - There is a full width separator line above the parent row, and below the last child.
- Selecting a child row shows a web view.

expanded

---
- For the `Search Engines` row: (tip: en.blog has this row)
  - Row selection shows all child rows.
  - If a child row is expandable, it looks like a parent row, i.e. text color is dark grey, and the chevron is pointed down.
  - When a child row is expanded, it exhibits the same behavior has listed above (`For an expandable row...`).

search_engines_init

search_engines_expanded

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
@@ -199,21 +199,21 @@ private extension StatsTotalRow {
             return
         }
 
+        let imageSize = rowData.statSection?.imageSize ?? StatSection.defaultImageSize
+        imageWidthConstraint.constant = imageSize
+
         imageView.isHidden = !hasIcon
 
         if let icon = rowData.icon {
-            imageWidthConstraint.constant = Constants.defaultImageSize
             imageView.image = icon
         }
 
         if let iconURL = rowData.socialIconURL {
-            imageWidthConstraint.constant = Constants.socialImageSize
             downloadImageFrom(iconURL)
         }
 
         if let iconURL = rowData.userIconURL {
-            imageWidthConstraint.constant = Constants.userImageSize
-            imageView.layer.cornerRadius = Constants.userImageSize * 0.5
+            imageView.layer.cornerRadius = imageSize * 0.5
             imageView.clipsToBounds = true
 
             // Use placeholder image until real image is loaded.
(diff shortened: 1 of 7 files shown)


WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
```diff
@@ -199,21 +201,21 @@ private extension StatsTotalRow {
...
 
         if let iconURL = rowData.socialIconURL {
-            imageWidthConstraint.constant = Constants.socialImageSize
             downloadImageFrom(iconURL)
         }
 
         if let iconURL = rowData.userIconURL {
-            imageWidthConstraint.constant = Constants.userImageSize
-            imageView.layer.cornerRadius = Constants.userImageSize * 0.5
+            imageView.layer.cornerRadius = imageSize * 0.5
```
stevebaranski: I know this code was in place previously, but it occurred to me that some might not understand the intent of this message.

If we're making the image view circular, it _might_ be more clear to set the path for a mask via one of the following methods:

- `init(arcCenter:radius:startAngle:endAngle:clockwise:)`
- `init(ovalIn:)`

I don't feel strongly about this, but wanted to share.

PR #11749, opened 2019-05-22
Keeping image overlay on Action sheet dismissal

### LINKED ISSUE #11428: Offline Support: Inconsistent Overlay
Adding media to a post - you can take a photo to add or insert one from the camera roll. An overlay is displayed over the image “failed to insert media, please tap for options.” Tapping the image brings up an action sheet with the options: remove image, retry, or dismiss. After dismissing the action sheet, the overlay disappears. Tapping on the image again brings up the action sheet. 
Steps:
1. Create new post while offline
2. Select media to upload
3. Tap the media to activate the action sheet
4. Tap outside of the action sheet
5. Return or add characters to the editing space
6. Scroll up and down in the view -- the overlay disappears
--

yaelirub:
When tapping an overlay and dismissing the action sheet by touching the screen out of the image a gesture recognizer is fired which in turn
calls the deselect method from the editor so the attachment gets removed.
I added the method to reprocess the media attachment upon dismissal so in case there is still an error, the attachment will show.

Fixes #11428 

To test:
1. Create new post while offline
2. Select media to upload
3. Tap the media to activate the action sheet
4. Tap outside of the action sheet
5. Return or add characters to the editing space
6. Scroll up and down in the view -- see that the overlay is still displayed 

**Before**

image_overlay1

**After**
image_overlay

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -2972,6 +2972,7 @@ extension AztecPostViewController: TextViewAttachmentDelegate {
         if let mediaAttachment = attachment as? MediaAttachment {
             self.resetMediaAttachmentOverlay(mediaAttachment)
             richTextView.refresh(mediaAttachment)
+            processMediaAttachments()
         }
     }
 


diegoreymendez:
@yaelirub - Shouldn't this PR update `RELEASE-NOTES.txt` since these changes are user-facing?

yaelirub:
@diegoreymendez, I wasn't sure if this one should be included since the overly was already there so it's not a change per se but I will create a PR for that if you think it's necessary

PR #11655, opened 2019-05-08
Allow WordPress.com API URLs to be overridden with a launch argument

jtreanor:
## Description

Currently, `https://public-api.wordpress.com` and similar urls are hardcoded everywhere that we use WordPressKit and WordPressAuthenticator. This makes it very difficult to mock out network requests for testing purposes without resorting to swizzling or similar hacks. Recent changes to those pods (see the PRs below) allow the URLs to be passed in while defaulting to the correct value.

Here is what I have done:

- Added `wordPressComApiBase` to `Environment`. It defaults to `public-api.wordpress.com` but can be overridden.
- In an extension, I have defined `WordPressComRestApi.defaultApi()` which can be used everywhere that we want an API instance. It uses the value from `Environment.current.wordPressComApiBase`.
- Added support for a `-wpcom-api-base-url` launch argument that overrides the API endpoint used.

In a subsequent PR, I will take advantage of this by adding mocking for our UI tests as I have done in https://github.com/wordpress-mobile/WordPress-Android/pull/9422.

Related PRs:

- WordPressKit: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/141
- WordPressAuthenticator: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/93

## Testing:

First, check that nothing is broken:

1. Build and run the app.
2. Do a smoke test, making sure things work as expected.

Now test that the URL injection works:

1. Enable the `-wpcom-api-base-url http://localhost:8080/` launch argument (WordPress -> Edit Scheme -> Run -> Arguments -> Check the box).
2. Build and run the app.
3. Now you will see from the logs that the app is trying (and failing) to reach `http://localhost:8080/` for the API requests. This is correct, as there is nothing running at `http://localhost:8080/`.

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #11762, opened 2019-05-23
Use Environment.current.wordPressComApiBase when initializing WPStatsService

jtreanor:
This extends the changes made in https://github.com/wordpress-mobile/WordPress-iOS/pull/11655 to `WPStatsService`. Once this change is made it will be possible to point the stats service at a different URL. We'll use this for mocking purposes.

To test:

- Smoke test stats to make sure things look as they should.

Update release notes:

### DIFF
--- a/WordPressComStatsiOS/WordPressComStatsiOS/UI/WPStatsViewController.m
+++ b/WordPressComStatsiOS/WordPressComStatsiOS/UI/WPStatsViewController.m
@@ -50,7 +50,7 @@
         self.statsTableViewController = tableVC;
         tableVC.statsDelegate = self.statsDelegate;
         tableVC.statsProgressViewDelegate = self;
-        tableVC.statsService = [[WPStatsService alloc] initWithSiteId:self.siteID siteTimeZone:self.siteTimeZone oauth2Token:self.oauth2Token andCacheExpirationInterval:5 * 60];
+        tableVC.statsService = [[WPStatsService alloc] initWithSiteId:self.siteID siteTimeZone:self.siteTimeZone oauth2Token:self.oauth2Token andCacheExpirationInterval:5 * 60 apiBaseUrlString:[WordPressComRestApi apiBaseURLString]];
     } else if ([segue.identifier isEqualToString:@"InsightsTableEmbed"]) {
         InsightsTableViewController *insightsTableViewController = (InsightsTableViewController *)segue.destinationViewController;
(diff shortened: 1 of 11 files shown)


WordPressComStatsiOS/WordPressComStatsiOS/UI/WPStatsViewController.m
```diff
@@ -50,7 +50,7 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
         self.statsTableViewController = tableVC;
         tableVC.statsDelegate = self.statsDelegate;
         tableVC.statsProgressViewDelegate = self;
-        tableVC.statsService = [[WPStatsService alloc] initWithSiteId:self.siteID siteTimeZone:self.siteTimeZone oauth2Token:self.oauth2Token andCacheExpirationInterval:5 * 60];
+        tableVC.statsService = [[WPStatsService alloc] initWithSiteId:self.siteID siteTimeZone:self.siteTimeZone oauth2Token:self.oauth2Token andCacheExpirationInterval:5 * 60 apiBaseUrlString:[WordPressComRestApi apiBaseURLString]];
```
jtreanor: I haven't figured how we could inject the URL here 🤔

@ScoutHarris Do you know what scenario we could hit this code path and not have a valid `self.statsService` to pass through?

We could get away with leaving this as is, but I would prefer to cover all bases.

WordPressComStatsiOS/WordPressComStatsiOS/UI/WPStatsViewController.m
```diff
@@ -63,7 +63,7 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender
         }
         if (insightsTableViewController.statsService == nil)
         {
-            insightsTableViewController.statsService = [[WPStatsService alloc] initWithSiteId:self.siteID siteTimeZone:self.siteTimeZone oauth2Token:self.oauth2Token andCacheExpirationInterval:5 * 60];
+            insightsTableViewController.statsService = [[WPStatsService alloc] initWithSiteId:self.siteID siteTimeZone:self.siteTimeZone oauth2Token:self.oauth2Token andCacheExpirationInterval:5 * 60 apiBaseUrlString:[WordPressComRestApi apiBaseURLString]];
```
jtreanor: Above comment applies here too.

PR #11757, opened 2019-05-22
Implement Stats Refresh analytics events

stevebaranski:
### Summary
Fixes #11730.

To test:
- Checkout the branch and confirm that existing tests pass, including those in `WPAnalyticsTrackerAutomatticTracksTests`.
- Review the code changes, and confirm that everything looks square.
- As a user, launch the app on-device or in the Simulator, logged in with _Collect Information_ toggle enabled in _Me : App Settings : Privacy Settings_.
- Step through the Stats functionality, taking note that these events are logged* as expected.

It's probably easiest to validate events in one of the following two ways:
1. The event and properties are logged to the console [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m#L72-L76).
1. You can search for the events in post via _Tracks Live View_. Once the events are captured, be sure they are sent across the wire. This is easiest to accomplish by backgrounding the app, initiated [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Automattic-Tracks-iOS/TracksService.m#L269).

Update release notes:

- [X] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

### Discussion
Absent explicit requirements, it took some time & energy to distill the specific changes that needed to be made. For this reason, I've chosen to chronicle my approach below. Please advise if I missed something!

1. Examined the Android implementation of analytics; the PRs for these changes are noted in the issue.
1. Defined the net new events in `WPAnalytics.h` (see `WordPressShared`).
1. Absorbed these new events in WPiOS, ensuring that `WPAnalytics.m` prescribes `eventName` properties that match the aforementioned Android implementation.
1. Implemented the events below:

#### "Legacy" Events
- `WPAnalyticsStatStatsInsightsAccessed`
- `WPAnalyticsStatStatsPeriodDaysAccessed`
- `WPAnalyticsStatStatsPeriodWeeksAccessed`
- `WPAnalyticsStatStatsPeriodMonthsAccessed`
- `WPAnalyticsStatStatsPeriodYearsAccessed`
- `WPAnalyticsStatStatsScrolledToBottom`
- `WPAnalyticsStatStatsSinglePostAccessed`

#### _Latest Post Summary_ Events
- `WPAnalyticsStatStatsItemTappedLatestPostSummaryNewPost`
- `WPAnalyticsStatStatsItemTappedLatestPostSummaryPost`
[...]

PR #11769, opened 2019-05-23
Restored period bar visibility and set to use default period count

stevebaranski:
This PR addresses two issues:

1. It addresses a regression introduced via #11757, where the period bar was not visible on the Period screens. Regrettably, the implementation of the `UIScrollViewDelegate` to capture the "scroll to bottom" analytics event was hijacking `UITableViewDelegate` (which now conforms to `UIScrollViewDelegate`). Restoring the tracked event will be tackled in a subsequent PR.
1. Now that #11758 has been merged, this PR removes the workaround introduced via #11719 to allow the period bar and chart navigation to remain in sync for a longer horizon (i.e., 14 periods).

To test:
- Navigate to Stats for a given site, and review the Period Charts presented. 
- Observe that navigating with the Period bar updates the chart, as well as the highlighted bar when applicable.
- Rinse & repeat for a couple of different periods, dimensions, device types & orientations, confirming no regressions in behavior or appearance.

Update release notes:

Screen Shot 2019-05-23 at 3 43 50 PM

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
@@ -79,8 +79,7 @@ class SiteStatsPeriodTableViewController: UITableViewController {
             return nil
         }
 
-        let periodCount = 10
-        cell.configure(date: selectedDate, period: selectedPeriod, delegate: self, expectedPeriodCount: periodCount)
+        cell.configure(date: selectedDate, period: selectedPeriod, delegate: self)
         viewModel?.statsBarChartViewDelegate = cell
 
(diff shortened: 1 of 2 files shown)


stevebaranski:
For what it's worth, I elected not to preserve the event scroll-to-bottom event here, as it's going to take a little more consideration.

- `UITableViewDelegate` hasn't always, but does now extend `UIScrollViewDelegate`.
- Both `SiteStatsInsightsTableViewController` & `SiteStatsPeriodTableViewController` currently leverage `ImmuTable`, and the `ImmuTableViewHandler` does not currently implement `UIScrollViewDelegate` delegate methods.

@danielebogo with this additional context, would you feel comfortable reviewing this?

stevebaranski:
Thanks for the review @danielebogo!

PR #11776, opened 2019-05-24
Issue/10623 add dependency attribution

markpar:
Fixes wordpress-mobile#10623.

This fix implements the great suggestion from @jtreanor to convert the 3rd-party acknowledgements markdown created for us by CocoaPods into HTML and then display that in a web view. Using the commonmarker gem, we convert the markdown to HTML during the `post_install` phase of `pod install`, adding some lightweight global styling along the way. The resulting HTML is embedded as a resource in the main app bundle as `acknowledgements.html` and this is then displayed in a web view when the user taps `Acknowledgements` on the `About WordPress for iOS` screen.

Unit tests in Xcode (Product > Test) passed.

To test:

* Boot app.
* Me > App Settings > About WordPress for iOS > Acknowledgements.

Expect: A list of licenses for third-party libraries we use. Content should be identical to Pods/Target Support Files/Pods-WordPress/Pods-WordPress-acknowledgements.markdown

Update release notes:

- [x]  If there are user facing changes, I have added an item to RELEASE-NOTES.txt.


markpar:
Here are screenshots for the Acknowledgements entry and the resulting HTML. Styling suggestions welcome!

image

Podfile
```diff
@@ -169,6 +169,24 @@ target 'WordPress' do
...
+        
+        markdown = File.read('Pods/Target Support Files/Pods-WordPress/Pods-WordPress-acknowledgements.markdown')
+        rendered_html = CommonMarker.render_html(markdown, :DEFAULT)
+        styled_html = "<style>
+                         body {
+                           font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+                         }
+                      </style>
+                      <body>
+                        #{rendered_html}
```
markpar: I had the wrong destination repo in my initial PR.  This is a comment by @jtreanor from that discussion: https://github.com/markpar/WordPress-iOS/pull/1#discussion_r287318728

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -2317,6 +2318,7 @@
 		4645AFC41961E1FB005F7509 /* AppImages.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; name = AppImages.xcassets; path = Resources/AppImages.xcassets; sourceTree = "<group>"; };
 		46F84612185A8B7E009D0DA5 /* PostContentProvider.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = PostContentProvider.h; sourceTree = "<group>"; };
 		48690E659987FD4472EEDE5F /* Pods-WordPressNotificationContentExtension.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-WordPressNotificationContentExtension.release.xcconfig"; path = "../Pods/Target Support Files/Pods-WordPressNotificationContentExtension/Pods-WordPressNotificationContentExtension.release.xcconfig"; sourceTree = "<group>"; };
+		4D520D4E22972BC9002F5924 /* acknowledgements.html */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.html; name = acknowledgements.html; path = "../Pods/Target Support Files/Pods-WordPress/acknowledgements.html"; sourceTree = "<group>"; };
```
markpar: Another comment by @jtreanor from the errant PR.

Podfile
markpar: > Could we add a <title> here too so that it is displayed above the Web View?
 
That's a great suggestion!  We can do it in `Podfile`'s `post_install` hook, along with a quick call to `sub` to remove the existing `<h1>Acknowledgements</h1>`.  Here's a (snipped) screenshot:

image

> Of course, this file isn't localized but considering the generated file isn't either, maybe that would be okay. Alternatively, we could add it in the Swift code and use NSLocalizedString.

I started looking into a way to make the title localizable and thought one way to do would be to first render a token (instead of the string `Acknowledgements`) in the `<title>` during `pod install`.  Then later, in Swift, load the HTML into memory and do a find & replace with the localized title string.  

At this point, we could do something like writing the in-memory HTML to a temp file on the device and then pass that URL to `displayWebView`.  This would work, but it feels a little more complicated than it needs to be.   

I found that `WKWebView` has a method `loadHTMLString`; however, it might involve touching `WebKitViewController` and/or `WebViewControllerConfiguration` to support it.  While I am always in favor of localization, this is the point where I would stop and ask others to weigh in on whether it merits further investigation.  🙂

jtreanor: Thanks for digging into the details here.

Let's start simple by adding the `<title>` in `post_install` without localization as you mention and explore those other approaches if we feel there is a need.

megsfulton:
For the rendering of the HTML, just a few small changes to make the text on the page more easily readable. Apologies for not including these details on the original ticket. 

- Add 20px margin/padding to the left, top, and right side so the text isn't against the edge of the screen
- When the body text is displayed on a mobile browser it should have a font size of 16px and a color of #1A1A1A. 

Acknowledgements Redlines

Podfile
markpar: 👍 That sounds like a great plan.

markpar:
Thanks so much!  Here's a screenshot with the new styling.

image

megsfulton:
>I used margin instead of explicitly specifying a top, left, and right, which means we also have a bottom margin of 20px. It can easily be changed to explicit values for top, left, and right if we don't want a bottom margin.

Having a margin on all sides makes sense! 

The text on the screen is still rendering quite small. @markpar could you try 2xing it 32 and see what that looks like?

markpar:
Sure thing, @megsfulton!  

Well, this is weird!  I bumped from `16px` to `32px` and this is the result.  
image

When I try an iPad simulator, it looks good.  This is `32px`, which on iPad looks a little large.  
image

Here's `16px` on an iPad.
image


I'm wondering if this might be an issue with the iPhone simulator?   It appears not to be rendering this web view content correctly, perhaps?  My hypothesis is that your initial `16px` design was right on the money and the simulator is having a problem.  I have a physical iPhone Xr I can try it on, however I'm not on the WP team, so I don't have a provisioning profile and can't deploy it to my device.  

@jtreanor, what are your thoughts on how to proceed?  I am searching the internets to see if this might be a known issue with WKWebView and local HTML content, and am continuing my attempts to deploy to my physical iPhone.

markpar:
Turns out the iPhone simulator is, of course, not the problem at all.  🤦‍♂ 😆   There are loads of `<pre>` tags in the HTML and they don't wrap by default.   Adding some CSS to set the value of `white-space` to `pre-wrap` restores the wrapping functionality.  

After that, there was a single problem child that remained.  The license for `glog` contains this URL `http://www.google.com/codesearch/p?hl=en#dR3YEbitojA/COPYING&amp;q=GetSystemTimeAsFileTime%20license:bsd`, and it was not wrapping.  All the HTML and CSS voodoo I tried did not cause the URL to wrap.  I finally resorted to inserting a hardcoded `<br>` and that did the trick.

Once that was taken care of, I added a `meta` tag for the `viewport` and the resulting HTML is the best so far.  @megsfulton, would love your feedback!

iPhone
image
 
iPad
image

megsfulton:
Looks great now! 🎉

markpar:
Awesome!  Thanks so much!  @jtreanor, I think we are ready to merge.  👍 🙂

PR #11783, opened 2019-05-25
Implements an initial version of UploadsManager class and Uploader protocol

### LINKED ISSUE #11653: Offline Support: Implement UploadsManager.

### LINKED ISSUE #11694: Define the Uploader protocol.
## Description:

This issue was spawned from the discussion here: https://github.com/wordpress-mobile/WordPress-iOS/issues/11640

Based on this graph:

57587998-167a7380-74e4-11e9-925f-408b4636f77d

We need to define the `Uploader` protocol.

diegoreymendez:
This PR implements an initial version of the `UploadsManager` class and the `Uploader` protocol.  It also wires the events that should trigger the `UploadsManager.

**Important:** This PR is aimed at a feature branch because we still have to do more work before this is ready for production.  Other PRs to complete this feature should target the same destination branch.

**Important 2:** There was some discussion on Friday with @yaelirub that @jklausa would be taking some of this over.  The reason why I pushed this over the weekend is because I had some of this code already implemented and it felt bad to duplicate the effort.  The feature branch should allow us to sync on Monday to see how to move this forward together next week.  Apologies if this was confusing @yaelirub and @jklausa!

## Affected issues

Closes #11653 
Closes #11694 

## What's included and what's not?

This PR set things up so we can resume uploads due to certain triggers.

This PR does NOT include the logic to make the uploads resume yet.  This should be tackled incrementally into this feature branch before we merge back to `develop`.

I've opened these issues to track the works that remains to be done:

https://github.com/wordpress-mobile/WordPress-iOS/issues/11784
https://github.com/wordpress-mobile/WordPress-iOS/issues/11785

## Testing:

Since this code isn't really doing anything yet, the changes proposed here should be mostly harmless.  We're also targetting a feature branch so we can do further testing before merging back to develop.


jklausa:
This looks really good and mostly what I had in mind too! 

One thing I'm not sure about: this could be extended to incorporate the responsibilities of `ShareExtensionSessionManager`, right? 

I was thinking hoping we could also piggy-back off of already-existing `MediaUploadOperation` / `MediaUploadOperation` — I think the models might need changing a bit, but they seem to be along the lines of what we'll need here.

diegoreymendez:
The question in my mind is if we need to record the pending uploads in the database (like `MediaUploadOperation`), or if we can figure out a way to query our existing records to know what needs to be uploaded and how.

Both posts and media have a `remoteStatus` property, which should make it possible for us to know if something needs to be uploaded (I believe).

jklausa:
I'm torn on this. 

On one hand, I see your point about us already sorta-kinda having the required information (`remoteStatus`), without needing to duplicate it in another place `UploadOperation`.

Having the information split in two places can lead to weird bugs when those data sources get out of sync down the road, and in general, it feels _bad_ to have that information duplicated.

On the other hand, look just how many fields does does `UploadOperation` define:

```
@objc(UploadOperation) public class UploadOperation : NSManagedObject {

    @NSManaged private var uploadStatus: Int32

    @NSManaged public var groupID: String?

    @NSManaged public var backgroundSessionTaskID: Int32

    @NSManaged public var backgroundSessionIdentifier: String?

    @NSManaged public var siteID: Int64

    @NSManaged public var created: NSDate?
}
```
We probably won't need _all of those_, but I imagine needing to keep track of `backgroundSessionTaskID`, `backgroundSessionIdentifier` and `groupID` will end up something we'll need to tackle. And just shoving those back onto `Post` or `Media` which already have too many properties/responsibilities doesn't feel great either.

I think I'm slightly leaning towards the later solution, since (and I know this is probably premature-optimisation), we could reuse similar architecture to handle all kinds of retry-able network operations — I could see something like `LikeUploadOperation` or `CommentUploadOperation` popping up in the future.

diegoreymendez:
I'll merge even if Peril is falining, because:

1. The error shown gives me no chance to fix the issue.
2. I'm merging to a feature branch so there's no risk for now.

Issue #11646, opened 2019-05-06
Post List Revamp Master Issue


leandroalonso:
This is a master issue to track the post list revamp on iOS.

# Changes

* View toggle to switch between card/list view
* Updated card designs

# Blueprint

![post_list_blueprint](https://user-images.githubusercontent.com/7040243/57258653-6ca66d00-7033-11e9-83e3-51f01f179cba.png)

# Subtasks

- [x] Replace the "More" behavior to show options in an Action Sheet #11661
- [x] Update card designs and do not display author name when "Me" is selected https://github.com/wordpress-mobile/WordPress-iOS/pull/11705
- [x] Display "skeleton post card" when loading posts #11789
- [x] Adapt cards so it can be displayed in a compact way #11816
- [x] Change search to always happen in a list view #11830
- [x] Enable the list view option #11816
- [x] Improvements and polishes #11974
- [ ] Merge with `develop` and integrate

leandroalonso:
Note: action sheets are inverted for `scheduled` and `trashed` posts in the blueprint. "View" option only appears when showing the compact mode.

Issue #9993, opened 2018-08-15
Site Pages and Posts: no results view displays before loading view


ScoutHarris:
### Expected behavior
On a slow network, when selecting Site Pages or Blog Posts, the loading view should show first.

### Actual behavior
On a slow network, when selecting Site Pages or Blog Posts, the "you don't have any xxx" view displays first, then the loading view, then the final view.

I'm noting both Site Pages and Blog Posts here because their logic and behavior is the same. 

Because `syncHelper.isSyncing` is `false` initially, the no results view displays immediately. Of course when it updates to `true`, _then_ the loading view is displayed.

### Steps to reproduce the behavior
- On a slow network, select either Site Pages or Blog Posts.
- Notice the views displayed:

![nrv](https://user-images.githubusercontent.com/1816888/44177512-853cb480-a0ab-11e8-81be-12f65422ab4d.gif)

designsimply:
I tested from a slow connection and found that I didn't see the no results view before the loading view when opening the posts list using WP Internal 13.1.0.20190814 on iPhone 6S iOS 12.4. ([1m27s](https://cloudup.com/coqj4I0j8Qi)) 

I believe this was fixed when the posts list was revamped in 12.9 (see https://github.com/wordpress-mobile/WordPress-iOS/pull/12087).

<img src="https://user-images.githubusercontent.com/1119271/63185714-fd6bec00-c017-11e9-9ad8-c4e225e4c6c3.PNG" width="270" alt="IMG_2934"> <img src="https://user-images.githubusercontent.com/1119271/63186277-4bcdba80-c019-11e9-86b7-c65fd6f101bc.jpg" width="270" alt="fetching-pages">
<sup>Tested with WP Internal 13.1.0.20190814 on iPhone 6S iOS 12.4.</sup>

PR #11789, opened 2019-05-28
Display "ghost post card" when loading posts

leandroalonso:

This PR shows a "ghost post card" when loading posts:

To test:
- Check if the ghost cells appear when loading fresh content
- All the previous error and messages showed by `NoResultsController` should remain the same
- Check it in iPad and with RtL screens


SylvesterWilmott:
Looks great @leandroalonso , thanks!

leandroalonso:
@nheagy currently, in production, the loader of NoResultsController is also flashed when there's no content:



I deliberately added this behavior to override the previous loader. If I remove it, the behavior shown on the previous image will happen.

If we want no loader at all, this will need extra work on the `NoResultsViewController` and will affect all the other screens (Pages, Comments...).

What do you think @SylvesterWilmott?

SylvesterWilmott:
>I deliberately added this behavior to override the previous loader. If I remove it, the behavior shown on the previous image will happen. If we want no loader at all, this will need extra work on the NoResultsViewController and will affect all the other screens (Pages, Comments...).

I think replacing the current loader is what is expected from this PR, if the flashing is part of a more fundamental issue it shouldn't be tackled here.

ScoutHarris:
FWIW, I think this is related: https://github.com/wordpress-mobile/WordPress-iOS/issues/9993.

PR #11788, opened 2019-05-28
Network mocking for all UI and screenshot tests

jtreanor:
This PR updates all of our UI and screenshot tests to use mocked networking using WireMock. The mocking functionality is shared with WPAndroid from the [WordPressMocks repo](https://github.com/wordpress-mobile/WordPressMocks).

It relies on the following open PRs:

- https://github.com/wordpress-mobile/WordPressMocks/pull/3

Other related PRs:

- https://github.com/wordpress-mobile/WordPress-iOS/pull/11655
- https://github.com/wordpress-mobile/WordPress-iOS/pull/11717
- https://github.com/wordpress-mobile/WordPress-iOS/pull/11762
- https://github.com/wordpress-mobile/WordPress-iOS/pull/11807
- https://github.com/wordpress-mobile/WordPress-Android/pull/9422
- https://github.com/wordpress-mobile/WordPressKit-iOS/pull/141
- https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/93
- https://github.com/wordpress-mobile/WordPressMocks/pull/1

To test:

- Check out this branch. Run `rake dependencies` as usual.
- Run `rake mocks` to start the mock server.
- Run the UI tests in the `WordPressUITests` target with Xcode.
- All tests should pass and there should be no errors/warnings in the terminal when you ran the mock server.

**Note:** The screenshot tests are broken in `develop` due to the recent stats changes. I think we should fix that in a separate change.

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #11807, opened 2019-05-29
Fix WordPressScreenshotGeneration build errors

jtreanor:
Fixes persistent build errors when running the screenshot tests from Xcode. It will fix errors that look like this:

image (1).

It includes two fixes:

- A new file, `Logger.swift`, was added to the `WordPressUITests` target. I've added it to `WordPressScreenshotGeneration` too to prevent build failures.
- For some reason, the `WordPressScreenshotGeneration` scheme was set to build and run with the `Release` configuration by default. However, it tests `WordPress` in the DEbug configuration so CocoaPods gets mixed up with linking as some things are `Debug` and some things are `Release`. We didn't spot this because the tests are usually run with `fastlane` which passes an explicit build configuration. It now defaults to `Debug`.

This took me way longer to figure out than I'd like to admit 😅 

To test:

- Run the screenshot tests the target from Xcode (select the `WordPressScreenshotGeneration` scheme and press ⌘+U).

There may be some failures due to flakiness that will hopefully be solved by https://github.com/wordpress-mobile/WordPress-iOS/pull/11788.

Update release notes:


jkmassel:
Merged this myself to close it for the `12.6 milestone`. Thanks @jtreanor!!

PR #11811, opened 2019-05-29
Fix Notices from hidden ViewControllers being shown in the Editor

### LINKED ISSUE #11810: Notice from Post List is shown even when editing a Post
### Expected behavior

The notice should not be displayed.

### Actual behavior

The notice is displayed. The message is out of context.

1. Go to Post List and edit a Post. 
2. After the Editor is shown, go offline. 

Tested on iPhone XS, iOS 12.2, WPiOS develop (e8ccef3).

shiki:
Fixes #11810.  There is a corresponding PR here: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/35.

## Findings

There are 2 sources of the Notices that are shown in the Editor:

- `PostListViewController`
- `ReaderStreamViewController`

They are both missing an `isViewOnScreen()` check in `NetworkAwareUI.shouldPresentAlert()`. 

The `isViewOnScreen()` also does not detect whether the ViewController (A) is presenting another ViewController (B), which essentially makes ViewController (A) not visible and shouldn't show a Notice. I updated `isViewOnScreen()` in https://github.com/wordpress-mobile/WordPressUI-iOS/pull/35. 

## Solution

I changed the `NetworkAwareUI` protocol to require `UIViewController` so we can add `isViewOnScreen()` in the default `shouldPresentAlert()` implementation. With this, we don't have to remember to override `shouldPresentAlert()` in new ViewControllers. 

| Before | After |
|--------|-------|
|        |  |

### Scenario 1

1. Go to Post List and edit a Post. 
2. After the Editor is shown, go offline. No Notice should be shown.

### Scenario 2

1. Go to Post List 
2. Go offline. A Notice should be shown.
3. Immediately edit a Post. The Notice should disappear. 

## Release Notes


WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
```diff
@@ -1089,8 +1089,13 @@ class AbstractPostListViewController: UIViewController,
         return ReachabilityUtils.noConnectionMessage()
     }
 
-    func shouldPresentAlert() -> Bool {
-        return !connectionAvailable() && !contentIsEmpty() && isViewOnScreen()
+    // MARK: - Others
+
+    override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        // Override to hide Notice when we are showing the Editor. We hide Notices in
```
yaelirub: I think the comment is not very clear (or at least its not to me 😬 )
I would add "This override purpose is..." or something like that.  and "Ths noticce should be dismissed/ hidden  in viewDidDisappear" 

>but it is not called when presenting a ViewController.

What is "it"?

WordPress/Classes/ViewRelated/System/NetworkAware.swift
```diff
@@ -22,7 +22,7 @@ extension NetworkAware {
 
 extension NetworkAwareUI {
```
yaelirub: Instead of conforming the protocol to UIViewController can we specify on the extension:
`extension NetworkAwareUI where Self: UIViewController`
?

shiki: 🤦‍♂ That is so much better. Thank you for spotting that! Fixed in 8132aa3bf7d61ae0408e85a3ea46ac1909a1ed31.

WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
shiki: Thank you for letting me know. I updated it in 9bea3d94bb25fe026cc1312d7de10c5e2d91e9e4. How does that look? The mention of `viewDidDisappear` is out of context so I just removed it.

shiki:
@yaelirub Thank you for the feedback. This is ready for another look.

PR #11816, opened 2019-05-30
Adds the option to display posts in a compact way

leandroalonso:

This PRs adds an option to see the post list in a more compact way:

And also:

* Always displays the posts cells based on the user choice (compact or card);
* Shows a compact ghost cell if the user chooses the compact view;
* When seeing compact cells, the action sheet always shows a "View" option

To test:
- Check if the ghost cells appear when loading fresh content;
- Check if your preference is saved: if you select compact cells it should always display compact cells unless you change it - the default is the comfortable cell;
- Check if a "View" option is shown in the action sheet for compact cells - but not for the card cell;


WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -355,19 +407,21 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
...
 
         configurePostCell(cell)
+        configureRestoreCell(cell)
     }
 
     fileprivate func cellIdentifierForPost(_ post: Post) -> String {
         var identifier: String
 
         if recentlyTrashedPostObjectIDs.contains(post.objectID) == true && filterSettings.currentPostListFilter().filterType != .trashed {
-            identifier = type(of: self).postCardRestoreCellIdentifier
```
nheagy: Ha, I missed this first time around. Nice cleanup.

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -376,10 +430,15 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
     private func configurePostCell(_ cell: UITableViewCell) {
         guard let cell = cell as? PostCell else { return }
 
-        cell.setActionSheetDelegate(self)
         cell.isAuthorHidden = showingJustMyPosts
     }
 
+    private func configureRestoreCell(_ cell: UITableViewCell) {
+        guard let cell = cell as? RestorePostTableViewCell else { return }
```
nheagy: We prefer our `return` statements on their own line, even for a simple guard

SylvesterWilmott:
Looks great @leandroalonso !

PR #11820, opened 2019-05-30
[Crash] NSRangeException in BlogDetailsViewController

### LINKED ISSUE #11819: Crash: NSRangeException in BlogDetailsViewController
Sentry Issue: https://sentry.io/share/issue/96a43e29c8e645079d5fc76cd31612cb/

```
NSRangeException: None
  File "BlogDetailsViewController.m", line 507, in -[BlogDetailsViewController setRestorableSelectedIndexPath:]
  File "BlogDetailsViewController.m", line 243, in -[BlogDetailsViewController decodeRestorableStateWithCoder:]
  File "main.m", line 11, in main
...
(38 additional frame(s) were not displayed)
```

danielebogo:
Fixes #11819 

This fixes a crash I introduced in _12.4_. It happened when the `restorableSelectedIndexPath` decoded by `- (void)decodeRestorableStateWithCoder:(NSCoder *)coder` is out of bounds with the `tableSections`.

To fix that I check in the section index is out of bound or not.

I wasn't able to reproduce it, so I forced it setting different values to the `restorableSelectedIndexPath` property when `- (void)decodeRestorableStateWithCoder:(NSCoder *)coder` is called.

Like this:
```
- (void)decodeRestorableStateWithCoder:(NSCoder *)coder
{
//    NSIndexPath *indexPath = [coder decodeObjectForKey:WPBlogDetailsSelectedIndexPathKey];
//    if (indexPath) {
        self.restorableSelectedIndexPath = [NSIndexPath indexPathForRow:0 inSection:999999];
//    }

    [super decodeRestorableStateWithCoder:coder];
}
```

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
@@ -504,7 +504,7 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 {
     _restorableSelectedIndexPath = restorableSelectedIndexPath;
 
-    if (restorableSelectedIndexPath != nil) {
+    if (restorableSelectedIndexPath != nil && restorableSelectedIndexPath.section < [self.tableSections count]) {
         BlogDetailsSection *section = [self.tableSections objectAtIndex:restorableSelectedIndexPath.section];
         self.selectedSectionCategory = section.category;
     }
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
@@ -506,7 +506,19 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 
     if (restorableSelectedIndexPath != nil && restorableSelectedIndexPath.section < [self.tableSections count]) {
         BlogDetailsSection *section = [self.tableSections objectAtIndex:restorableSelectedIndexPath.section];
-        self.selectedSectionCategory = section.category;
+        switch (section.category) {
+            case BlogDetailsSectionCategoryQuickStart:
+            case BlogDetailsSectionCategoryDomainCredit: {
+                self.selectedSectionCategory = BlogDetailsSectionCategoryGeneral;
+                NSUInteger section = [self findSectionIndexWithSections:self.tableSections category:self.selectedSectionCategory];
+                _restorableSelectedIndexPath = [NSIndexPath indexPathForRow:0 inSection:section];
+            }
+                break;
+                
+            default:
+                self.selectedSectionCategory = section.category;
+                break;
+        }
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
@@ -514,7 +514,6 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
                 _restorableSelectedIndexPath = [NSIndexPath indexPathForRow:0 inSection:section];
             }
                 break;
-                
             default:
                 self.selectedSectionCategory = section.category;
                 break;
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
```diff
@@ -504,9 +504,20 @@ - (void)setRestorableSelectedIndexPath:(NSIndexPath *)restorableSelectedIndexPat
 {
     _restorableSelectedIndexPath = restorableSelectedIndexPath;
```
shiki: I have a somewhat related concern with `_restorableSelectedIndexPath` being modified in its `- (NSIndexPath *)restorableSelectedIndexPath` `getter`. It does not look like that will affect the crash though.

shiki:
> Should we set the restorable value at all if it's not valid, namely restorableSelectedIndexPath.section < [self.tableSections count]?

I tested this part by giving it an invalid value but it didn't seem to matter. 🤷‍♂

WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
```diff
@@ -504,9 +504,20 @@ - (void)setRestorableSelectedIndexPath:(NSIndexPath *)restorableSelectedIndexPat
...
 
-    if (restorableSelectedIndexPath != nil) {
+    if (restorableSelectedIndexPath != nil && restorableSelectedIndexPath.section < [self.tableSections count]) {
         BlogDetailsSection *section = [self.tableSections objectAtIndex:restorableSelectedIndexPath.section];
-        self.selectedSectionCategory = section.category;
+        switch (section.category) {
```
jkmassel: Is it possible for Quick Start or Domain Credit to be removed between launches? As a for-instance, if a user redeemed a domain credit in Calypso, or did Quick Start on another device? 

I don't understand the fine details of this code, but just wanted to check on that possibility.

nheagy: > Quick Start on another device

fwiw, this scenario isn't possible, as Quick Start doesn't sync between devices. The Quick Start menu should require the user to specifically dismiss it, however I'm not sure how that might interact with state restoration. Perhaps save state with QS, dismiss QS, then crash the app before it can save state?

danielebogo:
@shiki @elibud I pushed a little commit.
> Should we set the restorable value at all if it's not valid, namely restorableSelectedIndexPath.section < [self.tableSections count]?

Yes I set the value to nil in this case since the getter set returns a _default value_ when its ivar is nil.

> I have a somewhat related concern with _restorableSelectedIndexPath being modified in its - (NSIndexPath *)restorableSelectedIndexPath getter. It does not look like that will affect the crash though.

The crash seems related to the setter only. We left the getter to return/set a default value if its ivar is nil to avoid the previous hardcode value. Also because the _Stats_ section can change position and because the `decodeRestorableStateWithCoder` doesn't set anything if the restored value is nil.

WordPress/Classes/ViewRelated/Blog/BlogDetailsViewController.m
shiki: I played with Quick Start for a bit. I couldn't make it crash.

danielebogo: I think that might be a scenario only with credit domain. Like save the state with a domain credit, make a purchase online and then open the app again. The stored index should be out of bounds. As @nheagy said not with QS.

jkmassel: Wasn't so much thinking about a crash, as that it might restore in an invalid state if the index wasn't out of bounds, but wasn't valid either. If the index for domain credit is `2`, and we store that, then it goes away, and now "Site Pages" is at index `2`, we might be restoring the wrong thing?

Even if that _is_ the case, it's probably not critical right now though.

danielebogo: >  If the index for domain credit is 2, and we store that, then it goes away, and now "Site Pages" is at index 2, we might be restoring the wrong thing?

With QS and Domain credit it shouldn't happen. But it can with another section, since the value stored is an index and not something else that can be addressed to the right section. We added a small map/inverse map to handle the dynamic section (Domain credit and QS) but it needs to be improved (and extended).

danielebogo:
Thanks everyone for testing this!

PR #11830, opened 2019-05-31
Change post list search to always happen in a list view

leandroalonso:
Depends on https://github.com/wordpress-mobile/WordPress-iOS/pull/11816

This PR changes the post list search. As follow:

* Always display results in a compact cell
* Group posts by status (Published, Pending, Trashed)

search

To test:
- Check if search **always** display compact cells
- The behavior of the search should remain the same
- Test in iPad and RtL screens


leandroalonso:
About issue 1: we're using the same header as "Site Pages" section uses. If we're going to change that, it would be good to apply this change in both Pages and Posts section to keep it consistent. Do you agree, @SylvesterWilmott? What about the dimensions, any change?

@nheagy I wasn't able to reproduce the 2nd issue. :( Can you describe the steps or paste your debug output here?

SylvesterWilmott:
>The headings should be Published and Drafts. Please consult the mockup on these headings, as they also appear to not have the same vertical spacing as the mockups.

That's correct, although as @leandroalonso stated, we're using these labels in Pages they are actually incorrect. This should be changed in both places. @leandroalonso let's use the labels "Published" and "Drafts" for this PR.

The spacing looks good, these headers should mimic the design and behaviour of those found in "Activity Log".

nheagy:
> these headers should mimic the design and behaviour of those found in "Activity Log"

@leandroalonso there is `ActivityListSectionHeaderView` that can be reused here.

> I wasn't able to reproduce the 2nd issue

I'll try to recreate it, too 🤞

PR #11832, opened 2019-05-31
Fixes warnings regarding deprecated api.

aerych:
Fixes a couple of warnings about calling deprecated API : `automaticallyAdjustsScrollViewInsets`. 

In `MenuItemEditingViewController.m` the call is simply removed as it doesn't appear its needed.
In `MediaLibraryViewController.swift` the call is replaced with the recommended substitute.

To test:
Note how the UI looks, paying particular attention to the scroll views, on the following two screens:
- My Sites > Site > Menus > Select a Menu > Tap to Edit one of the menu's items. 
- My Sites > Site > Media
Check on an iPad and on an iPhone.  Bonus for also checking on the XS Max.

Apply the patch and check the screens again.  There should be no noticeable difference.
Note that there is some preexisting unrelated UI wonkiness with Menus. :(  I'll open issues for these when I get time.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Media/MediaLibraryViewController.swift
+++ b/WordPress/Classes/ViewRelated/Media/MediaLibraryViewController.swift
@@ -83,7 +83,7 @@ class MediaLibraryViewController: WPMediaPickerViewController {
 
         title = NSLocalizedString("Media", comment: "Title for Media Library section of the app.")
 
-        automaticallyAdjustsScrollViewInsets = false
+        collectionView?.contentInsetAdjustmentBehavior = .never
 
         registerChangeObserver()
         registerUploadCoordinatorObserver()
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Media/MediaLibraryViewController.swift
```diff
@@ -83,7 +83,7 @@ class MediaLibraryViewController: WPMediaPickerViewController {
 
         title = NSLocalizedString("Media", comment: "Title for Media Library section of the app.")
 
-        automaticallyAdjustsScrollViewInsets = false
+        collectionView?.contentInsetAdjustmentBehavior = .never
```
ScoutHarris: Interestingly, this change makes the scroll indicator disappear. Reverting it shows the scroll indicator.

From `develop`:
indicator

I'm not sure it's worth holding up this change though. I'll leave it to you; just let me know how you'd like to proceed.

aerych: Well spotted!  It looks like it might be running behind the search bar when its scrolled to the top.  Setting the property to `.automatic` seems to resolve the issue.   Since this is the default value I think we can just get rid of the line altogether.

aerych: Updated! Ready for another peek!

aerych:
Thanks @ScoutHarris !!

PR #11839, opened 2019-06-01
Implements Uploader protocol in MediaCoordinator and PostCoordinator.

### LINKED ISSUE #11784: Implement UploaderProtocol in MediaCoordinator.
Implement the [`Uploader` protocol](https://github.com/wordpress-mobile/WordPress-iOS/blob/9585b9a0cc7b3b77c23b398df7c76982743d58bc/WordPress/Classes/Services/MediaCoordinator.swift#L9) in `MediaCoordinator`.

This implementation should:

- Get a list of media that need to be synchronized.
- For those media items that have running uploads, do nothing.
- For those media items that have no running uploads, make sure to upload them.

### LINKED ISSUE #11785: Implement UploaderProtocol in PostCoordinator.
Implement [the Uploader protocol](https://github.com/wordpress-mobile/WordPress-iOS/blob/9585b9a0cc7b3b77c23b398df7c76982743d58bc/WordPress/Classes/Services/PostCoordinator.swift#L233) in PostCoordinator.

This implementation should:

- Get a list of posts that need to be synchronized.
- For those posts that have running uploads, do nothing.
- For those posts that have no running uploads, make sure to upload them.

diegoreymendez:
Fixes #11785 
Fixes #11784 

Implements the Uploader protocol in both coordinator, and wires it so that it's triggered when the App is opened, when it comes to foreground, and when `Reachability` detects there's a connection.

Update release notes:

## Known Issues / Scope limitations:

Reachability isn't working very well, but from my investigation this is not related to my PR.  Reachability is simply failing to call the closures it should call.

I'll need to [investigate this separately](https://github.com/wordpress-mobile/WordPress-iOS/issues/11840#issuecomment-498194915), as it goes beyond the scope of this specific PR.

### Test 1:

1. Create a post with media while offline.
2. Put the app in the background.
3. Go back online.
4. Bring the app back to foreground.
5. Make sure the post is uploaded.

### Test 2:

1. Create a post with media while offline.
2. Close the App.
3. Go back online
4. Launch the app again.
5. Make sure the post is uploaded.

### DIFF
--- a/WordPress/Classes/Services/MediaCoordinator.swift
+++ b/WordPress/Classes/Services/MediaCoordinator.swift
@@ -5,11 +5,7 @@ import WordPressFlux
 /// items, independently of a specific view controller. It should be accessed
 /// via the `shared` singleton.
 ///
-class MediaCoordinator: NSObject, Uploader {
-    func resume() {
-    }
-
-
+class MediaCoordinator: NSObject {
     @objc static let shared = MediaCoordinator()
 
     private(set) var backgroundContext: NSManagedObjectContext = {
@@ -602,6 +598,22 @@ extension MediaCoordinator: MediaProgressCoordinatorDelegate {
 
         if mediaProgressCoordinator != mediaLibraryProgressCoordinator {
             removeCoordinator(mediaProgressCoordinator)
+        }
+    }
+}
+
+extension MediaCoordinator: Uploader {
+    func resume() {
+        let service = MediaService(managedObjectContext: mainContext)
+
+        service.getFailedMedia { [weak self] mediaArray in
+            guard let self = self else {
+                return
+            }
+
+            for media in mediaArray {
+                self.retryMedia(media)
+            }
         }
     }
 }
--- a/WordPress/Classes/Services/MediaCoordinator.swift
+++ b/WordPress/Classes/Services/MediaCoordinator.swift
@@ -606,14 +606,12 @@ extension MediaCoordinator: Uploader {
     func resume() {
         let service = MediaService(managedObjectContext: mainContext)
 
-        service.getFailedMedia { [weak self] mediaArray in
+        service.getFailedMedia { [weak self] media in
             guard let self = self else {
                 return
             }
 
-            for media in mediaArray {
-                self.retryMedia(media)
-            }
+            media.forEach() { self.retryMedia($0) }
         }
     }
 }
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -255,9 +255,7 @@ extension PostCoordinator: Uploader {
                 return
             }
 
-            for post in posts {
-                self.retrySave(of: post)
-            }
+            posts.forEach() { self.retrySave(of: $0 ) }
         }
     }
 }
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -248,9 +248,16 @@ class PostCoordinator: NSObject {
 
 extension PostCoordinator: Uploader {
     func resume() {
-        // Resume the upload of all posts that are not synched.
-        //
-        // 1. Query posts with status == .failed
-        // 2. Call retrySave() for each post
+        let service = PostService(managedObjectContext: mainContext)
+
+        service.getFailedPosts { [weak self] posts in
+            guard let self = self else {
+                return
+            }
+
+            for post in posts {
+                self.retrySave(of: post)
+            }
+        }
     }
 }
--- a/WordPress/Classes/Uploads/UploadsManager.swift
+++ b/WordPress/Classes/Uploads/UploadsManager.swift
@@ -1,27 +1,11 @@
 import Foundation
+import Reachability
 
 /// Takes care of coordinating all uploaders used by the app.
 ///
 class UploadsManager: NSObject {
     private let uploaders: [Uploader]
-
-    /// The reason why this property is lazy is that it's basically a closure with a self reference.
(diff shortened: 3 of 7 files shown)


yaelirub:
Hey @diegoreymendez, wonderful work here!
Test 1:
The media upload resumed when the app returned to foreground but the image overlay stayed.
media_background

Test 2:
media_from_load
This seem to happen only in Aztec. Gutenberg works

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -606,6 +602,22 @@ extension MediaCoordinator: MediaProgressCoordinatorDelegate {
...
+extension MediaCoordinator: Uploader {
+    func resume() {
+        let service = MediaService(managedObjectContext: mainContext)
+
+        service.getFailedMedia { [weak self] mediaArray in
+            guard let self = self else {
+                return
+            }
+
+            for media in mediaArray {
```
yaelirub: A suggestion for a 1 liner: 
`mediaArray.forEach {self.retryMedia($0)}`

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -248,9 +248,16 @@ class PostCoordinator: NSObject {
...
-        // 1. Query posts with status == .failed
-        // 2. Call retrySave() for each post
+        let service = PostService(managedObjectContext: mainContext)
+
+        service.getFailedPosts { [weak self] posts in
+            guard let self = self else {
+                return
+            }
+
+            for post in posts {
```
yaelirub: same 1 line comment from above.

WordPress/Classes/Uploads/UploadsManager.swift
```diff
@@ -2,9 +2,12 @@ import Foundation
 
 /// Takes care of coordinating all uploaders used by the app.
 ///
-@objc
 class UploadsManager: NSObject {
     private let uploaders: [Uploader]
+
+    /// The reason why this property is lazy is that it's basically a closure with a self reference.
```
yaelirub: Are you avoiding using Reachability cause we are planning on abandoning it?

yaelirub:
>Reachability isn't working very well, but from my investigation this is not related to my PR. Reachability is simply failing to call the closures it should call.

Diego, have you looked into using [NWPathMonitor](https://www.hackingwithswift.com/example-code/networking/how-to-check-for-internet-connectivity-using-nwpathmonitor) from Apple's Network framework?

WordPress/Classes/Uploads/UploadsManager.swift
```diff
@@ -31,6 +34,9 @@ class UploadsManager: NSObject {
         self.uploaders = uploaders
 
         super.init()
+
+        /// This just makes sure the reachabilityObserver property is initialized here.
+        _ = reachabilityObserver
     }
 
     deinit {
```
yaelirub: AFAIK You don't have to remove the observer since we are targeting ios > 9.0. From Apple docs:

> If your app targets iOS 9.0 and later or macOS 10.11 and later, you don't need to unregister an observer in its dealloc method. Otherwise, you should call this method or removeObserver(_:name:object:) before observer or any object specified in addObserver(forName:object:queue:using:) or addObserver(_:selector:name:object:) is deallocated.

https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

just in case that's the reason you are saving it as a property

WordPress/Classes/Uploads/UploadsManager.swift
diegoreymendez: Apologies!  I forgot to push my last commit, which made some changes to how `UploadsManager` uses reachability.

WordPress/Classes/Uploads/UploadsManager.swift
diegoreymendez: Apologies! I forgot to push my last commit, which made some changes to how UploadsManager uses reachability.

WordPress/Classes/Services/MediaCoordinator.swift
diegoreymendez: Changed.

WordPress/Classes/Services/PostCoordinator.swift
diegoreymendez: Changed.

diegoreymendez:
> Diego, have you looked into using NWPathMonitor from Apple's Network framework?

Only available in iOS 12 or newer, which makes it a no-go :(

One other option is probably [this](https://github.com/ashleymills/Reachability.swift) but I'd like to keep that as a separate task since I think fixing reachability is beyond the scope of this specific PR.

I've opened this issue for it: https://github.com/wordpress-mobile/WordPress-iOS/issues/11843

diegoreymendez:
@yaelirub - Awesome feedback... I've opened a new issue to tackle that problem separately here: https://github.com/wordpress-mobile/WordPress-iOS/issues/11846.

I've decided to split it because it's going to be easier for me to make progress on this incrementally, and because this issue is only present in the feature branch.

Rest assured we'll make sure everything is working before merging the feature branch.

Thanks for the review!

WordPress/Classes/Uploads/UploadsManager.swift
```diff
@@ -31,17 +18,24 @@ class UploadsManager: NSObject {
...
+        setupReachableBlock()
     }
 
-    deinit {
-        NotificationCenter.default.removeObserver(reachabilityObserver)
+    // MARK: Reachability
+
+    private func setupReachableBlock() {
+        reachability.reachableBlock = { _ in
+            self.resume()
```
yaelirub: Should this be a weak self?

diegoreymendez: Yep.  Will submit a PR to fix.

PR #11851, opened 2019-06-04
Make universal link router values non optional

diegoreymendez:
`UniversalLinkRouter` values were not optional even before this PR.  The `NavigationAction` protocol [was allowing `values` to be optional](https://github.com/wordpress-mobile/WordPress-iOS/blob/7d72a3b99c8e5357d0ed6157c94e1fe2b590bf19/WordPress/Classes/Utility/Universal%20Links/Route.swift#L16), causing unnecessary overhead in the classes that implemented this protocol.

This PR simply modifies the protocol so that [the `values` parameter is no longer optional](https://github.com/wordpress-mobile/WordPress-iOS/blob/3e2037011ff30859d19af92003f7a78674e12ed4/WordPress/Classes/Utility/Universal%20Links/Route.swift#L16), and changes the implementations of this protocol to adopt the change.

## Testing:

Make sure the code builds, and that unit tests don't fail.

Update release notes:


diegoreymendez:
Ty @frosty

PR #11857, opened 2019-06-04
Change quick start highlight color to white

### LINKED ISSUE #11688: Site Creation: Blue text in notices can't be tapped.
## Description:

The blue text shown by the notices in the Site Setup guided steps can't be tapped, but looks like it can be.

## Image:

IMG_9E2B15E903E0-1

nheagy:
Fixes #11688 

Changes the Quick Start notices to have a white highlight

image

To test:
- Start a Quick Start tour (let me know if you need help with this)
- Ensure the notices use white as a highlight color

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTours.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTours.swift
@@ -349,7 +349,7 @@ private extension String {
     private enum Constants {
         static let iconOffset: CGFloat = 1.0
         static let iconSize: CGFloat = 16.0
-        static let highlightColor = WPStyleGuide.lightBlue()
+        static let highlightColor: UIColor = .white
         static var highlightFont: UIFont {
             get {
                 return WPStyleGuide.fontForTextStyle(.subheadline, fontWeight: .semibold)


nheagy:
Looks like the font should be bold, too. Also appears that `WPStyleGuide.fontForTextStyle` isn't creating bold fonts for `subheadline`.

diegoreymendez:
@pinarol - It seems [this recent change](https://github.com/wordpress-mobile/WordPress-iOS/blob/761644daed94e906fc8131cfe2b73da2e3b7df2c/WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift#L244) caused a regression in which the bold font for the highlight in `QuickStartTours.swift` is replaced by a regular font.

Is there a reason why we're forcing a font there?  Can we revert this change?

pinarol:
@diegoreymendez That was done to adapt NoticeView to system font size changes. https://github.com/wordpress-mobile/WordPress-iOS/pull/11614

Couldn't understand what's going on exactly but I guess you can try to put that line back to configureLabels() and see if it changes anything. If it is still responsive to system font size changes then everything should be OK.

diegoreymendez:
Thanks for the context @pinarol, I'll look further into this.

PR #11858, opened 2019-06-05
Make highlightColor on QS white

### LINKED ISSUE #11688: Site Creation: Blue text in notices can't be tapped.

elibud:
Fixes #11688 

To test:

Run Quick Start and check that there's no more blue text.

Update release notes:

cc: @diegoreymendez

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTours.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTours.swift
@@ -349,7 +349,7 @@ private extension String {
     private enum Constants {
         static let iconOffset: CGFloat = 1.0
         static let iconSize: CGFloat = 16.0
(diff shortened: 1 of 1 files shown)


elibud:
@nheagy on this PR I'm just making the text white, but there's still a problem here which is that the font weight is not being taken into account on the NSAttributedString. Can you please look into that and fix it in this PR?

diegoreymendez:
Hey folks, I think this is the same change as: https://github.com/wordpress-mobile/WordPress-iOS/pull/11857

Is that correct?

elibud:
Yes it's the same, closing in favor of Nate's PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/11857/

PR #11879, opened 2019-06-07
[Stats Refresh] Update Y axis chart formatting to match Android's behaviour

### LINKED ISSUE #11859: [Stats Refresh] Charts: improve axis number logic
Reported by @SylvesterWilmott  -

One final thing I think we should align on across platforms is the chart axis number logic, as it stands there is a big difference between the platforms in this regard with Android using slightly more “sensible” numbers than iOS. I believe it’s just the highest number is being rounded to the next “sensible” number that is divisible by the number of horizontal lines on the chart but maybe @vojtasmrcek can explain this better.

### **Period Overview chart examples**
(Note: Period Overview and Post Stats uses the same chart)

**iOS:**
ios

**Android:**
and

### **Latest Post Summary chart examples**

**iOS:**
ios_lps

**Android:**
and_lps

frosty:
Fixes #11859. This PR updates the y axis formatting of the stats charts to better match Android's behaviour. Main changes:

* Updating the logic of the `LargeValueFormatter` to match Android's implementation. On Android this is bundled with the library, but on iOS there's a Swift version with the library's demo app which has different behaviour. We'll now only show a decimal point for values between 1000 and 100,000.
* Rounding the y axis values to produce a nicer set of rounded values on the Y axis. e.g. if the highest value is 96, we would previously have had axis lines at 24, 48, 72, 96. Now they'll be 25, 50, 75, 100. Again, this matches Android.
* Charts with no data now have a default y axis of 0, 1, 2, 3, 4. Previously we were showing 0, 0.2, 0.5, 0.8, 1.0.

chart-y-axis

**To test:**

* Build and run
* Ideally, if you have an Android device to hand, build and run and check the y values match for the various period charts.
* Otherwise, check they look similar to the screenshots above.

cc @SylvesterWilmott 

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Charts/Charts+LargeValueFormatter.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Charts/Charts+LargeValueFormatter.swift
@@ -11,8 +11,8 @@ public class LargeValueFormatter: NSObject, IValueFormatter, IAxisValueFormatter
 
     /// Suffix to be appended after the values.
     ///
-    /// **default**: suffix: ["", "k", "m", "b", "t"]
-    public var suffix = ["", "k", "m", "b", "t"]
+    /// **default**: suffixes: ["", "k", "m", "b", "t"]
+    public var suffixes = ["", "k", "m", "b", "t"]
 
     /// An appendix text to be added at the end of the formatted value.
     public var appendix: String?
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Stats/Charts/Charts+LargeValueFormatter.swift
```diff
@@ -22,33 +22,38 @@ public class LargeValueFormatter: NSObject, IValueFormatter, IAxisValueFormatter
...
 
-        if let appendix = appendix {
-            r += appendix
-        }
+        // Replace the exponent with the correct suffix
+        let replaced = string.replacingOccurrences(of: "E[0-9][0-9]", with: suffixes[exponentInt / 3], options: .regularExpression)
 
-        return r
-    }
+        return replaced
```
ScoutHarris: 👍

frosty:
For reference here's the `LargeValueFormatter` on Android: https://github.com/PhilJay/MPAndroidChart/blob/master/MPChartLib/src/main/java/com/github/mikephil/charting/formatter/LargeValueFormatter.java

PR #11883, opened 2019-06-10
Fix dates math in Stats graph.

### LINKED ISSUE #11873: [Stats Refresh] Period Overview: Views & Visitors are incorrect for previous periods
### Expected behavior
The Views count should be accurate for all periods.

### Actual behavior
The Views count is only accurate for the current period (d, w, m, or y). But for weeks, months, and years, selecting a different period from the date bar or the chart, the Views count is always incorrect. Days is sometimes correct, but not always.

Also, Visitors is sometimes incorrect.

For example:

Calypso:
calpyso

iOS:
ios

Go to Stats > Period. Change the period using either the date bar or the chart. Compare the numbers with Calypso.

### LINKED ISSUE #11882: [Stats Refresh] cannot tap chart bars to change data on first view
Steps to reproduce:

1. Go to the Days/Weeks/Months/Years view for Stats on any site.
1. Tap the bars in the first chart you see and check that the data changes in the modules below.
1. Switch to a different site.
1. Go to the Days/Weeks/Months/Years view for Stats.
1. Check to see if tapping the bars changes the data below.
1. Close and re-open the app.
1. Go to the Days/Weeks/Months/Years view for Stats on any site.
1. Check to see if tapping the bars changes the data below.

<strong>Result:</strong> the first time you open stats after closing the app or switching sites, tapping on the bars in the chart does not change the data in the modules. If you use an arrow to change bars or if you change to a different time period (DWMY), then it starts working normally. ([2m29s](https://cloudup.com/c7s7YmxzQ-S))

IMG_2822
<sup>Tested with WP Internal 12.6.0.20190604 on iPhone 6S iOS 12.3.1</sup>

jklausa:
Fixes #11873, pretty sure also fixes #11882.

I hate calendar math.

To test:

1. Go to stats.
2. Tap on the `<` and `>` buttons and verify that they move you between periods correctly.
3. Tap on random graph elements and verify that it moves you between period corectly.
4. Double check values displayed under the graph with what Calypso is showing.


jklausa:
(This _really_ could use a test harness around it, but it's 5AM round these parts and I'm _not_ looking at dates code for a single minute more today. I'll try to add some on Tuesday in separate PR.)

jklausa:
@ScoutHarris I actually think this is a major enough issue it should probably be `DDLogError`, not `Info`, but good point about not using `NSLog`. Updated :)

PR #11892, opened 2019-06-11
[Stats Refresh] change Post Stats chart delegate

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/11882#issuecomment-500704850

The problem is that sometimes the date bar is created _after_ the overview chart. This resulted in the chart delegate being `nil` because the delegate was the date bar.

This changes the chart delegate to be `PostStatsTableViewController`. Which then inform the date bar when a chart bar has been selected.

To test:
- Access Post Stats via both:
  - Insights > Latest Post Summary > View more.
  - Period > Post and Pages > row selection.
- Verify selecting a chart bar updates the data.

Update release notes:


ScoutHarris:
@jklausa @danielebogo if one of you could review this as soon as you get a chance, I'd appreciate it!

PR #11895, opened 2019-06-12
Don't show blocking alert on set feature image

### LINKED ISSUE #11433: Offline Support: Set Feature Imaged showing blocking alert.
## Description:

Post Settings > Set Featured Image shows an alert that there was a problem accessing your Media Library photos.  Tapping OK dismissed the media picker.

## Matching Android issue:

https://github.com/wordpress-mobile/WordPress-Android/issues/9560

jklausa:
Fixes #11433.

Before, trying to set a Featured image on a post while offline/on a slow connection, would throw up this ugly modal pop-up.

Now, it lets you pick an image and tries to upload it anyway — displaying a notice if it fails.

|before|after|
|---|---|
|Simulator Screen Shot - iPhone Xs Max - 2019-06-12 at 05 40 33|Simulator Screen Shot - iPhone Xs Max - 2019-06-12 at 05 33 20|

To test:
1. Put your device/sim offline.
2. Go to Posts
3. Open editor for any given Post
4. Go to Post Settings
5. Try to set a featured image
6. Verify that the app lets you try to do that, even when offline

### DIFF
--- a/WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
+++ b/WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
@@ -138,11 +138,23 @@
     }
     // try to sync from the server
     MediaCoordinator *mediaCoordinator = [MediaCoordinator shared];
-    [mediaCoordinator syncMediaFor:self.blog success:^{
-        if (!localResultsAvailable && successBlock) {
-            successBlock();
-        }
-    } failure:failureBlock];
+
+    __weak __typeof(self) weakSelf = self;
+    [mediaCoordinator syncMediaFor:self.blog
+                           success:^{
+                               if (!localResultsAvailable && successBlock) {
+                                   successBlock();
+                               }
+                           } failure:^(NSError * _Nonnull error) {
+                               if ([error.domain isEqualToString:NSURLErrorDomain] && weakSelf.ignoreNetworkLoadingErrors && successBlock) {
+                                   successBlock();
+                                   return;
+                               }
+
+                               if (failureBlock) {
+                                   failureBlock(error);
+                               }
+                           }];
 }
 
 - (void)notifyObserversWithIncrementalChanges:(BOOL)incrementalChanges removed:(NSIndexSet *)removed inserted:(NSIndexSet *)inserted changed:(NSIndexSet *)changed moved:(NSArray<id<WPMediaMove>> *)moved
--- a/WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
+++ b/WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
@@ -139,14 +139,14 @@
     // try to sync from the server
     MediaCoordinator *mediaCoordinator = [MediaCoordinator shared];
 
-    __weak __typeof(self) weakSelf = self;
+    __block BOOL ignoreSyncError = self.ignoreSyncErrors;
     [mediaCoordinator syncMediaFor:self.blog
                            success:^{
                                if (!localResultsAvailable && successBlock) {
                                    successBlock();
                                }
                            } failure:^(NSError * _Nonnull error) {
-                               if ([error.domain isEqualToString:NSURLErrorDomain] && weakSelf.ignoreNetworkLoadingErrors && successBlock) {
+                               if (ignoreSyncError && successBlock) {
                                    successBlock();
                                    return;
                                }
--- a/WordPress/Classes/ViewRelated/Media/MediaProgressCoordinatorNoticeViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Media/MediaProgressCoordinatorNoticeViewModel.swift
@@ -5,7 +5,7 @@ enum MediaNoticeUserInfoKey {
 }
 
 struct MediaProgressCoordinatorNoticeViewModel {
-    static let ErrorNoticeTag: Notice.Tag = "MediaProgressCoordinatorNoticeViewModel.tag"
+    static let uploadErrorNoticeTag: Notice.Tag = "MediaProgressCoordinatorNoticeViewModel.UploadError"
 
     private let mediaProgressCoordinator: MediaProgressCoordinator
     private let successfulMedia: [Media]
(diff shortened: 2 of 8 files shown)


WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
```diff
@@ -138,11 +138,23 @@ - (void)loadDataWithOptions:(WPMediaLoadOptions)options success:(WPMediaSuccessB
...
-    } failure:failureBlock];
+
+    __weak __typeof(self) weakSelf = self;
+    [mediaCoordinator syncMediaFor:self.blog
+                           success:^{
+                               if (!localResultsAvailable && successBlock) {
+                                   successBlock();
+                               }
+                           } failure:^(NSError * _Nonnull error) {
+                               if ([error.domain isEqualToString:NSURLErrorDomain] && weakSelf.ignoreNetworkLoadingErrors && successBlock) {
```
yaelirub: two questions:
1. Do we need to check the error domain here? or can we handle the same for any error?
2. Why are we checking the successBlock here?

WordPress/Classes/ViewRelated/Media/MediaProgressCoordinatorNoticeViewModel.swift
```diff
@@ -5,6 +5,8 @@ enum MediaNoticeUserInfoKey {
 }
 
 struct MediaProgressCoordinatorNoticeViewModel {
+    static let ErrorNoticeTag: Notice.Tag = "MediaProgressCoordinatorNoticeViewModel.tag"
```
yaelirub: Can we change this to class+reason to be consistent with the way we defined other tags like:
"WPError.Networking"
or 
ReachabilityUtils.NoConnection"
?

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -1328,14 +1328,19 @@ - (void)showMediaPicker
     options.badgedUTTypes = [NSSet setWithObject: (__bridge NSString *)kUTTypeGIF];
     options.preferredStatusBarStyle = UIStatusBarStyleLightContent;
     WPNavigationMediaPickerViewController *picker = [[WPNavigationMediaPickerViewController alloc] initWithOptions:options];
-    self.mediaDataSource = [[WPAndDeviceMediaLibraryDataSource alloc] initWithPost:self.apost
-                                                             initialDataSourceType:MediaPickerDataSourceTypeMediaLibrary];
-    picker.dataSource = self.mediaDataSource;
+
+    WPAndDeviceMediaLibraryDataSource *mediaDataSource = [[WPAndDeviceMediaLibraryDataSource alloc] initWithPost:self.apost
+                                                                                           initialDataSourceType:MediaPickerDataSourceTypeMediaLibrary];
+    mediaDataSource.mediaLibraryIgnoresNetworkLoadingErrors = YES;
```
yaelirub: Not sure I understand what we're doing here. If we are always setting the mediaLibraryIgnoresNetworkLoadingErrors to true why even use it?

jklausa: `WPAndDeviceMediaLibraryDataSource` is also used in different places and I didn't want to silently change the behavior for other screens in the app

yaelirub: Oh! I see it's used in site creation. What happens there when picking media in offline?

yaelirub:
@jklausa , I think this PR should include an update to release notes as well.

WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
jklausa: I was being cautious here — I think it might be slightly dangerous to outright ignore any errors this method can return — if there's a failure when parsing the response or a Core Data error, I don't think we want to swallow them the same way. 

I'm open to changing this and being more liberal if you think it's fine though :)

yaelirub: Will we still see the annoying popup alert error on any of the other cases (like core data error). I believe we made a general design decision to not block the user with these annoying alerts on errors anymore unless the errors should block progress.
What do you think?

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
jklausa: the lower-level `MediaLibraryPickerDataSource` is also used in bunch of other places, like `GutenbergMediaPickerHelper` and `AztecPostViewController`, `PostEditor`, etc.

I don't want to change the behavior for those.

the `WPAndDeviceMediaLibraryDataSource` that wraps both local images and `MediaPicker` is also used in picking the site icon — it's actually a great point you raised and I changed it to work in both places correctly.

WordPress/Classes/ViewRelated/Media/MediaProgressCoordinatorNoticeViewModel.swift
jklausa: done!

WordPress/Classes/ViewRelated/Media/MediaLibraryPickerDataSource.m
jklausa: Yeah, you're right. I changed it to handle any errors here.

jklausa:
>The copy for the error message seems a bit off to me. I'm not sure we need the message. Maybe we could just have the title of "Couldn't upload feature image" or something like "Error occurred while uploading featured image"?

I'll add a "Needs Copy Review" label, but for what it's worth it's not a new message — I'm reusing what was in the app already.

mikedang:
> I'll add a "Needs Copy Review" label, but for what it's worth it's not a new message — I'm reusing what was in the app already.

The current copy is grammatically off. We could say something like:

>There was a problem with accessing your media. Please try again.

or

>An error occurred while uploading your media. Please try again.

jklausa:
@mikedang sorry for being unclear here. 

I meant that, in the screenshot on the right, indicating "after", that is gonna be displayed after this change, I'm using a copy that was already in use elsewhere in the app — I didn't write it just for this. If you have suggestions as to how improve that — I'm all ears :)

jklausa:
@yaelirub should be ready for another go!

jklausa:
awesome! if there are copy changes suggested by @mikedang, I'll file them as a separate PR :)

mikedang:
Ah, I see. Thanks for clarifying, @jklausa. I think the current message works fine.

PR #11901, opened 2019-06-12
Resume uploads

diegoreymendez:
Merges our UploadManager work into `develop`.

## Testing

Make sure uploads are resumed properly when reopening the app, or bringing it to foreground.

Update release notes:


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 12.7
 -----
 * Block Editor: Video, Quote and More blocks are available now.
+* Offline Improvements: Posts will now be auto-uploaded when reopening the App or when a connection is detected.
```
yaelirub: Should we say, failed to upload posts...? or something like "Post that failed to upload due to connectivity issues will be auto-uploaded..."?

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -210,10 +224,11 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
         PushNotificationsManager.shared.registerForRemoteNotifications()
 
         // Deferred tasks to speed up app launch
-        DispatchQueue.global(qos: .background).async {
+        DispatchQueue.global(qos: .background).async { [weak self] in
             MediaCoordinator.shared.refreshMediaStatus()
             PostCoordinator.shared.refreshPostStatus()
             MediaFileManager.clearUnusedMediaUploadFiles(onCompletion: nil, onError: nil)
+            self?.uploadsManager.resume()
```
yaelirub: Could there be any conflict with the media and posts refresh tasks and the upload manager resume being performed async?

WordPress/Classes/Uploads/UploadsManager.swift
```diff
@@ -0,0 +1,44 @@
...
+
+        super.init()
+
+        setupReachableBlock()
+    }
+
+    // MARK: Reachability
+
+    private func setupReachableBlock() {
+        reachability.reachableBlock = { _ in
```
yaelirub: Two questions here:
1. Shouldn't this be `{ [weak self] in
self?.resume()
}

2. Shouldn't we check if reachability isReachable() and resume if it is?
like  `reachability.reachableBlock = { [weak self] reachability in
guard let reachability = reachability else {
return
}
if reachability.isReachable() {
}`

WordPress/Classes/Uploads/UploadsManager.swift
```diff
@@ -0,0 +1,44 @@
...
+
+        reachability.startNotifier()
+    }
+
+    // MARK: Interacting with Uploads
+
+    /// Resumes all uploads handled by the uploaders.
+    ///
+    func resume() {
+        for uploader in uploaders {
```
yaelirub: this can also be `uploaders.forEach { $0.resume() }`

WordPress/Classes/Uploads/UploadsManager.swift
diegoreymendez: 1. Yep.  Will change.
2. Not really because in theory `reachableBlock` is only called when a connection is detected (as opposed to `unreachableBlock`).

yaelirub:
@diegoreymendez , since we've been working on offline I had multiple draft posts and updates to published posts that failed to upload. I checked your branch and all of those posts were uploaded or updated just now. Even ones that were failed to upload a couple of months a go. 

I like it! I think that's how it should work but I know we're having discussions on whats the expected behavior is and so I just wanted to mention this.

Brilliant work on this!!! I'm super excited about this huge improvement!

diegoreymendez:
Thanks @yaelirub - I'll address all feedback tomorrow morning so we can merge this.

WordPress/Classes/System/WordPressAppDelegate.swift
diegoreymendez: If there is I'm not aware of any.  Both `PostCoordinator` and `MediaCoordinator` have their own operation queues, which makes me think they're prepared to be thread safe.

If we find any issue with this we can tackle it incrementally.

RELEASE-NOTES.txt
diegoreymendez: Changed, and also fixed a typo in another item.

WordPress/Classes/Uploads/UploadsManager.swift
diegoreymendez: Fixed 1.  There should be no need to fix 2 as described above.

diegoreymendez:
@yaelirub - Ready for another look.

WordPress/Classes/Uploads/UploadsManager.swift
diegoreymendez: Changed.

yaelirub:
>"I checked your branch and all of those posts were uploaded or updated just now. Even ones that were failed to upload a couple of months a go.

I like it! I think that's how it should work but I know we're having discussions on whats the expected behavior is and so I just wanted to mention this."

@diegoreymendez, did I miss your response to this?

diegoreymendez:
>> I checked your branch and all of those posts were uploaded or updated just now. Even ones that were failed to upload a couple of months a go.
I like it! I think that's how it should work but I know we're having discussions on whats the expected behavior is and so I just wanted to mention this.

> @diegoreymendez, did I miss your response to this?

Nope I took that as a comment, since this PR is meant to upload failed posts, which is what you're seeing.  Is there a question I'm missing?

diegoreymendez:
After chatting with @yaelirub via DM she clarified to me that there's an ongoing discussion about whether posts should be auto-published or not when back online.  We decided to still move ahead and merge this, and deal with any necessary changes incrementally (mostly because there's no clarity or decisions on this yet).

PR #11923, opened 2019-06-13
[Refresh Stats] Fix crash on rotation

### LINKED ISSUE #11922: [Stats Refresh] iPhone XR crashes on rotation
On the iPhone XR when in any Stats view, rotating causes the app to crash with this:

```
2019-06-13 11:51:05.376423-0600 WordPress[43891:49723930] *** Assertion failure in -[UIPageViewController willAnimateRotationToInterfaceOrientation:duration:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/UIKitCore_Sim/UIKit-3698.103.12/UIPageViewController.m:1089
2019-06-13 11:51:05.409589-0600 WordPress[43891:49723930] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'No view controllers'
```

danielebogo:
Fixes #11922 

This fix the crash when the Stats section is selected and the device rotates.
This bug should affect all the devices where you have a single screen in portrait and a split view in landscape.

## To test:
• Run the branch on and iPhone XR, X, 8 Plus, iPad
• Open Stats from a selected blog
• Rotate the device. It shouldn't crash.
• Change also period and rotate

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/SiteStatsDashboardViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/SiteStatsDashboardViewController.swift
@@ -28,7 +28,9 @@ class SiteStatsDashboardViewController: UIViewController {
     }
 
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
-        updatePeriodView(oldSelectedPeriod: currentSelectedPeriod, reloadData: false)
+        if traitCollection.verticalSizeClass == .regular, traitCollection.horizontalSizeClass == .compact {
+            updatePeriodView(oldSelectedPeriod: currentSelectedPeriod)
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


danielebogo:
@ScoutHarris I fixed the reload data while the device rotates. Would you mind to have another look?

ScoutHarris:
Hey @danielebogo .

OK, I've made one change. The problem I noticed was when changing periods, it would use the selected date from the previous period. It's easy to notice when doing this:
- Go to Years. Select one that is not _2019_.
- Go to Days. The day corresponded to the day from the selected year.

What I've done is only use the selected date in `viewWillTransition` (basically, on these _special_ devices). Otherwise, use the current date.

ScoutHarris:
Hey @jkmassel . Can you include this in the next 12.6 release? Thank you!!

danielebogo:
@ScoutHarris thanks for the fix! It looks good!

PR #11933, opened 2019-06-17
removing displayHasFailedMediaAlert

### LINKED ISSUE #11429: Offline Support: Publishing a Post with Media in Aztec.
## Description:

You can't publish / save a post that has local images while offline.

When trying to save / publish a post with media from the camera roll an alert dialog displays:

> Uploads failed / Some media uploads failed. This action will remove all failed media uploads from the post. Save anyway?

`Not now` and `Yes` are the options.

This was reproduced in Aztec, but not in Gutenberg.

## Matching Android Issue:

None.

yaelirub:
As we are showing the snack bar on a fail upload and giving the option to retry in multiple places we no longer need the popup alert
removing related code as suggested in: https://github.com/wordpress-mobile/WordPress-iOS/issues/11429#issuecomment-501737362

Fixes #11429 

To test:
1. Be offline
2. Add image to an existing Aztec post
3. See retry overlay and snack bar with error message
4. Choose update 
5. See the snack bar with "Error occurred while saving"

IMG_4430

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,8 +6,8 @@
 * Stats Insights: New two-column layout for This Year stats.
 * Stats Insights: added details option for This Year stats.
 * Stats Insights: New two-column layout for Most Popular Time stats.
-
 * Improved messaging experience on failure to upload media.
+
 12.7
 -----
 * Block Editor: Video, Quote and More blocks are available now.
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 * Stats Insights: added details option for This Year stats.
 * Stats Insights: New two-column layout for Most Popular Time stats.
 
+* Improved messaging experience on failure to upload media.
 12.7
 -----
 * Block Editor: Video, Quote and More blocks are available now.
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -39,16 +39,6 @@ extension PostEditor where Self: UIViewController {
             return
         }
 
-        // If there is any failed media allow it to be removed or cancel publishing
-        if hasFailedMedia {
-            displayHasFailedMediaAlert(then: {
-                // Failed media is removed, try again.
-                // Note: Intentionally not tracking another analytics stat here (no appropriate one exists yet)
-                self.publishPost(action: action, dismissWhenDone: dismissWhenDone, analyticsStat: analyticsStat)
-            })
-            return
-        }
-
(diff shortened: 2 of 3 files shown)


yaelirub:
While woking on this PR. 
I have discovered a few other issues.
The overlay seems to disappear on Action sheet dismissal.
Steps:
1. Be offline
2. Add image to post in Aztec
3. see failure overlay
4. tap on overlay
5. choose retry or edit in action sheet
6. see overlay disappears when action sheet is dismissed

overlay_dissapear

Another issue I notice is that there is inconsistency between the options presented in the action sheet in Aztec (Remove image/ Edit/ Dismiss) and Gutenberg (Retry/ Dismiss).

Aztec
IMG_4432

Gutenberg
IMG_4433

I think we should update Aztec to be like Gutenberg as the  edit option is not relevant and the I personally don't like the option to remove the image. 
Would love your thoughts here, @osullivanchris. If you agree, I can create a ticket and update the action sheet.

pinarol:
hey @yaelirub 👋 I did a quick test on this and I think it is acting a bit misleading for the user now. Here are my steps:

- Go offline
- Add a media from device 
- Wait until it fails
- Go online
- Tap Update(or publish)

After these steps I see that the failed message is cleared and the media is there like it is uploaded successfully but actually it is not. If we open the page preview we don't see the media there. If we go back to the posts list page we see a red label saying `Upload failed`. But when we open the post we see the media. So it looked to me as the action sheet was necessary to prevent this but I am not fully informed about the history behind it. 

Publish:

failedMediaAztec

Update:

This was even more strange the thumbnail got lost after the Update then came back once I do a pull to refresh on posts-list page.

failedMediaAztec2


I see on html mode that it is just the local thumbnail we see here:

```
<img src="file:///Users/pinarolguc/Library/Developer/CoreSimulator/Devices/109C7051-9728-458D-BFA6-79E3C5683377/data/Containers/Data/Application/2CC9598C-CE7E-4274-9B02-2536B83C4A22/Library/Caches/Media/thumbnail-p537-1792x1792.jpeg" class="size-full" data-wp_upload_id="x-coredata://ECDA7D29-9874-4736-8387-52235EA981BC/Media/p537">
```

yaelirub:
@pinarol , we have some upcoming changes to post and media upload when regaining connectivity so hopefully that will solve the first issue you mentioned. 
I will look into this more and will modify if necessary after the new media uploader manager is merged

RELEASE-NOTES.txt
```diff
@@ -7,6 +7,7 @@
 * Stats Insights: added details option for This Year stats.
 * Stats Insights: New two-column layout for Most Popular Time stats.
 
+* Improved messaging experience on failure to upload media.
```
diegoreymendez: The blank space above and below this line seems off when compared to other release notes.

yaelirub:
Closing this as @jklausa is working on affecting implementation

PR #11942, opened 2019-06-18
Merge 12.7 into `develop`

jkmassel:
Fixes #

To test:

Update release notes:


WordPress/Resources/release_notes.txt
```diff
@@ -1,13 +1,22 @@
...
-* In Account Settings, your Primary Site now displays the site domain if the site has no name. (Takes the pressure off having to think of The Perfect Name immediately. It’ll come to you!)
-* Wondering about the code the powers the app? We’ve added acknowledgements for all the third-party code libraries we use — you can find them in your settings.
-* We fixed a crash that could happen when setting up your blog, so now you can actually, y’know, set it up.
+* Block Editor: Video, Quote and More blocks are available now.
+* Block Editor: Copy/Paste of text with attributes( bold, italic, ...) will be respected on the editor.
+* Block Editor: Updated color scheme.
+* Block Editor: Nested lists are now available on the toolbar.
+* Post Settings: Setting a Featured Image on a Post/Site should now work better in poor netowrk conditions.
+* Stats Insights: New two-column layout for All-Time stats.
+* Stats Insights: New two-column layout for Today stats.
```
jtreanor: Shouldn't this just be the release notes for 12.7 and not every version?

jkmassel: The script generated this one – I've updated the PR so that it only changes the 12.7 ones, but I could pull out the others if that's what we're after? tbh, I'm not 100% sure what the desired behaviour is here. 

If we decide we only ever want this to be the most recent, I can update the script accordingly!

jtreanor: I believe we only ever want the most recent ones here. The script isn't smart enough to do that yet.

jkmassel: Should be good to go

PR #11946, opened 2019-06-18
Muriel Colors: Add new API and navigation bar changes

nheagy:
Refs #11683 

This sets up the main API for using Muriel colors in the apps. As an intial example, this build changes the colors of the navigation bar to the colors, behind a feature flag.

muriel navbar mk2

To test:
- Read the API, see what you think. It's all strongly typed Swift.
- Turn the flag on/off and see how the navigation bar responds.

Update release notes:
- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`. (**No changes**)


WordPress/Classes/Extensions/UIColor+MurielColors.swift
```diff
@@ -0,0 +1,150 @@
...
+    case shade700 = 700
+    case shade800 = 800
+    case shade900 = 900
+
+    var description: String {
+        return "\(rawValue)"
+    }
+}
+
+struct MurielColorIdentifier {
```
frosty: How about just calling this `MurielColor` to make things a little more concise? Then it's made up of a name and shade.

WordPress/Classes/Extensions/UIColor+MurielColors.swift
```diff
@@ -0,0 +1,150 @@
...
+    class func muriel(color colorID: MurielColorIdentifier) -> UIColor {
+        let assetName = colorID.assetName()
+        guard let color = UIColor(named: assetName) else {
+            return .red
+        }
+        return color
+    }
+
+    /// Muriel accent color
+    static var accent = muriel(color: .accent)
```
frosty: I feel like perhaps these var names are all a bit general to sit directly in UIColor without any kind of name spacing. The easiest solution might just be to prepend `muriel` to them all, although the duplication is a bit annoying. 

Other ideas: 

* Put them all in a `Muriel` struct within `UIColor`, but then you'll have to explicitly reference `UIColor.` at any call sites.
* Add a `color` var to `MurielColorIdentifier` to return the color directly from there. Again you'll have duplication when using it though, as we'd have to do `MurielColorIdentifier(name: .accent).color`.

Just some ideas! I may also just be nitpicking!

WordPress/Classes/Extensions/WPStyleGuide+ApplicationStyles.swift
```diff
@@ -2,15 +2,36 @@ import Foundation
...
 
-    @objc public class func navigationBarShadowImage() -> UIImage {
+    public class func navigationBarShadowImage() -> UIImage {
         return UIImage(color: UIColor(fromHex: 0x007eb1))
     }
+
+    // MARK: - style the navigation appearance using Muriel colors
+    class func configureNavigationAppearance() {
+        let navigationAppearance = UINavigationBar.appearance()
+        navigationAppearance.barTintColor = wordPressBlue()
```
frosty: Should this be using the `.navigationBar` color?

nheagy:
Thanks for the re-review @frosty!

Let's keep our eye on the color naming debate—if we decide what I'd done here is too simplistic, I'll rename them 😀

Issue #11852, opened 2019-06-04
Stats Refresh v2 Master Issue


ScoutHarris:
#### Shared
- [ ] Hide navigation bar on scroll (#10851)
- [ ] Add additional Tracks events (#11775)
- [ ] Update Widget - awaiting input from Sly.

#### Insights
- [ ] Insights Management
  - [ ] Initially, display these stats in this order: _Posting Activity_, _Today_, _All-Time_, _Most Popular Time_, _Comments_
  - [ ] Add an `Edit` button at the bottom of the Insights table to access the `Manage Insights` page.
  - [ ] Manage Insights:
    - [ ] Add to Insights.
    - [ ] Remove from Insights.
    - [ ] Reorder displayed Insights.
  - [ ] Ellipsis button on card headers has options to:
    - [ ] Move up
    - [ ] Move down
    - [ ] Remove from Insights
  - [ ] If all Insights removed, show empty state with `Manage Insights` button.
  - [ ] `News` card: announce IM and scrolls down to the Edit button.
- [ ] Insights _Comments_ details: change to a single non-filtered list (#11713)
  - [ ] Implement support for `/sites/$site/stats/comment-followers` query.
  - [ ] Update details view to show results from `comment-followers`, remove `FilterTabBar`.
- [ ] Two column block type:
  - [ ] _Today_ (renamed from Today's Stats)
  - [ ] _All-time_ (renamed from All Time Stats)
  - [ ] _This Year_ (renamed from Annual Site Stats)
  - [ ] _Most Popular Time_
  - [ ] _Follower Totals_
- [ ] Add `View more` details for _This Year_ (after convert to new block type).

#### DWMY
- [ ] _Countries_: Add map (#11083) (@danielebogo)
- [ ] Remove _Published_? - awaiting input from Sly.

#### Issues
- [ ] Stats page showing error when user doesn't have permissions to view stats on jetpack sites (#11823)
- [ ] Retrieve country flags from the server
- [ ] Retain cycle with StatsTotalRows (https://github.com/wordpress-mobile/WordPress-iOS/pull/10743#pullrequestreview-189348264)
- [ ] Emojis in post titles here get messed up (https://github.com/wordpress-mobile/WordPress-iOS/pull/10831#pullrequestreview-193454300)
- [ ] New Stats section needs Dynamic Type review. (#10845)
[...]

PR #11960, opened 2019-06-20
[Stats Refresh] Countries section - fix map separator height

danielebogo:
Refs. #11852

This PR fixes the Countries map cell separator line. If the map is visible, the bottom separator line should have the same height of the default separator line.

Simulator Screen Shot - iPhone X - 2019-06-20 at 11 00 26

Simulator Screen Shot - iPhone X - 2019-06-20 at 11 00 39

## To test:
• Select a site that contains data to display and select a stats period
• Scroll until the countries section is displayed and check the separator line between the map and the countries list. The separator height should be the same.
• Do the same test selecting a site with no data.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/CountriesCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/CountriesCell.swift
@@ -13,6 +13,8 @@ class CountriesCell: UITableViewCell, NibLoadable {
     @IBOutlet weak var subtitlesStackViewTopConstraint: NSLayoutConstraint!
     @IBOutlet weak var rowsStackViewTopConstraint: NSLayoutConstraint!
 
+    @IBOutlet private var topSeparatorLineHeightConstraint: NSLayoutConstraint!
+
     private weak var siteStatsPeriodDelegate: SiteStatsPeriodDelegate?
     private var dataRows = [StatsTotalRowData]()
     private typealias Style = WPStyleGuide.Stats
@@ -69,6 +71,7 @@ private extension CountriesCell {
         }
 
         rowsStackViewTopConstraint.constant = !dataRows.isEmpty ? subtitleHeight : 0
+        topSeparatorLineHeightConstraint.constant = dataRows.isEmpty ? 0.5 : 0.33
     }
 
 }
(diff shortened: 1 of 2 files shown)


SylvesterWilmott:
Looks good @danielebogo thanks!

WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/CountriesCell.swift
```diff
@@ -13,6 +13,8 @@ class CountriesCell: UITableViewCell, NibLoadable {
     @IBOutlet weak var subtitlesStackViewTopConstraint: NSLayoutConstraint!
     @IBOutlet weak var rowsStackViewTopConstraint: NSLayoutConstraint!
 
+    @IBOutlet private var topSeparatorLineHeightConstraint: NSLayoutConstraint!
```
ScoutHarris: Hey @danielebogo . I know I approved this, but I just noticed this. Just wondering - why is this `private var` instead of `weak var`?

danielebogo: @ScoutHarris `private` because the property doesn't need to be exposed and accessible from outside its class. Also this is not `weak` because the all subviews should be strongly retained. It won't generate any retain cycle. This ensures all the subviews will exist until its parent is deallocated. Having a weak var implicitly unwrapped it's also dangerous because it can be accessed while its status is nil and cause a crash. Apple says _the only time you really need to make an outlet weak is if you have a custom view that references something back up the view hierarchy and in general that's not recommended_.

ScoutHarris: > Apple says the only time you really need to make an outlet weak is if you have a custom view that references something back up the view hierarchy and in general that's not recommended.

Oh, well, it's a good thing Xcode defaults to `weak` when an outlet is created. 🤦‍♂

danielebogo: Ah I think XCode keeps the last _configuration_ you used when an outlet is added. I had the same problem IBAction and IBOutlet

ScoutHarris: TIL.

Issue #11222, opened 2019-03-08
[Sharing] Share Extension: Improve text import from other apps


frosty:
When sharing text to our share extension from other apps (I tested Bear and Ulysses), one of two things happens:

1. The share extension is launched with the shared text inserted as a block quote
2. The share extension is launched with an empty editor

In Bear, to reproduce number 1, simply tap the system share icon in the top right of a note. To reproduce number 2, tap the (i) icon in the top right of a note, tap TXT at the bottom, and attempt to share the text version of the note.

There are two issues here:

### Text inserted as a blockquote

This is a conscious choice we made when we first built the share extension. If only text is being shared to the extension, we insert it as a blockquote. One use case for this is for example sharing snippets of text from a web page (highlight the text in Safari, and choose Share from the popup menu). It seems that we have no way to detect whether a user is sharing selected text or an entire document. 

I think we have a couple of options here, and I'd be interested to get input from @wordpress-mobile/automattic-design:

* Revisit this decision, and perhaps never auto-quote shared text
* Auto-quote text only if it's below a certain character / word count (but how do you decide the limit?)
* Allow the user to choose whether imported text is quoted or not. How would we surface this? A setting in the main app's App Settings?

### No text inserted

I saw the same behaviour from both Ulysses and Bear, and it looks like it's a bug on our end. In this situation, the `NSItemProviders` sent to the share extension contain two types of content: `public.file-url` (a URL for a local file containing the shared content), and (in Bear's case, but not Ulysses') the shared content as `public.plain-text`.

When we attempt to extract shared content (in ShareExtractor.swift), we [iterate through a number of different extractors that we've written](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressShareExtension/ShareExtractor.swift#L126). When we reach an extractor that can extract valid content, we stop checking the other extractors.

In this case, the URL extractor says it can extract the URL content, so we never look at the text. However, the URL extractor [bails out](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressShareExtension/ShareExtractor.swift#L233) later if it detects that the URL is a file URL. So we end up with no content.

There are a couple of solutions here:

* Make the parsing / extracting more robust, and perhaps prefer text content over URL content if it's available.
* Add the ability for the URL extractor to copy content from a file URL

[...]

PR #11965, opened 2019-06-21
Share Extension: Plain Text is no longer Blockquoted

jleandroperez:
### Details:
When importing Plain Text by means of the Share Extension, by default the text gets blockquoted.

We'd like to update this behavior, so that this is no longer the case.

@nheagy May I bug you with a super small PR?
Thanks in advance!!

cc @bummytime YAY for one line fixes!

### Testing:
1. Install both WPiOS and Simplenote
2. Log into both apps
3. Launch Simplenote
4. Open any random note
5. Press over the top right `( i )` button
6. Press over the `Send` button that shows up in the share sheet
7. Pick WordPress iOS from the App List
8. Verify the imported text is no longer blockquoted

### DIFF
--- a/WordPress/WordPressShareExtension/ShareExtractor.swift
+++ b/WordPress/WordPressShareExtension/ShareExtractor.swift
@@ -578,7 +578,7 @@ private struct PlainTextExtractor: TypeBasedExtensionContentExtractor {
             url.absoluteString.count == payload.count {
             returnedItem.url = url
         } else {
-            returnedItem.selectedText = payload
+            returnedItem.importedText = payload
         }
         return returnedItem
     }


jkmassel:
I'm bumping this to `12.9` as part of the `12.8` code freeze. If we need to land it in the `12.8` timeframe, let me know and we can release a new beta release containing it once it lands.

frosty:
@nheagy Didn't we go back and forth with this one with with the recent Share changes? I can't find any recent discussion about it, but I thought it was decided to keep it this way. Could / should we change this just for Simplenote imports?

nheagy:
@jleandroperez I'm going to close this one, as we discussed.

Here's an explanation in case someone comes across this in the future: blockquoting was [debated previously](https://github.com/wordpress-mobile/WordPress-iOS/issues/11222#issuecomment-472424623) and ultimately we decided that text selection to blockquotes served the common use case of sharing articles and the like. Our assumption is external editors will share a local file instead of a text object. We tested this assumption with Bear and Ulysses.

The file formats [WPiOS accepts are in ShareExtractor.swift.](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressShareExtension/ShareExtractor.swift#L350)

Anyway—thanks for understanding. I can't wait to see improved integration between WordPress and Simplenote!

jleandroperez:
Thank you @nheagy!! And thank you @frosty and @bummytime for jumping over! 

I didn't account for the scenarion in which Safari shares selected text, which is quite important IMO (and definitely should be blockquoted). For documentation purposes, as discussed over Slack, we'll attack this problem the other way round (and patch up the way Simplenote shares it's stuff!).

Again, thank you all!! 🔥

PR #11967, opened 2019-06-21
Muriel colors: Tab bar and table view styling methods

nheagy:
Refs #11946 
Relies on https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/209 and https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/104

### Two changes here:
**1. Tab bar**: uses the system colors now, except for the unselected state which uses `Gray 300` because we have to use an image for the reader tab's icon. This is to be able to make the badge appear. For dark mode we'll add to new versions of the image to the asset file.
**2. Table views**: this doesn't make _every_ table view look correct. It moves the table view styling from the `Shared` pod into the app itself. This was coordinated with the Woo team, but I'll be double checking that both apps still build.

muriel color flag comparison tables mk3

To test:
- Check that the tab bar and table view colors look right:
  - table view backgrounds and text colors. Icon colors are mostly unaffected
- With the flag off, ensure the app still look right (aka quite blue)

Update release notes:


nheagy:
@frosty thanks for the review! Ready for another go
- Reorganized tab bar styling
- Turned off tab bar translucency until I can fix reader edges in another PR
- Removed semicolons
- Fixed colors on Me and Notification tabs
- Fixed the Me tab badge size 🎉

nheagy:
@frosty nice find, and thanks for the fix!

This is ready for another peak 😉

nheagy:
Thanks @frosty!

PR #11970, opened 2019-06-21
setting remote post url when its created from abstract post

### LINKED ISSUE #11952: Editor: Updating a post on a self-hosted site makes preview unavailable
### Expected behavior

I expect that I will be able to save a draft post and preview it repeatedly while I am working on it.

### Actual behavior

On a self-hosted site (without Jetpack), I can only preview the post when I first create/save it or when I first open a saved draft. If I update a saved draft, when I try to preview it I always get the "Preview Unavailable" screen.

The app logs show this error:

```
2019-06-19 13:10:52:158 displayPreview: Post permalink is unexpectedly nil
```

1. Select a self-hosted site in the app.
2. Start a new post and enter title/content.
3. Save the draft.
4. Preview the post. The preview loads.
5. Go back to the editor and make a change to the post.
6. Update the draft.
7. Preview the post. Result: The "Preview Unavailable" screen appears.

yaelirub:
Fixes #11952 

When we upload a post on a request to preview we call the uploadPost method in PostService. 
That call creates a remotePost in https://github.com/wordpress-mobile/WordPress-iOS/blob/e26bb0f8ba71e7650a4149b145748aca68e66d59/WordPress/Classes/Services/PostService.m#L720

We then call updatePost with this remotePost. The issue is that we don't assign value to the URL property of the remote post.

This is not an issue with WP sites because the updatePost method in PostServiceRemoteRest returns a remote post object (has a URL value) that we update our local post with. 

When the site is self hosted updatePost in PostServiceRemoteXMLRPC is called and it returns a boolean. We then invoke a callback with the remote post we made the call with (that has no URL).

2 changes in this PR:
1.   Setting the remote post URL with the abstract post permalink upon its creation. 
2.  Pointing to a branch in [PR](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/163) for WordPressKit. With this change, we are fetching the post after it's update like we do in createPost.

To test:
1. Select a self-hosted site in the app.
2. Start a new post and enter title/content.
3. Save the draft.
4. Preview the post. The preview loads.
5. Go back to the editor and make a change to the post.
6. Update the draft.
7. Preview the post.

Before:
post_update_before

After:
post_update

Update release notes:


WordPress/Classes/Services/PostService.m
```diff
@@ -734,6 +734,9 @@ - (RemotePost *)remotePostWithPost:(AbstractPost *)post
     remotePost.authorAvatarURL = post.authorAvatarURL;
     remotePost.excerpt = post.mt_excerpt;
     remotePost.slug = post.wp_slug;
+    if(post.permaLink != nil) {
```
aerych: I think the change in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/163 renders this change moot since we're now fetching the canonical remote post.  Probably fine to keep as is, just making a note for future reference.

yaelirub: @aerych , I left this change since we are returning the captured post in the case of failure to fetch and I wanted it to still have the url value

yaelirub:
Closed with: https://github.com/wordpress-mobile/WordPress-iOS/pull/11985

PR #11973, opened 2019-06-22
[Stats Refresh] Remove unused row types

ScoutHarris:
Fixes #n/a

This a bit of code cleanup/refactor to remove row types that are no longer necessary.

With the conversion to two-column stat cards, `SimpleTotalsStatsRow` was no longer used, so that's gone.
That left only Insights _Publicize_ using `SimpleTotalsCell` (via `SimpleTotalsStatsSubtitlesRow`). With a small change, it's now using `TopTotalsCell`, and thus `SimpleTotalsCell` has been removed.

As an added bonus, this fixes an issue where the _Publicize_ subtitles didn't show until you scrolled up and back down. 🎉 for accidental fixes!

To test:
- Go to Insights Publicize.
- Verify it renders correctly. And with subtitles on the first try!

publicize

Update release notes:


WordPress/Classes/ViewRelated/Stats/SiteStatsTableViewCells.swift
```diff
@@ -206,10 +158,18 @@ struct TopTotalsInsightStatsRow: ImmuTableRow {
             return
         }
 
+        let limitRowsDisplayed: Bool = {
+            if let statSection = dataRows.first?.statSection, statSection == .insightsPublicize {
+                return false
+            }
+            return true
```
danielebogo: What do you think if this can be reduced to:
`return !(dataRows.first?.statSection == .insightsPublicize)`

ScoutHarris: Ah, yes, that is much cleaner. It shall be done. Thanks!

ScoutHarris:
Hey @danielebogo . Ready for another review please.

PR #11974, opened 2019-06-23
Post list revamp improvements

leandroalonso:
Ref #11646

This PR:

* Remove all force unwraps
* Ditches a lot of old Obj-C codes and migrate some of them to Swift
* Moves post styles that were in `ViewController` to `WPStyleGuide+Posts.swift`
* Unify iPad/iPhone post margins to be `16pt` removing some code complexity
* Move `Pages` styles to its own file

To test:

* "Blog posts" section in iPhone and iPads


WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    class func applyActionBarButtonStyle(_ button: UIButton) {
+        button.flipInsetsForRightToLeftLayoutDirection()
+        button.setImage(button.imageView?.image?.imageWithTintColor(grey()), for: .normal)
+        button.setTitleColor(grey(), for: .normal)
+        button.setTitleColor(darkGrey(), for: .highlighted)
+        button.setTitleColor(darkGrey(), for: .selected)
+    }
+
+    class func applySelectedBackgroundView(_ selectedBackgroundView: UIView, topMargin: CGFloat) {
```
nheagy: I know there are a lot of `apply…` method names here, but I feel something like `insertSelectedBackgroundSubview` would be more accurate. Thoughts?

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
```diff
@@ -0,0 +1,140 @@
...
+        let marginMask = UIView()
+        selectedBackgroundView.addSubview(marginMask)
+        marginMask.translatesAutoresizingMaskIntoConstraints = false
+        marginMask.leadingAnchor.constraint(equalTo: selectedBackgroundView.leadingAnchor).isActive = true
```
nheagy: Could we cleanup these into a single `activateConstraints` call?

WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
```diff
@@ -345,17 +345,17 @@ class PageListViewController: AbstractPostListViewController, UIViewControllerRe
...
         }
 
         let sectionInfo = _tableViewHandler.resultsController.sections?[section]
         let nibName = String(describing: PageListSectionHeaderView.self)
-        let headerView = Bundle.main.loadNibNamed(nibName, owner: nil, options: nil)![0] as! PageListSectionHeaderView
+        let headerView = Bundle.main.loadNibNamed(nibName, owner: nil, options: nil)?.first as? PageListSectionHeaderView
 
         if let sectionInfo = sectionInfo {
-            headerView.setTitle(PostSearchHeader.title(forStatus: sectionInfo.name))
+            headerView?.setTitle(PostSearchHeader.title(forStatus: sectionInfo.name))
```
nheagy: Should we unwrap the `headerView` here too?

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
```diff
@@ -0,0 +1,140 @@
+/// A WPStyleGuide extension with styles and methods specific to the Posts feature.
+///
+extension WPStyleGuide {
```
nheagy: I'm glad to see you converted these methods to Swift!

I've made some comments on places I think we could make it a bit more Swift-y without too much effort.

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
```diff
@@ -0,0 +1,140 @@
...
+        let lineHeight = 1.6 * font.pointSize
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.minimumLineHeight = lineHeight
+        paragraphStyle.maximumLineHeight = lineHeight
+        return [.paragraphStyle: paragraphStyle, .font: font]
+    }
+
+    // MARK: - Nav Bar Styles
+
+    class var navigationBarButtonRect: CGRect {
```
nheagy: This var doesn't need a getter now that it's Swift 😀. Same with `spacingBetweeenNavbarButtons`.

Maybe even `postCardBorderColor()`, what do you think?

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    class func buttonForBar(with image: UIImage,
+                            target: Any?, selector: Selector) -> WPButtonForNavigationBar {
+        let button = WPButtonForNavigationBar(frame: navigationBarButtonRect)
+        button.tintColor = .white
+        button.setImage(image, for: .normal)
+        button.addTarget(target, action: selector, for: .touchUpInside)
+        button.removeDefaultLeftSpacing = true
+        button.removeDefaultRightSpacing = true
+        button.rightSpacing = spacingBetweeenNavbarButtons / 2
```
nheagy: I realise this value matches what was in ObjC, but what do you think about rounding the result of `spacingBetweeenNavbarButtons / 2`?

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
```diff
@@ -0,0 +1,140 @@
...
+        let font = notoBoldFontForTextStyle(.headline)
+        let lineHeight = 1.4 * font.pointSize
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.minimumLineHeight = lineHeight
+        paragraphStyle.maximumLineHeight = lineHeight
+        return [.paragraphStyle: paragraphStyle, .font: font]
+    }
+
+    class var postCardSnippetAttributes: [NSAttributedString.Key: Any] {
+        let textStyle: UIFont.TextStyle = UIDevice.isPad() ? .callout : .subheadline
```
nheagy: Do we want to do a device check here? Is the font intended to be the same on iPad in slide over mode as well?

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
leandroalonso: A lot better! Renamed.

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
leandroalonso: Done.

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
leandroalonso: This is a really good question!

What I did here was to translate the previous existent logic. What are your thoughts on this, @SylvesterWilmott?

JFYI, the Reader uses `.subheadline` regardless of the device.

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
leandroalonso: Any special reason to round this value?

I mean, `spacingBetweeenNavbarButtons` is now `40`, which means that this value will always be `20` regardless of rounding it or not.

But if the value changes for, let's say, `41`, we want a rounded number to be used as spacing?

nheagy: I think with `40`, it's fine, just thinking of the weird case if it becomes `39` and `19.5` isn't pixel-aligned on `@3x` devices. Anyway—pretty minor. I wouldn't hold up this PR for this particular comment ✅

WordPress/Classes/ViewRelated/Post/WPStyleGuide+Posts.swift
nheagy: I'd like to heard what Sylvester thinks but I know he's at a meetup this week. I think we can leave this for now and hopefully come back to it in the future.

PR #11977, opened 2019-06-24
Update Gutenberg to 1.8.0

Tug:
This PR updates the gutenberg-mobile reference to point to the latest release, v1.8.0.

The reference is currently pointing to the release/1.8.0 branch on https://github.com/wordpress-mobile/gutenberg-mobile. Relevant PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1172.

To test:
The block editor should work as normal.

Update release notes:


koke:
One issue I found with quotes:

1. Insert a quote block
2. Write some text in the quote
3. Tap in the citation field
4. Write some text
5. Press return
6. Press backspace, nothing happens.

There seems to be 2 lines added when you press return, and backspace won't work while in the middle one

PR #11981, opened 2019-06-24
[Stats Refresh] Some VoiceOver improvements for StatsTotalRow

frosty:
This PR makes a few small improvements to VoiceOver support for StatsTotalRow and its container TopTotalsCell.

* StatsTotalRow should now read the name of the cell followed by the value, and if tappable state that it's a button and to tap for more detail.
* TopTotalsCell should read that it's a table, followed by the names of the columns.

For example:

IMG_4836

* Tapping / selecting this entire table will read "Table showing Link and Clicks".
* Tapping the row shown on the screenshot will read "WordPress.com, 53. Button. Tap for more detail."

**To test:**

* Build and run and enable VoiceOver
* Navigate through a stats period screen and check that the rows are read correctly. 

Update release notes:

There are user facing changes but as we're going to have a large dump of VoiceOver changes, let's just add a single item for them at the end.

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
@@ -178,11 +178,11 @@ class StatsTotalRow: UIView, NibLoadable, Accessible {
     // MARK: - Accessibility
     func prepareForVoiceOver() {
         isAccessibilityElement = true
-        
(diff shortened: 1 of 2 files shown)


frosty:
Thanks for the review! And you're right, I forgot to mention the limitations of this PR – the View More row isn't covered, and neither is 'No data available'. Let's do this incrementally and we can address them in another PR.

PR #11982, opened 2019-06-24
[Stats Refresh] Change stats difference value to grey colour (12.7)

### LINKED ISSUE #11976: [Stats Refresh] Show value difference in grey for current period
On Android, we show the current period value difference in grey because the current period is not complete yet. As a result, this value is far more likely to be coloured red than green which can be demotivating. We should do the same on iOS.

This applies to:

- DWMY tabs
- Post stats

Screenshot 2019-06-24 at 07 40 04

Colour used is `grey`

frosty:
(Same as #11980 but targeting 12.7)

Fixes #11976. This PR updates the color of the difference value in stats to grey to de-emphasize how positive / negative the change is.

Screenshot 2019-06-24 at 14 35 11

**To test:**

* Build and run, check the appearance of the difference label

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,8 @@
 * Stats Insights: New two-column layout for This Year stats.
 * Stats Insights: added details option for This Year stats.
 * Stats Insights: New two-column layout for Most Popular Time stats.
+* Post preview: Fixed issue with preview for self hosted sites not working. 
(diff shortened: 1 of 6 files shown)


frosty:
Got it! Thanks for that, I hadn't noticed the distinction. Will push a fix shortly.

frosty:
@ScoutHarris Round two! I extracted some logic from SiteStatsTableHeaderView, as it was already calculating if we were in the latest period to update the < > arrows.

PR #11996, opened 2019-06-25
Improve UI test output (logs and screenshots)

rachelmcr:
This PR improves some of the UI test output, to help us better understand some flaky test failures we have seen in recent test runs:

* When a UI test fails, we'll take a screenshot after the failure. That way we can see the state of the app when it fails.
* It [groups some test steps](https://developer.apple.com/documentation/xctest/activities_and_attachments/grouping_tests_into_substeps_with_activities) into named activities, to clarify in the logs what step we're at in the test and why certain actions are being taken.

To test:
* Start the mocking server with `rake mocks`.
* Open the simulator.
* Run the UI tests in the `WordPressUITests` target. In the test logs, confirm you see the new named activities as expected.
* Force a test to fail (e.g. remove one of the test steps and re-run that test) and confirm that the test report includes a screenshot showing the state of the app when the test failed (taken during tear down).

Update release notes:

### DIFF
--- a/WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
+++ b/WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
@@ -52,6 +52,14 @@ extension XCTestCase {
         systemAlertHandler(alertTitle: "“WordPress” Would Like to Access Your Photos", alertButton: "OK")
     }
 
+    public func takeScreenshotOfFailedTest() {
+        if let failureCount = testRun?.failureCount, failureCount > 0 {
+            XCTContext.runActivity(named: "Take a screenshot at the end of a failed test") { (activity) in
+                add(XCTAttachment(screenshot: XCUIApplication().windows.firstMatch.screenshot()))
+            }
+        }
+    }
+
     public func systemAlertHandler(alertTitle: String, alertButton: String) {
         addUIInterruptionMonitor(withDescription: alertTitle) { (alert) -> Bool in
             let alertButtonElement = alert.buttons[alertButton]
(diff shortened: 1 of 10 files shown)


WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
```diff
@@ -52,6 +52,14 @@ extension XCTestCase {
         systemAlertHandler(alertTitle: "“WordPress” Would Like to Access Your Photos", alertButton: "OK")
     }
 
+    public func takeScreenshotOfFailedTest() {
+        if let failureCount = testRun?.failureCount, failureCount > 0 {
```
jtreanor: Is this reset after each test? Does it work for multiple failures?

rachelmcr: > Is this reset after each test?

Yes, this is reset after each test (`testRun` is specific to each test execution); it will only count failures in the test that was just run.

> Does it work for multiple failures?

Our tests are set up to stop after the first failure (`continueAfterFailure = false` in the test setup) so the teardown is done right after a failure — there won't be multiple failures in a test. It wouldn't work the same way if we allowed multiple failures; we'd have to take a different approach in that case. I can see how that check could be misleading if we ever change that setting — I could add a note about it only working in tandem with `continueAfterFailure` set to false.

But this reminds me that it won't screenshot any failures that happen _during_ the teardown (only failures that occur before that point). Do you think that's ok for now or should I continue looking for a broader approach?

jtreanor: Ah, that makes sense. I think this is fine as is. If we change `continueAfterFailure` we can reconsider.

Giving it another pass now.

PR #12001, opened 2019-06-25
Use the Tracks Sentry integration

jkmassel:
Integrate Sentry using Tracks

**To Test:**
- Add these lines to a file and run the app in a simulator:

```swift
import AutomatticTracks

CrashLogging.logMessage("This is a test!")
```

Run the app in a way that causes the code above to run, and verify that it shows up in Sentry.


jtreanor:
There seems to be some failing tests here. I re-ran but it still seems to be red. `develop` is fine.

jkmassel:
Well this was a little more complicated than I first expected.

Tracked down the issue here.

We use https://github.com/wordpress-mobile/NSObject-SafeExpectations which adds the `objectWithKeyPath` method to `NSDictionary`. As it turns out, Sentry also adds an `objectWithKeyPath` method. So when it's bundled statically, the method signatures conflict. 

I've filed an issue with Sentry (https://github.com/getsentry/sentry-cocoa/issues/314), because I don't think they're deliberately exporting these symbols for use outside the library. 

In the mean time, I've disabled static compilation for Sentry and Tracks for the moment, but if we don't think that's an acceptable trade-off, we could see where Sentry lands with this and proceed at that time? We have an implementation that works, it's just inconsistent with the others. That bit of tech debt might be worth it for the optimizations provided by static compilation.

WDYT?

jtreanor:
Ah, that is unexpected. This is a perfect example of why categories in libraries [should always be prefixed](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW4) 😄

> I've disabled static compilation for Sentry and Tracks for the moment

I don't love this but I accept this may be the best trade off right now. Why does it need to be disabled for Tracks as well as Sentry? Perhaps I'm missing something.

PR #12004, opened 2019-06-26
Muriel Colors: Update Image Assets

nheagy:
Refs #11683

This PR updates a bunch of images in our `xcassets` file. It does this by:
- merging `AppImages.xcassets` and `Images.xcassets`
- Creating a duplicate asset collection called `MurielImages.xcassets`
- Switching between the two using the `EXCLUDED_SOURCE_FILE_NAMES` build setting
- Also deletes a few unused assets from the new catalogue because why not

This means that the asset catalogue is chosen irrespective of the `.murielColors` feature flag. Debug builds will use the muriel set, all others the old set.

**These changes can be hard to notice because many of our icons are tinted when used.** Here's some example changes:
Screen Shot 2019-06-25 at 11 38 55 PM
Screen Shot 2019-06-25 at 11 40 06 PM

To test:
- Build as debug, checkout the new icon color, notice a few images around the app have new colors, too.
- Either change the `EXCLUDED_SOURCE_FILE_NAMES` setting or build for release or something, ensure that the old assets are used

Update release notes:


nheagy:
Thanks for the reviews @frosty @jkmassel!

> it looks like they're created in code

Yup 😀 

> how long might we need to keep both around?

At most one more sprint.

PR #12006, opened 2019-06-26
[Stats Refresh] Update Countries Map pod and fix autolayout warnings

danielebogo:
Refs. #11852

Because the original project doesn’t appear to be maintained, this PR updates the `FSInteractiveMap ` pod pointing to the [wordpress-mobile fork](https://github.com/wordpress-mobile/FSInteractiveMap).
It also fixes some autolayout warning I saw testing it. 

## To test:
• Open Stats from a selected site
• Select a Period
• Scroll until the map is displayed
• Make sure everything works fine
• Having a site with one country only would be great to test if the map still crashes (it shouldn't 😊)

Update release notes:


danielebogo:
Thanks for reviewing this!
@ScoutHarris it should be fixed now! It's ready for another look!

PR #11964, opened 2019-06-21
[Fix] Fix NSRangeException in Stats running on iOS 11

danielebogo:
Fixes #11949 

This crash happened on iOS 11 only. Calling the `performBatch` while scrolling the tableview caused a crash in Stats periods and insights. `performBatch` should be called only when the row is touched.

## To test:
- Run the branch on iOS 11
- Open Stats Insights and Period and scroll the overview.
- Expand and close the cells.
- Make the same test using a device with iOS 12 to make sure everything is working as expected.

Update release notes:

- [ ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #12010, opened 2019-06-26
[Stats Refresh] Stats details: fix expanding rows

ScoutHarris:
Fixes #n/a

With #11964, the details table no longer updated as it was calling a delegate method that changed. The result was rows didn't expand and child rows were not shown.

broken

This restores the table update by calling the new delegate method.

To note, this also restores all chevrons updating when a row is expanded, but hey at least they expand.

To test:
- Go to any details view that has expandable rows. 
- Expand a row.
- Verify child rows appear.

fixed

Update release notes:


ScoutHarris:
@jkmassel  - this will need to go in the next 12.7 release please.

PR #12023, opened 2019-06-28
Added Create Post button to empty results view displayed on no connectivity from post list

### LINKED ISSUE #11997: "Create Post" button should be shown when offline too.
This issue was spawned from a proposal by @shiki in: https://github.com/wordpress-mobile/WordPress-iOS/issues/11430#issuecomment-481835803

The proposal is that the "Create a Post" button should also be shown when there is no data AND we're offline.

## Decision

Original proposal accepted: **If there is no cached data and the device is offline, we should still show the Create a Post button.**
Headline:
Unable to load posts right now.

Subheading:
Check your network connection and try again. Or draft a post.

Button:
Create post

No punctuation on buttons, naturally, though I feel like our sentences here need terminal punctuation. Also, button text is sentence case per the brand guide.

## Original Proposal

I think this needs some discussion on how empty states should be handled. From what I can see, when loading either _Published_ or _Drafts_:

| Has Cached Data? | Online | Offline |
|--------|-------|-------|
|  NO  |   |   |
| YES   |   |  |

My opinion is that if there is no cached data and the device is offline, we should still show the _Create a Post_ button.

yaelirub:
When displaying  NoResultsView due to no connection from post lists
we want to display a specific copy and a create post button.

Fixes #11997 

Before:
 

After:
Screen Shot 2019-06-27 at 5 36 03 PM

To test:
1. Be offline
2. Go to Published or Drafts tab 
3. See the "no results view" with the right copy
4. See the "Create Post" button
5. Tap create post button
6. See that when exiting the editor or from the menu you have save as draft option
7. Tap save as draft
8. See the post was created as draft but fails to upload.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
@@ -147,17 +147,27 @@ import Reachability
     ///   - accessoryView:      View to show instead of the image. Optional.
     ///
     @objc func configure(title: String,
+                         noConnectionTitle: String? = nil,
                          buttonTitle: String? = nil,
                          subtitle: String? = nil,
+                         noConnectionSubtitle: String? = nil,
                          attributedSubtitle: NSAttributedString? = nil,
                          attributedSubtitleConfiguration: AttributedSubtitleConfiguration? = nil,
                          image: String? = nil,
                          subtitleImage: String? = nil,
                          accessoryView: UIView? = nil) {
         let isReachable = reachability?.isReachable()
-        titleText = isReachable == false ? NoConnection.title : title
-        subtitleText = isReachable == false ? NoConnection.subTitle : subtitle
-        attributedSubtitleText = isReachable == false ? NSAttributedString(string: NoConnection.subTitle) : attributedSubtitle
+        if (isReachable == false) {
+            titleText = noConnectionTitle != nil ? noConnectionTitle : NoConnection.title
+            let subtitle = noConnectionSubtitle != nil ? noConnectionSubtitle : NoConnection.subTitle
+            subtitleText = subtitle
+            attributedSubtitleText = NSAttributedString(string: subtitleText!)
+        } else {
+            titleText = title
+            subtitleText = subtitle
+            attributedSubtitleText = attributedSubtitle
+        }
+
         configureAttributedSubtitle = attributedSubtitleConfiguration
         buttonText = buttonTitle
         imageName = isReachable == false ? NoConnection.imageName : image
@@ -174,14 +184,6 @@ import Reachability
     func configureForNoSearchResults(title: String) {
         configure(title: title)
         displayTitleViewOnly = true
-    }
-
-    func configureForNoConnectionInPostList() {
-        titleText = NoConnection.titlePostList
-        subtitleText = NoConnection.subTitlePostList
-        attributedSubtitleText = NSAttributedString(string: NoConnection.subTitlePostList)
-        buttonText = NoConnection.buttonTitlePostList
-        imageName = NoConnection.imageName
     }
 
     /// Public method to remove No Results View from parent view.
@@ -550,9 +552,6 @@ private extension NoResultsViewController {
     struct NoConnection {
         static let title: String = NSLocalizedString("Unable to load this page right now.", comment: "Title for No results full page screen displayed when there is no connection")
         static let subTitle: String = NSLocalizedString("Check your network connection and try again.", comment: "Subtitle for No results full page screen displayed when there is no connection")
-        static let titlePostList: String = NSLocalizedString("Unable to load posts right now.", comment: "Title for No results full page screen displayedfrom post list when there is no connection")
-        static let subTitlePostList: String = NSLocalizedString("Check your network connection and try again. Or draft a post.", comment: "Subtitle for No results full page screen displayed from post list when there is no connection")
-           static let buttonTitlePostList: String = NSLocalizedString("Create Post", comment: "Button title, encourages users to create a post.")
         static let imageName = "cloud"
     }
 }
--- a/WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
@@ -175,7 +175,7 @@ import Reachability
         configure(title: title)
         displayTitleViewOnly = true
     }
-    
+
     func configureForNoConnectionInPostsList() {
         titleText = NoConnection.titlePostList
         subtitleText = NoConnection.subTitlePostList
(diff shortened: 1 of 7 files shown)


diegoreymendez:
@yaelirub - Remember to tick the "Update release notes" checkbox.

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
```diff
@@ -542,6 +550,9 @@ private extension NoResultsViewController {
     struct NoConnection {
         static let title: String = NSLocalizedString("Unable to load this page right now.", comment: "Title for No results full page screen displayed when there is no connection")
         static let subTitle: String = NSLocalizedString("Check your network connection and try again.", comment: "Subtitle for No results full page screen displayed when there is no connection")
+        static let titlePostList: String = NSLocalizedString("Unable to load posts right now.", comment: "Title for No results full page screen displayedfrom post list when there is no connection")
+        static let subTitlePostList: String = NSLocalizedString("Check your network connection and try again. Or draft a post.", comment: "Subtitle for No results full page screen displayed from post list when there is no connection")
+           static let buttonTitlePostList: String = NSLocalizedString("Create Post", comment: "Button title, encourages users to create a post.")
```
jklausa: The whitespace on this line is off.

diegoreymendez: The indentation seems to be off here.

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
```diff
@@ -176,6 +176,14 @@ import Reachability
         displayTitleViewOnly = true
     }
 
+    func configureForNoConnectionInPostList() {
```
diegoreymendez: Having this kind of customization that's specific to one of the App's VCs in `NoResultsViewController.swift` isn't great, because it means we have some of the logic for the post list VC in one place, and part of in another place.

In my opinion this class should be mostly abstract and not have logic that's specific to any parent VC.

Considering there's already a `configure` method that allows us to configure all this, can we call that the post list customizations all live in `PostListViewController.swift`?

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
```diff
@@ -542,6 +550,9 @@ private extension NoResultsViewController {
     struct NoConnection {
         static let title: String = NSLocalizedString("Unable to load this page right now.", comment: "Title for No results full page screen displayed when there is no connection")
         static let subTitle: String = NSLocalizedString("Check your network connection and try again.", comment: "Subtitle for No results full page screen displayed when there is no connection")
+        static let titlePostList: String = NSLocalizedString("Unable to load posts right now.", comment: "Title for No results full page screen displayedfrom post list when there is no connection")
+        static let subTitlePostList: String = NSLocalizedString("Check your network connection and try again. Or draft a post.", comment: "Subtitle for No results full page screen displayed from post list when there is no connection")
```
diegoreymendez: This message sounds a bit strange due to the "Or draft a post." bit, can I ask you @osullivanchris to share your opinion on it?

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
diegoreymendez: Actually I just understood why you're not using the existing configure method.  I'll ping you over DM to suggest an alternative.

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
yaelirub: This _was_ @osullivanchris suggestion. https://github.com/wordpress-mobile/WordPress-iOS/issues/11997#issuecomment-506047116

yaelirub:
@diegoreymendez , @jklausa updated according to code review.
Now not specifying for post list view controller so had to update caller methods in obj-c

WordPress/Classes/ViewRelated/Blog/NoResultsViewController.swift
osullivanchris: The message was actually written by editorial. If you want me to raise it with them I can? I think its ok! Its slightly awkward, but might just be the nature of what we're trying to say.

diegoreymendez: That's fine - the reason I was asking was to double check as the message looks a bit awkard indeed!  I'm marking this as resolved.

diegoreymendez:
I [opened a separate issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/12049) for my observation that this needs to be done for the pages list too.

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -597,8 +597,7 @@ private extension PostListViewController {
     func handleRefreshNoResultsViewController(_ noResultsViewController: NoResultsViewController) {
 
         guard connectionAvailable() else {
-            noResultsViewController.configure(title: noConnectionMessage(),
-                                              image: noResultsImageName)
+            noResultsViewController.configure(title: "", noConnectionTitle: NoResultsText.noConnectionTitle, buttonTitle: NoResultsText.noConnectionButtonTitle, subtitle: nil, noConnectionSubtitle: NoResultsText.noConnectionSubtitle, attributedSubtitle: nil, attributedSubtitleConfiguration: nil, image: nil, subtitleImage: nil, accessoryView: nil)
```
diegoreymendez: Considering that `NoResultsViewController` is prepared to handle the offline state, why not avoid this guard here entirely and send the offline customizations to the calls below?

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -661,5 +660,8 @@ private extension PostListViewController {
         static let noScheduledTitle = NSLocalizedString("You don't have any scheduled posts", comment: "Displayed when the user views scheduled posts in the posts list and there are no posts")
         static let noTrashedTitle = NSLocalizedString("You don't have any trashed posts", comment: "Displayed when the user views trashed in the posts list and there are no posts")
         static let noPublishedTitle = NSLocalizedString("You haven't published any posts yet", comment: "Displayed when the user views published posts in the posts list and there are no posts")
+        static let noConnectionTitle: String = NSLocalizedString("Unable to load posts right now.", comment: "Title for No results full page screen displayedfrom post list when there is no connection")
+        static let noConnectionSubtitle: String = NSLocalizedString("Check your network connection and try again. Or draft a post.", comment: "Subtitle for No results full page screen displayed from post list when there is no connection")
+        static let noConnectionButtonTitle: String = NSLocalizedString("Create Post", comment: "Button title, encourages users to create a post.")
```
shiki: Should we reuse and change the message in `buttonTitle` instead?

yaelirub: Yeah we can

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
diegoreymendez: As discussed let's skip this for now as understanding the existing code ain't simple.

designsimply:
Tested and confirmed with 12.9 beta (WP Internal 12.9.0.20190717) that I can create a draft while offline, that it fails to upload while offline, and that it does upload after going back online. ([53s](https://cloudup.com/cV5u7CU46Fv))

PR #12005, opened 2019-06-26
Remove extra editor_session_end tracks event

etoledom:
This PR removes an extra track of the event `editor_session_end` on save/update posts.

As showed [in this comment](https://github.com/wordpress-mobile/WordPress-Android/issues/10098#issuecomment-505755839), this event was being double tracked on save/update events.

This event is already properly tracked for all interactions here:
https://github.com/wordpress-mobile/WordPress-iOS/blob/c504303b0962001b8f77fbb3cf071e633e908b26/WordPress/Classes/ViewRelated/Post/PostEditor%2BPublish.swift#L92

To test:

Even though this should only affect save/update actions, to be sure let's check that in all possible editor closing actions, `editor_session_end` is tracked only once.

**Tip:** To clear the console: 

- Activate the `OS_ACTIVITY_MODE: disable` environment variable in the Run scheme editor.
- On `GutenbergViewController:212 (init)` add `RCTSetLogThreshold(RCTLogLevel.warning)`
- On `GutenbergViewController:gutenbergDidEmitLog`(around L:512), comment `.trace` and `.info` logs (I think this one helps when running metro server).

A PR will come soon to help with these verbose logs from RN.

- Discard new post
    - New post
    - Discard post
- Discard changes on Draft post
    - Open a saved draft
    - Make some changes
    - Close via X button saving changes
- Close published post
    - Open a published post
    - Close it without any changes
- Discard changes on Published post
    - Open a published post
    - Make some changes
    - Close the post discarding changes
- Publish new post
    - Create a new post
    - Add content
    - Publish the post via Publish button
- Publish draft post
    - Open a draft
    - Press the menu (…) button
    - Publish now
    - Publish now
- Save new post as draft
  - Create new post
  - Add some content
  - Press X button
  - Save as Draft
- Update existing draft post
  - Open existing draft
  - Edit content
  - Press X button
  - Update draft
- Update published post
  - Open a published post
  - Edit content
  - Press X button
  - Update post

Repeat for pages.

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

cc @daniloercoli

PR #12026, opened 2019-06-28
Editor Analytics ContentType now is set to new when it should

etoledom:
This PR fixes the issue spotted here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12005#issuecomment-506250177

`hasContent()` method checks for both Title and Body of the post, so it's better suited to check if a post is new.
~With an empty body, `containsGutenbergBlock()` was failing (there are no blocks - yet), but the editor information is already given by the `Editor` enum instance, being a more reliable source for this information.~
**Edit:** As discussed in the comment, the current behavior is the desired behavior

To test:

- Create a new post.
- Check that `editor_session_start` is tracked with `content_type: new`.
- Edit the title and save as draft.
- Open the saved post using **gutenberg**:
  - Check that `editor_session_start` is tracked with `content_type: classic`
- Open the saved post using **Aztec**:
  - Check that `editor_session_start` is tracked with `content_type: classic`

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
@@ -14,7 +14,7 @@ struct PostEditorAnalyticsSession {
         currentEditor = editor
         postType = post.analyticsPostType ?? "unsupported"
         blogType = post.blog.analyticsType.rawValue
-        contentType = ContentType(post: post).rawValue
+        contentType = ContentType(for: post, and: editor).rawValue
     }
 
     mutating func start(hasUnsupportedBlocks: Bool) {
@@ -87,10 +87,10 @@ extension PostEditorAnalyticsSession {
         case gutenberg
         case classic
 
-        init(post: AbstractPost) {
-            if post.isContentEmpty() {
+        init(for post: AbstractPost, and editor: Editor) {
+            if !post.hasContent() {
                 self = .new
-            } else if post.containsGutenbergBlocks() {
+            } else if editor == .gutenberg {
                 self = .gutenberg
             } else {
                 self = .classic
--- a/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
@@ -14,7 +14,7 @@ struct PostEditorAnalyticsSession {
         currentEditor = editor
         postType = post.analyticsPostType ?? "unsupported"
         blogType = post.blog.analyticsType.rawValue
-        contentType = ContentType(for: post, and: editor).rawValue
+        contentType = ContentType(post: post).rawValue
     }
 
     mutating func start(hasUnsupportedBlocks: Bool) {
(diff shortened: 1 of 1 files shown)


koke:
> With an empty body, `containsGutenbergBlock()` was failing (there are no blocks - yet), but the editor information is already given by the `Editor` enum instance, being a more reliable source for this information.

`containsGutenbergBlock` should return `false` if the content is empty. The `content_type` property was intentionally about the content, not the editor used. Right now, there's a 1:1 match between the content_type and the editor that opens for that content, but that might change in the future. For instance, if we start opening classic posts with Gutenberg.

If we don't think that distinction is valuable, then `content_type` doesn't make much sense anymore beyond `new|existing`, as the editor opened is already provided by the `editor` property.

etoledom:
> The content_type property was intentionally about the content, not the editor used

Right! I misunderstood it. it looked weird having editor: gutenberg, and content_type: classic, but it makes sense.

I'll roll back that change. Thanks for noticing it!

etoledom:
Thanks for the review @koke !
Updated the PR description and it's ready for another look.

koke:
Oh I see it now, I opened a post with just a title:

```
editor_session_start, properties: {
    "blog_type" = jetpack;
    "content_type" = classic;
    editor = gutenberg;
    "has_unsupported_blocks" = 0;
    "post_type" = post;
    "session_id" = "3F8C7322-5306-4902-87FB-82A6728E95C6";
}
```

This is misleading as well, since it says `editor=gutenberg` and `content_type=classic`. Sorry for changing my mind again, but what do you think about introducing a new `content_type=empty`?

- Using `new` is misleading because it's an existing post.
- Using `gutenberg` is not accurate because the content doesn't have blocks.
- Using `classic` might be ok depending on the interpretation, but it's surprising to see a "classic" post opened with Gutenberg at this stage.

etoledom:
> what do you think about introducing a new content_type=empty

I agree! and I like the idea. Probably it's is a wording problem.

Alternatively we could rename `classic` and `gutenberg` (that makes think about the the editor) to something like `html` or `plane_html`, in contrast of `gutenberg_blocks` content.

I'm not sure how changing a prop name might affect the previous events. I'm OK with both alternatives.

Also cc @daniloercoli to coordinate this change with Android.

koke:
I think it might be fine as it is? 🤔 

Reading your suggestions, I remember the first attempt at describing the content was more like just adding `has_blocks`, but we also wanted to differentiate when the user was editing a classic post from a new one.

I'm not sure the distinction between a completely new post and an existing one with just a title is that important, as the editor to use would still match the default one: `EditorFactory` uses `post.isContentEmpty` to determine "new" posts in terms of using the app setting.

So if I did this again, I might rename `new` to `empty` to be more accurate, but I don't think renaming the property is worth the trouble at this stage. I'd say let's leave it as it is.

etoledom:
> I'd say let's leave it as it is.

So, if I understand correctly, we should not merge this PR?

> I'm not sure the distinction between a completely new post and an existing one with just a title is that important, as the editor to use would still match the default one

That's a good point.

If `content_type` new doesn't bump the new posts event it's probably fine. Still the names are a bit confusing (others could also think about classic and gutenberg referring to the editor), but you have the best context and experience with the editor analytics to see that it's not a big issue 👍 

Let me know to close the PR!

koke:
> So, if I understand correctly, we should not merge this PR?

Yes, let's close it and keep the current behavior, even if the naming is a bit confusing.

etoledom:
Perfect! Thank you @koke for keeping an eye on this 🙏

PR #12027, opened 2019-06-28
Add Alternative App Icons

frosty:
This PR adds some alternative app icons using iOS 10.3's `alternate icon` feature. This is just a fun feature to allow a bit of customisation for users. Here's the initial batch:

Screenshot 2019-06-28 at 11 33 51

icon-change-1

The app's icon can be changed by going to **Me** > **App Settings** > **App Icon** (down the bottom), and choosing an icon. Some notes on implementation:

* Alternate icons have to be listed in the Info.plist. Unfortunately, they can't be included in an asset catalog, so they have to be added to the project the old fashioned way as standard files.
* I added code in `AppIconViewController.swift` to dynamically fetch the list of icon names from the Info.plist instead of hardcoding them directly.
* I added some tracking as it'd be nice to see how much use this feature gets, but I didn't want to link specific users to specific icons so I'm just sending generic 'icon changed' and 'icon reset' events (associated PR here: https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/211).
* The Jetpack icon is only visible in the picker if you have a Jetpack site added to the app.

**To test:**

* Build and run
* Navigate to Me > App Settings > App Icon, and try changing icons
* Ensure you test switching back to the original icon as that's a slightly different case
* Check the Jetpack icon appears if you have a Jetpack site, and doesn't appear for a normal self-hosted site (you can create one with https://poopy.life if you don't have one).

Update release notes:


WordPress/Classes/ViewRelated/Me/AppIconViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        guard let iconDict = infoPlistIconsDict else {
+            self.icons = icons
+            return
+        }
+
+        // Add them (sorted) to the default key – first any prefixed with WordPress, then the rest.
+        let keys = iconDict.keys
+        let wordPressKeys = keys.filter({$0.hasPrefix("WordPress")}).sorted()
+        let otherKeys = keys.filter({!$0.hasPrefix("WordPress")}).sorted()
+        icons.append(contentsOf: (wordPressKeys + otherKeys))
```
danielebogo: Not a big deal but if you treat `keys` as a _Set_, you can get the `otherKeys` subtracting `wordPressKeys`.
```
let keys = Set(iconDict.keys)
let wordPressKeys = keys.filter { $0.hasPrefix(_CONSTANT_PREFIX_) }.sorted()
let otherKeys = keys.subtracting(wordPressKeys).sorted()
```

WordPress/Classes/ViewRelated/Me/AppIconViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+
+        // Load the names of the alternative app icons from the info plist
+        guard let iconDict = infoPlistIconsDict else {
+            self.icons = icons
+            return
+        }
+
+        // Add them (sorted) to the default key – first any prefixed with WordPress, then the rest.
+        let keys = iconDict.keys
+        let wordPressKeys = keys.filter({$0.hasPrefix("WordPress")}).sorted()
```
danielebogo: Can you use `Constants. defaultIconName` to check the prefix?

WordPress/Classes/ViewRelated/Me/AppIconViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        let otherKeys = keys.filter({!$0.hasPrefix("WordPress")}).sorted()
+        icons.append(contentsOf: (wordPressKeys + otherKeys))
+
+        // Only show the Jetpack icon if the user has a Jetpack-connected site in the app.
+        if BlogService(managedObjectContext: ContextManager.shared.mainContext).hasAnyJetpackBlogs() == false {
```
danielebogo: Quick question: should we allow WP.com users to use the Jetpack icon too?

WordPress/Classes/ViewRelated/Me/AppIconViewController.swift
frosty: Ah, nice!

WordPress/Classes/ViewRelated/Me/AppIconViewController.swift
frosty: You could, but it'd take a bit more dictionary diving to get to the list of icon files for each icon.

WordPress/Classes/ViewRelated/Me/AppIconViewController.swift
frosty: I guess we could do!

PR #12031, opened 2019-06-29
remove blocking spinner on upload and autosave before preview 

### LINKED ISSUE #11812: "Generating Preview..." HUD is blocking user interaction. Use progress bar instead.
This is spun off of #11722 
detailed comment here:  https://github.com/wordpress-mobile/WordPress-iOS/pull/11722#pullrequestreview-243054234

The loading HUD blocks any interaction, so I can't get away from the preview screen until the request times out.
> 
> Screen Shot 2019-05-29 at 08 09 41
> Screen Shot 2019-05-29 at 08 08 46

Steps to reproduce:
1. Edit published post
2. Turn to offline mode on simulator or Network Link Conditioner enabled and 100% packet loss on device.
3. Tap to preview

## Decision
We are going to implement something similar to uploading a media with a progress bar 
IMG_1287
It shows a loading state, with a description, and doesn't block the user from using the UI

Generating preview loader

yaelirub:
Fixes #11812 

As described in the issue when saving a draft or autosaving a post before preview we used to show a blocking spinner which would block the user from performing any actions till preview is generated. 
This issue is prominent in low connectivity.

This PR aims to improve the experience by showing the saving/generating status in the navigation bar the same way we do when uploading media. 

generating

**This improvement brings up the need to also remove the blocking spinner when loading the preview (web view loading). It will be tackled in a separate [PR](https://github.com/wordpress-mobile/WordPress-iOS/issues/12032)**

To test:
1. Go to device settings -> Developer and turn on Network Link Conditioner with "Very Bad Network"
POSTS:
2. Open the app and go to published posts
3. Update a post
4. tap preview
5. See "Generating Preview" in the nav bar.
6. Make sure you can exist the editor and perform other actions. 
7. Go back from the preview and see that the navigation bar in the editor is back to default.

DRAFTS:
2. Open the app and go to drafts
3. Update a draft
4. tap preview
5. See "Saving Draft" in the nav bar.
6. Make sure you can exist the editor and perform other actions. 
7. Go back from the preview and see that the navigation bar in the editor is back to default.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -3190,7 +3190,7 @@ extension AztecPostViewController {
         static let defaultMargin            = CGFloat(20)
         static let blogPickerCompactSize    = CGSize(width: 125, height: 30)
         static let blogPickerRegularSize    = CGSize(width: 300, height: 30)
-        static let savingDraftButtonSize      = CGSize(width: 130, height: 30)
+        static let savingDraftButtonSize    = CGSize(width: 130, height: 30)
         static let uploadingButtonSize      = CGSize(width: 150, height: 30)
         static let moreAttachmentText       = "more"
         static let placeholderPadding       = UIEdgeInsets(top: 8, left: 5, bottom: 0, right: 0)
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -3192,7 +3192,6 @@ extension AztecPostViewController {
         static let blogPickerRegularSize    = CGSize(width: 300, height: 30)
         static let savingDraftButtonSize      = CGSize(width: 130, height: 30)
         static let uploadingButtonSize      = CGSize(width: 150, height: 30)
-        static let generatingPreviewButtonSize = CGSize(width: 170, height: 30)
         static let moreAttachmentText       = "more"
         static let placeholderPadding       = UIEdgeInsets(top: 8, left: 5, bottom: 0, right: 0)
         static let headers                  = [Header.HeaderType.none, .h1, .h2, .h3, .h4, .h5, .h6]
@@ -3298,10 +3297,6 @@ extension AztecPostViewController: PostEditorNavigationBarManagerDelegate {
 
     var uploadingButtonSize: CGSize {
         return Constants.uploadingButtonSize
-    }
-
-    var generatingPreviewButtonSize: CGSize {
-        return Constants.generatingPreviewButtonSize
     }
 
     var savingDraftButtonSize: CGSize {
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -3299,11 +3299,11 @@ extension AztecPostViewController: PostEditorNavigationBarManagerDelegate {
     var uploadingButtonSize: CGSize {
         return Constants.uploadingButtonSize
     }
-    
+
     var generatingPreviewButtonSize: CGSize {
         return Constants.generatingPreviewButtonSize
     }
-    
+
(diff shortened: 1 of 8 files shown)


yaelirub:
>It's kinda hard to notice the fact something has changed in the navbar - I'm not sure if I didn't know where to look, I'd notice anything at all. Users will probably get used to it though.

>A bigger problem however, is that the amount of text we can fit in that amount of space is extremely limited - I'm worried about localization issues here. With double-length pseudolanguage enabled, it already doesn't fit:

Hey @jklausa , thank you so much for these comments. @osullivanchris do you have any input on these?

>(and causes UI glitch with the spinner)

I will update the code to fix this. Wish we were using autolayout there.

>I imagine this will get solved in the future PR you mentioned for async preview loading - but I thought I'd mention it just in case - maybe something to keep in mind for the next PR :)

It will be solved in a future PR. Thanks for mentioning though. 🙏

osullivanchris:
> Hey @jklausa , thank you so much for these comments. @osullivanchris do you have any input on these?

Hey I think these concerns from @jklausa are totally valid! I thought it was effective to grab something for free here to unblock us. But if I was designing this pattern from scratch that's not the way I would do it. I actually don't know if I've seen that in any app before :D 

It depends whether we think there is value in creating a new pattern for non-blocking loading indication. I'd be happy to take a look at it, but perhaps it should be a separate issue?

jklausa:
Yeah —  I think it's fine to ship this as-is, since it's still an improvement over current state — and we can revisisit and revamp it down the line if we have time :)

yaelirub:
@jklausa , I have created a new view LoadingStatusView that is meant to replace the WPUplaodStatusButton. 
It is using autolayout instead of frames so layout should be less problematic. 

I tested this in English and Hebrew.

Please test like you did before and let me know what you think. 

If it seems good to you I can open another PR to replace  WPUplaodStatusButton with LoadingStatusView

WordPress/Classes/ViewRelated/Post/PostEditorNavigationBarManager.swift
```diff
@@ -73,6 +75,24 @@ class PostEditorNavigationBarManager {
         return button
     }()
 
+    /// Preview Generating Button
+    ///
+    private lazy var previewGeneratingView: LoadingStatusView = {
+        let view = LoadingStatusView(title: NSLocalizedString("Generating Preview", comment: "Message to indicate progress of generating preview"))
+        view.translatesAutoresizingMaskIntoConstraints = false
```
shiki: ⚠️  This looks unnecessary since `LoadingStatusView` already does this in its initializer.

WordPress/Classes/ViewRelated/Post/PostEditor+MoreOptions.swift
```diff
@@ -72,27 +64,29 @@ extension PostEditor where Self: UIViewController {
 
     func displayPreview() {
         savePostBeforePreview() { [weak self] previewURLString, error in
-            if error != nil {
-                let title = NSLocalizedString("Preview Unavailable", comment: "Title on display preview error" )
-                self?.displayPreviewNotAvailable(title: title)
+            guard let self = self else {
```
shiki: Nice improvement!

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -3190,6 +3190,7 @@ extension AztecPostViewController {
         static let defaultMargin            = CGFloat(20)
         static let blogPickerCompactSize    = CGSize(width: 125, height: 30)
         static let blogPickerRegularSize    = CGSize(width: 300, height: 30)
+        static let savingDraftButtonSize      = CGSize(width: 130, height: 30)
```
shiki: Looks like this is misaligned. 🙂

yaelirub:
Following [PR](https://github.com/wordpress-mobile/WordPress-iOS/issues/12101) to replace WPUplaodStatusButton with LoadingStatusView

designsimply:
I tested with 21.9 beta and can confirm that when I tap Preview I can see a saving/generating status in the navigation bar.

preview generating-preview
<sup>Tested with WP Internal 12.9.0.20190717 on iPhone 6S iOS 12.3.1 on a WordPress 5.2.2 Jetpack 7.5.3 site.</sup>

PR #12041, opened 2019-07-01
[Stats Refresh] Tapping Chart bars doesn't work

### LINKED ISSUE #12037: Stats: Day/Week/Month/Year: Tapping date bars doesn't change data
### Expected behavior
When looking at stats in Day/Month/Week/Year, tapping each bar should change the stats shown to reflect the day I'm trying to look at.

### Actual behavior
In some cases, tapping the bar changes which bar is highlighted, but doesn't change the stats shown.

For example, this screenshot shows 1 view and 1 visitor, but the bar selected shows close to 30 visitors.
Screen Shot 2019-07-01 at 4 30 12 PM

This has been tricky to reproduce, but I've had most luck with opening a site I've looked at stats for recently. Sometimes, the app loads the stats (it will give a loading screen between tapping on bars) and everything is fine. Sometimes, it doesn't try to load anything. 
Tapping the < and > arrows seems to work even when tapping the bars doesn't.

Here's a video:
statsstuck

##### Tested on iPhone SE, (iOS 12.2) running WPiOS 12.6 and 12.7
Four user reports so far, #2127720-zen, #2145011-zen, #2142342-zen, #2145194-zen
More detail on all those reports: https://wp.me/p7cLQ7-Da

danielebogo:
Fixes #12037 

This is an _hotfix_ and it fixes a problem when a user is trying to change a period selecting the chart bar.
I saw the delegate injected in the Chart View was nil. I think this is caused by the element passed trough the configuration object. This is owned by the function scope, which means it will be released when the function returns. Injecting the delegate reference in the view init fixes the problem.

## To test:
• Open Stats and select one of the period DWMY
• Tap the bars to change the date
• It should reload the data

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Charts/StatsBarChartConfiguration.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Charts/StatsBarChartConfiguration.swift
@@ -3,16 +3,15 @@ struct StatsBarChartConfiguration {
     let data: BarChartDataConvertible
     let styling: BarChartStyling
     let analyticsGranularity: BarChartAnalyticsPropertyGranularityValue?
-    let delegate: StatsBarChartViewDelegate?
     let indexToHighlight: Int?
 }
 
 extension StatsBarChartConfiguration {
     init(data: BarChartDataConvertible, styling: BarChartStyling) {
-        self.init(data: data, styling: styling, analyticsGranularity: nil, delegate: nil, indexToHighlight: nil)
+        self.init(data: data, styling: styling, analyticsGranularity: nil, indexToHighlight: nil)
     }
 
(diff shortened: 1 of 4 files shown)


danielebogo:
Thanks a lot @jklausa

jkmassel:
I’m not 100% sure the issue is resolved – it didn’t start reloading data until I changed the screen within the stats page, then it worked the way I expected.

stats-issue-video

danielebogo:
Thanks @jklausa and @jkmassel !
I made a small change. I finally got where the delegate was nil. It should be ok now!
Would you mind to give it another try?

frosty:
I should add, we can address that in a separate PR, but I think it really needs tackling ^

danielebogo:
@frosty absolutely!

frosty:
I created an issue for this: https://github.com/wordpress-mobile/WordPress-iOS/issues/12048

danielebogo:
Thanks everyone for reviewing this!

PR #12050, opened 2019-07-02
Unit Tests to Tracks analytics

etoledom:
Given the importance of analytics, it would be good to have an easy way to create unit tests for tracks events.

In this PR, this is achieved via adding a new `AppAnalytics` protocol to the AppEnvironment.

Events will need to be tracked via `Environment.current.analytics.track`.

A mock `TestAnalytics` can easily be injected in a test case setup:
`Environment.replaceEnvironment(analytics: TestAnalytics.self)`

And use `TestAnalytics.tracked` to assert for tracked events.

This PR also adds a few examples with tests for `PostEditorAnalyticsSession`.

**To test:**
- CI has to pass ✅ 
- Check if you like this approach 👍

### DIFF
--- a/WordPress/Classes/Utility/Environment/Environment.swift
+++ b/WordPress/Classes/Utility/Environment/Environment.swift
@@ -8,6 +8,9 @@ struct Environment {
 
     // MARK: - Globals
 
+    /// A type to track events for analytics
+    let analytics: AppAnalytics.Type
+
     /// A type that helps tracking whether or not a user should be prompted for an app review
     let appRatingUtility: AppRatingUtilityType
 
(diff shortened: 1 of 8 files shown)


koke:
I think the approach would be fine on its own, but `WPAnalytics` is already designed to support multiple trackers via the `WPAnalyticsTracker` protocol. I wonder if we could use that instead, and avoid registering the regular trackers on the unit test target, and registering a mock one instead, similar to your `TestAnalytics` one.

etoledom:
Thank you @koke for the review!

I wasn't aware about the `WPAnalyticsTracker`. I refactored TestAnalytics to conform to that protocol and rolled back the Environment changes.

This looks better now, since we don't have to edit the `WPAnalytic.track()` calls 🎉

etoledom:
Thank you!

PR #12051, opened 2019-07-02
Fixes an issue with the tour guide not initializing some values in time.

### LINKED ISSUE #11687: Site Creation: "Upload a site icon" tutorial step not clear.
## Description:

The step that explains to the user how to "Upload a site icon" is not explaining clearly to the user what needs to be tapped.

## Images:

This is how it looks in iOS:

IMG_9C9240FED0E9-1

This is how it looks in Android (notice the blue dot showing where to tap):

Screenshot_20190513-213714

diegoreymendez:
Fixes #11687 

This PR also fixes two issues that could cause the spotlight effect to never go away:

1. If we run the "Upload an Icon" tutorial multiple times in a row without ever tapping the icon.
2. If we run the "Upload an Icon" tutorial and then a different one without ever tapping on the icon.

## Testing:

1. Create a WP.com site.
2. When offered to, request assistance to set the site up.
3. Make sure the spotlight dot effect is visible when the "Upload an Icon" tutorial is running.
4. Try multiple combinations of running "Upload an Icon" tutorial multiple times in a row with and without tapping on the icon.  Make sure it behaves well.
5. Try multiple combinations of running "Upload an Icon" tutorial and other tutorials multiple times in a row with and without tapping on the icon.  Make sure it behaves well.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -101,7 +101,8 @@ open class QuickStartTourGuide: NSObject {
                                 self?.currentSuggestion = nil
 
                                 if accepted {
-                                    self?.start(tour: tour, for: blog)
+                                    self?.prepare(tour: tour, for: blog)
+                                    self?.begin()
                                     cancelTimer(false)
                                     WPAnalytics.track(.quickStartSuggestionButtonTapped, withProperties: ["type": "positive"])
                                 } else {
@@ -116,16 +117,13 @@ open class QuickStartTourGuide: NSObject {
         WPAnalytics.track(.quickStartSuggestionViewed)
     }
 
-    /// Starts the specified tour.
+    /// Prepares to begin the specified tour.
     ///
     /// - Parameters:
-    ///     - tour: the tour to start.
+    ///     - tour: the tour to prepare for running.
     ///     - blog: the blog in which the tour will take place.
-    ///     - immeditately: whether the tour must start immediately or not.  If this is `false`
-    ///         the tour will be initialized but the first step won't be executed.  The caller of
-    ///         this method must also call `run()` to effectively start the tour.
     ///
-    func start(tour: QuickStartTour, for blog: Blog, immediately: Bool = true) {
+    func prepare(tour: QuickStartTour, for blog: Blog) {
         endCurrentTour()
         dismissSuggestion()
 
@@ -135,17 +133,12 @@ open class QuickStartTourGuide: NSObject {
             fallthrough
         default:
             currentTourState = TourState(tour: tour, blog: blog, step: 0)
-
-            if immediately {
-                showCurrentStep()
-            }
         }
     }
 
-    /// Runs the current tour.  Only works after calling `startTour` with parameter `immediately` set to
-    /// `false`, and if the tour is in the first step.  Otherwise this method will do nothing.
+    /// Begins the prepared tour.  Should only be called after `prepare(tour:for:)`.
     ///
-    func run() {
+    func begin() {
         guard let state = currentTourState,
             state.step == 0 else {
 
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -116,7 +116,16 @@ open class QuickStartTourGuide: NSObject {
         WPAnalytics.track(.quickStartSuggestionViewed)
     }
 
-    func start(tour: QuickStartTour, for blog: Blog) {
+    /// Starts the specified tour.
+    ///
+    /// - Parameters:
+    ///     - tour: the tour to start.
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
```diff
@@ -116,7 +116,16 @@ open class QuickStartTourGuide: NSObject {
...
+    /// Starts the specified tour.
+    ///
+    /// - Parameters:
+    ///     - tour: the tour to start.
+    ///     - blog: the blog in which the tour will take place.
+    ///     - immeditately: whether the tour must start immediately or not.  If this is `false`
+    ///         the tour will be initialized but the first step won't be executed.  The caller of
+    ///         this method must also call `run()` to effectively start the tour.
+    ///
+    func start(tour: QuickStartTour, for blog: Blog, immediately: Bool = false) {
```
nheagy: Why add a parameter that is never used? Unless I'm mistaken, after this PR `immediately` is always false.

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
```diff
@@ -126,10 +135,26 @@ open class QuickStartTourGuide: NSObject {
...
+            if immediately {
+                showCurrentStep()
+            }
         }
     }
 
+    /// Runs the current tour.  Only works after calling `startTour` with parameter `immediately` set to
+    /// `false`, and if the tour is in the first step.  Otherwise this method will do nothing.
+    ///
+    func run() {
```
nheagy: `run` is a pretty ambiguous name, and doesn't fit the verbs used elsewhere in this class.

What about putting this guard at the start of `showCurrentStep` instead?

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
diegoreymendez: Actually this was a mistake in my PR, as the default value was meant to be `true`.  I've just fixed it.

There are two calls to this method.  The one that delays running the tutorial:

https://github.com/wordpress-mobile/WordPress-iOS/blob/e3ee8cd321feb74e37f69d012346aff898a64484/WordPress/Classes/ViewRelated/Blog/QuickStartChecklistViewController.swift#L102-L107

And the one that I didn't want to modify, which was previously and still is immediate:

https://github.com/wordpress-mobile/WordPress-iOS/blob/e3ee8cd321feb74e37f69d012346aff898a64484/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift#L104-L107

I could make the immediate one still make both calls if we prefer, but I wanted to bring up both scenarios are currently used.  Thoughts?

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
diegoreymendez: I considered using `initialize()` and `start()`, thus removing the `immediate` param and forcing both to always be called (immediately or not, it'd be up to the caller).  Maybe that's a better option?  Any suggestion to make the method names clearer is of course welcome!

As to moving the guard in this method into `showCurrentStep()`, I'm not sure I understand: these methods have different purposes (`showCurrentStep()` should be called for all steps, whereas `run()` should only be called to run the tutorial from the first step forward.  But maybe this point would become less important if we go for something like `initialize` / `start`.

diegoreymendez: Maybe `prepare` and `start`?

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
nheagy: I'd prefer that both use the same process to start a tour, as it keeps the API simpler. I'd like to break up `start(tour:blog:)` into two methods. The immediate start version can call both methods in sequence.

I propose:
`start(tour:blog:)` become `prepare(tour:blog:)`
`run()` become `begin()`

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
nheagy: Yeah, I liked those names. I prefer `begin` to `start` but it's a minor difference—start sounds immediate to me.

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
diegoreymendez: Changed the process so that it's now the same for both cases.

WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
diegoreymendez: Changed to `prepare()` and `begin()`.

diegoreymendez:
@nheagy - Ready for another look.

diegoreymendez:
Oh nm, I missed the approval.  Thanks for the review!

PR #12055, opened 2019-07-03
[Stats Refresh] Avoid delay when a chart bar is selected

### LINKED ISSUE #12048: Stats: Chart bars take a long time to reflect user interaction
### Actual behavior 

Related to #12037, after tapping a bar in a stats chart (for data that has already been cached), there's a long 1 - 2 second delay before the bar is highlighted to show the tap.

stats-bar-tap-1

### Expected behavior

I would expect the bar highlight to update the instant I tap it to show that I did tap it successfully (even if it takes a moment to update the data labels).

danielebogo:
Fixes #12048 

This PR fixes the delay caused when the chart bar is selected (only for cached data).
Also it introduces an improvement when the cache data is loaded at the beginning and a fix to avoid displaying an empty table if the overview is fetching and a new request is triggered.

sr-chart

## To test:
- Select a site and open Stats
- Select one of the period DWMY
- Select different dates using the chart bars
- If the data is cached it should update the view immediately

Update release notes:

### DIFF
--- a/WordPress/Classes/Stores/StatsPeriodStore.swift
+++ b/WordPress/Classes/Stores/StatsPeriodStore.swift
@@ -309,6 +309,7 @@ private extension StatsPeriodStore {
         // make a network call for whatever reason, we still want to load the data we have cached.
 
         guard shouldFetchOverview() else {
+            fetchingOverviewListener?(true, false)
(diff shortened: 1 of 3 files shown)


danielebogo:
Thanks @jkmassel

PR #12056, opened 2019-07-03
Gutenberg: Compatibility with React Native 0.60.0

jtreanor:
Gutenberg PR is here: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1184

To test:

Update release notes:


WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -517,6 +517,8 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
             DDLogWarn(message)
         case .error:
             DDLogError(message)
+        case .fatal:
+            break
```
jtreanor: I'm not sure why the build was failing for me here. I was seeing an error that said the switch statement must be exhaustive.

etoledom: I did an update to the bridge without merging the change to WPiOS 😞 
Since there are not `DDLogFatal()`, I'd rather mix the `.error` and `.fatal` in the same case.

Something like:
` case .error, .fatal:`

etoledom:
Left a comment but I don't think it's a blocker for this PR.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
jtreanor: I've updated this now.

Issue #11846, opened 2019-06-03
Refresh issues when UploadManager uploads images in Aztec.


diegoreymendez:
This issue was spawned off the discussion here: https://github.com/wordpress-mobile/WordPress-iOS/pull/11839#issuecomment-498223891

## Original Report

The original report by @yaelirub is as follows:

Test 1:
The media upload resumed when the app returned to foreground but the image overlay stayed.
![media_background](https://user-images.githubusercontent.com/1335657/58799167-372e7880-860d-11e9-9d54-8a248e737f5d.gif)

~Test 2:
The media upload resumes when the app was relaunched but the image was never attached to the post and only a placeholder is visible.
![media_from_load](https://user-images.githubusercontent.com/1335657/58799308-adcb7600-860d-11e9-9a19-3e7370a7ae6c.gif)
This seem to happen only in Aztec. Gutenberg works~

Test 2 was moved to a separate issue: #12060.

## Subtasks

- [x] #12058 Fix _Test 1_ only
- [ ] #12110 Try to add tests

shiki:
I created a separate issue for the _Test 2_ since it seems like an entirely different issue. 

→ https://github.com/wordpress-mobile/WordPress-iOS/issues/12060

PR #12058, opened 2019-07-03
Aztec Media: Clear existing errors when restarting

shiki:
Closes part of #11846. 

When a media upload fails and the upload is retried, the error label “Failed to insert media” never disappears even if the retry succeeded. This covers this scenario by clearing the error label when an upload is started or restarted.

| Before | After |
|--------|-------|
| old       |   new    |

## Testing

1. Go offline.
2. Create a post and add an image. Do not leave the editor. 
3. Go online. The image upload should be restarted. 

Confirm that the error label is no longer shown. 

## Release Notes


shiki:
Thank you, Yael!

PR #12059, opened 2019-07-03
Fix alternate app icons

jkmassel:
Fixes an issue causing the app to be rejected by Apple. Also fixes an issue causing custom icons not to work on iPad.

**To test:**
- Run on iPhone simulator and ensure all alternate icons work
- Run on iPad Pro 12.9" simulator and ensure all alternate icons work
- In App Store Connect, note that `12.8.0.1.1` passes validation

Update release notes:


frosty:
Thanks for addressing this! I can’t review properly as I don’t have my laptop with me, but a few questions based on looking at the code:

* Are the AppIconViewController changes necessary? The non-iPad plist section should still be present on both platforms?
* I think the ‘pt’ suffix is missing off some of the 83.5 filenames in the plist

jkmassel:
@frosty – good questions!

>Are the AppIconViewController changes necessary? The non-iPad plist section should still be present on both platforms?

While I was testing, this was the only way to get it to work. I'm not sure what changed, but you're correct – I've removed the additional code path!

> I think the ‘pt’ suffix is missing off some of the 83.5 filenames in the plist

I figured out what was up with that – most of the `83.5` files don't have a `pt` on them (possibly to keep the file name lengths the same?) The pride file was the only one with the `pt` suffix. I removed the suffix on that filename to bring it in line with the others, and it seems to be working!

PR #12066, opened 2019-07-04
Default to Gutenberg upon opening a block based post

etoledom:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1186

This PR flips the "default for new posts" switch to true the first time the block editor is launched, if the switch is off.

The alert text was changed to match https://github.com/wordpress-mobile/gutenberg-mobile/issues/1186#issuecomment-508231868

The logic is:
- The alert will be shown if:
  - The Gutenberg default switch is off
  - The post has gutenberg blocks
  - Is first time opening a gutenberg post

When the alert is shown, we automatically flip the switch located in App Settings.

This will trigger the tracks event `AppSettingsGutenbergEnabled` but I'd say that it is still desirable, unless we want a new tracks event for this particular case?

I'm unsure if we should add this to the release notes.

**To test:**

- Set `UserDefaults.standard.setValue(false, forKey: "Gutenberg.InformativeDialog")` to be triggered on app lunch. (on `applicationWillEnterForeground` is helpful to easily reset it)
- Make sure the gutenberg switch is off on App Settings
- Open a new post
  - It should open Aztec
- Open a post without gutenberg blocks
  - It should open Aztec
- Open a post with gutenberg blocks
  - It should open gutenberg and show the alert
- Open again the same post
  - It should open gutenberg without showing the alert
- Open a new post
  - It should open gutenberg
- Check the switch on AppSettings
  - It should be ON
- Flip the AppSettings switch to OFF
- Open a gutenberg post
  - It should open gutenberg
  - It should NOT show the alert
  - it should NOT flip the switch back to ON

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/KeyValueDatabase.swift
+++ b/WordPress/Classes/Services/KeyValueDatabase.swift
@@ -7,9 +7,14 @@ import Foundation
 /// list objects.
 protocol KeyValueDatabase {
     func object(forKey defaultName: String) -> Any?
-    func bool(forKey: String) -> Bool
     func set(_ value: Any?, forKey defaultName: String)
     func removeObject(forKey defaultName: String)
+}
+
+extension KeyValueDatabase {
+    func bool(forKey key: String) -> Bool {
+        return object(forKey: key) as? Bool ?? false
+    }
 }
 
 // MARK: - Storage implementations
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Services/KeyValueDatabase.swift
```diff
@@ -7,6 +7,7 @@ import Foundation
 /// list objects.
 protocol KeyValueDatabase {
     func object(forKey defaultName: String) -> Any?
+    func bool(forKey: String) -> Bool
```
koke: This should be in a protocol extension since it’s just a wrapper that will be the same for all cases

etoledom: Nice catch! Updated 👍

etoledom:
> Since it seems like an odd case, and it would only happen once, I think it's fine?

As agreed, we will merge this as a short lived feature, so we won't worry about these edge cases.
Thank you!

PR #12067, opened 2019-07-04
[Stats Refresh] Avoid deselect action when a chart bar is selected

danielebogo:
Refs. #11852

This PR removes the possibility to deselect a selected chart bar. Unfortunately there's no property to enable/disable this feature and there's no public method to override the behaviour.
For this reason I replaced the default tap gesture implementing a new one owned by the `StatsBarChartView` itself.

sr-deselect-bar

## To test:
- Open a DWMY period and select a date from the chart
- Try to tap on the same bar verifying the bar is not deselected

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Charts/StatsBarChartView.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Charts/StatsBarChartView.swift
@@ -120,7 +120,7 @@ class StatsBarChartView: BarChartView {
 private extension StatsBarChartView {
     func setupGestures() {
         gestureRecognizers = gestureRecognizers?.filter { gesture in
-            if let gesture = (gesture as? UITapGestureRecognizer) {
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Stats/Charts/StatsBarChartView.swift
```diff
@@ -117,6 +118,26 @@ class StatsBarChartView: BarChartView {
 // MARK: - Private behavior
 
 private extension StatsBarChartView {
+    func setupGestures() {
+        gestureRecognizers = gestureRecognizers?.filter { gesture in
+            if let gesture = (gesture as? UITapGestureRecognizer) {
```
nheagy: `(gesture as? UITapGestureRecognizer)` doesn't need to be in parentheses.

danielebogo:
Thanks @nheagy !

Issue #12047, opened 2019-07-02
Site Creation: Pressing enter sometimes not doing anything


diegoreymendez:
This issue was initially reported by @aerych 

### Expected behavior

Pressing ENTER in a field should always perform an action, be it moving first responder to the next field, or submitting the form.

### Actual behavior

Sometimes pressing ENTER does nothing.

This happens in step 1, when entering the topic for the blog.
This also happens in step 2 for both fields (title and tagline).

### Steps to reproduce the behavior

Just create a new WP.com site and try this in steps 1 and 2.

##### Tested on iPhone Xs, iOS 12.3.1, WPiOS 12.7 (DEV)

diegoreymendez:
We implemented this for the verticals screen.  We still need to implement this for the Basic Information screen.

PR #12071, opened 2019-07-05
Pressing enter in the verticals screen now performs an action.

diegoreymendez:
This is the first PR to fix: https://github.com/wordpress-mobile/WordPress-iOS/issues/12047

Pressing enter in the Site Verticals screen will now find attempt to submit the vertical selection with a perfect match.

enterInSiteVerticals

## Testing:

1. Go to the sites list.
2. Create a new WP.com site.
3. At the verticals selection screen type any input and press enter / tap return.
4. If you have a connection, the form should submit.  If you have no connection, nothing should happen.

Try combining pressing enter with tapping on a cell almost simultaneously - it shouldn't crash.

## Update release notes:


aerych:
@diegoreymendez 
 Is there a better text option for the button other than "return"?  Maybe "Next" since tapping it or pressing enter should move you to the next step?

diegoreymendez:
@aerych - I though I had changed it to "Next" but I'll need to double check when I'm back from AFK... in the meanwhile and just to verify, have you seen this after running the demo or in the gif?

Bottom line: suggestion makes a ton of sense, and it was in my plans.  We'll need to review the design after this PR though as there's a "Next" button right above the keyboard too.

diegoreymendez:
Thanks @aerych.  I deleted a comment I had posted since it was off... I'll review when back from AFK.

diegoreymendez:
@aerych - The return button is now "Done".

I realize "Next" may feel like the obvious option here, but "Done" is currently more consistent with another PR I'm pushing for the next step (in which we have two fields and "Next" is used for moving to the next field.

diegoreymendez:
I'm pushing since there's an approval (and so I can push another PR), but feel free to keep commenting @aerych and I'll address.

aerych:
> The return button is now "Done".

Works for me!

> I'm pushing since there's an approval

Its all good sir! :)  Cheers!

PR #12077, opened 2019-07-08
Numerous color palette tweaks

nheagy:
Refs #11683

This changes numerous colors throughout the app to improve the look with the new color palette. I can provide a PDF of changes for review, if you'd like.

image
image
image

To test:
- Build app and look through all the top-level screens. Particularly:
  - all links should be the new blue
  - all table views should have the new gray background
  - more text should be very dark, and a more true neutral, no dark-blue text, etc.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
+++ b/WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
@@ -245,10 +245,13 @@ extension WPStyleGuide {
         let followIcon = Gridicon.iconOfType(.readerFollow, withSize: size)
         let followingIcon = Gridicon.iconOfType(.readerFollowing, withSize: size)
 
-        let tintedFollowIcon = followIcon.imageWithTintColor(.primary(shade: .shade40))
-        let tintedFollowingIcon = followingIcon.imageWithTintColor(.success(shade: .shade50))
+        let normalColor = UIColor.primary
+        let highlightedColor = UIColor.primaryDark
+        let selectedColor = UIColor.success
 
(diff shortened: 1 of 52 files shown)


WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
```diff
@@ -245,10 +232,13 @@ extension WPStyleGuide {
...
         let followingIcon = Gridicon.iconOfType(.readerFollowing, withSize: size)
 
-        let tintedFollowIcon = followIcon.imageWithTintColor(.primary(shade: .shade40))
-        let tintedFollowingIcon = followingIcon.imageWithTintColor(.success(shade: .shade50))
+        let normalColor = UIColor.primary
+        let highlightedColor = UIColor.primaryDark
+        let selectedColor = UIColor.success
 
-        let highlightIcon = followingIcon.imageWithTintColor(.primaryLight)
+        let tintedFollowIcon = followIcon.imageWithTintColor(normalColor)
```
frosty: Is there a reason why these (and the other reader icons) are set as tinted images rather than setting the button tint color?

nheagy: Only because I'm trying to make minimal changes on the frozen branch. Thoughts?

nheagy:
Updated to fix the test, @frosty—let me know what you think.

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
frosty: Okay, sure - can we make a note to maybe look at updating that next time?

PR #12083, opened 2019-07-08
[Stats Refresh] Period details: Show all data on rotation

### LINKED ISSUE #12072: [Stats Refresh] Posts and Pages > View more: list truncate to 10 posts when switching to landscape
User reported an issue with the post display under Stats in the WordPress.com iOS App.

### Expected behavior
All posts will list correctly in landscape stats view and scroll down as normal.

### Actual behavior
Upon switching to landscape, list is truncated to 10 posts and scrolling past that is not possible without refreshing the page by pulling down from the top.

Switching back to portrait does not fix the issue, and also requires refreshing the page.

Issue does not appear if you start in Landscape mode.

1. Go to Stats
2. Switch view to "Days" or another time-based view.
3. Click "View More" to see the full "posts and pages" list.
4. Switch to landscape mode.

ScoutHarris:
Fixes #12072 

This fixes an issue where on rotation the Period details views would revert to showing only 10 rows.

The issue was caused by Period being refreshed via `viewWillTransition` in `SiteStatsDashboardViewController`, resulting in the data to be overwritten in `StatsPeriodStore`.

This changes Period to only refresh data if the view is visible. 

To note, the original issue specifically states Posts and Pages, but it actually affected _all_ Period details.

To test:
- On a site with a lot of data (so you get more than 10 detail data rows), go to Period, and select `View more` for any stat.
- Verify all the data is shown (i.e. is not limited to 10 rows).

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -15,6 +15,7 @@
 * Stats Periods: Fixed crash when the Countries map displayed one country only
 * Added a selection of user customizable app icons. Change it via Me > App Settings > App Icon.
 * Update the app's colors using the Muriel color pallete.
+* Stats Periods detail views: Fixed an issue where rotation would truncate data.
 
 12.7
 -----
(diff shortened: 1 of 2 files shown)


ScoutHarris:
Hey @jkmassel - this will need to go in the next 12.8 release. Thank you!!

PR #12043, opened 2019-07-02
Fix/fix code after merge

danielebogo:
This PR fixes some broken methods after the merge 12.7 -> develop.
The affected classes are:
• OverviewCell
• SiteStatsPeriodViewModel
• SiteStatsTableHeaderView

Update release notes:

- [ ] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #12089, opened 2019-07-09
[Stats Refresh] Fix Insights Annual details date bar

ScoutHarris:
Ref #11962, #12043

This fixes the date bar on Insights Annual details view that was broken when 12.7 was merged to `develop`. An attempt was made to fix it on #12043 but it wasn't quite right.

To test:
- Follow the steps noted on #11962 .
- Specifically what was broken - `mostRecentDate` is used to prevent the date bar from going past the site's most recent year. For example, on a site with no date for 2019, we don't want to use `Date()` for the current date as that will result in a broken view. So, on a site with no data for 2019:
   - Go to Insights > This Year > View more.
   - Verify the date bar only goes up to the year the site has data for.

this_year

details

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Helpers/StatsPeriodHelper.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Helpers/StatsPeriodHelper.swift
@@ -7,9 +7,9 @@ class StatsPeriodHelper {
         return cal
     }()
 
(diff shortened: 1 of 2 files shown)


ScoutHarris:
Sorry @jkmassel . One more for 12.8.

frosty:
Thanks for the fix, @ScoutHarris! Apologies we didn't get this quite right!

PR #12093, opened 2019-07-10
Issue/11646 post list revamp fixes

leandroalonso:
Ref #11646

This PR fixes the icon problem and the post list scrolling to the top after viewing a post.


nheagy:
Merging despite failing test because that test is also failing in the target branch. 

Thanks @leandroalonso!

PR #12094, opened 2019-07-10
[Stats Refresh] Correct DWMY period date 

danielebogo:
Refs. #11852

This fixes a problem when a period interval is selected. Sometimes the calculated date used to load the data was wrong, mostly when the selected period was `week`.
When a period is selected the API returns the whole intervals list. Every interval already has the correct start date, which is the right one to use or, in case the selected period is `week`, it's the right date to calculate the right week end date.

`StatsPeriodHelper` will return the right end date using 2 methods:
```
func calculateEndDate(from currentDate: Date, offsetBy count: Int = 1, unit: StatsPeriodUnit) -> Date?
```
and
```
func endDate(from intervalStartDate: Date, period: StatsPeriodUnit, offsetBy count: Int = 1) -> Date
```
The last one is used when a chart interval start date is passed as argument. Because this fix has been applied on the overview only I didn't delete the firs one. As soon the same logic to get the start date from the intervals list is applied to the Stats detail and Post detail, the first method can be deleted.

## To test:
- Select a blog with data (en.blog is a perfect example)
- Test every DWMY period
- Switch between the dates using the arrows or tapping on the chart bar
- Compare the loaded values with Calypso

Update release notes:


ScoutHarris:
Hey @danielebogo . 

In comparing this branch vs `release/12.8`, one thing I've noticed is the `Likes` is not getting updated correctly. It seems to happen only on weeks, and I could only repro it on en.blog. Maybe because it has so much data? Anyway, it only happen when there is no cached data. 

If you switch periods and return, the correct `Likes` will eventually show. It may be that `Likes` is now getting update _before_ the query actually finishes?

The dashes are shown correctly as the query is running. However, it then shows 0 afterwards. 

likes_loading

likes_done

danielebogo:
I saw the likes API takes a lot sometimes to respond. This happens on the `12.7.1` too.
> I haven't gone through the code yet, but functionally the selection of bars is definitely fixed for me, and the figures are showing correctly. I am seeing the same 0 likes issue as @ScoutHarris though.

I think that might happen because the summary API call, which contains **0** as likes value, finishes after the Likes API call.

> I've also seen an issue when I switched from Weeks to Days and also to Years. The view appeared showing 0 data, and selecting any bars did nothing at all. I had to leave the tab, visit another tab, then come back again and it would reload and work ok. Any ideas?

Not sure but this might be for the same reason where one response overrides the others and/or because the store doesn't save the data using a tree structure in its state.

danielebogo:
Sometimes I saw that a request for summary or likes in a specific period can be finalised after the period is changed. This also might be the reason why the **0** appears.

ScoutHarris:
Hey @danielebogo .

I don't think this is related to this PR because it's happening on `release/12.8` as well, but it appears the Overview chart somehow got changed to use weeks Sun - Sat instead of Mon - Sun. 

It _seems_ simple enough. I don't suppose it could be fixed in this PR? If not, we can spin up another fix.

overview_dates

WordPress/Classes/ViewRelated/Stats/Shared Views/Date Chooser/SiteStatsTableHeaderView.swift
```diff
@@ -4,6 +4,10 @@ protocol SiteStatsTableHeaderDelegate: class {
     func dateChangedTo(_ newDate: Date?)
 }
 
+protocol SiteStatsTableHeaderDateButtonDelegate: SiteStatsTableHeaderDelegate {
```
ScoutHarris: > As soon the same logic to get the start date from the intervals list is applied to the Stats detail and Post detail

I'm assuming one of these protocols will be removed when :point_up:  is done? And the accompanying (now defunct) `dateChangedTo` method in `SiteStatsPeriodTableViewController`?

ScoutHarris: Oh, nevermind. I see what's happening. Sorry!

danielebogo: Exactly, after the overview is done the idea is to start with the post chart and stats detail and use the new protocol, moving the logic from the header to the view model. After that the other delegate and `func calculateEndDate(from currentDate: Date, offsetBy count: Int = 1, unit: StatsPeriodUnit) -> Date?` can be removed.

danielebogo:
Thanks everyone for reviewing this!

PR #12095, opened 2019-07-10
Site Creation: RETURN now performs an action in the Basic Information screen.

### LINKED ISSUE #12047: Site Creation: Pressing enter sometimes not doing anything

diegoreymendez:
Fixes #12047 

Pressing enter now performs an action in the Basic Information screen.

nextAndDone

## Known issues:

There is a bit of a design problem here since there's a physical button named "Next" that could be confusing for the user, since it doesn't do the same thing as the "Next" button in the virtual keyboard.

This is out of scope for the purpose of this issue, but I went ahead and opened an issue to tackle this separately in our next iteration: https://github.com/wordpress-mobile/WordPress-iOS/issues/12117

## Testing:

Simply follow the steps shown in the animation above.

## Update release notes:


WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
```diff
@@ -231,7 +258,7 @@ extension SiteInformationWizardContent: UITableViewDataSource {
     }
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return Rows.count()
+        return Row.count()
```
yaelirub: Rows worked better here but Row works better for everything else. Just a note.

WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
```diff
@@ -245,15 +272,19 @@ extension SiteInformationWizardContent: UITableViewDataSource {
     }
 
     private func configure(_ cell: InlineEditableNameValueCell, index: IndexPath) {
-        if Rows.title.matches(index.row) {
+        if Row.title.matches(index.row) {
             cell.nameLabel.text = TableStrings.site
             cell.valueTextField.attributedPlaceholder = attributedPlaceholder(text: TableStrings.site)
+            cell.valueTextField.delegate = self
+            cell.valueTextField.returnKeyType = .next
```
yaelirub: I'm going to leave a comment about this in the separate ticket for the design issue

WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
diegoreymendez: Yeah, I thought the same thing.

The reason I changed it anyway is because enums are normally singular, and additionally this would normally be:

```swift
return Row.allCases.count
```

which makes it ok that it's singular.  I can change this line to be as described if you prefer.

WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
```diff
@@ -293,6 +324,35 @@ extension SiteInformationWizardContent: InlineEditableNameValueCellDelegate {
     }
 }
 
+extension SiteInformationWizardContent: UITextFieldDelegate {
+    func textFieldShouldReturn(_ textField: UITextField) -> Bool {
+        for (index, row) in Row.allCases.enumerated() {
+            guard let rowTextField = valueTextField(for: row) else {
+                let errorMessage = "We expect all rows to have `valueTextField`.  Please review the logic."
```
yaelirub: Would it be useful to add the row it's failing for?

WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
yaelirub: I think `Row.allCases.count` is better

WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
```diff
@@ -293,6 +324,35 @@ extension SiteInformationWizardContent: InlineEditableNameValueCellDelegate {
...
+                CrashLogging.logMessage(errorMessage, properties: nil, level: .error)
+                assertionFailure(errorMessage)
+                continue
+            }
+
+            guard rowTextField == textField else {
```
yaelirub: Could we bring this guard up to `guard let rowTextField = valueTextField(for: row)` ?

diegoreymendez: The problem with doing that is that this case shouldn't trigger an error message, while the case above should.

yaelirub: 👍

WordPress/Classes/ViewRelated/Site Creation/SiteInfo/SiteInformationWizardContent.swift
```diff
@@ -293,6 +324,35 @@ extension SiteInformationWizardContent: InlineEditableNameValueCellDelegate {
...
+                continue
+            }
+
+            let indexPath = IndexPath(row: index + 1, section: 0)
+
+            guard let nextTextField = valueTextField(at: indexPath) else {
+                goNext()
```
yaelirub: Ok now this is a little confusing. 😕 I thought something was off here until I realized that goNext is to next page. Can we update the name here to goNextStep or nextStep or something like that?

diegoreymendez: Yes of course, I'll DM you to agree on a better name.

diegoreymendez:
Thanks a lot for the thorough review @yaelirub ! :)

PR #12120, opened 2019-07-14
Allow building with Xcode 11 by conditionally disabling Zendesk

nheagy:
Zendesk's SDK is the only thing stopping the code from building with Xcode 11. This is @frosty's `try/ios-13` branch with one change to move the `XCODE11` flag into `xcconfig` files.

This will be helpful for working on iOS 13-specific projects such as dark mode and `Sign In with Apple`.

Thanks to @davedelong for coming up with the `_DD_IOS_*` solution.

To test:
- build with Xcode 10
- ensure the app still works fine
- ensure Zendesk still works
- build with Xcode 11
- ensure the app still works fine
- if the build worked then Zendesk was disabled successfully
  - consider if everything is just _better that way_

Update release notes:
- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`. **(there weren't)**


loremattei:
Tested and worked for me as well!

nheagy:
Thanks, @frosty, @loremattei

PR #12123, opened 2019-07-15
[Gutenberg] GutenbergSettings refactor

etoledom:
**UPDATE:**

This PR (now) is a refactor of GutenbergSettings and the auto-enable gutenberg Dialog logic, moving all the decision logic out of the `InformativeDialog` extension, making the logic of `auto-enable gutenberg` more generic and accessible.

To test:
- The test cases from https://github.com/wordpress-mobile/WordPress-iOS/pull/12066 should still hold true.
- The case described on https://github.com/wordpress-mobile/WordPress-iOS/pull/12066#pullrequestreview-257903408 should have been fixed with these changes.

You can reset the gutenberg selection calling:
`UserDefaults().set(nil, forKey: GutenbergSettings.Key.enabled)`

----
**Old Description:**

This PR is the WPiOS version of https://github.com/wordpress-mobile/WordPress-Android/pull/10209

WordPressKit side PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/166

On this PR:
- Editor preference is sent to the remote server each time the user switch the pref in App Settings
- The preference is sent to the remote server each time the editor is opened in the app. This was necessary since some users may have already switched to GB, and the new prompt dialog is not shown in these cases. We may want to revise this decision, and send data to the remote server when the dialog is shown only by putting a little bit of logic around it.
- The local app preference is set for all of the wpcom powered sites available in the app

I took the opportunity to move all the decision logic out of the `InformativeDialog` extension, making the logic of `auto-enable gutenberg` more generic and accessible.

To test:
I left a small log for testing purposes.

- Preferably, log in with an alternative account that doesn't have hundreds of posts linked.
- Switch the editor to gutenberg / aztec
  - Check in the report card for that user that the posts are being set accordingly on the backend.
- Open the editor (Gutenberg or Aztec)
  - Check that the backend sets the mobile editor setting accordingly with the app switch selection (not necessarily with the editor that was opened)
  - As an example: If you switch to enable gutenberg, and open non-gutenberg post, it will open Aztec, but since the switch is set to Gutenberg, it should set the backend to Gutenberg anyway.

The test cases from https://github.com/wordpress-mobile/WordPress-iOS/pull/12066 should still hold true, and the case described on https://github.com/wordpress-mobile/WordPress-iOS/pull/12066#pullrequestreview-257903408 should have been fixed with these changes.

Update release notes:


jkmassel:
👋 I'm bumping this to `13.0` as part of the `12.9` code freeze. If you'd like to release it as part of `12.9`, just merge it to `release/12.9` and DM me – I'll be happy to cut a new beta release.

etoledom:
Thank you @jkmassel !
We really want this to land on 12.9, so I changed the target branch.

I'll let you know when it's ready to merge 👍

hypest:
FWIW, I tested both the app-settings and the via-editor flows and I verified that the correct setting is found on the user's report card on the backend 🎉.

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -49,18 +83,54 @@ class GutenbergSettings {
...
+
+        queue.async {
+            allBlogs.forEach({ (blog) in
+                self.postEditorSettings(currentSettings, to: blog)
+                Thread.sleep(forTimeInterval: delay)
+            })
+        }
+    }
+
+    private func getAllBlogs() -> [Blog] {
```
koke: I think you only want WordPress.com blogs here, so maybe check `defaultAccount.blogs` instead?

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -22,17 +27,46 @@ class GutenbergSettings {
...
+        set(enabled) {
+            if enabled {
+                wasGutenbergEnabledOnce = true
+            }
+            database.set(enabled, forKey: Key.enabled)
+        }
+    }
+
+    /// True if gutenberg editor has been enabled at least once
+    var wasGutenbergEnabledOnce: Bool {
```
koke: I think this approach works, but I believe we could also check that `database.object(forKey: Key.enabled) != nil` to check if the value has ever been set

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -49,18 +83,54 @@ class GutenbergSettings {
...
+    ///
+    func setToRemote() {
+        let currentSettings: EditorSettings = isGutenbergEnabled ? .gutenberg : .aztec
+        let allBlogs = getAllBlogs()
+        let delay: TimeInterval = allBlogs.count > 5 ? 0.3 : 0
+
+        queue.async {
+            allBlogs.forEach({ (blog) in
+                self.postEditorSettings(currentSettings, to: blog)
+                Thread.sleep(forTimeInterval: delay)
```
koke: Why are we adding this delay?

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: Nice! Seems to work great 🎉

Podfile.lock
```diff
@@ -18,11 +18,6 @@ PODS:
   - DoubleConversion (1.1.5)
   - Down (0.6.6)
   - Folly (2018.10.22.00):
-    - boost-for-react-native
-    - DoubleConversion
-    - Folly/Default (= 2018.10.22.00)
-    - glog
-  - Folly/Default (2018.10.22.00):
```
etoledom: These changes came from a `rake dependencies`
Doing the same on `develop` generates the same diff.

There might have been a bad merge at some point

etoledom: CI is failing and seems related to this 🤔

etoledom: Removing `Pods` folder did the trick.

etoledom:
Thank you @koke for the review and for bringing up the conversation yesterday.

Following yesterday's conversation, I temporarily rolled back the sync to Remote from this PR, but kept the refactor. I think this refactor will help us on the next steps.

Also moved the target branch back to `develop` and updated the title and description accordingly.

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -22,17 +23,38 @@ class GutenbergSettings {
...
+    var wasGutenbergEnabledOnce: Bool {
+        return database.object(forKey: Key.enabled) != nil
+    }
+
+    func shouldAutoenableGutenberg(for post: AbstractPost) -> Bool {
+        return  post.containsGutenbergBlocks() && !wasGutenbergEnabledOnce
+    }
+
+    func setGutenbergEnabledIfNeeded() {
+        if isGutenbergEnabled == false {
```
koke: What do you think about getting rid of `toggleGutenberg()` and moving the analytics to the `set` accessor of the property? I find the "toggle" semantics awkward since we know her we want to enable it (if it's disabled), and on app settings we also know the specific value to be set from the UI.

WordPress/Classes/ViewRelated/Me/AppSettingsViewController.swift
```diff
@@ -200,7 +200,9 @@ class AppSettingsViewController: UITableViewController {
 
     func toggleGutenberg() -> (Bool) -> Void {
         return { [weak self] _ in
-            GutenbergSettings().toggleGutenberg()
+            let settings = GutenbergSettings()
+            settings.toggleGutenberg()
```
koke: I'm curious, is this just a style improvement, or was something not behaving as expected?

etoledom: Right! It was calling the method to sync the setting with remote too (a second call on the same instance) but now that second call was removed so this is not really needed anymore. 👍

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: Sounds good to me 👍 
`.toogleGutenberg()` sounds practical for the switch, but since it holds the analytics logic, is not the best for other uses. I'll give it a try to remove it, or its dependency on tracks.

etoledom:
@koke - Pushed a commit that handles the last coupe of comments. Ready for another look!

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -22,21 +23,34 @@ class GutenbergSettings {
...
-    func isGutenbergEnabled() -> Bool {
-        return database.object(forKey: gutenbergEditorEnabledKey) as? Bool ?? false
+    /// True if gutenberg editor is currently enabled
+    var isGutenbergEnabled: Bool {
+        get {
+            return database.bool(forKey: Key.enabled)
+        }
+        set(isEnabled) {
+            if isGutenbergEnabled != isEnabled {
+                trackSettingChange(to: isEnabled)
```
koke: `WPAnalytics.refreshMetadata()` will read from `isGutenbergEnabled` to set the `gutenberg_enabled` property, so it needs to be called after the new value is set

etoledom: That's right! My bad 🙏 
Fixed in the last commit.

etoledom:
Thank you!

PR #12124, opened 2019-07-15
Posts and Pages search: Update to Muriel color scheme

leandroalonso:
This PR fixes the color and some design details in posts and pages search.

To test:

- In the posts section, perform a search and check if the design matches the blueprint of #11646
- In the pages section, perform a search and check if the design matches the blueprint of #11646

| Before  | After |
|    |  |


WordPress/Classes/ViewRelated/Post/PostCompactCell.swift
```diff
@@ -83,6 +86,12 @@ class PostCompactCell: UITableViewCell, ConfigurablePostView {
         contentView.backgroundColor = innerView.backgroundColor
     }
 
+    private func setupSeparator() {
+        WPStyleGuide.applyBorderStyle(separator)
+        contentView.heightAnchor.constraint(equalToConstant: contentView.frame.height + WPStyleGuide.separatorHeight).isActive = true
```
frosty: I'm not sure this should be necessary, if you ensure the InnerView is pinned to the top of the content view, and the bottom of the InnerView is pinned to the top of the separator?

WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
```diff
@@ -282,6 +282,17 @@ class AbstractPostListViewController: UIViewController,
             stopGhostIfConnectionIsNotAvailable()
             showNoResultsView()
         }
+
+        updateBackgroundColor()
+    }
+
+    // Update controller's background color to avoid white lines
```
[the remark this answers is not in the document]
leandroalonso: Changed! Let me know what you think.

WordPress/Classes/ViewRelated/Post/PostCompactCell.swift
leandroalonso: This code was adding the separator height to the cell height and realign the labels (to maintain the same margins as before the hardcoded separator). But yeah, complex and not so clear.

I removed it and managed to achieve the same result with a heigh constant in the xib.

PR #12133, opened 2019-07-16
Update release-toolkit and its dependency scheme

loremattei:
This PR updates `Pluginfile` to handle the changes in https://github.com/wordpress-mobile/release-toolkit/pull/60. 
`RMagick` is not a default dependency for `release-toolkit`, so this PR adds it as an optional one to be installed when the promo screenshot feature is required. 

# To Test
1. Run `bundle install --without screenshots` and verify that RMagick is not installed. 
2. Run `bundle install` and verify that RMagick is not installed. 
3. Run `bundle install --with screenshots` and verify that RMagick is installed. 

Update release notes:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,13 +1,18 @@
 GIT
   remote: https://github.com/wordpress-mobile/release-toolkit
-  revision: 074bac6f11ae07a293d73eed71eede716480ef6e
-  tag: 0.2.0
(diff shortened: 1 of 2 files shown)


loremattei:
Thank you for the review @jkmassel.
I merged the fix in `release-toolkit`, cut a release branch and updated the reference here.

PR #12134, opened 2019-07-16
[Hotfix] Fix launch image background color

frosty:
We noticed that the released version of 12.8 wasn't showing a launch screen for fresh installs of the app (users would just see black until the app loads). This PR fixes that by updating the background color of the launch image storyboard. It was still referencing an old asset color that no longer exists:

image

I've updated it to use `WordPress-50`. cc @nheagy 

**To test:**

* Build and run a fresh install (simulator is fine).
* Check you see the launch image when the app launches


frosty:
Thanks @etoledom!

PR #12137, opened 2019-07-16
[Stats Refresh] Insights: fix missing stats

ScoutHarris:
Fixes #n/a

The WPKit PR (https://github.com/wordpress-mobile/WordPressKit-iOS/pull/169) fixes issues where parsing Insights json would fail if any piece of data was missing. 

I noticed the issue for these stats:
- This Year
- Most Popular
- Posting Activity

I didn't actually see these stats fail, but I did apply the same fix here as well:
- All Time
- Today

To test:
- On a site with very little data (but not _no_ data), go to Insights.
- For the stats listed above, verify they appear if they have data (can be verified in Calypso).

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -43,8 +43,8 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    #pod 'WordPressKit', '~> 4.2.0'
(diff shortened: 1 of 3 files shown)


ScoutHarris:
@jkmassel - this PR, along with WPKit PR https://github.com/wordpress-mobile/WordPressKit-iOS/pull/169, need to go in 12.9 please.

PR #11294, opened 2019-03-18
Auth: Updates so an account can be created with just an authtoken.

aerych:
This PR updates WPiOS to accommodate changes in WordPressAuthenticator-iOS that remove the `username` field from the credentials enum.  These changes allow an account to be created with _just_ an auth token.  Taken together it will allow a user to use magic links requested from outside of the mobile app (e.g. Calypso, wp-admin, etc.). 

To test:
- Test the wpcom login flows.  Ensure you can log in with an email address + password, and by requesting a magic link.  Ensure that no information is missing from the epilogue. 
- Shut down the simulator then relaunch it.
- Confirm you are still logged in in the Me tab.

Bonus: _After_ requesting a magic link uninstall and reinstall the app, then try the magic link. It _should_ log you in correctly. 

Needs review: @ScoutHarris here is the companion to the WordPressAuthentication-iOS PR.

I've opened this as a draft since the Podfile is pointing to a commit hash. I'll update it once the auth pod PR lands.

PR #12139, opened 2019-07-16
Issues/fixes login tracking

aerych:
This PR addresses an issue where tracks events were being miss-reported or lost during log in. For additional background see p9OQAC-h0-p2.

In https://github.com/wordpress-mobile/WordPress-iOS/pull/11294 I added changes to support authenticating with magic links generated from outside of the app. To limit the changes needed in the client code I relied on a temporary username, just a UUID, to create the initial WPAccount instance since the account was required to fetch account details. The actual username was obtained when syncing account details.   I missed how this could impact our tracks events. 

Problem one: We refresh tracks meta data when creating an account during login. This meant that tracks events would be de-anonymized with a UUID for a username. Ouch.
Problem two:  Even though we refresh meta data when updating account details, its done asynchronously leading to some events being dispatched before meta data was updated. 

This PR fixes the issue by ensuring we do not refresh meta data when there is a temporary username, and that meta data is refreshed before we call the success block when refreshing account details. 

To test:
You can confirm the glich by signing in with a magic link in the simulator.  I used a test account with a gmail email address and just checked gmail in safari in the simulator.  Check the tracks events in the tracks live view and confirm you only see the wpcom generated magic link events.  (Be paitent. It will take a few minutes for events to be picked up.)

Apply the patch and log in with a magic link again. Confirm that you see all expected wpios client events in the live view. 

I've targeted `develop` but we probably want to cherry pick these in to the `release/12.9` branch.

@astralbodies, would you do the honors?

### DIFF
--- a/WordPress/Classes/Services/AccountService.m
+++ b/WordPress/Classes/Services/AccountService.m
@@ -278,10 +278,10 @@ NSString * const WPAccountEmailAndDefaultBlogUpdatedNotification = @"WPAccountEm
         [self setupAppExtensionsWithDefaultAccount];
         dispatch_async(dispatch_get_main_queue(), ^{
             [WPAnalytics refreshMetadata];
+            if (success) {
+                success();
+            }
         });
-        if (success) {
-            success();
-        }
     } failure:^(NSError *error) {
         DDLogError(@"Failed to fetch user details for account %@.  %@", account, error);
         if (failure) {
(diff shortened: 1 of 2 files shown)


aerych:
Closing in favor of https://github.com/wordpress-mobile/WordPress-iOS/pull/12140

PR #12150, opened 2019-07-17
[Gutenberg] - Send unsupported blocks list to Tracks

etoledom:
WPiOS side of https://github.com/wordpress-mobile/gutenberg-mobile/issues/1174
Analog of Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10237

**UPDATE:**
The new property has been added just to EDITOR_SESSION_START event.

**Old description:**
A new property unsupported_blocks is added to the EDITOR_SESSION_* events, and it does hold the list of unsupported blocks found in the post.

**Note:**

Since the unsupported blocks list is stored at the beginning of the editor session, subsequent editions to the post won't affect this property.

If the user remove all blocks, then save and close the post, tracks will still track the original list of unsupported blocks. and also track: `has_gutenberg_blocks: 0` creating an inconsistency.

Talking with @daniloercoli , on Android the value of `has_gutenberg_blocks` is also recorded at the beginning of the editing session, so there won't be inconsistencies, even though the data is not accurate.

I decided to keep the real `has_gutenberg_blocks` value for now. Ideally we could update the list of unsupported blocks when we request the updated content to gutenberg.

To test:

- Start the app and open a GB post that does contain unsupported blocks.
- Check that the events `🔵 Tracked: editor_session*` record the list of unsupported blocks accordingly.
- Open a gutenberg post without unsupported blocks.
- Check that the property `unsupported_blocks` is not recorded.

Update release notes:

cc @daniloercoli

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -480,7 +480,7 @@ class AztecPostViewController: UIViewController, PostEditor {
         }
 
         if !editorSession.started {
-            editorSession.start(hasUnsupportedBlocks: false)
+            editorSession.start()
         }
     }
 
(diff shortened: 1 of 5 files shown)


koke:
As mentioned in https://github.com/wordpress-mobile/WordPress-Android/pull/10237#discussion_r304767178 let's make sure the property is set even if there are no unsupported blocks. Other than that it's looking good

etoledom:
Oh my bad! I misinterpreted that comment thinking that you preferred to keep it null.

Sent a new commit with the new logic 👍

etoledom:
Thank you!

PR #12151, opened 2019-07-17
Muriel: Add a `tableBackground` semantic color

frosty:
This PR just adds a new `tableBackground` semantic color to the Muriel palette, as we were referencing gray-0 all over the place. This should make it simpler if ever we need to update it in future.

**To test:**

* Check the code
* Build and run and check backgrounds look okay!

Update release notes:


frosty:
@danielebogo, would you be able to take a look at this as @nheagy is AFK?

frosty:
Thanks @danielebogo!

PR #12159, opened 2019-07-18
Stats: Use site timezone

### LINKED ISSUE #12018: [Stats Refresh] Use Site timezone instead of User timezone
From https://github.com/wordpress-mobile/WordPress-Android/issues/10117:

When using the value of "Today", the app should use site time zone instead of the app time zone. This is relevant for the "Overview" card which could (in some cases) be showing tomorrow's date as the last date in the graph. The same goes for the "Today" insights where we might be showing insights for "Tomorrow" in some cases and for the "Posting activity" where we need the date to set a range.
We need to let the user know that we're showing insights for a different day.

Design from [this](https://wp.me/paCL20-dH) P2: 

The date bar would, therefore, have two states

- When site timezone and local timezone are in agreement – in this condition, we show only the selected period (the current design)
- When site timezone and local timezone are in disagreement – in this condition, we show selected period and details about the site timezone offset (see below)

ios_date_bar

ScoutHarris:
Fixes #12018 

This changes all of Stats to use the site timezone instead of the device's timezone.

NOTE: The Post Stats endpoint does not honor the site timezone, but is in UTC. The result is that, depending on your local TZ, site TZ, and timing, the Post Stats and Latest Post Summary charts could be missing a day or have an extra day.

Corresponding WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/170

To test:

Test the following items on site with the timezone:
  - Before your local.
  - After your local (preferably a day ahead).
  - In UTC.
  - Same as your local.

**Date Bar:**
- [ ] If your timezone differs from the site, a 'Site timezone' label will appear in the date bar showing the offset from UTC.
- [ ] If your timezone is the same as the site, the 'Site timezone' label will not appear.

**Periods:**
- [ ] Verify changing the date via the bar works as expected.
- [ ] Verify changing the date via the chart works as expected.
- [ ] Compare data with Calypso. Especially Days as that is the one most affected.

**Insights:**
- [ ] Verify Today shows today's stats.
- [ ] Go to This Year > View more. Verify the date bar behaves as expected.
- [ ] Verify the Latest Post Summary chart is correct (taking the note above into consideration).

**Post Stats:**
Access Post Stats via Posts and Pages and Latest Post Summary.
- [ ] Verify the date bar behaves as expected.
- [ ] Verify the chart behaves as expected.
- [ ] Verify the chart is correct (taking the note above into consideration).

**Timezone:**
- [ ] Change site timezone via site settings. Verify the date bar is updated accordingly.

Update release notes:


ScoutHarris:
@SylvesterWilmott - screenshots:

iphone

ipad

frosty:
Thanks for all your work getting this together @ScoutHarris! I haven't had much of a chance to test yet, but I just had a few thoughts regarding the style of the date bar having seen it in action (partly for @SylvesterWilmott):

* I think we should reduce the size of the timezone label text – maybe `subhead` or `footnote`?
* I think the standard way of showing UTC offsets would be with no spaces either side of the + / - – e.g. `UTC+7`.
* However, we also need to account for locations that have a 30 or 45 minute offset. Possibly `+0730` or `+07:30` in those cases?

SylvesterWilmott:
>I think we should reduce the size of the timezone label text – maybe subhead or footnote?

It should be `subhead` in the screenshots above. Is the issue in how large the bar is because it is slightly larger than it should be. @ScoutHarris currently the bar is at 68pt, can we reduce it to 60pt?

>I think the standard way of showing UTC offsets would be with no spaces either side of the + / - – e.g. UTC+7.

Looking into this, that is correct. The correct formatting would be `UTC+1` `UTC-2` etc

>However, we also need to account for locations that have a 30 or 45 minute offset. Possibly +0730 or +07:30 in those cases?

It looks like there's a few standard formats for this. For readability let's go with:

`UTC+07:30`

frosty:
> It should be subhead in the screenshots above. Is the issue in how large the bar is because it is slightly larger than it should be. @ScoutHarris currently the bar is at 68pt, can we reduce it to 60pt?

I think it's `callout`, but it doesn't look much smaller than the main date text, just a lighter weight. It felt like it could be smaller?

Screenshot 2019-07-19 at 13 35 22

Thanks for the clarification on the other items :)

ScoutHarris:
@frosty , @SylvesterWilmott - To get it to fit in 60pt I had to squinch it a bit. Meaning:
- Remove 1pt from top and bottom margins (from 12 to 11).
- Reduce space between date and timezone labels to 1pt (was 2).
- Use the `footnote` font (was `callout`). 

tz_1
tz_2
tz_3


(@frosty - thank you for noticing the minutes issue.)

ScoutHarris:
@danielebogo - Since the timezone format has been changed, I closed #12163 and updated the `accessibilityLabel` here.

danielebogo:
Hey @ScoutHarris ! LGTM!
I tested it again and everything looks great! Even the accessibility label.
Little note about the conflicts with `SiteStatsInsightsTableViewController`: keep only `let year = viewModel?.annualInsightsYear()` from develop.
:shipit:

SylvesterWilmott:
>To get it to fit in 60pt I had to squinch it a bit.

Does it still look the same as before when the new label isn't included? Ie 44pt without the new label and 60pt with it. I'm not 100% on the new text size, could you include full screenshots here of iPhone X and iPad please?

ScoutHarris:
Hey @SylvesterWilmott .

> Does it still look the same as before when the new label isn't included? Ie 44pt without the new label and 60pt with it.

Yes.

> include full screenshots

Here ya go.

iphone_date

iphone_with_tz

ipad_date

ipad_with_tz

ScoutHarris:
@SylvesterWilmott this is a pretty significant change. So I'm going to go ahead and merge it. If there are any required changes to the timezone display, just let me know and I'll do them asap.

SylvesterWilmott:
>this is a pretty significant change. So I'm going to go ahead and merge it. If there are any required changes to the timezone display, just let me know and I'll do them asap.

Sure. The only change is really to move the text size of the site timezone label back to what we had before but to retain the 60pt height of the bar.

PR #12166, opened 2019-07-19
[Gutenberg] - PR 01 - Move gutenberg switch to site settings

etoledom:
This PR moves the Editor selector (Gutenberg switch) from App Settings to Site Settings.

Even though it makes sense to add it as part of the WRITING section, I decided to make it its own section to be able to keep displaying the footer. (cc @iamthomasbishop )
editor_setting

This is a UI only change, and the setting is still stored globally.
On a next PR, we will make the setting per-site.

To test:
- Check that the Switch displays correctly on the Site Settings screen.
- Check that the switch behaves correctly in the same way than it did before.

Update release notes:
(Will update release notes when the whole feature is ready)
- [*] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.


koke:
> This is a UI only change, and the setting is still stored globally

I took a look at the code and it seems good, but I'm hesitant to merge this with that behavior. If we want to keep PRs small maybe we can start on the other side: keep the global switch but move the storage to be per site, and have the app settings toggle save on all sites (😬). Then this PR would start toggling each specific site.

etoledom:
Updated this PR to be dependent on https://github.com/wordpress-mobile/WordPress-iOS/pull/12171, which implements Editor settings per Site and Sync with remote

etoledom:
@koke would you mind taking another look?
I'm using a feature branch now.

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -6,10 +6,13 @@ import Foundation
...
-        guard let api = blog.wordPressComRestApi(), let siteID = blog.dotComID?.intValue else {
-            let error = NSError(domain: "EditorSettingsService", code: 0, userInfo: [NSDebugDescriptionErrorKey: "Api or dotCom Site ID not found"])
-            failure(error)
-            return
+        guard let api = blog.wordPressComRestApi() else {
+            // SelfHosted non-jetpack sites won't sync with remote.
+            return success()
+        }
+        guard let siteID = blog.dotComID?.intValue else {
+            let error = NSError(domain: "EditorSettingsService", code: 0, userInfo: [NSDebugDescriptionErrorKey: "dotCom Site ID not found"])
```
koke: Not a blocker, but a block with an account should always have an ID, right? We might want to add an assertion here besides the `guard`

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -96,3 +96,16 @@ class GutenbergSettings {
...
+@objc(GutenbergSettings)
+class GutenbergSettingsBridge: NSObject {
+    @objc(setGutenbergEnabled:forBlog:)
+    static func setGutenbergEnabled(_ isEnabled: Bool, for blog: Blog) {
+        GutenbergSettings().setGutenbergEnabled(isEnabled, for: blog)
+    }
+
+    @objc(postSettingsToRemoteForBlog:)
+    static func postSettingsToRemote(for blog: Blog) {
+        GutenbergSettings().postSettingsToRemote(for: blog)
```
koke: Is there any case where we want to set the value locally and not save to the server? Do we need two separate methods or should `setGutenbergEnabled` automatically post to remote?

WordPress/Classes/ViewRelated/Blog/SiteSettingsViewController+Swift.swift
```diff
@@ -100,25 +100,34 @@ extension SiteSettingsViewController {
...
     }
 
+    @objc(getEditorSettingsSectionFooterView)
+    func editorSettingsSectionFooterView() -> UIView {
+        let footer = commonFooterView()
+        footer.textLabel?.text = NSLocalizedString("Edit new posts and pages with the block editor.", comment: "Explanation for the option to enable the block editor")
+        return footer
+    }
+
+    private func commonFooterView() -> UITableViewHeaderFooterView {
```
koke: Nitpicky, but when I read `commonFooterView` I imagine a view shared for all footers. Considering the usage, I'd suggest naming the method `makeFooterView` ([API Guidelines](https://swift.org/documentation/api-design-guidelines/#begin-factory-name-with-make)), or have a `configure*` method like we do with `WPStyleGuide`

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: > should setGutenbergEnabled automatically post to remote?

This was the first implementation. I decided to keep them separated since they are two different tasks (local storage/remote sync).

It's also helpful for unit tests

> Is there any case where we want to set the value locally and not save to the server?

Technically yes for self-hosted non-jetpack sites, but that's being handled internally anyway.

WordPress/Classes/ViewRelated/Blog/SiteSettingsViewController+Swift.swift
etoledom: There's a `WPStyleGuide.configureTableViewSectionFooter` that I preferred not to modify since it's widely used.
`makeFooterView` songs great! 👍 
A local `configure*` sounds good too but a `make*` will save the extra init call. Just a small detail I guess? :)

I'll push the name change.

WordPress/Classes/Services/EditorSettingsService.swift
etoledom: Right, we already guarded against a `wordPressComRestApi` api, so at this point it should definitely have a dotComID.
I'll add the assertion 👍

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
koke: They are two different tasks, but I don't think the UI that calls this should care. Now that we have both persistence and remote sync, `GutenbergSettings` is starting to act more as a Service class, and I wonder if we should fold all the logic into `EditorSettingsService`, although that's more of a question for the next PR 😁

Issue #12138, opened 2019-07-16
Stats: Better Error Messaging for Jetpack Remote Install Failure


thehenrybyrd:
### Expected behavior

When I try to install Jetpack on my site, I expect an informative error if it fails.

### Actual behavior

Instead all reasons for failure give:

![image-43-1](https://user-images.githubusercontent.com/15107387/61315657-eadb7880-a807-11e9-9810-6b253e9a3f7b.jpg)

Known reasons for failure include blocked XML-RPC (such as a 403 at /xmlrpc.php) and Jetpack already existing in the database (the user gets "the jetpack plugin is already installed” when trying to install from wp-admin).
It would be very helpful to return errors that describe the problem we had with installing Jetpack.

designsimply:
Raised in priority due to several user reports and a steady uptick in metrics for `install_jetpack_remote_failed `. 

<sup>(internal reference: p4a5px-2pQ-p2)</sup>

nheagy:
Not sure how to implement this one. Hoping to involve design to give some direction.

rachelmcr:
I opened a separate issue https://github.com/wordpress-mobile/WordPress-iOS/issues/12497 to track the case where Jetpack is already installed and connected, since it seems like that should be handled differently from installation errors.

designsimply:
To test:

1. Add a self-hosted site with no plugins to the WordPress for iOS app.
1. Install the [All In One WP Security & Firewall](https://wordpress.org/plugins/all-in-one-wp-security-and-firewall/) plugin.
1. Go to WP Security > Firewall.
1. Check the box for "Completely Block Access To XMLRPC" and save settings.
1. Go to `/xmlrpc.php` for the site and verify a 403 Forbidden error is returned.
1. On the web, activate Jetpack and connect it to a WordPress.com account.
1. In the app, go to Stats and try to finish the Jetpack Setup there.

**Result:** Jetpack install flow fails with the generic error "Jetpack could not be installed at this time" and that error could be more helpful. ([2min](https://cloudup.com/cGLqbWPl759))
<sup>Tested with WPAndroid 13.5-rc-1 on Pixel 3 Android 10 on `uptight-gorillas.jurassic.ninja` with xmlrpc.php returning 403 Forbidden.</sup>

Would it be possible to detect the 403 and respond with something like the following in this case?

[...]

PR #12168, opened 2019-07-19
[Jetpack Remote Install] Add more info in analytics failure event 

### LINKED ISSUE #12158: Analytics: need more info about the `install_jetpack_remote_failed` stat for debugging Jetapack Install failures
We are seeing ~180 failed Jetpack Install requests per day in the `install_jetpack_remote_failed` stat (internal reference: p4a5px-2pQ-p2#comment-9969).

Can we please add add the site URL to the tracks props for success and failure events? This will allow for qualitatively debugging install issues by looking at the actual login screen.

Is there anything else we can add to the error logs to help debug Jetpack Install failures? (also see internal reference: p7cLQ7-Dx-p2#comment-1252).

h/t @gravityrail for the suggestion!

danielebogo:
Fixes #12158

This PR adds the jetpack error name and the site URL as properties when the Jetpack Install fails.
If we want more info from the error we get from the API response we need to update WP Kit as well.

## To test:
- Try to install Jetpack on a website that might fail (Use a slow connection profile to generate a timeout)

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Jetpack/Install/JetpackRemoteInstallViewController.swift
+++ b/WordPress/Classes/ViewRelated/Jetpack/Install/JetpackRemoteInstallViewController.swift
@@ -73,7 +73,9 @@ private extension JetpackRemoteInstallViewController {
             case .success:
                 WPAnalytics.track(.installJetpackRemoteCompleted)
             case .failure(let error):
-                WPAnalytics.track(.installJetpackRemoteFailed)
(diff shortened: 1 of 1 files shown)


elibud:
I think happiness was also asking for a more user visible error, can you check with them?

danielebogo:
@elibud this **Issue** [12138](https://github.com/wordpress-mobile/WordPress-iOS/issues/12138)?

elibud:
Yup, that one, can we please address it too? Thanks

PR #12171, opened 2019-07-19
[Gutenberg] Sync editor setting per site

etoledom:
This PR is a work-in-progress, targeting a feature branch to implement https://github.com/wordpress-mobile/gutenberg-mobile/issues/1233

This PR adds `mobileEditor` and `webEditor` to the Blog model, and are synchronized with the server.

Related `WordPressKit` PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/166 

The editor selection setting is now per site, but the switch is still App-wide, so (just for now), changing the `Use Block Editor` switch will iterate through all the user's blogs and set the appropriate value. This will change when we merge https://github.com/wordpress-mobile/WordPress-iOS/pull/12166, that move the switch UI to the Site Settings section.

**Note:** As mentioned on https://github.com/wordpress-mobile/gutenberg-mobile/issues/1233:
> Check the values of the web and mobile preferences. If both are set to gutenberg launch GB for new post, otherwise Aztec.

This means that switching the `Gutenberg Editor` ON, won't necessarily open gutenberg unless the web setting is also set to gutenberg.

To test:
- Check that the editor setting get's sync from remote when the app lunches or a site is selected.
- Check that the proper value is stored when switching the `Use Block Editor` switch.

Update release notes:

### DIFF
--- a/WordPress/Classes/Models/Blog+Editor.swift
+++ b/WordPress/Classes/Models/Blog+Editor.swift
@@ -1,0 +1,27 @@
+import Foundation
+
+enum MobileEditor: String {
+    case aztec
+    case gutenberg
+}
+
+enum WebEditor: String {
+    case classic
+    case gutenberg
+}
+
+extension Blog {
+    struct Editor {
+        fileprivate let blog: Blog
+        var mobile: MobileEditor? {
+            return MobileEditor(rawValue: blog.mobileEditor ?? "")
+        }
+        var web: WebEditor? {
+            return WebEditor(rawValue: blog.webEditor ?? "")
+        }
+    }
+
+    var editor: Editor {
+        return Editor(blog: self)
+    }
+}
--- a/WordPress/Classes/Models/Blog+Editor.swift
+++ b/WordPress/Classes/Models/Blog+Editor.swift
@@ -27,9 +27,7 @@ extension Blog {
     var editor: Editor {
         return Editor(blog: self)
     }
-}
 
-@objc extension Blog {
     @objc var isGutenbergEnabled: Bool {
         return self.editor.mobile == .gutenberg
     }
--- a/WordPress/Classes/Models/Blog+Editor.swift
+++ b/WordPress/Classes/Models/Blog+Editor.swift
@@ -28,3 +28,9 @@ extension Blog {
         return Editor(blog: self)
     }
 }
+
+@objc extension Blog {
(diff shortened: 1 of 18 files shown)


WordPress/Classes/Models/Blog+Editor.swift
```diff
@@ -0,0 +1,34 @@
...
+        var web: WebEditor? {
+            return WebEditor(rawValue: blog.webEditor ?? "")
+        }
+        func setMobileEditor(_ newValue: MobileEditor) {
+            blog.mobileEditor = newValue.rawValue
+        }
+    }
+
+    var editor: Editor {
+        return Editor(blog: self)
```
ctarda: Sorry if this is a stupid question, but would this be a reference cycle?

etoledom: I don't think so? 🤔 
Blog.Editor is a Value type. It does hold a reference to a Blog object, but then Blog just construct an Editor once and gives it back to the caller, and its lifetime will die together with the stack frame, so the ref count for Blog should go down accordingly.

I did a small test just to be sure and I didn't notice any Blog object leaking.

Please let me know if I'm missing something!

etoledom:
Thank you @ctarda !

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -0,0 +1,51 @@
...
+            failure(error)
+            return
+        }
+
+        let service = EditorServiceRemote(wordPressComRestApi: api)
+        service.getEditorSettings(siteID, success: { (settings) in
+            do {
+                try self.saveRemoteEditorSettings(settings, on: blog)
+            } catch {
+                failure(error)
```
koke: While reviewing #12166, I found that this will cause a crash. When the mobile setting is not set, the `saveRemoteEditorSettings` will throw an error. This will call the `failure` handler but it doesn't return, so it will also call `success` a couple of lines later.

This means that the handler in `BlogService` will call `dispatch_group_leave` once too many times and the app will crash.

etoledom: Good catch @koke !
I got that crash while working on the other PRs, and it was fixed here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12189

Good thing that we are working on a feature branch :)

PR #12175, opened 2019-07-19
Reader: Fix card borders to account for display scale

### LINKED ISSUE #12153: Reader: borders on posts appear blurry, border under Manage bar is too large.
In the Reader there are two issues with borders: the border under the Reader appears 3px when viewing a 3x screenshot, it should only be 1 actual px to be consistent with other borders/separators.

Second, the borders on posts in the Reader appear blurry because it appears the borders are sitting on half-pixels or something causing them to not appear sharp. There's a light gray "inner" border above or below each real border.

On this screen (post content w/ PII removed)
Artboard Copy 2

Zoomed in:
Screen Shot 2019-07-17 at 5 59 38 PM
Screen Shot 2019-07-17 at 6 00 07 PM

frosty:
Fixes #12153, by reducing the width of the border on Reader cards.

reader-borders-1

**To test:**

* Build and run, and view a stream in the Reader
* Ensure that the cards have single pixel borders on both 2x and 3x screen types (e.g. iPhone XS and iPhone XS Max).

@etoledom, would you mind reviewing?

cc @mattmiklic 

Update release notes:

(I didn't update release notes and I think it's so minor it can just be included in "Improved color scheme consistency" that I've added in another branch)

### DIFF
--- a/WordPress/Classes/Extensions/UIView+Borders.swift
+++ b/WordPress/Classes/Extensions/UIView+Borders.swift
@@ -3,7 +3,7 @@ extension UIView {
         let borderView = makeBorderView(withColor: bgColor)
 
         NSLayoutConstraint.activate([
-            borderView.heightAnchor.constraint(equalToConstant: 1.0 / UIScreen.main.scale),
+            borderView.heightAnchor.constraint(equalToConstant: .hairlineBorderWidth),
             borderView.topAnchor.constraint(equalTo: topAnchor),
             borderView.centerXAnchor.constraint(equalTo: centerXAnchor),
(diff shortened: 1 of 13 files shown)


mattmiklic:
This looks good!

@etoledom good catch! In addition to reducing the line weight, any card that sits directly adjacent to the top toolbar should only have a border on the bottom, not the top or sides.

frosty:
Thanks for catching that @etoledom! I updated the border in that area too, and I extracted all of these `1.0 / UIScreen.main.scale` usages to a `.hairlineBorderWidth` property on `CGFloat`.

PR #12182, opened 2019-07-22
Stats: fix card border heights

### LINKED ISSUE #12152: Stats: Inconsistent borders on backgrounds
While examining some of the new color work I noticed that in the new Stats design, the borders on tops of the "cards" that contain each section are larger than the borders on the bottom. When viewing 3x screenshots, the bottom border is 1 actual px tall, but the top border is 2px. I think in iOS terms that's 0.5px and 1px, but I'm not sure. The smaller border is the correct one, as it's consistent with what's seen in other sections of the app.

As seen on this screen (PII removed):
Artboard

Zoomed in with a pixel grid turned on:
Screen Shot 2019-07-17 at 5 52 08 PM
Screen Shot 2019-07-17 at 5 57 24 PM

ScoutHarris:
Fixes #12152 

The stat separator lines are now scaled according to the device (as we do other lines).

To test:
- Run the app on devices of different [scale](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/image-size-and-resolution/).
- Verify the top and bottom lines on stat cards scale accordingly. Ex:
  - iPhone XS: lines are .33pt.
  - iPhone XR: lines are .5pt.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Extensions/WPStyleGuide+Stats.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Extensions/WPStyleGuide+Stats.swift
@@ -36,6 +36,8 @@ extension WPStyleGuide {
 
         static func configureViewAsSeparator(_ separatorView: UIView) {
             separatorView.backgroundColor = separatorColor
+            separatorView.constraints.first(where: { $0.firstAttribute == .height })?.isActive = false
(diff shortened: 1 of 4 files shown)


ScoutHarris:
**iPhone XS:**

xs_1
xs_2

---

**iPhone XR:**

xr_1
xr_2

SylvesterWilmott:
Thanks for the fix @ScoutHarris

PR #12185, opened 2019-07-23
Strip "+" and "00" from country prefix code before sending it to backend for validation

### LINKED ISSUE #12181: Domain Credit: domain registration phone validation fails if you use a "+" symbol in the country code
To prepare:

1. Setup a WordPress.com blog and apply a plan to it.
1. Do not register the domain yet.
1. Have an external keyboard connected to your device.

To test:

1. In the app, go to My Sites > {site} > Register Domain.
1. Fill out the "Domain contact information" form.
1. Using an external keyboard, enter "+1" for the country code.
1. Tap the "Register domain" button.

**Result:** if a "+" symbol is included in the country code in the "Domain contact information" form then the form fails saying, "Please enter a valid phone number." ([42s](https://cloudup.com/cwLRUkbYXm5))

**Suggestions:** either allow users to enter a plus symbol in the country code field and account for that in form validation or change the country code field into a dropdown to match Calypso.

please-enter-a-valid-phone-number
<sup>Tested with WPiOS 12.8.1 on iPhone 6S iOS 12.3.1</sup>

<sub>h/t @rfaile313 for the report in Slack p1563811379004200-slack-cellular-happiness</sub>

jklausa:
Fixes #12181. 

To test:
* Go to `RegisterDomainDetailsViewModel.swift`, line 190 and add `return;` there, to prevent you from actually redeeming the domain, and put a breakpoint there
* Follow the steps outlined in #12181 
* Verify that no validation error gets shown and the breakpoint you set gets hit


WordPress/Classes/ViewRelated/Domains/Register/RegisterDomainDetails/ViewModel/RegisterDomainDetailsViewModel.swift
```diff
@@ -464,8 +464,17 @@ class RegisterDomainDetailsViewModel {
     private func formattedPhoneNumber() -> String {
         let section = sections[SectionIndex.phone.rawValue]
         let countryCode = section.rows[CellIndex.PhoneNumber.countryCode.rawValue].editableRow?.value ?? ""
+
+        let strippedCountryCode = countryCode
+            .replacingOccurrences(of: Constant.phoneNumberCountryCodePrefix, with: "")
+            .replacingOccurrences(of: "00", with: "")
```
pinarol: we can add a check here to replace the 00 only if it is the prefix, it is a very very rare case but I see a country code as 500 here https://www.internationalcitizens.com/international-calling-codes/

jklausa:
@pinarol thanks for the great catch with only stripping the leading-zeros! 

after chatting briefly with @elibud I also added a mapping between country-codes and the prefixes, so we'll pre-populate the field when users change their country :)

ready for another look!

WordPress/Classes/ViewRelated/Domains/Register/RegisterDomainDetails/ViewModel/RegisterDomainDetailsViewModel+CountryDialCodes.swift
```diff
@@ -0,0 +1,277 @@
...
+        "VN": 84,
+        "VU": 678,
+        "WF": 681,
+        "WS": 685,
+        "YE": 967,
+        "ZA": 27,
+        "ZM": 260,
+        "ZW": 263]
+
+    func countryCodePrefix(for countryCode: String) -> Int? {
```
pinarol: Country Code is better represented with String so we can return String here

jklausa: Good point!

jklausa:
@pinarol thanks for the review!

PR #12187, opened 2019-07-23
Implement UISwipeActions / Remove MGSwipeTableCell

frosty:
Now that we've bumped our base SDK up to iOS 11, we can drop MGSwipeTableCell and use native UISwipeActions instead. This PR does just that. Apologies that it's a little long, but the majority of the changes are just cookie cutter across the different types of notifications and their tests.

before-after-swipes

A few notes:

* I was kind of confused by our existing setup for these actions. From what I can tell, every notification action type (such as Like Comment, Like Post, Reply, Spam) implemented a method returning a button, but it seems as though we were only using Approve and Trash (and mark as read, but that's separate). I've implemented contextual actions for all of them for consistency, but I'm not sure they're all needed right now. When testing, please ensure the actions still show up in places like the comment detail screen.
* You may notice the text size by default is a little smaller than previously. It's not configurable by us, but it does support dynamic type, so that should be fine.
* It also doesn't appear to support accessibility labels or hints unfortunately, but system swipe actions are accessible whereas I don't think MGSwipeTableCell was. If you're interested in testing this, with Voice Over active: tap a row to select it, make a two finger rotate gesture to bring up the 'rotor', and change it to Actions, then swipe up and down with one finger. It should read the available actions such as Approve, Trash, etc.

**To test:**

* Build and run
* Check unit tests pass and look ok (I had to update them)
* Ensure that you can do the same things you could previously do with notifications:
- [ ] Swipe to mark as read
- [ ] Swipe to mark as unread
- [ ] Swipe to approve
- [ ] Swipe to unapprove
- [ ] Swipe to trash
* Check you can undo a trash
* Check you can still trigger notification actions from other places: comment notification detail views, and push notification actions

Update release notes:

(I don't think this is directly user facing as it should be largely invisible)

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -149,7 +149,6 @@ target 'WordPress' do
     pod 'Gifu', '3.2.0'
     pod 'GiphyCoreSDK', '~> 1.4.0'
     pod 'HockeySDK', '5.1.4', :configurations => ['Release-Internal', 'Release-Alpha']
-    pod 'MGSwipeTableCell', '1.6.8'
     pod 'MRProgress', '0.8.3'
     pod 'Starscream', '3.0.6'
     pod 'SVProgressHUD', '2.2.5'
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -56,7 +56,6 @@ PODS:
     - HockeySDK/DefaultLib (= 5.1.4)
   - HockeySDK/DefaultLib (5.1.4)
   - lottie-ios (2.5.2)
-  - MGSwipeTableCell (1.6.8)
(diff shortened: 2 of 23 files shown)


frosty:
@ctarda @etoledom I may have just thought of an approach I prefer here...  will ping you when I've updated.

frosty:
@ctarda @etoledom I did a refactor! Instead of implementing a handler, the actions now just implement a title and an optional color. It's less code, and it means the actions don't need any knowledge of the UI context in which they're going to be displayed.

PR #12189, opened 2019-07-23
[Gutenberg] - PR 02 - Editor default and migration

etoledom:
Third step to implement https://github.com/wordpress-mobile/gutenberg-mobile/issues/1233
Based on https://github.com/wordpress-mobile/WordPress-iOS/pull/12166

This PR implements the editor default logic and the migration of the local setting to remote, when the remote setting is empty.

To test:
- Open the test user's report card
  - Check that nothing is set on `SELECTED EDITORS` for mobile platform.
  - You can press `reset` to remove the setting if there's any.

**Sync and Migration from local to remote:**
- Login with a test user.
  - Check that the locally defined default editor was set on the report card.
- Toggle the switch 
  - Check that the new value was set on the report card.

**Sync from remote to local**
- Set a blog on Aztec and a blog on Gutenberg
- Uninstall the app, install and login again
- Check that the editor switch correspond to the editor set on the report card for both blogs

**Auto-enabled dialog**
- Open a Gutenberg post on a blog originally set to use Aztec (from a clean install)
  - Check that the dialog appears and both the switch and the report card is set to gutenberg for that blog
- Open the other blog set to gutenberg.
- Switch it back to Aztec.
- Open a blog with gutenberg blocks.
  - Check that it opens with Gutenberg, but the switch and the report card remain as Aztec, and no dialog appears.

**Migration from global setting to per-site setting**
- On 12.9 set to Gutenberg:
  - Updating to 13.0, all blogs should default to gutenberg
  - No autoenable-dialog should appear for any blog.

- On 12.9 set to Aztec:
  - Updating to 13.0 should: 
    - Keep Aztec on older users (most users)
    - Auto-enable gutenberg for newer users

**New installs from 13.0 and beyond**
- The default editor should be Gutenberg for all newer users
- The default editor should be Aztec for all older users

**Note:** A newer user is defined as having a userID > 149287441

cc @daniloercoli

### DIFF
--- a/WordPress/Classes/Models/Blog+Editor.swift
+++ b/WordPress/Classes/Models/Blog+Editor.swift
@@ -29,6 +29,9 @@ extension Blog {
     }
 
     @objc var isGutenbergEnabled: Bool {
-        return self.editor.mobile == .gutenberg
+        guard let selectedEditor = editor.mobile else {
+            return GutenbergSettings().getDefaultEditor(for: self) == .gutenberg
+        }
+        return selectedEditor == .gutenberg
     }
 }
--- a/WordPress/Classes/Models/Blog+Editor.swift
+++ b/WordPress/Classes/Models/Blog+Editor.swift
@@ -11,27 +11,38 @@ enum WebEditor: String {
 }
 
 extension Blog {
-    struct Editor {
-        fileprivate let blog: Blog
-        var mobile: MobileEditor? {
-            return MobileEditor(rawValue: blog.mobileEditor ?? "")
+    static let mobileEditorKeyPath = "mobileEditor"
+    static let webEditorKeyPath = "webEditor"
+
+    /// The stored setting for the default mobile editor
+    ///
+    var mobileEditor: MobileEditor? {
+        get {
+            return rawValue(forKey: Blog.mobileEditorKeyPath)
         }
-        var web: WebEditor? {
-            return WebEditor(rawValue: blog.webEditor ?? "")
-        }
-        func setMobileEditor(_ newValue: MobileEditor) {
-            blog.mobileEditor = newValue.rawValue
+        set {
+            setRawValue(newValue, forKey: Blog.mobileEditorKeyPath)
         }
     }
 
-    var editor: Editor {
-        return Editor(blog: self)
+    /// The stored setting for the default web editor
+    ///
+    var webEditor: WebEditor? {
+        get {
+            return rawValue(forKey: Blog.webEditorKeyPath)
+        }
+        set {
+            setRawValue(newValue, forKey: Blog.webEditorKeyPath)
+        }
+    }
+
+    /// The editor to use when creating a new post
+    ///
+    var editor: MobileEditor {
+        return mobileEditor ?? GutenbergSettings().getDefaultEditor(for: self)
     }
 
     @objc var isGutenbergEnabled: Bool {
-        guard let selectedEditor = editor.mobile else {
-            return GutenbergSettings().getDefaultEditor(for: self) == .gutenberg
-        }
-        return selectedEditor == .gutenberg
+        return editor == .gutenberg
     }
 }
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -64,6 +64,17 @@ import Foundation
         }, failure: failure)
     }
 
+    func syncEditorSettingsForAllBlogs() {
+        let blogService = BlogService(managedObjectContext: managedObjectContext)
+        guard let blogs = blogService.blogsForAllAccounts() as? [Blog] else {
+            return
+        }
+        let blogsWithAccount = blogs.filter({ $0.account != nil })
+        for blog in blogsWithAccount {
+            syncEditorSettings(for: blog, success: {}, failure: { _ in })
+        }
+    }
+
     func api(for blog: Blog) -> WordPressComRestApi? {
         return blog.wordPressComRestApi()
     }
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -19,7 +19,7 @@ import Foundation
         let service = EditorServiceRemote(wordPressComRestApi: api)
         service.getEditorSettings(siteID, success: { (settings) in
             do {
-                try self.saveRemoteEditorSettings(settings, on: blog)
+                try self.update(blog, remoteEditorSettings: settings)
                 ContextManager.sharedInstance().save(self.managedObjectContext)
                 success()
             } catch EditorSettingsServiceError.mobileEditorNotSet {
@@ -30,12 +30,12 @@ import Foundation
         }, failure: failure)
     }
 
-    private func saveRemoteEditorSettings(_ settings: EditorSettings, on blog: Blog) throws {
+    private func update(_ blog: Blog, remoteEditorSettings settings: EditorSettings) throws {
+        blog.webEditor = WebEditor(rawValue: settings.web.rawValue)
         guard settings.mobile != .notSet else {
             throw EditorSettingsServiceError.mobileEditorNotSet
         }
         blog.mobileEditor = MobileEditor(rawValue: settings.mobile.rawValue)
-        blog.webEditor = WebEditor(rawValue: settings.web.rawValue)
     }
 
     private func migrateLocalSettingToRemote(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -71,7 +71,9 @@ import Foundation
         }
         let blogsWithAccount = blogs.filter({ $0.account != nil })
         for blog in blogsWithAccount {
-            syncEditorSettings(for: blog, success: {}, failure: { _ in })
+            syncEditorSettings(for: blog, success: {}, failure: { (error) in
+                DDLogError("Error saving editor settings: \(error)")
+            })
         }
     }
 
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -5,16 +5,13 @@ import Foundation
 }
 
 @objc class EditorSettingsService: LocalCoreDataService {
-
-    let database: KeyValueDatabase
     let wpcomApi: WordPressComRestApi?
 
     @objc convenience override init(managedObjectContext: NSManagedObjectContext) {
         self.init(managedObjectContext: managedObjectContext, wpcomApi: nil)
     }
 
-    init(managedObjectContext: NSManagedObjectContext, database: KeyValueDatabase = UserDefaults(), wpcomApi: WordPressComRestApi? = nil) {
-        self.database = database
+    init(managedObjectContext: NSManagedObjectContext, wpcomApi: WordPressComRestApi? = nil) {
         self.wpcomApi = wpcomApi
         super.init(managedObjectContext: managedObjectContext)
     }
@@ -54,7 +51,7 @@ import Foundation
 
     private func migrateLocalSettingToRemote(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
         if blog.editor.mobile == nil {
-            let settings = GutenbergSettings(database: database)
+            let settings = GutenbergSettings()
             let defaultEditor = settings.getDefaultEditor(for: blog)
             settings.setGutenbergEnabled(defaultEditor == .gutenberg, for: blog)
         }
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -5,8 +5,26 @@ import Foundation
 }
 
 @objc class EditorSettingsService: LocalCoreDataService {
-    @objc func syncEditorSettings(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
-        guard let api = blog.wordPressComRestApi(), let siteID = blog.dotComID?.intValue else {
+
+    let database: KeyValueDatabase
+    let wpcomApi: WordPressComRestApi?
+
+    @objc convenience override init(managedObjectContext: NSManagedObjectContext) {
+        self.init(managedObjectContext: managedObjectContext, wpcomApi: nil)
+    }
+
+    init(managedObjectContext: NSManagedObjectContext, database: KeyValueDatabase = UserDefaults(), wpcomApi: WordPressComRestApi? = nil) {
+        self.database = database
+        self.wpcomApi = wpcomApi
+        super.init(managedObjectContext: managedObjectContext)
+    }
+
+    @objc(syncEditorSettingsForBlog:success:failure:)
+    func syncEditorSettings(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
+        guard
+            let api = wpcomApi ?? blog.wordPressComRestApi(),
+            let siteID = blog.dotComID?.intValue
+        else {
             let error = NSError(domain: "EditorSettingsService", code: 0, userInfo: [NSDebugDescriptionErrorKey: "Api or dotCom Site ID not found"])
             failure(error)
             return
@@ -16,11 +34,13 @@ import Foundation
         service.getEditorSettings(siteID, success: { (settings) in
             do {
                 try self.saveRemoteEditorSettings(settings, on: blog)
+                ContextManager.sharedInstance().save(self.managedObjectContext)
+                success()
+            } catch EditorSettingsServiceError.mobileEditorNotSet {
+                self.migrateLocalSettingToRemote(for: blog, success: success, failure: failure)
             } catch {
                 failure(error)
             }
-            ContextManager.sharedInstance().save(self.managedObjectContext)
-            success()
         }, failure: failure)
     }
 
@@ -32,11 +52,20 @@ import Foundation
         blog.webEditor = settings.web.rawValue
     }
 
+    private func migrateLocalSettingToRemote(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
+        if blog.editor.mobile == nil {
+            let settings = GutenbergSettings(database: database)
+            let defaultEditor = settings.getDefaultEditor(for: blog)
+            settings.setGutenbergEnabled(defaultEditor == .gutenberg, for: blog)
+        }
+        postEditorSetting(for: blog, success: success, failure: failure)
+    }
+
     func postEditorSetting(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
         guard
             let selectedEditor = blog.editor.mobile,
             let remoteEditor = EditorSettings.Mobile(rawValue: selectedEditor.rawValue),
-            let api = blog.wordPressComRestApi(),
+            let api = wpcomApi ?? blog.wordPressComRestApi(),
             let siteID = blog.dotComID?.intValue
         else {
             let error = NSError(domain: "EditorSettingsService", code: 0, userInfo: [NSDebugDescriptionErrorKey: "Api or dotCom Site ID not found"])
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -45,12 +45,12 @@ import Foundation
         guard settings.mobile != .notSet else {
             throw EditorSettingsServiceError.mobileEditorNotSet
         }
-        blog.mobileEditor = settings.mobile.rawValue
-        blog.webEditor = settings.web.rawValue
+        blog.mobileEditor = MobileEditor(rawValue: settings.mobile.rawValue)
+        blog.webEditor = WebEditor(rawValue: settings.web.rawValue)
     }
 
     private func migrateLocalSettingToRemote(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
-        if blog.editor.mobile == nil {
+        if blog.mobileEditor == nil {
             let settings = GutenbergSettings()
             let defaultEditor = settings.getDefaultEditor(for: blog)
             settings.setGutenbergEnabled(defaultEditor == .gutenberg, for: blog)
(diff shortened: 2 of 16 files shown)


WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -5,8 +5,26 @@ import Foundation
 }
 
 @objc class EditorSettingsService: LocalCoreDataService {
-    @objc func syncEditorSettings(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
-        guard let api = blog.wordPressComRestApi(), let siteID = blog.dotComID?.intValue else {
+
+    let database: KeyValueDatabase
```
etoledom: Temporary addition to be able to inject a Mock KeyValueDatabase back to GutenbergSettings on Unit Tests.
This will be removed on https://github.com/wordpress-mobile/WordPress-iOS/pull/12196

etoledom:
@koke - Updated with a core data migration.
Now the global to per-site local setting is migrated via this custom core data migration.
The same logic apply:

- If an editor was previously chosen by the user, we keep that value for all their sites. Otherwise:
- WPCom/Jetpack sites will default to gutenberg.
- Self-Hosted will default to Aztec.

I also removed the hard-coded userID to separate between older and newer user accounts as discussed.

WordPress/Classes/Utility/Migrations/87-88/BlogToBlogMigration87to88.swift
```diff
@@ -0,0 +1,24 @@
...
+
+class BlogToBlogMigration87to88: NSEntityMigrationPolicy {
+    override func createRelationships(forDestination dInstance: NSManagedObject, in mapping: NSEntityMapping, manager: NSMigrationManager) throws {
+        try super.createRelationships(forDestination: dInstance, in: mapping, manager: manager)
+
+        guard let sourceBlog = manager.sourceInstances(forEntityMappingName: "BlogToBlog", destinationInstances: [dInstance]).first else {
+            return
+        }
+
+        let editor: MobileEditor
```
koke: I think migrations generally should not have external dependencies and use plain strings instead of `MobileEditor` or `GutenbergSettings`. Those objects might change in the future and we want migrations to be always the same.

etoledom: Right! That makes sense. Changed on https://github.com/wordpress-mobile/WordPress-iOS/pull/12189/commits/5c4e23be4782864d2c6927b8c4f3868c9e0f4bfc

etoledom:
Last update:
- The informative dialog should appear once per user, not per blog.
- The informative dialog should appear when the user opens gutenberg for the very first time, even if it's a new post.
- Not changed: If the informative dialog was shown on 12.9, it will not be shown again on 13.0
- Not changed: It should NOT show the informative dialog if the user manually opted-in

WordPress/Classes/Utility/Migrations/87-88/BlogToBlogMigration87to88.swift
```diff
@@ -0,0 +1,24 @@
+import CoreData
+
+class BlogToBlogMigration87to88: NSEntityMigrationPolicy {
```
koke: Should we be syncing settings for all blogs after the migration is completed? I was looking at the report card and I found surprising that the settings for each site weren't saved until I navigated to that site.

etoledom: On previous experiences, when we do a bunch of calls to this api, for some reason the setting is not stored properly on remote.

For example, from 5 blogs that we post to be set as `gutenberg`, one or two of them might be set to `aztec`. This happened on both iOS and Android.

That was the reason why we added the delay in the bulk post on previous iterations. That helped on reliability but didn't solve the problem completely.

So I'm thinking that is more reliable to lazily sync each site when it's needed.

WordPress/Classes/Models/Blog+Editor.swift
```diff
@@ -29,6 +29,9 @@ extension Blog {
     }
 
     @objc var isGutenbergEnabled: Bool {
-        return self.editor.mobile == .gutenberg
+        guard let selectedEditor = editor.mobile else {
+            return GutenbergSettings().getDefaultEditor(for: self) == .gutenberg
```
koke: This felt like it should go inside the `Editor` initialization, but then the whole `Editor` struct felt a bit superfluous so I'd like to propose a different API for the blog editor settings: 5a2600e8cc8ef0983fd3e4f91dcf9704424c6508

etoledom: Absolutely! Looks a lot simpler and more elegant.
You can make a small PR and I'll ✅ right away?

Thanks for the suggestion 🙏

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -5,8 +5,20 @@ import Foundation
...
 @objc class EditorSettingsService: LocalCoreDataService {
-    @objc func syncEditorSettings(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
-        guard let api = blog.wordPressComRestApi() else {
+    let wpcomApi: WordPressComRestApi?
+
+    @objc convenience override init(managedObjectContext: NSManagedObjectContext) {
+        self.init(managedObjectContext: managedObjectContext, wpcomApi: nil)
+    }
+
+    init(managedObjectContext: NSManagedObjectContext, wpcomApi: WordPressComRestApi? = nil) {
```
koke: I thought passing an API here that overrides the specific blog one for testing was a bit unexpected. I looked around and what I see other services doing is have a `remoteForBlog:` method that gets overriden by the tests. Something like c19ae6ef4fa749328edcf9da121c70d2f87697f1

etoledom: Nice trick! Absolutely :shipit:

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -19,11 +31,13 @@ import Foundation
         service.getEditorSettings(siteID, success: { (settings) in
             do {
                 try self.saveRemoteEditorSettings(settings, on: blog)
```
koke: I missed this in the previous PR, but `saveRemoteEditorSettings` seems like it's going to actually save the settings. Looking at the pattern in other services, I'd suggest `update(blog, remoteSettings: settings)`

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -35,11 +49,20 @@ import Foundation
         blog.webEditor = settings.web.rawValue
     }
 
+    private func migrateLocalSettingToRemote(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
+        if blog.editor.mobile == nil {
+            let settings = GutenbergSettings()
+            let defaultEditor = settings.getDefaultEditor(for: blog)
+            settings.setGutenbergEnabled(defaultEditor == .gutenberg, for: blog)
```
koke: I don't completely follow this part. It seems that if there is no mobile setting stored we would store what we decide is the default one (Gutenberg for .com/Jetpack, Aztec for self hosted). However we always set a value for this in the migration, so it should never be nil?

I'm wondering if we should make `Blog.mobileEditor` non-optional in the data model, and ensure it has the `getDefaultEditor` value when a blog instance is created.

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -19,11 +31,13 @@ import Foundation
         service.getEditorSettings(siteID, success: { (settings) in
             do {
                 try self.saveRemoteEditorSettings(settings, on: blog)
+                ContextManager.sharedInstance().save(self.managedObjectContext)
+                success()
+            } catch EditorSettingsServiceError.mobileEditorNotSet {
+                self.migrateLocalSettingToRemote(for: blog, success: success, failure: failure)
```
koke: As mentioned in Slack, we bail early from `saveRemoteEditorSettings`, and we never set the `webEditor` value in this branch, until the `syncEditorSettings` is called the next time

WordPress/Classes/Models/Blog+Editor.swift
etoledom: Cherry-picked the commit for easier integration with this PR

WordPress/Classes/Services/EditorSettingsService.swift
etoledom: Cherry-picked the commit for easier integration with this PR

WordPress/Classes/Utility/Editor/EditorFactory.swift
```diff
@@ -26,6 +26,7 @@ class EditorFactory {
         if gutenbergSettings.shouldAutoenableGutenberg(for: post) {
             gutenbergSettings.setGutenbergEnabled(true, for: post.blog)
             gutenbergSettings.postSettingsToRemote(for: post.blog)
+            gutenbergSettings.didAutoenableGutenberg()
```
koke: If I understood the code correctly, this will set the global `Key.appWideEnabled` to true, which means `wasGutenbergEnabledOnce` will always return true for all blogs and `shouldAutoenableGutenberg` will return `false` for any other blog.

etoledom: That's true. This was the behavior we agreed on.
We want to show the dialog just one per user.

I kept the per-site flag to avoid showing the dialog when the user manually opted in to gutenberg in that blog, but still show it on the next blog where gutenberg was auto-enabled.

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -68,12 +75,17 @@ class GutenbergSettings {
...
     func wasGutenbergEnabledOnce(for blog: Blog) -> Bool {
-        return blog.editor.mobile != nil
+        // If gutenberg was "globaly" enabled before, will take precedence over the new per-site flag
+        return database.object(forKey: Key.appWideEnabled) != nil || database.object(forKey: Key.enabledOnce(for: blog)) != nil
     }
 
     /// True if gutenberg should be autoenabled for the blog hosting the given post.
     func shouldAutoenableGutenberg(for post: AbstractPost) -> Bool {
-        return  post.containsGutenbergBlocks() && !wasGutenbergEnabledOnce(for: post.blog) && post.blog.editor.web == .gutenberg
+        return !wasGutenbergEnabledOnce(for: post.blog)
```
koke: I think there are some possible scenarios where gutenberg might be enabled but `wasGutenbergEnabledOnce` would be false (e.g. migration runs for a WordPress.com site with the app setting being `nil`). Should this bail early and return `false` if gutenberg is already enabled for this site?

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -89,12 +101,25 @@ class GutenbergSettings {
...
+    private func shouldUseGutenbergForNewPosts(on blog: Blog) -> Bool {
+        guard let userSelectedEditor = blog.editor.mobile else {
+            return getDefaultEditor(for: blog) == .gutenberg
+        }
+
+        return userSelectedEditor == .gutenberg
+    }
+
+    func getDefaultEditor(for blog: Blog) -> MobileEditor {
+        // Default to gutenberg on WPCom/Jetpack sites
```
koke: I'm not sure I know the reason for this, but it'd be good to expand a bit on the comment for future reference

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -89,12 +101,25 @@ class GutenbergSettings {
...
         if post.isContentEmpty() {
-            return blog.editor.mobile == .gutenberg && blog.editor.web == .gutenberg
+            return shouldUseGutenbergForNewPosts(on: blog)
         } else {
             // It's an existing post
             return post.containsGutenbergBlocks()
         }
     }
+
+    private func shouldUseGutenbergForNewPosts(on blog: Blog) -> Bool {
```
koke: I missed this in 3e8bd9daaa9354ce4c7ddc9d1e33b30c9fa90f04 but this function does the exact same thing as `Blog.isGutenbergEnabled` right?

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: I think this is part of what it's explained here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12189#issuecomment-515111020

In the scenario you mention, the user got gutenberg auto-enabled via the migration, so we do want to show the informative dialog in this case, even on a new post

Probably with the many changes on this behavior, the naming got confusing 🤔

WordPress/Classes/Services/EditorSettingsService.swift
etoledom: Pushed https://github.com/wordpress-mobile/WordPress-iOS/pull/12189/commits/5fef12b96b156fa4fc0103c08e430a4bd9ec7cff

WordPress/Classes/Services/EditorSettingsService.swift
etoledom: Pushed https://github.com/wordpress-mobile/WordPress-iOS/pull/12189/commits/5fef12b96b156fa4fc0103c08e430a4bd9ec7cff

WordPress/Classes/Services/EditorSettingsService.swift
etoledom: > However we always set a value for this in the migration, so it should never be nil?

It will be nil on new installs, right? 🤔 

> I'm wondering if we should make Blog.mobileEditor non-optional in the data model, and ensure it has the getDefaultEditor value when a blog instance is created.

Sounds like a good alternative. I don't think I will have time to check it out today though

WordPress/WordPressTest/CoreDataMigrationTests.m
```diff
@@ -967,6 +967,100 @@ - (void)testMigrate74to75
...
+
+    // WPCom/Jetpack sites should migrate to Gutenberg
+
+    NSManagedObject *wpcomBlog = [results firstObject];
+    XCTAssertEqualObjects([wpcomBlog valueForKey:@"mobileEditor"], @"gutenberg");
+
+    // SelfHosted (non-jetpack) should migrate to Aztec
+
+    NSManagedObject *selfHostedBlog = [results lastObject];
+    XCTAssertEqualObjects([selfHostedBlog valueForKey:@"mobileEditor"], @"aztec");
```
koke: After running the app on the simulator, this test fails for me because the migration is setting the value according to `kUserDefaultsGutenbergEditorEnabled`

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: True!
I like `blog.isGutenbergEnabled` :)

`shouldUseGutenbergForNewPosts` was created/renamed to follow up with Android.  `blog.isGutenbergEnabled` feels more Swifty ✨

WordPress/WordPressTest/EditorSettingsServiceTests.swift
```diff
@@ -0,0 +1,122 @@
...
+
+        waitForExpectations(timeout: 0.1) { (error) in
+            // The default value should be now on local and remote
+            XCTAssertEqual(blog.mobileEditor, .gutenberg)
+        }
+    }
+}
+
+extension EditorSettingsServiceTest {
+    func blogWith() -> Blog {
```
koke: That's an odd name, maybe this used to have arguments?

WordPress/WordPressTest/CoreDataMigrationTests.m
etoledom: Right. We need user defaults for the migration.

The first thing that comes into mind is to use user default from the Environment introduced in https://github.com/wordpress-mobile/WordPress-iOS/pull/12196

Do you have other options?

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: https://github.com/wordpress-mobile/WordPress-iOS/pull/12189/commits/4fcabaf6beae7db9077647b36c8c2a10b9dc5564

WordPress/WordPressTest/EditorSettingsServiceTests.swift
etoledom: Wops, yes. Refactor fail!

WordPress/Classes/Utility/Migrations/87-88/BlogToBlogMigration87to88.swift
koke: That makes me curious, but it's probably not worth the effort of fully investigating, the current approach seems like it should be fine

WordPress/Classes/Utility/Editor/EditorFactory.swift
koke: > We want to show the dialog just one per user

I can agree with that, but just to double check, this is not just about presenting the dialog but automatically enabling gutenberg when you open a gutenberg post. In previous iterations this was done for all sites when the dialog was shown. Now it only enables it for the current site (which is better), but any other site where the user opens a Gutenberg post will not get gutenberg auto-enabled for new posts. Is this the behaviour we want?

WordPress/WordPressTest/EditorSettingsServiceTests.swift
etoledom: https://github.com/wordpress-mobile/WordPress-iOS/pull/12189/commits/4263c317e7cfd9d580bc98052aadbec37cd378d6

WordPress/WordPressTest/CoreDataMigrationTests.m
koke: That's probably the sensible route to take. I thought of maybe trying to inject the `database` into the `userInfo` of `NSMigrationManager`, but that would require too many changes to `ALIterativeMigrator` and I don't like the risk of breaking things there.

Maybe `OCMock` can replace the standard user defaults, since the tests are Obj-C?

koke:
> The default editor should be Gutenberg for all newer users
> The default editor should be Aztec for all older users
> Note: A newer user is defined as having a userID > 149287441

I think it was decided not to do the check for user ID and just default to Gutenberg for new installs. The PR was updated in 032c14f0d420f740e0774bbf3b68993a8c7cc5e8 but the description was not, so leaving a not here for completeness

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController+InformativeDialog.swift
```diff
@@ -5,7 +5,7 @@ extension GutenbergViewController {
 
     enum InfoDialog {
         static let message = NSLocalizedString(
-            "This post uses the block editor, which is the default editor for new posts. To enable the classic editor, go to Me > App Settings.",
+            "This post uses the block editor, which is the default editor for new posts. To enable the classic editor, go to Site Settings.",
```
hypest: Let's update the copy to match the latest agreed over at the Android side PR: `You're currently using the block editor to create new [posts/pages]. Want to use the classic editor by default? Head to Site Settings.` https://github.com/wordpress-mobile/WordPress-Android/pull/10254#issuecomment-515334567

etoledom: Updated!

etoledom:
Checked https://github.com/wordpress-mobile/WordPress-iOS/pull/12189/commits/eebe0650f37dacb265793bd221910159291062bc and it looks good! 👍

hypest:
Tried:

## WP.com/Jetpack sites
❌ Create a new WP.com site on mobile and check Site Settings → should be Gutenberg: it's Aztec (setting, backend)
❌ Create a new WP.com site on mobile and start a new post → should be Gutenberg: it's Aztec

✅ Clean app, clean backend, start new post → should be Aztec
✅ Clean app, "Aztec" in backend, start new post → should be Aztec
⚠️ Clean app, "Gutenberg" in backend, start new post → should be Gutenberg: it's GB but with the popup (I wasn't expecting the popup)
✅ Clean  app, clean backend, open Gutenberg post → auto-enables Gutenberg
✅ Clean  app, clean backend, open Gutenberg post (auto-enables GB), start new post → should be Gutenberg
✅ Clean  app, clean backend, open Gutenberg post (auto-enables GB), open a GB post again→ Gutenberg but no popup
⏳ Clean app, clean backend, open Gutenberg post (auto-enables GB), new post in other site → should be Aztec
⏳ Clean app, clean backend, open Gutenberg post (auto-enables GB), open GB post in other site → auto-enables Gutenberg
⏳ Toggling the setting in a site doesn't toggle it in others
⏳ Migrate from v12.9 with GB ON, start new post → should be Gutenberg in all sites: `seems to be Aztec on current site and Gutenberg on other site`
⏳ Migrate from v12.9 with GB OFF and been auto-enabled, start new post → should be Aztec in all sites
⏳ Migrate from v12.9 with GB OFF and never auto-enabled, start new post → should be Aztec in all sites
⏳ Migrate from v12.9 with GB OFF and never auto-enabled, open a GB post (should auto-enable GB), change to other site and start new post → should be Aztec
⏳ Migrate from v12.9 with GB OFF and never auto-enabled, open a GB post (should auto-enable GB), change to other site and open GB post → auto-enables Gutenberg on the site

## Selfhosted sites

⏳ Self-hosted site added, start new post → should be Aztec
⏳ Self-hosted site added, open a GB post → auto-enables Gutenberg
⏳ Self-hosted site added, open a GB post (should auto-enable GB), start new post → should be Gutenberg
⏳ Self-hosted site added, open a GB post (should auto-enable GB), clean app data, re-add self-hosted site, start new post → should be Aztec
⏳ Self-hosted site added, open a GB post (should auto-enable GB), clean app data, re-add self-hosted site, open a GB post → auto-enables Gutenberg
⏳ Migrate a selfhosted from v12.9 with GB OFF and been auto-enabled, start new post → should be Aztec in all selfhosted sites
⏳ Migrate from v12.9 with GB OFF and never auto-enabled, start new post → should be Aztec in all selfhosted sites
⏳ Migrate from v12.9 with GB OFF and never auto-enabled, open a GB post (should auto-enable GB), change to other site and start new post → should be Aztec
⏳ Migrate from v12.9 with GB OFF and never auto-enabled, open a GB post (should auto-enable GB), change to other site and open GB post → auto-enables Gutenberg on the site

hypest:
Added the `and show popup` detail to the create-site-and-start-post case.

etoledom:
Thank you @koke for all your effort and help! 🙏

PR #12192, opened 2019-07-23
Update Fastlane

jkmassel:
Patches the minimagick advisory
Should also provide better Xcode 11 support

**To Test:**
Do an internal release from `develop`. Don't worry about version numbers, just delete it from Hockey when you're done. This should test the majority of the WPiOS fastlane tooling.


jkmassel:
If I `rm -rf vendor/ && bundle install` in the project, I can't run do an internal release because the secret store checks fail currently.

If we're good to merge this as-is, I'm fine with that, but I wanted to get that fixed before submitting this for testing by other folks.

Issue #171, opened 2013-10-09
Post Media: Deleting the only media looses tab title.


aerych:
To reproduce.

Edit an existing post with on media item.
Tap on the media tab.
Tap on the media item and delete it.
The media tab is empty and has lost its title and is empty.

Since there is no media to view a better experience might be to return the user to the post editor. 

iOS 6 / iPhone 5

PR #12195, opened 2019-07-24
[Jetpack] Jetpack Remote Install Error log

### LINKED ISSUE #12138: Stats: Better Error Messaging for Jetpack Remote Install Failure

danielebogo:
Refs. #12138 

This PR adds more log info about the JRI error received.

New log example:
```
Jetpack Remote Install error for site SITE_URL – The request timed out. (-1001: unknown)
```

I will update the Podfile as soon [#171](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/171) is merged.

## To test:
• Try to install Jetpack on a website that might fail (Use a slow connection profile to generate a timeout)

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -43,9 +43,9 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.3.0-beta.1'
+#    pod 'WordPressKit', '~> 4.3.0-beta.1'
     #pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => 'issue/stats_timezone'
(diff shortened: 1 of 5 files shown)


danielebogo:
Thanks @frosty

PR #12199, opened 2019-07-24
Stats: adjust 'relativeStringInPast' logic

### LINKED ISSUE #12172: Stats LPS summary: single periods have plural descriptors
In the Latest Post Summary summary text, when the period is 1, the time period is plural (hours, days, weeks, months, years). 

lps_summary

It would be nice if they were singular where applicable.

ScoutHarris:
Fixes #12172 

This modifies the logic used to determine `relativeStringInPast` (used by Latest Post Summary and Followers).

- When a time falls into the `AndaHalf` categories, the time is now rounded accordingly.
  - Ex: previously 1.75 days would display as "1 days". Now it displays as "2 days".
- 1 day and 1 hour are now displayed as "a day" and "an hour". 
  - Previously, they would display as "1 days" and "1 hours".
- Updated month breakpoints to be more inline with Calypso.

To test:
- On sites with varying periods since the Latest Post Summary was published, go to Insights LPS.
- Verify the periods are as expected.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Helpers/StatsDataHelper.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Helpers/StatsDataHelper.swift
@@ -187,57 +187,57 @@ extension Date {
         let hours = components.hour ?? 0
         let minutes = components.minute ?? 0
 
-        if days >= DateFormattingBreakpoints.aboutYearAndAHalf.rawValue {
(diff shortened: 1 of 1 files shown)


ScoutHarris:
> what do you think if DateFormattingBreakpoints contains static vars rather than cases?

I like it, and I did it. 😄 Thanks @danielebogo !

PR #12214, opened 2019-07-26
Don't repeatedly show failed upload notices after auto retry

### LINKED ISSUE #12213: Repeating notices about failed uploads
We've gotten a report from number of users (2204495-zen, https://ios.forums.wordpress.org/topic/strange-behavior-with-uploading-pics#post-15436) about them repeatedly seeing notices in the format of "1 post, 1 file not uploaded", which are probably connected to the recent changes to how we handle file uploads.

jklausa:
Fixes #12213.

Okay, so fixing this one was a doozy. What helped me get on the right path was a log submitted by a user in 2204495-zen. 

A relevant excerpt is:

```
019-07-25 10:22:10:808 🔵 Tracked: media_service_upload_failed <error_code: 2, error_description: Error Domain=Alamofire.AFError Code=2 "The system returned an error while checking the provided URL for reachability. 
URL: file:///var/mobile/Containers/Data/Application/8E539821-37D0-47C1-A310-791005ADD3D2/Documents/Media/IMG_5144.JPG 
Error: Error Domain=NSCocoaErrorDomain Code=260 "The file “IMG_5144.JPG” couldn’t be opened because there is no such file." UserInfo={NSURL=file:///var/mobile/Containers/Data/Application/8E539821-37D0-47C1-A310-791005ADD3D2/Documents/Media/IMG_5144.JPG, NSFilePath=/var/mobile/Containers/Data/Application/8E539821-37D0-47C1-A310-791005ADD3D2/Documents/Media/IMG_5144.JPG, NSUnderlyingError=0x283ed5410 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}", error_domain: Alamofire.AFError> 
```

It looks for whatever reason, sometimes we have a `Media` object that wasn't uploaded properly, that has it's underlying backing file disappear from the disk.

That's not, what we usually call, "good".

This "fixes" it in two ways:

1) by silencing all notices created from errors caused by auto-retries of Media. the user hasn't explicitely initiated that action, so we shouldn't bother it with it. 
// ^- this is VERY MUCH open for discussion if it's the right thing to do in the long run. cc @wordpress-mobile/ravenclaw

2) by adding a error handler to check if an upload failed specifically because of the underlying file being gone, that then sends a log to sentry and deletes the `Media` object if that happens, since there isn't really a meaningful way for us to recover from that.

___

This is, like everything to do with this issue, a nightmare to test. The "simplest" way I found to trigger the underlying bug is so far this:

To test (*must* be done on Simulator):
1. Enable Network Link Conditioner on your Mac, with a `Very Bad Network` or `100% Packet Loss` Presets
1. Build latest `develop`
2. Go to Media Library
3. Add a new picture from your device (using the "Choose from My Device") option
4. Verify it fails to upload
5. Locate the file on disk (I usually just do `po NSHomeDirectory()` and `cd` to `Documents/Media` in iTerm) and delete it.
5. Turn off Network Link Conditioner
6. Relaunch the app
7. Verify that the "1 post, 1 file not uploaded" Notice shows up.

Now that you've been able to _reproduce_ the issue,
1. build the branch this PR is pointing to
2. launch the app
3. verify that the notice no longer shows up
4. go to media library and verify that the "broken" media object is no longer there
5. cry inside a bit how about how complicated media handling inherently is
6. **REMEMBER TO TURN OFF YOUR NETWORK LINK CONDITIONER**

### DIFF
--- a/WordPress/Classes/Services/MediaCoordinator.swift
+++ b/WordPress/Classes/Services/MediaCoordinator.swift
@@ -31,13 +31,6 @@ class MediaCoordinator: NSObject {
         return coordinator
     }()
 
-    /// Tracks uploads that don't belong to a specific post _and_ are auto-retries of failed uploads
-    private lazy var autoRetryProgressCoordinator: MediaProgressCoordinator = {
-        let coordinator = MediaProgressCoordinator()
-        coordinator.delegate = self
-        return coordinator
-    }()
-
     /// Tracks uploads of media for specific posts
     private var postMediaProgressCoordinators = [AbstractPost: MediaProgressCoordinator]()
 
@@ -166,10 +159,8 @@ class MediaCoordinator: NSObject {
     /// Retry the upload of a media object that previously has failed.
     ///
     /// - Parameter media: the media object to retry the upload
-    /// - Parameter isAutomaticRetry: indicates whether an upload is a manual action taken by the user
-    /// or an automagical one done via the `Uploader` protocol.
     ///
-    func retryMedia(_ media: Media, analyticsInfo: MediaAnalyticsInfo? = nil, isAutomaticRetry: Bool = false) {
+    func retryMedia(_ media: Media, analyticsInfo: MediaAnalyticsInfo? = nil) {
         guard media.remoteStatus == .failed else {
             DDLogError("Can't retry Media upload that hasn't failed. \(String(describing: media))")
             return
@@ -177,12 +168,7 @@ class MediaCoordinator: NSObject {
 
         trackRetryUploadOf(media, analyticsInfo: analyticsInfo)
 
-        let coordinator: MediaProgressCoordinator
-        if isAutomaticRetry {
-            coordinator = autoRetryProgressCoordinator
-        } else {
-            coordinator = self.coordinator(for: media)
-        }
+        let coordinator = self.coordinator(for: media)
 
         coordinator.track(numberOfItems: 1)
         let uploadProgress = uploadMedia(media)
@@ -607,7 +593,13 @@ extension MediaCoordinator: MediaProgressCoordinatorDelegate {
     func mediaProgressCoordinatorDidFinishUpload(_ mediaProgressCoordinator: MediaProgressCoordinator) {
         // We only want to show an upload notice for uploads initiated within
         // the media library, or if there's been a failure.
-        if mediaProgressCoordinator == mediaLibraryProgressCoordinator || mediaProgressCoordinator.hasFailedMedia {
+        // If the errors are causes by a missing file, we want to ignore that too.
+
+        let mediaErrorsAreMissingFilesErrors = mediaProgressCoordinator.failedMedia.allSatisfy { $0.hasMissingFileError }
+
+        if !mediaErrorsAreMissingFilesErrors,
+            mediaProgressCoordinator == mediaLibraryProgressCoordinator || mediaProgressCoordinator.hasFailedMedia {
+
             let model = MediaProgressCoordinatorNoticeViewModel(mediaProgressCoordinator: mediaProgressCoordinator)
             if let notice = model?.notice {
                 ActionDispatcher.dispatch(NoticeAction.post(notice))
@@ -633,7 +625,7 @@ extension MediaCoordinator: Uploader {
 
             media.forEach() {
                 self.addObserverForDeletedFiles(for: $0)
-                self.retryMedia($0, isAutomaticRetry: true)
+                self.retryMedia($0)
             }
         }
     }
@@ -646,23 +638,15 @@ extension MediaCoordinator {
     // We want to collect more data about that, so we're going to log that info to Sentry,
     // and also delete the `Media` object, since there isn't really a reasonable way to recover from that failure.
     func addObserverForDeletedFiles(for media: Media) {
-        addObserver({ (media, state) in
-            guard
-                case .failed(let error) = state,
-                let afError = error as? AFError,
-                case .multipartEncodingFailed(let encodingFailure) = afError else {
-                    return
-            }
-
-            switch encodingFailure {
-            case .bodyPartFileNotReachableWithError,
-                 .bodyPartFileNotReachable:
-                CrashLogging.logMessage("Deleting a media object that's failed to upload because of a missing local file.",
-                                        properties: ["underlyingError": afError])
-                self.cancelUploadAndDeleteMedia(media)
-            default:
+        addObserver({ [weak self] (media, _) in
+            guard let mediaError = media.error,
+                media.hasMissingFileError else {
                 return
             }
+
+            CrashLogging.logError(mediaError,
+                                  userInfo: ["description": "Deleting a media object that's failed to upload because of a missing local file."])
+            self?.cancelUploadAndDeleteMedia(media)
 
         }, for: media)
     }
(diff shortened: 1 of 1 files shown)


maxme:
We wrote a fix in wpandroid that limits auto-upload to 10 auto-retries. We never hide upload errors (automatically started or manually started). Some of these errors might only be solved by some user action ([1]). Also note we can't rely on errors on our side to decide if a user action is required, so the user will also see other errors as well (errors produced by bad network connectivity, timeouts, etc.)

[1] One simple case to test is to have a site with no remaining media quota and try to upload a post with some media on there (I can add users to such a test site if you want to test this case). It will fail, we can retry a thousand times (manually or not), it will always fail, unless the user do something (remove existing media or upgrade his plan).

jklausa:
@maxme We should probably have a bigger discussion about it elsewhere, but I think there’s a difference between actionable errors that we should surface to the user, even if encountered during auto-upload, and network failures - and for sure there’s a space for more nuanced implementation here. 

I agree that we should surface errors like you mentioned (no storage space on remote) to the users. 

However, I think the most common error case by far (though I’ll readily admit that I don’t have actual numbers on that, and this is mostly based on my hunch!) will be due to flaky/no internet. There’s very little users can actively do to change that in most cases. I think showing notices about bad internet and failed uploads _when users haven’t initiated any uploads themselves_ is of limited utility at best, and actively confusing at worst — which the existence of the ticket this PR is responding to would seem to confirm. 

I’m super open to discussing this more next week :)

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -156,16 +166,24 @@ class MediaCoordinator: NSObject {
     /// Retry the upload of a media object that previously has failed.
     ///
     /// - Parameter media: the media object to retry the upload
+    /// - Parameter isAutomaticRetry: indicates whether an upload is a manual action taken by the user
+    /// or an automagical one done via the `Uploader` protocol.
     ///
-    func retryMedia(_ media: Media, analyticsInfo: MediaAnalyticsInfo? = nil) {
+    func retryMedia(_ media: Media, analyticsInfo: MediaAnalyticsInfo? = nil, isAutomaticRetry: Bool = false) {
```
frosty: Nitpick, but could possibly shrink this to 'isAutomatic`

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -1,6 +1,9 @@
 import Foundation
 import WordPressFlux
 
+import class AutomatticTracks.CrashLogging
```
frosty: Huh, I don't recall seeing us use this syntax anywhere else – I didn't even know it existed. Is there any advantage doing this over just importing `AutomatticTracks` and `Alamofire`?

jklausa: There are 2 advantages that I see:

1) It only makes the imported symbols available in the MediaCoordinator: I don't need this class to have access to the entirety of Alamofire, it just needs to know about AFError. Same with CrashLogging — I don't need the entirety of Tracks — all actual tracking needs should probably be handled by `WPAppAnalytics`
2) It indicates intent for the next person to read this code — nobody will ask "huh, why is this random class pulling in Alamofire for?" if it's clear that we just need that one specific object.


If you think the being unfamiliar syntax negates those, I'm happy to remove it :)

frosty: Thanks for the explanation – I think it probably makes sense to keep it in that case.

jkmassel:
I'm moving this PR's milestone forward to `13.1`, which is the next unfrozen release branch. If we need it released sooner, please merge it into `release/12.9` or `release/13.0` and DM me and I'll be happy to include it in a beta or hotfix, as required.

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -28,6 +31,13 @@ class MediaCoordinator: NSObject {
         return coordinator
     }()
 
+    /// Tracks uploads that don't belong to a specific post _and_ are auto-retries of failed uploads
+    private lazy var autoRetryProgressCoordinator: MediaProgressCoordinator = {
```
diegoreymendez: Could you provide some information on why this change is necessary / desirable?

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -611,11 +631,43 @@ extension MediaCoordinator: Uploader {
...
+                case .failed(let error) = state,
+                let afError = error as? AFError,
+                case .multipartEncodingFailed(let encodingFailure) = afError else {
+                    return
+            }
+
+            switch encodingFailure {
+            case .bodyPartFileNotReachableWithError,
+                 .bodyPartFileNotReachable:
+                CrashLogging.logMessage("Deleting a media object that's failed to upload because of a missing local file.",
```
diegoreymendez: Considering this is an abnormal (as in "shouldn't happen") condition, why not log this with `CrashLogging.logError(...)`.

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -156,16 +166,24 @@ class MediaCoordinator: NSObject {
...
             DDLogError("Can't retry Media upload that hasn't failed. \(String(describing: media))")
             return
         }
 
         trackRetryUploadOf(media, analyticsInfo: analyticsInfo)
 
-        let coordinator = self.coordinator(for: media)
+        let coordinator: MediaProgressCoordinator
+        if isAutomaticRetry {
+            coordinator = autoRetryProgressCoordinator
```
diegoreymendez: Two questions:

1. Is this change needed?  Why?
2. The original code went through a cache of existing media coordinators.  The new code doesn't - so there's a risk we may be changing logic that shouldn't be changed.

WordPress/Classes/Services/MediaCoordinator.swift
diegoreymendez: Agree on this change.

jklausa:
I've simplified and updated this patch, according to notes from Diego. @yaelirub would you be so kind and could take a peek at this one?

diegoreymendez:
Can we retarget or recreate this PR as a @jklausa?  I think this issue is becoming quite annoying for our users.

jklausa:
uh that extremely did not go well.

@diegoreymendez after this is approved i'll post a separate PR to cherry-pick those changes to 12.9 (or 13.0?) and will work with platforms to get it released

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -587,7 +593,13 @@ extension MediaCoordinator: MediaProgressCoordinatorDelegate {
     func mediaProgressCoordinatorDidFinishUpload(_ mediaProgressCoordinator: MediaProgressCoordinator) {
         // We only want to show an upload notice for uploads initiated within
         // the media library, or if there's been a failure.
-        if mediaProgressCoordinator == mediaLibraryProgressCoordinator || mediaProgressCoordinator.hasFailedMedia {
+        // If the errors are causes by a missing file, we want to ignore that too.
+
+        let mediaErrorsAreMissingFilesErrors = mediaProgressCoordinator.failedMedia.allSatisfy { $0.hasMissingFileError }
+
+        if !mediaErrorsAreMissingFilesErrors,
+            mediaProgressCoordinator == mediaLibraryProgressCoordinator || mediaProgressCoordinator.hasFailedMedia {
```
yaelirub: Why are we not checking hasFailedMedia at the beginning of this function and only if it does handle accordingly?

jklausa: This method is also responsible for presenting success notices.

yaelirub: Yep. so I was thinking `if mediaProgressCoordinator.hasFailedMedia { (handle failure) }  else { handle success }`

PR #12215, opened 2019-07-26
Limits auto-uploading temporarily.

diegoreymendez:
In this PR we're limiting auto-uploading of posts temporarily to new drafts or existing drafts that haven't changed status.

## To test:

Make sure you comment these lines out, so that we can have a post in CoreData that failed to publish without its status being changed to draft.

https://github.com/wordpress-mobile/WordPress-iOS/blob/c8403092a99194637e7917c0f24ca16d27ba650b/WordPress/Classes/Services/PostService.m#L263-L266

Make sure you test this using a device, as the simulator seems to have some issues with reachability.

1. Go offline
2. Create a new post
3. Publish the post
4. Create a new draft
5. Save the draft
6. Make sure both are marked as failed.
7. Reconnect the device and make sure that the published post isn't uploaded, but the draft is.

## Update release notes:

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -256,7 +256,7 @@ extension PostCoordinator: Uploader {
             }
 
             posts.forEach() { post in
-                let shouldRetry = post.status == .draft && (!post.hasRemote() || post.original?.status == .draft)
+                let shouldRetry = post.status == .draft && !post.hasRemote()
 
                 if shouldRetry {
                     self.retrySave(of: post)
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -255,7 +255,13 @@ extension PostCoordinator: Uploader {
                 return
             }
 
-            posts.forEach() { self.retrySave(of: $0 ) }
+            posts.forEach() { post in
+                let shouldRetry = post.status == .draft && (!post.hasRemote() || post.original?.status == .draft)
+
+                if shouldRetry {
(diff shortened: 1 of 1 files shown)


WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -255,7 +255,13 @@ extension PostCoordinator: Uploader {
                 return
             }
 
-            posts.forEach() { self.retrySave(of: $0 ) }
+            posts.forEach() { post in
+                let shouldRetry = post.status == .draft && (!post.hasRemote() || post.original?.status == .draft)
```
shiki: @diegoreymendez This works great. 

@maxme gave a good suggestion that we can make this safer. Android only has this condition:

```swift 
let shouldRetry = post.status == .draft && !post.hasRemote()
```

The `post.original?.status == .draft` can cause a published post to be reverted back to draft. Here's an example scenario where I was able to reproduce this:

1. iOS: Create a `DRAFT` post and upload it.
2. iOS: Go offline.
3. Web: Publish the post. 
4. iOS: While still offline, edit the post and save it. Do not publish. 
5. iOS: Go online and let the auto-uploader finish. 

When the upload is done, the post is reverted back to a draft.

diegoreymendez: For context: as we discussed privately this is going to always be an issue unless we get API support.

But if we add a timeout to these auto-uploads, they're going to be much safer to allow.  So we'll scale back this solution to what @shiki proposed until we can get more safeguards in place.

The decision is:

1. Upload local drafts only in the PR.
2. Discuss the timer and other possible safeguards as soon as possible.

Issue #10097, opened 2018-09-04
Rich Notifications : Add custom actions to authentication flow notifications


ghost:
When executing the 2FA flow via the WP app, the user is presented with a basic push notification. 

The notification displayed on the phone provides no interactivity - the user must tap the notification to open the app. Upon doing so, the user is present with a static [alert](https://developer.apple.com/design/human-interface-guidelines/ios/views/alerts/) that [offers two options](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/PushAuthenticationManager.swift#L137-L150): **Approve** or **Ignore**. 

**Note**: _a similar request is noted in #10023, with an emphasis on Apple Watch behavior._

aerych:
Closed with #15253

PR #12216, opened 2019-07-26
Try/approve 2fa from notification

aerych:
Refs: #10097

I'm tinkering with enabling a Push Notification action to approve a 2fa challenge from the notification interface.  Would love an early peek from @jleandroperez or @frosty who are more familiar with this feature for a sanity check on the direction I'm heading in.

To test:

Update release notes:

### DIFF
--- a/WordPress/Classes/Utility/InteractiveNotificationsManager.swift
+++ b/WordPress/Classes/Utility/InteractiveNotificationsManager.swift
@@ -80,6 +80,10 @@ final class InteractiveNotificationsManager: NSObject {
             return handleLocalNotificationAction(with: identifier, category: category, userInfo: userInfo, responseText: responseText)
         }
 
+        if NoteActionDefinition.twoFactorAuthApprove == NoteActionDefinition(rawValue: identifier) {
+            return approveAuthChallenge(userInfo)
+        }
+
         guard AccountHelper.isDotcomAvailable(),
             let noteID = userInfo.object(forKey: "note_id") as? NSNumber,
             let siteID = userInfo.object(forKey: "blog_id") as? NSNumber,
@@ -255,6 +259,10 @@ private extension InteractiveNotificationsManager {
         })
     }
 
+    func approveAuthChallenge(_ userInfo: NSDictionary) -> Bool {
+        return PushNotificationsManager.shared.handleAuthenticationNotification(userInfo, completionHandler: nil)
+    }
+
 
     /// Returns a collection of *UNNotificationCategory* instances, for each one of the
     /// supported NoteCategoryDefinition enum case's.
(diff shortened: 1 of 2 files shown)


jleandroperez:
Everything looks great @aerych !! That's **exactly** the way things were meant to be extended sir.

I'm subscribing to this PR, please do let me know if I can help you anyhow!!

WordPress/Classes/Utility/InteractiveNotificationsManager.swift
```diff
@@ -329,7 +340,7 @@ private extension InteractiveNotificationsManager {
                 options: [])
         }
 
-        static var allDefinitions = [commentApprove, commentLike, commentReply, commentReplyWithLike, mediaUploadSuccess, mediaUploadFailure, postUploadSuccess, postUploadFailure, shareUploadSuccess, shareUploadFailure]
+        static var allDefinitions = [commentApprove, commentLike, commentReply, commentReplyWithLike, mediaUploadSuccess, mediaUploadFailure, postUploadSuccess, postUploadFailure, shareUploadSuccess, shareUploadFailure, .twoFactorAuthChallenge]
```
frosty: The other categories in here don't use a `.` before the category name – does it work without?

WordPress/Classes/Utility/InteractiveNotificationsManager.swift
```diff
@@ -365,6 +377,8 @@ private extension InteractiveNotificationsManager {
                 return NSLocalizedString("View", comment: "Opens the post epilogue screen to allow sharing / viewing of a post.")
             case .shareEditPost:
                 return NSLocalizedString("Edit Post", comment: "Opens the editor to edit an existing post.")
+            case .twoFactorAuthApprove:
+                return NSLocalizedString("Approve", comment: "Approve login attempt (verb)")
```
frosty: Should we clarify the comment a little more? "Approve two-factor login attempt (verb)"?

WordPress/Classes/Utility/InteractiveNotificationsManager.swift
```diff
@@ -432,7 +446,7 @@ private extension InteractiveNotificationsManager {
             }
         }
 
-        static var allDefinitions = [commentApprove, commentLike, commentReply, mediaWritePost, mediaRetry, postRetry, postView, shareEditPost]
+        static var allDefinitions = [commentApprove, commentLike, commentReply, mediaWritePost, mediaRetry, postRetry, postView, shareEditPost, .twoFactorAuthApprove]
```
frosty: Same comment as above regarding `.`

WordPress/Classes/Utility/PushAuthenticationManager.swift
```diff
@@ -63,14 +63,17 @@ class PushAuthenticationManager {
             return
         }
 
-        showLoginVerificationAlert(message) { approved in
```
frosty: I know this PR isn't final, but I think we'll need to leave code in here to show an alert if the app is currently open when the notification arrives? We might need to split this method into two parts – one we can reuse in the 'backgrounded app' case, and one which calls into that but also does the UI part in the case where the app is in the foreground?

aerych: Good point about about the need for the alert if the app is open.  I'll make sure that functionality is retained.

aerych:
Thanks @frosty and  @jleandroperez !! I appreciate the early peek.

PR #12220, opened 2019-07-29
Add + icon back to Posts list

### LINKED ISSUE #12219: [Post List] Add "+" icon to navigation bar
The new post list uses a view changer in the upper right-hand corner of the screen which has replaced the "+" button / the action to create a new post.

Let's re-include the "+" button to the top right corner:

Screenshot 2019-07-29 at 11 28 04

frosty:
Fixes #12219. This PR reintroduces the + button to the Posts list. I also switched both Pages and Posts to use a gridicon icon for consistency, and just use the system spacing instead of doing our own thing.

Screenshot 2019-07-29 at 14 04 13

**To test:**

Update release notes:

I haven't updated the release notes yet. If we can get this into 12.9 too then a release note won't be necessary as nobody will ever notice the button was gone.


SylvesterWilmott:
Looks good @frosty thanks!

PR #12222, opened 2019-07-29
Fix/posts add button

frosty:
This PR is the same as #12220, but targeted at 12.9. Same steps to test, and the code is identical.


frosty:
cc @jkmassel

frosty:
Thanks @danielebogo!

PR #12231, opened 2019-07-30
UITests: Fix editor tests

etoledom:
This PR fixes the broken UITests by moving the switch editor logic to Site Settings.

To test:
- Run UI tests locally
- Check that UI tests pass


etoledom:
I decided to keep the `AppSettingsScreen` screen in case is needed in the future, even though is unused right now.

Let me know if you think is better to just remove the whole file 👍

WordPress/WordPressUITests/Screens/SiteSettingsScreen.swift
```diff
@@ -0,0 +1,48 @@
...
+        case .off:
+            if isBlockEditorEnabled() {
+                blockEditorToggle.tap()
+            }
+        }
+        return self
+    }
+
+    func goBackToMySite() -> MySiteScreen {
+        navBackButton.tap()
```
rachelmcr: Before tapping the back button, we should check if we're on iPhone (and only go back in that case). On iPad, we can have both MySiteScreen and SiteSettingsScreen visible at the same time, and going back takes us to the sites list instead.

jtreanor:
> There are some mocked network requests that need to be updated/added given the Gutenberg changes. However, it looks like @jtreanor has suggested fixes for those in wordpress-mobile/WordPress-Android#10299 — James, will you also be making those upstream changes in WordPressMocks? That would resolve this for the iOS tests as well.

Yes, once that PR is merged we can update the `WordPressMocks` pod too 😄

etoledom:
Thank you @rachelmcr for the review!
I made the requested changes.

Tested on iPad but for reason is crashing locally in the `testSelfHostedUsernamePasswordLoginLogout` test. 

Again not sure if is a local problem of mine or something else is happening.

rachelmcr:
> Tested on iPad but for reason is crashing locally in the testSelfHostedUsernamePasswordLoginLogout test.

Hm, that's odd. I can't reproduce that crash — tests pass on iPad for me now with no issues — and since it shouldn't be related to the changes in the PR let's go ahead and merge this. If we start to see crashes for that test on CI (or get reports from real users) we can investigate that separately.

etoledom:
Thank you!

Issue #9705, opened 2018-07-05
Add Ability to Change Username


SiobhyB:
We receive questions around changing usernames frequently in support.

At the moment, we have to guide users away from the app and to a web browser to change their username.

It'd be convenient if users were able to edit their usernames directly from the app itself, and it may also reduce the volume of requests we receive.

Further discussion can be found at p7cLQ7-uk-p2.

SiobhyB:
Android issue: https://github.com/wordpress-mobile/WordPress-Android/issues/8002

SiobhyB:
We chatted about this in p1562572814187500-slack-mobile-support and confirmed it's still something that comes up in support a lot.

thehenrybyrd:
I run into requests like this _all the time_. I looked up tickets about this based on the snippet I use + assigned to me, and found 9 from this week alone.
While the workflow of 'send a snippet' doesn't make this feel like a "bug" or "problem", it's clearly a common request from users. Would be great to let them stay in the app for this! ❤

designsimply:
@SiobhyB @thehenrybyrd can you document these requests as you find them so I can use that info as part of prioritizing issues? 

The format I find most helpful is to include the shorthand (internal reference) and a small quote from the user mentioning the pain point or (especially helpful) **why** they want the change in one note per comment (this helps with searching by number of comments later). In this case specifically, I think it would also be interesting to know if the users most commonly requesting to be able to change their username from the WP apps are brand new users who have made a mistake in spelling or just changed their mind about a username or if they have come back to blogging after a break or if they're asking because of another reason I haven't thought of yet.

Thank you for raising the issue! I've added it to Groundskeeping `To Do` and it will be considered for prioritization within the next few months.

thehenrybyrd:
#2178404-zen (internal reference)

> How to change the username?

The user doesn't specify why - and this is the case for almost all cases I come across. This account has been around for two years, so I'm thinking more like your second case.
I can start asking why they want to change their username, while sending instructions, but honestly I don't expect to get a lot of response. 😕

thehenrybyrd:
Another in #2196455-zen.

> I've got absolutely the wrong username, [@username] is not my name 😆

thehenrybyrd:
# 2199257-zen

> can i change my username?

The account is two months old, and the user just changed their site address.

sophiegyo:
Another (Italian) in 2233788-zen.
Quick translation:

>Is it possible to change my username?

[...]

PR #8711, opened 2018-02-24
Social signup: username changing

nheagy:
Refs #8109

Allows users to change their username during social signup.

simulator screen shot - iphone x - 2018-02-26 at 11 51 49

To test:
- Signup for a new account
- Tap the username item on the new epilogue

Note:
- Needs a design review for the confirmation button and warning at the bottom of the screen
- Needs the same for copy
- PR isn't yet finished, not ready for code review

PR #12234, opened 2019-07-30
[Account Settings] Add Ability to Change Username

danielebogo:
Refs. #9705

username-change-2

This PR adds the ability to change the username. The feature is based on the _Flux + View Model_ pattern. The view controller is a subclass of `SignupUsernameTableViewController` in order to keep the same UI structure.

## To test:
- Select the Tab _Me_ -> _Account Settings_
- Tap on the _username_ to open the new screen and change the username
- If the user is an A8C user you shouldn't be able allowed to change the username

Update release notes:


WordPress/Classes/ViewRelated/Me/Change Username/ChangeUsernameViewController.swift
```diff
@@ -0,0 +1,153 @@
...
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        usernameTextfield.resignFirstResponder()
+    }
+}
+
+private extension ChangeUsernameViewController {
+    func setupViewModel() {
+        viewModel.reachabilityListener = { [weak self] in
+            DispatchQueue.main.async {
```
frosty: Should we just ensure that each of these listener blocks is called on the main queue from the viewModel?

WordPress/Classes/Stores/AccountSettingsStore.swift
```diff
@@ -0,0 +1,134 @@
+import WordPressFlux
+
+enum AccountSettingsState: Equatable {
+    case stationary
```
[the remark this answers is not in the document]
danielebogo: Oh I like `idle`!!! Thanks a lot!

danielebogo:
Thanks @frosty for reviewing it. I update everything. I also added the local validation considering the alphanumeric charset.

I added the alert to confirm the save action. I'm not sure about the copy though.

CC @SylvesterWilmott 

Simulator Screen Shot - iPhone X - 2019-08-01 at 16 25 16

Simulator Screen Shot - iPhone X - 2019-08-01 at 17 20 29

WordPress/Classes/ViewRelated/Me/Change Username/View Model/ChangeUsernameViewModel.swift
```diff
@@ -132,6 +143,13 @@ private extension ChangeUsernameViewModel {
...
             static let saveUsername = NSLocalizedString("There was an error saving the username", comment: "Text displayed when there is a failure saving the username.")
             static let emptyValue = NSLocalizedString("Username must be at least 4 characters.", comment: "Error displayed if the username input is an empty string")
+            static let alphanumeric = NSLocalizedString("Username can only contain lowercase letters (a-z) and numbers.", comment: "Error displayed if the username doesn't contain numbers or digits")
         }
     }
 }
+
+private extension String {
+    var isAlphanumeric: Bool {
+        return range(of: "[^a-zA-Z0-9]", options: .regularExpression) == nil
```
frosty: Are usernames allowed accented characters at all? (I don't know the answer to this) I suspect not, but if they are we may want to do something like

`rangeOfCharacter(from: CharacterSet.alphanumerics.inverted) == nil`

danielebogo: >rangeOfCharacter(from: CharacterSet.alphanumerics.inverted) == nil

This was my first attempt. Using characters sets to check the string, but it includes the accents. The regex then was the simple solution to get only alphanumerics characters ignoring diacritics.

danielebogo:
@frosty @SylvesterWilmott Calypso uses to add a confirmation textfield as soon you start typing in the _Username_ field. Although I think this is still the best experience, because with an alert the user can accidentally confirm the action, a second textfield (plus the actions row we can add in future) pushes the text to the bottom below the keyboard.
I was thinking if at this point we can use a button at the bottom of the page rather than the save button in the navigation bar. This forces the user to scroll the page and read the paragraph.
What do you think?

folletto:
Any reason why we aren't reusing the name suggestions from the signup flow to avoid people getting frustrated in guessing an available username?

danielebogo:
@folletto  we we are using the same experience of Calypso. There's an API call to validate what you type but not to suggest a new username. Should we stay with it or make a different flow? @SylvesterWilmott @frosty

folletto:
I haven't checked in a while so I'm not sure, but we already have a screen to do that in the signup flow! Already designed, with suggestions. 

See #8711 

To me _this_ is introducing a different flow, not the other way around, as the users are on mobile, and have already (maybe) experienced the other UI.

WordPress/Classes/ViewRelated/Me/Change Username/ChangeUsernameViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    enum Constants {
+        static let actionButtonTitle = NSLocalizedString("Save", comment: "Settings Text save button title")
+        static let username = NSLocalizedString("Username", comment: "The header and main title")
+        static let cellIdentifier = "SearchTableViewCell"
+
+        enum Alert {
+            static let loading = NSLocalizedString("Loading usernames", comment: "Shown while the app waits for the username suggestions web service to return during the site creation process.")
+            static let title = NSLocalizedString("Careful!", comment: "Alert title.")
+            static let message = NSLocalizedString("You are changing your Username to %@. Changing your username will also affect your Gravatar profile and IntenseDebate profile addresses. \nConfirm your new Username to continue.", comment: "Alert message.")
```
frosty: Nitpick, but a couple of occurrences of 'username' in here have a capital U. I think they should all be lowercase?

WordPress/Classes/ViewRelated/Me/Change Username/View Model/ChangeUsernameViewModel.swift
```diff
@@ -0,0 +1,140 @@
+import Reachability
+import WordPressFlux
+
+class ChangeUsernameViewModel {
+    typealias VoidListener = () -> Void
```
frosty: Perhaps just `Listener`?

WordPress/Classes/Services/AccountSettingsService.swift
```diff
@@ -18,6 +18,7 @@ protocol AccountSettingsRemoteInterface {
     func getSettings(success: @escaping (AccountSettings) -> Void, failure: @escaping (Error) -> Void)
     func updateSetting(_ change: AccountSettingsChange, success: @escaping () -> Void, failure: @escaping (Error) -> Void)
     func changeUsername(to username: String, success: @escaping () -> Void, failure: @escaping () -> Void)
+    func validateUsername(to username: String, success: @escaping () -> Void, failure: @escaping (Error) -> Void)
```
frosty: Are we still using the validate methods now we changed to the suggestion model? If not, should we remove them?

frosty:
> Should we disable the Save button until a different (new) username is selected?

I can't recreate this at the moment, so I'm not quite sure how I got into that state 🤔

One other thing I forgot to mention: on the confirmation alert, if you type the new username incorrectly we just dismiss the alert and fail silently. We should probably let the user know that their username wasn't changed.

danielebogo:
Hey @frosty thanks for your review! I pushed some change and it's ready for another round!
Here you can find some screenshot about the alert with the button disabled/enabled:

Simulator Screen Shot - iPhone X - 2019-08-08 at 15 47 02

Simulator Screen Shot - iPhone X - 2019-08-08 at 15 47 06

danielebogo:
Thanks @frosty ! It's ready again 😊

PR #12221, opened 2019-07-29
[Gutenberg] Migration to per-site editor setting and sync with server

etoledom:
To test:

## WP.com/Jetpack sites
- [x]  Create a new WP.com site on mobile and check Site Settings → should be Gutenberg.
- [x] Create a new WP.com site on mobile and start a new post → should be Gutenberg.
- [x] Clean app, clean backend, start new post → should be Aztec
- [x] Clean app, "Aztec" in backend, start new post → should be Aztec
- [x] Clean app, "Gutenberg" in backend, start new post → should be Gutenberg.
- [x] Clean  app, clean backend, open Gutenberg post → auto-enables Gutenberg
- [x] Clean  app, clean backend, open Gutenberg post (auto-enables GB), start new post → should be Gutenberg
- [x] Clean  app, clean backend, open Gutenberg post (auto-enables GB), open a GB post again→ Gutenberg but no popup
- [x] Clean app, clean backend, open Gutenberg post (auto-enables GB), new post in other site → should be Aztec
- [x] Clean app, clean backend, open Gutenberg post (auto-enables GB), open GB post in other site → auto-enables Gutenberg.
- [x] Toggling the setting in a site doesn't toggle it in others
- [x] Migrate from v12.9 with GB ON, start new post → should be Gutenberg in all sites
- [x] Migrate from v12.9 with GB OFF and been auto-enabled, start new post → should be Aztec in all sites
- [x] Migrate from v12.9 with GB OFF and never auto-enabled, start new post → should be Aztec in all sites
- [x] Migrate from v12.9 with GB OFF and never auto-enabled, open a GB post (should auto-enable GB), change to other site and start new post → should be Aztec
- [x] Migrate from v12.9 with GB OFF and never auto-enabled, open a GB post (should auto-enable GB), change to other site and open GB post → auto-enables Gutenberg on the site.

## Selfhosted sites

- [x] Self-hosted site added, start new post → should be Aztec
- [x] Self-hosted site added, open a GB post → auto-enables Gutenberg
- [x] Self-hosted site added, open a GB post (should auto-enable GB), start new post → should be Gutenberg
- [x] Self-hosted site added, open a GB post (should auto-enable GB), clean app data, re-add self-hosted site, start new post → should be Aztec
- [x] Self-hosted site added, open a GB post (should auto-enable GB), clean app data, re-add self-hosted site, open a GB post → auto-enables Gutenberg

[...]

PR #12235, opened 2019-07-30
[Gutenberg] Fix showing autoenable dialog after migration

etoledom:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/pull/12221#issuecomment-516079036

I tried to avoid adding the change on the migration to not need to modify the key used for per-site gutenberg enabled once. But the migration alternative feels a bit more "correct" (?).

I also added a set of tests to cover our backs regarding the autoenable logic and requirements.
If you notice there are some cases not covered by the tests please let me know.

To test:
- Migrate from v12.9 with GB ON, start new post → should be Gutenberg in all sites
- Should not show the autoenable dialog on new gutenberg post.

Update release notes:

### DIFF
--- a/WordPress/Classes/Utility/Editor/EditorFactory.swift
+++ b/WordPress/Classes/Utility/Editor/EditorFactory.swift
@@ -27,6 +27,7 @@ class EditorFactory {
             gutenbergSettings.setGutenbergEnabled(true, for: post.blog)
             gutenbergSettings.postSettingsToRemote(for: post.blog)
             gutenbergVC.shouldPresentInformativeDialog = true
+            gutenbergSettings.willShowDialog(for: post.blog)
         }
 
         return gutenbergVC
--- a/WordPress/Classes/Utility/Editor/GutenbergSettings.swift
+++ b/WordPress/Classes/Utility/Editor/GutenbergSettings.swift
@@ -87,12 +87,16 @@ class GutenbergSettings {
 
     /// True if gutenberg editor has been enabled at least once on the given blog
     func wasGutenbergEnabledOnce(for blog: Blog) -> Bool {
-        return database.object(forKey: Key.enabledOnce(for: blog)) != nil
+        return database.object(forKey: Key.enabledOnce(for: blog)) != nil || database.object(forKey: Key.appWideEnabled) != nil
     }
 
     /// True if gutenberg should be autoenabled for the blog hosting the given post.
     func shouldAutoenableGutenberg(for post: AbstractPost) -> Bool {
         return !wasGutenbergEnabledOnce(for: post.blog)
+    }
+
+    func willShowDialog(for blog: Blog) {
+        database.set(true, forKey: Key.enabledOnce(for: blog))
     }
 
     // MARK: - Gutenberg Choice Logic
--- a/WordPress/Classes/Utility/Editor/GutenbergSettings.swift
+++ b/WordPress/Classes/Utility/Editor/GutenbergSettings.swift
@@ -7,7 +7,7 @@ class GutenbergSettings {
     enum Key {
         static let appWideEnabled = "kUserDefaultsGutenbergEditorEnabled"
         static func enabledOnce(for blog: Blog) -> String {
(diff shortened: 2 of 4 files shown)


WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -87,14 +87,18 @@ class GutenbergSettings {
 
     /// True if gutenberg editor has been enabled at least once on the given blog
     func wasGutenbergEnabledOnce(for blog: Blog) -> Bool {
-        return database.object(forKey: Key.enabledOnce(for: blog)) != nil
+        return database.object(forKey: Key.enabledOnce(for: blog)) != nil || database.object(forKey: Key.appWideEnabled) != nil
```
koke: I don't think we want this, it means that the dialog will show only once per user and not per site. If you migrate from 12.9 and create a new site, it's getting gutenberg for new posts but not showing the dialog.

WordPress/Classes/Utility/Editor/EditorFactory.swift
```diff
@@ -27,6 +27,7 @@ class EditorFactory {
             gutenbergSettings.setGutenbergEnabled(true, for: post.blog)
             gutenbergSettings.postSettingsToRemote(for: post.blog)
             gutenbergVC.shouldPresentInformativeDialog = true
+            gutenbergSettings.willShowDialog(for: post.blog)
```
koke: I don't understand why we're adding this, since `setGutenbergEnabled` will also set the flag

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
etoledom: >  If you migrate from 12.9 and create a new site, it's getting gutenberg for new posts but not showing the dialog.

True! I wasn't aware that this was required, but I confirmed internally with @hypest 

I'll give the other option setting the flag on the migration itself a try.

WordPress/WordPressTest/GutenbergSettingsTests.swift
```diff
@@ -113,4 +119,107 @@ class GutenbergSettingsTests: XCTestCase {
...
+        database.set(nil, forKey: GutenbergSettings.Key.appWideEnabled)
+        migration()
+
+        post.content = gutenbergContent
+
+        XCTAssertTrue(mustUseGutenberg)
+        XCTAssertTrue(shouldAutoenableGutenberg)
+    }
+
+    func migration() {
```
etoledom: These tests are probably better suited to be run with the actual migration test, but we need to inject an UserDefaults double in the migration itself first.

etoledom:
Answering to https://github.com/wordpress-mobile/WordPress-iOS/pull/12235#discussion_r309245638:

We have a problem with new blogs:
- We set gutenberg via`softSetGutenberg` to avoid setting the enabledOnce key and show the dialog on new posts.
- But when the blog is selected in the app, there's a settings sync, and the "hard" `setGutenbergEnabled` is called, setting the enabledOnce key
- To avoid that, we will only set the key if the new `isEnabled` value is different than the previous.
- This bring us the problem that, for new blogs, the enabledOnce key won't be set on auto-enabling.
- So we "manually" set the enabledOnce key when we run the auto-enabling process via `willShowDialog`.

This problem seems to happen just on new blogs.

I tried looking for a solution to be implemented internally on `GutenbergSettings`, but it seems like we are missing a key specific to the auto-migration/dialog shown, independent of the gutenberg enabled key. And having this key, we would still need a new function to set it.

WordPress/Classes/Utility/Editor/EditorFactory.swift
etoledom: Answered here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12235#issuecomment-517660414 to have more space :)

etoledom:
@koke could you please take another look?

etoledom:
Thank you!

Issue #12218, opened 2019-07-28
“Clear Device Media Cache” doesn't clear everything 


AtrumGeost:
Hi folks!

I had a user who's WordPress.com app was using around 21.4 GB. We asked the user to go to **Me > App Settings** and select **Clear Device Media Cache**.

Also, we verified the user didn't have any offline drafts.

### Expected behavior
We expected the app to delete any local content it could be storing.

### Actual behavior
Even though the app was using a lot of space the "Media Cache Size" was "Empty". And selecting **Clear Device Media Cache** didn't help.

### Workaround

Re-installing the WordPress.com deleted the local content and the storage the app was using was reduced to 81.7 MB. 

### Steps to reproduce the behavior
There aren't specific steps for this. This happens over regular use and—I'm guessing—the constant upload of images to the WordPress.com site via the app.

##### Tested on iPhone 7, iOS 12.3.1, WPiOS 12.8.2
(User's devive info)

For internal reference: p1564148616088500-slack-mobile-support

frosty:
Should be fixed by #12241

PR #12241, opened 2019-07-31
Allow clearing of unused media uploads

frosty:
Refs #12218.

@etoledom, I wasn't sure who to ask to review this – ideally I'd ask @SergioEstevao, but he's still AFK for a while. I know you've worked on some media code in the past, but let me know if you'd rather I pass it to somebody else.

The issue referenced mentions reports of user device storage being used up and never cleared when we clear the media cache. I noticed two issues that I think contribute to this:

### Automatic upload clearing

We have code to clear the media uploads folder (in Documents) on app launch, but in my testing this is never actually happening. The predicate to find media that should be excluded from the deletion was:

`purgeMediaFiles(exceptMedia: NSPredicate(format: "blog != NULL || remoteURL == NULL")`

along with the comment:

```
/// Clear the local Media directory of any files that are no longer in use or can be fetched again,
/// such as Media without a blog or with a remote URL
```

The issue is that unless something has gone very wrong, media should never have a nil `blog`. Because practically every Media has a blog, this predicate was matching all media, and thus preventing any media from being deleted.

I've updated the predicate to only exclude items which **have** a blog AND don't have a remote URL – this should indicate that they haven't been uploaded yet:

`"blog != NULL && remoteURL == NULL"`

This should allow all uploaded media with a remoteURL to be cleared, and keep anything that hasn't been uploaded yet. Media without a blog will now be cleared, which makes sense to me as we can never do anything with it. I think this is more in line with the intention of the comment I quoted above.

### Clear uploads on manual cache clear

The second issue was that the "Clear Device Media Cache" button in App Settings only clears the temporary downloaded image cache and not the uploads directory. I've updated that call to run the same upload clearing that happens on startup.

In the screenshot below, the top folder is the uploads directory which has 3 successful uploaded images and one failed image. The bottom folder is the normal image caches directory. After tapping the clear cache button, everything is cleared except for the failed upload.

media-cache

**To test:**

* Build and run on the simulator (easiest to access the caches directories)
* Pause the app and `po NSHomeDirectory()`. Open this folder in Finder, and navigate into Documents > Media. This is the uploads store.
* Upload a few images in the media library and see that they get copied into the uploads folder.
* Upload another image but cause it to fail. You could use Network Link Conditioner set to 100% packet loss, or kill your network connection before the upload completes. The image for this should also be present in the uploads folder.
* Relaunch the app, and verify that all images except for the failed image are removed from the uploads directory. The failed image should automatically attempt to re-upload thanks to the work by Ravenclaw.
* Repeat the above steps, but instead of restarting the app, use the Me > App Settings > Clear Device Media Cache button. It should have the same effect.

cc @charliescheer 

Update release notes:

### DIFF
--- a/WordPress/Classes/Utility/Media/MediaFileManager.swift
+++ b/WordPress/Classes/Utility/Media/MediaFileManager.swift
@@ -200,7 +200,9 @@ class MediaFileManager: NSObject {
     ///
     @objc class func clearAllMediaCacheFiles(onCompletion: (() -> Void)?, onError: ((Error) -> Void)?) {
         let cacheManager = MediaFileManager(directory: .cache)
-        cacheManager.clearFilesFromDirectory(onCompletion: onCompletion, onError: onError)
+        cacheManager.clearFilesFromDirectory(onCompletion: {
+            MediaFileManager.clearUnusedMediaUploadFiles(onCompletion: onCompletion, onError: onError)
+        }, onError: onError)
     }
 
     /// Helper method for getting the default upload directory URL.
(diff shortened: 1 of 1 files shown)


charliescheer:
awesome!  Thanks @frosty

Issue #12240, opened 2019-07-30
[Auto-upload Published] Full publishing and cancellation flow


shiki:
This covers the following part of #12227: 

<img width="1243" alt="60972121-8ad36680-a31d-11e9-930b-e5d3f3fd79fc jpg (1320×3178) 2019-07-30 16-14-40" src="https://user-images.githubusercontent.com/198826/62169329-2e90b080-b2e5-11e9-9ca4-203eaeab6e71.png">

### Drafts

 Test Case | Status 
---|----
| <p>Post stays in Drafts and SnackBar appears</p><p>Status: Android and iOS show different messages.</p> |  |
| Local drafts are automatically uploaded **as drafts** | ✔️ #11839 #12215 | 
 Drafts with un-confirmed changes are remote-auto-saved | 
 Drafts with confirmed changes are automatically uploaded | 

### Publishing

 Test Case | Status 
---|----
 When offline:<ul><li> Post stays in Drafts or previous remote status<li> SnackBar appears: "Post will be published next time your device is online"</ul> | 
 Post List:<ul><li>Status message: "Post will be published next time your device is online"" <li>Cancel button is shown" </li> | 
 Cancellation<ul><li>Post stays in Drafts or previous remote status<li>SnackBar: "Changes will not be published"</ul> | 
 Published posts with confirmed changes are automatically uploaded. |  |
 Published posts with UN-confirmed changes are remote-auto-saved | 
 Cancelled posts are remote-auto-saved not auto-uploaded. | 
 It should be possible to revert a published post to a draft. | 
When a draft is uploaded, pressing on the notice's _Publish_ button should auto-upload the post as published | 
Tapping on _Publish_ in the Post List should auto-upload the post | 

### Additional scenarios

- A post with a failed media upload

## Plan

It looks like all parts have to be there in order to merge these to `develop`. We will create multiple PRs but target a master branch instead. 

Most of the work has been done by @diegoreymendez in PR #12178. We're going to take it apart into smaller chunks. And then we're going to need to test everything before merging to `develop`.

Targeting the `PUBLISHED` is deliberate to decrease the risks of regressions. If everything works great, we can apply the same to other statuses. 

## Known Tasks 

- [x] @shiki Update Post List to show failed published posts in Drafts. Done in #12252.
- [x] @shiki Auto-uploading PUBLISHED posts. Done in #12252.
- [ ] #12136 Safety for auto-uploading. 
[...]

Issue #12227, opened 2019-07-29
[Epic] Offline Posting


shiki:
This covers the planned offline posting epic as described by this:

![](https://user-images.githubusercontent.com/50576199/60972121-8ad36680-a31d-11e9-930b-e5d3f3fd79fc.jpg)

This epic links to all the related sub-issues.

Related posts:

- paCBwp-bF-p2
- paCBwp-c7-p2

## Out of Scope

Deletion while offline is currently out of scope.

### General Test Cases 

- Upgrading from an old version with failed posts and/or local drafts
- Local drafts with failed media
- Core Data migrations. We added migrations in #12347.

shiki:
Updated the cancelation messages based on the messages used in https://github.com/wordpress-mobile/WordPress-Android/pull/10343. This was agreed by the team to be the right message to use.

shiki:
Closing this since the [board](https://github.com/orgs/wordpress-mobile/projects/29) now clearly reflects what needs to be done.

PR #12252, opened 2019-07-31
Auto-upload local published posts

shiki:
Part of #12240. This is targeting the master branch `issue/12240-master-branch`. 

## Changes

Flow@2x

### Auto-uploading local `published` posts

Local drafts that the user set to `published` will now be automatically uploaded and **published** when the device is online. Previously, local `published` posts were changed to `DRAFT` and were automatically uploaded as `draft`.

This change was based on @diegoreymendez's work in #12178. 

This deliberately targets `published` only. Scheduled, private, and other statuses' behaviors are left unchanged. I believe this lowers the risk of regression. After everything is done, we can also test the whole publishing flow before fully committing with all statuses.

Please see #12240 for the current plan. 

### Post stays in Draft

This change also completes the _"Post stays in draft"_ requirement. This is a missing portion of #12178, as described in https://github.com/wordpress-mobile/WordPress-iOS/pull/12178#issuecomment-514307641.

- Local drafts that failed to upload will be shown in the Drafts list.
- Existing Published posts that were edited but failed in uploads will remain in the Published list. Please see https://github.com/wordpress-mobile/WordPress-iOS/issues/12240#issuecomment-516878607.

## Reviewing

Since this involves auto-uploading, it would probably be better if at least two developers can review this. 

## Testing

1. Go offline.
2. Create a post and publish it. Verify that the post will stay in Drafts.
3. Go online.
4. The post should be automatically published and will be moved to the Published list.

Please also test with other statuses like `scheduled/future`, `private`, and `pending`. Their behavior should not change.

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -254,7 +254,7 @@ extension PostCoordinator: Uploader {
                 return
             }
 
-            posts.forEach({ self.retrySave(of: $0) })
+            posts.forEach { self.retrySave(of: $0) }
         }
     }
 }
@@ -276,13 +276,13 @@ extension PostCoordinator {
             let allowedStatuses: [BasePost.Status] = [.draft, .publish]
 
             postService.getFailedPosts { posts in
-                let postsToRetry = posts.filter({ post -> Bool in
+                let postsToRetry = posts.filter { post -> Bool in
                     guard let status = post.status else {
                         return false
                     }
 
                     return allowedStatuses.contains(status) && !post.hasRemote()
-                })
+                }
 
                 result(postsToRetry)
             }
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -248,19 +248,37 @@ class PostCoordinator: NSObject {
 
 extension PostCoordinator: Uploader {
     func resume() {
-        let service = PostService(managedObjectContext: mainContext)
-
-        service.getFailedPosts { [weak self] posts in
+        let fetcher = FailedPostsFetcher(mainContext)
+        fetcher.getPostsToRetry { [weak self] posts in
             guard let self = self else {
                 return
             }
 
-            posts.forEach() { post in
-                let shouldRetry = post.status == .draft && !post.hasRemote()
+            posts.forEach({ self.retrySave(of: $0) })
+        }
+    }
+}
 
-                if shouldRetry {
-                    self.retrySave(of: post)
-                }
+extension PostCoordinator {
+    /// Fetches failed posts that should be retried when there is an internet connection.
+    class FailedPostsFetcher {
+        private let postService: PostService
+
+        init(_ postService: PostService) {
+            self.postService = postService
+        }
+
+        init(_ managedObjectContext: NSManagedObjectContext) {
+            postService = PostService(managedObjectContext: managedObjectContext)
+        }
+
+        func getPostsToRetry(result: @escaping ([AbstractPost]) -> Void) {
+            postService.getFailedPosts { posts in
+                let postsToRetry = posts.filter({ post -> Bool in
+                    return post.status == .draft && !post.hasRemote()
+                })
+
+                result(postsToRetry)
             }
         }
     }
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -273,9 +273,15 @@ extension PostCoordinator {
         }
 
         func getPostsToRetry(result: @escaping ([AbstractPost]) -> Void) {
+            let allowedStatuses: [BasePost.Status] = [.draft, .publish]
+
             postService.getFailedPosts { posts in
                 let postsToRetry = posts.filter({ post -> Bool in
-                    return post.status == .draft && !post.hasRemote()
+                    guard let status = post.status else {
+                        return false
+                    }
+
+                    return allowedStatuses.contains(status) && !post.hasRemote()
                 })
 
                 result(postsToRetry)
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -259,6 +259,17 @@ const NSUInteger PostServiceDefaultNumberToSync = 40;
             Post *postInContext = (Post *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
             if (postInContext) {
                 postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
+                // If the post was not created on the server yet we convert the post to a local
+                // draft post with the current date. However, if it's PUBLISHED or DRAFT, we will
+                // not reset the status so that when retrying the upload, the post will have the
+                // status that the user originally intended.
+                //
+                // When we support auto-uploading of all statuses, we will fully remove this block.
+                if (!postInContext.hasRemote && ![postInContext.status isEqualToString:PostStatusPublish]) {
+                    postInContext.status = PostStatusDraft;
+                    postInContext.dateModified = [NSDate date];
+                }
+
                 [[ContextManager sharedInstance] saveContext:self.managedObjectContext];
             }
             if (failure) {
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -259,11 +259,6 @@ const NSUInteger PostServiceDefaultNumberToSync = 40;
             Post *postInContext = (Post *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
             if (postInContext) {
                 postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
-                // If the post was not created on the server yet we convert the post to a local draft post with the current date.
-                if (!postInContext.hasRemote) {
-                    postInContext.status = PostStatusDraft;
-                    postInContext.dateModified = [NSDate date];
-                }
                 [[ContextManager sharedInstance] saveContext:self.managedObjectContext];
             }
             if (failure) {
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -259,12 +259,14 @@ const NSUInteger PostServiceDefaultNumberToSync = 40;
             Post *postInContext = (Post *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
             if (postInContext) {
                 postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
-                // If the post was not created on the server yet we convert the post to a local
-                // draft post with the current date. However, if it's PUBLISHED or DRAFT, we will
-                // not reset the status so that when retrying the upload, the post will have the
-                // status that the user originally intended.
+                // If the post was not created on the server yet we convert the post to a local draft
+                // with the current date. This post upload will be automatically retried later as a draft.
                 //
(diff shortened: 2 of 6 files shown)


WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -248,19 +248,43 @@ class PostCoordinator: NSObject {
 
 extension PostCoordinator: Uploader {
     func resume() {
-        let service = PostService(managedObjectContext: mainContext)
-
-        service.getFailedPosts { [weak self] posts in
+        let fetcher = FailedPostsFetcher(mainContext)
```
shiki: I extracted the logic into a class so it can be tested easily. See 60dc5a721173eba04885031d178eb8c8bcd141c8.

yaelirub:
@shiki , before I dive into this PR I have a question for you and @osullivanchris .
We keep saying that we will not publish without the user's confirmation.
In one of our latest post, looking back on old posts getting auto uploaded, @diegoreymendez wrote:
"Make sure the user has confirmed publishing a post." as one of our realizations. 
@osullivanchris  said in another case that we should treat "only publish on confirmation" as a principal. 
I understand that it is not included in the mocks but I wonder if we should show a confirmation dialog before publishing.
Any thoughts?

shiki:
> I understand that it is not included in the mocks but I wonder if we should show a confirmation dialog before publishing.

@yaelirub Do you mean **before** automatically publishing? Or when pressing the _Publish_ button in the editor?

diegoreymendez:
Just a small clarification: whenever I say "we will not publish without the user's confirmation", I'm specifically referring to what @malinajirka has been doing in Android with the hash to verify that we're uploading what the user intended to.

The other safety measure I'm vouching for is having a timeout for not publishing things that were confirmed ages ago.

It is my belief that these two could be enough.

osullivanchris:
> @osullivanchris said in another case that we should treat "only publish on confirmation" as a principal.

@yaelirub Do you mean before automatically publishing? Or when pressing the Publish button in the editor?

- When the user hits 'publish' they should see a confirmation dialog. It already exists in iOS appearing as a bottom/action sheet "Are you sure you want to publish now". That covers this requirement that we had. I assume @shiki just didn't include it in the diagram
- In the image in this pull request the snackbar/toast message is not the same as my mocks in #12240 - is that just because this is a subtask and the snackbar/toast message will be updated in a separate task? I would not want this to go live by itself as I think the new messaging "Post will be published next time your device is online" is important to set the correct expectations, along with the ability to cancel if that is not desired.

Let me know if any more clarification needed from my side or iOS specific mocks, but the flow should be the same.

shiki:
is that just because this is a subtask and the snackbar/toast message will be updated in a separate task?

@osullivanchris Correct. This is targetting a master branch. We'll be adding more to that branch. Once everything's been done, we'd like to go over it with you and everyone else before releasing it.

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -248,19 +248,43 @@ class PostCoordinator: NSObject {
...
+        fetcher.getPostsToRetry { [weak self] posts in
             guard let self = self else {
                 return
             }
 
-            posts.forEach() { post in
-                let shouldRetry = post.status == .draft && !post.hasRemote()
+            posts.forEach({ self.retrySave(of: $0) })
```
jklausa: style nitpick, could be just `posts.forEach { self.retrySave(of: $0) }`

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -248,19 +248,43 @@ class PostCoordinator: NSObject {
...
+
+        init(_ managedObjectContext: NSManagedObjectContext) {
+            postService = PostService(managedObjectContext: managedObjectContext)
+        }
+
+        func getPostsToRetry(result: @escaping ([AbstractPost]) -> Void) {
+            let allowedStatuses: [BasePost.Status] = [.draft, .publish]
+
+            postService.getFailedPosts { posts in
+                let postsToRetry = posts.filter({ post -> Bool in
```
jklausa: style nitpick, could omit the `()` here

WordPress/Classes/Services/PostService.m
```diff
@@ -259,11 +259,17 @@ - (void)uploadPost:(AbstractPost *)post
             Post *postInContext = (Post *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
             if (postInContext) {
                 postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
-                // If the post was not created on the server yet we convert the post to a local draft post with the current date.
-                if (!postInContext.hasRemote) {
+                // If the post was not created on the server yet we convert the post to a local
+                // draft post with the current date. However, if it's PUBLISHED or DRAFT, we will
```
jklausa: This is super nitpicky, but it took me three reads of this comments to really understand it. Could we maybe reword the following section
>However, if it's PUBLISHED or DRAFT, we will

into something that talks about user intent, and not the state of a property?

so something like:

>However, if the post was supposed to be published or was a draft (e..g. has a `PostStatusPublish` or `PostStatusDraft`)

or something along those lines?

A sentence somewhere indicating something like "This is required because we _currently_ only want to automatically retry uploads of posts that are blah blah blah" near the end would also be great!

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -248,19 +248,43 @@ class PostCoordinator: NSObject {
...
+        init(_ postService: PostService) {
+            self.postService = postService
+        }
+
+        init(_ managedObjectContext: NSManagedObjectContext) {
+            postService = PostService(managedObjectContext: managedObjectContext)
+        }
+
+        func getPostsToRetry(result: @escaping ([AbstractPost]) -> Void) {
+            let allowedStatuses: [BasePost.Status] = [.draft, .publish]
```
jklausa: should this include `.publishPrivate`? if not, why not?

WordPress/Classes/ViewRelated/Post/PostListFilter.swift
```diff
@@ -68,7 +68,7 @@ import Foundation
         let filterType: Status = .published
         let statuses: [BasePost.Status] = [.publish, .publishPrivate]
 
-        let predicate = NSPredicate(format: "status IN %@", statuses.strings)
+        let predicate = NSPredicate(format: "postID > 0 AND status IN %@", statuses.strings)
```
jklausa: I know you explain why this is needed like 15 lines below in `draftFilter` comment, but adding a short line like "see predicate on `draftFilter()` on why this is needed would be great — when debugging things, it's easy to develop a tunnel vision and miss things like that :)

WordPress/Classes/Services/PostService.m
```diff
@@ -259,11 +259,17 @@ - (void)uploadPost:(AbstractPost *)post
...
+                // not reset the status so that when retrying the upload, the post will have the
+                // status that the user originally intended.
+                //
+                // When we support auto-uploading of all statuses, we will fully remove this block.
+                if (!postInContext.hasRemote && ![postInContext.status isEqualToString:PostStatusPublish]) {
                     postInContext.status = PostStatusDraft;
```
jklausa: Apologize if this is a dumb question, but I'm confused how this line (setting `postInContext.status` to `draft`  if a upload failure occured, is supposed to work with the new filtering on `getPostsToRetry`. 

If the only statuses a failed post can have is either:
a) `publish`, because we explicitly check for it and don't touch it
b) `draft`, because we blindly overwrite it in every other case

...then isn't the filtering on `getPostsToRetry()` kinda pointless?

shiki: Great observation! And it's not a dumb question. You're absolutely right. As it is right now, it does look pointless. 

In the next following PRs, we will be adding safety features like: 

- Do not auto-upload if the failure was more than 48 hours ago. Reference: https://github.com/wordpress-mobile/WordPress-Android/issues/9555#issuecomment-510844042 
- Do not auto-upload if the user has not confirmed intent to publish https://github.com/wordpress-mobile/WordPress-iOS/pull/12252#issuecomment-517480673. 
- Possibly remote auto-save instead if we cannot auto-upload.

These are currently proposals. We'd love to hear what you think about those. 

At a minimum, `FailedPostsFetcher` will look like this in the future:

```swift
postService.getFailedPosts { posts in
    let postsToRetry = posts.filter({ post -> Bool in
        guard let status = post.status else {
	    return false
        }
    
        if failureHappenedMoreThan48HoursAgo() {
    	    return false
        }
    
        if userDidNotConfirmIntent() {
    	    return false
        }
    
        // This will no longer be needed if we support all statuses
        // return allowedStatuses.contains(status) && !post.hasRemote()

        return true
    })
    
    result(postsToRetry)
}
```

WordPress/Classes/Services/PostCoordinator.swift
shiki: Instead of fully implementing #12227, I wanted to limit the scope to just #12240 which will only support published posts for now. There are only a few steps towards supporting `.publishPrivate` like adding a different message for the Snack bar and Post List:

> Private post will be published next time your device is online

Instead of: 

> Post will be published next time your device is online

Not adding `.publishPrivate` for now decreases the amount of code which decreases the amount of testing we will need to do in order to achieve the full workflow. 



I have listed the known tasks in #12240. 

What do you think about this plan, @jklausa?

WordPress/Classes/Services/PostService.m
shiki: @jklausa  Thanks for pointing that out. I left the original first sentence there without realizing how the whole paragraph reads. It is quite confusing. 

I'm not sure if this is better, but please let me know what you think: 

```swift
// If the post was not created on the server yet we convert the post to a local draft 
// with the current date. This post upload will be automatically retried later as a draft. 
// 
// However, if the post was supposed to be published or draft, we will leave it as is. 
// This is intentional because we currently want to automatically retry posts that 
// are either published or drafts. In the future, we will automatically retry all statuses. 
//
// Automatic uploads happen in `PostCoordinator.resume()`.
```

jklausa: that's much better, thank you!

WordPress/Classes/Services/PostService.m
jklausa: alrighty, with more context that makes much more sense — i was being perhaps overly cautious due to the recent problems ;) thanks for the explanation!

WordPress/Classes/Services/PostCoordinator.swift
shiki: Thanks for this. Done in e18bb15645d45205b9da4eed99baa44ed08c1dc5.

WordPress/Classes/Services/PostCoordinator.swift
shiki: Done in e18bb15645d45205b9da4eed99baa44ed08c1dc5.

WordPress/Classes/Services/PostService.m
shiki: Thank you! Done in 76cfa6c9bac57fc96fa855951dcc7857cd2aa588.

WordPress/Classes/ViewRelated/Post/PostListFilter.swift
shiki: Thanks for pointing it out. I also found a small bug here. Fixed and updated the comment in 4c13ac72896785c6d46dd8f1ed2ecf437477603b.

yaelirub:
@shiki , I tested according to the instructions and it seems to be working as expected. Im good with merging if @jklausa approves

PR #12254, opened 2019-07-31
Fix notification request crash on iOS 13

nheagy:
Refs #11221

This fixes a crash after approving push notifications on iOS 13.

To test:
- Install a fresh install of WPiOS
- Login
- Re-launch (why isn't the request happening the first time? 🤔)
- Approve push notifications
- Ensure nothing crashes

Update release notes:


nheagy:
Ya, I think we should look into not requesting the first launch next 😬

Thanks, @frosty!

PR #12255, opened 2019-08-01
Now showing loading status in navigation bar when preview is loading 

### LINKED ISSUE #12032: Loading preview spinner is blocking user interaction. Show status in navigation bar instead
Similar to https://github.com/wordpress-mobile/WordPress-iOS/issues/11812 when the preview is loading we currently show a spinner that is blocking user interaction. 
This becomes prominent on low connectivity and can leaves the user stuck till the preview loads.

We need to handle this in the same way we do when saving/ generating before showing the preview, but displaying the status in the navigation bar.

yaelirub:
This replaces the UI blocking HUD. Prominent in low connectivity. 
Fixes #12032 

To test:
1. Enable Network Link Conditioner for bad network
2. Update a post or a draft
3. Choose preview
4. See the loading status in the navigation bar instead of a blocking HUD
5. Make sure you can navigate back while the spinner is loading 

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
+++ b/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
@@ -7,7 +7,6 @@
 #import "WordPress-Swift.h"
 
 @import Gridicons;
-@import SVProgressHUD;
 @import WordPressUI;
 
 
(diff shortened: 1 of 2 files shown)


jklausa:
I second the visual overload concerns - just the activity indicator would work fine IME, but maybe @osullivanchris has other ideas :)?

yaelirub:
Apologies @shiki  and @jklausa , I forgot to push the changes I made to make it look like in the gif where we only show the title after the loading is done

yaelirub:
@shiki , @jklausa , ready for another look

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -7,7 +7,6 @@
 #import "WordPress-Swift.h"
 
 @import Gridicons;
-@import SVProgressHUD;
```
jklausa: love this!

PR #12257, opened 2019-08-01
Fix crash after logout

### LINKED ISSUE #12242: NSRangeException in SiteSettingsViewController
Sentry Issue: [WORDPRESS-IOS-586](https://sentry.io/organizations/a8c/issues/1056232079/?referrer=github_integration)

```
NSRangeException: *** -[__NSArrayM objectAtIndexedSubscript:]: index 3 beyond bounds [0 .. 2]
  File "SiteSettingsViewController.m", line 784, in -[SiteSettingsViewController tableView:titleForHeaderInSection:]
...
(56 additional frame(s) were not displayed)
```

etoledom:
Fixes #12242 

**Reason of the crash:**

With the site settings controller loaded, after the user logs out, the blog's account becomes nil.
This makes the `tableSections` return to change, while the table still expected the old values.
For some reason the system will refresh the sections (or possibly, all visible cells) when the app goes to background, and given the state previously described, the crash will happen. 💥 

**Fix proposal:**
This PR forces to reload the table when there default account changes (i.e. user logout), updating the table's data source.

Maybe another option could be to manually remove the controller when a logout happens.

To test:
- Follow the instructions from #12242 
- Check that the app doesn't crash.
- Run UI tests locally on iPad
- Check that all UI tests pass

Update release notes:

cc @jtreanor

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/SiteSettingsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/SiteSettingsViewController.m
@@ -17,6 +17,7 @@
 #import "WordPress-Swift.h"
 #import "WPWebViewController.h"
 #import <wpxmlrpc/WPXMLRPC.h>
+#import "AccountService.h"
 @import WordPressKit;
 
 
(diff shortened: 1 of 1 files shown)


etoledom:
Thank you @rachelmcr !

@jleandroperez - Would you mind taking a look to the code changes? :)

etoledom:
Thank you so much both of you! :)

Issue #11863, opened 2019-06-05
[Stats Refresh] accessing Stats from a Blog shows old post stats


ScoutHarris:
### Expected behavior
Accessing Stats from a Blog in Blog Posts should show the new post stats.

### Actual behavior
Accessing Stats from a Blog in Blog Posts shows the old post stats.

### Steps to reproduce the behavior
- Go to Blog Posts.
- On a blog, select `More`.
- Select `Stats`. Notice the old post stats are shown.

<img width="400" alt="blog" src="https://user-images.githubusercontent.com/1816888/58995018-beb3ed00-87af-11e9-91a2-8fcd4845a27d.png">

<img width="400" alt="old_stats" src="https://user-images.githubusercontent.com/1816888/58995024-c4a9ce00-87af-11e9-9101-7a573fc34e8f.png">

ScoutHarris:
Fixed with https://github.com/wordpress-mobile/WordPress-iOS/pull/11864.

PR #12259, opened 2019-08-01
Change page list labels (private, etc) to be gray

nheagy:
Refs #11863

> Yellow "private" text should be gray, preceded by a •

Before | After
----- | -----
| image | image |

To test:
- Open the pages list, find or create a private page
- Ensure that the new appearance matches the description

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Pages/PageListTableViewCell.m
+++ b/WordPress/Classes/ViewRelated/Pages/PageListTableViewCell.m
@@ -111,8 +111,7 @@ static CGFloat const FeaturedImageSize = 120.0;
     self.titleLabel.adjustsFontForContentSizeCategory = YES;
     
     self.titleLabel.textColor = [UIColor murielNeutral70];
-    self.timestampLabel.textColor = [UIColor murielNeutral30];
-    self.badgesLabel.textColor = [UIColor murielWarningDark];
+    self.badgesLabel.textColor = [UIColor murielNeutral];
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Pages/PageListTableViewCell.m
```diff
@@ -145,27 +144,22 @@ - (void)configureBadges
...
-            badgesString = [badgesString stringByAppendingString:@" · "];
-        }
-        badgesString = [badgesString stringByAppendingString:NSLocalizedString(@"Local changes", @"Title of the Local Changes Badge")];
+        [badges addObject:NSLocalizedString(@"Local changes", @"Title of the Local Changes Badge")];
     }
     
-    self.badgesLabel.text = badgesString;
-}
-
-- (void)configureTimeStamp
```
danielebogo: This needs to be removed also in `- (void)setPost:(AbstractPost *)post` where `[self configureTimeStamp]`

nheagy:
@danielebogo thanks! Fixed that up 👍

mattmiklic:
Looking good!

mattmiklic:
Actually one small thing; instead of `•` could you use the `·` character? That's Unicode U+00B7. That would be the more typographically correct symbol. :)

nheagy:
Thanks @danielebogo @mattmiklic!

PR #12263, opened 2019-08-02
Update several screen's background colors

nheagy:
Refs #11683

This fixes the background colors for: Activity, Menu, Themes, and Media screens. It also includes some text color fixes for activity details view.

image
image
image

To test:
- Check out the screens mentioned above
- Verify that they all have the same background color, which is the same as the blog details view (and hopefully, by now, most of the rest of the app)

Update release notes:


nheagy:
@mattmiklic how do those colors look?

PR #12272, opened 2019-08-05
Update colors on plugin screens

nheagy:
Refs #11683

This updates colors on the plugin screens:

Directory | List | Detail
--- | --- | ---
Simulator Screen Shot - iPhone Xʀ - 2019-08-05 at 22 55 42|Simulator Screen Shot - iPhone Xʀ - 2019-08-05 at 22 52 07|Simulator Screen Shot - iPhone Xʀ - 2019-08-05 at 22 52 00

To test:
- On a Jetpack site, tap Plugins 
- Ensure the colors in this section have been updated

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Plugins/PluginListCell.swift
+++ b/WordPress/Classes/ViewRelated/Plugins/PluginListCell.swift
@@ -24,8 +24,8 @@ class PluginListCell: UITableViewCell {
         }
     }
 
-    override func layoutSubviews() {
-        super.layoutSubviews()
+    override func awakeFromNib() {
+        super.awakeFromNib()
 
         nameLabel.textColor = .text
         authorLabel.textColor = .textSubtle
--- a/WordPress/Classes/ViewRelated/Plugins/PluginListCell.swift
+++ b/WordPress/Classes/ViewRelated/Plugins/PluginListCell.swift
@@ -23,6 +23,14 @@ class PluginListCell: UITableViewCell {
             view.leadingAnchor.constraint(greaterThanOrEqualTo: accessoryViewContainer.leadingAnchor).isActive = true
         }
     }
+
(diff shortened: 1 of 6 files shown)


WordPress/Classes/ViewRelated/Plugins/PluginListCell.swift
```diff
@@ -23,6 +23,14 @@ class PluginListCell: UITableViewCell {
             view.leadingAnchor.constraint(greaterThanOrEqualTo: accessoryViewContainer.leadingAnchor).isActive = true
         }
     }
+
+    override func layoutSubviews() {
```
jklausa: more of a question rather than a nitpick, but: why the override of `layoutSubview` here, instead of `awakeFromNib` or something else?

jklausa:
Looks good overall, just one question and then it's g2g!

mattmiklic:
Two small things on the design front:

- The "Active" and "Needs updates" labels: the icons are the correct color, but we should use Gray 50 for the text as we're no longer using green/yellow/red for text colors to ensure minimum contrast.
- The gray background behind the Plugin icon/name/author is a different shade of gray than the background between the grouped tables below it. All of them should be Gray 0.

WordPress/Classes/ViewRelated/Plugins/PluginListCell.swift
nheagy: tbh I copy & pasted from elsewhere in the app then edited. Since it's _only_ colors, I think you're right that `awakeFromNib` is a better fit 👍

nheagy:
Alright @jklausa @mattmiklic this is ready for another look (I've also updated the screenshots in the description)

nheagy:
Thanks @mattmiklic @jklausa!

PR #12276, opened 2019-08-06
[Gutenberg] Implement new single request to bulk update user's mobile editor preference

etoledom:
This PR implements the new request from https://github.com/wordpress-mobile/WordPressKit-iOS/pull/176

To test:
- Run the migration from 12.9 to 13.0
- Check that the settings on server corresponds with the expectation:
  - All blogs without mobile editor set, should be set to the new value given
  - All blogs with a mobile editor set, should remain unchanged
  - The local settings for all blogs should map the final state of mobile editor settings on server

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/EditorSettingsService.swift
+++ b/WordPress/Classes/Services/EditorSettingsService.swift
@@ -31,23 +31,6 @@ import Foundation
         }, failure: failure)
     }
 
-    private func update(_ blog: Blog, remoteEditorSettings settings: EditorSettings) throws {
-        blog.webEditor = WebEditor(rawValue: settings.web.rawValue)
-        guard settings.mobile != .notSet else {
-            throw EditorSettingsServiceError.mobileEditorNotSet
-        }
-        GutenbergSettings().setGutenbergEnabled(settings.mobile == .gutenberg, for: blog)
-    }
-
-    private func migrateLocalSettingToRemote(for blog: Blog, success: @escaping () -> Void, failure: @escaping (Swift.Error) -> Void) {
-        if blog.mobileEditor == nil {
-            let settings = GutenbergSettings()
-            let defaultEditor = settings.getDefaultEditor(for: blog)
-            settings.setGutenbergEnabled(defaultEditor == .gutenberg, for: blog)
-        }
-        postEditorSetting(for: blog, success: success, failure: failure)
-    }
-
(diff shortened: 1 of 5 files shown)


WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -65,47 +65,50 @@ import Foundation
...
-                self.syncEditorSettings(for: blog, success: {
-                    next()
-                }, failure: { (error) in
-                    DDLogError("Error saving editor settings: \(error)")
-                    next()
-                })
-            }
+
+        let isGutenbergEnabledAppWide = database.bool(forKey: GutenbergSettings.Key.appWideEnabled)
+        let editor: EditorSettings.Mobile = isGutenbergEnabledAppWide ? .gutenberg : .aztec
```
etoledom: I'm thinking on having this as a parameter, so this method is more generic and can be reused later if needed

WordPress/Classes/Services/EditorSettingsService.swift
```diff
@@ -65,47 +65,50 @@ import Foundation
         }, failure: failure)
     }
 
-    func syncEditorSettingsForAllBlogs() {
-        let blogService = BlogService(managedObjectContext: managedObjectContext)
-        guard let blogs = blogService.blogsForAllAccounts() as? [Blog] else {
+    /// This method is intended to be called only after the db 87 to 88 migration
+    ///
+    func postAppWideEditorSettingToRemoteForAllBlogsAfterMigration(database: KeyValueDatabase = UserDefaults.standard) {
```
koke: This is OK but a bit verbose, how about `migrateGlobalSettingToRemote`. And since the migration calling is already looking up the value, why not pass a `Bool` with the setting instead of the database?

etoledom: > And since the migration calling is already looking up the value, why not pass a Bool 

That sounds good! I was thinking about something like that on https://github.com/wordpress-mobile/WordPress-iOS/pull/12276/files#r311444419 👍

etoledom:
Thank you @koke !
I did the proposed refactor if you want to take a final look. No behavioral changes.
The methods meant to be overwritten on tests can't be private, but the rest were moved to the extension 👍

koke:
Ah, I forgot about the tests for a moment and thought all of those were meant to be private 🤦‍♂ 
Changes look good :shipit:

PR #12287, opened 2019-08-08
Move Notification Settings to Notifications tab

frosty:
This PR implements an item from the IA project – moving Notification Settings from the Me tab to the Notifications tab. Android PR is here: https://github.com/wordpress-mobile/WordPress-Android/pull/10335

notifications

**To test:**

* Build and run
* Ensure that Notification Settings is no longer present in the Me tab
* Ensure that the cog icon is present in the Notifications screen when: you are signed into a wpcom site or a Jetpack site, but not when you are just signed into a non-Jetpack self-hosted site
* Tap the icon and ensure Notification Settings are displayed and usable
* Check universal links still work – with the simulator open, in Terminal run: `xcrun simctl openurl booted https://wordpress.com/me/notifications`. This should launch Notification Settings.

Update release notes:


iamthomasbishop:
Design-wise this looks mostly solid, and this will make folks using the app happy 😄 One question regarding:

> Ensure that Notification Settings is no longer present in the Me tab

My feelings on this aren't strong, but I think we might want to consider leaving the Notifications Settings cell on the *Me* page temporarily, so we can measure its usefulness against the new entry point. I understand keeping the Notification Settings in that table would mean two entry points, but I think it might be a good thing to measure. I'd love to hear others' opinions on the topic, though.

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1190,6 +1220,10 @@ private extension NotificationsViewController {
         return AccountHelper.isDotcomAvailable() == false
     }
 
+    var shouldDisplaySettingsButton: Bool {
+        return AccountHelper.isDotcomAvailable() == true
```
danielebogo: Can this one returns directly `AccountHelper.isDotcomAvailable()`?

danielebogo: I know it's not related to this PR but I think also some line above you can do something like:
```
var shouldDisplayJetpackPrompt: Bool {
        return !AccountHelper.isDotcomAvailable()
    }
```

frosty: Right, for the second item I think some folks do `== false` instead of `!` because it can be more obvious to read at a glance (it could be easy to not see the `!` at the start).

osullivanchris:
> My feelings on this aren't strong, but I think we might want to consider leaving the Notifications Settings cell on the _Me_ page temporarily, so we can measure its usefulness against the new entry point. I understand keeping the Notification Settings in that table would mean two entry points, but I think it might be a good thing to measure. I'd love to hear others' opinions on the topic, though.

Hey @iamthomasbishop back from AFK and catching up on this one. I agree it would be a good thing to measure. But of all the things we _could_ measure I'm not sure this is one to prioritise. My reasons:

- Its unlikely we will change our approach
- User testing indicated this was a clearer entry point
- Two entry points while small is a complexity we have to remember, support, and clean up later
- Its not perfect but we can by proxy understand the effect by seeing how many more people use the feature after the change (I know this isn't very good science but should be enough in this case)

Hope you're ok with that!

iamthomasbishop:
@osullivanchris All good points, my suggestion was mainly out of curiosity, to run as a short test in production – in other words, keep both to gather some data for a short period of time (~1 week?) to see the effects. This isn't a strong rationale, however, so it's fine by me if we decide to skip this altogether.

loremattei:
I'm moving this to 13.2 since 13.1 has been cut. If you want this to make it to 13.1, please move it back, target the `release/13.1` branch and ping me to build a new beta.

frosty:
@osullivanchris I've added the footer to the Me screen with the suggested text. I made one slight tweak to the text, adding 'now' in there to indicate that we just moved it. Look ok?

Simulator Screen Shot - iPhone X - 2019-08-15 at 15 31 25

iamthomasbishop:
@frosty I like this, but what do you think about shortening the label to something along the lines of `Notification Settings are now in the Notifications tab` or `Notification Settings can be found in the Notifications tab`?

PR #12290, opened 2019-08-09
Notifications: Add support for `providesAppNotificationSettings`

frosty:
I just discovered a Notifications API that I figured we should probably implement. It allows us to hook into system notifications settings UI, providing a way for users to jump into Notification Settings in our app. There are two places it surfaces:

1. In system **Settings > WordPress > Notifications**, there's now a **WordPress Notification Settings** item:

s

2. When you receive a notification, if you either expand it and tap the `...` button or swipe it and tap Manage, you're shown a system menu for Deliver Quietly, Turn Off... or Settings. Now, if you tap Turn Off... you'll be presented with a new option to jump into the app to manage your settings:

Slice

**To test:**

* Build and run
* Navigate to system settings and check that item 1 above exists and works.
* Receive a notification, and bring up the menu shown in item 2. Tap Turn Off... and then choose Configure in WordPress... Check you're taken to notification settings in the app.

Note, you may need the debug notification server running to receive notifications – please let me know if you need help with this!

Update release notes:

### DIFF
--- a/WordPress/Classes/Utility/InteractiveNotificationsManager.swift
+++ b/WordPress/Classes/Utility/InteractiveNotificationsManager.swift
@@ -52,8 +52,14 @@ final class InteractiveNotificationsManager: NSObject {
         defer {
             WPAnalytics.track(.pushNotificationOSAlertShown)
         }
+
+        var options: UNAuthorizationOptions = [.badge, .sound, .alert]
+        if #available(iOS 12.0, *) {
+            options.insert(.providesAppNotificationSettings)
+        }
+
         let notificationCenter = UNUserNotificationCenter.current()
-        notificationCenter.requestAuthorization(options: [.badge, .sound, .alert]) { (allowed, _)  in
+        notificationCenter.requestAuthorization(options: options) { (allowed, _)  in
             DispatchQueue.main.async {
                 if allowed {
                     WPAnalytics.track(.pushNotificationOSAlertAllowed)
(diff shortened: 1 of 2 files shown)


WordPress/Classes/Utility/InteractiveNotificationsManager.swift
```diff
@@ -502,4 +508,8 @@ extension InteractiveNotificationsManager: UNUserNotificationCenterDelegate {
             completionHandler()
         }
     }
+
+    func userNotificationCenter(_ center: UNUserNotificationCenter, openSettingsFor notification: UNNotification?) {
+        MeNavigationAction.notificationSettings.perform([:])
```
danielebogo: I saw `func perform(_ values: [String: String], source: UIViewController? = nil)` doesn't use the value at the moment. What do you think if we remove it or add `[:]` as default value in order to use `.perform()` only?

frosty: This actually used to be optional, but @diegoreymendez changed it recently: https://github.com/wordpress-mobile/WordPress-iOS/pull/11851/files

While we can't add a default parameter to the protocol (I think?) I just added one to MeNavigationAction.

frosty:
> Do you know if we can in some way force the authorization again to add the new value to the options?

Hmm, good question. I think we should leave it for this PR, but maybe we're not doing notification permissions correctly? As far as I was aware, you're supposed to call the requestAuthorization method every time the app launches, but it seems like we only ever do it once...

danielebogo:
>  I think we should leave it for this PR, but maybe we're not doing notification permissions correctly?

Absolutely. This one can be merged with no problem!

PR #12303, opened 2019-08-13
Fixes an issue that causing our previews to cancel loading.

### LINKED ISSUE #10677: Preview of draft post/View of published post is not loading images every time
### Expected behavior

Previewing a draft post loads all its content.

### Actual behavior

When there is an image on a draft post, it is not shown on preview. This happens when the draft was just created on and the site is private. Sometimes previewing an old post, the image is shown.

- Create a new post.
- Write some title.
- Add an image.
- Save as draft.
- Make sure that the `Update` button is inactive (the post has no changes).
- Press the `more` (`...`) button.
- Select `Preview`.
- The image won't show.

preview

##### Tested on iPhone XS, iOS 12.1, WPiOS 11.3.1

diegoreymendez:
Fixes #10677 

The issue seems to be that our `UIWebView` delegate methods call `stopLoading`.  This would be fine if the delegate methods were called once for the web view, but they can be called multiple times for any resources (like embeds) that the page loads.

## Testing requirements / suggestions

@shiki hasn't been able to reproduce this reliably, while I've managed to reproduce it every single time.  When testing this I suggest to use the network link conditioner and try to make sure it's set to have a very slow connection.

The reason for this is that the bug relies on the second request being started BEFORE the first one completes.  On a very fast connection, the first request could complete really fast, and make it look like everything's fine.

## To test:

@rachelmcr - Can I ask you to test this fix on several different site types with and without custom domains?

It should be enough to:

1. Have a slow connection.
2. Preview any post with an image, to make the post loading interruption more apparent (as crazy as it sounds, this bug isn't really specific to images though - but they do make it easy to test the bug).

## Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
+++ b/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
@@ -95,7 +95,6 @@
 - (void)viewWillDisappear:(BOOL)animated
 {
     [super viewWillDisappear:animated];
-    [self stopLoading];
     [self stopWaitingForConnectionRestored];
 }
 
@@ -112,20 +111,21 @@
     [self.view addSubview:self.webView];
 }
 
-#pragma mark - Loading
+#pragma mark - Loading Animations
 
-- (void)startLoading
+- (void)startLoadAnimation
 {
     [self.navigationItem setLeftBarButtonItem:[self statusButtonItem] animated:YES];
     self.navigationItem.title = nil;
 }
 
-- (void)stopLoading
+- (void)stopLoadAnimation
 {
     self.navigationItem.leftBarButtonItem = self.navigationItem.backBarButtonItem;
     self.navigationItem.title  = NSLocalizedString(@"Preview", @"Post Editor / Preview screen title.");
-    [self.webView stopLoading];
 }
+
+#pragma mark - Reachability
 
 - (void)reloadWhenConnectionRestored
 {
@@ -154,6 +154,7 @@
 - (void)webViewDidFinishLoad:(UIWebView *)awebView
 {
     DDLogMethod();
+    [self stopLoadAnimation];
 }
 
 - (void)webView:(UIWebView *)webView didFailLoadWithError:(NSError *)error
@@ -178,6 +179,8 @@
         return;
     }
 
+    [self stopLoadAnimation];
+    
     [self.generator previewRequestFailedWithReason:[NSString stringWithFormat:@"Generic web view error Error. Error code: %d, Error domain: %@", error.code, error.domain]];
 }
 
@@ -224,7 +227,7 @@
 }
 
 - (void)preview:(PostPreviewGenerator *)generator attemptRequest:(NSURLRequest *)request {
-    [self startLoading];
+    [self startLoadAnimation];
     [self.webView loadRequest:request];
     [self.noResultsViewController removeFromView];
 }
--- a/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
+++ b/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
@@ -154,7 +154,6 @@
 - (void)webViewDidFinishLoad:(UIWebView *)awebView
 {
     DDLogMethod();
-    [self stopLoading];
 }
 
 - (void)webView:(UIWebView *)webView didFailLoadWithError:(NSError *)error
@@ -179,7 +178,6 @@
         return;
     }
 
-    [self stopLoading];
     [self.generator previewRequestFailedWithReason:[NSString stringWithFormat:@"Generic web view error Error. Error code: %d, Error domain: %@", error.code, error.domain]];
 }
 
@@ -212,26 +210,6 @@
     }
     
     if (navigationType == UIWebViewNavigationTypeFormSubmitted) {
-        return NO;
-    }
-    
-    // Something is causing our web view to try and load a different URL after the initial URL is already loading.
-    //
-    // Example of first URL: https://diegotest4.wordpress.com/?p=748
-    // Example of second URL: https://public-api.wordpress.com/wp-admin/rest-proxy/#https://diegotest4.wordpress.com
-    //
-    // The second URL is probably loaded by Javascript code or is part of a frame, as it's also loaded by Safari
-    // according to Charles.app (not just in our app in WPiOS).
-    //
-    // The problem is that our code is failing to filter this second URL, and allowing it to load.  This fix
-    // implements a final safeguard in the logic of this method, after all other checks are above pass.
-    //
-    // We won't allow a URL that's not the base document to load at this point.
-    // It could be worth our effort to try removing this check once we migrate our web view to WKWebView.
-    //
-    // More information and context on this issue can be found here: https://git.io/fj5Px
-    //
-    if (![request.URL isEqual:request.mainDocumentURL]) {
         return NO;
     }
     
--- a/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
+++ b/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
@@ -214,6 +214,27 @@
     if (navigationType == UIWebViewNavigationTypeFormSubmitted) {
         return NO;
     }
+    
+    // Something is causing our web view to try and load a different URL after the initial URL is already loading.
+    //
+    // Example of first URL: https://diegotest4.wordpress.com/?p=748
+    // Example of second URL: https://public-api.wordpress.com/wp-admin/rest-proxy/#https://diegotest4.wordpress.com
+    //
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Fyi, @elibud

loremattei:
I'm moving this to 13.2 since 13.1 has been cut. If you want this to make it to 13.1, please move it back, target the `release/13.1` branch and ping me to build a new beta.

rachelmcr:
While logged in to an account with 2FA, whenever I try to view/preview a post I end up on a 2FA screen like this (and I can't get past the 2FA to view the post):



This also happens on `develop` in a simulator, so it isn't specific to the changes in this PR, but I wanted to highlight it in case it could cause issues later on. (I don't see it in the beta release on my device so maybe it's just a simulator-specific issue? I'm not sure and I'm not set up to test dev branches on device right now — let me know if you'd like me to follow up by testing this on device.)

I'll test various scenarios on an account without 2FA to confirm the fix that way.

diegoreymendez:
> I'm not sure and I'm not set up to test dev branches on device right now — let me know if you'd like me to follow up by testing this on device

That's a really good / interesting find!

Unfortunately I can't see this issue on my end, even thought I'm testing on an account with 2FA.  It sounds to me like the site is challenging you to re-enter the 2FA code, to validate your identity.  My initial hunch is that this is a security challenge that responds to some criteria that isn't great in mobile - but would need to digg deeper to know.

In my modest opinion it's in our best interest that we record this issue separately unless we feel it's better to couple it in this PR.  If this is already in develop it could help us assess its impact as well.

It would make sense to test this in a device, if you can reproduce it easily.  I'd believe this is more of a state condition related to the API wanting to re-validate who you are, but again, that's just an initial guess.

rachelmcr:
> It would make sense to test this in a device, if you can reproduce it easily.

I tested this branch on a device and couldn't reproduce it there, so there must be something odd with the state on my simulator. (I'll keep an eye out for this in the wild and open a separate issue if I can reproduce it on a device.)

I confirmed that with a 2FA account on my device the previews are all working on the same site types I tested with a non-2FA account. From a testing perspective this looks good! 🎉

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -214,6 +214,27 @@ - (BOOL)webView:(UIWebView *)awebView
...
+    //
+    // The problem is that our code is failing to filter this second URL, and allowing it to load.  This fix
+    // implements a final safeguard in the logic of this method, after all other checks are above pass.
+    //
+    // We won't allow a URL that's not the base document to load at this point.
+    // It could be worth our effort to try removing this check once we migrate our web view to WKWebView.
+    //
+    // More information and context on this issue can be found here: https://git.io/fj5Px
+    //
+    if (![request.URL isEqual:request.mainDocumentURL]) {
```
shiki: @diegoreymendez I did some more tests and found that this prevents Youtube embeds from being loaded. 

The steps that I did:

1. Create a post in Calypso with a Youtube embed and some images. 
2. Publish it. 
3. View the post in WPiOS.

The images load fine but the Youtube embed didn't.  The post is here: https://yellingyaktest001.home.blog/2019/08/14/videos/. And this is what I see:



The Youtube embed shows up using the preview shown if I click on _Visit_ in the Reader.

I validated that the Youtube embed was canceled by adding a breakpoint on line 235. These URLs were ignored:

- `https://www.youtube.com/embed/oB6b5P6bcZg?version=3&rel=1&fs=1&autohide=2&showsearch=0&showinfo=1&iv_load_policy=1&wmode=transparent`
- `https://widgets.wp.com/likes/master.html?ver=20190321#ver=20190321`

shiki: @rachelmcr Would you mind testing Youtube embeds too? Maybe it's possible that it's just me experiencing this. 😬

diegoreymendez: Awesome feedback.  I'll try to dig deeper into how to make this safe for us.  Eli proposed checking for the specific URL.

rachelmcr: Oh, good catch! I didn't think to test posts with embeds, but when I do there are a large number of embeds that never load (YouTube but also embeds like Slideshare, Spotify, Tumblr, VideoPress). I should note though that (I guess due to the original issue) most of these embeds also don't load for me in the production release.

diegoreymendez: Thanks for that info @rachelmcr!

diegoreymendez:
Thanks to @shiki 's observation I managed to spot that, while my assessment partially right (loading other resources was causing the web view to cancel loading), my fix to this issue wasn't right!  Thanks!

I'm now proposing what I feel is the correct fix for this issue.

That said, @koke do you happen to remember if there was a reason why we were calling `stopLoading` in the delegates?  (I couldn't spot anything too obvious)

diegoreymendez:
I'm also adding you as a reviewer again @rachelmcr - to try and validate that the new fix works fine.  Apologies for the back and forth!

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -154,7 +154,6 @@ - (void)refreshWebView
 - (void)webViewDidFinishLoad:(UIWebView *)awebView
 {
     DDLogMethod();
-    [self stopLoading];
```
shiki: @diegoreymendez This works great for me even on slow connections. However, I think we should still call the `[self stopLoading]` but **not call** the `self.webView stopLoading` part of it:

https://github.com/wordpress-mobile/WordPress-iOS/blob/cc53bb295e3c5759715028bc9666814b42a98fe6/WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m#L123-L128

The other lines of `[self stopLoading]` are for resetting the header. Right now, the header is stuck at _Loading_. This has been going on for a few minutes after everything was loaded:

koke:
> That said, @koke do you happen to remember if there was a reason why we were calling `stopLoading` in the delegates? (I couldn't spot anything too obvious)

I think `stopLoading` originally was only stopping the loading indicator (https://github.com/wordpress-mobile/WordPress-iOS/commit/5fe6b446765aafb2bfc00b4979017d5bccb523e1), but later on it also started calling `stopLoading` on the web view (https://github.com/wordpress-mobile/WordPress-iOS/commit/39b3f04ab97ba0badef139b2150db0b3a1f25218)

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
diegoreymendez: Fixed.

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -95,7 +95,6 @@ - (void)viewDidAppear:(BOOL)animated
 - (void)viewWillDisappear:(BOOL)animated
 {
     [super viewWillDisappear:animated];
-    [self stopLoading];
```
diegoreymendez: Fyi, this was removed on purpose.  During my tests I found that it looks nicer if we let the preview keep loading even as this view disappears.

This can be tested by opening a preview, and switching to another section of the App (using the bottom tab bar).

- With the previous code, once you go back to "My Sites" the preview is still blank.
- With this change, the preview is properly loaded.

In both scenarios the preview is refreshed, but the latter looks much better IMHO.

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
```diff
@@ -112,21 +111,22 @@ - (void)setupWebView
...
 
-- (void)startLoading
+- (void)startLoadAnimation
 {
     [self.navigationItem setLeftBarButtonItem:[self statusButtonItem] animated:YES];
     self.navigationItem.title = nil;
 }
 
-- (void)stopLoading
+- (void)stopLoadAnimation
```
diegoreymendez: I think part of the issue we had is that the previous naming for these methods was confusing.  These methods now have a clearer naming which specifically relates to the loading UI changes / animations.  The idea is to better separate responsibilities.

Any piece of code that wants to make the web view stop loading, can do so directly.

diegoreymendez:
This is ready for review, @shiki.

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
shiki: Good thinking! 👏 

And thank you for explaining.

WordPress/Classes/ViewRelated/Post/PostPreviewViewController.m
shiki: ❤️ Thanks for renaming that!

Issue #12136, opened 2019-07-16
[Auto-upload Published] Safety: Content hash and timeout for offline actions


diegoreymendez:
Parent issue: #12240. 

Came up on: #9555 (comment)

## Goal 

The goal is to avoid auto-uploading posts if the user has not **confirmed** publishing by pressing the _Publish_ button in the editor.

The work should be targeting the master branch, `issue/12240-master-branch`.

## Known Tasks 

- [x] Use content hash code to decide if the user has confirmed publishing/uploading. The example is what @malinajirka prototyped in https://github.com/wordpress-mobile/WordPress-Android/commit/024436d78cf8028247a9f6e62ac7dc21aef78c2d. The code for the content hash code can be seen in this PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/1328.
- [x] Do not auto-upload if the post's modified date is greater than 2 days.

## Matching Android Issue

https://github.com/wordpress-mobile/WordPress-Android/issues/10214

shiki:
Moving this to Backlog. It's planned as part of #12240. 

To note: 1 idea is to use content hash codes like what @malinajirka prototyped in https://github.com/wordpress-mobile/WordPress-Android/commit/024436d78cf8028247a9f6e62ac7dc21aef78c2d.

shiki:
Completed by #12347 and #12352. Thank you, Jan!

PR #12317, opened 2019-08-14
[Auto-upload Published] Cancelation Flow

shiki:
Part of #12240. In #12252, we added auto-uploading of locally published posts. In #12278, we updated the Post List and Notice messages for it. This PR adds the cancelation flow. 

image

Credit goes to @diegoreymendez. Most of this was taken from his work on #12178.  

## What Cancelation Means

In the context automatic uploads, canceling means:

- Leave the changes as local changes
- Do not automatically upload

Canceling should not discard the local changes. In WPAndroid, cancelation is implemented as [turning off the _confirmation flag_](https://github.com/wordpress-mobile/WordPress-Android/blob/1bdae360a9c6213eea84d6bbfc71c3c582a630c6/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadUtils.java#L442-L449). 

## UploadActionUseCase

In this PR, I introduced a new class, `UploadActionUseCase`. Its goal is to encapsulate the business logic of:

- Should a post be auto-uploaded or just remote auto-saved? 
- Can the user cancel the auto-upload? 

It is shared between `PostCoordinator` and Post List classes.

image

I also envision this to be the location of the safety features that we need to add. And perhaps a feature flag to disable auto-uploading. 

## What's Missing

There are still a few parts missing here. I wanted to focus on the main flow of confirmation → upload and confirmation → cancel. Here are some that were not added in this PR:

- Notice should show Cancel button #12319 
- Show notice after canceling #12325 

And for completing #12240, we still need to implement these:

- Safety: Content hash and timeout for offline actions #12136 
- Show Publish button in Post List #12321
- Automatically upload published and draft posts with existing remote #12324 
- Remote auto-save unconfirmed draft and published posts #12322 
- Fix the “Upload failed“ shown after canceling. 

Some are in the checklists of #12240. 

### Stub for Confirmation Property

Jan is working on #12136 to adding a content hash property that will be used to determine whether the user confirmed or canceled the auto-upload. In this PR, I added a stub:

https://github.com/wordpress-mobile/WordPress-iOS/blob/504b9153cc2ed2ac29e9468b746d64919bfc74cc/WordPress/Classes/Services/PostCoordinator%2BUploadActionUseCase.swift#L82

I riddled the code with `#warning()` so I won't forget to remove or replace them. The stub works by adding or removing a `"<C>"` to the `postTitle`. For example, if the user entered `"Llamas"` as the title:

Action | Saved `postTitle`
--------|-------
Confirmed auto-upload        |       `<C> Llamas`
Canceled | `LLamas`

@jklausa I'd like to get your input on how this will integrate with what you've implemented so far for #12136. I'll help you out in integrating the work. 

## Testing

1. Go offline.
2. Create a post and publish it. 
3. Press the _Cancel_ button in the Post List. 

The post should not be auto-uploaded. 

## Reviewing 

Since this is a critical part of the publishing flow, it would be great if 2 developers can review it. Please let me know if you think a sync call explaining what's been done so far would be beneficial. We can arrange that. 

## Release Notes


shiki:
> I did find the filtering issue we discussed via DM (which is most likely not introduced in this PR though)

@diegoreymendez I created a separate issue for it: #12329

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
```diff
@@ -0,0 +1,100 @@
...
+
+        /// Returns what action should be executed when we retry a failed upload.
+        ///
+        /// In some cases, we do not want to automatically upload a post if the user has not
+        /// given explicit confirmation. Users "confirm" automatic uploads by pressing the
+        /// Publish or Update button in the editor.
+        ///
+        /// If we do not receive a confirmation, which can happen if the editor crashed, we will
+        /// try to upload a revision instead.
+        func autoUploadAction(for post: AbstractPost) -> UploadAction {
```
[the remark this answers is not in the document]
shiki: If we end up renaming the class as discussed in [this comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/12317/files#r314378334), I think we can stick with `autoUploadAction(for:)`.

yaelirub:
Hey @shiki, I'm seeing issue with the consistency of the appearance of the cancel option. 
When I create a post by tapping the "+" outside of the editor while offline and publish it and then go to the post list "draft tab", I don't see the option to cancel.
If I switch to a different tab and then come back to the draft tab, the option to cancel then appears.

shiki:
@yaelirub Thanks. This look like Core Data making one's life a misery. I'll take a look.

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
```diff
@@ -0,0 +1,95 @@
...
+            guard post.isFailed,
+                let status = post.status,
+                UploadActionUseCase.allowedStatuses.contains(status),
+                !post.hasRemote() else {
+                return .nothing
+            }
+
+            if post.isLocalDraft || post.confirmedAutoUpload {
```
yaelirub: Should this be a && instead of || ?

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
```diff
@@ -0,0 +1,95 @@
...
+        /// This can be used to determine if the app should show the Cancel button in the Post List.
+        ///
+        /// - SeeAlso: autoUploadAction(for:)
+        func canCancelAutoUpload(of post: AbstractPost) -> Bool {
+            guard autoUploadAction(for: post) == .upload else {
+                return false
+            }
+
+            // Local drafts are always automatically uploaded
+            if post.isLocalDraft {
```
yaelirub: since autoUploadAction will return .upload if post.isLocalDraft || post.confirmedAutoUpload then if isLocalDraft == false then it means that post.confirmedAutoUpload == true. 

So we can just  do `return !post.isLocalDraft`, right?

shiki: You're right! 👍

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /// Temporary method to support old _Retry_ upload functionality.
+        func canRetryUpload(of post: AbstractPost) -> Bool {
+            guard post.isFailed,
+                let status = post.status else {
+                return false
+            }
+
+            return !UploadActionUseCase.allowedStatuses.contains(status)
```
yaelirub: Just want to see if I understand. If the post status is draft or publish we will hide the retry button?

WordPress/Classes/ViewRelated/Post/PostCardCell.swift
```diff
@@ -278,6 +286,10 @@ class PostCardCell: UITableViewCell, ConfigurablePostView {
         retryButton.setImage(Gridicon.iconOfType(.refresh, withSize: CGSize(width: 18, height: 18)), for: .normal)
         retryButton.isHidden = true
 
+        cancelAutoUploadButton.setTitle(NSLocalizedString("Cancel", comment: "Labal for the auto-upload cancelation button in the post list. Tapping will prevent the app from auto-uploading the post."),
```
yaelirub: misspelling: Labal should be Label

WordPress/WordPressTest/Services/PostCoordinatorFailedPostsFetcherTests.swift
```diff
@@ -74,8 +74,8 @@ private extension PostCoordinator.FailedPostsFetcher {
     func getPostsToRetrySync() -> [AbstractPost] {
         var result = [AbstractPost]()
```
yaelirub: Will this test be updated later on to return the result as a dictionary of posts and their actions?

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
shiki: Updated in 17bcaba4d70a3222fcd721e1b8eaad8adc9a3f75. Thanks for spotting that!

WordPress/WordPressTest/Services/PostCoordinatorFailedPostsFetcherTests.swift
shiki: Probably not. There will be more specific tests for the actions in `PostCoordinatorUploadActionUseCaseTests`. The `FailedPostsFetcher` just tests the Core Data query for now and I think that's fine.

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
shiki: Yes. The Retry button will only be shown for those Post status types that we do not currently auto-upload like `publishedPrivate` and `scheduled`. We shouldn't need the Retry button anymore since they will be automatically uploaded.

I'm sorry. In my quest to go breadth-first, implementing the whole flow first for `.published` and `.draft` instead of everything all at once, the old features are still there and can be confusing.

WordPress/Classes/ViewRelated/Post/PostCardCell.swift
shiki: Good eye. Fixed in ae0da70ca240835c98e832d5b02d184e3b4964a7.

yaelirub:
Hey @shiki , I've tested thoroughly and didn't notice anything other than my comment above and Diego's comment. I'm good with merging after Diego's comments are resolved as you see fit.

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
```diff
@@ -0,0 +1,100 @@
...
+        ///
+        /// If the post was published locally, it will be published in the server.
+        case upload
+        /// Upload a revision in the server.
+        case remoteAutoSave
+        case nothing
+    }
+
+    /// Provides business logic for automatic uploads.
+    final class UploadActionUseCase {
```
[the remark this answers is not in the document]
diegoreymendez: The issue I find with the term `UseCase` is that it means: "a specific situation in which a product or service could potentially be used", meaning that it's named after where we need to use this code from, rather than what this code is.  This'll work fine if we only ever use this code for the auto-upload use case, but will look weird if we ever want to use any of it to introspect posts for other reasons.

That said, I don't want to hold back this PR on my disagreement on the naming, so it's ok that you make a final call and move ahead.

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
```diff
@@ -0,0 +1,100 @@
...
+        /// Returns what action should be executed when we retry a failed upload.
+        ///
+        /// In some cases, we do not want to automatically upload a post if the user has not
+        /// given explicit confirmation. Users "confirm" automatic uploads by pressing the
+        /// Publish or Update button in the editor.
+        ///
+        /// If we do not receive a confirmation, which can happen if the editor crashed, we will
+        /// try to upload a revision instead.
+        func autoUploadAction(for post: AbstractPost) -> UploadAction {
+            guard post.isFailed else {
```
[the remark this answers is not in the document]
diegoreymendez: Well, I understand the point as I've done the same in the past.

FWIW to achieve better readability my current opinion is that the conditions can be grouped up in a local variable or private method.  Thinking of something like `canAutoUpload`:

Example with private method:

```swift
guard canAutoUpload(post) else {
    return .nothing
}
```

Example with var:

```swift
let canAutoUpload = post.isFailed,
	                let status = post.status,
	                UploadActionUseCase.allowedStatuses.contains(status),
	                !post.hasRemote()

guard canAutoUpload else {
    return .nothing
}
```

diegoreymendez:
@yaelirub - Really nice find. :)

@shiki - I think this is the change you need to fix the UI issues @yaelirub found: https://github.com/wordpress-mobile/WordPress-iOS/pull/12178/files#diff-f09fce653dd62692afc4dbc2dce721eeL279

WordPress/Classes/Services/PostCoordinator+UploadActionUseCase.swift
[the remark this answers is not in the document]
shiki: Thank you for the explanation. You're right. Now, I'm lost at what to name it. Because I do not agree with `*Logic`. I feel like it's at the same level as `*Utils` which gets abused. 😐 Anyway, in the interest of getting this PR merged to unblock the other pending issues, I used `*Interactor` for now. We can geek out on the name later in the life of the master branch. 😄

Issue #12320, opened 2019-08-14
Dark Mode Master Issue


nheagy:
## Goals
- Add support for iOS 13's [dark mode](https://www.apple.com/ios/ios-13-preview/)

![dark-mode-i2](https://user-images.githubusercontent.com/517257/63059004-0dae8a80-bea3-11e9-8118-c7856b242dfa.png)

### Global

- [x] Add `XCODE11` build flag to support parallel development between Xcode 10 & 11 (#12120)
- [x] Create initial set of preview screenshots
- [x] First-pass at dark mode support on main tab bar screens
- [x] Dark mode versions for badged state of tab bar images
- [x] Fancy alerts have white backgrounds
- [x] Revisions section has white backgrounds
- [x] QuickStart has some white background
- [x] Add new site flow has some white background
- [ ] Update fancy button border style (https://github.com/wordpress-mobile/WordPress-iOS/issues/12320#issuecomment-527646730) #12592
- [x] Change `textPlaceholder` Muriel color to use `tertiaryLabel` on iOS 13.
- [x] Register domain has white backgrounds
- [x] Edit comment is white text on white background.
- [x] Pending comments have a white background.
- [x] Gravatar cropping screen has a white background

### Regressions

- [x] Reader tag area shows visible side gradients: https://github.com/wordpress-mobile/WordPress-iOS/issues/12457
- [x] Web preview has blue toolbar on iOS 12

### NUX
- [x] Dark mode for NUX
- [x] Login NUX button has black spinner while signing in
- [x] Header text on sign up bottom sheet isn't readable
- [x] Header and footer on signup epilogue are white and should be listForeground
- [ ] New account epilogue has a different color behind 'new account' which shouldn't be there https://github.com/wordpress-mobile/WordPress-iOS/issues/12593
- [ ] Modal HUD shown after hitting Continue on New Account screen should use system grey background. https://github.com/wordpress-mobile/WordPress-iOS/issues/12595
- [ ] Activity indicator on secondary NUX buttons appears white-on-white in light mode. [#140](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/140)

### Editor
- [x] Dark mode for editor 
- [x] Dark text on dark background in Aztec HTML mode
- [ ] Post Settings: Tags screen has white to the left and right of the tags field https://github.com/wordpress-mobile/WordPress-iOS/issues/12596
[...]

PR #12323, opened 2019-08-14
Dark Mode: initial commit

nheagy:
Refs #12320 

This is the intial dark mode commit, which lays the foundations of the rest of the changes. This initial set was focussed on creating these major functionality screenshots:

Stats | Reader | Notifications
--- | --- | ---
Simulator Screen Shot - iPhone Xʀ - 2019-08-14 at 19 24 58 | Simulator Screen Shot - iPhone Xʀ - 2019-08-12 at 15 35 30 | Simulator Screen Shot - iPhone Xʀ - 2019-08-14 at 12 36 15

To test:
- Build and run in Xcode 11
- Switch to dark mode (Settings > Developer in the Simulator)
- Check the screens posted above
- Ensure they basically match the mockups. You'll see lots of white backgrounds, but these screens should be more-or-less accurate, with perhaps some remaining issues when scrolling down

Update release notes:


WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
```diff
@@ -230,38 +151,175 @@ extension UIColor {
...
 
-    /// MARK: Muriel colors for buttons
+    /// For icons that are present in a table view, or similar list
+    static var listIcon: UIColor {
+        let fallbackColor = UIColor(light: .neutral(shade: .shade20), dark: .neutral(shade: .shade50))
+        #if XCODE11
+        if #available(iOS 13, *) {
+            return .secondaryLabel
+        } else {
+            return fallbackColor
```
danielebogo: I have a doubt here: because for the dark mode it returns `.secondaryLabel` should the _fallbackColor_ be just `.neutral(shade: .shade20)`? Or the _listIcon_ should be just `UIColor(light: .neutral(shade: .shade20), dark: .neutral(shade: .shade50))`?

WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
```diff
@@ -294,3 +352,35 @@ extension UIColor {
...
+                    return light
+                }
+            }
+        } else {
+            self.init(color: light)
+        }
+        #else
+            // in older versions of iOS, we assume light mode
+            self.init(color: light)
+        #endif
```
danielebogo: I think this can be simplified by removing the 2 _else_ case. You only need to check if it's _XCODE11_ and _iOS 13_. In all the other cases do the same thing.
```
#if XCODE11
        if #available(iOS 13, *) {
            self.init { traitCollection in
                if traitCollection.userInterfaceStyle == .dark {
                    return dark
                } else {
                    return light
                }
            }
        }
#endif
self.init(color: light)
```
You can simplify this also in the colors properties above.

nheagy: Ah ya, I like that a lot better!

nheagy: I wasn't able to use this pattern for `init(light: UIColor, dark: UIColor)` because `self.init` doesn't return, however, it looks great for all the color properties.

nheagy:
@frosty I think this is ready for re-review. Some notes:

> chart bars are lighter
_Pretty_ sure these look the same as before the changes. Are you comparing to `develop`?

> table sections are darker than normal
> action buttons seems slightly lighter than normal

I'm not seeing these, unless I don't perceive them. Make sure you're lauching the app in light mode with Settings > Developer, as not everything is changing correctly dynamically. (I consider full dynamic support a separate problem as it may be a huge can of worms).

nheagy:
@frosty: 

> the main tableview background color has changed? Is this intentional?

Yes! On iOS 13 we're using the system colors when possible. iOS 11 & 12 will continue to use the Muriel colors. That should be what's going on with section headers as well.

That safe for later icon IS weird. I'll look into that _and_ the failing tests ASAP. Thanks!

frosty:
> Yes! On iOS 13 we're using the system colors when possible 😍iOS 11 & 12 will continue to use the Muriel colors. That should be what's going on with section headers as well.

Of course! D'oh. Sorry for the false alarm on that one!

nheagy:
Green lights appeared! Thanks @frosty and @danielebogo!

PR #12326, opened 2019-08-15
Fix for media overlay disappears on action sheet dismissal in Classic editor

### LINKED ISSUE #11937: Image overlay disappears when action sheet is dismissed in Aztec
Spun off of: #11933
And related to: #11749

Steps:
1. Be offline
2. Add image to post in Aztec
3. see failure overlay
4. tap on overlay
5. choose retry or edit in action sheet
6. see overlay disappears when action sheet is dismissed

overlay_dissapear

yaelirub:
- Added a mediaFor(uploadID: String) which searches for the media by upload id within the post object since the mediaCoordinator keeps returning nil for the media that failed to upload

- updated overlay reset method to reset only if the media doesn't have an error. 

- updated the reset overlay method name to communicate that it will only be reset if there is no error

Fixes #11937 

To test:
1. Open the editor on an existing post or start a new post in the classic editor
2. Be offline
3.  Add media to the post
4. See the failure overlay 
5. Tap on overlay
6. Choose discard or tap on title (or other content in the post)
7. See that the overlay is still there
8. Become online
9. Wait till the media is auto uploaded
10. See that the overlay disappears 

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -973,6 +973,15 @@ extension AztecPostViewController {
 
         toolbar.selectItemsMatchingIdentifiers(identifiers.map({ $0.rawValue }))
     }
+    
+    private func mediaFor(uploadID: String) -> Media? {
+        for media in post.media {
+            if media.uploadID == uploadID {
+                return media
+            }
+        }
+        return nil
+    }
 }
 
 
@@ -2477,7 +2486,7 @@ extension AztecPostViewController {
     }
 
     private func handleUploadStarted(attachment: MediaAttachment) {
-        resetMediaAttachmentOverlay(attachment)
+        resetMediaAttachmentOverlayIfNoError(attachment)
         attachment.progress = 0
         richTextView.refresh(attachment, overlayUpdateOnly: true)
     }
@@ -2494,6 +2503,7 @@ extension AztecPostViewController {
             guard let attachment = self.findAttachment(withUploadID: mediaUploadID) else {
                 return
             }
+            resetMediaAttachmentOverlayIfNoError(attachment)
             attachment.uploadID = nil
             attachment.progress = nil
             if let imageAttachment = attachment as? ImageAttachment {
@@ -2617,7 +2627,7 @@ extension AztecPostViewController {
     }
 
     fileprivate func retryFailedMediaUpload(media: Media, attachment: MediaAttachment) {
-        resetMediaAttachmentOverlay(attachment)
+        resetMediaAttachmentOverlayIfNoError(attachment)
         attachment.progress = 0
         richTextView.refresh(attachment)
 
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -2916,10 +2927,14 @@ extension AztecPostViewController {
...
     }
 
-    fileprivate func resetMediaAttachmentOverlay(_ mediaAttachment: MediaAttachment) {
-        mediaAttachment.overlayImage = nil
-        mediaAttachment.message = nil
-        mediaAttachment.shouldHideBorder = false
+    fileprivate func resetMediaAttachmentOverlayIfNoError(_ mediaAttachment: MediaAttachment) {
+        if let uploadID = mediaAttachment.uploadID,
+        let media = self.mediaFor(uploadID: uploadID),
+        media.error == nil {
```
shiki: ⚠️ So, I'm not sure if we need reset only if there's an error. It seems that this is causing a regression. 

1. Edit a post that has images.
2. Select an image. The action sheet should show.
3. Dismiss the action sheet. 

The overlay will not disappear. 

[Here](https://drive.google.com/file/d/1-PiLnH8H-bOf9F49ygA9sRVYkv5dBzat/view?usp=sharing) is a video.

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -973,6 +973,15 @@ extension AztecPostViewController {
 
         toolbar.selectItemsMatchingIdentifiers(identifiers.map({ $0.rawValue }))
     }
+
+    private func mediaFor(uploadID: String) -> Media? {
```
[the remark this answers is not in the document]
yaelirub: >Were you able to figure the reason for this?
I know Diego has been working on figuring out some issues where the mediaCoordinator is being flaky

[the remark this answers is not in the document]
yaelirub: >Also, it's probably worth it to see if we can use findAttachment(withUploadID uploadID: String) -> MediaAttachment? instead.

Why is that better? I was following the way Gutenberg does it https://github.com/wordpress-mobile/WordPress-iOS/blob/347a12fc06e3042b92f379455cfaa4de35d56860/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaInserterHelper.swift#L81

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
yaelirub: > So, I'm not sure if we need reset only if there's an error. It seems that this is causing a regression.

The reset is only if there is *no* error `media.error == nil`
The vidoe doesn't load for some reason. Is it a big file?

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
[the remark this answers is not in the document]
shiki: Oh, not necessarily better. I just thought if they're essentially the same, then we don't need to add another method to maintain. 🙂

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
shiki: >  Is it a big file

It's 16 MB. Sorry. I should learn how to optimize videos soon. 😅

yaelirub:
@shiki , updated and ready for another look

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -2916,10 +2927,19 @@ extension AztecPostViewController {
         restoreInputAssistantItems()
     }
 
-    fileprivate func resetMediaAttachmentOverlay(_ mediaAttachment: MediaAttachment) {
-        mediaAttachment.overlayImage = nil
-        mediaAttachment.message = nil
-        mediaAttachment.shouldHideBorder = false
+    fileprivate func resetMediaAttachmentOverlayIfNoError(_ mediaAttachment: MediaAttachment) {
```
shiki: Should we revert the method name back to `resetMediaAttachmentOverlay`? This new name no longer reflects what the method does.

PR #12332, opened 2019-08-15
Add 'Sign in with Apple' entitlement.

ScoutHarris:
Fixes #n/a

This adds the 'Sign in with Apple' entitlement. It's not used for anything yet, so testing is primarily to ensure the app still works.

To test:
- Open the app in Xcode 11.
  - Verify it runs with no problems.
  - Go to the project > WordPress target > Signing & Capabilities. 
  - Verify the 'Sign in with Apple' entitlement appears (probably at the bottom).
- Run the app in Xcode 10. Verify it runs with no problems.

Update release notes:

### DIFF
--- a/WordPress/WordPress-Alpha.entitlements
+++ b/WordPress/WordPress-Alpha.entitlements
@@ -2,10 +2,6 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>com.apple.developer.applesignin</key>
-	<array>
-		<string>Default</string>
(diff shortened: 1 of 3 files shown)


astralbodies:
I ended up removing the entitlements for the internal builds. I don't think we can set up Sign in with Apple in the enterprise developer portal easily. Well, I think we could, but I'm not sure it's worth the effort and I don't want to do anything to confuse Apple's server-side configuration at this point. We'll need to exclude the button and such from internal builds or at least provide a pop-up saying it's not enabled.

ScoutHarris:
Hey @astralbodies .

> I ended up removing the entitlements for the internal builds. I don't think we can set up Sign in with Apple in the enterprise developer portal easily.

Thank you for fixing that!! Much appreciated!

PR #12344, opened 2019-08-19
Dark mode: Fix no results background color

nheagy:
Refs #12320

This updates the background color set in `configureTableViewColors(view: UIView?)` which will affect several screens, but most notably the no results view.

image

To test:
- open stats or reader
- note that the background color for the no results view is now black in dark mode


nheagy:
Thanks @ScoutHarris!

PR #12345, opened 2019-08-19
Dark mode: fixes for stats: maps and view more

nheagy:
Refs #12320

Fixes two problems in stats: the map colors, and the background color for "view more" rows.

image

To test:
- open stats in dark mode
- ensure view more rows have the proper dark background
- ensure the map looks about right

### Note:
The map doesn't entirely transition dynamically when changed between light and dark mode. I intend to leave that for a future improvement.

Update release notes:


WordPress/Classes/ViewRelated/Stats/Period Stats/Countries/Map/CountriesMapView.swift
```diff
@@ -41,4 +52,24 @@ class CountriesMapView: UIView, NibLoadable {
...
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        colors = mapColors(traits: traitCollection)
+        if let countries = countries {
+            setData(countries)
+        }
+    }
+
+    private func mapColors(traits: UITraitCollection?) -> [UIColor] {
```
danielebogo: The function argument `traits` is never used because the if statement uses the `traitCollection`. I also see the view `traitCollection` is passed as argument above. Can we remove the argument or use traits in the if statement? I don't know if the function needs to accept different traits than the view trait.

nheagy:
Great suggestion @danielebogo! Ready for another look 😀

mattmiklic:
This is looking good. Just one request; let's use `systemGray5` instead of Muriel's Gray 70 there; it will complement the background color and the pink better.

nheagy:
map view
 @mattmiklic here it is with `systemGray5`

mattmiklic:
Looking good. 👍

nheagy:
Thanks @mattmiklic @danielebogo!

PR #12346, opened 2019-08-20
Merge develop into 12240-master-branch

jklausa:
This is just for bookkeeping — merging latest `develop` into the `12240-master-branch`.

My changes for #12136 modify the Core Data stores, so we need the latest changes in here, and without it the diff was completely unreadable.


jklausa:
I think the UI tests are failing because of some changes that are in this branch already and haven't updated the test. I can take a stab at fixing those now, or we can wait until the master branch is almost ready to go and fix all of those in one big swoop — I'm open to doing it either way :)

PR #12352, opened 2019-08-21
Offline Posting: Expire AutoUploads after 48h

jklausa:
This is a second part of #12136 — adding support for expiring the scheduled autouploads after 48hs from confirmation.

Note that since this is using the same Core Data model number as #12347, it might crash on launch if you ran the #12347 branch in your simulator before. You'll need to delete and reinstall the app from the sim/device.
This is expected — I don't want this to create multiple migrations that will have to happen at runtime for our users, and all the migrations in the master branch should be coalesced into one before merging into `develop`.

To test:

Scenario 1 — making sure "young posts" are correctly autouploaded:

1. Go offline
2. Change the date on your Mac/device to yesterday 
3. Compose a post, try to post it
4. Change the date on your Mac/device back to the correct one
5. Go back online
6. Verify that the Post gets correctly uploaded / is marked in the Post list as "will be published next time your device is online

Scenario 2 — expiration

1. Go offline
2. Change the date on your Mac/device to something more than 48hrs in the past
3. Compose a post, try to post it
4. Change the date on your Mac/device back to the correct one
5. Go back online
6. Verify that the Post gets doesn't get automatically uploaded and it's displayed in the post list as "Upload failed"

Update release notes:

### DIFF
--- a/WordPress/Classes/Models/AbstractPost.m
+++ b/WordPress/Classes/Models/AbstractPost.m
@@ -5,6 +5,21 @@
 #import "BasePost.h"
 @import WordPressKit;
 
+@interface AbstractPost ()
+
+/**
+ The following pair of properties is used to confirm that the post we'll be trying to automatically retry uploading,
+ hasn't changed since user has tapped on "confirm", and that we're not suddenly trying to auto-upload a post that the user
+ might have already forgotten about.
+
+ The public-facing counterparts of those is the `shouldAttemptAutoUpload` property.
+ */
+
+@property (nonatomic, strong, nullable) NSString *confirmedChangesHash;
+@property (nonatomic, strong, nullable) NSDate *confirmedChangesTimestamp;
+
+@end
+
 @implementation AbstractPost
 
 @dynamic blog;
@@ -16,6 +31,7 @@
 @dynamic featuredImage;
 @dynamic revisions;
 @dynamic confirmedChangesHash;
+@dynamic confirmedChangesTimestamp;
 
 @synthesize restorableStatus;
 
@@ -412,13 +428,6 @@
     return [self originalIsDraft] && [self dateCreatedIsNilOrEqualToDateModified];
 }
 
-- (BOOL)shouldAttemptAutoUpload {
-    BOOL hashesEqual = self.confirmedChangesHash != nil && ([self.confirmedChangesHash isEqualToString:self.changesConfirmedContentHashValue]);
-
-    return hashesEqual;
-}
-
-
 - (NSString *)authorNameForDisplay
 {
     return [NSString makePlainText:self.author];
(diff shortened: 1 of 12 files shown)


jklausa:
actually, since Jayson mentioned he's drowning in review — @yaelirub, would you mind taking a peek at this?

yaelirub:
Hey @jklausa ,
Following scenario 1, I don't see the right status label:
Today's date is 21 I'm changing the date to 20
>Verify that the Post gets correctly uploaded / is marked in the Post list as "will be published next time your device is online

Instead I see Upload Failed.

WordPress/Classes/Models/AbstractPost.m
```diff
@@ -583,6 +592,32 @@ - (void)markRemoteStatusFailed
...
 
+- (BOOL)shouldAttemptAutoUpload {
+    if (!self.confirmedChangesTimestamp || !self.confirmedChangesHash) {
+        return NO;
+    }
+
+    NSTimeInterval timeDifference = [[NSDate date] timeIntervalSinceDate:self.confirmedChangesTimestamp];
+
+    BOOL timeDifferenceWithinRange = timeDifference <= (60 * 60 * 24 * 2);
+    // I know, I know. This isn't how a date comparison _should_ be done!
```
yaelirub: I know it's pretty obvious but can you add a comment saying something like "Post that is older than 48 hours will not be auto upload) ?
I would personally also get rid of the current comment. And instead just say that this method can potentially be called multiple times.

WordPress/WordPressTest/Services/PostAutoUploadInteractorTests.swift
```diff
@@ -116,7 +116,7 @@ private extension PostCoordinatorUploadActionUseCaseTests {
         }
 
         if confirmedAutoUpload {
-            post.confirmedChangesHash = post.changesConfirmedContentHashValue()
+            post.shouldAttemptAutoUpload = true
```
yaelirub: Any chance we can add tests for the 48h expiration functionality?

yaelirub:
Hey @jklausa ,
Tested following the scenarios and it works as expected.

I noticed 2 issues though:
1. When updating an existing post offline the notice says "Post will be published next time you're online" while the status label says "Upload Failed" and the post doesn't get uploaded when I'm back online.


2. The retry button is not visible but I delete the post and undo then its visible for a split second and then disappears



I also left a couple of neat picky comments

WordPress/Classes/Models/AbstractPost.m
```diff
@@ -583,6 +592,32 @@ - (void)markRemoteStatusFailed
...
+    // However, going thru NSCalendar can get really expensive and this method can potentially be called a lot during
+    // scrolling of a Post List — and for our specific use-case, being slightly innacurate here in terms of
+    // leap seconds or other calendrical oddities doesn't actually matter.
+
+    BOOL hashesEqual = [self.confirmedChangesHash isEqualToString:[self calculateConfirmedChangesContentHash]];
+
+    return hashesEqual && timeDifferenceWithinRange;
+}
+
+- (void)setShouldAttemptAutoUpload:(BOOL)shouldAttemptAutoUpload {
```
shiki: While testing #12358 which uses this as the base branch, I found that the Cancel button in the Post List stopped working. I think this is the cause. We're not using the `shouldAttemptAutoUpload` parameter. 🙂

jklausa: 🤦‍♂ I don't know how I missed that.

shiki:
Just for clarity, my concern in https://github.com/wordpress-mobile/WordPress-iOS/pull/12352#pullrequestreview-278425386 seems to be fixed. I didn't look at the other parts of this PR. I'll let @yaelirub finish the review. 🙂

PR #12358, opened 2019-08-21
Add Cancel button in notice shown after failing to publish a post

### LINKED ISSUE #12319: [Auto-upload Published] Notice should show Cancel button
📣 Part of #12240. This is a continuation of the #12240 PR and requires that to be merged first. Any fixes should target the `issue/12240-master-branch` branch. 

In #12240, we added a _Cancel_ button on the Post List for locally published posts. To follow the spec described in #12240, the notice should be changed so that it will show a _Cancel_ button instead of _Retry_. Pressing the button should have the same effect as the one in the Post List.

shiki:
Closes #12319.

⚠️ This is targeting the branch of #12136. That has to be merged first.  

This adds the _Cancel_ button for the notice that is shown after publishing while offline. 

Before | After 
--------|-------
retry     |   cancel button
    
## Testing

1. Go offline.
2. Create a post and publish it.
3. A notice should be shown with a _Cancel_ button.

Confirm that tapping on the _Cancel_ button will prevent the app from auto-uploading the post. 

## Reviewing 

## Submitter Checklist

## Tasks

- [ ] #12136 Must be merged first. After that, change the target branch to `issue/12240-master-branch`

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -244,6 +244,20 @@ class PostCoordinator: NSObject {
             try? post.managedObjectContext?.save()
         }
     }
+
+    /// Cancel active and pending automatic uploads of the post.
+    func cancelAutoUploadOf(_ post: AbstractPost) {
+        cancelAnyPendingSaveOf(post: post)
+
+        // nil-ing this out will make the `shouldAttemptAutoUpload` property to return false.
+        post.confirmedChangesHash = nil
+
+        let moc = post.managedObjectContext
+
+        moc?.perform {
+            try? moc?.save()
+        }
+    }
 }
 
 // MARK: - Automatic Uploads
@@ -267,20 +281,6 @@ extension PostCoordinator: Uploader {
                     return
                 }
             }
-        }
-    }
-
-    /// Cancel active and pending automatic uploads of the post.
-    func cancelAutoUploadOf(_ post: AbstractPost) {
-        cancelAnyPendingSaveOf(post: post)
-
-        // nil-ing this out will make the `shouldAttemptAutoUpload` property to return false.
-        post.confirmedChangesHash = nil
-
-        let moc = post.managedObjectContext
-
-        moc?.perform {
-            try? moc?.save()
         }
     }
 }
--- a/WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
@@ -42,13 +42,20 @@ struct PostNoticeViewModel {
     }
 
     private var failureNotice: Notice {
+        let failureAction = self.failureAction
+
         return Notice(title: failureTitle,
                       message: message,
                       feedbackType: .error,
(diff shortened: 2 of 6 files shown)


WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -244,6 +244,20 @@ class PostCoordinator: NSObject {
             try? post.managedObjectContext?.save()
         }
     }
+
+    /// Cancel active and pending automatic uploads of the post.
```
shiki: I just moved this so it can be overridden in mocks.

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -244,6 +244,20 @@ class PostCoordinator: NSObject {
...
+    func cancelAutoUploadOf(_ post: AbstractPost) {
+        cancelAnyPendingSaveOf(post: post)
+
+        // nil-ing this out will make the `shouldAttemptAutoUpload` property to return false.
```
yaelirub: This was updated in Jans PR: https://github.com/wordpress-mobile/WordPress-iOS/blob/0e42cff404559bcc9bcc0b759f6ed08ea339fec0/WordPress/Classes/Services/PostCoordinator.swift#L277

WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
```diff
@@ -5,7 +5,13 @@ enum PostNoticeUserInfoKey {
 }
 
 struct PostNoticeViewModel {
-    let post: AbstractPost
+    private let post: AbstractPost
+    private let postCoordinator: PostCoordinator
+
+    init(post: AbstractPost, postCoordinator: PostCoordinator = PostCoordinator.shared) {
```
yaelirub: Interesting. Why pass the postCoordinator as a parameter? Won't we always use PostCoordinator.shared?

WordPress/Classes/Services/PostCoordinator.swift
shiki: Yeah, I figured there are going to be some conflicts. 😄

WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
shiki: It's for dependency injection in the tests. In the tests, we pass a mocked `PostCoordinator`.

WordPress/Classes/Services/PostCoordinator.swift
shiki: I was wrong! I picked the wrong branch. I should have been targeting #12352. Thank you for the heads up. Marking this as Not Ready for Review for now to avoid the confusion.

shiki:
@yaelirub I've updated the base branch since #12352 was merged. I also fixed the conflicts. 

Ready for another look!

yaelirub:
Hey @shiki , Pressing on the cancel button doesn't seem to work for me

jklausa:
Sorry for being late to this! I also took it out for spin and looked at the code and can confirm the 👍

PR #12360, opened 2019-08-22
[SIWA] Using methods to show new account epilogue

ScoutHarris:
Fixes #n/a

WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/182
WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/117

This uses the changes in both pods listed above to show the epilogue after a WP account is created with Apple ID credentials.

**NOTE**: The endpoint isn't updated yet, so account creation will fail. Thus the epilogue doesn't actually show. But it does use the same logic as Google, so.... 🤞 .

To test:
There's not much additional to test, it's mostly making sure the code looks sane. But, to be sure nothing broke:

Run the app in Xcode 11.
- Log out if necessary.
- Select Log In > Continue with Apple.
- Walk through the process.
- After your password is entered, since account creation doesn't work yet:
  - The view will be dismissed.
  - A failure error will appear in the console. Ex:
`Apple Authenticator: createWPComUserWithApple failure. error: Error Domain=WordPressKit.WordPressComRestApiError Code=7 "The given service is not valid."`

Run the app in Xcode 10. Verify the app runs and does not show the 'Continue with Apple' login button.
Update release notes:


WordPress/Classes/ViewRelated/NUX/SignupEpilogueViewController.swift
```diff
@@ -73,7 +73,18 @@ class SignupEpilogueViewController: NUXViewController {
     // MARK: - analytics
 
     private func tracksProperties() -> [AnyHashable: Any] {
-        let source = socialService != nil ? "google" : "email"
+        let source: String = {
+            guard let service = socialService else {
+                return "email"
+            }
+            switch service {
```
frosty: You could just return `service.rawValue` here

ScoutHarris: Actually, `SocialService` does not have a `rawValue` (you may be thinking of `SocialServiceName`, which does).

frosty: D'oh, thanks :)

PR #12371, opened 2019-08-22
While offline, attach media even after upgrading the app

### LINKED ISSUE #12143: While offline, media is not attached after upgrading the app
This was found in #12060. 

## Expected behavior

Media is attached to the post when the app is relaunched. 

## Actual behavior

This will be easier to reproduce if running the app from Xcode. 

1. Go offline.
2. Run the app in Xcode.
3. Create a post with an image.
4. Exit the Editor and save the draft.
5. Stop app by stopping Xcode. 
6. Run the app in Xcode again.
7. Open the post again. Notice that the image will no longer be shown, just the placeholder. 

## Findings

The main issue here is the usage of full path URLs. 

After we attach an image and we save the post, the content will look like this:

```html
<p>
    <img
        src="file:///var/mobile/Containers/Data/Application/92BED2B8-0498-41DF-BE0B-D0B24F3EAC9F/Library/Caches/Media/thumbnail-p3-2436x2436.jpeg"
        class="size-full"
        data-wp_upload_id="x-coredata://D5CF762D-7FB1-450D-B5A2-989DC0FC70B9/Media/p3"
    >
</p>
```

Notice that the `` `src` is the full path to the image in the file system. The issue here is that we cannot fully rely on the application data path represented by this part of that path:

```
/var/mobile/Containers/Data/Application/92BED2B8-0498-41DF-BE0B-D0B24F3EAC9F/
```

That UUID folder will **change** every time we run the app on Xcode. For example, when I run the app in Xcode again and create a new draft, the post content will look like this:

```html
<p>
    <img
        src="file:///var/mobile/Containers/Data/Application/6B2A2AF7-40B1-4ACB-BE59-A4D0095B288F/Library/Caches/Media/thumbnail-p4-2436x2436.jpeg"
        class="size-full"
        data-wp_upload_id="x-coredata://D5CF762D-7FB1-450D-B5A2-989DC0FC70B9/Media/p4"
    >
</p>
```

Notice that the `src` is pointing to a different base path with a different UUID folder. This is the reason why the image is never shown again when the user reopens the post. Contrary to my previous guess, the image file was not deleted, the base folder (UUID) was just renamed to a different one. We were just pointing to the old file path. 

Based on some light testing, it looks like this happens when the user **upgrades the app** from the App Store too. This does not happen when the user just force closes the app or the app crashes.

### References

These observations are corroborated by these references:

- [Changes To App Containers In iOS 8](https://developer.apple.com/library/archive/technotes/tn2406/_index.html) -- mentions to always use `-URLsForDirectory:inDomains:` when locating a file.
- [Xcode 6 keeps renaming my app's directory in iOS8 simulator after each run](https://stackoverflow.com/questions/25233386/xcode-6-keeps-renaming-my-apps-directory-in-ios8-simulator-after-each-run)
- [Xcode changes location of applications folder on build](https://stackoverflow.com/questions/25233386/xcode-6-keeps-renaming-my-apps-directory-in-ios8-simulator-after-each-run)

### Possible Solution

It looks like the right solution would be to not rely on the full path and use a relative path. For example, for this post content:

```html
<img
    src="file:///var/mobile/Containers/Data/Application/6B2A2AF7-40B1-4ACB-BE59-A4D0095B288F/Library/Caches/Media/thumbnail-p4-2436x2436.jpeg"
>
```

We will replace the `src` with:

```html
<img src="file://Library/Caches/Media/thumbnail-p4-2436x2436.jpeg" >
```

We will then convert this back to the full path when the post is reopened in the Editor. This will guarantee that we will be using the current application data path. This type of reference will survive across app upgrades.

It does not stop here though. It looks like the Post List also relies on the full path to show the image by using `Post.pathForDisplayImage` which is filled in by `updatePathForDisplayImageBasedOnContent`:

https://github.com/wordpress-mobile/WordPress-iOS/blob/4b71c77527ce91d39b898690275467dc997503d0/WordPress/Classes/Models/AbstractPost.m#L638-L654

There might be more so we'll have to make adjustments to those places too.

leandroalonso:
Fixes #12143 

### To test:

1. Go offline
2. Run the app in Xcode
3. Create a post with an image
4. Exit the Editor and save the draft
5. Stop app by stopping Xcode
6. Run the app in Xcode again
7. Check if the image thumbnail appears
8. Open the post again, check that all images are visible

### Additional information

* Before editing a post, the post's content is updated to reflect the correct local URLs - this will only happen with posts with `isFailed = true`
* This also fixes non-image assets, like videos for example
* The `pathForDisplayImage` property now always return a correct path for a local image, otherwise, it will return `nil`

Update release notes:


WordPress/Classes/Extensions/AbstractPost+fixedMediaLocalURL.swift
```diff
@@ -0,0 +1,42 @@
...
+
+            extractFilenameAndAbsolutePath(from: media)?.forEach { filename, absolutePath in
+                content = content.replacingMatches(of: "src\\s*=\\s*\"(file://.[^\"]+?/\(NSRegularExpression.escapedPattern(for: filename)))\"", with: "src=\"\(absolutePath)\"")
+            }
+        }
+
+        self.content = content
+    }
+
+    private func extractFilenameAndAbsolutePath(from media: Media) -> [(String, String)]? {
```
jklausa: The method signature here could use a touch-up —  there's no way for this method to actually ever return `nil`, so can we get rid of the `Optional` here?

Additionally, I'd love to have the tuple parameters be named, like `-> [(localURL: String, absoluteLocalURL: String` or so.

WordPress/Classes/Extensions/AbstractPost+fixedMediaLocalURL.swift
```diff
@@ -0,0 +1,42 @@
...
+        var namesAndPaths: [(String, String)] = []
```
jklausa: This is also misleadingly named — AFAICT it doesn't store any names?

WordPress/Classes/Extensions/AbstractPost+fixedMediaLocalURL.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+extension AbstractPost {
+    /// If a post is in the failed state it can contain references to local files.
+    /// When updating the app through the App Store the local paths can change.
+    /// This will fix any outdated local path with the correct one.
+    ///
+    func fixedMediaLocalURL() {
```
jklausa: Grammar nitpick: "fix**ed**" implies to me that it would fix something, and return a corrected version as a return value.

This method however, doesn't return any values and acts on the object itself — I think a better name would be `fixMediaLocalURL()` (or even `fixLocalMediaURLs()`, I think that reads easier)

WordPress/Classes/Extensions/URL+Helpers.swift
```diff
@@ -145,6 +145,29 @@ extension URL {
         let components = pathComponents.filter({ $0 != "/" })
         return components.count == 4 && hasWordPressDotComHostname
     }
+
+    /// In case of an app migration, the UUID of local paths can change.
+    /// This method returns the correct path for a given file URL.
+    ///
+    func fixedMediaLocalURL() -> URL? {
```
jklausa: It looks like the only place we're using this (apart from tests) is in the override for `BasePost.featuredImageURL()` — and the logic is also _very similar_ to `-[Media absoluteURLForLocalPath:  cacheDirectory:]`.

If it's only used in one place, I'd rather keep in there, instead of polluting the namespace of `URL` globally. What'd you think?

WordPress/Classes/Extensions/AbstractPost+fixedMediaLocalURL.swift
```diff
@@ -0,0 +1,42 @@
...
+            return
+        }
+
+        media.forEach { media in
+            guard !media.hasRemote else {
+                return
+            }
+
+            extractFilenameAndAbsolutePath(from: media)?.forEach { filename, absolutePath in
+                content = content.replacingMatches(of: "src\\s*=\\s*\"(file://.[^\"]+?/\(NSRegularExpression.escapedPattern(for: filename)))\"", with: "src=\"\(absolutePath)\"")
```
jklausa: I'm suuuuuper worried about the use of regexes here. I tried reading this and I _think_ it's correct, but I still don't like it.

There's some precedent in the app on how to handle this, the `PostCoordinator` does something really similar for when a `Media` objects gets correctly uploaded, and it's piggy-backing on the `HTMLProcessor` from `Aztec`: https://github.com/wordpress-mobile/WordPress-iOS/blob/b71fc07ee9a9345ec24dbaa08b931a618402f021/WordPress/Classes/Services/PostCoordinator.swift#L184-L213

It's much more verbose than your option, but it's been extensively used elsewhere already, making me much confident in it being reliable.

WordPress/Classes/Models/BasePost.swift
```diff
@@ -34,6 +34,21 @@ extension BasePost {
         case trash = "trash"
         case deleted = "deleted" // Returned by wpcom REST API when a post is permanently deleted.
     }
+
+    static let pathForDisplayImageKeyPath = "pathForDisplayImage"
+    var pathForDisplayImage: String? {
```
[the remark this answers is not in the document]
jklausa: Is there a reason why we'd only want to do it after an version update? 

Reliably detecting that is a PITA, doing changes to `posts` seems like it could be potentially dangerous when mixed with Core Data issues and seems like a recipe for some race conditions that will haunt us forever.

This is fairly straightfoward bit of string comparisons, which, unless the posts are HUGE and/or have thousands of images, shouldn't take too long — doing it on-demand, when a failed image is loaded into the editor seems like a reasonable enough thing for me.

leandroalonso:
@jklausa the PR is available to review again. Let me know what you think about https://github.com/wordpress-mobile/WordPress-iOS/pull/12371#discussion_r319100401

WordPress/Classes/Extensions/AbstractPost+fixedMediaLocalURL.swift
jklausa: Those seem like good reason to do it the way it's currently done then :)

So, @leandroalonso, I think I'd love for two things to happen:
a) as @diegoreymendez mentioned, splitting out the regex to a separate property/method with explanation of what it does
b) I'd also love to see a comment explaining the trade-off you mentioned above directly in the code, so when someone stumbles upon it in the future, they don't have to track down this PR discussion

after that, I think this is ready to go in :)

PR #12372, opened 2019-08-22
Dark mode: semantic color names v2

nheagy:
Refs #12320 

This PR makes two main changes to the muriel color API:
- **add** method `gray(_: MurielColorShade)` to `UIColor` to explicitly pick a shade of muriel gray
- **refactor** method `neutral(…)` to be explicit that the color values ramp in the opposite way in dark mode. Previously this was done in the `xcassets` file, and I think that was misleading.

image

**Note:**
The inverted values for `neutral` in dark mode does a lot of the heavy lifting for dark mode support, though as we use more specific semantic colors, this is having less impact. Also, The xcassets file still has alternate values for `Blue` and `WordPressBlue` for dark and light modes. Ideally, at some point, color-studio will provide dark mode values for these.

To test:
- fire up the app
- make sure it still looks about right

Update release notes:

### DIFF
--- a/WordPress/Classes/Extensions/Colors and Styles/MurielColor.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/MurielColor.swift
@@ -34,6 +34,7 @@ enum MurielColorShade: Int, CustomStringConvertible {
     case shade70 = 70
     case shade80 = 80
     case shade90 = 90
+    case shade100 = 100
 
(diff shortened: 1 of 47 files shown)


nheagy:
@frosty not intentional! Good catch, fixed.

nheagy:
Thanks @frosty!

PR #12373, opened 2019-08-22
Aztec dark mode

frosty:
Refs #12320. This PR implements dark mode for Aztec.

Simulator Screen Shot - iPhone Xs - 2019-08-22 at 22 24 06

There are a few small changes in light mode due to using the semantic or system colors now. Please let me know if this isn't acceptable!

editor-1

**To test:**

* Build and run in iOS 13 with dark mode enabled. Check the UI in the editor looks good and matches the screenshot above.
* Run in light mode on iOS 13 and check everything looks correct.
* Run on iOS 12 and check everything looks correct.

Update release notes:


iamthomasbishop:
Looking great, @frosty! I think Dark Mode is looking fantastic. My only request on dark mode was that we consider removing the bottom border on the formatting toolbar and/or making them a little more subtle. But this isn't a deal-breaker, it looks fine as it is.

I'm not sure if you'd want to address light mode issues in a separate ticket, but I would consider a couple of changes: 

* The blockquote background on light mode could be changed to a more neutral gray – it looks a little odd using the old blueish gray
* The non-active icons on the formatting toolbar for iOS 12 (this branch) seem a bit too light. We might want to consider darkening those slightly – perhaps `gray 20` or `gray 30`

That's all I've got for now – this is looking great, esp dark mode! Well done!

frosty:
Thanks for the reviews!

@iamthomasbishop:

* I opted to leave the dividers as-is for now, just because we're using the `divider` semantic color. Ideally removing the dividers would be great, but that can be an iteration as it's more involved.
* I also slightly darkened the inactive icons in light mode.

PR #12375, opened 2019-08-23
Displaying publish button in post card 

### LINKED ISSUE #12321: [Auto-upload Published] Show Publish button in Post List
📣 Part of #12240. To decrease the risk of conflicts, this should use the #12317 PR branch as the base branch. Ideally, that branch should be merged first. Any fixes should target the `issue/12240-master-branch` branch. 

In #12317, we added the _Cancel_ button in the Post List. When the user presses the _Cancel_ button, the app will not try to automatically upload the post anymore. The app will then hide the _Cancel_ button and only show _Edit_ and _More_. 

There used to be a _Retry_ button in the middle but I removed it since it seems to be confusing if it is shown after cancelation. 

In WPAndroid, we show a _Publish_ button after canceling.

image

## Proposal

image

## Questions

- [x] The _Publish_ button is part of the spec described in #12240. Should we add this as part of the scope for iOS? Or will the current experience be good enough? 

@osullivanchris Hoping to get your input here. 🙂 

## Decisions 

- 2019-Aug-15 We'll include this in the scope as decided in https://github.com/wordpress-mobile/WordPress-iOS/issues/12321#issuecomment-521570525. 
- 2019-Aug-22 We had a chat about https://github.com/wordpress-mobile/WordPress-iOS/issues/12321#issuecomment-523392568 and we decided that we will be implementing **Option B** described in https://github.com/wordpress-mobile/WordPress-iOS/issues/12321#issuecomment-523097247. 
- 2019-Aug-26 We had some confusion around where to show the Publish button. We decided on having only a maximum of 3 buttons on the Post List. If there are more, the rest, preferably the _Publish_ and _Trash_, should be hidden under _More_. Please see [this illustration](https://user-images.githubusercontent.com/198826/63712173-5e56a980-c7fa-11e9-858e-0f44847d07ff.png).

## Discovered Tasks 

- [x] ~Pressing on the Publish button should mark the post as “confirmed for auto-uploading“. If offline, the post will be automatically uploaded later.~ #12382

yaelirub:
Closes #12321. Please follow the discussion on the issue as all considerations are well described there.

## Changes

In this PR, we added the Publish button for these conditions:

- The post was locally published and was going to be automatically uploaded but the user canceled it. 
- The post is a draft.

### Examples

| Failed locally published |Draft with remote | Local draft 
|--------|-------|------|
IMG_9222        |        IMG_9223 | IMG_9224 |

### Architecture 

If you noticed on the examples above, the Publish and Trash button can now be shown on the card or in the _More_ action sheet. To accommodate this new behavior, we added a data representation of these buttons in `PostCardStatusViewModel`:

https://github.com/wordpress-mobile/WordPress-iOS/blob/4fd80baf39030cfc18d20bc0306900cd67bb138f/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift#L9-L26

The `PostCardStatusViewModel`, specifically its [`buttonGroups`](https://github.com/wordpress-mobile/WordPress-iOS/blob/4fd80baf39030cfc18d20bc0306900cd67bb138f/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift#L124-L192) property, now decides the visibility and placement of the buttons. 

image

### Drafts online

1. Go to drafts tab and see the publish action in the Action Sheet (tapping More)
2. Tap "Publish Now" and see that the post is published and moved to the published list 

### Drafts offline

1. Be offline
2. Create a post
3. Published
4. See the message "Post will be published..."
5. Hit cancel
6. see the publish button
7. Go back online 
8. See that the post can be published by hitting the publish button in action bar

Cut the gif before the last action because I had to enter my passcode.

### Possible Regressions

Since this changes the architecture for when the buttons should be accessible, please watch for regressions on this. The buttons' visibility should stay the same. 

## Known Issues

As described in https://github.com/wordpress-mobile/WordPress-iOS/pull/12375#issuecomment-524126919, the Publish button does not currently work if the user is offline. 

## Reviewing 

## Submitter Checklist

## Tasks 

- [x] ~Remove _Publish Now_ button?~
- [x] Add tests 
- [x] Fix font sizes seem off


yaelirub:
@shiki , I would love you opinion, tests and expertise on this PR. Please consider the following as well:

1. When we hit cancel and then try to publish while offline we get the "No internet connection" alert. because the publish method in PostListViewController checks for connectivity before trying to publish and we definitely need to change that . I wonder what is the right decision here 
A. Show a notice saying "No internet connection" instead of the alert.
B. Try and upload again and then when it fails cause it's still offline, update the post status label to `postWillBePublished` and allow to cancel again?
C. Anything else? (Maybe @osullivanchris has an ideal scenario in mind? and what does Android do?) 

2. I will update the tests accordingly

WordPress/Classes/Services/PostAutoUploadInteractor.swift
```diff
@@ -64,4 +64,14 @@ final class PostAutoUploadInteractor {
 
         return !PostAutoUploadInteractor.allowedStatuses.contains(status)
     }
+
+    /// Returns true if any of the following conditions are true:
+    ///
+    /// * The post is a draft.
+    /// * The post failed to upload and has local changes but the user canceled auto-uploading
+    /// * The upload failed and the user cannot Cancel it anymore. This happens when we reached the maximum number of retries.
+    func canPublish(of post: AbstractPost) -> Bool {
```
shiki: This seems to be working well. 👍 

So, when I was working on `PostAutoUploadInteractor`, I was aiming for it to only contain logic for the auto-uploading flow: 

- What actions are performed
- If the action can be canceled

You can see that `canRetryUpload(of:)` doesn't fit this. Because, as described in the comment, it is supposed to be temporary. 😐 I figured this was okay for now since it's supposed to be removed and it was easier to find. But maybe that was bad judgement on my part. I could have used `#warning()` instead.

I believe `canPublish(of:)` is also the same. It's not really part of auto-upload. It uses the `PostAutoUploadInteractor` logic to derive its complete business logic. With this, we should just keep it inside `PostCardStatusViewModel`:

```swift
var canPublish: Bool {
    let isNotCancelablePublishedWithFailedToUploadChanges: Bool = post.isFailed && post.hasLocalChanges() && !canCancelAutoUpload(of: post)
    return post.isDraft() || isNotCancelablePublishedWithFailedToUploadChanges
}
```

yaelirub: I''l move it to the view model.

yaelirub:
@shiki , updated and ready for another look.
For drafts, I have removed the more option and added a trash option instead as all other options are revealed in the action bar already.

osullivanchris:
@yaelirub @shiki 
Hey, I don't think we ever have 4 post actions visible. It looks tight for space and will create problems in other languages.

From my conversations with @shiki 

1. My original intention was to have iOS replicate Android functionality. I didn't want to add any new UI for this task, I just wanted to make sure the post behaved like a draft again once 'cancelling' occurred. 
2. We thought iOS did not have the publish action anywhere in the list of cards, so we couldn't replicate the functionality.
3. After @shiki had a look around we found out that user success team had already taken care of that and added 'Publish now' action in the latest internal build. In Android the 'Publish' button is directly available on the card, rather than in the 'more' menu. That's probably what I would have designed on iOS if it didn't exist at all. But I'm ok with it being in the 'more menu' as long as its there. I trust they have made the right call about which action occurs where. I chatted to @iamthomasbishop and he explained that 'View' was a more common action and 'Publish' a one-off, thus why we would have 'View' available on the card and 'Publish now' in the more section.
4. So I am happy with what already was built and think it covered our needs. I would keep the 'publish now' action in the more menu and not add a new one on the card. 

There might have been a gap in communication.  So I hope this clears it up and its not too much wasted effort / additional work to resolve.

cc @frosty

osullivanchris:
@shiki  could you post a gif or screenshots with the updated changes?

WordPress/WordPressTest/ViewRelated/Post/Views/PostCardStatusViewModelTests.swift
```diff
@@ -0,0 +1,51 @@
...
+            // Local published draft with confirmed auto-upload.
+            PostBuilder(context).published().with(remoteStatus: .failed).confirmedAutoUpload().build():
+                ButtonGroups(primary: [.edit, .cancelAutoUpload, .more], secondary: [.moveToDraft, .trash]),
+            // Local published draft with canceled auto-upload.
+            PostBuilder(context).published().with(remoteStatus: .failed).build():
+                ButtonGroups(primary: [.edit, .publish, .more], secondary: [.moveToDraft, .trash]),
+            // Published post
+            PostBuilder(context).published().withRemote().build():
+                ButtonGroups(primary: [.edit, .view, .more], secondary: [.stats, .moveToDraft, .trash]),
+        ]
```
leandroalonso: I believe we can replace those comments in favor of multiple tests methods, what do you think?

This will make explicit which case is failing and why.

leandroalonso:
@shiki everything looks good to me, I just made some minor comments. Let me know what you think. :)

WordPress/Classes/ViewRelated/Post/PostActionSheet.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
+import AutomatticTracks
 
 @objc protocol PostActionSheetDelegate {
-    func showActionSheet(_ post: AbstractPost, from view: UIView)
+    func showActionSheet(_ postCardStatusViewModel: PostCardStatusViewModel, from view: UIView)
```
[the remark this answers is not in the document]
leandroalonso: Oh, I haven't noticed that this was pointing to the Offline master branch.

Sorry about that. I agree.

WordPress/WordPressTest/ViewRelated/Post/Views/PostCardStatusViewModelTests.swift
leandroalonso: Ok! My only concern is with tests that could possibly involve more logic and make it harder to understand them. We don't want to write tests for our tests, right? 🤣 

Anyway, this case seems pretty simple.

WordPress/WordPressTest/ViewRelated/Post/Views/PostCardStatusViewModelTests.swift
```diff
@@ -0,0 +1,85 @@
...
+
+    override func tearDown() {
+        context = nil
+        contextManager = nil
+        super.tearDown()
+    }
+
+    func testExpectedButtonGroupsForVariousPostAttributeCombinations() {
+        // Arrange
+        print(String(describing: PostCardStatusViewModel.Button.edit))
```
leandroalonso: This `print` is necessary?

shiki: Removed in 76e470f2929814e98eb51bcba0999d42890e2bd7

PR #12376, opened 2019-08-23
Update comments swipe actions to UISwipeActions

### LINKED ISSUE #12212: Comments: Swipe to delete has no title
### Expected behavior

When I swipe on a table cell that has custom swipe actions, I expect to see a title for each action so I know what it does.

### Actual behavior

When you swipe a row in the Comments list, there's a swipe to delete action but it's unlabelled:

image

* Go to the Comments section of the app and swipe a row

lorenzolewis:
Fixes #12212 

It looks like there are swipe actions in the code but the swipe option wasn't working correctly (tested against iOS 12.4 and 11.4).

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
@@ -251,57 +251,50 @@ static NSString *CommentsLayoutIdentifier                       = @"CommentsLayo
 
 - (BOOL)tableView:(UITableView *)tableView canEditRowAtIndexPath:(NSIndexPath *)indexPath
 {
-    return NO;
+    return YES;
 }
 
-- (UITableViewCellEditingStyle)tableView:(UITableView *)tableView editingStyleForRowAtIndexPath:(NSIndexPath *)indexPath
-{
-    return UITableViewCellEditingStyleDelete;
-}
-
-- (NSArray *)tableView:(UITableView *)tableView editActionsForRowAtIndexPath:(NSIndexPath *)indexPath
-{
+- (UISwipeActionsConfiguration *)tableView:(UITableView *)tableView trailingSwipeActionsConfigurationForRowAtIndexPath:(NSIndexPath *)indexPath {
+    
     Comment *comment = [self.tableViewHandler.resultsController objectAtIndexPath:indexPath];
-    NSMutableArray *actions = [NSMutableArray array];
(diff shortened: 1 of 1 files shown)


lorenzolewis:
I think I got most of it converted over, I'm having issues getting the event handlers to work (I think it's because the old buttons use UITableViewRowAction whereas UISwipeAction uses UIContextualActionHandler. I'm pretty rough in Objective C so I could use some help for that part.

lorenzolewis:
Alright, I think it's all working again. First time in Objective C so that was an adventure

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -254,54 +254,50 @@ - (BOOL)tableView:(UITableView *)tableView canEditRowAtIndexPath:(NSIndexPath *)
...
+            }];
+            completionHandler(YES);
+        }];
         
         approve.backgroundColor = [UIColor murielPrimary];
         [actions addObject:approve];
     }
     
-    return actions;
+    UISwipeActionsConfiguration *swipeActions = [UISwipeActionsConfiguration configurationWithActions:actions];
```
frosty: Could we set `performsFirstActionWithFullSwipe` to `NO` on the `swipeActions`? This would prevent the user from accidentally trashing a comment by swiping too far or too quickly, and make it a more explicit action.

lorenzolewis:
No problem, @frosty, good call out. Went ahead and added in the performsFirstActionWithFullSwipe line and tested it out.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -297,6 +297,7 @@ - (UISwipeActionsConfiguration *)tableView:(UITableView *)tableView trailingSwip
     }
     
     UISwipeActionsConfiguration *swipeActions = [UISwipeActionsConfiguration configurationWithActions:actions];
+    swipeActions.performsFirstActionWithFullSwipe = FALSE;
```
frosty: Sorry, one last request – in Objective-C, it's most common to write `NO` instead of `FALSE`. If you could update this, I'll get it merged in :)

frosty:
Great! Just one last small comment!

lorenzolewis:
@frosty noted for the future and fixed for now :)

Issue #7886, opened 2017-10-02
Duplicate Entries In “My Sites”


khaykov:
From mobilerequests:

> I’m getting multiple entries for each of my sites. Version 8.5.0.20170929 on iOS 11.0. This has been going on for a while though. I haven’t cleared the cache or anything so that this could be better debugged.

<img src="https://user-images.githubusercontent.com/728822/31061627-3114f972-a75e-11e7-8b84-cd517c732f81.png" width="270" alt="mysites">

elibud:
From the OP
> I deleted the app and reinstalled it and all is well now.

thehenrybyrd:
We've got another mention of this is #2073587-z. 
Looks like it's probably the same issue as p4a5px-2fv-p2 (seeing some similar logs).

fresatomica:
Another one in #2239935-zen

designsimply:
Tested and confirmed using WP Internal 13.1.0.20190819 that I can see duplicate sites in My Sites on my iPad.

<img src="https://user-images.githubusercontent.com/1119271/63449234-e99bfd80-c3fc-11e9-9396-8929a587a3ab.PNG" width="400" alt="IMG_0161">

It's easy to see in a search too:

<img src="https://user-images.githubusercontent.com/1119271/63449260-fa4c7380-c3fc-11e9-8f58-d94d136bacf2.PNG" width="400" alt="IMG_0159">

Deleting and re-installing the app makes the problem go away. I can't remember the exact timeline, but it's possible this started showing up after upgrading to iOS 12.4.

* App version: 13.1 beta
* Country Code: us
* Device model: iPad Mini 4 (WiFi)
* OS: iOS 12.4

designsimply:
I also noticed that, when sites were showing up as duplicate in the My Sites list, the list in Reader > Followed Sites > Manage also showed duplicates. Delete+reinstall made the problem go away.

<img src="https://user-images.githubusercontent.com/1119271/63450209-bce8e580-c3fe-11e9-8df7-66045e77bc29.PNG" width="400" alt="IMG_0156">
<sup>Tested with WP Internal 13.1.0.20190819 on iPad Mini 4 (WiFi) (iPad5,1) iOS 12.4.</sup>

jklausa:
@designsimply did you have to log-in again to the app when updating to 13.0 / 13.1?

designsimply:
[...]

PR #12377, opened 2019-08-23
De-duplicate blogs after syncing

koke:
This doesn't prevent the site duplication from happening, but if it has happened it will clean it up in the next blog sync

To help with #7886

To test:

To set up the scene, check out `release/13.1` and run the app, but before set up a breakpoint in `-[BlogService updateBlogWithRemoteBlog:account:]` so it always creates a duplicate when syncing:

Screen Shot 2019-08-23 at 11 32 52

Then go to the sites list and pull to refresh until you start seeing duplicates.

Add some local drafts in different blogs by creating new posts and relaunching the app before they get saved on the server.

localDrafts

Now checkout `fix/deduplicate-blogs` and run the app. Go to the sites list and wait for a bit, it should sync automatically. If not, pull to refresh. The duplicates should be gone. If you go to the posts list, every local draft should be there.

localDraftsRecovered

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/BlogService+Deduplicate.swift
+++ b/WordPress/Classes/Services/BlogService+Deduplicate.swift
@@ -13,9 +13,15 @@ extension BlogService {
     @objc(deduplicateBlogsForAccount:)
     func deduplicateBlogs(for account: WPAccount) {
         // Group all the account blogs by ID so it's easier to find duplicates
-        let blogsById = Dictionary(grouping: account.blogs, by: { $0.dotComID })
+        let blogsById = Dictionary(grouping: account.blogs, by: { $0.dotComID?.intValue ?? 0 })
         // For any group with more than one blog, remove duplicates
-        for (_, group) in blogsById where group.count > 1 {
+        for (blogID, group) in blogsById where group.count > 1 {
+            assert(blogID > 0, "There should not be a Blog without ID if it has an account")
+            guard blogID > 0 else {
+                DDLogError("Found one or more WordPress.com blogs without ID, skipping de-duplication")
+                continue
+            }
+            DDLogWarn("Found \(group.count - 1) duplicates for blog with ID \(blogID)")
             deduplicate(group: group)
         }
     }
@@ -31,9 +37,11 @@ extension BlogService {
             // If there are other blogs with local drafts, we reassing them to the blog that
             // is not going to be deleted
             for draft in localDrafts(for: blog) {
+                DDLogInfo("Migrating local draft \(draft.postTitle ?? "<Untitled>") to de-duplicated blog")
                 draft.blog = candidate
             }
             // Once the drafts are moved (if any), we can safely delete the duplicate
+            DDLogInfo("Deleting duplicate blog \(blog.logDescription())")
             managedObjectContext.delete(blog)
         }
     }
--- a/WordPress/Classes/Services/BlogService+Deduplicate.swift
+++ b/WordPress/Classes/Services/BlogService+Deduplicate.swift
@@ -27,7 +27,7 @@ extension BlogService {
         let candidate = group[candidateIndex]
 
         // We look through every other blog
-        for (index, blog) in group.enumerated() where index != candidateIndex  {
+        for (index, blog) in group.enumerated() where index != candidateIndex {
             // If there are other blogs with local drafts, we reassing them to the blog that
             // is not going to be deleted
             for draft in localDrafts(for: blog) {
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Services/BlogService+Deduplicate.swift
```diff
@@ -0,0 +1,61 @@
...
+    /// If there's more than one blog in each group with local drafts, those will be reassigned
+    /// to the remaining blog.
+    ///
+    @objc(deduplicateBlogsForAccount:)
+    func deduplicateBlogs(for account: WPAccount) {
+        // Group all the account blogs by ID so it's easier to find duplicates
+        let blogsById = Dictionary(grouping: account.blogs, by: { $0.dotComID?.intValue ?? 0 })
+        // For any group with more than one blog, remove duplicates
+        for (blogID, group) in blogsById where group.count > 1 {
+            assert(blogID > 0, "There should not be a Blog without ID if it has an account")
```
jklausa: Wouldn't this crash in cases where a XML-RPC blog would get duplicated? They will have a `dotComID` of nil, which would get coalesced to `0` in the line above and then trigger this `assert`.

jklausa: Ah, but this is based on `Account` anyway, so they have to be `REST` blogs anyway. Nevermind.

WordPress/Classes/Services/BlogService+Deduplicate.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private func deduplicate(group: [Blog]) {
+        // If there's a blog with local drafts, we'll preserve that one, otherwise we pick up the first
+        // since we don't really care which blog to pick
+        let candidateIndex = group.firstIndex(where: { !localDrafts(for: $0).isEmpty }) ?? 0
+        let candidate = group[candidateIndex]
+
+        // We look through every other blog
+        for (index, blog) in group.enumerated() where index != candidateIndex {
```
jklausa: I had no idea you could do a `where` together with `for... in`! That's so cool!

WordPress/Classes/Services/BlogService+Deduplicate.swift
```diff
@@ -0,0 +1,61 @@
...
+            DDLogInfo("Deleting duplicate blog \(blog.logDescription())")
+            managedObjectContext.delete(blog)
+        }
+    }
+
+    private func localDrafts(for blog: Blog) -> [AbstractPost] {
+        // The original predicate from PostService.countPostsWithoutRemote() was:
+        //   "postID = NULL OR postID <= 0"
+        // Swift optionals make things a bit more verbose, but this should be equivalent
+        return blog.posts?.filter({ (post) -> Bool in
```
jklausa: There's a `isDraft` property on `AbstractPost` you might be able to use here, though it uses different logic — not sure which is more applicable.

WordPress/Classes/Services/BlogService.m
```diff
@@ -671,6 +671,17 @@ - (void)mergeBlogs:(NSArray<RemoteBlog *> *)blogs withAccount:(WPAccount *)accou
...
+    /*
+     Sometimes bad things happen and blogs get duplicated. 👭
+     Hopefully we'll fix all the causes and this should never happen again 🤞🤞🤞
+     But even if it never happens again, it has already happened so we need to clean up. 🧹
+     Otherwise, users would have to reinstall the app to get rid of duplicates 🙅‍♀️
+
+     More context here:
+     https://github.com/wordpress-mobile/WordPress-iOS/issues/7886#issuecomment-524221031
+     */
+    [self deduplicateBlogsForAccount:account];
```
jklausa: Should we do it on every `syncBlogsForAccount:` call? Seems like once during an app startup _should_ be enough?

WordPress/Classes/Services/BlogService+Deduplicate.swift
koke: I follow the same logic we use to show an alert warning that you have local drafts when signing out

WordPress/Classes/Services/BlogService.m
koke: We don't really know for sure how blogs get duplicated, and in any case I think it belongs conceptually in the sync method: when you call sync you expect the local list to reflect what's on the server.

The implementation will bail early if it detects no duplicates, so it shouldn't be too much overhead.

I wouldn't mind revisiting this once we have all the other fixes in place though.

PR #12379, opened 2019-08-23
Gutenberg DarkMode

etoledom:
PR for Gutenberg DarkMode

GutenDark

Known Issues: 
- ToolBar button colors are not changed. Will need some extra work.
- The SafeArea border colors on Landscape can get some more ❤️
- Dynamically changing between Dark and Light mode with the editor open, some parts of the editor won't change accordingly (bottom-sheet and a few elements in the toolbar)

All these will be worked on, on future PRs.

To test:
- `rake dependencies`
- Build and run using Xcode 11 (beta)
- Set the device to Dark Appearance
- Open gutenberg editor
- Check that is dark 🌑 

Test for a solved crash:
- Open gutenberg editor
- Close gutenberg editor
- Go to iOS settings and change the appearance mode.
- Go back to WordPress app
- Check that it does not crash (there was a crash at this point before)

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -113,7 +113,8 @@ def gutenberg_dependencies(options)
         'react-native-keyboard-aware-scroll-view',
         'react-native-safe-area',
         'react-native-video',
-        'RNSVG'
+        'RNSVG',
+        'ReactNativeDarkMode',
     ]
     if options[:path]
         podspec_prefix = options[:path]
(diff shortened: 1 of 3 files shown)


etoledom:
Thank you for the review @frosty !

Yes, those issues with the toolbar button colors are the ones I was referring too. It will take some refactoring on the JS side to make the proper change, so the idea is to handle that on another PR.

>  I think it'd be good if image placeholders and dimmed out uploading images were dark too

Good catch! And I agree that it's better for a future PR too.

etoledom:
Thank you!

(Also cc @iamthomasbishop about the image placeholder)

frosty:
@etoledom Do we have @iamthomasbishop's list of changes documented anywhere else (in an issue, maybe)? Just don't want to lose track of these :)

PR #12386, opened 2019-08-24
Dark mode: NUX improvements

frosty:
Refs #12320. This PR implements some improvements to the appearance of the login flow for dark mode.

* Updated primary and secondary button styles
* Fixed invisible text in text fields, and increased padding between icon and text
* Selectively show dark or light Sign in with Apple button based on user interface style

dark-before-after

Shouldn't be noticeable changes in light mode.

**To test:**

* Build and run in iOS 13, light and dark mode, and in iOS 12
* Check things look as expected

Update release notes:

cc @mattmiklic


WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
```diff
@@ -258,6 +258,37 @@ extension UIColor {
...
     static var primaryButtonDownBackground = muriel(color: .accent, .shade80)
     static var primaryButtonDownBorder = muriel(color: .accent, .shade90)
+
+    static var secondaryButtonBackground: UIColor {
+        #if XCODE11
+        if #available(iOS 13, *) {
+            return UIColor(light: .white, dark: .systemGray5)
+        }
+        #endif
+        return UIColor(light: .white, dark: .neutral(.shade80))
```
nheagy: We don't need dark mode colors if this isn't in Xcode 11. You can return `.white` here. Same principle applies to the next three.

frosty:
Thanks @nheagy!

frosty:
@nheagy I fixed the issue you mentioned, and I noticed the NUX epilogue also needed some love so I pushed up some fixes there – would you mind checking that too please?

login-nux

mattmiklic:
> Updated primary and secondary button styles
> Fixed invisible text in text fields, and increased padding between icon and text
> Selectively show dark or light Sign in with Apple button based on user interface style

All of this looks good! There are some additional updates to be made to the prologue and the button styles but I think we can address them in separate PRs.

PR #12389, opened 2019-08-25
Dark mode: tab bar colors

nheagy:
Refs #12320

This fixes the colors of the tab bar icons with badges on them, as well as changes the colors to match the iOS system defaults:

new tab colors

To test:
- check the tab bar icons in both light and dark mode
- check them out with and without badges (I fiddle with `WPTabBarController.m`)

Update release notes:


WordPress/ColorPalette.xcassets/Blue50.colorset/Contents.json
```diff
@@ -27,10 +27,10 @@
       "color" : {
         "color-space" : "srgb",
         "components" : {
-          "red" : "0.322",
```
frosty: Are these changes intentional?

PR #12400, opened 2019-08-27
Issues/7886 bug 3

aerych:
Refs #7886 

This PR addresses "Bug 3" as mentioned by @koke [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/7886#issuecomment-524221031). 

I've reworked the sync logic to sync account details before creating an account and syncing blogs, removing the use-case for the temp user id.  With this change any existing blog should be detected by its username when calling `[AccountService createOrUpdateAccountWithUsername:authToken:]`.

To test:

Scenario A 
- Start logged out. 
- Log in to a wpcom account via magic link or password
- Confirm the log in is successful. 
- Confirm the epilogue shows the correct username and blogs. 
- After logging in confirm that the sites are not duplicated. 

Scenario B
- Start logged out
- Use the self-hosted flow to log in to a self-hosted account that is connected Jetpack connected.
- Go to Stats
- Log in to the connected WPcom account. 
- Confirm the log in is successful. 
- Confirm the epilogue shows the correct username and blogs. 
- After logging in confirm that the sites are not duplicated. 

Scenario C
- Start logged out
- Log into a wpcom account that is NOT connected to Jetpack.
- Add a self-hosted account that is connected to Jetpack via a different wpcom account. 
- Go to Stats for the self-hosted account.
- Log in to the connected WPcom account. 
- Confirm the log in is successful. 
- Confirm the epilogue shows the correct username and blogs. 
- After logging in confirm that the sites are not duplicated. 
- Confirm that the only wpcom blogs shown are the ones belonging to the account that was added first.

Update release notes:

@jklausa could I trouble you with this one?

### DIFF
--- a/WordPress/Classes/Services/AccountService.h
+++ b/WordPress/Classes/Services/AccountService.h
@@ -85,17 +85,6 @@ extern NSNotificationName const WPAccountEmailAndDefaultBlogUpdatedNotification;
 ///-----------------------
 
 /**
- A convenience method for creating an Account from an auth token and remote user.
-
- This method makes a pass through call to `createOrUpdateAccountWithUsername:authToken:`, see it for details.
-
- @param remoteUser a RemoteUser instance.
- @param authToken the OAuth2 token returned by signIntoWordPressDotComWithUsername:authToken:
- @return a WordPress.com `WPAccount` object
- */
-- (WPAccount *)createOrUpdateAccountWithUserDetails:(RemoteUser *)remoteUser authToken:(NSString *)authToken;
-
-/**
  Creates a new WordPress.com account or updates the password if there is a matching account
  
  There can only be one WordPress.com account per username, so if one already exists for the given `username` its password is updated
@@ -145,7 +134,7 @@ extern NSNotificationName const WPAccountEmailAndDefaultBlogUpdatedNotification;
  @param success A success block.
  @param failure A failure block.
  */
-- (void)syncAccountDetailsAndCreateAccount:(NSString *)authToken
+- (void)createOrUpdateAccountWithAuthToken:(NSString *)authToken
                                    success:(void (^)(WPAccount * _Nonnull))success
                                    failure:(void (^)(NSError * _Nonnull))failure;
 
--- a/WordPress/Classes/Services/AccountService.h
+++ b/WordPress/Classes/Services/AccountService.h
@@ -136,6 +136,11 @@ extern NSNotificationName const WPAccountEmailAndDefaultBlogUpdatedNotification;
                             success:(nullable void (^)(void))success
                             failure:(nullable void (^)(NSError *error))failure;
 
+
+- (void)syncAccountDetailsAndCreateAccount:(NSString *)authToken
+                                   success:(void (^)(WPAccount * _Nonnull))success
+                                   failure:(void (^)(NSError * _Nonnull))failure;
+
 /**
  Initializes the WordPress iOS Extensions with the WordPress.com Default Account.
  */
--- a/WordPress/Classes/Services/AccountService.h
+++ b/WordPress/Classes/Services/AccountService.h
@@ -137,6 +137,14 @@ extern NSNotificationName const WPAccountEmailAndDefaultBlogUpdatedNotification;
                             failure:(nullable void (^)(NSError *error))failure;
 
 
+/**
+ Syncs the details for the account associated with the provided auth token, then
+ creates or updates a WPAccount with the synced information.
+
+ @param authToken The auth token associated with the account being created/updated.
+ @param success A success block.
+ @param failure A failure block.
+ */
 - (void)syncAccountDetailsAndCreateAccount:(NSString *)authToken
                                    success:(void (^)(WPAccount * _Nonnull))success
                                    failure:(void (^)(NSError * _Nonnull))failure;
--- a/WordPress/Classes/Services/AccountService.h
+++ b/WordPress/Classes/Services/AccountService.h
@@ -4,6 +4,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @class WPAccount;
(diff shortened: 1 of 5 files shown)


WordPress/Classes/Services/WordPressComSyncService.swift
```diff
@@ -15,9 +15,39 @@ class WordPressComSyncService {
...
     ///
     func syncWPCom(authToken: String, isJetpackLogin: Bool, onSuccess: @escaping (WPAccount) -> Void, onFailure: @escaping (Error) -> Void) {
+        syncAccountDetailsAndCreateAccount(authToken: authToken, onSuccess: { account in
+            self.syncOrAssociateBlogs(account: account, isJetpackLogin: isJetpackLogin, onSuccess: onSuccess, onFailure: onFailure)
+        }, onFailure: { error in
+            onFailure(error)
+        })
+    }
+
+    func syncAccountDetailsAndCreateAccount(authToken: String, onSuccess: @escaping (WPAccount) -> Void, onFailure: @escaping (Error) -> Void) {
```
koke: I wonder if we should have this new sync-and-create method inside `AccountService`. I found surprising to be passing a `RemoteUser` in a `.h` file, and it seems it's a first.

Seeing that this just calls `AccountServiceRemoteREST` and pass the result to another method in `AccountService` makes me think this belongs in there.

WordPress/Classes/Services/AccountService.m
```diff
@@ -187,10 +187,12 @@ - (void)requestVerificationEmail:(void (^)(void))success failure:(void (^)(NSErr
 /// @name Account creation
 ///-----------------------
 
-- (WPAccount *)createOrUpdateAccountWithAuthToken:(NSString *)authToken
+- (WPAccount *)createOrUpdateAccountWithUserDetails:(RemoteUser *)remoteUser authToken:(NSString *)authToken
 {
-    NSString *tempUsername = [[NSUUID UUID] UUIDString];
-    return [self createOrUpdateAccountWithUsername:tempUsername authToken:authToken];
+    NSString *username = remoteUser.username;
+    WPAccount *account = [self createOrUpdateAccountWithUsername:username authToken:authToken];
```
koke: This tries to find an account matching the username/email, which should be fine in most cases, but should we be trying to match the account ID instead/also, which is the only piece of data about the account that won't ever change?

WordPress/Classes/Services/AccountService.m
```diff
@@ -262,6 +267,20 @@ - (WPAccount *)findAccountWithUserID:(NSNumber *)userID
     return [results firstObject];
 }
 
+- (void)syncAccountDetailsAndCreateAccount:(NSString *)authToken
+                                   success:(void (^)(WPAccount * _Nonnull))success
+                                   failure:(void (^)(NSError * _Nonnull))failure
+{
+    WordPressComRestApi *api = [[WordPressComRestApi alloc] initWithOAuthToken:authToken userAgent:[WPUserAgent defaultUserAgent]];
```
jtreanor: Just creating a new `WordPressComRestApi` object directly means it can't be mocked (it will always hit `public-api.wordpress.com`). It should be replaced with this (what we do everywhere else):

```suggestion
    WordPressComRestApi *api = [WordPressComRestApi defaultApiWithOAuthToken:authToken userAgent:[WPUserAgent defaultUserAgent] localeKey:[WordPressComRestApi LocaleKeyDefault]];
```

This allows the mocked base URL to be injected.

WordPress/Classes/Services/AccountService.h
```diff
@@ -84,16 +85,15 @@ extern NSNotificationName const WPAccountEmailAndDefaultBlogUpdatedNotification;
...
+ A convenience method for creating an Account from an auth token and remote user.
 
  This method makes a pass through call to `createOrUpdateAccountWithUsername:authToken:`, see it for details.
 
+ @param remoteUser a RemoteUser instance.
  @param authToken the OAuth2 token returned by signIntoWordPressDotComWithUsername:authToken:
  @return a WordPress.com `WPAccount` object
  */
-- (WPAccount *)createOrUpdateAccountWithAuthToken:(NSString *)authToken;
+- (WPAccount *)createOrUpdateAccountWithUserDetails:(RemoteUser *)remoteUser authToken:(NSString *)authToken;
```
koke: Since this is only called from `AccountService.m` now, I think we can revert this bit on the `.h` file

WordPress/Classes/Services/AccountService.h
```diff
@@ -136,6 +136,19 @@ extern NSNotificationName const WPAccountEmailAndDefaultBlogUpdatedNotification;
...
+
+/**
+ Syncs the details for the account associated with the provided auth token, then
+ creates or updates a WPAccount with the synced information.
+
+ @param authToken The auth token associated with the account being created/updated.
+ @param success A success block.
+ @param failure A failure block.
+ */
+- (void)syncAccountDetailsAndCreateAccount:(NSString *)authToken
```
koke: Nitpicking names, but I'd propose just `createOrUpdateAccountWithAuthToken:success:failure:`. I think this method says "I give you an oAuth2 token, go create an account with that" and it's up to the service to figure out what details it needs to do that.

WordPress/Classes/Services/WordPressComSyncService.swift
```diff
@@ -28,23 +45,21 @@ class WordPressComSyncService {
         }
 
         let onSuccessInternal = {
```
jklausa: I think we can now get rid of the whole property here, it reads a bit confusing to me if it's only just that one line.

aerych: Don't we need the block to wrap the account?

aerych:
Thanks @koke  and @jklausa !

PR #12405, opened 2019-08-27
[Stats Insights] Load Insights asynchronously 

### LINKED ISSUE #12146: [Stats Refresh] Insights: Load overview block independently
On Insights we should load the LPS block first, then all the other blocks

danielebogo:
Fixes #12146

This PR enables the async loading block in Insights.

insights-async

At the moment the app displays the _No Result View_ until the store finishes to load the whole overview. This can take some times and block the user experience.
With the async block loading every table section is now independently handled based on a _row status_ (`StoreFetchingStatus`). If the row status is _loading_ and there's no cached data for that section a _Ghost view_ is displayed instead.

The _No Result View_ will be displayed only if there's no cached data and all the API calls will fail.

Also the chart view has priority in the loading queue. This means the other sections will be loaded as soon the LPS is loaded.

This feature is behind the feature flag `statsAsyncLoading`.

## To test:
- To test the async loading you should open a site that doesn't have any stored data
- Also use a link conditioner to simulate a slow connection
- In `FeatureFlag` return false in `case .statsAsyncLoading:` to test everything works fine as before
- If you want to see all the ghost views enabled, return `return rowBlock()` in `SiteStatsInsightsViewModel`
- Test with iOS 13 and Dark mode to check the new colors

Update release notes:


WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -820,19 +810,90 @@ extension StatsInsightsStore {
...
-            state.fetchingAnnualAndMostPopularTime ||
-            state.fetchingDotComFollowers ||
-            state.fetchingEmailFollowers ||
-            state.fetchingPublicize ||
-            state.fetchingTodaysStats ||
-            state.fetchingPostingActivity ||
-            state.fetchingCommentsInsight ||
-            state.fetchingTagsAndCategories
+        let mirror = Mirror(reflecting: state)
+        return mirror.children.compactMap { $0.value as? StoreFetchingStatus }.first { $0 == .loading } != nil
```
frosty: Could you perhaps add a comment to this code just clarifying what it's doing?

danielebogo:
Hey @frosty ! Thanks for reviewing this! I fixed the bug in the error row and I removed the action button if the NRV reachability is false. Also I added a comment about the reflection used to map the store members.

WordPress/Classes/ViewRelated/Stats/Shared Views/GhostViews/StatsGhostTableViewRows.swift
```diff
@@ -43,6 +43,6 @@ struct StatsGhostChartImmutableRow: StatsRowGhostable {
 
 private enum GhostCellStyle {
     static let muriel = GhostStyle(beatDuration: TimeInterval(0.75),
-                                   beatStartColor: .neutral(shade: .shade0),
-                                   beatEndColor: .neutral(shade: .shade5))
+                                   beatStartColor: .neutral(.shade0),
```
frosty: We should maybe think about defining these in the Muriel UIColor extension – `ghostStart` and `ghostEnd` or something.

danielebogo: @nheagy made it already in [#12415](https://github.com/wordpress-mobile/WordPress-iOS/pull/12415) 😊
I'm going to update this as soon `13.2` will be merged in `develop`. I added a reminder in the Master Issue.

PR #12407, opened 2019-08-27
Update CocoaPods to the 1.7 series for Xcode 11

jkmassel:
It's all in the title 😃

**To Test:**
- Run `rm -rf Pods && bundle install && bundle exec pod install` on this branch, then build and run the app.
- Ensure all tests continue to pass.


WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -9998,39 +10016,131 @@
...
+				"${BUILT_PRODUCTS_DIR}/1PasswordExtension/OnePasswordExtension.framework",
 				"${BUILT_PRODUCTS_DIR}/Alamofire/Alamofire.framework",
+				"${BUILT_PRODUCTS_DIR}/AlamofireNetworkActivityIndicator/AlamofireNetworkActivityIndicator.framework",
+				"${BUILT_PRODUCTS_DIR}/Automattic-Tracks-iOS/AutomatticTracks.framework",
+				"${BUILT_PRODUCTS_DIR}/Charts/Charts.framework",
 				"${BUILT_PRODUCTS_DIR}/CocoaLumberjack/CocoaLumberjack.framework",
+				"${BUILT_PRODUCTS_DIR}/DoubleConversion/DoubleConversion.framework",
+				"${BUILT_PRODUCTS_DIR}/Down/Down.framework",
+				"${BUILT_PRODUCTS_DIR}/FSInteractiveMap/FSInteractiveMap.framework",
+				"${BUILT_PRODUCTS_DIR}/Folly/folly.framework",
```
jtreanor: It looks like everything is being included as frameworks after the upgrade. I think we will need to tweak our `Podfile` slightly.

jtreanor: Yep, it looks like this was changed in 1.7. We need to change our logic to look something like [this](https://github.com/leavez/cocoapods-static-swift-framework/blob/master/lib/cocoapods-static-swift-framework/patch/static_pod.rb#L16).

Ping me if you'd like a hand!

PR #12409, opened 2019-08-28
Blogs: Do not sync while syncing.

aerych:
Refs #7886 

This PR introduces some new state flags to track when the blog list is syncing, ~and the last time it was synced~.  This addresses Bugs 5 and 7 in #7886 where multiple syncs (and multiple derived contexts) could occur simultaneously, leading to duplicated blogs. 

It would be nice if only a single derived context could exist but I'm not too keen on keeping a reference to a context while syncing (and beyond).  Our pattern thus far is that contexts, like services, would be ephemeral. Open to other ideas tho.

Test Prep: 
- Use Network Link Conditioner to slow requests. 
- Optionally, edit [BlogListViewController viewWillAppear:] to call `autoSyncBlogs` multiple times. 
- ~Optionally adjust the time interval for auto syncs to avoid waiting a full 5 minutes between auto syncs.~
- Add break points where isSyncing is set.

To Test:
Scenario 1:
- Visit the blog list. ~Confirm it auto syncs blogs the first time.~
- Tap to view a blog detail.
- Tap back to the blog list.  ~Confirm that it does not sync blogs when tapping back.~
- ~Wait the time interval for auto syncs.~
- Tap to view a blog detail, then tap back to the blog list. Confirm that blogs are synced.

Scenario 2:
- On the blog list view controller, pull to refresh.  Confirm that blogs sync. 
- Repeat the pull to refresh.  Confirm that a sync occurs and you didn't have to wait for the autosave time interval.

Scenario 3:
- This one is a little tricky.  Confirm that you can not trigger two simultaneous syncs:
- Launch the app and view the blog list. 
- While blogs are syncing try to pull to refresh. 
- Confirm a second sync does not start. 
- Alternatively, you can add a `syncBlogs` call in `viewWillAppear` after the auto sync call, and confirm that a second sync does not start.

Update release notes:

Needs Review: @koke, would you mind taking a peek?

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
@@ -7,6 +7,7 @@ static NSInteger HideAllMinSites = 10;
 static NSInteger HideAllSitesThreshold = 6;
 static NSTimeInterval HideAllSitesInterval = 2.0;
 static NSInteger HideSearchMinSites = 3;
+static NSTimeInterval AutoSyncInterval = 300;
 
 @interface BlogListViewController () <UIViewControllerRestoration,
                                         UIDataSourceModelAssociation,
@@ -30,7 +31,8 @@ static NSInteger HideSearchMinSites = 3;
 @property (nonatomic) NSDate *firstHide;
 @property (nonatomic) NSInteger hideCount;
 @property (nonatomic) BOOL visible;
-
+@property (nonatomic) BOOL isSyncing;
+@property (nonatomic, strong) NSDate *lastSyncDate;
 @end
 
 @implementation BlogListViewController
@@ -146,7 +148,7 @@ static NSInteger HideSearchMinSites = 3;
     [self maybeShowNUX];
     [self updateViewsForCurrentSiteCount];
     [self validateBlogDetailsViewController];
-    [self syncBlogs];
+    [self autoSyncBlogs];
     [self setAddSiteBarButtonItem];
     [self updateCurrentBlogSelection];
 }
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
```diff
@@ -7,6 +7,7 @@
 static NSInteger HideAllSitesThreshold = 6;
 static NSTimeInterval HideAllSitesInterval = 2.0;
 static NSInteger HideSearchMinSites = 3;
+static NSTimeInterval AutoSyncInterval = 300;
```
koke: Why are we limiting this to 5 minutes? I'm not sure if we should do it or if that's the right interval, but it seems unrelated to the sync locking changes and I'd prefer to leave it out of this PR.

Since we just added #12377 to de-duplicate blogs on sync, I'd rather sync too often, at least until this issue is sorted out.

WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
```diff
@@ -338,34 +340,55 @@ - (void)validateBlogDetailsViewController
     }
 }
 
+- (void)autoSyncBlogs
+{
+    if (!self.lastSyncDate || fabs([self.lastSyncDate timeIntervalSinceNow]) > AutoSyncInterval) {
+        [self syncBlogs];
```
koke: I think we should be calling `[self.tableView.refreshControl beginRefreshing]` from here?

koke: According to [the docs](https://developer.apple.com/documentation/uikit/uirefreshcontrol/1624842-beginrefreshing):

> Call this method when an external event source triggers a programmatic refresh of your scrolling view. In a table view, for example, if you use an instance of Timer to refresh the contents of the table view periodically, you would call this method as part of your timer handler. This method updates the state of the refresh control to reflect the in-progress refresh operation. When the refresh operation ends, be sure to call the endRefreshing() method to return the control to its default state.

I understand what this does is telling the refresh control that we are already syncing. That way if the user starts dragging to refresh, a loading indicator would appear instead of triggering a second refresh (or trying to).

aerych:
Thanks @koke!

PR #12410, opened 2019-08-28
[Fix] Author filter issues

### LINKED ISSUE #12330: Author filter issues
### Expected behavior

The author tooltip filter should have a tiny arrow on top, and be responsive.
IMG_8600

### Actual behavior

The arrow is not present on iOS 13 and the tooltip is super unresponsive.
IMG_1776

On an iOS 13, multi-author blog, tap on the avatar on the top left column. Notice the arrow missing and the unresponsiveness.

danielebogo:
Fixes #12330 

This PR wants to fix the popover tableview inset while the app is running on iOS 13.
Investigating the problem I found here ([1](https://forums.developer.apple.com/thread/121729) | [2](https://twitter.com/steipete/status/1157666057424912384)) that 
popovers in iOS 13 use safe area. This implies I had to fix the top inset and avoid the bottom safe area over the last row.

## To test:
- Run the App on iOS 13
- Open the Post list and tap on the Author filter
- Make the same test with iOS 12

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/AuthorFilterViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/AuthorFilterViewController.swift
@@ -62,6 +62,8 @@ class AuthorFilterViewController: UITableViewController {
         get {
             let height = CGFloat(tableView(self.tableView, numberOfRowsInSection: 0)) * Metrics.rowHeight
             if #available(iOS 13, *) {
+                // Popovers in iOS 13 use safe area. This means the safe area is over the last row.
+                // I need to add the safe area bottom inset to the height.
                 let bottomSafeInset = UIApplication.shared.keyWindow?.safeAreaInsets.bottom ?? 0.0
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Post/AuthorFilterViewController.swift
```diff
@@ -48,6 +48,11 @@ class AuthorFilterViewController: UITableViewController {
         tableView.separatorColor = .neutral(.shade10)
         tableView.isScrollEnabled = false
         tableView.showsVerticalScrollIndicator = false
+        if #available(iOS 13, *) {
+            tableView.contentInset = .zero
+        } else {
+            tableView.contentInset = UIEdgeInsets(top: -13.0, left: 0, bottom: 0, right: 0)
```
frosty: Let's use Metrics.topInset here :)

danielebogo:
Thanks @frosty for the review! I fixed it! Ready for another quick look!

danielebogo:
Thanks @frosty !

PR #12412, opened 2019-08-28
Issues/7886 bug 2

aerych:
Refs #7886 

This PR attempts to address Bug 2 as @koke described [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/7886#issuecomment-524221031). Incidentally it should also sort out bug 1.

With these changes, when a call is made to retrieve the default account and the default can not be retrieved, the remaining accounts in the app are examined to find one that is a likely candidate to be the default. Accounts that were added strictly to connect a self-hosed blog to Jetpack, but were not default accounts themselves, are ignored to avoid surprising the user.

These changes leave the edge case where a user was logged in with a Jetpack connected account with a self-hosted site but no wpcom blogs of its own, but this seems pretty fringe (I think?).

To test:
Run AccountServiceTests.  Ensure all pass. 

Scenario 1:
Confirm you can log in to an account, and log back out without errors. No regressions!

Scenario 2:
Be logged in to a WordPress.com account.
Put a break point at the beginning of the app delegate's `willFinishLaunchingWithOptions` method. 
Start the simulator, and when the break point is reached, clear the default account key from user defaults by running the following in the debugger:
`po UserDefaults.standard.removeObject(forKey: "AccountDefaultDotcomUUID")`
Continue execution.  Confirm you are still logged in. 

Scenario 3:
Log in to a WordPress.com account
Add a self-hosted site that is connected to a different WordPress.com account. 
Sign in to Jetpack on that WordPress.com account. 
Switch to the Me tab and log out. 
Confirm that you are actually logged out of the first WP com account and you are not logged in to the Jetpack connected account. 

Scenario 4:
Log in to a WordPress.com account
Add a self-hosted site that is connected to a different WordPress.com account. 
Sign in to Jetpack on that WordPress.com account. 
Repeat the steps in Scenario 2 to clear the key from user defaults. 
Confirm the initial WordPress.com account is restored and not the Jetpack connected account. 

Update release notes:

@koke Can I trouble you with another?

### DIFF
--- a/WordPress/Classes/Services/AccountService.m
+++ b/WordPress/Classes/Services/AccountService.m
@@ -35,30 +35,15 @@ NSString * const WPAccountEmailAndDefaultBlogUpdatedNotification = @"WPAccountEm
 {
     NSString *uuid = [[NSUserDefaults standardUserDefaults] stringForKey:DefaultDotcomAccountUUIDDefaultsKey];
     if (uuid.length > 0) {
-        NSFetchRequest *fetchRequest = [NSFetchRequest fetchRequestWithEntityName:@"Account"];
-        NSPredicate *predicate = [NSPredicate predicateWithFormat:@"uuid == %@", uuid];
-        fetchRequest.predicate = predicate;
-
-        NSError *error = nil;
-        NSArray *fetchedObjects = [self.managedObjectContext executeFetchRequest:fetchRequest error:&error];
-        if (fetchedObjects.count > 0) {
-            WPAccount *defaultAccount = fetchedObjects.firstObject;
-            defaultAccount.displayName = [defaultAccount.displayName stringByDecodingXMLCharacters];
-            return defaultAccount;
+        WPAccount *account = [self accountWithUUID:uuid];
+        if (account) {
+            return account;
         }
     }
 
     // Attempt to restore a default account that has somehow been disassociated.
-    NSSortDescriptor *sort = [NSSortDescriptor sortDescriptorWithKey:@"blogs.@count" ascending:NO];
-    NSArray *accounts = [[self allAccounts] sortedArrayUsingDescriptors:@[sort]];
-
-    for (WPAccount *account in accounts) {
-        // Skip accounts that were likely added to Jetpack-connected self-hosted
-        // sites, while there was an existing default wpcom account.
-        if ([self isJetpackAccount:account]) {
-            continue;
-        }
-
+    WPAccount *account = [self findDefaultAccountCandidate];
+    if (account) {
         // Assume we have a good candidate account and make it the default account in the app.
         // Note that this should be the account with the most blogs.
         // Update user defaults here vs the setter method to avoid potential side-effects from dispatched notifications.
@@ -280,7 +265,7 @@ NSString * const WPAccountEmailAndDefaultBlogUpdatedNotification = @"WPAccountEm
  @param account The account to inspect.
  @return True if used only for a Jetpack connection.
  */
-- (BOOL)isJetpackAccount:(WPAccount *)account
+- (BOOL)accountHasOnlyJetpackBlogs:(WPAccount *)account
 {
     if ([account.blogs count] == 0) {
         // Most likly, this is a blogless account used for the reader or commenting and not Jetpack.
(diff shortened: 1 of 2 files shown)


WordPress/Classes/Services/AccountService.m
```diff
@@ -242,6 +258,39 @@ - (NSUInteger)numberOfAccounts
...
+/**
+ Checks an account to see if it is just used to connect to Jetpack.
+
+ @param account The account to inspect.
+ @return True if used only for a Jetpack connection.
+ */
+- (BOOL)isJetpackAccount:(WPAccount *)account
+{
+    // If an account has multiple blogs, or zero blogs, then it is not an account
+    // that was connected via Jetpack login while another default wpcom account existed.
```
koke: I don't think that's accurate. With the logging changes in #12411 I tested:

1. Installed the app and logged in as kokejb7
2. Added self-hosted sandbox.koke.me, connected to `kokeme`. Logged in from stats
3. Added self-hosted nosenada.es, connected to `kokeme`. Stats worked right away without a second login
4. Signed out from Me > Sign out

On relaunch, this is the current state:

```
All accounts and blogs:
<Account username: kokeme ID: 442746, Email: koke@amedias.org (verified) ObjectID: x-coredata://26A6E50E-DCDD-4053-A67F-FA7E48A92DB1/Account/p6>
└─ <Blog Name: Koke's .org Sandbox URL: https://sandbox.koke.me XML-RPC: https://sandbox.koke.me/xmlrpc.php wp.com account: kokeme blogId: 137551970 plan: Free (2002) ObjectID: x-coredata://26A6E50E-DCDD-4053-A67F-FA7E48A92DB1/Blog/p2176>
└─ <Blog Name: Sólo sé que no sé nada URL: https://nosenada.es XML-RPC: https://nosenada.es/xmlrpc.php wp.com account: kokeme blogId: 94917218 plan: Free (2002) ObjectID: x-coredata://26A6E50E-DCDD-4053-A67F-FA7E48A92DB1/Blog/p2177>
```

I think that we should just see if an account has any WordPress.com blogs and promote that to default.

aerych: Oh good catch there. 😬 
> I think that we should just see if an account has any WordPress.com blogs and promote that to default.

There's still a situation where a user might have had a default account with 0 blogs.  Lemme tweak the logic a bit. I like the idea of checking for wpcom hosted blogs, and we can add an extra check for 0 blogs case.

aerych:
Heya @koke, I tweaked the logic a bit.  Thoughts?

WordPress/Classes/Services/AccountService.m
```diff
@@ -34,25 +34,41 @@ @implementation AccountService
 - (WPAccount *)defaultWordPressComAccount
 {
     NSString *uuid = [[NSUserDefaults standardUserDefaults] stringForKey:DefaultDotcomAccountUUIDDefaultsKey];
-    if (uuid.length == 0) {
-        return nil;
+    if (uuid.length > 0) {
+        NSFetchRequest *fetchRequest = [NSFetchRequest fetchRequestWithEntityName:@"Account"];
```
koke: This block could be extracted as `findAccountWithUUID:`

WordPress/Classes/Services/AccountService.m
```diff
@@ -34,25 +34,41 @@ @implementation AccountService
...
-    if (fetchedObjects.count > 0) {
-        defaultAccount = fetchedObjects.firstObject;
-        defaultAccount.displayName = [defaultAccount.displayName stringByDecodingXMLCharacters];
-    } else {
-        [[NSUserDefaults standardUserDefaults] removeObjectForKey:DefaultDotcomAccountUUIDDefaultsKey];
+    // Attempt to restore a default account that has somehow been disassociated.
+    NSSortDescriptor *sort = [NSSortDescriptor sortDescriptorWithKey:@"blogs.@count" ascending:NO];
+    NSArray *accounts = [[self allAccounts] sortedArrayUsingDescriptors:@[sort]];
+
+    for (WPAccount *account in accounts) {
```
koke: And this part could be `findDefaultAccountCandidate`

WordPress/Classes/Services/AccountService.m
```diff
@@ -242,6 +258,44 @@ - (NSUInteger)numberOfAccounts
...
+    return fetchedObjects;
+}
+
+/**
+ Checks an account to see if it is just used to connect to Jetpack.
+
+ @param account The account to inspect.
+ @return True if used only for a Jetpack connection.
+ */
+- (BOOL)isJetpackAccount:(WPAccount *)account
```
koke: The logic seems good, but I think the naming is confusing. Perhaps `accountHasOnlyJetpackBlogs`? The current name would imply to me that this would return true if the account had any Jetpack blogs, regardless of other conditions.

koke:
Interesting effect: I started testing this in a device that already had duplicate accounts. When I signed out, it started asking me the password for the account and would not let me cancel until I entered the password

aerych:
Thanks Koke! Great suggestions, as always. I'll rename and shuffle some code around then merge this. 🙇

PR #12416, opened 2019-08-28
Cancel media auto-uploads after 3 failures

### LINKED ISSUE #12369: [Offline Posting] Do not automatically retry upload after 3 failures
📣 Part of #12240. Please use the `issue/12240-master-branch` as the target branch. 

In Android, https://github.com/wordpress-mobile/WordPress-Android/issues/10059, we ran into an issue where a post can be automatically uploaded indefinitely. There can be many reasons for this but one of the reasons can be that the user doesn't have permission to upload to the site. 

We should also add this for iOS to preemptively avoid this scenario. 

See [UploadActionUseCase](https://github.com/wordpress-mobile/WordPress-Android/blob/ec1fb997c3d3113048c2c83b03863d5f77f20642/WordPress/src/main/java/org/wordpress/android/ui/uploads/UploadActionUseCase.kt#L46-L49) for how it is implemented on Android.

## Known Tasks  

- [x] #12504 [Auto-upload Published] Do not automatically retry media uploads after 3 failures
- [ ] Add maximum retry for posts
- [ ] Add tests

diegoreymendez:
Moves forward #12369 for Media.

This PR makes sure that media uploads are not retried after the 3rd automatic retry fails.

PS: @shiki since we have model 90 in our feature branch, this PR will most likely collide - unfortunately there's no easy way for me to make CD changes, target master, and not break our feature branch.

## Important Notes:

- This PR is purposedly aimed at `develop` since the changes should be fairly safe, and also valuable to reduce the impact of some recent issues with failing uploads.
- We'll need to resolve CoreData conflicts when merging this into our feature branch.  That's unfortunately unavoidable on my end.
- The PR is kind of simple, but it took me a while to wrap it up.  There were several small things that made the adventure more complicated than it should have been.  For this reason, this PR has no automated testing now, and I'll analyze tomorrow if I can add some.  Please though, go ahead and start reviewing the PR.

### Preparation

(A) Plase a breakpoint here: https://github.com/wordpress-mobile/WordPress-iOS/blob/ea2c870222e723512dd5d19315fa3a3e1d2fb8a1/WordPress/Classes/Services/MediaService.m#L258

(B) Plase a breakpoint here: https://github.com/wordpress-mobile/WordPress-iOS/blob/3dc497cb38038d664c13dd8994ebe4b3361c39a8/WordPress/Classes/Services/MediaService.m#L226

Make sure you have no media pending auto-upload.  This can be tricky to verify, but considering the logic introduced in this PR, it should be enough to run & kill the App a few times while online to resolve (zero out) all pending media uploads.

### Test 1: Check that a successful manual upload doesn't increase the failure counter.

1. Create a new post
2. Insert an image
3. Make sure that breakpoint (A) is hit as soon as the upload starts.

### Test 2: Check that a failed manual upload leaves the failure counter at 1.

1. Be offline
2. Create a new post
3. Insert an image
4. Make sure that breakpoint (A) is hit as soon as the upload starts.
5. Make sure that breakpoint (B) is NOT hit when the upload fails (can take a while).

### Test 3: Check that manual retries reset the failure counter.

1. Do all the steps from test 2.
2. When the notice with the error comes up, tap "Retry".
3. Make sure that breakpoint (A) is hit as soon as the upload starts.
4. Make sure that breakpoint (B) is NOT hit when the upload fails (can take a while).

### Test 4: Check that automatic retries further increment the counter.

1. Do all the steps from test 2.
2. Stop the app and relaunch it with Xcode (still being offline).
3. Make sure (A) is not hit, but (B) is, and the counter increments to 2.
4. Stop the app and relaunch it with Xcode (still being offline).
5. Make sure (A) is not hit, but (B) is, and the counter increments to 3.
6. Stop the app and relaunch it with Xcode (still being offline).
7. Make sure (A) is not hit, but (B) is, and the counter increments to 4.
8. Stop the app and relaunch it with Xcode (still being offline).
9. Make sure no other breakpoints are hit.

### Test 5:

1. For this test, remove the breakpoints.
2. Do all steps from test 4.
3. Tap "Retry" on the failed post.
4. Make sure the post is uploaded just fine.

## Update release notes:

## Tasks 

- [x] RFC on https://github.com/wordpress-mobile/WordPress-iOS/pull/12416#issuecomment-526238430
- [x] RFC on https://github.com/wordpress-mobile/WordPress-iOS/pull/12416#pullrequestreview-281642098
- [ ] Add automated tests


diegoreymendez:
@elibud, @loremattei - This PR should help mitigate the media-failure-Sentry-issues considerably, and avoid the same from happening as heavily in the future.

shiki:
@diegoreymendez Some quick questions as I'm reviewing this. In #12369, we mentioned that the retry limit was implemented on **posts** and not on **media**. Considering that you marked this as "Fixes #12369", did you conclude that it's not necessary to limit the retry count on posts? Or are you planning to implement a similar thing for posts later?

leandroalonso:
@diegoreymendez It looks good me, just made 2 simple comments. :)

diegoreymendez:
@shiki - I've updated the PR description which now reads:

> Moves forward #12369 for Media.

Apologies for the confusion though.  Sometimes when you go too deep into an issue you lose sight of what's around it, and for some reason I was entirely focused on media.  I'll tackle posts next.

diegoreymendez:
Update:

- All comments were addressed.
- Media (CoreData) property `uploadFailureCount` was renamed to `autouploadFailureCount` so you may need to delete the App before running this again.
- The description of this PR was updated to address the changes.
   - Test 2 and 3 were updated.
   - Test 5 was added.

I'll now be adding automated tests, but please go ahead and provide feedback on the existing code.

diegoreymendez:
Marking this as "Not ready for review" so I can change the code a bit after an amazing suggestion by @leandroalonso, and after I add some automated testing.

diegoreymendez:
@shiki @leandroalonso - I'm currently having some trouble with test 5.  I have to upload twice for the upload to work... but otherwise this is working well.

Can I ask you to give this PR another look?  I'll probably need help next week so that we can reach a favorable conclusion.

diegoreymendez:
This PR is blocked by https://github.com/wordpress-mobile/WordPress-iOS/pull/12450

The good thing is that the proposed change is straightforward and we should be able to merge it soon.

WordPress/WordPressTest/MediaServiceTests.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    /// Requesting media for manual attempt, no previous automated retry failures
+    ///
+    func testThatFailedMediaForUploadWorks() {
+        let media = mediaBuilder.with(remoteStatus: .failed).build()
+
+        let failedMediaForUpload = mediaService.failedMediaForUpload(automatedRetry: false)
+
+        XCTAssertTrue(failedMediaForUpload.contains(media))
+        XCTAssertEqual(failedMediaForUpload.count, 1)
```
leandroalonso: We've been using Nimble expectations in some tests. I know this is a personal taste, but I found the syntax so much descriptive:

```
expect(failedMediaForUpload).to(contain(media))
expect(failedMediaForUpload.count).to(equal(1))
```

What do you think?

WordPress/Classes/Models/AbstractPost.swift
```diff
@@ -61,6 +61,26 @@ extension AbstractPost {
...
+
+    /// Updates the post after an upload failure.
+    ///
+    /// - Important: This logic could have been placed in the setter for `remoteStatus`, but it's my belief
+    ///     that our code will be much more resilient if we decouple the act of setting the `remoteStatus` value
+    ///     and the logic behind processing an upload failure.  In fact I think the `remoteStatus` setter should
+    ///     eventually be made private.
+    ///
+    @objc
+    func processUploadFailure() {
```
leandroalonso: Maybe we can rename this method to `failedToUpload()`?

leandroalonso:
@diegoreymendez 2 more simple comments (besides the issue we're seeing in `Test 5`).

diegoreymendez:
This was closed automatically by mistake.  Reopening.

@shiki, @leandroalonso - I think I resolved the remaining issues.  Can you re-test this?

WordPress/WordPressTest/MediaServiceTests.swift
diegoreymendez: Right now we only have around 7 tests using Nimble, vs the rest of the codebase not using it.

I'd say it's more important for me to stay close in this case to our standard.  I'm not sure if there was a discussion around Nimble but it could be good to try and see if there's consensus in using it, first?

For more context on my thoughts: the problem with not following standards is that our codebase becomes more fragmented, harder to maintain and it also makes submitting code harder because it's difficult to know what the preference of the reviewer may be.

WordPress/Classes/Models/AbstractPost.swift
diegoreymendez: Done.

leandroalonso:
Everything works as expected. :)

There's only one issue I'm seeing: after hitting the "Retry" button, the image is uploaded but the post's content is not updated with the remote URL and I end up with a content like this:

```
<img src="file:///Users/leandroalonso/Library/Developer/CoreSimulator/Devices/E690FA1D-AE36-4267-905D-8F6E71F4FA31/data/Containers/Data/Application/92DFEF2D-73F0-45DA-8835-6B791A475D85/Library/Caches/Media/thumbnail-p48-1792x1792.jpeg" class="size-full" data-wp_upload_id="x-coredata://6A84C812-D895-4EF7-85F3-FAC36D74667E/Media/p48">

A
```

But, as we talked, the same issue happens on `develop` - we don't need to fix it on this PR.

WordPress/WordPressTest/MediaServiceTests.swift
leandroalonso: Ok!

diegoreymendez:
This PR is blocked by: https://github.com/wordpress-mobile/WordPress-iOS/pull/12459

leandroalonso:
:shipit:

PR #12423, opened 2019-08-29
Remove the old-style app icon

jkmassel:
I realized the old App Icon is still present – I don't think we need it anymore with the new Muriel colour style? 

**To Test:**
- Ensure that CI tests are green before starting.
- Ensure that the main icon works correctly.
- Ensure that alternate icons work correctly.
- Build and **archive** the `WordPress Internal` target. Find the archive in the organizer, choose "Show in Finder", and see that the banner is present (it should be `13.1.0.12190814`)


frosty:
@jkmassel I think we can actually just remove the whole AppImages.xcassets file now that the new color scheme feature flag is removed. Perhaps we should rename the Muriel asset catalog now too.

jkmassel:
`AppImages.xcassets` has been removed entirely now and `MurielImages.xcassets` has taken its place. 

Testing steps updated, because now we should check a few things :)

PR #12425, opened 2019-08-29
Only publish a post after all media has been uploaded

### LINKED ISSUE #12408: When a post is updated and a media is being uploaded, it loses its core data reference
### Expected behavior

When I tap "Update" I would expect the post to reflect my changes in my website, including the images I uploaded.

### Actual behavior

If a media is being uploaded and I tap "Update" the post may end up with content like this - thus, the images not being shown:

```
<img src="///var/mobile/Containers/Data/Application/18DF75D5-F673-4ED7-B187-0C3DABD69A3F/Library/Caches/Media/thumbnail-p125-2436x2436.jpeg" class="size-full">
<img src="///var/mobile/Containers/Data/Application/18DF75D5-F673-4ED7-B187-0C3DABD69A3F/Library/Caches/Media/thumbnail-p126-2436x2436.jpeg" class="size-full">
```

These images references will never be replaced because we rely on the `data-wp_upload_id` tag that was removed.

I reproduced this issue following 2 different approaches:

#### Bad connection

1. Use a bad connection (you can use the throttle in Charles proxy or some tool like Network Conditioner)
2. Create a post with some text and publish it
3. Add some images to the post you just published
4. Tap "Update"
5. Close the app before the images were uploaded
6. Open the app
7. Go to the post and tap to edit it
8. Restore your connection and tap "Update"
9. Check the HTML: all `data-wp_upload_id` references were removed

#### Offline

1. Open the app and publish a post
2. Close the app and go offline
3. Edit the post and add an image
4. Go back online and tap "Update"
5. Check the HTML: all `data-wp_upload_id` references were removed

##### More info

`2275839-ze`

leandroalonso:
Fixes #12408

To test:

#### Bad connection

1. Use a bad connection (you can use the throttle in Charles proxy or some tool like Network Conditioner)
2. Create a post with some text and publish it
3. Add some images to the post you just published
4. Tap "Update"
5. Close the app before the images were uploaded
6. Open the app
7. Go to the post and tap to edit it
8. Restore your connection and tap "Update"
9. Check that `data-wp_upload_id` is still present or if the path of the images were all replaced for its remote URL

#### Offline

1. Open the app and publish a post
2. Close the app and go offline
3. Edit the post and add an image
4. Go back online and tap "Update"
5. Check that `data-wp_upload_id` is still present or if the path of the images were all replaced for its remote URL

### Additional information

The issue was happening if the posts had all images in a failed state PLUS not being uploaded. If those conditions were satisfied the post was published right away and all core data identifiers were removed.

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -36,7 +36,7 @@ class PostCoordinator: NSObject {
 
         change(post: post, status: .pushing)
 
-        if mediaCoordinator.isUploadingMedia(for: post) {
+        if mediaCoordinator.isUploadingMedia(for: post) || post.hasFailedMedia {
             change(post: post, status: .pushingMedia)
             // Only observe if we're not already
             guard !isObserving(post: post) else {
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -44,6 +44,7 @@ extension PostEditor where Self: UIViewController {
             displayHasFailedMediaAlert(then: {
                 // Failed media is removed, try again.
                 // Note: Intentionally not tracking another analytics stat here (no appropriate one exists yet)
+                self.removeFailedMedia()
                 self.publishPost(action: action, dismissWhenDone: dismissWhenDone, analyticsStat: analyticsStat)
             })
             return
@@ -133,7 +134,6 @@ extension PostEditor where Self: UIViewController {
     fileprivate func displayHasFailedMediaAlert(then: @escaping () -> ()) {
         let alertController = UIAlertController(title: FailedMediaRemovalAlert.title, message: FailedMediaRemovalAlert.message, preferredStyle: .alert)
         alertController.addDefaultActionWithTitle(FailedMediaRemovalAlert.acceptTitle) { [weak self] alertAction in
-            self?.removeFailedMedia()
             then()
         }
 


shiki:
This solution has an unintended consequence of **auto-uploading published posts**. We don't want that until #12227. To reproduce:

1. Go offline.
2. Create a post. Add 2 images from the device.
3. Press _Publish_ on the top right and confirm. The editor will be closed. Notice that the post is now in the Published tab of the Post List. 
4. Go online. 

The post will be auto-uploaded **and published**.  This does not happen in the current `develop`.

leandroalonso:
@shiki maybe we should target the Offline master branch instead of `develop`?

If I understood correctly, a post should never be published before all the associated `Media` were uploaded.

Also, if the user hits `Publish` is correct to assume that the user intends to actually publish the post? @osullivanchris any thoughts on that?

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -36,7 +36,7 @@ class PostCoordinator: NSObject {
 
         change(post: post, status: .pushing)
 
-        if mediaCoordinator.isUploadingMedia(for: post) {
+        if mediaCoordinator.isUploadingMedia(for: post) || post.hasFailedMedia {
```
shiki: So I think this is the cause of the post getting published. Originally, if the post upload fails, we were resetting the `status` to `.draft` so the post will **not be published**. That happens here:

https://github.com/wordpress-mobile/WordPress-iOS/blob/72a217d5b2fec206747e93976eeb323faeb1006f/WordPress/Classes/Services/PostService.m#L259-L268

But since the `upload(post: post)` in line 84 is not getting called anymore, there is no chance for the failure block (above) to ever get called and reset the `status` to `.draft`. 

Also, the condition of `post.hasFailedMedia` may be correct but it does not account for the fact that after this line:

1. We observe for any pending media uploads
2. After all pending media uploads are finished, upload the post again

The problem is, we are only assuming that the media are "pending" and not "failed". In the event that all media are failed, there are no media to **observe and wait for** and so the `self.upload(post: post)` in line 57 is never called again.

shiki: An idea from @diegoreymendez: We can probably merge these:

https://github.com/wordpress-mobile/WordPress-iOS/blob/72a217d5b2fec206747e93976eeb323faeb1006f/WordPress/Classes/Services/PostCoordinator.swift#L29

https://github.com/wordpress-mobile/WordPress-iOS/blob/72a217d5b2fec206747e93976eeb323faeb1006f/WordPress/Classes/Services/PostCoordinator.swift#L149

...since `retrySave()` does the restarting of failed media before uploading the post.

leandroalonso:
Closed in favor of #12452

Issue #11398, opened 2019-04-03
Fatal Exception: NSRangeException in WPRichContentView deinit


jkmassel:
From Fabric:

```
Fatal Exception: NSRangeException
*** NSRunStorage, _NSBlockNumberForIndex(): index (9223372036854722427) beyond array bounds (9223372036854722427)

Fatal Exception: NSRangeException
0  CoreFoundation                 0x22ded8ec4 __exceptionPreprocess
1  libobjc.A.dylib                0x22d0a9a50 objc_exception_throw
2  CoreFoundation                 0x22dddf594 -[NSCache init]
3  UIFoundation                   0x23855477c _NSBlockNumberForIndex
4  UIFoundation                   0x23854bdd4 _NSDeleteGlyphRangeFromTextBlocks
5  UIFoundation                   0x2384f89f8 -[NSLayoutManager setTextStorage:]
6  CoreFoundation                 0x22ddbe5dc -[__NSArrayM enumerateObjectsWithOptions:usingBlock:]
7  UIFoundation                   0x2384ef4ac -[NSTextStorage dealloc]
8  UIFoundation                   0x2384f4b90 -[NSConcreteTextStorage dealloc]
9  UIKitCore                      0x25b4ffd84 -[UITextView .cxx_destruct]
10 libobjc.A.dylib                0x22d0a8478 object_cxxDestructFromClass(objc_object*, objc_class*)
11 libobjc.A.dylib                0x22d0b871c objc_destructInstance
12 libobjc.A.dylib                0x22d0b8784 object_dispose
13 UIKitCore                      0x25b1f0fac -[UIResponder dealloc]
14 UIKitCore                      0x25b63fdcc -[UIView dealloc]
15 UIKitCore                      0x25b5ec9c8 -[UIScrollView dealloc]
16 UIKitCore                      0x25b4f08a0 -[UITextView dealloc]
17 WordPress                      0x102206b80 WPRichContentView.__deallocating_deinit (WPRichContentView.swift:159)
18 WordPress                      0x102206bbc @objc WPRichContentView.__deallocating_deinit (<compiler-generated>)
19 CoreFoundation                 0x22de370d4 cow_cleanup
20 CoreFoundation                 0x22ddbf408 -[__NSArrayM dealloc]
21 libobjc.A.dylib                0x22d0c338c (anonymous namespace)::AutoreleasePoolPage::pop(void*)
[...]

PR #12427, opened 2019-08-30
RichContent: Tweaks deinit logic.

aerych:
Refs #11398 

This PR is an attempt to address the crash in #11398. 

Rather than call `WPRichTextImage.clean` during `WPRichContentView.deinit`, let's call `CachedAnimatedImageView.clean` from `WPRichTextImage.deinit`.  This will avoid the call to `CachedAnimatedImageView.prepareForReuse`, which doesn't make sense during `deinit` and still cancel any inflight session tasks. 

You can see from the comments in the issue that its not clear how the deinit call could be related to the crash, but we might at least be able to rule out a suspect.

To test:
Launch the app and view the Reader. 
Find posts or comments with animated images. 
Confirm there are no regressions and that WPRichTextImage.deinit is called.

Update release notes:

@jleandroperez Would you be game for a quick review of this one?


aerych:
Thanks @jleandroperez !

PR #12428, opened 2019-08-30
Fix Scroll Indicator in login epilogue screen

### LINKED ISSUE #12162: Scroll Indicator not visible on login epilogue screen
#11791 # Expected behavior
After logging in, if I have a long list of sites, I'd expect to be able to scroll through them and see the scroll indicator fully.

### Actual behavior
The scroll indicator is mostly off-screen. 

1) Log into any WordPress.com account.
2) Scroll within the list of sites (only tested on iPhone Xs hardware and simulator)
3) Note the lack of (or very hard to see) scroll indicator

### Potential Solution
I haven't researched this issue deeply, but I wonder if it's related to the site list being pinned to the device edges instead of the safe area insets? 
 
IMG_7EDAF672F3FF-1

Simulator Screen Shot - iPhone Xs - 2019-08-30 at 13 17 35

^^^ This screenshot is redacted – it only shows the scrollbar. In the original, all rows looked fine.

lorenzolewis:
Fixes #12162 

I traced this back to heighForRowAt in the LoginEpilogueTableViewController. If that is disabled, the scrollbar works as expected. This causes the side affect of the rows not being the proper height (screenshot below)

Not really sure where to go from here, if this would be setting a different heighForRowAt value, or adjusting the constrains in the storyboard. 

B87AD15C-E12E-4CB5-97D9-AFDCAA902857

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/LoginEpilogueTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/NUX/LoginEpilogueTableViewController.swift
@@ -123,9 +123,9 @@ extension LoginEpilogueTableViewController {
         return Settings.headerHeight
     }
 
-    override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
-        return UITableView.automaticDimension
-    }
+//    override func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+//        return UITableView.automaticDimension
+//    }
 
     override func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
         return UITableView.automaticDimension


frosty:
Hey again @lorenzolewis :) 

So I think we will need to keep the `heightForRow` definition, or we can just define `tableView.rowHeight = UITableView.automaticDimension` in `viewDidLoad`. As you identified, without this autoresizing cells don't work properly.

It might be worth experimenting with the autolayout constraints in `LoginEpilogue.storyboard`. It seems as though the table view controller is inset by -20 on the left and right, which I think would have the effect of pushing the scrollbar out:

image

I'm not sure if there's a reason this is like it is, but you could see what the effect is of adjusting those. If they do need to be outset, another idea would be to adjust the tableView's `scrollIndicatorInsets`to bring the scroll indicator back in without adjusting the rest of the tableView.

Hope this gives you something to go on!

frosty:
I'm going to close this for now to clear up our PR queue, as it's not been addressed in many months. Please feel free to re-open a new PR if you'd like to work on this again.

Issue #11389, opened 2019-04-01
ReaderPostServiceRemote fetchPostsFromEndpoint Crash


jkmassel:
From Fabric:

```
Fatal Exception: NSInvalidArgumentException
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[1]

Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x19dcbcea4 __exceptionPreprocess
1  libobjc.A.dylib                0x19ce8ca50 objc_exception_throw
2  CoreFoundation                 0x19dc34384 _CFArgv
3  CoreFoundation                 0x19dbba870 -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]
4  CoreFoundation                 0x19dbab3b4 +[NSDictionary dictionaryWithObjects:forKeys:count:]
5  WordPressKit                   0x1028405f8 -[ReaderPostServiceRemote fetchPostsFromEndpoint:algorithm:count:before:success:failure:] (ReaderPostServiceRemote.m:98)
6  WordPress                      0x1006cae14 -[ReaderPostService fetchPostsForTopic:earlierThan:deletingEarlier:success:failure:] (ReaderPostService.m:118)
7  WordPress                      0x1006ca5c4 -[ReaderPostService fetchPostsForTopic:earlierThan:success:failure:] (ReaderPostService.m:48)
8  WordPress                      0x100a25cac closure #1 in ReaderStreamViewController.loadMoreItems(_:failure:) (<compiler-generated>)
9  WordPress                      0x100a2d778 partial apply for closure #1 in ReaderStreamViewController.loadMoreItems(_:failure:) (<compiler-generated>)
10 WordPress                      0x1008202c4 thunk for @escaping @callee_guaranteed () -> () (<compiler-generated>)
11 CoreData                       0x1a0915e88 developerSubmittedBlockToNSManagedObjectContextPerform
12 libdispatch.dylib              0x19d6f5484 _dispatch_client_callout
13 libdispatch.dylib              0x19d69cbd0 _dispatch_lane_serial_drain$VARIANT$mp
14 libdispatch.dylib              0x19d69d718 _dispatch_lane_invoke$VARIANT$mp
15 libdispatch.dylib              0x19d6a5eb8 _dispatch_workloop_worker_thread
16 libsystem_pthread.dylib        0x19d8d80dc _pthread_wqthread
[...]

PR #12430, opened 2019-08-30
Reader: Adjust guard statement to check for nil date.

aerych:
Refs #11389 

This PR is an attempt at resolving #11389 by adjusting a guard statement to check for a nil date--a possible cause of the crash based on the stack trace.

To test:
- Go to the Reader. 
- Scroll to load more. 
- Ensure more posts load (assuming there were more), and there are no regressions in this feature.

Update release notes:

@ctarda could I trouble you for a quick review?


aerych:
Thanks @ctarda !!

Issue #11384, opened 2019-04-01
WPTableViewHandler controllerDidChangeContent Crash


jkmassel:
From Fabric:

```
Crashed: com.apple.main-thread
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000000

Crashed: com.apple.main-thread
0  UIKitCore                      0x23befcd44 __46-[UITableView _updateWithItems:updateSupport:]_block_invoke + 80
1  UIKitCore                      0x23befce5c __46-[UITableView _updateWithItems:updateSupport:]_block_invoke.1100 + 228
2  UIKitCore                      0x23befc5b4 -[UITableView _updateWithItems:updateSupport:] + 1976
3  UIKitCore                      0x23bef5248 -[UITableView _endCellAnimationsWithContext:] + 14944
4  UIKitCore                      0x23bf0b4e8 -[UITableView endUpdates] + 88
5  WordPress                      0x10084a54c -[WPTableViewHandler controllerDidChangeContent:] (WPTableViewHandler.m:618)
6  CoreData                       0x212366f74 __82-[NSFetchedResultsController(PrivateMethods) _core_managedObjectContextDidChange:]_block_invoke + 5016
7  CoreData                       0x212329b28 developerSubmittedBlockToNSManagedObjectContextPerform + 156
8  CoreData                       0x2121f42dc -[NSManagedObjectContext performBlockAndWait:] + 252
9  CoreData                       0x212200308 -[NSFetchedResultsController(PrivateMethods) _core_managedObjectContextDidChange:] + 116
10 CoreFoundation                 0x20f7095bc __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 20
11 CoreFoundation                 0x20f709588 ___CFXRegistrationPost_block_invoke + 64
12 CoreFoundation                 0x20f708a7c _CFXRegistrationPost + 392
13 CoreFoundation                 0x20f708728 ___CFXNotificationPost_block_invoke + 96
14 CoreFoundation                 0x20f682524 -[_CFXNotificationRegistrar find:object:observer:enumerator:] + 1496
15 CoreFoundation                 0x20f7081d8 _CFXNotificationPost + 696
16 Foundation                     0x2100f0814 -[NSNotificationCenter postNotificationName:object:userInfo:] + 68
[...]

PR #12440, opened 2019-08-30
ReaderComments: Adjust comment service save behavior.

aerych:
Refs: #11384 

This PR is an attempt at addressing the crash in #11384.  There is some background that gives context to these changes in the issue, but the tl;dr is ReaderComments were saved on a derived context, but not via the intended mechanism.  Its theorized this allowed for circumstances that result in the WPTableViewHandler crash. 

The PR does two things:
- Ensures a failure block is called from the main thread.
- Ensures that a derived context is properly saved.  Some code was slightly refactored to allow for this.

To test:
Open the reader and view comments for several posts. 
Ensure comments are correctly loaded. 
Try liking or replying to a comment. 
Ensure content updates and nothing goes 💥 .

Update release notes:

@astralbodies since this deals with main and derived contexts I'd really appreciate your eyes on this one if you have some free time.


aerych:
Thanks @astralbodies !

PR #12450, opened 2019-09-02
Makes the remoteStatus change synchronous for media uploads.

diegoreymendez:
While trying to resolve #12416, I realized that media being uploaded wasn't properly marked as having `remoteStatus == .pushing`.  Upon debugging this issue I realized the problem was that we are setting this status asyncrhonously here:

https://github.com/wordpress-mobile/WordPress-iOS/blob/534556756a7d11efb3e10000aad6860493bc7d6b/WordPress/Classes/Services/MediaService.m#L245-L252

In this PR I change that block to be executed synchronously, as well as the save operation at the end of the block.

## Testing:

Place breakpoint A [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/534556756a7d11efb3e10000aad6860493bc7d6b/WordPress/Classes/Services/MediaService.m#L248).
Place breakpoint B [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/534556756a7d11efb3e10000aad6860493bc7d6b/WordPress/Classes/Services/MediaService.m#L253)

### Test:

1. Launch the editor and insert an image in a post.
2. Breakpoint A should be hit.
3. Do `po mediaInContext.remoteStatus` and make sure the remote status is `MediaRemoteStatusLocal`.
4. Step once, `po mediaInContext.remoteStatus` and make sure it's now `MediaRemoteStatusPushing`.
5. Tap play, breakpoint B should hit.
6. Do `po mediaInContext.remoteStatus` and make sure the remote status is still `MediaRemoteStatusPushing`. - ie: it was synchronously saved

## Update release notes:

### DIFF
--- a/WordPress/Classes/Services/MediaService.m
+++ b/WordPress/Classes/Services/MediaService.m
@@ -242,12 +242,12 @@ NSErrorDomain const MediaServiceErrorDomain = @"MediaServiceErrorDomain";
         return;
     }
 
-    [self.managedObjectContext performBlock:^{
+    [self.managedObjectContext performBlockAndWait:^{
         Media *mediaInContext = (Media *)[self.managedObjectContext existingObjectWithID:mediaObjectID error:nil];
         if (mediaInContext) {
             mediaInContext.remoteStatus = MediaRemoteStatusPushing;
             mediaInContext.error = nil;
-            [[ContextManager sharedInstance] saveContext:self.managedObjectContext];
+            [[ContextManager sharedInstance] saveContextAndWait:self.managedObjectContext];
         }
     }];
     void (^successBlock)(RemoteMedia *media) = ^(RemoteMedia *media) {
--- a/WordPress/Classes/Utility/ContextManager.m
+++ b/WordPress/Classes/Utility/ContextManager.m
@@ -135,6 +135,23 @@ static ContextManager *_override;
 
 - (void)saveContextAndWait:(NSManagedObjectContext *)context
 {
+    __block NSManagedObjectContext *currentContext = context;
+    
+    while (currentContext.parentContext != nil) {
+        [currentContext performBlockAndWait:^{
+            NSError *error;
+            
+            if (![currentContext save:&error]) {
+                DDLogError(@"Fatal Core Data Error encountered — throwing an exception. Underlying error is:\n %@", error);
+                @throw [NSException exceptionWithName:@"Unresolved Core Data save error"
+                                               reason:[NSString stringWithFormat:@"Unresolved Core Data save error - derived context. Core Data Error Domain: %@, code: %i", error.domain, error.code]
+                                             userInfo:error.userInfo];
+            }
+            
+            currentContext = currentContext.parentContext;
+        }];
+    }
+    
     [context performBlockAndWait:^{
         [self internalSaveContext:context];
     }];


diegoreymendez:
I'm going ahead and merging this @astralbodies as it's pretty straightforward and it's blocking another big PR from being merged.  Can you let me know if you see anything off and I'll address it afterwards?  Thanks!

diegoreymendez:
Reverted here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12453

I'm restoring this branch as I need to investigate a deadlock report by @leandroalonso .

WordPress/Classes/Services/MediaService.m
```diff
@@ -242,12 +242,12 @@ - (void)uploadMedia:(Media *)media
         return;
     }
 
-    [self.managedObjectContext performBlock:^{
+    [self.managedObjectContext performBlockAndWait:^{
```
shiki: @diegoreymendez @leandroalonso @jklausa 

I performed some experiments and it looks like `performBlockAndWait` is the real culprit of the deadlock. To reiterate, the deadlock can be reproduced by:

1. Go offline. 
2. Create a post and add an image. 
3. Exit the editor and save the post as a draft. 
4. Stay in the Post List. Go online. 
5. The media will be automatically uploaded but there will be a deadlock and the app will freeze. 

The reason why the deadlock only happens during this scenario is because of where this method, `uploadMedia()`, is called when this happens. Consider this stack trace:

image

In there, you would notice that one of the originating methods is `getFailedPosts`. Looking at `getFailedPosts` it has a `performBlock` on **the same context**.

https://github.com/wordpress-mobile/WordPress-iOS/blob/38b278b207fdeb2f409e378b25417dbf474d0c77/WordPress/Classes/Services/PostService.m#L69-L70

The same happens if the `uploadMedia()` is called from `getFailedMedia`: 

https://github.com/wordpress-mobile/WordPress-iOS/blob/38b278b207fdeb2f409e378b25417dbf474d0c77/WordPress/Classes/Services/MediaService.m#L512-L513

### What I think this means

This means that the call to `self.managedObjectContext performBlockAndWait` queues the block below but the block is **waiting for the (same) current thread**. And it is never called because the same thread is blocked. 

1. Thread A: `getFailedMedia` → `performBlock`
2. Thread A: `retryMedia()` → `uploadMedia()`
3. Thread A: `uploadMedia()` → `performBlockAndWait`. Waits for **Thread A**. But we're in Thread A. 

### Proof

This is a shallow and not-even-close-to-production ready fix. In `getFailedMedia` and `getFailedPosts`, if you call the `result()` block in a background thread, the deadlock disappears:

```objc
        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
            if (!results) {
                result(@[]);
            } else {
                result(results);
            }
        });
```

### Recommendation

With all that said, I don't think we should be calling `performBlockAndWait` anyway. Especially since the `uploadMedia()` can be called on the **main thread**. If we want to ensure that the saving is done first before the routines below this, perhaps we can make use of the completion block? 

```objc
        Media *mediaInContext = (Media *)[self.managedObjectContext existingObjectWithID:mediaObjectID error:nil];
        if (mediaInContext) {
            mediaInContext.remoteStatus = MediaRemoteStatusPushing;
            mediaInContext.error = nil;
            [[ContextManager sharedInstance] saveContext:self.managedObjectContext withCompletionBlock:^{
                // DO THE ROUTINES IN HERE
            }];
        }
    }];
```

I haven't tried this yet but probably worth a try?

diegoreymendez: Thanks for the info @shiki.  This is very valuable and clear.

I'm not sure I agree with the recommendation, though, because I would definitely expect that a call to `uploadMedia()` would result in the media's `remoteStatus` to be `.pushing` immediately and for any thread.

The reason why this is particularly important in my opinion, is because the moment the media starts uploading, we most likely want any code (literally anywhere in our codebase) to be updated on that.  If this change is only available to the caller via a completion block, then we get a chance that other code elsewhere may query CD and get an outdated `remoteStatus` - which could result in other very obscure / hard to track issues.

I'll try to see if I can spend some time with your analysis and come up with an alternative solution to propose.

PR #11245, opened 2019-03-12
Improve re-authentication prompts

koke:
Currently, when the oAuth2 token is revoked, the app doesn't notice unless you refresh the posts list.
This PR will add a handler to, whenever we receive an `invalid_token` response from WordPress.com, present the login screen and remove that token from the keychain

To test:

1. Log in to the app
2. Go to https://wordpress.com/me/security/connected-applications and disconnect WordPress for iOS
3. Navigate somewhere in the app, any network request should trigger the login screen

If you kill and relaunch the app, you should also get a prompt to log in

PR #12451, opened 2019-09-02
Issues/7886 merge dup accounts

aerych:
Refs #7886

This PR attempts to merge any duplicate accounts in the app.  At launch the accounts in the app are inspected, duplicates are grouped and then merged. If a duplicate is the default account it is favored over other accounts. Blogs are merged by moving all blogs to the account to keep, then de-duplicating the blogs.  I did not opt to merge account settings since this should just be synced at launch anyway. 

To test:
Run tests and confirm they pass.

Update release notes:

@koke Would you please? :)

### DIFF
--- a/WordPress/Classes/Services/AccountService+MergeDuplicates.swift
+++ b/WordPress/Classes/Services/AccountService+MergeDuplicates.swift
@@ -9,10 +9,7 @@ extension AccountService {
         }
 
         let accountGroups = Dictionary(grouping: accounts) { $0.userID }
-        for key in accountGroups.keys {
-            guard let group = accountGroups[key], group.count > 1 else {
-                continue
-            }
+        for group in accountGroups.values where group.count > 1 {
             mergeDuplicateAccounts(accounts: group)
         }
 
@@ -34,10 +31,7 @@ extension AccountService {
             destination = defaultAccount
         }
 
-        for account in accounts {
-            if account == destination {
-                continue
-            }
+        for account in accounts where account != destination {
             mergeAccount(account: account, into: destination)
         }
 
@@ -48,7 +42,6 @@ extension AccountService {
     private func mergeAccount(account: WPAccount, into destination: WPAccount) {
         // Move all blogs to the destination account
         destination.addBlogs(account.blogs)
-        account.removeBlogs(account.blogs)
         managedObjectContext.delete(account)
     }
 
--- a/WordPress/Classes/Services/AccountService+MergeDuplicates.swift
+++ b/WordPress/Classes/Services/AccountService+MergeDuplicates.swift
@@ -3,11 +3,11 @@ import Foundation
 extension AccountService {
 
     func mergeDuplicatesIfNecessary() {
-        let accounts = allAccounts()
-        guard accounts.count > 1 else {
+        guard numberOfAccounts() > 1 else {
             return
         }
 
+        let accounts = allAccounts()
         let accountGroups = Dictionary(grouping: accounts) { $0.userID }
         for group in accountGroups.values where group.count > 1 {
             mergeDuplicateAccounts(accounts: group)
--- a/WordPress/Classes/Services/AccountService+MergeDuplicates.swift
+++ b/WordPress/Classes/Services/AccountService+MergeDuplicates.swift
@@ -1,0 +1,55 @@
+import Foundation
+
+extension AccountService {
+
+    func mergeDuplicatesIfNecessary() {
+        let accounts = allAccounts()
+        guard accounts.count > 1 else {
+            return
+        }
+
+        let accountGroups = Dictionary(grouping: accounts) { $0.userID }
+        for key in accountGroups.keys {
+            guard let group = accountGroups[key], group.count > 1 else {
+                continue
+            }
+            mergeDuplicateAccounts(accounts: group)
+        }
+
+        if managedObjectContext.hasChanges {
+            ContextManager.sharedInstance().save(managedObjectContext)
+        }
+    }
+
+    private func mergeDuplicateAccounts(accounts: [WPAccount]) {
+        // For paranoia
+        guard accounts.count > 1 else {
+            return
+        }
+
+        // If one of the accounts is the default account, merge the rest into it.
+        // Otherwise just use the first account.
+        var destination = accounts.first!
+        if let defaultAccount = defaultWordPressComAccount(), accounts.contains(defaultAccount) {
+            destination = defaultAccount
+        }
+
+        for account in accounts {
+            if account == destination {
+                continue
+            }
+            mergeAccount(account: account, into: destination)
+        }
+
+        let service = BlogService(managedObjectContext: managedObjectContext)
+        service.deduplicateBlogs(for: destination)
+    }
+
+    private func mergeAccount(account: WPAccount, into destination: WPAccount) {
+        // Move all blogs to the destination account
+        destination.addBlogs(account.blogs)
+        account.removeBlogs(account.blogs)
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Services/AccountService+MergeDuplicates.swift
```diff
@@ -0,0 +1,55 @@
...
+extension AccountService {
+
+    func mergeDuplicatesIfNecessary() {
+        let accounts = allAccounts()
+        guard accounts.count > 1 else {
+            return
+        }
+
+        let accountGroups = Dictionary(grouping: accounts) { $0.userID }
+        for key in accountGroups.keys {
```
koke: It doesn't seem like you need the key, why not `for group in accountGroups.values where group.count > 1`?

WordPress/Classes/Services/AccountService+MergeDuplicates.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        // If one of the accounts is the default account, merge the rest into it.
+        // Otherwise just use the first account.
+        var destination = accounts.first!
+        if let defaultAccount = defaultWordPressComAccount(), accounts.contains(defaultAccount) {
+            destination = defaultAccount
+        }
+
+        for account in accounts {
```
koke: I think you can also use `where` here:

```swift
for account in accounts where account != destination {
    mergeAccount(account: account, into: destination)
}
```

WordPress/Classes/Services/AccountService+MergeDuplicates.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        let service = BlogService(managedObjectContext: managedObjectContext)
+        service.deduplicateBlogs(for: destination)
+    }
+
+    private func mergeAccount(account: WPAccount, into destination: WPAccount) {
+        // Move all blogs to the destination account
+        destination.addBlogs(account.blogs)
+        account.removeBlogs(account.blogs)
```
koke: I wonder if that line is actually needed. I believe the previous one would replace each `Blog`'s `account` property. If, for some mysterious reason, it didn't, the next line should nullify that property.

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -56,6 +56,7 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, willFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil) -> Bool {
         window = UIWindow(frame: UIScreen.main.bounds)
 
+        AccountService.init(managedObjectContext: ContextManager.shared.mainContext).mergeDuplicatesIfNecessary()
```
koke: Instead of doing this on launch, what do you think about putting it right before syncBlogs? That way it's closer to the other de-duplication code, and we can trigger it via pull-to-refresh.

WordPress/Classes/Services/AccountService+MergeDuplicates.swift
aerych: I forget about `where` syntax sometimes 😬

WordPress/Classes/System/WordPressAppDelegate.swift
aerych: Dunno. It shouldn't need to be called so frequently and if there are multiple accounts then there is a slight overhead calling it. I'd prefer to not call it aggressively but if you have a strong feeling about it we could change it.

aerych:
Hound is being cranky about the inline arrays as method parameters. I think this is more of a quirk with the linter vs a problem with code style and I'm keen to ignore hound in this case--inline arrays seem tidier than the alternative.  Thoughts?

WordPress/Classes/System/WordPressAppDelegate.swift
koke: I can't make up my mind, I see benefits for both cases. If it stays here though, I'd remove from the initialization critical path. Maybe put it async on a background context like we do others:

```swift
// Deferred tasks to speed up app launch
DispatchQueue.global(qos: .background).async { [weak self] in
    MediaCoordinator.shared.refreshMediaStatus()
    PostCoordinator.shared.refreshPostStatus()
    MediaFileManager.clearUnusedMediaUploadFiles(onCompletion: nil, onError: nil)
    self?.uploadsManager.resume()
}
```

And you can drop the `.init` 😁

koke:
> Hound is being cranky about the inline arrays as method parameters. I think this is more of a quirk with the linter vs a problem with code style and I'm keen to ignore hound in this case--inline arrays seem tidier than the alternative. Thoughts?

Not sure if I'm missing a specific error, but I think Hound is OK with inline arrays, it just wants a space after every comma.

WordPress/Classes/System/WordPressAppDelegate.swift
aerych: > If it stays here though, I'd remove from the initialization critical path. 

Hmm.  I think we'd want to run this before the authtoken fixer tho.

aerych:
>  I think Hound is OK with inline arrays, it just wants a space after every comma.

https://cldup.com/be5NzfcBe8.png

WordPress/Classes/System/WordPressAppDelegate.swift
koke: I'm not sure I fully understand this scenario, but what I think you're saying is that:

1. There is a duplicate account
2. The auth token is missing
3. The token fixer asks the user to re-authenticate
4. If the user cancels, the default account is removed
5. If the de-duplication runs after that it would promote the duplicate as the default account but would still have no token

If that's the case, since #11245 we would just prompt for re-authentication again on the first network request that failed, so I'm not too worried about that.

aerych: Fair point.

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -234,6 +234,7 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
 
         // Deferred tasks to speed up app launch
         DispatchQueue.global(qos: .background).async { [weak self] in
+            AccountService(managedObjectContext: ContextManager.shared.mainContext).mergeDuplicatesIfNecessary()
```
koke: This is called on a background queue with a main context, it should be wrapped in a `context.performBlock()`

aerych: Good point. To keep things tidy I've moved this into a separate method to wrangle the perform block.

WordPress/Classes/Services/AccountService+MergeDuplicates.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+
+extension AccountService {
+
+    func mergeDuplicatesIfNecessary() {
+        let accounts = allAccounts()
+        guard accounts.count > 1 else {
```
koke: One last nitpick, for performance. Since most of the times this will not happen, and `allAccounts` still has to fetch all the accounts, I'd suggest to use `numberOfAccounts() > 1` for this check. This uses `countForFetchRequest` instead of `executeFetchRequest`, which should be much faster, as it doesn't have to read all the data from disk.

aerych: Ooooo good call.

aerych:
Thank you so much @koke.  I made that final change.  Once CI is happy I'll merge it up.

Issue #12362, opened 2019-08-22
[Auto-upload Published] Publish button in draft's success notice should auto-upload it


shiki:
📣 Part of #12240. Please use the `issue/12240-master-branch` as the target branch. 

**Waited on : https://github.com/wordpress-mobile/WordPress-iOS/pull/12466**

When a draft is successfully uploaded, we show a notice like this:

<img src="https://user-images.githubusercontent.com/198826/63533362-e02c9700-c4c9-11e9-9739-8bb14c50f33a.png" width="320">

Tapping on the **Publish** button on that notice should mark the post as “confirmed for auto-uploading“. If the user is offline, the post will be automatically published later. 

## Tasks 

- [ ] Possibly add a test in `PostNoticeViewModelTests`. Mocks for `PostCoordinator` have been added in #12358 so perhaps we should wait for that to be merged.

yaelirub:
@shiki,
So this seems pretty simple and should be fixed by adding `post.shouldAttemptAutoUpload = true`and save to context
in `PosticeNoticeViewModel's publishPost()` but since we are checking for hasRemote in:
https://github.com/wordpress-mobile/WordPress-iOS/blob/c1d32840deca04240a7ff1d26a843b5b37add607/WordPress/Classes/Services/PostAutoUploadInteractor.swift#L30
Posts that has been uploaded as draft will return `.nothing` for their action (as oppose to .upload which the status the autoUploader checks for auto uploading. 

My question is why do we have the hasRemote check?

yaelirub:
closed with #12523

PR #12466, opened 2019-09-04
Automatically upload published and draft posts

### LINKED ISSUE #12324: [Auto-upload Published] Automatically upload published and draft posts with existing remote
Part of #12240. 

This requires #12317 and #12136 to be completed first. Any changes should target the `issue/12240-master-branch` branch. 

All the work that's been done up to #12136 only supports drafts and published posts created on the device. If the post was uploaded to the server and the user edits it, we do not automatically it. The goal of this task is to support this.

Keep in mind, that there are rules for when we should auto-upload or remote auto-saved. They are described in the **Test Specs** section of #12240. 

## Other Tasks 

- [x] Waiting for responses on https://github.com/wordpress-mobile/WordPress-Android/issues/10461#issuecomment-526843789
- [ ] We should not automatically upload or remote auto-save when the post is in conflict with remote. We should add a unit test for this. [Android ref](https://github.com/wordpress-mobile/WordPress-Android/blob/6b0e9f286b2cd1744c826011dcca940b94248bd5/WordPress/src/test/java/org/wordpress/android/ui/uploads/UploadActionUseCaseTest.kt#L201-L217).

shiki:
Closes #12324. 

This adds auto-uploading support for published and draft posts that were previously uploaded to the server. The main change is the removal of [`!post.hasRemote()` in `PostAutoUploadInteractor`](https://github.com/wordpress-mobile/WordPress-iOS/commit/873bd85000c3970bc20d4c5fafe61fdbd0ceb10f#diff-ebdb8497b6444c39da3f2b426dc2a78cL29-L30). 

This also incorporates the related changes as described in the pending tasks in Android:

- https://github.com/wordpress-mobile/WordPress-Android/issues/10465 - Show “Cancel” after editing existing drafts. 
- https://github.com/wordpress-mobile/WordPress-Android/issues/10461#issuecomment-526843789 - Show “Draft will be uploaded next time your device is online” for drafts.
- https://github.com/wordpress-mobile/WordPress-Android/issues/10463 - Show “Changes will be published...” for existing posts. I needed some confirmation on this so the message is the temporary ”Changes will be upload next time your device is online”. We can update this in a separate PR.  
 
The rest of the changes are automated tests. 

### Local draft

Before | After 
--------|-------
before local draft  |  local draft

1. Go offline. 
2. Create a post. 
3. Tap on **X** and tap _Save Draft_.
3. Confirm the Post List and the notice says “Draft will be uploaded next time your device is online”
4. Go online. Confirm the post is automatically uploaded. 

### Editing an existing draft

Before | After 
--------|-------
before existing draft     |     existing draft

1. Go offline. 
2. Edit a draft that has been previously uploaded to the server. Make some changes.
3. Tap on **X** and tap _Update Draft_.
3. Confirm the Post List and the notice says “Changes will be uploaded next time your device is online”
4. Go online. Confirm the draft is **automatically uploaded**. 

### Publishing a local draft 

Before | After 
--------|-------
before published draft        |       published draft

1. Go offline.
2. Create a post. 
3. Tap on _Publish_.
4. Confirm the Post List and notice says “Post will be published next time your device is online”
5. Go online. Confirm the post is **automatically published**.

### Editing an existing published post

Before | After 
--------|-------
before existing published post  |       existing published post

1. Go offline. 
2. Edit a published post that has been previously uploaded to the server. Make some changes.
3. Tap on **X** and tap _Update Post_.
3. Confirm the Post List and the notice says “Changes will be uploaded next time your device is online”
4. Go online. Confirm the draft is **automatically published**. 

## Reviewing 

## Submitter Checklist

## Tasks

- [ ] Fix _Update_ button marking the post as confirmed https://github.com/wordpress-mobile/WordPress-iOS/pull/12466#issuecomment-529694557

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
@@ -71,9 +71,9 @@ class PostCardStatusViewModel: NSObject {
 
             switch postStatus {
             case .draft:
-                return canCancelAutoUpload ? StatusMessages.draftWillBeUploaded : StatusMessages.localChanges
+                return canCancelAutoUpload ? StatusMessages.changesWillBeUploaded : StatusMessages.draftWillBeUploaded
             case .publish:
-                return StatusMessages.postWillBePublished
+                return post.hasRemote() ? StatusMessages.changesWillBeUploaded : StatusMessages.postWillBePublished
             default:
                 return defaultFailedMessage
             }
@@ -240,5 +240,7 @@ class PostCardStatusViewModel: NSObject {
         static let localChanges = NSLocalizedString("Local changes", comment: "A status label for a post that only exists on the user's iOS device, and has not yet been published to their blog.")
         static let draftWillBeUploaded = NSLocalizedString("Draft will be uploaded next time your device is online",
                                                            comment: "Message shown in post list when a draft is scheduled to be automatically uploaded.")
+        static let changesWillBeUploaded = NSLocalizedString("Changes will be uploaded next time your device is online",
+                                                             comment: "Message shown in post list when a post's local changes will be automatically uploaded when the device is online.")
     }
 }
--- a/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
@@ -68,12 +68,15 @@ class PostCardStatusViewModel: NSObject {
             if autoUploadInteractor.autoUploadAction(for: post) != .upload {
                 return defaultFailedMessage
             }
+            if post.hasRemote() {
+                return StatusMessages.changesWillBeUploaded
+            }
 
             switch postStatus {
             case .draft:
-                return canCancelAutoUpload ? StatusMessages.changesWillBeUploaded : StatusMessages.draftWillBeUploaded
+                return StatusMessages.draftWillBeUploaded
             case .publish:
-                return post.hasRemote() ? StatusMessages.changesWillBeUploaded : StatusMessages.postWillBePublished
+                return StatusMessages.postWillBePublished
             default:
                 return defaultFailedMessage
             }
--- a/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
@@ -71,7 +71,7 @@ class PostCardStatusViewModel: NSObject {
 
             switch postStatus {
             case .draft:
-                return StatusMessages.draftWillBeUploaded
+                return canCancelAutoUpload ? StatusMessages.draftWillBeUploaded : StatusMessages.localChanges
             case .publish:
                 return StatusMessages.postWillBePublished
             default:
--- a/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
@@ -213,11 +213,9 @@ class PostCardStatusViewModel: NSObject {
     ///
     /// This is a helper method for `status`.
     private func generateFailedStatusMessage() -> String {
-        assert(post.remoteStatus == .failed, "This should only be used if the remoteStatus is .failed")
-
         let defaultFailedMessage = StatusMessages.uploadFailed
 
-        guard post.remoteStatus == .failed, let postStatus = post.status else {
+        guard post.isFailed, let postStatus = post.status else {
             return defaultFailedMessage
         }
 
--- a/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
@@ -56,15 +56,27 @@ class PostCardStatusViewModel: NSObject {
         if MediaCoordinator.shared.isUploadingMedia(for: post) {
             return NSLocalizedString("Uploading media...", comment: "Message displayed on a post's card while the post is uploading media")
         } else if post.isFailed {
-            if canCancelAutoUpload {
-                return StatusMessages.postWillBePublished
+            let defaultFailedMessage = StatusMessages.uploadFailed
+            guard let postStatus = post.status else {
+                return defaultFailedMessage
             }
 
             if post.wasAutoUploadCancelled {
                 return StatusMessages.localChanges
             }
 
-            return StatusMessages.uploadFailed
+            if autoUploadInteractor.autoUploadAction(for: post) != .upload {
+                return defaultFailedMessage
+            }
+
+            switch postStatus {
+            case .draft:
+                return StatusMessages.draftWillBeUploaded
+            case .publish:
+                return StatusMessages.postWillBePublished
+            default:
+                return defaultFailedMessage
+            }
         } else if post.remoteStatus == .pushing {
             return NSLocalizedString("Uploading post...", comment: "Message displayed on a post's card when the post has failed to upload")
         } else {
@@ -98,7 +110,8 @@ class PostCardStatusViewModel: NSObject {
         }
 
         if post.isFailed {
-            return (canCancelAutoUpload || post.wasAutoUploadCancelled) ? .warning : .error
+            let autoUploadAction = autoUploadInteractor.autoUploadAction(for: post)
+            return (autoUploadAction == .upload || post.wasAutoUploadCancelled) ? .warning : .error
         }
 
         switch status {
@@ -225,5 +238,7 @@ class PostCardStatusViewModel: NSObject {
         static let postWillBePublished = NSLocalizedString("Post will be published next time your device is online",
                                                            comment: "Message shown in the posts list when a post is scheduled for publishing")
         static let localChanges = NSLocalizedString("Local changes", comment: "A status label for a post that only exists on the user's iOS device, and has not yet been published to their blog.")
+        static let draftWillBeUploaded = NSLocalizedString("Draft will be uploaded next time your device is online",
+                                                           comment: "Message shown in post list when a draft is scheduled to be automatically uploaded.")
     }
 }
--- a/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
@@ -56,30 +56,7 @@ class PostCardStatusViewModel: NSObject {
         if MediaCoordinator.shared.isUploadingMedia(for: post) {
             return NSLocalizedString("Uploading media...", comment: "Message displayed on a post's card while the post is uploading media")
         } else if post.isFailed {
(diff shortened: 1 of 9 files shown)


shiki:
@osullivanchris I thought I'd send this to you first before I ask someone to review the code. Would you mind taking a quick look at the description and the screenshots above? Please let me know what you think. 🙂

leandroalonso:
@shiki When I tap "Update" (the right-upper button) on a published post, it shows me an alert containing an "Error occurred during saving".

If I close the app and open it again I can see that the content was saved and the post was scheduled to be auto-published. Do you think this is something to be fixed inside the scope of this task?

WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
```diff
@@ -56,15 +56,30 @@ class PostCardStatusViewModel: NSObject {
...
 
-            return StatusMessages.uploadFailed
+            if autoUploadInteractor.autoUploadAction(for: post) != .upload {
+                return defaultFailedMessage
+            }
+            if post.hasRemote() {
+                return StatusMessages.changesWillBeUploaded
+            }
+
+            switch postStatus {
```
leandroalonso: This method is getting bigger, maybe we can extract the failure status to its own property? Maybe `failedStatus` or something like that.

WordPress/WordPressTest/PostCardCellTests.swift
```diff
@@ -263,9 +264,21 @@ class PostCardCellTests: XCTestCase {
         XCTAssertFalse(postCell.separatorLabel.isHidden)
     }
 
-    func testShowsWarningMessageForFailedPublishedPosts() {
+    func testShowsChangesWillBeUploadedWarningForFailedPublishedPostsWithRemote() {
         // Given
-        let post = PostBuilder().published().with(remoteStatus: .failed).confirmedAutoUpload().build()
+        let post = PostBuilder(context).published().withRemote().with(remoteStatus: .failed).confirmedAutoUpload().build()
+
+        // When
```
leandroalonso: I think we can get rid of the Given/When/Then comments, I find them more intrusive than helpful. If our team agrees that we're following this pattern the line break should be enough. What do you think?

WordPress/WordPressTest/Services/PostAutoUploadInteractorTests.swift
```diff
@@ -31,10 +31,12 @@ class PostCoordinatorUploadActionUseCaseTests: XCTestCase {
             createPost(.publish, confirmedAutoUpload: true): .upload,
             // Published local drafts with no confirmation will be remote auto-saved
             createPost(.publish): .autoSave,
-            // Draft and published posts with remote are currently unsupported. This will be
-            // fixed soon.
-            createPost(.draft, hasRemote: true): .nothing,
-            createPost(.publish, hasRemote: true): .nothing,
+            // Posts with remote that do not have confirmation will be remote auto-saved
```
leandroalonso: I completely understand the desire of adding more test cases with less-effort but in this specific case, I had to understand the *logic* of this test in order to understand it.

In general, I'm in favor of "DAMP (Descriptive And Meaningful Phrases) not DRY (Don't repeat yourself)" in tests. Usually, I thought of the tests as documentation, and I'd like the reader to understand what the test is doing as quickly as possible, without thinking too much — and if there's duplicated code in the process I really don't mind.

Take the first case as an example, it will be like this without the parametrization:

```
    func testReturnsUploadAsActionForDraftedPost() {
        let post = createPost(.draft)

        let action = interactor.autoUploadAction(for: post)

        expect(action).to(equal(.upload))
    }
```

What do you think? I found it much more expressive.

We would end up with a lot of test functions? Probably. But I tend to find them more meaningful than the comments.

Anyway, this is just an addition to an existed test. So, I don't think we need to change it right now. We can save for a discussion in the near future. :)

WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
```diff
@@ -7,6 +7,7 @@ enum PostNoticeUserInfoKey {
 struct PostNoticeViewModel {
     private let post: AbstractPost
     private let postCoordinator: PostCoordinator
+    private let autoUploadInteractor = PostAutoUploadInteractor()
```
leandroalonso: By reading `PostAutoUploadInteractor` I assumed that this `Interactor` was responsible for auto-uploading the post, which doesn't seem to be what is really doing.

Anyway, I couldn't figure out any other better name... ¯\_(ツ)_/¯

WordPress/WordPressTest/Services/PostAutoUploadInteractorTests.swift
shiki: I think we've discussed this in the previous PR and I understand your argument. 

> I had to understand the logic of this test in order to understand it

It could be that I'm used to parameterized tests that I don't notice this. I'm curious to know how you _read_ this code and made it difficult to understand. Was it the use of dictionary? Was it the variable name (`postsAndExpectedActions`) that was not clear? Which logic did you find was difficult to understand? Is there something that we can do to make it easier for you? 

What did you think of the `Expectation` struct? I concocted that with you in mind. I knew you were gonna comment on this part. 🙂 

https://github.com/wordpress-mobile/WordPress-iOS/blob/2c6380a68b583dafe3eba8bb7cccaa69790bbf5b/WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift#L28-L34

> We would end up with a lot of test functions? Probably. But I tend to find them more meaningful than the comments.

That's true. If I was going to write this that way, I might still end up making a code generator for it. 😉 Or what do you think about grouping them by statuses? One method for `.draft` cases, and another for `.publish`, and so on.

WordPress/WordPressTest/PostCardCellTests.swift
shiki: If the `Given` statement is just one line or the test method is short, it's fine to leave it out. But as the lines get longer, it's harder to see which line is the part being tested. I think it's important in this case:

https://github.com/wordpress-mobile/WordPress-iOS/blob/2c6380a68b583dafe3eba8bb7cccaa69790bbf5b/WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift#L97-L115

The pattern acts as a reminder not to test too many scenarios in one method. And also, just to organize the statements more clearly. 

> If our team agrees that we're following this pattern

That could be added to the team contract. 😉

WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
shiki: Yep. Here's a history of its naming: https://github.com/wordpress-mobile/WordPress-iOS/pull/12317#discussion_r314378334. 😁

WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
```diff
@@ -56,15 +56,30 @@ class PostCardStatusViewModel: NSObject {
...
+            guard let postStatus = post.status else {
+                return defaultFailedMessage
             }
 
             if post.wasAutoUploadCancelled {
                 return StatusMessages.localChanges
             }
 
-            return StatusMessages.uploadFailed
+            if autoUploadInteractor.autoUploadAction(for: post) != .upload {
```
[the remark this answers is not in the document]
shiki: Thank you for investigating, @leandroalonso. The `shouldAttemptAutoUpload` returns `false`, that's why the action is `.autoSave` instead of `.upload`. The `shouldAttemptAutoUpload` is initially set to `true` here: 

https://github.com/wordpress-mobile/WordPress-iOS/blob/2c6380a68b583dafe3eba8bb7cccaa69790bbf5b/WordPress/Classes/ViewRelated/Post/PostEditor%2BPublish.swift#L87

However, something in our code changed the `status` to `.failed` which resulted in the hash changing, and ultimately the `shouldAttemptAutoUpload` return value.

https://github.com/wordpress-mobile/WordPress-iOS/blob/2c6380a68b583dafe3eba8bb7cccaa69790bbf5b/WordPress/Classes/Models/AbstractPost%2BHashHelpers.m#L7-L25

This is a case we probably did not find during #12252. This is a separate issue though since it happens in the master branch too. I created https://github.com/wordpress-mobile/WordPress-iOS/issues/12498 for it.

WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
shiki: I create a method instead (adb140b380549637a0aa7cf957ef186ad5f02428). Adding another property makes it harder to determine how the class should be used. What do you think?

shiki:
@leandroalonso Thank you for the review. I've responded to your comments. This is now ready for another look.

WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
```diff
@@ -216,6 +209,39 @@ class PostCardStatusViewModel: NSObject {
         return [status, sticky].filter { !$0.isEmpty }.joined(separator: separator)
     }
 
+    /// Determine what the failed status message should be and return it.
+    ///
+    /// This is a helper method for `status`.
+    private func generateFailedStatusMessage() -> String {
+        assert(post.remoteStatus == .failed, "This should only be used if the remoteStatus is .failed")
```
leandroalonso: Somehow this failed for me when offline (there were a post uploading media):

https://cld.wthms.co/PJZ12T

And when debugging:

```
(lldb) po post.remoteStatus == .local
true
```

shiki: The assertion should have been `post.isFailed`. But I removed it anyway. We'll just rely on the `guard` instead. Done in 77369515f0c31cd21b0455390a24a80b6c6e4e9e.

shiki:
@leandroalonso 

> When I tap "Update" (the right-upper button) on a published post, it shows me an alert containing an "Error occurred during saving".

Good find here. That _Update_ button requires the device to be online and I believe we decided not to touch it and keep it out of scope. However, we shouldn't mark the post as “confirmed for auto-upload” since we're not showing any message like “Changes will be published next time your device is online”. I'll take a look.

shiki:
@leandroalonso Actually, mind if we merge this and I'll fix that in a separate PR? I have a feeling the fix will take longer and with more tests. I reckon it's going to be related to #12498 too. 😬

yaelirub:
Since his issue is blocking https://github.com/wordpress-mobile/WordPress-iOS/issues/12362 and has been approved and confirmed for merge by @leandroalonso , I am merging this.
Hope thats ok, @shiki .

WordPress/WordPressTest/Services/PostAutoUploadInteractorTests.swift
shiki: Discussed tests with Leandro earlier today. Marking this as resolved.

shiki:
@leandroalonso Looping back on https://github.com/wordpress-mobile/WordPress-iOS/pull/12466#issuecomment-529695271. I created an issue for it: https://github.com/wordpress-mobile/WordPress-iOS/issues/12589.

PR #12483, opened 2019-09-05
Add dark mode to acknowledgements

jkmassel:
Ensures that the autogenerated acknowledgements look right in dark mode. Screenshot is what it looks like:

Simulator Screen Shot - iPhone Xs - 2019-09-05 at 12 51 04


iamthomasbishop:
@jkmassel Is it just me or does the navigation bar title appear north of center vertically? Maybe it's because the navigation bars on some of these new sheets are 56pt tall? Either way, I think it would be nice to clean up.

Also, I'm not sure if it's limited to this view, but it looks like we're using a mix of brand and system grays – especially on the navigation and bottom toolbars – so the tints clash a little.

Otherwise, looks good to me!

mattmiklic:
> Is it just me or does the navigation bar title appear north of center vertically?

Yeah, would be good to fix this -- but it's like this in the current version too, fwiw, so not specific to this PR.

jkmassel:
@iamthomasbishop  - it was an easy enough fix. Also got rid of the repeated "Acknowledgements" heading since I was in there.

Simulator Screen Shot - iPhone Xʀ - 2019-09-06 at 09 54 37

iamthomasbishop:
@jkmassel awesome! Thanks for the fix! It looks aligned to me:

D701414C-1D1D-4ECE-895C-993543DE7F0D

iamthomasbishop:
Ah, maybe the text **is** 1pt south of center, but it could be something with line height?

jkmassel:
It's possible! I think this component needs adjustment for the new popover style, so it might make more sense to adjust it in that PR – the component is used elsewhere so I'd rather not risk breaking it at this point? 

WDYT?

iamthomasbishop:
Definitely makes sense to adjust separately 👍

mattmiklic:
Agreed; actually it looks pretty close to perfect to my eyes already. The modal sheet presentation is a lot nicer than the previous full-screen view.

PR #12489, opened 2019-09-06
[Gutenberg] Autosave

etoledom:
This PR implements the Gutenberg autosave mechanism replacing the previously implemented client-side autosave.

**Update:** The new logic implemented is described on https://github.com/wordpress-mobile/WordPress-iOS/pull/12489#issuecomment-535434070
**Updated Gutenberg PR**: https://github.com/WordPress/gutenberg/pull/17548

- `gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1351
- `gutenberg` PR: https://github.com/WordPress/gutenberg/pull/17329
- `WPAndroid` PR: https://github.com/wordpress-mobile/WordPress-Android/pull/10483
- Original issue (Android): https://github.com/wordpress-mobile/WordPress-Android/issues/10415

**To test (updated):**
- **Run metro server** on `gutenberg-mobile:develop`
- Create/open a Gutenberg post.
- Write some content.
  - You can check if `editorDidAutosave()` is invoked on every content change.
  - Check that `debouncerCallback` is invoked after 1 second of inactivity.
  - Check that `debouncerCallback` is invoked after typing 50 characters without break.
- Kill the app and restart it again.
- Check that the post was automatically saved.

Update release notes:


WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -169,11 +163,11 @@ class GutenbergViewController: UIViewController, PostEditor {
...
     }()
 
-    /// For autosaving - The debouncer will execute local saving every defined number of seconds.
-    /// In this case every 0.5 second
+    /// For autosaving - We are now using the Gutenberg side autosaving mechanism, so we don't need
+    /// a local scheduler any longer. We keep the debouncer to reuse the shared PostEditor+Publish save code.
     ///
     fileprivate(set) lazy var debouncer: Debouncer = {
-        return Debouncer(delay: PostEditorDebouncerConstants.autoSavingDelay, callback: debouncerCallback)
+        return Debouncer(delay: 0, callback: debouncerCallback)
```
koke: By using a `Debouncer` here with a zero delay we are still using a timer. I'm not sure if this would have any unintended side effects.

I think I'm overthinking this but I'll drop the suggestion just in case, what if this was a protocol and we could have a different implementation when we wanted a false debouncer that just runs immediately?
https://gist.github.com/koke/988287b7c7125af9b88f46ac6032e37a

mchowning:
@etoledom : do we want this to target the (just created) `gb/release-1.13.0` branch?

etoledom:
> @etoledom : do we want this to target the (just created) `gb/release-1.13.0` branch?

Sounds good! Just did the change 👍

etoledom:
> I think the previous behavior was safer, as it ensured that if anything crashed you would lose at most 5 seconds of work. Now you could create a lot of content without pausing, and so without autosaving, and lose it all if the app crashes.

I see what you mean. We could always reduce the inactivity time to 1 second? I'd guess it's quite a lot less likely for someone to write a big piece of content without any one second of inactivity.

@daniloercoli - what do you think?

daniloercoli:
> I see what you mean. We could always reduce the inactivity time to 1 second? I'd guess it's quite a lot less likely for someone to write a big piece of content without any one second of inactivity.

> @daniloercoli - what do you think?

Yes, we can change it to 1 sec of inactivity, or even set a timer, or follow another logic we think is more appropriate for a mobile app with local storage. With this PR we're adding the autosaving to mobile-gb, before this PR nothing was autosaved, so I think there are room for improvement in another PR.

etoledom:
> With this PR we're adding the autosaving to mobile-gb, before this PR nothing was autosaved, so I think there are room for improvement in another PR.

We did have a timer on WPiOS requesting the current content to save it. That's the reason of @koke's comment [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/12489#pullrequestreview-284775918).

> we can change it to 1 sec of inactivity, or even set a timer, or follow another logic we think is more appropriate for a mobile app with local storage.

Is nice that we have that flexibility. I would say to merge this change on WPiOS as it is now to use the same mechanism as WPAndroid, and as @daniloercoli mentioned, keep improving it. 👍 

What do you think @koke ?

WordPress/Classes/ViewRelated/Post/Autosaver.swift
```diff
@@ -0,0 +1,38 @@
...
+        changesCount = 0
+        action()
+    }
+
+    /// Instantiates an instance of Autosaver
+    /// - Parameter changesThreshold: Ammount of changes allowed before autosaving. Default 50 changes.
+    /// - Parameter delay: Maximum time of inactivity before autosaving. Default 1 second.
+    /// - Parameter action: The action to be triggered when autosave is fired.
+    ///
+    init(changesThreshold: Int = 50, delay: Double = 1, action: @escaping () -> Void) {
```
etoledom: Defaults are taken from WPAndroid.

etoledom:
**Update:**

We have decided to implement a different approach for autosaving. This is to send content change events from Gutenberg, and let the native apps decide when is best to do an autosave.

We also changed the logic to decide when is best to autosave:
- We will wait for 1s of inactivity, or
- We will autosave after 50 continuous changes.

This new autosave logic has been implemented in both Gutenberg and Aztec, matching the logic on WPAndroid.

cc @daniloercoli @maxme

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
etoledom: This change is not needed anymore 👍

PR #12503, opened 2019-09-13
Update CircleCI for Xcode 11

jkmassel:
This PR fixes CircleCI with Xcode 11 (it used to be for re-enabling ZenDesk, but there was a lot of other stuff required for that)

**To test:**
- Checkout the branch and install pods
- Build and run (this will ensure that the pods installed right)
- Ensure that tests pass (this will ensure that we have CircleCI set up for XC11)

**Update release notes:**


Gemfile.lock
```diff
@@ -266,4 +266,4 @@ DEPENDENCIES
   xcpretty-travis-formatter!
 
 BUNDLED WITH
-   1.17.3
+   2.0.2
```
jtreanor: Should this change be made separately or is there a reason for this to happen now?

.ruby-version
```diff
@@ -1 +1 @@
-2.3.0
+2.5.5
```
jtreanor: We should communicate the potential implications of this, but the change is 👌

Gemfile.lock
jkmassel: `gem install bundler` will by default install Bundler 2 if we update ruby, which I think we have to do because of the CircleCI image. I figure given that fact, it makes sense to do this now, since we're in the same building with folks for a couple of days to help with fixes if needed? 😅

jtreanor: I guess, and we definitely want to upgrade but this seems like the kind of change which could cause issues 🤔

Anyway, if you’re comfortable with it, let’s go ahead.

PR #12510, opened 2019-09-14
Remove Xcode 11 ifdefs

jkmassel:
Removes all the XC11 conditionals from the app

**To test:**
- Ensure the tests are green
- Ensure the code looks good

**Please Note:** 

- This PR does not include changes to WPAuthenticator to remove the conditionals. That'll be changed in WPAuth as a PR, then included here in a forthcoming PR
- UI Tests are not expected to pass in this PR, but it should still turn green since they're not required.

Update release notes:


Podfile
```diff
@@ -191,23 +191,7 @@ target 'WordPress' do
     end
 
     
-    post_install do
-      
-        ## Append SDKVersions.xcconfig contents to WordPressAuthenticator config files
-        Dir.glob("Pods/Target Support Files/WordPressAuthenticator/*.xcconfig") do |xc_config_filename|
```
loremattei: From the PR description, it looks like we would still need it because of WPAuthenticator, but looking at the code it seems that the new version without conditionals has been already imported, right?

frosty: Right, if any pods are still using `#if XCODE11`, those will now fail because the condition will flip to false because we'll have removed the definition of Xcode 11.

PR #12511, opened 2019-09-14
Fix UI tests for iOS 13

jkmassel:
**To test:**
- Ensure UI Tests pass

**Update release notes:**

### DIFF
--- a/WordPress/WordPressUITests/Screens/BaseScreen.swift
+++ b/WordPress/WordPressUITests/Screens/BaseScreen.swift
@@ -53,7 +53,7 @@ class BaseScreen {
             safari.launch()
 
             // Select the URL bar when Safari opens
-            let urlBar = safari.otherElements["URL"]
+            let urlBar = safari.textFields.firstMatch
             waitFor(element: urlBar, predicate: "exists == true")
             urlBar.tap()
 
@@ -62,7 +62,7 @@ class BaseScreen {
             magicLinkComponents.path = "/magic-link"
             magicLinkComponents.queryItems = [URLQueryItem(name: "scheme", value: "wpdebug")]
 
-            safari.textFields["URL"].typeText("\(magicLinkComponents.url!.absoluteString)\n")
+            urlBar.typeText("\(magicLinkComponents.url!.absoluteString)\n")
 
             // Accept the prompt to open the deep link
             safari.scrollViews.element(boundBy: 0).buttons.element(boundBy: 1).tap()
--- a/WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
+++ b/WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
@@ -16,12 +16,11 @@ extension XCUIElement {
      */
     func clearTextIfNeeded() -> Void {
         let app = XCUIApplication()
-        let content = self.value as! String
 
-        if content.count > 0 && content != self.placeholderValue {
-            self.press(forDuration: 1.2)
-            app.menuItems.element(boundBy: 1).tap() // Select All (in the text edit menu)
-            app.menuItems.element(boundBy: 0).tap() // Cut (in the text edit menu)
+        self.tap()
+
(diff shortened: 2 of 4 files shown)


WordPress/WordPressUITests/Screens/BaseScreen.swift
```diff
@@ -53,7 +53,7 @@ class BaseScreen {
             safari.launch()
 
             // Select the URL bar when Safari opens
-            let urlBar = safari.otherElements["URL"]
+            let urlBar = safari.textFields.firstMatch
```
rachelmcr: Could we switch this to `let urlBar = safari.textFields["URL"]`? That'll be more precise, and avoid any issues with local runs where there might be another text field on the screen that could be selected instead of the URL field.

WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
```diff
@@ -16,12 +16,11 @@ extension XCUIElement {
...
-        let content = self.value as! String
 
-        if content.count > 0 && content != self.placeholderValue {
-            self.press(forDuration: 1.2)
-            app.menuItems.element(boundBy: 1).tap() // Select All (in the text edit menu)
-            app.menuItems.element(boundBy: 0).tap() // Cut (in the text edit menu)
+        self.tap()
+
+        for _ in 0...(self.stringValue?.count ?? 0) {
+            app.keys["delete"].tap()
```
rachelmcr: I noticed that deleting each character individually is fairly slow — what do you think of something like this instead:

```
if (self.stringValue?.count ?? 0) > 0 {
    self.press(forDuration: 1.2)
    app.keys["delete"].tap()
}
```

rachelmcr: Note that my suggestion is a full replacement of the changes here, so no `self.tap()` first — the long press without tapping on the text field will select all the text.

WordPress/WordPressUITests/Screens/BaseScreen.swift
jkmassel: This doesn't work in iOS 13 – it looks like they changed how the view hierarchy works. The test just stalls out on this line because it can't seem to find a match :(

rachelmcr: Hm, it works on my end — `otherElements["URL"]` doesn't work anymore but using `textFields["URL"]` everywhere instead was working for me.

WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
jkmassel: This works much better! :) I'll update the PR!

WordPress/WordPressUITests/Screens/BaseScreen.swift
rachelmcr: It's possible there's another factor I didn't notice that's making that work, though. If it doesn't work for you, I'm ok with your change here. It shouldn't ever be an issue on CircleCI in any case and it's pretty obvious what's happening if it fails on a local test run.

jkmassel: Ah – I might've been doing something wrong – I think I was still using `otherElements`. Updated this to your suggestion and it works locally – we'll wait for the ✅!

jkmassel: So weirdly, this breaks on iPad. In order to get the release moving, I'd like to merge this PR with the less-good version, then address this after, if that's ok with you?

jkmassel:
Admin merging for this to make the release

WordPress/WordPressUITests/Screens/BaseScreen.swift
rachelmcr: Sorry for the delay getting back to you about this one — I'll investigate and follow up on this.

PR #12512, opened 2019-09-14
Try/improve context saves

aerych:
This PR does two things:
- First it changes calls to `saveDerivedContext` to `saveContext` instead.  This will ensure that changes are propagated correctly.  The saveDerivedContext methods are no available.
- Second it changes the saveContext methods so calling saveContextAndWait with a derived context will be properly handled, and that performBlockAndWait is available when saving derived contexts. 

I'm hopeful that these changes will help sort out any issues with saving derived contexts that we haven't been able to track down.

I'm opening this PR as a draft for some early feedback before I sort out any tests that are broken by the changes.  @koke @astralbodies @frosty @jleandroperez, I would love your 👀  on it if you have some time. 

To test:
This is likely an impactful change.  I'm not sure how to best test beyond just heavy usage of features that relied on derived contexts and/or saveContextAndWait.  Media, Categories, Posts, Comments, Reader post list, etc.

Update release notes:


WordPress/Classes/Utility/ContextManager.m
```diff
@@ -103,63 +103,77 @@ - (NSManagedObjectContext *const)newChildContextWithConcurrencyType:(NSManagedOb
...
+        }];
+    }
+}
+
+- (void)saveDerivedContext:(NSManagedObjectContext *)context andWait:(BOOL)wait withCompletionBlock:(void (^)(void))completionBlock
+{
+    if (wait) {
+        [context performBlockAndWait:^{
+            [self internalSaveDerivedContext:context];
+            [self saveContext:self.mainContext andWait:wait withCompletionBlock:completionBlock];
```
koke: In the previous implementation, the completion block was called after saving the derived context and then the main context would save asynchronously. Was this change intentional?

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -103,63 +103,77 @@ - (NSManagedObjectContext *const)newChildContextWithConcurrencyType:(NSManagedOb
...
+            [self internalSaveContext:context withCompletionBlock:completionBlock];
+        }];
+    } else {
+        [context performBlock:^{
+            [self internalSaveContext:context withCompletionBlock:completionBlock];
+        }];
+    }
+}
+
+- (void)saveDerivedContext:(NSManagedObjectContext *)context andWait:(BOOL)wait withCompletionBlock:(void (^)(void))completionBlock
```
koke: I know this isn't exposed outside this file, but does it makes sense to have a method that allows both `wait == YES` and `completionBlock != nil`?

WordPress/Classes/Utility/ContextManager.m
aerych: It was. I'm thinking about cases where work is being done on a derived context, then subsequent work is being done immediately on the main context.  I thought it made sense to wait for the main to be saved before executing the callback.

jleandroperez:
I know this may be beyond the scope of the PR, **but** I've noticed both `internalSaveDerivedContext` and `internalSaveContext` have the exact same code (few minor differences apart). Maybe we could simplify and keep just one?. 

This is, OFC, unless I'm missing something!

Taking the app for a test drive, will report back shortly. Thanks for doing this Eric!!

aerych:
There is currently one test failing on ci.  It _was_ passing locally until I looked closer at it and realized that the test was passing despite a hidden core data crash.  Debugging this has been tricky -- multiple threads, unfamiliarity with the underlying code in PostCoordinator and MediaCoordinator, etc.  Consider it a blocker until it can be sorted.

WordPress/Classes/Utility/ContextManager.m
aerych: In that it allows the logic to be grouped into a single method to handle it vs scattered between several similar methods, I think so.  Open to other ideas tho!

aerych:
Still seeing a failure with the `testDoNotUploadAPostWithFailedMedia` test.  Its ... weird.   This is what I'm seeing when I try to debug:

1 Switch to the current develop.
2 Clean the build folder.
3 Launch the app in the simulator. 
4 Stop the simulator.
5 Run the `testDoNotUploadAPostWithFailedMedia` test. 
6 Note that the test passes. 🎉 
7 Run the test again.  Note that the test fails 🙃  
8 Run the test again to be sure. Yep, still fails. 
9 Launch the simulator again. 
10 Close the simulator and run the test. 
11 Note that the test passes. 🎉 
12 Run the test again. Note that the test fails again. 😢 
Go to step 8.

Can anyone reproduce this?  (cc @leandroalonso, any thoughts?)

jkmassel:
I have a consistently failing (AFAICT) over here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12514. We've seen this test being flaky a lot lately. It may or may not be a microcosm of any architecture changes that end up being part of this PR.

aerych:
With #12514 resolved and merged (h/t @jkmassel, @leandroalonso, @shiki) I've revisited this one and updated a few places where we were saving a context directly vs through the ContextManager.  I'm sure there are more, these are just the ones I found when some tests broke due to missing Blog relationships (h/t Koke for the extra core data error reporting you and Nate worked on recently). I've updated PostBuilder to assign a blog to the posts/media it wrangles. 

One Koke's recent threading PR lands I'll check this one again, and if good, I'll mark it ready for review.

WordPress/WordPressTest/PostBuilder.swift
```diff
@@ -6,12 +6,25 @@ class PostBuilder {
 
     var post: Post!
 
+    private static func bulidPost(context: NSManagedObjectContext) -> Post {
```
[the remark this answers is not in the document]
aerych: doh!

aerych:
Thanks @astralbodies!  
@koke, any final thoughts or do you think we're good to merge?

koke:
I don't have anything to add. My brain is too fried on a Friday to give this another full review, but it has two approvals already so I'll trust that 🙈

aerych:
Thanks everyone!  Merging up!

PR #12514, opened 2019-09-16
Towards fixing a flaky test

jkmassel:
This one has been failing here and there for a while.

Thanks @aerych for the ping to look at it.  IMHO calling `postCoordinator.save` ought to either:
- Have persisted its changes by the time it returns _or_
- Provide a callback that lets the caller know exactly when the save operation is complete.

Personally, I'd prefer the first approach, but right now it seems like the timing in this save behaviour is non-deterministic, which makes me wonder if it's part of / related to some of the Core Data instability we've seen recently.

Assuming this should be synchronous, I think we should refactor the tests not to use `toEventually(equal()) ` and instead should just use `to(equal())`.

**To test:** (not ready yet)
- Ensure tests are green in GitHub
- Erase Content + Settings in the Simulator then run the test suite
- Run the test suite a second time
- Re-run the tests a few times in CircleCI to ensure they're not flaky there 

**Update release notes:**

[Edit by @leandroalonso: The fix was made in a different way, explained here: https://github.com/wordpress-mobile/WordPress-iOS/pull/12514#issuecomment-533641629]

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -14,9 +14,7 @@ class PostCoordinator: NSObject {
 
     private var observerUUIDs: [AbstractPost: UUID] = [:]
 
-    private lazy var mediaCoordinator: MediaCoordinator = {
-        return MediaCoordinator.shared
-    }()
+    private var mediaCoordinator: MediaCoordinator
 
     private let backgroundService: PostService
 
@@ -24,7 +22,7 @@ class PostCoordinator: NSObject {
 
     // MARK: - Initializers
 
-    init(mainService: PostService? = nil, backgroundService: PostService? = nil) {
+    init(mainService: PostService? = nil, backgroundService: PostService? = nil, mediaCoordinator: MediaCoordinator? = nil) {
         let contextManager = ContextManager.sharedInstance()
 
         let mainContext = contextManager.mainContext
@@ -36,6 +34,7 @@ class PostCoordinator: NSObject {
 
         self.mainService = mainService ?? PostService(managedObjectContext: mainContext)
         self.backgroundService = backgroundService ?? PostService(managedObjectContext: backgroundContext)
+        self.mediaCoordinator = mediaCoordinator ?? MediaCoordinator.shared
     }
 
     // MARK: - Uploading Media
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -14,7 +14,7 @@ class PostCoordinator: NSObject {
 
     private var observerUUIDs: [AbstractPost: UUID] = [:]
 
-    private var mediaCoordinator: MediaCoordinator
+    private let mediaCoordinator: MediaCoordinator
 
     private let backgroundService: PostService
 
--- a/WordPress/WordPressTest/PostCoordinatorTests.swift
+++ b/WordPress/WordPressTest/PostCoordinatorTests.swift
@@ -20,15 +20,17 @@ class PostCoordinatorTests: XCTestCase {
 
     func testDoNotUploadAPostWithFailedMedia() {
         let postServiceMock = PostServiceMock()
-        let postCoordinator = PostCoordinator(mainService: postServiceMock, backgroundService: postServiceMock)
+        let mediaCoordinatorMock = MediaCoordinatorMock()
+        let postCoordinator = PostCoordinator(mainService: postServiceMock, backgroundService: postServiceMock, mediaCoordinator: mediaCoordinatorMock)
         let post = PostBuilder(context)
             .with(image: "test.jpeg", status: .failed)
             .with(remoteStatus: .local)
             .build()
+        mediaCoordinatorMock.mediaState = .failed(error: NSError())
 
         postCoordinator.save(post)
 
-        expect(post.remoteStatus).to(equal(.failed))
+        expect(postServiceMock.didCallMarkAsFailedAndDraftIfNeeded).toEventually(beTrue())
         expect(postServiceMock.didCallUploadPost).to(beFalse())
     }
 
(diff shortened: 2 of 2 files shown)


aerych:
Some observations I can add to this:

- While [troubleshooting](https://github.com/wordpress-mobile/WordPress-iOS/pull/12512#issuecomment-531812733) to understand how PostCoordinator does its thing I noticed that a break-point in the [the callback handling](https://github.com/wordpress-mobile/WordPress-iOS/blob/13.2.1/WordPress/Classes/Services/PostCoordinator.swift#L75) the `.failed` status is _never_ reached when the tests fail.   
- Setting the time out to 100 seconds did not make a difference in the behavior, suggesting its not anything attributable strictly to timing.

aerych:
I'm about convinced the cause is due to race conditions.  There are a number of async operations in the code that probably _need_ to happen synchronously--and probably do in a real-world situation with network latency--but can occur out of the intended order in tests. 

I think the expectation is:
- PostCoordinator.save is called
- Observers are registered with MediaCoordinator
- MediaCoordinator attempts to upload/save the media.
- Success or failure is dispatched.
- MediaCoordinator notifies its observers.
- Status is updated.

However, what I'm seeing is that the upload/save fails and the event is dispatched before the PostCoordinator is registered as an observer.

To better illustrate what I'm seeing, in a clean `develop` branch I added a bunch of `NSLog`s throughout the PostCoordinator and MediaCoordinator classes to report when things were happening.  [diff](https://gist.github.com/aerych/01fc6f41c42d0bba50044f4a148a07d2)

```
2019-09-16 17:13:33.115163-0400 WordPress[43446:3179216] Launching with XCTest injected. Preparing to run tests.
2019-09-16 17:13:33.553846-0400 WordPress[43446:3179216] Waiting to run tests until the app finishes launching.
Test Suite 'Selected tests' started at 2019-09-16 17:13:33.705
Test Suite 'WordPressTest.xctest' started at 2019-09-16 17:13:33.706
Test Suite 'PostCoordinatorTests' started at 2019-09-16 17:13:33.706
Test Case '-[WordPressTest.PostCoordinatorTests testDoNotUploadAPostWithFailedMedia]' started.
2019-09-16 17:13:33.746471-0400 WordPress[43446:3179216] POST COORDINATOR - SAVE
2019-09-16 17:13:33.753577-0400 WordPress[43446:3179216] MEDIA COORDINATOR - TELLS SERVICE TO UPLOAD MEDIA
2019-09-16 17:13:33.758322-0400 WordPress[43446:3179216] MEDIA COORDINATOR - NOTIFIY OBSERVERS FOR MEDIA OF STATE CHANGE: Uploading
2019-09-16 17:13:33.758335-0400 WordPress[43446:3179456] MEDIA COORDINATOR - SERVICE FAILURE BLACK CALLED
2019-09-16 17:13:33.758567-0400 WordPress[43446:3179456] MEDIA COORDINATOR - NOTIFIY OBSERVERS FOR MEDIA OF STATE CHANGE: Failed: Error Domain=MediaServiceErrorDomain Code=1 "Not enough space to upload" UserInfo={NSLocalizedDescription=Not enough space to upload}
2019-09-16 17:13:33.758820-0400 WordPress[43446:3179455] OBSERVERS: []
2019-09-16 17:13:33.759031-0400 WordPress[43446:3179216] POST COORDINATOR - SAVE -- ABOUT TO ADD AN OBSERVER
2019-09-16 17:13:33.759127-0400 WordPress[43446:3179455] OBSERVERS: []
2019-09-16 17:13:33.759167-0400 WordPress[43446:3179216] MEDIA COORDINATOR - ADD OBSERVER FOR POST
2019-09-16 17:13:33.759301-0400 WordPress[43446:3179455] MEDIA COORDINATOR - ADDING OBSERVER FOR POST IN ASYNC QUEUE 
2019-09-16 17:13:33.759307-0400 WordPress[43446:3179216] POST COORDINATOR - RECEIPT UUID : C7313FE9-F2BE-4DDC-93F0-97272110FD1E
/Users/aerych/Sandbox/xcode_ws/automattic/wpios-sandbox/WordPress/WordPressTest/PostCoordinatorTests.swift:31: error: -[WordPressTest.PostCoordinatorTests testDoNotUploadAPostWithFailedMedia] : expected to eventually equal <AbstractPostRemoteStatus>, got <AbstractPostRemoteStatus>

Test Case '-[WordPressTest.PostCoordinatorTests testDoNotUploadAPostWithFailedMedia]' failed (1.060 seconds).
Test Suite 'PostCoordinatorTests' failed at 2019-09-16 17:13:34.766.
	 Executed 1 test, with 1 failure (0 unexpected) in 1.060 (1.061) seconds
Test Suite 'WordPressTest.xctest' failed at 2019-09-16 17:13:34.767.
	 Executed 1 test, with 1 failure (0 unexpected) in 1.060 (1.061) seconds
Test Suite 'Selected tests' failed at 2019-09-16 17:13:34.767.
	 Executed 1 test, with 1 failure (0 unexpected) in 1.060 (1.062) seconds
```

Notice that in this particular example the PostCoordinator receives its observer receipt after the MediaCoordinator has tried and failed to upload media. 

It looks like there are several different asynchronous operations that can contribute to the test failing (or succeeding). 
- Observers are asynchronously registered via dispatch queue.  Even tho a call to addObserver returns a UUID immediately, the UUID is asynchronously added to the list of mediaObservers at some point in the future.
- Observers are asynchronously notified of changes in a way where change events can be received by observers out of order.  I've seen failed status events received _before_ pushing status events. 
- Nimble's `expect().toEventually()` periodically polls the main thread to check the status. In a situation where status failed and status pushing events are received out of order, there is a chance Nimble will miss the failed status. 

How is this sounding? Reasonable? Am I off track?  
Assuming the issue is a race condition with the failed media I'm curious/nervous that the same behavior might occur in production when a user has no network connection. 😬

shiki:
This is fairly new. I can take a look at it when I'm back from AFK. 

@leandroalonso perhaps you may have some ideas too.

leandroalonso:
@shiki @jkmassel I can take a look at that when I'm back.

It's a super recent test, we even saw it failing at first but then it was consistently working fine.

In the meantime, I would suggest to just ignore it to keep `develop` consistent. What you all think?

leandroalonso:
@aerych 

> Assuming the issue is a race condition with the failed media I'm curious/nervous that the same behavior might occur in production when a user has no network connection. 😬

Would that mean that if we turn the internet off while running this test it should always fail? I tried that (in different simulators and with clean installs) and the test still succeeds.

aerych:
> Would that mean that if we turn the internet off while running this test it should always fail? 

I think it would mean that it *might* fail, depending on the order in which the work in the different dispatch queues are performed.  The test still passes intermittently for Circle CI and I believe Jeremy said he had a good stretch of time when he was testing where it always passed. 🤷‍♂

leandroalonso:
@aerych @jkmassel 

I tried a different approach to make sure this test is more deterministic.

I mocked `MediaCoordinator.addObserver` and simulated a failed media state. This should be enough to prevent the callback from not being called because now we ensure that it will *always* be called.

"Ok, but why the `toEventually` still present?". This is because of `PostCoordinator.change`, that uses an async `perform` block. We can change that but it will probably cause odd side-effects and I want to prevent that.

@aerych can you please confirm if that fixes the intermittency? To be honest, I was really not able to reproduce the test failing (nor in Xcode 10 and/or 11).

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -14,17 +14,15 @@ class PostCoordinator: NSObject {
 
     private var observerUUIDs: [AbstractPost: UUID] = [:]
 
-    private lazy var mediaCoordinator: MediaCoordinator = {
-        return MediaCoordinator.shared
-    }()
+    private var mediaCoordinator: MediaCoordinator
```
shiki: This can be `let` now.

PR #12523, opened 2019-09-19
Marks an uploaded draft to be auto uploaded.

### LINKED ISSUE #12362: [Auto-upload Published] Publish button in draft's success notice should auto-upload it

yaelirub:
Fixes #12362 

To test:

1. Be online
2. Create a post with title
3. Exit editor
4. Save as draft
5. Put a break point here: https://github.com/wordpress-mobile/WordPress-iOS/blob/fb953e49825704046034ee1aba673e7c682fc83b/WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift#L30
6. Go offline
7. Continue run 
8. See notice "Post was uploaded" and tap publish in notice
9. Go online
10. See that the post is auto uploaded.

Update release notes:

### DIFF
--- a/WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
+++ b/WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
@@ -157,6 +157,7 @@ class PostNoticeViewModelTests: XCTestCase {
         override func cancelAutoUploadOf(_ post: AbstractPost) {
             cancelAutoUploadOfInvocations += 1
         }
+
         override func save(post postToSave: AbstractPost) {
 
         }
--- a/WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
+++ b/WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
@@ -113,7 +113,7 @@ class PostNoticeViewModelTests: XCTestCase {
         // Then
         expect(postCoordinator.cancelAutoUploadOfInvocations).to(equal(1))
     }
-    
+
     func testFailedPublishedUploadedDraftPostsPublishButtonWillMarkForAutoUpload() {
         // Given
         let context = ContextManager.shared.mainContext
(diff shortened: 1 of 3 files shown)


yaelirub:
@shiki , for the test that I added I had to use the main context. Any suggestions on how to mock the check for shouldAttemptAutoUpload on the derived context?

WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
```diff
@@ -114,6 +114,30 @@ class PostNoticeViewModelTests: XCTestCase {
...
+    func testFailedPublishedUploadedDraftPostsPublishButtonWillMarkForAutoUpload() {
+        // Given
+        let context = ContextManager.shared.mainContext
+        let post = PostBuilder(context)
+            .drafted()
+            .with(title: "I've been drafted!")
+            .withRemote()
+            .with(remoteStatus: .sync)
+            .with(userName: "Meee")
+            .with(password: "ThisIsABadPassword")
```
leandroalonso: I was wondering if we really needed username/password in the test. I tried removing it and got a `caught "NSInternalInconsistencyException", "Invalid parameter not satisfying: username != nil"`.

But we can fix that by mocking the `save` method in `MockPostCoordinator`:

````
private final class MockPostCoordinator: PostCoordinator {
        (...)

        override func save(post postToSave: AbstractPost) {

        }
    }
```

Which I think is the behavior we expect (we're really not interested in the save logics).

This will even make these tests faster. :)

yaelirub:
Thanks for the review and the helpful suggestion, @leandroalonso 
Ready for another look

WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
```diff
@@ -135,5 +157,8 @@ class PostNoticeViewModelTests: XCTestCase {
         override func cancelAutoUploadOf(_ post: AbstractPost) {
             cancelAutoUploadOfInvocations += 1
         }
+        override func save(post postToSave: AbstractPost) {
```
leandroalonso: There's a breakline missing here, I guess.

PR #12516, opened 2019-09-18
Merge 13.2.1 into develop

jkmassel:
Merges all the changes from the `13.2.1` series into `develop`.

PR #12525, opened 2019-09-20
Fix issue with gap in NotificationsController

### LINKED ISSUE #12495: Notifications: Persistent gap at top of notifications list in iOS 13
The Notifications list regularly has a large gap at the top:

File

**The issue is persistent on iOS 13.0 and WPiOS: 13.2.1**

It seems that this is due to the prompts we display in that area (prime for push, app rating prompt) – I guess it's not getting collapsed when it should be. Also related: https://github.com/wordpress-mobile/WordPress-iOS/issues/11531

yaelirub:
Fixes #12495 

in iOS 13, traitCollectionDidChange gets called when the view is created as oppose to previous versions where it was called when added to the view hierarchy. The different timing of the call created this issue.
Apple suggestion is to perform any work involving traits in one of the layout methods
and then rely on traitCollectionDidChange for any future size class changes.

> Trait environments, such as views and view controllers, now have their traitCollection property populated with traits during initialization. These initial traits represent a prediction of the ultimate traits that the trait environment will receive when it gets added to the hierarchy. Because the traits that are populated during initialization are just a prediction, they might differ from the traits that are received once actually in the hierarchy. Therefore, when possible you should wait to perform work that uses the traitCollection until the view, or view controller’s view, has moved into the hierarchy — meaning window returns a non-nil value — so that you don’t have to throw away any work done using the predicted traits if the actual traits are different. The best time to use the traitCollection is during layout, such as inside layoutSubviews(), viewWillLayoutSubviews(), or viewDidLayoutSubviews().

For further read: https://developer.apple.com/documentation/ios_ipados_release_notes/ios_13_release_notes

To test:
Go to notifications tab and notice no gap. 

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -1293,7 +1293,6 @@ internal extension NotificationsViewController {
     }
 }
 
-
 // MARK: - Sync'ing Helpers
 //
 private extension NotificationsViewController {
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -42,6 +42,10 @@ class NotificationsViewController: UITableViewController, UIViewControllerRestor
     /// NoResults View
     ///
     private let noResultsViewController = NoResultsViewController.controller()
+    
+    /// Will be used in viewWillLayoutSubviews to ensure layout is only preformed once
+    ///
+    private var layoutForTheFirstTIme = true
 
     /// All of the data will be fetched during the FetchedResultsController init. Prevent overfetching
     ///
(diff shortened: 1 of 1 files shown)


yaelirub:
Hey @frosty , thanks so much for the quick review and for providing a much better solution. 
I applied your changes and tested.
Ready for another look (basically reviewing your own code 😄 )

jtreanor:
I have fixed the base branch here to `release/13.2.2` so that we can release a hotfix when it is merged.

jkmassel:
Howdy @jtreanor @yaelirub!

So I can see how it might not have been obvious, but https://github.com/wordpress-mobile/WordPress-iOS/pull/12515 and https://github.com/wordpress-mobile/WordPress-iOS/pull/12516 actually merged `13.2.1` _and_ `13.2.2`.

Because of this, we should be targeting a hot fix release of `13.2.3`, and rebasing this branch on that. If you check https://github.com/wordpress-mobile/WordPress-iOS/pull/12515/commits, you'll note that the four commits in question were part of `13.2.2`. 

Hope that helps

koke:
I’m still seeing this in the TestFlight 13.3.0.0 release. Did we miss it or should it make it there eventually after the hot fix is released?

jtreanor:
@koke I will be releasing a new beta today to include the fixes from the hotfix

PR #12532, opened 2019-09-23
Fix localize.py for Xcode 11 and string issues

jtreanor:
Localisation failed for the 13.3 code freeze for two reasons:

- There were some new strings that were not literals.
- Something changed in the behaviour of the command line tools for Xcode 11, causing `find` to have a trailing newline and therefore `genstrings` was looking for a file called `""`, and failing.

I have fixed both of these issues so we can proceed with the code freeze.

To test:

- Run `./Scripts/localize.py` from the root of the repo. If should succeed.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
@@ -199,8 +199,9 @@ class StatsTotalRow: UIView, NibLoadable, Accessible {
 
         switch (showDisclosure, hasChildRows) {
         case (true, true):
-            let hint = expanded ? "Expanded. Tap to collapse." : "Collapsed. Tap to expand."
-            accessibilityHint = NSLocalizedString(hint, comment: "Accessibility hint")
+            accessibilityHint = expanded
+                                    ? NSLocalizedString("Expanded. Tap to collapse.", comment: "Accessibility hint")
+                                    : NSLocalizedString("Collapsed. Tap to expand.", comment: "Accessibility hint")
         case (true, false):
             accessibilityHint = NSLocalizedString("Tap for more detail.", comment: "Accessibility hint")
         default:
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
```diff
@@ -199,8 +199,9 @@ class StatsTotalRow: UIView, NibLoadable, Accessible {
 
         switch (showDisclosure, hasChildRows) {
         case (true, true):
-            let hint = expanded ? "Expanded. Tap to collapse." : "Collapsed. Tap to expand."
-            accessibilityHint = NSLocalizedString(hint, comment: "Accessibility hint")
+            accessibilityHint = expanded
```
loremattei: Is this related to this PR?

WordPress/Classes/ViewRelated/Stats/Shared Views/TopTotalsCell.swift
```diff
@@ -336,12 +336,12 @@ extension TopTotalsCell: Accessible {
         let dataTitle = dataSubtitleLabel.text
 
         if let itemTitle = itemTitle, let dataTitle = dataTitle {
-            let description = String(format: "Table showing %@ and %@", itemTitle, dataTitle)
-            accessibilityLabel = NSLocalizedString(description, comment: "Accessibility of stats table. Placeholders will be populated with names of data shown in table.")
+            let descriptionFormat = NSLocalizedString("Table showing %@ and %@", comment: "Accessibility of stats table. Placeholders will be populated with names of data shown in table.")
```
loremattei: Is this related to this PR?

WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
jtreanor: Yes, I mentioned it in the description. This is one of the things that was causing `localize.py` to fail. I'll update the PR title to be clear that the issues were not just Xcode 11 related.

PR #12541, opened 2019-09-24
Auto-upload published posts with failed media

### LINKED ISSUE #12498: [Auto-upload Published] Auto-upload published posts with failed media
📣 Part of #12240. Changes should target `issue/12240-master-branch`. 

In #12466, we found that locally published posts are not set to be automatically uploaded. We added auto-uploading in #12252 but we seemed to have missed this case. 

## Steps to reproduce

1. Use `issue/12240-master-branch`
2. Go offline. 
3. Create a post and add an image.
4. Tap on _Publish_ in the top right and publish the post. 

Notice that the Post List says “Upload failed” instead of “Post will be published...”.

## Investigation

Described from the comment https://github.com/wordpress-mobile/WordPress-iOS/pull/12466#discussion_r321806426, the `shouldAttemptAutoUpload` returns `false`. This is why the action is `.autoSave` instead of `.upload`. The `shouldAttemptAutoUpload` is initially set to `true` here: 

https://github.com/wordpress-mobile/WordPress-iOS/blob/2c6380a68b583dafe3eba8bb7cccaa69790bbf5b/WordPress/Classes/ViewRelated/Post/PostEditor%2BPublish.swift#L87

However, something in our code changed the `status` to `.failed` which resulted in the hash changing, and ultimately the `shouldAttemptAutoUpload` return value.

https://github.com/wordpress-mobile/WordPress-iOS/blob/2c6380a68b583dafe3eba8bb7cccaa69790bbf5b/WordPress/Classes/Models/AbstractPost%2BHashHelpers.m#L7-L25

This is a case we probably did not find during #12252. Though it's worth investigating if this worked in #12252 and we somehow changed the location of when we set `shouldAttemptUpload` to true.

leandroalonso:
Fixes #12498

## To test

While offline:
- Create a new post (using Aztec)
- Add some images
- Tap publish
- Check that the message "post will be published next time..." is appearing

## Additional info

We had an issue with Aztec (the same doesn't happen with Gutenberg) that was not updating `post.content` after an image was added. This explains why sometimes we ended up with `` in the post. This fix also prevents this issue from happening again.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -2293,7 +2293,7 @@ extension AztecPostViewController {
         case .processing:
             DDLogInfo("Creating media")
         case .thumbnailReady(let url):
-            handleThumbnailURL(url, attachment: attachment)
+            handleThumbnailURL(url, attachment: attachment, then: { self.post.content = self.getHTML() })
         case .uploading:
             handleUploadStarted(attachment: attachment)
         case .ended:
@@ -2451,7 +2451,8 @@ extension AztecPostViewController {
         return videoAttachment
     }
 
-    private func handleThumbnailURL(_ thumbnailURL: URL, attachment: MediaAttachment) {
+    private func handleThumbnailURL(_ thumbnailURL: URL, attachment: MediaAttachment,
+                                    then afterCallback: (() -> ())? = nil) {
         DispatchQueue.main.async {
             if let attachment = attachment as? ImageAttachment {
                 attachment.updateURL(thumbnailURL)
@@ -2461,6 +2462,7 @@ extension AztecPostViewController {
                 attachment.posterURL = thumbnailURL
                 self.richTextView.refresh(attachment)
             }
+            afterCallback?()
         }
     }
 
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -2293,7 +2293,7 @@ extension AztecPostViewController {
         case .processing:
             DDLogInfo("Creating media")
         case .thumbnailReady(let url):
-            handleThumbnailURL(url, attachment: attachment, then: { self.mapUIContentToPostAndSave(immediate: true) })
+            handleThumbnailURL(url, attachment: attachment, savePostContent: true)
         case .uploading:
             handleUploadStarted(attachment: attachment)
         case .ended:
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -2292,7 +2292,7 @@ extension AztecPostViewController {
         case .processing:
             DDLogInfo("Creating media")
         case .thumbnailReady(let url):
-            handleThumbnailURL(url, attachment: attachment)
+            handleThumbnailURL(url, attachment: attachment, then: { self.post.content = self.getHTML() })
```
yaelirub: Shouldn't this be [weak self] in the block?

yaelirub:
Hey @leandroalonso , thanks for working on this. 
While testing this I noticed that if I edit an image after a failed upload, it shows "Upload Failed" when I publish instead of "Post will be published..."
Also, if I successfully upload an image and then fail to publish it also say "Upload failed"

here are steps for the 2 scenarios:

1. Edit failed to upload image before publish
**Steps**
    1. Be offline
    2. Create post
    3. Upload Image
    4. Wait till image fails to upload
    5. Tap on the image
    6. tap "Edit" in action sheet
    7. Add caption
    8. Tap "Done"
    9. Tap publish
   10. Post shows "Upload Failed"

aztec_post1


2. Successfully upload image and then fail to publish post
**Steps**
    1. Be online 
    2. Create a post
    3. upload image
    4. Wait till the image successfully uploaded 
    5. Be offline
    6. Tap publish 
    7. Post shows "Upload Failed"

Due to XC11 certificate issues I am unable to test these on the device so I'm using NLC on simulator to emulate connection. These issues are not too bad if the posts are actually not being uploaded when regaining connectivity but still thought I should mention and we could create issues for these if needed.

leandroalonso:
Good catch, @yaelirub!

I fixed both problems, can you give it another look?

Unfortunately, I wasn't able to add a unit test to both of the cases you mentioned. The cost would be too high and will require multiple refactorings. I prefer to not do that right now, hope this is reasonable to you.

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
leandroalonso: I don't think `weak` is necessary in this case since we want to make sure that `self` still exists by the time the closure is called I (which should be right away).

yaelirub: If you want to make sure it exists  you can do `[weak self]
if let self = self {
}`

Otherwise you're risking retention cycle

yaelirub:
Hey @leandroalonso , seems like scenario 1 is fixed. As for the second scenario, the post is not automatically uploaded when regaining connectivity. Can you please check that?

leandroalonso:
@yaelirub I didn't realize the upload was failing, apologies for that.

It should be fixed by now. Can you give it a try?

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
leandroalonso: I changed the implementation to keep the consistency of one-line calls inside this `switch` statement. Let me know what you think.

yaelirub:
Tested all the described scenarios here and they all seem to work as expected.

PR #12546, opened 2019-09-24
Accounts: Ensure authtoken is updated for existing account.

aerych:
Refs #12537

While testing #12537 I noticed that auth tokens were not updating as expected.  This PR restores the expected behavior. 

To test:
- Apply the patch
- Start logged into the app
- You want to clear your existing auth token.  One way to do this is set a break point in the `[BlogListViewController setEditing:]`, go to the My Sites tab and view the blog list.  Tap the Edit button. Then run `po [WordPressAuthenticationManager showSigninForWPComFixingAuthToken];`. 
- If you use the above method, you can pull to refresh to invoke the authenticator to fix the auth toekn.  Otherwise you can close and relaunch the app. 
- Enter your password to login when prompted by the auth flow. 
- Confirm you are properly logged in and an NSAssert is not tripped due to the account not having an auth token. 

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/AccountService.m
+++ b/WordPress/Classes/Services/AccountService.m
@@ -190,11 +190,8 @@ NSString * const WPAccountEmailAndDefaultBlogUpdatedNotification = @"WPAccountEm
 
 - (WPAccount *)createOrUpdateAccountWithUserDetails:(RemoteUser *)remoteUser authToken:(NSString *)authToken
 {
-    WPAccount *account = [self findAccountWithUserID:remoteUser.userID];
-    if (!account) {
-        NSString *username = remoteUser.username;
-        account = [self createOrUpdateAccountWithUsername:username authToken:authToken];
-    }
+    NSString *username = remoteUser.username;
+    WPAccount *account = [self createOrUpdateAccountWithUsername:username authToken:authToken];
     [self updateAccount:account withUserDetails:remoteUser];
     return account;
 }
--- a/WordPress/Classes/Services/AccountService.m
+++ b/WordPress/Classes/Services/AccountService.m
@@ -190,8 +190,16 @@ NSString * const WPAccountEmailAndDefaultBlogUpdatedNotification = @"WPAccountEm
 
 - (WPAccount *)createOrUpdateAccountWithUserDetails:(RemoteUser *)remoteUser authToken:(NSString *)authToken
 {
-    NSString *username = remoteUser.username;
-    WPAccount *account = [self createOrUpdateAccountWithUsername:username authToken:authToken];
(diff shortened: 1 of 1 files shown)


koke:
Maybe I'm just not remembering the details, but I'm not understanding why the previous code wouldn't find an account by ID.

aerych:
> why the previous code wouldn't find an account by ID.

It would, but it wasn't updating the auth token.  After sleeping on it, I think I want to make a change to this patch and keep the lookup by userID. I'll just make sure we always update the authtoken when an account found.

koke:
Ah, I get it now, I think I was actually more confused about the previously existing code than the changes.

So `createOrUpdateAccountWithUsername:authToken:` is kind of the "designated initializer" of accounts, which everything else uses. ~This made sense in the past but not so much now since I don't think the username is such a critical property of an account since we added login by email and 3rd parties.~ We are still using the username as the main way to tie an account to a password in the keychain, so I guess it's kind of important.

The current code still looks a bit confusing since we actually always want to update `authToken`. It's in that `if` block because `createOrUpdateAccountWithUsername:authToken:` would already update it so why do it twice.

However, `updateAccount:withUserDetails:` will also update the `username` a second time so why not? 😂 

```objc
- (WPAccount *)createOrUpdateAccountWithUserDetails:(RemoteUser *)remoteUser authToken:(NSString *)authToken
{
    WPAccount *account = [self findAccountWithUserID:remoteUser.userID];
    if (!account) {
        NSString *username = remoteUser.username;
        account = [self createOrUpdateAccountWithUsername:username authToken:authToken];
    }
    // Even if we find an account we should still update its token
    account.authToken = authToken;
    [self updateAccount:account withUserDetails:remoteUser];
    return account;
}
```

aerych:
> The current code still looks a bit confusing since we actually always want to update authToken. It's in that if block because createOrUpdateAccountWithUsername:authToken: would already update it so why do it twice.
> However, updateAccount:withUserDetails: will also update the username a second time so why not? 

Fair point, but where `[self updateAccount:account withUserDetails:remoteUser];` is going to update more than just the username, calling `account.authToken = authToken;` outside the if block is purely repetitive.  We save a bit of work by avoiding the checks in `[SFHFKeychainUtils storeUsername:andPassword:forServiceName:accessGroup:updateExisting:error]` by keeping it in  the if block.  It doesn't seem too weird to me to have the assignment nested. (Granted, I wrote it so I'm a bit biased.  ;) )

koke:
@frosty was suggesting we might want to do this on a hotfix instead, since that will hit everyone getting a new phone and restoring from a backup 🙈

aerych:
Thanks Koke! 

> @frosty was suggesting we might want to do this on a hotfix instead

Totes game to cherry pick this into a hotfix or release branch.

PR #12548, opened 2019-09-24
Issue/fix url selection bug

SergioEstevao:
This PR works around what it looks to be a bug on iOS 13 around link interactions in UITextView.

According to the [documentation](https://developer.apple.com/documentation/uikit/uitextviewdelegate/1649337-textview).

```
Links in text views are interactive only if the text view is selectable but noneditable. 
That is, if the value of the UITextView isSelectable property is true and the isEditable property is false.
```

Aztec textview has the isSelectable property to true but we are getting interactive links anyway.

So this PR works around it by not allowing interaction and then selecting the initial position of the URL range.

To test:
 - Open or create a post with a link on the content using the Aztec editor
 - Tap on the link
 - Make sure that Safari is not open to the page of the tapped link.

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -142,7 +142,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :tag => 'v1.11.0'
+    gutenberg :commit => '7735be55ca14bd3aaf0a43dde9bf85c2dd92d557'
 
     ## Third party libraries
     ## =====================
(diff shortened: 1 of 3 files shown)


jtreanor:
I'll let @etoledom weigh in on this with his review. Once merged I will submit the hotfix for review.

Podfile
```diff
@@ -142,7 +142,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :tag => 'v1.11.0'
+    gutenberg :commit => '7735be55ca14bd3aaf0a43dde9bf85c2dd92d557'
```
etoledom: We should probably do a small Gutenberg v1.11.1 release 🤔

Issue #12535, opened 2019-09-23
SIWA: listen to sign outs


ScoutHarris:
Add a listener to the SIWA notification when the connection is removed outside of the app.

PR #12549, opened 2019-09-24
[SIWA] Handle invalid Apple user

### LINKED ISSUE #12399: SIWA: Check for token validity on launch
Make sure the SIWA token is valid upon launch.

ScoutHarris:
Fixes #12399 
Related WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/137

This uses the WPAuth changes to:
- Save the Apple user id to the keychain.
- On app launch, get the user id from the keychain and check it's state.
- If the user id is no longer authorized, log out of the account (using the same logic as logging out from the Me tab).

To note, this only addresses the scenario in the testing steps, i.e. checking state on app launch. We're not yet actively listening for Apple ID disconnects (https://github.com/wordpress-mobile/WordPress-iOS/issues/12535). Stay tuned!

To test:

This should be tested on a real device for proper keychain handling.

- Log in / sign up with Apple.
- Disconnect in your Apple account on the device.
  - Go to iOS Settings.
- Rerun the app.
- The app should log you out.
- This message should appear in the logs: `checkAppleIDCredentialState: Unauthorized Apple ID. User signed out.`

Update release notes:


WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -429,6 +433,46 @@ extension WordPressAppDelegate {
...
+    @available(iOS 13.0, *)
+    func checkAppleIDCredentialState() {
+
+        // If not logged in, remove the Apple User ID from the keychain, if it exists.
+        guard AccountHelper.isLoggedIn else {
+            do {
+                try SFHFKeychainUtils.deleteItem(forUsername: WPAppleIDKeychainUsernameKey,
+                                                 andServiceName: WPAppleIDKeychainUsernameKey)
+            } catch {
+                DDLogDebug("Error while removing Apple User ID: \(error)")
```
frosty: Do we want to log this as an 'error'? If I'm reading correctly it could simply be that the user has never logged in using SIWA, so there just wouldn't be a keychain item for Apple so we'll be logging out "Error while removing..." on every launch.

ScoutHarris: Hey @frosty .

> user has never logged in using SIWA, so there just wouldn't be a keychain item for Apple so we'll be logging out "Error while removing..." on every launch.

That is correct. That's why I used `DDLogDebug`. But now I see that the actual message can be misleading. I can change it to be less alarming, maybe something like:
`DDLogDebug("Unable to remove Apple User ID from keychain: \(error)")`

I didn't want to _not_ log it since there could be an actual error.

ScoutHarris:
Hey @frosty . I'm going to go ahead and merge these PRs to streamline my branchiness. My next task is working with the same code, so if the log message is still not acceptable I'd be more than happy to change it in the next PR.

frosty:
Thanks @ScoutHarris! The error message is better. As an additional change, we could not log the message if it's just that no item was found (which is the main non-error state):

```
} catch let error as NSError {
    if error.code != errSecItemNotFound {
        DDLogDebug("Unable to remove Apple User ID from keychain: \(error.localizedDescription)")
    }
}
```

We can roll this into a future PR, if you think it makes sense :)

ScoutHarris:
Heya @frosty .

Yep, yep that makes sense. 🤦‍♂ I'll roll it into https://github.com/wordpress-mobile/WordPress-iOS/pull/12552. Thanks!

PR #12552, opened 2019-09-25
[SIWA] Handle Apple Credential Revoked notification

### LINKED ISSUE #12535: SIWA: listen to sign outs

ScoutHarris:
Fixes #12535 
WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/138

This uses the changes in WPAuth to:
- Start and stop listening to Apple's revoked notification.
- When received, the account is logged out.

Also, the call to check the Apple credential state is now made in `applicationDidBecomeActive`. This allows the check to be done when the app returns from background.

---
Notes:
- This should be tested on a real device as it seems the revoked notification does not work in simulators.
- The revoked notification is only triggered when switching apps via the app switcher.

To test the revoked notification:
- Sign up / log in with SIWA.
- Open app switcher (i.e. double tap the home button).
- Select iOS Settings.
- Disconnect your Apple account (see steps below).
- Open app switcher (i.e. double tap the home button).
- Select WordPress.
- You should be logged out.
- This log message should appear: 
`Apple credentialRevokedNotification received. User signed out.`

---
To test checking the credential state:
- Sign up / log in with SIWA.
- Background WordPress.
- Go to iOS Settings.
- Disconnect your Apple account (see steps below).
- Return to WordPress.
- You should be logged out.
- This log message should appear: 
`checkAppleIDCredentialState: Unauthorized Apple ID. User signed out.`

---
Steps to disconnect in your Apple account on a device.
  - Go to iOS Settings.

---
Update release notes:


ScoutHarris:
Hey @frosty .

The only thing I found is that the revoked notification doesn't seem to be triggered in simulators. I've added a note in the description. Otherwise, I am getting it when app switching.

PR #12571, opened 2019-09-27
Publish button in Post List should auto-upload the post

### LINKED ISSUE #12382: [Offline Posting] Publish button in Post List should auto-upload the post
📣 Part of #12240. This should be done after #12375 and #12352 is merged. Please target the `issue/12240-master-branch` branch. 

In #12375, we added a Publish button in the Post List. Pressing on the Publish button should mark the post as “confirmed for auto-uploading“. If offline, the post will be automatically uploaded later.

leandroalonso:
Fixes #12382

# To test

### Case 1

1. Being offline, create a new post
2. Tap publish
3. Tap Cancel button
4. Tap Publish button
5. Check that post will be published next time device is online
6. Go online and make sure post is published

### Case 2

1. Being offline, create a new post
2. Tap publish
3. Tap Cancel button
4. Go online
5. Tap Publish
6. Check that post is published right away

Update release notes:

- [x] If it's feasible, I have added unit tests (I wasn't able to add a test to the `PostListViewController`, unfortunately)

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
@@ -909,8 +909,9 @@ class AbstractPostListViewController: UIViewController,
             apost.date_created_gmt = Date()
             apost.status = .publish
             apost.shouldAttemptAutoUpload = true
-            self.uploadPost(apost)
-            self.updateFilterWithPostStatus(.publish)
+            self.uploadPost(apost) { [weak self] in
+                self?.updateFilterWithPostStatus(.publish)
+            }
         }
 
         present(alertController, animated: true)
@@ -924,8 +925,8 @@ class AbstractPostListViewController: UIViewController,
         updateFilterWithPostStatus(.draft)
     }
 
-    fileprivate func uploadPost(_ apost: AbstractPost) {
-        PostCoordinator.shared.save(apost)
+    fileprivate func uploadPost(_ apost: AbstractPost, success successCallback: (() -> ())? = nil) {
+        PostCoordinator.shared.save(apost, success: successCallback)
     }
 
     @objc func viewPost(_ apost: AbstractPost) {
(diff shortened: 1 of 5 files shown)


yaelirub:
Hey @leandroalonso , thanks for working on this. The publish button in the post card seems to immediately publish scheduled posts. Is this the expected behavior? 
If you tap publish in the editor for a scheduled post, it will add it to the scheduled list.
An auto-upload of a scheduled post will add it to the scheduled list as well. 

I'm worried that this might be confusing to the users, thinking that the post will be scheduled to be published and not immediately published.

Thoughts?

leandroalonso:
@yaelirub I think we currently do not support scheduled posts, is that right?

In that case, this is a scenario for the future. What do you think?

shiki:
To clarify https://github.com/wordpress-mobile/WordPress-iOS/pull/12571#issuecomment-536646466 more, we're focusing on getting the flow done as described in https://github.com/wordpress-mobile/WordPress-iOS/issues/12240. But that is only for drafts and published posts. These are currently ignored:

- scheduled
- private
- pending

One of the reasons is to minimize the amount of manual testing and code reviews we have to do and focus on the flow instead of the post status. Second is adding support for the other post types is mainly going to be just: 

- adding messages for them
- testing the flow (now, for all post types) -- which we're supposed to do after everything is done anyway. 

These post types will be added in https://github.com/wordpress-mobile/WordPress-iOS/issues/12227. I was hoping that this strategy would save us some time. 😬

WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
```diff
@@ -908,6 +908,7 @@ class AbstractPostListViewController: UIViewController,
 
             apost.date_created_gmt = Date()
             apost.status = .publish
+            apost.shouldAttemptAutoUpload = true
             self.uploadPost(apost)
             self.updateFilterWithPostStatus(.publish)
```
shiki: As discussed with @leandroalonso, we will remove these auto-switching of tabs in a separate PR (#12584).

shiki:
I talked to Yael about this and we concluded that her concerns were not caused by this PR. We created a separate issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/12588.

:shipit:

PR #12576, opened 2019-09-30
Improve media retry

### LINKED ISSUE #11176: Media "Upload failed" message stuck for a post in the Blog Posts locally even after the image uploaded normally on a retry
While testing 11.9 alpha with my iPad I managed to get one post into a funny state where it's labeled with "Upload failed" locally on the post card in the Blog Posts list even though tapping retry fixed the image upload and the post seemed to get publish normally after that. The post itself is tagged locally with a red "Upload failed" message that won't go away even after restarting the app.

I took a few screenshots while creating the post originally: https://cloudup.com/cex0l-3oxwL 

Original failed media error as seen on my iPad:

img_0061
<sup>WP Internal 11.9.0.20190226 on iPad Mini 4 (WiFi) (iPad5,1) iOS 12.1.4</sup>

Stuck "Upload failed" message that appears locally on the Blog Posts list:

img_0080
<sup>WP Internal 11.9.0.20190226 on iPad Mini 4 (WiFi) (iPad5,1) iOS 12.1.4</sup>

Here are the app logs which looked relevant to me for the corresponding timestamp of the failed media upload: 

```
2019-02-28 19:23:00:926 Error occurred importing to Media with asset export error, code: 4, error: Error Domain=WordPress.MediaAssetExporter.AssetExportError Code=4 "The image could not be added to the Media Library." UserInfo={NSLocalizedDescription=The image could not be added to the Media Library.}
2019-02-28 19:23:18:378 🔵 Tracked: media_service_upload_failed, properties: {
    "error_code" = 0;
    "error_description" = "Error Domain=MediaServiceErrorDomain Code=0 \"Media doesn't have an associated file to upload.\" UserInfo={NSLocalizedDescription=Media doesn't have an associated file to upload.}";
    "error_domain" = MediaServiceErrorDomain;
}
```

<sup>(note that I uploaded two images successfully but deleted those and added a third and the third one is the only one that had an upload error)</sup>

More app logs: https://mc.a8c.com/pb/2084e/raw/

Checked it with my iPhone and it looks fine there:

img_2070
<sup>WP Internal 11.9.0.20190226 on iPhone 6S iOS 12.1.4</sup>

Site details

* `alleged-swift.jurassic.ninja`
* Jetpack 7.0.1
* WordPRess 5.2-alpha-44784
* Gutenberg toggled "on" in the App Settings 

Tested with WP Internal 11.9.0.20190226 on iPad Mini 4 (WiFi) (iPad5,1) iOS 12.1.4.

SergioEstevao:
Fixes #11176 and helps minimize #12353

This PR improves the media code when handling media upload fails and retries.

To test:

On the media library of a site try the following:
 - Upload new media
 - Upload new media and then turn off the network connection in the middle of the update
 - Check if failures are reported correctly
 - Turn the connection on again and retry uploads of media

Using the Aztec editor:

 - Create a post with media
 - While media is uploading turn off connection
 - Check if errors are reported on media
 - Check if you can retry uploads after connection is made

 - Create a new  post with media
 - While media is still uploading press Publish
 - Check if upload continues on Post List
 - Disconnect network in the middle
 - See if error reports correctly
 - Reconnect network and retry post publishing.

Update release notes:


WordPress/Classes/Utility/CoreDataHelper.swift
```diff
@@ -156,7 +156,7 @@ extension NSPersistentStoreCoordinator {
     /// - Returns: a NSManagedObjectID if the uri is valid or nil if not.
     ///
     public func safeManagedObjectID(forURIRepresentation uri: URL) -> NSManagedObjectID? {
-        guard let host = uri.host, host == "x-coredata" else {
+        guard let scheme = uri.scheme, scheme == "x-coredata" else {
```
frosty: whoa, did this ever work?

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -440,6 +440,7 @@ class AztecPostViewController: UIViewController, PostEditor {
 
         PostCoordinator.shared.cancelAnyPendingSaveOf(post: post)
         addObservers(toPost: post)
+        registerMediaObserver()
```
frosty: The observer is currently set when the post is set (which happens a few lines above this) – could this result in multiple observers being registered?

frosty: Hmm, in testing this seems fine. I think the didSet post one doesn't get called on initialisation?

SergioEstevao: > doesn't get called on initialisation?

Exactly, here is the relevant documentation: 

```The willSet and didSet observers of superclass properties are called when a property is set in a subclass initializer, after the superclass initializer has been called. They are not called while a class is setting its own properties, before the superclass initializer has been called.

For more information about initializer delegation, see Initializer Delegation for Value Types and Initializer Delegation for Class Types.```

WordPress/Classes/Utility/CoreDataHelper.swift
SergioEstevao: I don't think so! I guess we don't suffer a lot of bad connections or else this should have been evident for a long time.

PR #12585, opened 2019-09-30
[SIWA] Handle credential state changes

ScoutHarris:
Ref #12399 
Related WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/139

This uses the WPAuth changes to:
- Accept the `CredentialState` provided by WPAuth.
- If the user id is `revoked`, log out of the account.

There should be no visible difference from the previous implementation.

To test:

This should be tested on a real device for proper keychain handling.

- Log in / sign up with Apple.
- Background the app.
- Disconnect in your Apple account on the device.
  - Go to iOS Settings.
- Open the app.
- The app should log you out.

Update release notes:


WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -885,17 +885,21 @@ extension WordPressAppDelegate {
...
-                DDLogDebug("checkAppleIDCredentialState: Apple ID state not found: \(error.localizedDescription)")
-            }
+            DDLogDebug("checkAppleIDCredentialState: Apple ID state: \(state.rawValue)")
 
-            if !authorized {
-                DDLogInfo("checkAppleIDCredentialState: Unauthorized Apple ID. User signed out.")
+            switch state {
+            case .revoked:
+                DDLogInfo("checkAppleIDCredentialState: Revoked Apple ID. User signed out.")
                 self?.logOutDefaultWordPressComAccount()
```
frosty: We should also ensure we delete the Apple ID from the keychain on log out. I noticed an issue with my device where it somehow still has the Apple ID in the keychain, even though I logged out and logged back in with a different account. Because of this, each time we do the apple credential state check here, the app sees I have a revoked ID and logs me out again.

ScoutHarris: Hey @frosty . Nice catch! It has been done. See what you think now. Thanks!

PR #12540, opened 2019-09-24
[Dark Mode] Crash for unexpected state

danielebogo:
Fixes #12519 

This PR fixes a crash that happens when the app changes its app state to `.background`. 
Because the app [is notified that the trait is changed](https://twitter.com/smileyborg/status/1175199999367368704), it reloads the content to update the style. That generates an exception with the mutable attributed string used to generate the post content.

## To test:
• Run this branch and go to the reader
• Open a post and check everything works normally
• Switch user interface style (light to dark or vice versa)
• Put the app in background and check it doesn't crash
• Switch user interface style (light to dark or vice versa) and bring back the app
• Put the app in background again and check it doesn't crash

Update release notes:

- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12590, opened 2019-10-01
[Dark Mode] Re-enable screenshots when app is in background

### LINKED ISSUE #12554: Crash in Reader after putting the app in the background while viewing an individual post
Steps to reproduce:

1. Open an individual post in the Reader.
1. Minimize the app.

**Result:** the app crashes with a notice that says "WordPress Crashed Do you want to share additional information with the developer?" ([25s](https://cloudup.com/cTleV09vuvp))

wordpress-crashed
<sup>Tested with WPiOS 13.0.0.0 (TestFlight) on iPhone 6S iOS 13.1 beta 4.</sup>

danielebogo:
Fixes #12554 

This PR fix the problem introduced fixing [#12540](https://github.com/wordpress-mobile/WordPress-iOS/pull/12540) where the app wasn't able to keep the screenshot in background. Using @koke suggestion and the new _UIColor_ API `resolvedColor(with:)` I'm able to generate an attributed string for the dark and light user interface and assign it.

## To test:
• Run this branch on iOS 13 and go to the Reader
• Open a post and check everything works normally
• Switch user interface style (light to dark or vice versa)
• Put the app in background and check it doesn't crash
• On the App switch screen, change user interface style (light to dark or vice versa) and see if the app screenshot change
• Test on iOS 12 to check everything works fine

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
@@ -769,6 +769,7 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
         guard let post = post else {
             return
         }
+
         textView.isPrivate = post.isPrivate()
         textView.content = post.contentForDisplay()
 
@@ -780,6 +781,7 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
         guard let post = post else {
             return
         }
+
         if #available(iOS 13, *) {
             let isDark = traitCollection.userInterfaceStyle == .dark
             textView.attributedText = isDark ? darkTextViewAttributedString : lightTextViewAttributedString
--- a/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
@@ -102,6 +102,9 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
 
     private let bottomMarginAttachment = ReaderPlaceholderAttachment()
 
+    private var lightTextViewAttributedString: NSAttributedString?
+    private var darkTextViewAttributedString: NSAttributedString?
+
     @objc var currentPreferredStatusBarStyle = UIStatusBarStyle.lightContent {
         didSet {
             setNeedsStatusBarAppearanceUpdate()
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
```diff
@@ -765,22 +769,45 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
         guard let post = post else {
             return
         }
+
         textView.isPrivate = post.isPrivate()
         textView.content = post.contentForDisplay()
```
koke: Unrelated to this PR, but I noticed it the other day, and this ☝️ is doing the same thing that we're already doing afterwards and generating the attributed string twice.

WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
```diff
@@ -765,22 +769,45 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
...
+        guard let post = post else {
+            return
+        }
+
+        if #available(iOS 13, *) {
+            let isDark = traitCollection.userInterfaceStyle == .dark
+            textView.attributedText = isDark ? darkTextViewAttributedString : lightTextViewAttributedString
+        } else {
+            let attrStr = WPRichContentView.formattedAttributedStringForString(post.contentForDisplay())
+            textView.attributedText = attributedString(with: attrStr)
```
koke: It makes me a bit uncomfortable that we're pre-creating the attributed string on iOS 13 but generating it on demand on earlier versions. It would be interesting to pre-create it in both scenarios, with just the number of variants changing, but I can't think of a specific suggestion that doesn't seem to overcomplicate things ¯\\\_(ツ)\_/¯

WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
danielebogo: So `textView.content = post.contentForDisplay()` is not necessary because the attributed text is set after it right?

koke: Yes, the content setter actually does:

```swift
attributedText = WPRichContentView.formattedAttributedStringForString(newValue)
```

danielebogo:
Thanks @frosty @koke

PR #12594, opened 2019-10-01
Automatically upload published and draft posts with existing remote

leandroalonso:
Fixes a regression in #12324 

For tests instructions, please take a look in #12466 

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/PostService+MarkAsFailedAndDraftIfNeeded.swift
+++ b/WordPress/Classes/Services/PostService+MarkAsFailedAndDraftIfNeeded.swift
@@ -12,7 +12,7 @@
         guard post.remoteStatus != .failed else {
             return
         }
-        
+
         post.remoteStatus = .failed
 
         // If the post was not created on the server yet we convert the post to a local draft
--- a/WordPress/Classes/Services/PostService+MarkAsFailedAndDraftIfNeeded.swift
+++ b/WordPress/Classes/Services/PostService+MarkAsFailedAndDraftIfNeeded.swift
@@ -9,10 +9,6 @@
     ///     eventually be made private.
     ///
     func markAsFailedAndDraftIfNeeded(post: AbstractPost) {
-        guard post.remoteStatus != .failed, !post.hasRemote() else {
-            return
-        }
-
         post.remoteStatus = .failed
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/Services/PostService+MarkAsFailedAndDraftIfNeeded.swift
```diff
@@ -9,10 +9,6 @@
     ///     eventually be made private.
     ///
     func markAsFailedAndDraftIfNeeded(post: AbstractPost) {
-        guard post.remoteStatus != .failed, !post.hasRemote() else {
```
shiki: I think we can keep the first part

```suggestion
    guard post.remoteStatus != .failed else {
```

WordPress/Classes/Services/PostService.m
```diff
@@ -258,7 +258,7 @@ - (void)uploadPost:(AbstractPost *)post
         [self.managedObjectContext performBlock:^{
             Post *postInContext = (Post *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
             if (postInContext) {
-                [self markAsFailedAndDraftIfNeededWithPost:postInContext];
+                [self markAsFailedAndDraftIfNeededWithPost:post];
```
shiki: ⚠️ What is the reason for this change? I thought it should be `postInContext` since `post` may be coming from a different thread.

leandroalonso: Oh sorry, I shouldn't have committed this one.

I did that when I was looking for the cause, reverted.

WordPress/Classes/Services/PostService+MarkAsFailedAndDraftIfNeeded.swift
leandroalonso: Added back.

leandroalonso:
@shiki This wasn't a logic directly related to #12466 but in the end it affected it.

Adding a test for this specific regression in there would imply an integration test, with multiple parts not being mocked. I would try to avoid that.

I added a new test case to ensure this will not happen anymore. I'm pretty comfortable with it by now, what do you think?

shiki:
@leandroalonso Please merge it if you agree with my changes and the build passes. 🙂

PR #12607, opened 2019-10-02
Use fastlane match to manage certificates and profiles for AppStore and Enterprise builds

jtreanor:
Currently, we create and manage our certificates and provisioning profiles on an ad-hoc basis. This means that certs and profiles need to be created and renewed manually, but more importantly, they need to be downloaded and installed manually. This means there is no way for CI to easily install and use our certs/profiles.

This PR updates our internal (enterprise) and AppStore builds to use [Fastlane Match](https://docs.fastlane.tools/actions/match/) to create and manage their certs/profiles. They are [stored encrypted in Google Cloud](https://docs.fastlane.tools/actions/match/#google-cloud-storage).

This means there is no more need to renew profiles and certs manually and that they can easily be used on CI (I have a CircleCI build using it [here](https://circleci.com/gh/wordpress-mobile/WordPress-iOS/13403)).

We could extend this to also work for development certs/profiles but I think its a good idea to see how this goes first.

To test:

1. Run `bundle exec fastlane build_and_upload_release` from the `Scripts` directory.
2. It will check that certs/profiles are up to date and install them if needed.
3. See that the internal build is successfully uploaded to Hockey.
4. The AppStore build will fail at the final step because the current version already exists.
5. Delete the new build in HockeyApp.

Update release notes:


Scripts/fastlane/Fastfile
```diff
@@ -282,6 +267,35 @@ import "./ScreenshotFastfile"
     sh("cd .. && rm WordPress.app.dSYM.zip")
   end
 
+  ########################################################################
+  # Fastlane match code signing
+  ########################################################################  
+  private_lane :internal_code_signing do |options|
+    match(
+      type: "enterprise",
+      team_id: ENV["INT_EXPORT_TEAM_ID"],
```
jkmassel: I wonder – is there a way to read this out of the project configuration (or even hardcode it?) 

I had set up the new environment variable on my machine, which wasn't a huge deal, but wondering if we can shave a minute or two off of the setup this way.

jtreanor: I had the same thought, but just went with what was already here. @loremattei what do you think? Is there any reason not to hardcode the team id here?

jtreanor:
> One thing I noted – I was asked for my computer password by codesign a lot of times during the build. Given that this didn't happen before, I'm wondering if it's related to this change?

This happens any time you add a new certificate and then access it with xcodebuild. It is not match specific. Providing your password and choosing "Always Allow" should stop you from being asked again.

It is possible to prevent these prompts but only by providing your keychain password to match, which is roughly the same thing anyway. Its not an issue on CI because CircleCI unlocks the keychain for us.

Issue #11645, opened 2019-05-06
Webviews: tapping `tel:` or `mailto:` protocol links in a webview results in an "unsupported URL" error


designsimply:
Steps to reproduce:

1. Create some links in a post or sidebar widget with the `tel:` and `mailto:` link protocols.<br><sup>You can use the sample content below.</sup>
1. Save and publish the post.
1. Tap "View" in the snack bar that shows after the post is published to open a Preview.<br>Or you can tap "View" on the post in the Posts List.
1. Note that the links work in a Preview.
1. Go to Reader > Followed Sites and find the post.
1. Tap on the post to open the Reader view and try the links.
1. Tap the More menu at top right and select "Visit" to open a Webview and try the links.
1. Go to My Sites > View Site, find the post, and try the links.

<strong>Result:</strong> tapping on `tel:` or `mailto:` links from a Webview results in an "unsupported URL" error, but those same links work normally in Previews and from a mobile browser such as Safari. ([2m17s](https://cloudup.com/c6gz8rho0mE))

<strong>Sample content:</strong> 

```
<a href="tel:(415) 555 1212">Call (415) 555 1212</a>

<a href="mailto:someone@example.com">Send Email</a> 
```

<img src="https://user-images.githubusercontent.com/1119271/57255683-d9107480-7011-11e9-9fa8-757486003e50.png" width="270" alt="protocol-links-sample-content"> &nbsp; <img src="https://user-images.githubusercontent.com/1119271/57255682-d9107480-7011-11e9-820b-f5e12335bb86.png" width="270" alt="error-unsupported-url">
<p style="clear:both;"><sup>Tested with WP Internal 12.3.0.20190422 on iPhone 6S iOS 12.2 on `design5279.wordpress.com`.</sup></p>

<sub>(internal reference: p4a5px-2of-p2 and 1999324-zen)</sub>

designsimply:
Noting some possibly useful Apple and StackOverflow links:

* [Apple URL Scheme Reference](https://developer.apple.com/library/archive/featuredarticles/iPhoneURLScheme_Reference/Introduction/Introduction.html#//apple_ref/doc/uid/TP40007899-CH1-SW1)
* [How to make phone call in iOS 10 using Swift?](https://stackoverflow.com/a/40079079/1100719)

designsimply:
Added to the `Groundskeeping` project because we've been working recently to improve `Webviews` interactions.

brookewp:
[...]

PR #12612, opened 2019-10-02
Handle link tap inside WPRichContentView instead of in shouldInteractWith URL

### LINKED ISSUE #12568: Reader: Swiping with finger on a link opens the link instead of scrolling
### Expected behavior

When swiping to scroll through a post in the Reader, it shouldn't matter where you place your finger to perform the swipe action.

### Actual behavior

If you happen to place your finger on a link when you perform the swipe action, it opens the link instead of scrolling the post.

This is especially troublesome because you can't immediately close the webview. While the webview says "Loading..." the "x" in the top left doesn't do anything.

1. Open a post in the Reader.
2. Try to swipe when placing your finger over a link. The posts from this blog are great to test with because the content is very link-heavy: https://5blogs.wordpress.com

Result: No matter how quickly you perform the swipe action, the link opens.

yaelirub:
Fixes #12568 

Since iOS 13, the delegate method ```textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction)```
gets called not only when tapping on a link but also when scrolling.

As a result of this and the way we handle things inside that delegate method, links would open on scroll which feels especially buggy in link heavy posts like the ones in this blog: https://5blogs.wordpress.com

This PR adds a gesture recognizer to `WPRichContentView` where we now handle taps on links and attachments.

It also introduces a new WPRichContentViewDelegate method `interactWith(URL: URL)` to preform customization of the link presentation in the controller if needed. 

[Reported issues on SO](https://stackoverflow.com/questions/58189447/ios-13-uitextview-delegate-method-shouldinteract-called-when-scrolling-on-attach)

To test:

In reader post:
1. Open a post in reader (preferably with a lot of links)
2. Make sure that scrolling through doesn't open links
3. Make sure that tapping on a link opens the links as expected
4. Make sure that tapping on media is handled as expected

In reader post comments 
1. Open a post in reader with comments that have links 
2. Make sure that scrolling through the comment content doesn't open links
3. Make sure that tapping on a link in comment opens the links as expected
4. Make sure that tapping on media in comment is handled as expected

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
+++ b/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
@@ -14,7 +14,7 @@ import WordPressShared
 /// A subclass of UITextView for displaying HTML formatted strings.  Embedded content
 /// in tags like img, iframe, and video, are loaded manually and presented as subviews.
 ///
-class WPRichContentView: UITextView, UIGestureRecognizerDelegate {
+class WPRichContentView: UITextView {
 
     /// Used to keep references to image attachments.
     ///
@@ -33,6 +33,10 @@ class WPRichContentView: UITextView, UIGestureRecognizerDelegate {
         let side = max(bounds.size.width, bounds.size.height)
         return CGSize(width: side, height: side)
     }()
+    
+    private func setupTouchDetection() {
+        addGestureRecognizer(linkTapGestureRecognizer)
+    }
 
     @objc lazy var linkTapGestureRecognizer: UITapGestureRecognizer = { [unowned self] in
               let gestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(tapRecognized))
@@ -42,7 +46,6 @@ class WPRichContentView: UITextView, UIGestureRecognizerDelegate {
               gestureRecognizer.delegate = self
               return gestureRecognizer
           }()
-
 
     override var textContainerInset: UIEdgeInsets {
         didSet {
@@ -165,13 +168,6 @@ class WPRichContentView: UITextView, UIGestureRecognizerDelegate {
             }
         }
     }
-
-    private func setupTouchDetection() {
-          for gesture in gestureRecognizers ?? [] {
-              gesture.require(toFail: linkTapGestureRecognizer)
-          }
-          addGestureRecognizer(linkTapGestureRecognizer)
-      }
 
       @objc func tapRecognized(_ recognizer: UIGestureRecognizer) {
           let point = recognizer.location(in: self)
@@ -582,5 +578,22 @@ fileprivate extension String {
         let paragraphSeparators = [String(.carriageReturn), String(.lineFeed), String(.paragraphSeparator)]
 
         return paragraphSeparators.contains(endingString)
+    }
+}
+
+extension WPRichContentView: UIGestureRecognizerDelegate {
+    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
+        guard gestureRecognizer == linkTapGestureRecognizer else {
+            return true
+        }
+        let point = touch.location(in: self)
+        let characterIndex = self.layoutManager.characterIndex(for: point, in: self.textContainer, fractionOfDistanceBetweenInsertionPoints: nil)
+        // handle tap on link
+        if let linkAttribute = self.attributedText?.attribute(.link, at: characterIndex, effectiveRange: nil) {
+            return linkAttribute is URL
+        } else if let attachmentAttribute = self.attributedText?.attribute(.attachment, at: characterIndex, effectiveRange: nil) {
+            return attachmentAttribute is WPTextAttachment
+        }
+        return false
     }
 }
--- a/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
+++ b/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
@@ -177,6 +177,10 @@ class WPRichContentView: UITextView {
               let url = linkAttribute as? URL,
               let richDelegate = delegate as? WPRichContentViewDelegate {
               richDelegate.interactWith?(URL: url)
+          // handle tap on attachement
+          } else if let attachmentAttribute = self.attributedText?.attribute(.attachment, at: characterIndex, effectiveRange: nil),
+              let attachment = attachmentAttribute as? WPTextAttachment {
+              handleImageTapped(imageForAttachment(attachment))
           }
       }
 
--- a/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
+++ b/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
@@ -177,11 +177,7 @@ class WPRichContentView: UITextView {
               let url = linkAttribute as? URL,
               let richDelegate = delegate as? WPRichContentViewDelegate {
               richDelegate.interactWith?(URL: url)
-          // handle tap on attachement
-          } else if let attachmentAttribute = self.attributedText?.attribute(.attachment, at: characterIndex, effectiveRange: nil),
-              let attachment = attachmentAttribute as? WPTextAttachment {
-              handleImageTapped(imageForAttachment(attachment))
-          }
+        }
       }
 
     private func ensureLayoutForAttachment(_ attachment: NSTextAttachment, at range: NSRange) {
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
```diff
@@ -155,6 +166,27 @@ class WPRichContentView: UITextView {
...
+          let point = recognizer.location(in: self)
+          let characterIndex = self.layoutManager.characterIndex(for: point, in: self.textContainer, fractionOfDistanceBetweenInsertionPoints: nil)
+          // handle tap on link
+          if let linkAttribute = self.attributedText?.attribute(.link, at: characterIndex, effectiveRange: nil),
+              let url = linkAttribute as? URL,
+              let richDelegate = delegate as? WPRichContentViewDelegate {
+              richDelegate.interactWith?(URL: url)
+          // handle tap on attachement
+          } else if let attachmentAttribute = self.attributedText?.attribute(.attachment, at: characterIndex, effectiveRange: nil),
+              let attachment = attachmentAttribute as? WPTextAttachment {
```
koke: This doesn't seem to be getting called. When you tap on an image in a reader post, I'm seeing a `ReaderPlaceholderAttachment` instead of `WPTextAttachment`. Opening the image is working because the attachment manager is creating a `WPRichTextImage` which is a `UIControl` and acts as a button.

Unless I missed other details, we can get rid of this case.

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
```diff
@@ -155,6 +166,27 @@ class WPRichContentView: UITextView {
...
     }
 
+    private func setupTouchDetection() {
+          for gesture in gestureRecognizers ?? [] {
+              gesture.require(toFail: linkTapGestureRecognizer)
+          }
+          addGestureRecognizer(linkTapGestureRecognizer)
+      }
+
+      @objc func tapRecognized(_ recognizer: UIGestureRecognizer) {
```
koke: I haven't found an example where this goes wrong, but just in case, we should be implementing `gestureRecognizer(_:shouldReceive:)` in our delegate to make sure our recognizer only responds to taps on links.

I've been playing with it and this seems to work:

```swift
    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
        guard gestureRecognizer == linkTapGestureRecognizer else {
            return true
        }
        let point = touch.location(in: self)
        let characterIndex = self.layoutManager.characterIndex(for: point, in: self.textContainer, fractionOfDistanceBetweenInsertionPoints: nil)

        let linkAttribute = self.attributedText?.attribute(.link, at: characterIndex, effectiveRange: nil)
        return linkAttribute is URL
    }
```

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
```diff
@@ -155,6 +166,27 @@ class WPRichContentView: UITextView {
         }
     }
 
+    private func setupTouchDetection() {
+          for gesture in gestureRecognizers ?? [] {
+              gesture.require(toFail: linkTapGestureRecognizer)
```
koke: I thought this would be necessary but I commented it out and things still seem to work (after implementing the `shouldReceive` method). What was this trying to prevent?

Also, it seems to be interfering with the normal behavior of the text view. When I long press on a link, the share sheet pops up now ❤️ but the word I tapped also gets selected.

Simulator Screen Shot - iPhone 11 - 2019-10-03 at 09 37 57

koke:
While testing the tapping on images, I noticed something that seems unrelated to this issue, but worth noting, not sure if you know more about it. When you long press an image, it pops out as if you would be able to drag and drop, but the content is blank, and the moment you try to drag it, it disappears:

reader-image-long-press

I think we should probably support drag and drop and the share sheet on long press for reader images.

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
```diff
@@ -155,6 +166,27 @@ class WPRichContentView: UITextView {
...
+          }
+          addGestureRecognizer(linkTapGestureRecognizer)
+      }
+
+      @objc func tapRecognized(_ recognizer: UIGestureRecognizer) {
+          let point = recognizer.location(in: self)
+          let characterIndex = self.layoutManager.characterIndex(for: point, in: self.textContainer, fractionOfDistanceBetweenInsertionPoints: nil)
+          // handle tap on link
+          if let linkAttribute = self.attributedText?.attribute(.link, at: characterIndex, effectiveRange: nil),
+              let url = linkAttribute as? URL,
```
koke: This was probably an existing issue, but I wanted to test links in comments and somehow the first comment with a link I could find was an email address. `mailto:user@example.com` is a URL, but we don't want our in-app browser to handle that (and it can't)

Simulator Screen Shot - iPhone 11 - 2019-10-03 at 09 48 53

koke: Confirmed it was the same in latest WP Internal

koke:
> I have considered if this change warrants user-facing release notes and have added them to RELEASE-NOTES.txt if necessary.

I think this one should go in the release notes. At least definitely a mention in the call for testing, since I'm not sure what odd cases we might be missing for the gesture recognizer interactions.

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
yaelirub: Weird, it's getting called for me and if I remove this logic, tap on an image doesn't do anything.

koke: I was testing on the iOS 13.1 simulator

yaelirub: I'm on 13.1 device

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
yaelirub: And what about taps on images? I guess that's related to our comments above?

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
yaelirub: You're right! Removing it

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
koke: That code doesn't handle the images because I thought we didn't need to, but it's possible as well, just a bit more complicated. The logic is similar to the `tapRecognized`, except it only has to return if the touch should be handled.

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
yaelirub: Right. Out of scope for this PR. I'll create a ticket for it if one doesn't exist

yaelirub:
@koke , updated and ready for another look

yaelirub:
@koke , you're right!
Im removing the attachment from both methods

yaelirub:
The failing testDoNotUploadAPostWithFailedMedia was fixed for 13.4 https://github.com/wordpress-mobile/WordPress-iOS/pull/12514

jtreanor:
@yaelirub did you mean to target 13.3 with this? I submitted the build for review this morning. Could this wait for 13.4 or should we consider delaying the release?

yaelirub:
@jtreanor , I meant to target 13.3 and it has to be merged there since this bug is an annoying issue and have been reported by multiple users https://github.com/wordpress-mobile/WordPress-iOS/issues/12568#issuecomment-538017300 including Matt.

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
yaelirub: A ticket exists: https://github.com/wordpress-mobile/WordPress-iOS/issues/11645

aerych:
FWIW, possibly, maybe resolved in iOS 13.2.  More info linked [in this comment](https://github.com/Automattic/simplenote-ios/pull/470#issuecomment-542382370).

I haven't looked through the code here but FYI in case it would make sense to roll back with the updated OS (assuming its actually fixed)

PR #12609, opened 2019-10-02
Use credentials tool to copy/decrypt credentials intead of referencing the secrets repo path

jtreanor:
Currently, our credentials generation references the `~/.mobile-secrets` path directly. This is a problem for cases where the secrets repo doesn't exist (such as CI).

This is an update to use our existing credentials tool to copy/decrypt the relevant files within the repo. This will allow CI to decrypt these files if needed. It also simplifies the setup for third party contributors (no need to create a `~/.mobile-secrets` directory).

To test:

1. Run `rake dependencies`.
2. Confirm that `wpcom_alpha_app_credentials`, `wpcom_app_credentials` and `wpcom_internal_app_credentials` are present in the `WordPress/Credentials` directory.
3. Open the workspace in Xcode and build & run the app. Confirm WP.com login still works.

Update release notes:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12623, opened 2019-10-04
Revert "Revert "Use credentials tool to copy/decrypt credentials intead of referencing the secrets repo path""

jtreanor:
https://github.com/wordpress-mobile/WordPress-iOS/pull/12609 was reviewed and merged yesterday. Shortly after being merged @ScoutHarris encountered a problem where the credentials task wasn't being triggered, I reverted the change in wordpress-mobile/WordPress-iOS#12622.

This is a small addition to the original change to fix the issue @ScoutHarris was seeing. The problem was that the task checked for the presence of `~/.mobile-secrets/keys.json` which was only added relatively recently. It now checks for `~/.mobile-secrets/.git` which will be present for anyone who has the repo.

To test:

- Run `rake dependencies` and see that the `configure_apply` step is run.


Scripts/fastlane/Pluginfile
```diff
@@ -6,5 +6,5 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 3.2.0'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.7.1'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.7.2'
```
jtreanor: 0.7.2 had a bunch of improvements to the apply behaviour: https://github.com/wordpress-mobile/release-toolkit/releases/tag/0.7.2

ScoutHarris:
FWIW, I gave it a whirl. I did see `configure_apply` run and was able to log in.

ScoutHarris:
Hey @jtreanor . I just noticed these files were created locally. Is this expected? 

new_files

jtreanor:
@ScoutHarris yes, that is by design. They are also added to `.gitignore`

ScoutHarris:
Hey @jtreanor .

> yes, that is by design. 

Um, maybe a dumb question - but I'm not sure how to "clear" these files. They keep getting recreated and there is a high probability I'll commit them accidentally. Halp! 😄

jtreanor:
@ScoutHarris There is no need to clear the files. The reason you are seeing them is that you checked out this branch, ran `rake dependencies` and then switched to a branch where they were not ignored. This won't be a problem once this is merged (which I am doing now).

PR #12631, opened 2019-10-04
Add tracking for auto upload invoked

### LINKED ISSUE #12359: [Offline Posting] Add tracking
📣 Part of #12240. Any fixes should target the `issue/12240-master-branch` branch. 

We have to follow what events were added on Android. Please see https://github.com/wordpress-mobile/WordPress-Android/pull/10423.

yaelirub:
Fixes #12359 
Related PR on WPShared - https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/220

To test:

Update release notes:


WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -337,6 +337,9 @@ extension PostCoordinator: Uploader {
             }
 
             postsAndActions.forEach { post, action in
+                if action != .nothing, let status = post.status {
+                    self.trackAutoUpload(action: action, status: status)
+                }
```
leandroalonso: What do you think about moving these checks to the `trackAutoUpload` method?

It's a logic related to the track and this will make this call cleaner, without worrying about the `action` or if we have a `post.status`.

yaelirub: Hey Leandro, I'll follow your suggestion cause i like that approach as well though I thought it would be better than having an optional parameter

yaelirub:
@leandroalonso , ready for another look

Issue #12246, opened 2019-07-31
[Insights Management] Add 'Add New Stats Card' functionality


ScoutHarris:
Add `Add New Stats Card` functionality.

- [x] Add the `Add New Stats Card` view. (#12281)
- [x] Lists all Insights by category. (#12281)
- [x] Stats already displayed are greyed out and disabled. (#12633)
- [x] Accessed from: (#12281)
  - [x] The Customize card > Try it now.
  - [x] The add new card.
- [x] When selected, add stat to bottom of Insights (but before the add new card). (#12632)

From Zeplin:

![Insights list](https://user-images.githubusercontent.com/1816888/62239286-d663b800-b391-11e9-911c-6b7fbd3d54c5.png)

PR #12633, opened 2019-10-05
[Insights Management] Disable already displayed Insights

ScoutHarris:
Ref #12246 

To test:
- Go to Stats > Insights.
- Select 'Add new stats card'.
- Verify the stats already displayed are greyed out and disabled (i.e. selecting them does nothing).
- Select an available stat.
- Verify it appears in the Insights table.
- Return to 'Add new stats card'.
- Verify the newly selected stat is now greyed out.

selected_stats

Update release notes:


frosty (approved):
👍

PR #12650, opened 2019-10-08
[Insights Management] Add ability to organize cards

### LINKED ISSUE #12247: [Insights Management] Add ability to organize cards
On each stat card, show an ellipsis button that will present options to manage the card.

- NOTE: `StatsCellHeader` has a hidden `manageInsightButton` for this purpose.
- When the button is selected, these options are displayed in an action sheet:
  - Move up
    - Moves card up by one slot.
    - Hidden if the card is at the top.
  - Move down
    - Moves card down by one slot.
    - Hidden if the card is at the bottom.
  - Remove from Insights

From Zeplin:

options

ScoutHarris:
Fixes #12247 

This adds the ability to:
- Move a stat card up.
- Move a stat card down.
- Remove stat card from Insights.

To test:
- Go to Stats > Insights.
- Verify the ellipsis button appears in all headers.

manage_insight_button

- Tap the button.
- Select 'Move up'.
  - Verify the card moves up one slot.
  - If the card is at the top (or just below the Customize card), 'Move up' is not an option.
- Select 'Move down'.
  - Verify the card moves down one slot.
  - If the card is at the bottom, 'Move down' is not an option.
- Select 'Remove from insights'.
  - Verify the stat card is removed.
   - Select 'Add new stats card'.
  - Verify the newly removed stat is available to add.
- Please test on iPhone and iPad.

---
NOTE: A constraint warning appears when the action sheet is displayed. I tried other actions sheets in the app, and they do the same. It appears to be an issue with the action sheet itself.

---
Update release notes:


WordPress/Classes/ViewRelated/Stats/SiteStatsTableViewCells.swift
```diff
@@ -420,7 +462,7 @@ struct PostStatsEmptyCellHeaderRow: ImmuTableRow {
             return
         }
 
-        cell.configure(withTitle: "", adjustHeightForPostStats: true)
+        cell.configure(statSection: .postStatsGraph)
```
frosty: How come the empty cell header row is using the post stats graph section? (I may just be misunderstanding the code here)

frosty (approved):
Just one small query, otherwise this works excellently – nice job! 👍

WordPress/Classes/ViewRelated/Stats/SiteStatsTableViewCells.swift
ScoutHarris: On the `Post Stats` view, the `StatsCellHeader` above the graph has a specific height. While I was refactoring the `configure` method, I realized I didn't explicitly need the `adjustHeightForPostStats` param, all I needed to do was tell it the section it was for. By sending `postStatsGraph`, `StatsCellHeader` sizes itself correctly because:
1. `StatsCellHeader` knows it's for the graph header.
2. `postStatsGraph` doesn't have a title.

I can add a comment in my next PR to clarify that.

PR #12652, opened 2019-10-08
Make PostCoordinator.autoSave create non-draft posts if necessary and handle failures

### LINKED ISSUE #12630: [Offline Posting] Fix PostService.autoSave() or create a new one
⚒**Fix `PostService.autoSave()` or create a new one**

In `PostService.autoSave()`, upload the post as a draft by modifying `RemotePost` so we don't touch the `Post` `NSManagedObject` ever. We should probably make our own `autoSave` method because the existing one in `PostService` does not look useable for our goal. It looks like it's specific for remote previews.

Also, the current `autoSave` doesn't mark a post as failed when the saving fails. Which leads to this bug:

1. While offline, create a post and publish
2. Cancel the auto-upload
3. Restart the app
4. Check that the progress bar is appearing and that the label "Uploading post..." is visible

shiki:
Fixes #12630 

## Findings

Whenever a **new published post** is auto-saved, it gets stuck at “Uploading post...”. The cause was `PostCoordinator.autoSave` thinks that the post should be auto-saved instead of created in the server. 

https://github.com/wordpress-mobile/WordPress-iOS/blob/db3444be48d52294f83fc0024badeb0b19a7a114/WordPress/Classes/Services/PostService.m#L340

And because the post is _new_, the call to `[restRemote autoSave]` will fail and the `failureBlock` is called:

https://github.com/wordpress-mobile/WordPress-iOS/blob/db3444be48d52294f83fc0024badeb0b19a7a114/WordPress/Classes/Services/PostService.m#L336-L338

But the `failureBlock` never resets the `Post.remoteStatus` to `.failed`, so the post gets stuck at `.pushing`. Hence, the cause of the infinite “Uploading post...” message. 

## Solution

I opted not to create a variation of `PostService.autoSave()` and refactor it instead. This will benefit (hopefully positively) previewing too so that users who set their new posts to `.published` can still preview them. 

There are 2 main changes:

- Local posts, regardless of the status, will be automatically created
- If an auto-save fails, the post's `remoteStatus` will be changed to `.failed`. 

### Creating

1. While offline, create a post and publish it. 
2. In the Post List, cancel the auto-upload. 
3. Go online. The post should be automatically uploaded as a draft. 

### Editing 

1. While offline, edit an existing published post. 
2. Exit the editor → Update Post
3. In the Post List, cancel the auto-upload. 
4. Go online. The post should be auto-saved. Please verify that when editing the post in Calypso, you will be able to a message that there is an available auto-save of the post.

### Failures

1. While offline, edit an existing published post. 
2. Exit the editor → Update Post
3. In the Post List, cancel the auto-upload. 
4. Go online. While the _uploading UI_ is happening, immediately go back offline. The _uploading UI_ should disappear and the post should say “Local changes”. 

## Reviewing 

## Submitter Checklist

- [x] I have added unit tests.

### DIFF
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -329,17 +329,29 @@ forceDraftIfCreating:(BOOL)forceDraftIfCreating
          success:(void (^)(AbstractPost *post, NSString *previewURL))success
          failure:(void (^)(NSError * _Nullable error))failure
 {
+    NSManagedObjectID *postObjectID = post.objectID;
+
+    NSError *defaultError =
+        [NSError errorWithDomain:PostServiceErrorDomain
+                            code:0
+                        userInfo:@{ NSLocalizedDescriptionKey : @"Previews are unavailable for this kind of post." }];
+
+    void (^setPostAsFailedAndCallFailureBlock)(NSError *error) = ^(NSError *error) {
+        [self.managedObjectContext performBlock:^{
+            AbstractPost *postInContext = (AbstractPost *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
+            if (postInContext) {
+                postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
+            }
+
+            failure(error);
+        }];
+    };
+
     id<PostServiceRemote> remote = [self.postServiceRemoteFactory forBlog:post.blog];
-
-    NSError *defaultError = [NSError errorWithDomain:PostServiceErrorDomain
-                                                code:0
-                                            userInfo:@{ NSLocalizedDescriptionKey : @"Previews are unavailable for this kind of post." }];
-
     if ([remote isKindOfClass:[PostServiceRemoteREST class]]) {
         PostServiceRemoteREST *restRemote = (PostServiceRemoteREST*) remote;
         RemotePost *remotePost = [self remotePostWithPost:post];
-        NSManagedObjectID *postObjectID = post.objectID;
-        
+
         void (^successBlock)(RemotePost *post, NSString *previewURL) = ^(RemotePost *post, NSString *previewURL) {
             [self.managedObjectContext performBlock:^{
                 AbstractPost *postInContext = (AbstractPost *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
@@ -365,17 +377,6 @@ forceDraftIfCreating:(BOOL)forceDraftIfCreating
             }];
         };
         
-        void (^failureBlock)(NSError *error) = ^(NSError *error) {
-            [self.managedObjectContext performBlock:^{
-                AbstractPost *postInContext = (AbstractPost *)[self.managedObjectContext existingObjectWithID:postObjectID error:nil];
-                if (postInContext) {
-                    postInContext.remoteStatus = AbstractPostRemoteStatusFailed;
-                }
-
-                failure(error);
-            }];
-        };
-
         // The autoSave endpoint returns an exception on posts that do not exist on the server
         // so we'll create the post instead if necessary.
         BOOL mustBeCreated = ![post hasRemote];
@@ -384,7 +385,7 @@ forceDraftIfCreating:(BOOL)forceDraftIfCreating
             // Abort if the status is trashed/deleted. We'd rather not automatically create a
             // locally trashed post as drafts in the server.
             if ([post.status isEqualToString:PostStatusTrash] || [post.status isEqualToString:PostStatusDeleted]) {
-                failureBlock(defaultError);
+                setPostAsFailedAndCallFailureBlock(defaultError);
                 return;
             }
 
@@ -397,7 +398,7 @@ forceDraftIfCreating:(BOOL)forceDraftIfCreating
         } else {
             [restRemote autoSave:remotePost
                          success:successBlock
-                         failure:failureBlock];
+                         failure:setPostAsFailedAndCallFailureBlock];
         }
     } else if ([post originalIsDraft] && [post isDraft]) {
         [self uploadPost:post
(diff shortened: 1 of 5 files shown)


WordPress/Classes/Services/PostService.m
```diff
@@ -234,6 +254,20 @@ - (void)syncPostsOfType:(PostServiceType)postType
...
+      creationOption:CreationOptionAsIs
+             success:success
+             failure:failure];
+}
+
+/**
+ * @param creationOption The strategy to use if [remote createPost:] is invoked.
+ */
+- (void)uploadPost:(AbstractPost *)post
+    creationOption:(CreationOption)creationOption
```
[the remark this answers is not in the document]
shiki: I started with `forceDraft` but thought it would be confusing because `uploadPost` does both **updating** and **creating**. A `true` `forceDraft` could mean that even if we are just updating the post, we would change the post to a draft. Furthermore, a `false` `forceDraft` is also not clear in what it does without reading the code.  

This is why I opted to use an enum instead to clarify what to use _if creating_ a post: (a) change it to a draft, or (b) leave as is. I believe this, contrary to your argument of “making our code more explicit“, is very explicit. 🙂 

But I'd like to move on and I slightly agree that the enum is still confusing and even over-engineered. It's sort of a compromise instead of refactoring a whole lot of this class. So what do you think about `forceDraftIfCreating` instead? I think I'll have to add more comments to describe the effects of the boolean but it is what is. 😬

WordPress/Classes/Services/PostService.m
```diff
@@ -356,8 +415,7 @@ - (void)autoSave:(AbstractPost *)post
                      success(post, nil);
                  } failure:failure];
     } else {
-        NSDictionary *userInfo = @{ NSLocalizedDescriptionKey : @"Previews are unavailable for this kind of post." };
-        failure([NSError errorWithDomain:PostServiceErrorDomain code:0 userInfo:userInfo]);
+        failure(defaultError);
```
leandroalonso: Using an self-hosted site will end up in this call. However, this doesn't mark the post as failed which produces the bug of the "infinite uploading post...".

We need to mark the post as failed for each failure. :)

WordPress/Classes/Services/PostService.m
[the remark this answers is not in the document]
leandroalonso: I like `forceDraftIfCreating`. :)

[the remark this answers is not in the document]
shiki: Done in c3474abb01b9ba92e1009d2a23a9f72fe94e3ec3.

WordPress/Classes/Services/PostService.m
shiki: (╯°□°)╯︵ ┻━┻

Fixed in 928c4aae28fa5d7b230def8d4498cca694c4f4ad.

shiki:
@leandroalonso Ready for another look!

WordPress/Classes/Services/PostService.m
```diff
@@ -234,6 +234,30 @@ - (void)syncPostsOfType:(PostServiceType)postType
 - (void)uploadPost:(AbstractPost *)post
            success:(void (^)(AbstractPost *post))success
            failure:(void (^)(NSError *error))failure
+{
+    [self uploadPost:post
+forceDraftIfCreating:NO
```
leandroalonso: `forceDraftIfCreating` seems unaligned here. Is that intentional?

WordPress/Classes/Services/PostService.m
```diff
@@ -234,6 +234,30 @@ - (void)syncPostsOfType:(PostServiceType)postType
...
+ * Setting `forceDraftIfCreating` to `YES` is useful if we want to create a post in the server
+ * with the intention of making it available for preview. If we create the post as is, and the user
+ * has set its `status` to `.published`, then we would publishing the post even if we just
+ * wanted to preview it!
+ *
+ * Another use case of `forceDraftIfCreating` is to create the post in the background so we can
+ * periodically auto-save it. Again, we'd still want to create it as a `.draft` status.
+ */
+- (void)uploadPost:(AbstractPost *)post
+forceDraftIfCreating:(BOOL)forceDraftIfCreating
```
leandroalonso: Same as above.

WordPress/Classes/Services/PostService.m
```diff
@@ -333,31 +377,36 @@ - (void)autoSave:(AbstractPost *)post
...
+        if (mustBeCreated) {
+            // Abort if the status is trashed/deleted. We'd rather not automatically create a
+            // locally trashed post as drafts in the server.
+            if ([post.status isEqualToString:PostStatusTrash] || [post.status isEqualToString:PostStatusDeleted]) {
+                setPostAsFailedAndCallFailureBlock(defaultError);
+                return;
+            }
+
             [self uploadPost:post
+        forceDraftIfCreating:YES
```
leandroalonso: Same.

WordPress/Classes/Services/PostService.m
shiki: I made Xcode align it. I don't think it can align it anymore than that. 😬 The second parameter is longer than the first.

WordPress/Classes/Services/PostService.m
shiki: I made Xcode align it. I don't think it can align it anymore than that. 😬 The second parameter is longer than the first.

WordPress/Classes/Services/PostService.m
shiki: It's aligned correctly actually. The alignment is based on the colon.

WordPress/Classes/Services/PostService.m
leandroalonso: :shipit:

PR #12653, opened 2019-10-09
Fastlane: Split env files into project.env and user.env so that project.env can be shared

jtreanor:
As discussed, this splits the env file used for Fastlane (currently just `~/.wpios-env.default`) into two files: `.configure-files/project.env` and the existing `~/.wpios-env.default`. The latter is copied from the secrets repo and the former should be filled in with personal credentials. Example files for each can be found in `Scripts/fastlane/env`

I have also added a new method, `get_required_env()` which gets a value from `ENV`, but throws an error if it is missing. This ensures that if we add a key to `~/.wpios-env.default`, Fastlane will not continue if that key is missing. Yesterday, @jkmassel and I noticed that some Sentry keys were missing. It should prevent that from happening in future.

To test:

- Run `rake dependencies`. Confirm that `.configure-files/project.env` now exists.
- Ensuring you have an env file at `~/.wpios-env.default`, run any lane (e.g. `bundle exec get_pullrequests_list`). See that there are no errors.

Update release notes:


Scripts/fastlane/Fastfile
```diff
@@ -1,11 +1,18 @@
...
+  UI.user_error!("env/user.env not found: Please copy env/user.env-example to env/user.env and fill in the values") unless File.file?('env/user.env')
+  UI.user_error!("env/project.env not found: Make sure your configuration is up to date with `bundle exec fastlane run configure_apply`") unless File.file?('env/project.env')
+end
+
 platform :ios do
 ########################################################################
 # Environment
 ########################################################################
-Dotenv.load('~/.wpios-env.default')
+Dotenv.load('env/user.env')
```
jtreanor: Note that the user env file now lives in the repo and not at `~/.wpios-env.default`. If anyone has a preference, I am open to keeping it at the old location. Having it in the repo is more consistent with everything else we do.

loremattei: To be honest, I don't like this solution a lot. I mean, it's not specifically related to _this_ file.
I see we have added other secrets in the repo and I agree that we have to be consistent with the layout, so I wonder if there's a technical reason to handle all these files in the repo or if they could be places outside without adding technical efforts. 
With the files in the repo, there's always the risk that they get committed in an older branch that doesn't .gitignore them.

jtreanor: Thanks for your comment @loremattei! Those are all fair points. Lets chat about this later 👍

jtreanor: We discussed this offline and agreed to move all sensitive files to one directory to mitigate the issue. I have opened a PR for that in https://github.com/wordpress-mobile/WordPress-iOS/pull/12668. Once that is merged, I will update this PR to do the same.

jtreanor:
@loremattei I have now updated this following our discussion and the changes in https://github.com/wordpress-mobile/WordPress-iOS/pull/12668. Now there are two files: `~/.wpios-env.default` (the existing one should work fine) and `.configure-files/project.env` which is copied from the secrets repo.

I have updated the testing steps in the description.

PR #12663, opened 2019-10-09
[Stats] Update ghost cell colors for Dark Mode

ScoutHarris:
Ref #12405 

This updates the colors used in the Stats ghost cells.

To test:
- On a slow network, go to Stats > Insights.
- Verify the ghost colors are correct in dark mode.

ghost_colors

Update release notes:


frosty:
I did just notice that the ghosts are flickering every time a new card loads – I think this was pre-existing this PR, but it's something we should look at addressing.

danielebogo:
> I did just notice that the ghosts are flickering every time a new card loads – I think this was pre-existing this PR, but it's something we should look at addressing.

At the moment we reload the whole tableview and not a specific row/section. But since we have the type of the row/section and its state an inverse mapping to get the right index and reload it would be great!

PR #12690, opened 2019-10-14
Update button works while offline

### LINKED ISSUE #12589: [Offline Posting] “Update” button in editor should work while offline
📣 Part of #12227. Please target the `issue/12240-master-branch` branch. 

This is from https://github.com/wordpress-mobile/WordPress-iOS/pull/12466#issuecomment-529694557. 

## Steps to reproduce

1. Go offline
1. Edit an existing post
2. Make some changes
3. Press _Update_. The Snackbar should show “Error occurred during saving”.
5. Force close the app while the editor is still open. 
6. Open the app again and navigate to the Post List. Notice that the Post List says that the post will be automatically uploaded. 

~This is against the principles because we did not mention to the user (via Snackbar) that the post will be automatically uploaded.~

## ~Reason for Low Priority (Outdated)~

~I'm initially giving this a low priority because it seems like an edge case. Also, eventually, the user will end up pressing **X** → _Update Post_ and **confirming** the auto-upload.~

## Update (2019-Oct-10)

During the Design Review, we decided that the Update button should behave similarly to the _Update Post_ option that is shown in the dialog when exiting the editor. 

- It should work offline
- The same Snackbar message like “The post will be published...” should be shown
- ~A key difference is the button should not close the editor~
- ~The blocking progress dialog should be removed~

leandroalonso:
Fixes #12589

## To test:

1. Create and publish a post
2. While offline, edit the post
3. Tap "Update"
4. Tap X: note that no dialog should *not* appear if you already confirmed the changes
5. A snackbar message should appear
6. Check that the post will be updated when back online

Please, try different combinations:

* Add images
* Edit different type of posts: scheduled, drafts, pending
* Use Aztec and Gutenberg

## Additional info

Note that the original issue had 2 more tasks:

* A key difference is the button should not close the editor
* The blocking progress dialog should be removed

I decided to leave those to another task since it's not directly related to the Offline Support (and these behaviors are currently in `develop`).

Update release notes:

- [x] If it's feasible, I have added unit tests. (`PostEditor` stills hard to test)

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -72,14 +72,24 @@ class PostCoordinator: NSObject {
     }
 
     func save(_ postToSave: AbstractPost, automatedRetry: Bool = false) {
-        prepareToSave(postToSave, automatedRetry: automatedRetry) { post in
-            self.upload(post: post)
+        prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
+            switch result {
+            case .success(let post):
+                self.upload(post: post)
+            case .error(_):
+                break
+            }
         }
     }
 
     func autoSave(_ postToSave: AbstractPost, automatedRetry: Bool = false) {
-        prepareToSave(postToSave, automatedRetry: automatedRetry) { post in
-            self.mainService.autoSave(post, success: { uploadedPost, _ in }, failure: { _ in })
+        prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
+            switch result {
+            case .success(let post):
+                self.mainService.autoSave(post, success: { uploadedPost, _ in }, failure: { _ in })
+            case .error(_):
+                break
+            }
         }
     }
 
@@ -109,7 +119,7 @@ class PostCoordinator: NSObject {
     /// - Parameter automatedRetry: if this is an automated retry, without user intervenction
     /// - Parameter then: a block to perform after post is ready to be saved
     ///
-    private func prepareToSave(_ postToSave: AbstractPost, automatedRetry: Bool = false, then completion: @escaping (AbstractPost) -> ()) {
+    private func prepareToSave(_ postToSave: AbstractPost, automatedRetry: Bool = false, then completion: @escaping (Result<AbstractPost>) -> ()) {
         var post = postToSave
 
         if postToSave.isRevision() && !postToSave.hasRemote(), let originalPost = postToSave.original {
@@ -145,7 +155,7 @@ class PostCoordinator: NSObject {
                         // Let's check if media uploading is still going, if all finished with success then we can upload the post
                         if !self.mediaCoordinator.isUploadingMedia(for: post) && !post.hasFailedMedia {
                             self.removeObserver(for: post)
-                            completion(post)
+                            completion(.success(post))
                         }
                     }
                     switch media.mediaType {
@@ -172,7 +182,7 @@ class PostCoordinator: NSObject {
             return
         }
 
-        completion(post)
+        completion(.success(post))
     }
 
     func cancelAnyPendingSaveOf(post: AbstractPost) {
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -87,7 +87,7 @@ class PostCoordinator: NSObject {
             switch result {
             case .success(let post):
                 self.mainService.autoSave(post, success: { uploadedPost, _ in }, failure: { _ in })
-            case .error(_):
+            case .error:
                 break
             }
         }
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -76,12 +76,19 @@ class PostCoordinator: NSObject {
         return isPushingAllPendingMedia
     }
 
-    func save(_ postToSave: AbstractPost, automatedRetry: Bool = false, completion: ((Result<AbstractPost>) -> ())? = nil) {
+    func save(_ postToSave: AbstractPost,
+              automatedRetry: Bool = false,
+              defaultFailureNotice: Notice? = nil,
+              completion: ((Result<AbstractPost>) -> ())? = nil) {
+
         prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
             switch result {
             case .success(let post):
                 self.upload(post: post, completion: completion)
             case .error(let error):
+                if let notice = defaultFailureNotice {
+                    ActionDispatcher.dispatch(NoticeAction.post(notice))
+                }
                 completion?(.error(error))
             }
         }
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -71,13 +71,13 @@ class PostCoordinator: NSObject {
         return isPushingAllPendingMedia
     }
 
-    func save(_ postToSave: AbstractPost, automatedRetry: Bool = false) {
+    func save(_ postToSave: AbstractPost, automatedRetry: Bool = false, completion: ((Result<AbstractPost>) -> ())? = nil) {
         prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
             switch result {
             case .success(let post):
-                self.upload(post: post)
-            case .error(_):
-                break
+                self.upload(post: post, completion: completion)
+            case .error(let error):
+                completion?(.error(error))
             }
         }
     }
@@ -253,7 +253,7 @@ class PostCoordinator: NSObject {
         backgroundService.refreshPostStatus()
     }
 
-    private func upload(post: AbstractPost) {
+    private func upload(post: AbstractPost, completion: ((Result<AbstractPost>) -> ())? = nil) {
         mainService.uploadPost(post, success: { uploadedPost in
             print("Post Coordinator -> upload succesfull: \(String(describing: uploadedPost.content))")
 
@@ -261,9 +261,13 @@ class PostCoordinator: NSObject {
 
             let model = PostNoticeViewModel(post: uploadedPost)
             ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+
+            completion?(.success(uploadedPost))
         }, failure: { error in
             let model = PostNoticeViewModel(post: post)
             ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+
+            completion?(.error(error ?? SavingError.unknown))
 
             print("Post Coordinator -> upload error: \(String(describing: error))")
         })
@@ -419,5 +423,6 @@ extension PostCoordinator {
 extension PostCoordinator {
     enum SavingError: Error {
         case mediaFailure
+        case unknown
     }
 }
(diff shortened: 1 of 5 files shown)


WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -240,18 +253,22 @@ class PostCoordinator: NSObject {
...
 
             let model = PostNoticeViewModel(post: uploadedPost)
             ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+
+            completion?(.success(uploadedPost))
         }, failure: { error in
             let model = PostNoticeViewModel(post: post)
             ActionDispatcher.dispatch(NoticeAction.post(model.notice))
 
+            completion?(.error(error ?? SavingError.unknown))
```
yaelirub: What are we actually doing with this error? Can we send more info here?

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -336,14 +331,19 @@ extension PostEditor where Self: UIViewController {
...
-                generator.notificationOccurred(.error)
-            } else if let uploadedPost = uploadedPost {
+            switch result {
+            case .success(let uploadedPost):
                 self.post = uploadedPost
 
                 generator.notificationOccurred(.success)
+            case .error(let error):
+                guard !self.post.shouldAttemptAutoUpload else {
+                    break
```
yaelirub: Do you mean to return here instead of break?

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -336,14 +331,19 @@ extension PostEditor where Self: UIViewController {
...
-                ActionDispatcher.dispatch(NoticeAction.post(self.uploadFailureNotice(action: action)))
-                generator.notificationOccurred(.error)
-            } else if let uploadedPost = uploadedPost {
+            switch result {
+            case .success(let uploadedPost):
                 self.post = uploadedPost
 
                 generator.notificationOccurred(.success)
+            case .error(let error):
+                guard !self.post.shouldAttemptAutoUpload else {
```
yaelirub: Why are we doing this check here?

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -71,15 +71,25 @@ class PostCoordinator: NSObject {
...
 
     func autoSave(_ postToSave: AbstractPost, automatedRetry: Bool = false) {
-        prepareToSave(postToSave, automatedRetry: automatedRetry) { post in
-            self.mainService.autoSave(post, success: { uploadedPost, _ in }, failure: { _ in })
+        prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
+            switch result {
+            case .success(let post):
+                self.mainService.autoSave(post, success: { uploadedPost, _ in }, failure: { _ in })
+            case .error:
+                break
```
yaelirub: Should this be "return"? Why are we not doing anything with the error?

yaelirub:
Hey @leandroalonso , great work!
So far in my testing things are working as expected except for seeing the "local changes" label a lot which  I'm not sure is expected. 
Fo example:
1. Upload draft
2. tap edit
3. close the editor without updating
4. See local changes

OR
1. Upload draft / publish post
2. update draft / published
3. See local changes 

It happens with scheduled posts too.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
leandroalonso: If the post is not marked as `shouldAttemptAutoUpload` I wanna show an error to the user.

But I'm not sure this case will ever happen now that the "Update" button marks the post to be auto uploaded. 🤔 

Maybe makes sense to remove this error triggering logic?

WordPress/Classes/Services/PostCoordinator.swift
leandroalonso: In this case, `break` or `return` will produce the same result. Do you prefer `return`?

We're not handling the error because the `autoSave` operation is silent. If it fails, we don't show anything to the user. I just kept this behavior.

WordPress/Classes/Services/PostCoordinator.swift
leandroalonso: Hmm, I'm not sure I got your question.

Here we're sending the error through a block. If the `error` is `nil` then we send an `unknown` error, because we don't know what happened. Otherwise, the `error` contains all relevant data about what happened.

If that doesn't answer your question please let me know.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
leandroalonso: In this specific case, we need `break`. I want lines 344-346 to execute, if I use `return` here this will never happen.

leandroalonso:
Thanks @yaelirub, I was able to reproduce the issue you mentioned. I'm gonna take a look.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
shiki: So, the problem that I see here is that there are 2 places where we create notices, `PostEditor.uploadPost()` and `PostCoordinator.upload()`:

https://github.com/wordpress-mobile/WordPress-iOS/blob/3848353955553208bae43afa37d2c4e9b68c7283/WordPress/Classes/ViewRelated/Post/PostEditor%2BPublish.swift#L345

https://github.com/wordpress-mobile/WordPress-iOS/blob/3848353955553208bae43afa37d2c4e9b68c7283/WordPress/Classes/Services/PostCoordinator.swift#L266-L268

They're both connected and it's a bit tricky to see when the notice is created.

I was going to suggest that we move the notice-creation to `PostCoordinator.save()` like so:

```swift
prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
	switch result {
	case .success(let post):
		self.upload(post: post) {
			// show notices if necessary
		}
	case .error(let error):
		// show a default notice
		ActionDispatcher.dispatch(NoticeAction.post(Notice(title: "Unknown error")))
		completion?(.error(error))
	}
}
```

But I realized that that method is used in a lot of places. The change in behavior, not to mention the required lifting of `PostEditor.uploadFailureNotice`, might just cause more harm at the moment.

On the other hand, I'm dubious of the usage of `shouldAttemptAutoUpload` here. It requires digging and understanding of how `shouldAttemptAutoUpload` is related to _showing notices_. It's quite complex and could easily break.

## Suggestion

Figuring out why `PostCoordinator` shows notices in the first place is out of the question so we can just go along with it. A minimal impact I can think of is adding a `defaultFailureNotice` parameter:

```swift
func save(_ postToSave: AbstractPost,
		  automatedRetry: Bool = false,
		  defaultFailureNotice: Notice? = nil,
		  completion: ((Result<AbstractPost>) -> ())? = nil) {
	prepareToSave(postToSave, automatedRetry: automatedRetry) { result in
		switch result {
		case .success(let post):
			self.upload(post: post, completion: completion)
		case .error(let error):
			if let notice = defaultFailureNotice {
			}
			completion?(.error(error))
		}
	}
}
```

With this, we don't need to have the bleed of `shouldAttemptAutoUpload` logic. It's not the best but it's the least amount of effort I can think of at the moment.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -326,7 +321,7 @@ extension PostEditor where Self: UIViewController {
         SVProgressHUD.show(withStatus: action.publishingActionLabel)
         postEditorStateContext.updated(isBeingPublished: true)
 
-        uploadPost() { [weak self] uploadedPost, error in
+        PostCoordinator.shared.save(post) { [weak self] result in
```
shiki: One less function to maintain 🎉

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -370,23 +370,6 @@ extension PostEditor where Self: UIViewController {
         self.postEditorStateContext.updated(isBeingPublished: false)
     }
 
-    /// Uploads the post
-    ///
-    /// - Parameters:
-    ///     - completion: the closure to execute when the publish operation completes.
-    ///
-    private func uploadPost(completion: ((_ post: AbstractPost?, _ error: Error?) -> Void)?) {
-        mapUIContentToPostAndSave(immediate: true)
```
shiki: This line seems important to retain. Are we sure about not keeping it?

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -197,7 +186,13 @@ extension PostEditor where Self: UIViewController {
         ActionDispatcher.dispatch(NoticeAction.clearWithTag(uploadFailureNoticeTag))
         stopEditing()
 
-        if post.canSave() {
+        /// If a post is marked to be auto uploaded and can be saved, it means that the changes
+        /// had been already confirmed by the user. In this case, we just close the editor.
+        /// Otherwise, we'll show an Action Sheet with options.
+        if post.shouldAttemptAutoUpload && post.canSave() {
```
shiki: Like my comment below, I'm concerned with the usage of `shouldAttemptAutoUpload` here. However, I could not find a better alternative aside from tracking the “Update” button press or tracking if there were recent changes since the last save. The `Post.hasUnsavedChanges()` also doesn't work for our purpose. 😞 

So, yeah, we can keep this for now. Thank you for adding some comments.

WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -402,3 +419,10 @@ extension PostCoordinator {
         }
     }
 }
+
+extension PostCoordinator {
```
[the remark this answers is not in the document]
leandroalonso: Done in e1dbb0dec9b1539ee68db975595ffbda92739d0f

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
leandroalonso: Nice catch!

Done in 1303bbe4cc7c646364d795f3377c5a15f8ebeabd

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
leandroalonso: Yes, I completely agree with you. :(

I tried some other approaches but they were more complex or error-prone.

shiki:
I forgot to mention:

> A key difference is the button should not close the editor

This is working the way that we want it so it's already “done”. I guess my description in #12589 wasn't very clear. I should probably have added:

- For _new_ posts, the top-right button should close the editor.
- When editing existing posts, the top-right button should **not** close the editor.

I just thought of mentioning it to clarify not to change the existing behavior.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
leandroalonso: Done in af8ab6f8a6dfe907084c4ebf2e8bb9c9466fa9fa

leandroalonso:
Ready for another review, @shiki. 👁

shiki:
This looks good to me. @yaelirub, I'm letting you merge this in case you still have follow-ups on your concerns.

yaelirub:
Thanks @shiki , as of recent discussions, we're leaving the merge to the author. So, @leandroalonso , whenever you're ready

shiki:
Going to merge this since I need it as the base for another PR.

PR #12691, opened 2019-10-14
Gutenberg: Update to v1.15.0

etoledom:
This PR updates Gutenberg to v1.15.0
`gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1445

**To test:**
 * Block editor: Fix issue when "New Photo Post" shortcut won't add the selected photo to the post.
* Add Link Target (Open in new tab) to Image Block settings.
* DarkMode improvements (https://github.com/wordpress-mobile/gutenberg-mobile/issues/1391 and https://github.com/WordPress/gutenberg/pull/17356).
* **New block** "Media & Text".
* General Editor smoke test.

Update release notes:


koke:
Tested on iPhone 8 iOS 12. I have the "Increase Contrast" accessibility setting enabled and I noticed the inserter button has a different border color than fill color:

IMG_2BE502BAEA9D-1

If I disable that setting, it has the same color:

IMG_C1C4A573C8FC-1

Also, I noticed a faint white separation between the border and the fill. This is really not noticeable on screen, but it came up as I was cropping the previous screenshot and zoomed in:

Screen Shot 2019-10-15 at 15 48 04

koke:
On iPhone XS iOS 13, it's hard to spot but the toolbar has a left border, which is barely visible on light mode but produces an odd glow effect on dark mode:

inserter-light-dark

----

RichText doesn't show a capitalized keyboard the first time it gets focus:

1. Create a new post
2. Tap on the placeholder text, keyboard should show uppercase letters
3. Dismiss the keyboard
4. Tap on the paragraph again, keyboard shows lowercase letters

I found this when inserting a media-text block, uploading an image, then trying to write the text, but it seems more generic to RichText

etoledom:
Thank you @koke for the review!

>  I have the "Increase Contrast" accessibility setting enabled and I noticed the inserter button has a different border color than fill color:

Yes, we made a trick with borders to make the icon look thicker. Definitely we should use a proper icon to solve this. Will create a ticket to track it.

But it's not regression so I'd say it's not a blocker 👍 

>  it's hard to spot but the toolbar has a left border, which is barely visible on light mode but produces an odd glow effect on dark mode

This also have been noted by @iamthomasbishop before (I believe there should be a ticket somewhere?). Sadly it's true that it's a lot more noticeable in dark mode. Again, not a regression so probably not a blocker.

etoledom:
I found a couple of issues:

The Hx buttons from the Heading block have a different color in DarkMode.
H_icon_color

This one looks bad: Adding a photo by taking it with the camera, the image get shifted to the side.
This doesn't seem to be a problem with `gutenberg` though, since it also happens adding a photo directly to the Media Library.
Image

I wouldn't say any of these are blockers, so ✅ from my side on the release.

iamthomasbishop:
@koke 

> On iPhone XS iOS 13, it's hard to spot but the toolbar has a left border, which is barely visible on light mode but produces an odd glow effect on dark mode:

> This also have been noted by @iamthomasbishop before (I believe there should be a ticket somewhere?)

I've been hoping to get this removed for a while – [related issue here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/639).

------

@etoledom 👇 

> The Hx buttons from the Heading block have a different color in DarkMode.

Def not a blocker, but important to fix :)

> Adding a photo by taking it with the camera, the image gets rotated.
This doesn't seem to be a problem with gutenberg though, since it also happens adding a photo directly to the Media Library.

I think this isn't a blocker for this issue, but can we make sure to escalate this to ensure it's fixed asap?

etoledom:
> I think this isn't a blocker for this issue, but can we make sure to escalate this to ensure it's fixed asap?

There's already a PR for this:
https://github.com/wordpress-mobile/MediaPicker-iOS/pull/338

PR #12706, opened 2019-10-16
[Gutenberg] Image Size networking

etoledom:
This PR implements the changes on the gutenberg bridge to send the wpcom headers and site slug/id to gutenberg.

gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1453
gutenberg PR: https://github.com/WordPress/gutenberg/pull/17979
WPKit PR: 

To test:
- Add a image block to gutenberg.
- Select an image from the WordPress media Library.
- Select settings and choose a different image setting.
- Change to HTML mode.
- Check that the image URL has changes according to the selected image size.

Update release notes:

- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12714, opened 2019-10-17
[Gutenberg] Image Size networking (native)

etoledom:
Alternative to https://github.com/wordpress-mobile/WordPress-iOS/pull/12706

As a proof of concept, this PR implements the minimum necessary to make Image Size works.
In the future we might need to create a more generic api call with a dynamic request method (given from gutenberg). We also might need to handle other api version Paths.

Related PRs:
`gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1455

**To test:**
- **Run metro server on https://github.com/wordpress-mobile/gutenberg-mobile/pull/1455**
- Add a image block to gutenberg.
- Select an image from the WordPress media Library.
- Select settings and choose a different image setting.
- Change to HTML mode.
- Check that the image URL has changes according to the selected image size.

Update release notes:


WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    private let path: String
+    private unowned let blog: Blog
+
+    init(path: String, blog: Blog) {
+        self.path = path
+        self.blog = blog
+    }
+
+    func request(response: @escaping Response) {
```
koke: I found quite confusing the `response`/`Response` naming for a callback, when the argument to that callback will be the actual response. Maybe we can use the more generic `completion`?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
```diff
@@ -0,0 +1,62 @@
...
+            case .success(let responseObject):
+                response(.success(responseObject))
+            case .failure(let error):
+                response(.failure(error as NSError))
+            }
+        }
+    }
+
+    private var selfHostedPath: String {
+        let removedEditContext = path.replacingOccurrences(of: "context=edit", with: "context=view")
```
koke: Switching the context to `view` happens to work fine for this specific use case, but it likely won't for others. I don't like the idea of implementing it generically and pretending that `apiFetch` can now work for anything else, as it might lead to subtle undetected bugs.

Since I understand this is a temporary solution to an immediate problem, I think we should enforce that this system only works with a known set of API paths. So for any request other than getting media sizes, I would probably return a `Not implemented` error, or ignore the request and just log a warning.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
```diff
@@ -0,0 +1,62 @@
+import Alamofire
+
+struct GutenbergNetworkRequest {
```
koke: This is just a bit of a rant, but specially after doing JS for a while, I find interesting how reluctant we all are to use top level functions on Swift. There's no point in rewriting this, but it's something that keeps catching my attention 😁

WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
mchowning: I think adding this kind of guard makes sense. Assuming we always expect that platforms to be in sync, it would be nice to have that guard on the JS-side so that we don't have to replicate it on both platforms.

etoledom: I thought on having some kind of `white list` of paths that we accept to avoid surprises.
This implementation also only do `GET` requests so far, so there's another reason for this. 👍
I agree that is probably better to do is on JS side, then probably this context change should also go on the JS side?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
etoledom: I like the Struct to pack some properties inside it, and have methods with less arguments.
At first I did static functions (that would work as well as top level functions), but I didn't like having all the internal helper functions needing the whole set of arguments.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
etoledom: sounds good! 👍

WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
mchowning: >I agree that is probably better to do this on JS side, then probably this context change should also go on the JS side?

👍  The more path manipulation we move to the JS side, the better imo. I was planning to start investigating that once I got the Android implementation wrapped up.

etoledom: The SiteID is a WPCom thing, probably we don't want it on gutenberg repo.

mchowning: Good point @etoledom

mchowning:
I'm currently seeing an error when try to update the size of an image for a self-hosted site. I'll try to investigate this some tomorrow. I am seeing this show up in the logs, although I don't know whether it is related yet: 

```
2019-11-11 16:35:36.819134-0500 WordPress[15184:4171347] CredStore - performQuery - Error copying matching creds.  Error=-50, query={
    "m_Limit" = "m_LimitAll";
    ptcl = htps;
    srvr = "damp-ducks.jurassic.ninja";
    sync = syna;
}
```

mchowning:
Hey @etoledom ! I merged `gutenberg/release-1.17.0` into this and updated the PR to target that release branch.

etoledom:
There are a few known issues: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1455#issuecomment-554450519 - I'm hoping those are not blockers for this PR (I'm currently working on those).

Keeping that in mind, I believe this is ready for another 👀 
(cc @SergioEstevao @koke )

loremattei:
Hey there! I've moved this to 13.8 since 13.7 has been cut. If you want this to make it to 13.7, please feel free to ping me.

PR #12721, opened 2019-10-17
[Stats Refresh] Insights detail view async loading

### LINKED ISSUE #12424: [Stats Refresh] Insights: Load overview block independently in details view
On Insights we should load all the details view using the ghost animation

danielebogo:
Fixes #12424 

This PR removes the NRV from the Insights detail view and uses the async loading + ghost view approach instead.
The section affected are:
• _Followers_
• _Comments_
• _Tags & Categories_
• _This Year_

## To test:
• Select Stats from a site with data (en.blog is a good candidate) and select _Insights_
• Make sure the sections mentioned above are present in _Insights_
• For every section click on _View more_ and open the detail
• You should see the Ghost views before the content is loaded

Update release notes:


ScoutHarris:
FWIW, https://github.com/wordpress-mobile/WordPress-iOS/pull/11881 disabled showing 'View more' for _Comments_, so that is untestable.

ScoutHarris:
Hey @danielebogo .

When loading _This Year_ details, there is a grey gap at the top where the date bar is supposed to be. It looks kinda strange.

Loading:
this_year_loading

After loading:
this_year_loaded

WordPress/Classes/Stores/StatsStore+Cache.swift
```diff
@@ -0,0 +1,52 @@
...
+}
+
+extension StatsInsightsStore: StatsStoreCacheable {
+    func containsCachedData(for type: InsightType) -> Bool {
+        switch type {
+        case .latestPostSummary:
+            return state.lastPostInsight != nil
+        case .allTimeStats:
+            return state.allTimeStats != nil
+        case .followersTotals:
```
ScoutHarris: Is this necessary, since `default` returns `false`?

danielebogo:
@ScoutHarris thanks for the review! Header and enumeration are ok now! Ready for another look!

PR #12722, opened 2019-10-17
Show snackbar with "We'll publish the post" on failure to upload post with media

### LINKED ISSUE #12678: [Offline Posting] Show snackbar with "We'll publish the post" on failure to upload post with media
From Design Review: “We'll publish the post...” Snackbar is not shown when publishing a post with media while offline.

It shows a “1 post, 1 file not uploaded” instead.

This used to work so it seems that some of our latest work caused this regression.

Steps:
1. Be offline 
2. Create a post with media
3. publish the post
4. See the notice "1 post, 1 file not uploaded"

leandroalonso:
Fixes #12678

## Testing

Please, use Gutenberg and Aztek for testing.

1. While offline, create a post and add an image
2. Tap Publish (or Save, or Schedule)
3. Check that the snack bar message says the post will be published later

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -345,8 +345,10 @@ class PostCoordinator: NSObject {
     }
 
     private func dispatchNotice(for post: AbstractPost) {
-        let model = PostNoticeViewModel(post: post)
-        ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+        DispatchQueue.main.async {
+            let model = PostNoticeViewModel(post: post)
+            ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+        }
     }
 }
 
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -245,9 +245,8 @@ class PostCoordinator: NSObject {
             ActionDispatcher.dispatch(NoticeAction.post(model.notice))
 
             completion?(.success(uploadedPost))
-        }, failure: { error in
-            let model = PostNoticeViewModel(post: post)
-            ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+        }, failure: { [weak self] error in
+            self?.dispatchNotice(for: post)
 
             completion?(.error(error ?? SavingError.unknown))
 
@@ -316,9 +315,11 @@ class PostCoordinator: NSObject {
         guard let context = post.managedObjectContext else {
             return
         }
+
         context.perform {
             if status == .failed {
                 self.mainService.markAsFailedAndDraftIfNeeded(post: post)
+                self.dispatchNotice(for: post)
             } else {
                 post.remoteStatus = status
             }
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Services/PostCoordinator.swift
```diff
@@ -316,9 +315,11 @@ class PostCoordinator: NSObject {
         guard let context = post.managedObjectContext else {
             return
         }
+
         context.perform {
             if status == .failed {
                 self.mainService.markAsFailedAndDraftIfNeeded(post: post)
+                self.dispatchNotice(for: post)
```
shiki: Not sure if I agree with this. The `change()` method originally was supposed to manage transitions between statuses. I wasn't even sure about `self.mainService.markAsFailedAndDraftIfNeeded()` in here. 

Adding a UI call is a different story though. 😐 This method can potentially be used anywhere we need a `status` transition and it can be misleading and surprising that it also shows a notice. 

Would you mind explaining more why we have to do it this way? Is there another way?

leandroalonso: You're right.

Here's why: when you have a failed image in a post the `upload` method will never be called in `save`. Thus, `upload`'s failure block will also never be reached. This means that there's no snackbar message in this case.

I've added it to `change()` because of that, to fill this gap. I can add this call in all the places the post is changed to `failed` in `prepareToSave` but then we're back to the original problem.

Thoughts?

shiki: If we only need to handle `PostCoordinator.save()`, I think we can get away with adding the dispatch call there:

```diff
diff --git a/WordPress/Classes/Services/PostCoordinator.swift b/WordPress/Classes/Services/PostCoordinator.swift
index 9f90895da1..21122d2d5e 100644
--- a/WordPress/Classes/Services/PostCoordinator.swift
+++ b/WordPress/Classes/Services/PostCoordinator.swift
@@ -56,9 +56,16 @@ class PostCoordinator: NSObject {
             case .success(let post):
                 self.upload(post: post, completion: completion)
             case .error(let error):
-                if let notice = defaultFailureNotice {
-                    ActionDispatcher.dispatch(NoticeAction.post(notice))
+                switch error {
+                case SavingError.mediaFailure:
+                    let model = PostNoticeViewModel(post: postToSave)
+                    ActionDispatcher.dispatch(NoticeAction.post(model.notice))
+                default:
+                    if let notice = defaultFailureNotice {
+                        ActionDispatcher.dispatch(NoticeAction.post(notice))
+                    }
                 }
+
                 completion?(.error(error))
             }
         }
```

☝️ This code only _kinda_ works. The message is not correct but is probably worth it to look deeper into?

leandroalonso: I've tried that. Here and in `PostEditor+Publish` — which I think makes even more sense, since will get rid of UI stuff in `PostCoordinator`.

However, this doesn't work because of... synchronism (hello darkness). By the time this is called, `post.status` is still `pushing` and it's not yet marked to be auto uploaded. Thus, we always end up with "Post failed to upload" message. :( 

I can look deeper into it but this will probably lead us back to the sync/async stuff. I'm not sure if we wanna do that. 😅

shiki: Hmm. 🤔 I wonder if we can add a completion block to `change()`. It's a bit ugly but maybe it'll work? 

```swift
                    self.change(post: post, status: .failed) {
                        completion(.error(SavingError.mediaFailure))
                    }
```

It seems low risk (famous last words).

leandroalonso: @shiki I refactored to allow the `change()` method to execute a completion block.

It's still not something that I'm proud of but this prevents unwanted behaviors in `change()`. Let me know what you think.

PR #12739, opened 2019-10-21
Fix infinite upload

### LINKED ISSUE #12678: [Offline Posting] Show snackbar with "We'll publish the post" on failure to upload post with media

shiki:
Closes #12678 

This is a continuation of #12722. 

## Findings

During #12722, we found an issue where an infinite media upload is executed when using Gutenberg. Here is the scenario: 

1. Edit an existing post while offline
2. Add a media
3. Tap on _Update_

We found that Gutenberg was always adding a media after the tap. We initially fixed this by adding a workaround (7a106d1). 

## Solution

After further investigation, I found that `prepareToSave`'s `completion` block was being called multiple times if there are multiple `Media` attached to a `Post`. After fixing this in 
1cfa7d9, I found that this fixes the infinite upload as well. 

The workaround was removed in db1070c. There is still an outstanding issue to this where images can still get uploaded multiple times when running the same scenario but the infinite upload _while in the editor_ does not happen anymore. This (now) unrelated issue will be reported in Gutenberg. 

## Testing

- Please run the scenario described above and validate that an infinite upload no longer happens.
- Please also test that uploading with a media will still show the correct notice

## Reviewing 

- Only 1 reviewer is needed but anyone can review. 
- Please merge it if you approve of the change and are not requesting any changes. 

## Submitter Checklist


leandroalonso:
@shiki Looks good to me and it works as expected!

Can we create the Gutenberg issue before merging this PR? What do you think?

shiki:
@leandroalonso I created https://github.com/wordpress-mobile/WordPress-iOS/issues/12744.

PR #12741, opened 2019-10-21
Issue/fix rotation of photos

### LINKED ISSUE #12703: Aztec: vertical images appear horizontal in previews and on the web on self-hosted and Jetpack sites
Steps to reproduce:

1. Open a self-hosted or Jetpack site.
1. Create a new post and add a title. 
1. Add a horizontal image from the device.
1. Add a horizontal image by taking a photo.
1. Preview the post and observe whether the image appears horizontal or vertical.
1. Publish the post.
1. View the post in a browser and observe whether the image appears horizontal or vertical.

**Result:** vertical images uploaded to self-hosted or Jetpack sites appear horizontal in previews and on the web. ([3m28s](https://cloudup.com/cUY4lOz58Q1))

right-image-orientiation-in-aztec wrong-image-orientation-on-web-and-in-previews-for-images-added-via-aztec
<sup>Tested with WPiOS 13.4.0.1 (TestFlight) on iPhone 6S iOS 13.2.</sup>

SergioEstevao:
Fixes #12703 

This PR bings fixes for two issues:

 - Images that were captured inside the app using a non-standard orientation of the device (ex: upside down) were being saved incorrectly in the camera roll
- Images that have a non-standard orientation, capture inside the app or in the camera app, when updated to a WP.org site that was is bellow version 5.3 will be show incorrectly on the web.

For the first issue we are updating the WPMediaPicker version that has a fix to save the images correctly when taken inside the app

For the second issue we are now applying a transform on all images uploaded from the app to make sure that they all have orientation up independently of the size setting. 

To test:
 - Capture a photo inside the app with the device upside down
 - Make sure the image is recorded with the correct orientation and display properly inside the media library
 - Use that photo on a post and upload it to a WP.org site that is bellow version 5.3 ( it cannot be connected with Jetpack or else the Photon will fix/hide the issue)
- Upload/Publish the Post
- See the post on the web, using Safari, to see if photo shows with the correct orientation.

Update release notes:


designsimply:
I tested with `develop` with the using the iPhone 11 simulator and I still couldn't get it to work. ([1m31s](https://cloudup.com/cKIv3ups3Na))

@SergioEstevao I used App Settings > Max Image Upload Size = Original on a WP 5.2.4 self-hosted site and I should be seeing images shot vertically stay vertical in those conditions right?

SergioEstevao:
> I tested with develop with the using the iPhone 11 simulator and I still couldn't get it to work. (1m31s)
> @SergioEstevao I used App Settings > Max Image Upload Size = Original on a WP 5.2.4 self-hosted site and I should be seeing images shot vertically stay vertical in those conditions right?

It needs to be tested on `issue/fix_rotation_of_photos` branch

WordPress/Classes/Utility/Media/MediaImageExporter.swift
```diff
@@ -284,58 +284,48 @@ class MediaImageExporter: MediaExporter {
...
+
+            // Create a thumbnail of the image source.
+            guard let image = CGImageSourceCreateThumbnailAtIndex(source, 0, thumbnailOptions as CFDictionary) else {
+                throw ImageExportError.imageSourceThumbnailGenerationFailed
+            }
+
+            if nullifyGPSData == true {
+                // When removing GPS data for a thumbnail, we have to remove the dictionary
+                // itself for the CGImageDestinationAddImage method.
+                imageProperties.removeValue(forKey: kCGImagePropertyGPSDictionary)
```
frosty: I'm a bit uncertain here because we were previously removing the GPS data differently depending on whether we were dealing with a full sized image or a thumbnail. Do we still need the alternative code path too?

frosty: I'm a bit uncertain here because we were previously removing the GPS data differently depending on whether we were dealing with a full sized image or a thumbnail. Do we still need the alternative code path too?

WordPress/Classes/Utility/Media/MediaImageExporter.swift
```diff
@@ -284,58 +284,48 @@ class MediaImageExporter: MediaExporter {
...
             }
 
+            // Add the thumbnail image as the destination's image.
+            CGImageDestinationAddImage(destination, image, imageProperties as CFDictionary?)
+
+            // Get the dimensions from the CGImage itself
+            width = CGFloat(image.width)
```
frosty: Same goes for the width and height, we were getting this from the image properties for thumbnails?

SergioEstevao: Yes when doing this code path without a size set we will get the original size of the image. So it’s ok. You can confirm by  checking the size of the image uploaded to the server using the original size setting.

WordPress/Classes/Utility/Media/MediaImageExporter.swift
SergioEstevao: We are removing it from the properties so the new image should have any GPS info.

Issue #12686, opened 2019-10-11
Post Scheduling Enhancements Master Issue


ScoutHarris:
Expand options when creating a scheduled post ([#](https://wp.me/paCL20-ac-p2)).

- [ ] Add feature flag for local notification option.
- [ ] Add ability to set up local notification.
- [ ] Add feature flag for 'Add to calendar' option.
- [ ] Add ability to add to calendar.
- [ ] Ensure views are correct for Dark Mode.
- [ ] Add Accessibility support. (_Some work done in #12959_)
- [ ] Add Tracks.
- [ ] Enable feature flags for internal.
- [ ] Enable feature flags for all.
- [ ] Remove feature flags.

---
#### 14.0:
- [x] Post Scheduling: post date doesn't match messaging (#12992)
- [x] Post Scheduling: Calendar allows vertical scrolling while changing months (#13113)
- [x] Enable Feature Flag for Post Scheduling Calendar Picker (#13133)
- [x] Posts List uses device timezone while Post Settings uses site timezone (#13258)

#### 13.7:
- [x] Add feature flag for publish date/time. (https://github.com/wordpress-mobile/WordPress-iOS/pull/12820)
- [x] Add new 'Publish' flow to select publish date and time. (https://github.com/wordpress-mobile/WordPress-iOS/pull/12820)

---
![post_scheduling_ios](https://user-images.githubusercontent.com/1816888/66790758-6784fe00-eeae-11e9-8c8b-91bb84afb2dd.jpg)

ScoutHarris:
@osullivanchris @frosty @bjtitus

It occurred to me that the three main parts of this project can stand alone as independent features:
- [ ] Add new 'Publish' flow to select publish date and time.
- [ ] Add ability to set up local notification.
- [ ] Add ability to add to calendar.

Is there any reason we can't feature flag them separately, so each can be released as they're complete (after the trial)?
- @bjtitus already added a `postScheduling` flag to show the new 'Publish' view with the date and time selectors.
- Another flag for the notification option.
- Another flag for the calendar option.

@bjtitus is making great progress, but just in case, I didn't see why any incomplete part(s) should hold up the others.

Thoughts?

bjtitus:
I think that makes sense. The only thing that could be a blocker for notifications and calendar is the new screen. Right now the flag controls the screen appearing at all, so turning ok Notifications or Calendar Event would necessitate turning the new Publish flow on as well.

[...]

PR #12749, opened 2019-10-22
Add Calendar Picker + new Publish Settings screen

bjtitus:
Ref #12686

This addresses the first items from the master issue:
- [x] Add feature flag. (`postScheduling`)
- [x] Add new 'Publish' flow to select publish date and time.

<div align="center">
Napkin 2 10-21-19, 10 27 23 PM
</div>

To test:
- Navigate to the `Published` setting on a blog post.
- Tap on the Published cell
- Review new Publish settings screen

A few things to point out:
- Adds a new dependency on [JTAppleCalendar](https://github.com/patchthecode/JTAppleCalendar)
- The new `HalfScreenPresentationController` might be useful in other parts of the app and could possibly be moved to WordpressUI alongside its `FancyAlertPresentationController` parent

There are a few known issues I am hoping to address shortly:
- Rotation to landscape does not work well with the new `HalfScreenPresentationController`
- The bottom sheet should probably be in a popover but that may not be critical for shipping.

Please feel free to leave any feedback or ask any questions about the design. I tried to stick with in-code view layout/setup because most of the global styling stuff seemed much more easily accessible that way.


Podfile.lock
```diff
@@ -500,9 +504,9 @@ SPEC CHECKSUMS:
   WPMediaPicker: 1897f312c7b41114ffd239fb782431ae602134a1
   wpxmlrpc: 6ba55c773cfa27083ae4a2173e69b19f46da98e2
   yoga: 4e71c9a33abf45ba55af55ae9cbc86f4234bb2a9
-  ZendeskSDK: 07977c9aa708ea64461a1b18b7b3cfec3f496785
+  ZendeskSDK: 787414f9240ee6ef8cfe4ea0f00e8b4d01d2d264
```
ScoutHarris: I think `Podfile.lock` needs to be updated. When I installed pods in this branch I got a change on `ZendeskSDK`.

ScoutHarris:
Hey @bjtitus .

A few things I noticed.

Functional:
- Once the calendar is presented, it cannot be dismissed. Maybe add a tap action to dismiss (along with any changes)?

Design - Just some minor differences compared with Zeplin:

(Normally I'd include screenshots, but GH isn't letting me upload atm. Hopefully this all makes sense.)

- On the calendar, the separator line between the top buttons and "Choose a date" looks a bit thick.
- "Choose a time" time display: Zeplin has a comma instead of "at".
- "Date and Time" display on the "Publish" view: Zeplin displays the full date.
- Footer text on the "Publish" view: Zeplin's wording is slightly different than yours.

In the places that display the time, I do agree with the addition of AM/PM. 👍

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+                )
+            }
+        }
+
+        let footerText: String?
+
+        if let date = viewModel.date {
+            let publishedOnString = PublishSettingsController.dateFormatter.string(from: date)
+            let offsetLabel = viewModel.timeZone?.label ?? "Unknown Offset"
+            footerText = "Post will be published on \(publishedOnString) in your site time zone (\(offsetLabel))"
```
ScoutHarris: The `footerText` should be localized (including the optional "Unknown Offset"). 

Also worth noting, we don't use string interpolation with localized text since it causes issues for the translators. We typically use `String.localizedStringWithFormat` to accomplish both.

ScoutHarris:
Hey @bjtitus .

I noticed you added this notice message for differing time zones. (I don't recall seeing it in any of the designs.)

notice

It looks kind of odd to me, and I'm not sure it's necessary. But... @osullivanchris what do you think?

WordPress/Classes/ViewRelated/Post/CalendarViewController.swift
```diff
@@ -0,0 +1,156 @@
...
+        }
+        self.datePicker = datePicker
+        return datePicker
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        chosenValueRow.titleLabel?.text = NSLocalizedString("Choose a time", comment: "Label for Publish time picker")
+        chosenValueRow.detailLabel?.text = TimePickerViewController.dateFormatter.string(from: datePicker.date)
+        let doneButton = UIBarButtonItem(title: "Done", style: .done, target: self, action: #selector(done))
```
ScoutHarris: "Done" should also be localized.

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
```diff
@@ -0,0 +1,179 @@
+import Foundation
+import CocoaLumberjack
+import WordPressShared
+
+private enum PublishSettingsCell: CaseIterable {
+    case dateTime
+
+    var title: String {
```
ScoutHarris: Is this used? I commented it out with no ill effect. Maybe it's not needed anymore?

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+            timeZone = nil
+        }
+    }
+
+    fileprivate var cells: [PublishSettingsCell] {
+        switch state {
+        case .published, .immediately:
+            return [PublishSettingsCell.dateTime]
+        case .scheduled:
+            return PublishSettingsCell.allCases
```
ScoutHarris: OK, I admit, I don't quite understand why this is returning `allCases`. Can you explain a bit please?

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+            noticeMessage = NSLocalizedString("Site time zone differs from device time zone", comment: "Notice that time zones are different when scheduling")
+        }
+    }
+
+    func tableViewModelWithPresenter(_ presenter: ImmuTablePresenter) -> ImmuTable {
+        return mapViewModel(viewModel, presenter: presenter)
+    }
+
+    func refreshModel() {
+        // Don't need to refresh the model
```
ScoutHarris: Can we add more comments here explaining why this method can't be removed? Of course, the first thing I did was remove it, and _now_ I see why it has to be there.  😄

ScoutHarris:
Hey @bjtitus .

Just some general thoughts to consider.

I think it would be helpful to move all your files to one project directory so it's clear what they're for (including the ones that are currently in `Views`). Right now, since the file names are a bit generic, it's not immediately clear what they're for. Maybe under `Post` (where most of your files are currently), we can have a `Scheduling` folder or something? 

Regarding file/class names, we do try to make them a bit more specific, again so it's more clear what they're used for. Typically, we don't generalize files/classes "just in case". If, in the future, we decide to use them more widely, we can rename/move them accordingly.

I'll let you ponder that, but I do think at least putting them in a properly named directly will most definitely help.


And, finally, despite peppering you with changes, I think this is a great start! Great job figuring out our codebase and reusing existing components! This is looking very nice and already a huge improvement.

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
bjtitus: Some of this is from code I had prepared for the Notifications and Calendar cases. Was planning to add cell types for each of those and include them here. So in the scheduled case, this will include the `.notification` and `.calendar` type also. Might be better to be more explicit and just use an array for this:

`[PublishSettingsCell.dateTime, PublishSettingsCell.notifications, PublishSettingsCell.calendar]`

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
bjtitus: Sure, I can flesh this comment out. It's a required method for `SettingsController` that isn't really used here.

WordPress/Credentials/ApiCredentials.m
```diff
@@ -1,7 +1,7 @@
 #import "ApiCredentials.h"
 
-#define WPCOM_API_CLIENT_ID @""
-#define WPCOM_API_CLIENT_SECRET @""
+#define WPCOM_API_CLIENT_ID @"67049"
```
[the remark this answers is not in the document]
bjtitus: Seems like a lot of these got accidentally included >_< and merged wrong. I'll clean them up.

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
bjtitus: This isn't used currently but I was planning to use it later on when their are multiple cell states. I've removed if for now and will re-add in a future PR if needed.

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -12650,7 +12677,7 @@
 				MTL_ENABLE_DEBUG_INFO = NO;
 				PRODUCT_BUNDLE_IDENTIFIER = "org.wordpress.alpha.$(PRODUCT_NAME:rfc1034identifier)";
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				PROVISIONING_PROFILE_SPECIFIER = "match InHouse org.wordpress.alpha.WordPressNotificationContentExtension";
+				PROVISIONING_PROFILE_SPECIFIER = "WordPress Alpha Notification Content Distribution";
```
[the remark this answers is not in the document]
bjtitus: I believe this was the same bad merge. This isn't my provisioning profile specifier but I saw it had changed with the latest commits.

Podfile.lock
jkmassel: Just FYI – this one is ZenDesk's fault :(

bjtitus:
@ScoutHarris 

🔔 

> I noticed you added this notice message for differing time zones. (I don't recall seeing it in any of the designs.)

I added that notice message after being confused about the wrong date showing up for a site I had that was set to UTC. I thought it might gel with the original feature request to be more aware of when exactly a scheduled post is going to happen by alerting the user to the fact that their device is not in the same time zone they are scheduling for.

Happy to remove it if it's confusing or useless!

<hr>

📅 

> * "Choose a time" time display: Zeplin has a comma instead of "at".
> * Footer text on the "Publish" view: Zeplin's wording is slightly different than yours.

I believe these two text differences stem from a decision to use the built in Date Formatter styles. I'm a huge fan of using these system provided styles to match what the user expects to see in a formatted date. For the footer, that meant changing the first `at` to `on` to avoid repetition (`on October 15, 2013 at 7:55pm`).

Custom format strings often completely change the information to something the user doesn't expect to see. The Japanese calendar is one I typically check which uses the year of the current emperor with the emperor's name. The normal English date format string wouldn't account for this. The removal of AM/PM is also automatic based on the locale instead of having to decide whether to include the `a` format string component. There are a whole host of those kinds of small tweaks that the provided styles handle without needing to write logic for them.

### Custom Date Format String

**Code**
```swift
let formatStringFormatter = DateFormatter()
formatStringFormatter.calendar = Calendar(identifier: .japanese)
formatStringFormatter.locale = Locale(identifier: "ja")
formatStringFormatter.dateFormat = "MMMM dd, yyyy, hh:MM a"
formatStringFormatter.string(from: date)
```

**English Output**
```
October 22, 2019, 08:10 PM
```

**Japanese Output**
```
10月 22, 0001, 08:10 午後
```

### Date Formatter Styles

**Code**
```swift
let formatter = DateFormatter()
formatter.calendar = Calendar(identifier: .japanese)
formatter.locale = Locale(identifier: "ja")
formatter.dateStyle = .long
formatter.timeStyle = .short
formatter.string(from: date)
```

**English Output**
```
October 22, 2019 at 8:12 PM
```

**Japanese Output**
```
令和元年10月22日 20:06
```

Maybe @osullivanchris has ideas on whether it's more critical to keep the custom date format strings here (`<date>, <time>` vs. `<date> at <time>` in English) or use the system provided syntax that will adjust based on locale and calendar?

osullivanchris:
> Maybe @osullivanchris has ideas on whether it's more critical to keep the custom date format strings here (<date>, <time> vs. <date> at <time> in English) or use the system provided syntax that will adjust based on locale and calendar?

You've done some good thinking on this and I agree with you, we should go with the system to avail of the localisation etc that comes built in. I don't think we had made a big deliberate decision here, so I'll defer to the system.

> I added that notice message after being confused about the wrong date showing up for a site I had that was set to UTC. I thought it might gel with the original feature request to be more aware of when exactly a scheduled post is going to happen by alerting the user to the fact that their device is not in the same time zone they are scheduling for.

This is an interesting use case and it sounds like a good idea to me. I don't have a build to play with. When does it appear? Always on the screen, or after the user changes the time, or something different?

Not sure on the blue bar. Is it something we use elsewhere? Looks like system UI, but the blue is awkwardly similar but different to our nav blue. 

To ask a couple of teasing questions - do you think it needs to be that prominent and did you have any other ideas? For example, the text where we tell you "Post will be published at X days in your site timezone (UTC +1)" is small and not highlighted. Do you think when the timezone on the device is different to the site that it justifies being made extra prominent? I don't have the right answer I'm just thinking this through myself. Perhaps the alert could be tied to or appended to that text in some way.

bjtitus:
@osullivanchris Thanks for taking a look!

> When does it appear? Always on the screen, or after the user changes the time, or something different?

I just had it appearing any time the device timezone is different than the site. That may be overkill with a lot of shared sites where users from many different timezones are contributing.

> Not sure on the blue bar. Is it something we use elsewhere? Looks like system UI, but the blue is awkwardly similar but different to our nav blue.

I just ended up using an existing notice from the existing code in the app. I'm not sure where the color comes from but it's the same used for some data loading errors.

> Do you think when the timezone on the device is different to the site that it justifies being made extra prominent? I don't have the right answer I'm just thinking this through myself. Perhaps the alert could be tied to or appended to that text in some way.

I like the idea of adding it to the footer text. We could probably just add as a new line so it isn't quite so in your face.

I think I'll pull the alert for now and add it as a future issue so it's not blocking this PR. It's only ~4 lines of code.

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
ScoutHarris: That'd be great! Thanks!

WordPress/Classes/ViewRelated/Post/PublishSettingsViewController.swift
ScoutHarris: Oh! _Now_ I get it. Thanks for the explanation. When you add `notification` and `calendar` it will make more sense; then it will be obvious what it's doing. I'm not sure an explicit array will be needed then, but we'll see when you get to that part.

ScoutHarris:
@osullivanchris - just answer this query - 

> Not sure on the blue bar. Is it something we use elsewhere?

The blue notice seems to be intended to inform the user an action was taken, specifically in Settings pages. For example, it's used in Account Settings for `There is a pending change of your email` or `Username changed to`.

It's not really meant to just be informational. Given that (and it's lack of attractiveness), @bjtitus I very much agree with this plan. 👍 

> I like the idea of adding it to the footer text. We could probably just add as a new line so it isn't quite so in your face.

Podfile.lock
```diff
@@ -59,6 +59,7 @@ PODS:
   - HockeySDK (5.1.4):
     - HockeySDK/DefaultLib (= 5.1.4)
   - HockeySDK/DefaultLib (5.1.4)
+  - JTAppleCalendar (8.0.2)
```
ScoutHarris: `Podfile.lock` still needs to be updated. I got local changes again, this time on `Giphy`.

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarMonthView.swift
```diff
@@ -0,0 +1,118 @@
...
+
+        let nextButton = UIButton(frame: CGRect(x: 0, y: 0, width: 24, height: 24))
+        nextButton.setImage(Gridicon.iconOfType(.chevronRight).imageFlippedForRightToLeftLayoutDirection(), for: .normal)
+        nextButton.setContentHuggingPriority(.defaultHigh, for: .horizontal)
+        nextButton.addTarget(self, action: #selector(CalendarMonthView.nextMonth), for: .touchUpInside)
+
+        let headerLabel = UILabel()
+        headerLabel.text = "February, 2019"
+        headerLabel.textAlignment = .center
+//        headerLabel.font = UIFont.preferredFont(forTextStyle: .caption1)
```
ScoutHarris: Can this be removed?

ScoutHarris:
Hey @bjtitus . One minor visual glitch. When the 'Date and Time' row is selected, it remains grey:

row_selected

Podfile.lock
bjtitus: What's the practice you generally use for this? I used `pod update JTAppleCalendar` to avoid changes to other pods since the previous changes seemed like they were unwanted in this PR. If I delete the lockfile and re-run `pod install` I get quite a few updates. If I run `pod install` I see only this one.

bjtitus:
Thanks for checking the latest changes @ScoutHarris!

I think I've addressed everything except the Podfile.lock issue. I'm not quite sure what the best thing to do is there. If I `pod install` without touching the `Podfile.lock`, I don't get any further changes. If I delete the lockfile or run `pod update` I get significantly more updates than what you mention.

Also note that the row deselection change is a bit of a hack. The problem is that the `UITableView` (and all other code I can find in this app) rely on the view appearance methods, which are not called in a non-fullscreen presentation context. I added a manual deselection method but don't really love the solution. Let me know if you have any thoughts!

Podfile.lock
bjtitus: @ScoutHarris Ignore this. I ended up getting the new Zendesk checksum. Not sure why it didn't work the first few times.

ScoutHarris:
Hey @bjtitus .

> Also note that the row deselection change is a bit of a hack. 

So, I had a think about this, and I think maybe creating a custom `ImmuTableRow` might be a cleaner solution. That will let us set the `selectionStyle`, instead of `NavigationItemRow`.

We do this sort of thing often to customize cells for a particular view. And I don't think a `selectionStyle` is needed here. 

So in `PublishSettingsViewController`, we could add something like:
```
    private struct DateAndTimeRow: ImmuTableRow {
        static let cell = ImmuTableCell.class(WPTableViewCellValue1.self)
        
        let title: String
        let detail: String
        let action: ImmuTableAction?
        let accessibilityIdentifer: String
        
        init(title: String, detail: String, action: @escaping ImmuTableAction, accessibilityIdentifier: String) {
            self.title = title
            self.detail = detail
            self.action = action
            self.accessibilityIdentifer = accessibilityIdentifier
        }
        
        func configureCell(_ cell: UITableViewCell) {
            cell.textLabel?.text = title
            cell.detailTextLabel?.text = detail
            cell.selectionStyle = .none
            cell.accessibilityIdentifier = accessibilityIdentifer
            
            WPStyleGuide.configureTableViewCell(cell)
        }
    }
```

What I've done is basically creating a cell like `NavigationItemRow` does, removed the unnecessary properties, and set `selectionStyle = .none`. (To note, we do need to set `accessibilityIdentifier`. I didn't notice that initially.)

And then of course when the cell is created, use `DateAndTimeRow` instead of `NavigationItemRow`.

What do you think about this?

bjtitus:
@ScoutHarris Sounds good to me! I've made that change. Let me know if there's anything else I can change here.

ScoutHarris:
Hey @bjtitus .

Sorry, I just noticed the time selector isn't displaying the time anymore.

time

bjtitus:
@ScoutHarris Sorry about that! Replaced with the wrong date formatter when I moved them into a `Date` extension.

Issue #12750, opened 2019-10-22
Crash: Opening post settings after uploading a featured image from another client


rachelmcr:
### Expected behavior

I expect always to be able to open the post settings for a post.

### Actual behavior

If I add a featured image from a different client (web or another app), when I try to open the post settings for that post in the app it crashes.

The crash does _not_ happen if I first go to the Media section in the app and ensure the featured image is already synced there.

### Steps to reproduce the behavior

1. Create a new post in the app and save it as a draft.
2. Open the same post in the web editor.
3. Upload a featured image to the post and save it.
4. In the app, make sure the remote changes are synced (e.g. pull to refresh the post list) and then open the post again.
5. Open the post settings. Result: Crash.
6. Open the Media section and wait for the media to sync.
7. Open the post again, and open the post settings. Result: The post settings load.

##### Original report on iPhone SE, iOS 12.3.1, WPiOS 13.3, internal ref: 2435883-zen
h/t @melindahelt for raising the issue!

This crash also seems to generate numerous Sentry crash reports (one per crash, rather than grouping them) such as these:
https://sentry.io/share/issue/9b5b12566ad8477ea36c12374c310771/
https://sentry.io/share/issue/2cf5061c2c624562b598342a8186b585/

rachelmcr:
@designsimply I think this should probably be prioritized for a fix. It's hard to tell the exact frequency because Sentry isn't grouping these crashes, but searching Sentry for `featuredImageURLForDisplay` crashes returns 1000+ results.

shiki:
We confirmed that this was caused by https://github.com/wordpress-mobile/WordPress-iOS/pull/12371.

designsimply:
Quick note to say a 2-star review came in about this issue "Lacking a lot" by erikbrig – Oct 21, 2019 Version 13.3

> App crashes when pressing “post settings”, unable to format quotes, and unable to add description to featured image for photographers credit.

designsimply:
From a 1-star review "Click post-settings = CRASH" by Newer Reviewer Tour – Edited Oct 20, 2019 Version 13.3

> The app crashes and just vanishes if you click post-settings on a WordPress.org site you are trying to edit.

PR #12754, opened 2019-10-22
Fix post settings crash and ensure local image paths are always correct

leandroalonso:
Addendum to #12750

To test:

1. Create a new post in the app and save it as a draft.
2. Open the same post in the web editor.
3. Upload a featured image to the post and save it.
4. In the app, make sure the remote changes are synced (e.g. pull to refresh the post list) and then open the post again.
5. Open the post settings. The image should be loaded.

Additional info:

The featured image in pot settings never shows a local image (only an uploaded one). However, we decided to keep the code consistent and `featuredImageURLForDisplay` will always return the correct path.

Update release notes:


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,10 @@
 13.5
 -----
- 
+
+13.4.1
+-----
+* Post Settings with a featured image will not crash the app.
```
yaelirub: Maybe something more like: "Post Settings: Fixed a crash with featured image."

PR #12757, opened 2019-10-23
Replace GiphyCoreSDK with Giphy pod

etoledom:
GiphyCoreSDK has been deprecated (and deleted 😱), so we need to implement the new `Giphy` pod instead.

Sadly this pod also installs the Giphy UI library that we don't use.

This error started to appear in some [CI checks](https://circleci.com/gh/wordpress-mobile/WordPress-iOS/15175?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) already.
It should be able to reproduce the error by deleting the local `Pods/` folder and running `rake dependencies`.

To test:
- `rm -rf Pods/`
- `rake dependency`
- Build and Run.
- Check that you can download/add gifs to the Media Library or (Aztec) posts.

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -148,7 +148,7 @@ target 'WordPress' do
     pod '1PasswordExtension', '1.8.5'
     pod 'Charts', '~> 3.2.2'
     pod 'Gifu', '3.2.0'
-    pod 'GiphyCoreSDK', '~> 1.4.0'
+    pod 'Giphy', '1.1.3'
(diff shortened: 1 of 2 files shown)


hypest:
> and deleted 😱

Should I assume that now, trying to build an older hash of the WPiOS repo will fail since the GiphyCoreSDK repo is not there anymore?

etoledom:
Thank you all!

> Should I assume that now, trying to build an older hash of the WPiOS repo will fail since the GiphyCoreSDK repo is not there anymore?

Yes, looks like it.

Issue #12147, opened 2019-07-17
[Stats Refresh] DWMY: Load overview block independently


danielebogo:
On DWNY we should load the Chart block first, then all the other blocks

PR #12764, opened 2019-10-23
[Stats Refresh] DWMY Async loading operations and feature flag

danielebogo:
Refs. #12147 
Refs. #11852

This is the first step to allow DWMY to load the blocks independently adding a new feature flag `statsAsyncLoadingDWMY` and the `AsyncOperation` class.

## To test:
- [ ] Run `AsyncOperationTests`
- [ ] Run `StatsPeriodAsyncOperationTests`
- [ ] Run the branch:
1. Open Stats -> DWMY from a Site you never opened before (or from a fresh install)
2. Yous shouldn't see the NRV
3. You should be able to load the chart and Top Posts card
4. Disable the feature flag and make sure everything works as before

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
@@ -194,8 +194,6 @@ private extension SiteStatsPeriodTableViewController {
             return
         }
 
-        DDLogInfo("--- Refresh Table View")
-
         tableHandler.viewModel = viewModel.tableViewModel()
 
         if asyncLoadingActivated {
--- a/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
@@ -46,7 +46,9 @@ class SiteStatsPeriodTableViewController: UITableViewController, StoryboardLoada
                 refreshData()
             }
 
-            displayLoadingViewIfNecessary()
+            if !asyncLoadingActivated {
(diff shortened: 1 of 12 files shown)


ScoutHarris:
@danielebogo I assume the `Not Ready for Review` label is no longer applicable?

danielebogo:
@ScoutHarris yes! Just removed it!

ScoutHarris:
Hey @danielebogo .

Since this is a WIP, I assume views like this are expected? Where, while loading, the date bar is ghosted but the other stats either don't show at all or show old data?

|  |  |
|--------|-------|
| loading_init | loading_again |

WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
```diff
@@ -183,7 +194,17 @@ private extension SiteStatsPeriodTableViewController {
             return
         }
 
+        DDLogInfo("--- Refresh Table View")
```
ScoutHarris: Is this necessary? It's creating a bit of log noise, and doesn't really provide any information...

danielebogo: Yes I need it to actually verifying the table is reloaded immediately after an operation is executed.
I'm going to remove it here since I have it in another branch 👍

danielebogo:
Thanks @ScoutHarris for reviewing this.
> Since this is a WIP, I assume views like this are expected? Where, while loading, the date bar is ghosted but the other stats either don't show at all or show old data?

Yes exactly, this PR doesn't contains any placeholder view yet but only the new operations implemented for the chart and top posts only. So you should see only those cards populated in the list.

danielebogo:
Thanks @ScoutHarris @frosty

PR #12768, opened 2019-10-24
Support Landscape for Post Scheduling Calendar Picker

bjtitus:
Ref #12686

@ScoutHarris @osullivanchris Please take a look and provide any feedback. Wasn't too bad to squeeze this stuff in the compact height, but the calendar spacing is quite large.

To test:

- Navigate to the `Published on` / `Scheduled for` setting on a blog post.
- Tap on the `Published on` / `Scheduled for` cell
- Open the new calendar picker by tapping on the `Date and Time` cell
- Rotate to landscape orientation

Simulator Screen Shot - iPhone Xʀ - 2019-10-23 at 22 07 06

The downside to this layout is that their is no way to dismiss the calendar without changing the date or continuing through the process to keep the current date.


osullivanchris:
Hey @bjtitus thanks for the update. 

> Wasn't too bad to squeeze this stuff in the compact height, but the calendar spacing is quite large.

I'm not sure if I have access to the build to test. Is the screenshot the entire device? Is it possible to see a screenshot with any chrome that is visible around it (e.g. status bar)? Generally height wise looks ok.

Agree on the calendar spacing. I had a quick look at it. Here or two images, a quick mock and one with red-lines to help describe. One option would be just to add padding to to the left and right of the calendar area (in red). I've then just distributed the date rows within that new width. If we did that we would also need to add padding to the change month chevrons to keep the alignment I think (blue lines).

However given that it looks ok, I'd probably avoid adding this conditional logic and keep it the way it is. The only caveat is, how will it look on iPad? Will the width be even more exaggerated? Then we might need a sensible limit with padding as described.

cal_with_padding

cal_with_padding_redlines


> The downside to this layout is that their is no way to dismiss the calendar without changing the date or continuing through the process to keep the current date.

Interesting. I think its probably ok. "Continuing the process" really just means re-confirming what you already had, and the only additional step is bringing up the time, so its not too heavy for the user. Just to confirm, it retains the information from the previous time I selected. For example:
- The date is set to publish immediately
- I decide to select a scheduling time, choosing Friday 25th October at 3:00pm UTC.
- I confirm
- I open the calendar again
- Friday 25th October is still selected (can you confirm)
- I tap next
- 3:00pm UTC is still selected (can you confirm)
if it works like that, it shouldn't break the flow.

The only other idea I have is that "Publish Immediately" kind of acts like a 'back' button initially before you select a date. But once you've selected a date/time, it still says "Publish Immediately". We could change it to "Cancel". But again, maybe this is intended to be a shortcut to revert to publish immediately, rather than reverting. And additional logic doesn't seem like good value here.

ScoutHarris:
Hey @bjtitus . This should be branched off of `feature/12686-post_scheduling_calendar_picker`, which means the target branch should be that as well. Do you mind changing it so the PR changes are accurate (and don't contain _everything_ you've done)? Thank you!

bjtitus:
@osullivanchris 

No status bar or anything else appear. Here's a screenshot of the full device with chrome:

Screen Shot 2019-10-25 at 7 35 29 AM

This is similar to the post view where nothing else is visible along the top:

Screen Shot 2019-10-25 at 7 35 17 AM

I think I'm going to change to a popover on iPad to avoid the sizing issues if that's alright. It sounded like that's what you were envisioning anyway.

osullivanchris:
> No status bar or anything else appear. Here's a screenshot of the full device with chrome:

ah thanks for the screenshot. Looks good. If it rendered slightly differently would it be scrollable, if the user can't tap the last row?

> I think I'm going to change to a popover on iPad to avoid the sizing issues if that's alright. It sounded like that's what you were envisioning anyway.

Sure. If you can just check and see are we doing that for similar interactions, unfortunately I don't have an iPad to hand to test. And send a screenshot if you want me to review.

bjtitus:
@osullivanchris Looking at the rest of the app it seems like Form Sheets are used more often than Popovers. However, iOS does use popovers in similar situations (editing a Contact and adding a Related Date at the bottom). I think a Form Sheet might look a little better here but let me now what you think!

Note that the popover can be presented from any position along the row also, if you'd rather it point to the date label.

Napkin 2 10-27-19, 10 00 58 PM

ScoutHarris:
> The downside to this layout is that their is no way to dismiss the calendar without changing the date or continuing through the process to keep the current date.

My 2 cents - I think allowing calendar dismissal in landscape would be preferable.

> This is similar to the post view where nothing else is visible along the top

Well, that's not entirely accurate. The bar at the top does allow you to cancel editing (among other things).

@bjtitus @osullivanchris Can we adjust the calendar to allow dismissing it? Just as a quick hack, I moved the calendar down a bit to allow 1) the top bar to appear, and 2) dismiss by tapping. I didn't adjust the spacing accordingly, so it doesn't look quite right. But can we adjust the calendar in landscape to allow for this?

suggestion

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarMonthView.swift
```diff
@@ -73,9 +72,23 @@ class CalendarMonthView: UIView {
         ])
 
         addSubview(stackView)
+
+        stackView.addConstraints([
+            collectionView.heightAnchor.constraint(equalToConstant: 240),
+            collectionView.widthAnchor.constraint(equalToConstant: 375)
```
ScoutHarris: We try not to just have hard-coded values. We tend to use either a `struct` or `enum` to give constants indicative names. If you look for examples of `struct Constants` or `enum Constants`, that should give you an idea.

bjtitus:
@osullivanchris Given @ScoutHarris's feedback about being unable to dismiss in landscape on iPhone I was wondering if a close button might be added just for that case. This would end up looking like the Post screen; check out the screenshot below:

Screen Shot 2019-10-28 at 9 07 47 PM

I was only planning to include it in landscape on iPhones, but could also add to iPads if you would rather use the Form Sheet than Popover. I think the spacing gets extremely tight for the touch targets if extra spacing is added in iPhone landscape as shown in the comment above.

osullivanchris:
Hi just catching up after being away for a couple days

**Dismissing on landscape**
Reading the thread we have two options proposed

- Move the calendar down and use the skrim/overlay as a tapable area to dismiss. I feel like this is both removing real estate for the calendar UI, and its also a small awkward area to tap to dismiss
- Putting an 'x' in the header on the left like in posts. I don't think it works in this context because it is close to the 'Publish immediately' option, so it looks like they are the same action.

When I'm looking at sheets generally this is a good stress test to note - because my inclination would have been to use the overlay area as a tap to close action for bottom sheets and not worry about an explicit action. But on landscape it does feel like we need an explicit action to close.

To be honest I think what is causing the trouble is having the 'Publish immediately' action within the calendar UI. I think a separate step in the flow for this would potentially make it clearer. But I don't know how big a re-work it is. 

Here is a sketch of what I'm talking about. I am open to what UI we use in the new step 2. Maybe an action sheet is too heavy. But if we separate the decision to publish now / schedule, the step with the calendar UI doesn't have to work so hard, and it frees up space to put a simple 'cancel' action on the left of the sheet. 

IMG_4453

Its late here so I can look into it more Wednesday. As I share this, I'm realising the downside of this solution is that when it changes to 'scheduled', the step 2 UI might not really work. Will share the idea anyway while I'm here.

The other thing is maybe we could put the 'publish immediately' action somewhere else. To me it looks like a back action, which it is not if the post is currently scheduled. I will follow up Wednesday when I have time to think about it more.

> I think popover would be better here. Right now there isn't much difference, but when notifications and calendar options are added, I think popover will better indicate which one is being modified. I like where you showed it - anchored to the 'Date and Time' row.

I'm happy to go with this. The logic sounds good to me and @ScoutHarris is closer to it as I haven't been working on iPad.

bjtitus:
@ScoutHarris Check out my latest set of changes and let me know if there's anything else I can improve. I don't love the solution to allow the calendar view to break its height but I was mainly seeing unsatisfiable constraint warnings on rotation after fixing some of the smaller ambiguous warnings.

The remaining warnings seem to be in WPSplitViewController or possibly other code as the memory addresses don't match up to the views in the presented view controller.

ScoutHarris:
Hey @osullivanchris .

> The other thing is maybe we could put the 'publish immediately' action somewhere else. To me it looks like a back action, which it is not if the post is currently scheduled. I will follow up Wednesday when I have time to think about it more.

Oh I like that idea!

I was wondering if we could compromise for the time being? Since you're reconsidering the flow I was thinking maybe we could add the 'X' right now (it'd just be for iPhone landscape).  Or if you _really_ don't like the 'X', we could leave the calendar un-dismissable in iPhone landscape.

In either case, we could finish this out, and once you work out the flow, we can create another issue for that and address it in the future?

ScoutHarris:
Hey @bjtitus .

This is great! Thanks for all the changes!

The only other thing is it would be nice to have constants for these values, since it's not really obvious what's going on there. Sorry, I think I missed that from your first PR.

```
func configureCalendar(_ calendar: JTACMonthView) -> ConfigurationParameters {
    let startDate = Date(timeIntervalSinceReferenceDate: (-24*60*60)*365*50)
    let endDate = Date(timeIntervalSinceReferenceDate: (24*60*60)*365*50)
    return ConfigurationParameters(startDate: startDate, endDate: endDate)
}
```

Otherwise, the only item remaining is @osullivanchris 's preference about iPhone landscape.

osullivanchris:
> I was wondering if we could compromise for the time being? Since you're reconsidering the flow I was thinking maybe we could add the 'X' right now (it'd just be for iPhone landscape). Or if you really don't like the 'X', we could leave the calendar un-dismissable in iPhone landscape.

Agree! Good compromise. I haven't been able to prioritise it to do it the due dilligence, and this unblocks us. If it only appears on landscape that's great. I think there is space there to separate the 'x' and 'publish immediately' actions, whereas on portrait they would run into each other even more. 

> In either case, we could finish this out, and once you work out the flow, we can create another issue for that and address it in the future?

definitely. I'll add this to my groundskeeping list or something like that. Thanks for unblocking.

bjtitus:
@ScoutHarris Added those to an extension on Date and switched to using `DateComponents` for the other calculation that had time interval constants.

Also updated with code to add the close button in landscape on iPhone. It looks like the screenshot from above but I can add spacer bar button items if we need to.

WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
```diff
@@ -174,7 +174,6 @@ private struct DateAndTimeRow: ImmuTableRow {
             }
 
             if self?.viewController.traitCollection.userInterfaceIdiom == .pad {
-//                navigationController.modalPresentationStyle = .formSheet
```
ScoutHarris: 👍

WordPress/Classes/ViewRelated/Post/Scheduling/SchedulingCalendarViewController.swift
```diff
@@ -71,7 +69,26 @@ class SchedulingCalendarViewController: DatePickerSheet, DateCoordinatorHandler
         (segue.destination as? DateCoordinatorHandler)?.coordinator = coordinator
     }
 
-    @IBAction func publishImmediately() {
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        let closeButton = UIBarButtonItem(image: Gridicon.iconOfType(.cross), style: .plain, target: self, action: #selector(closeButtonPressed))
```
ScoutHarris: Hey you found Gridicons! 👍 ! 😄

PR #12772, opened 2019-10-24
Do full link selection fix only on 13.0

### LINKED ISSUE #12626: NSRangeException: Aztec.TextStorage attributedSubstringFromRange:: Out of bounds
Sentry Issue: [WORDPRESS-IOS-1W6P](https://sentry.io/share/issue/94e42b8dcb934785a6c2d37238fcb1d2/)

```
NSRangeException: NSMutableRLEArray replaceObjectsInRange:withObject:length:: Out of bounds
  ?, in <redacted>
...
(34 additional frame(s) were not displayed)
```

This one seems like it could be due to the device just not having enough memory🤔

SergioEstevao:
Fixes #12626

To test:
 - Check the scenario referred in the ticket to see if the bug still happens

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -1350,14 +1350,18 @@ extension AztecPostViewController: UITextViewDelegate {
     }
 
     func textView(_ textView: UITextView, shouldInteractWith URL: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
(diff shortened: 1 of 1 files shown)


designsimply:
I tested with `develop` @ `625c1df8d0` on an iPhone 6S iOS 13.1.3 and couldn't get the app to crash after trying to add, remove, move links including dragging and dropping text with links or just links in the Aztec editor.

koke:
Sorry I got to this one late, but I'm still not sure about this fix. The original crash was also happening on 13.0, so this won't fix the crash for those users. I still think it's better to avoid the crash (which could lead to content loss in an editor context) rather than avoid a link UX bug.

PR #12791, opened 2019-10-28
[Stats Refresh] DWMY Async loading and Ghost views

danielebogo:
Refs. #12147 

This PR continues the Load Blocks Independently started [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/12764), implementing a new operation, new operations in the queue and ghost views. It also refactored the method used to handling the blocks in Insights.

To keep the PR easy to review and because of this refactor, I implemented the operations and the ghost views for these blocks only:
• _summary_
• _topPostsAndPages_
• _topReferrers_
• _topPublished_
• _topClicks_

_topPublished_ uses a different API. For this reason there's a specific operation for that case: `StatsPublishedPostsAsyncOperation`

### Refactor
The method implemented by the Insights and Period view models is now a protocol called `AsyncBlocksLoadable`. It uses a Generic Value as return type, which means it can be anything (in our case a single or a collection of ImmutableRow).

## To test:
- Open Stats -> DWMY from a Site you never opened before (or from a fresh install)
- Yous shouldn't see the NRV
- You should be able to load the chart and Top Posts card
- Disable the feature flag and make sure everything works as before

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/SiteStatsViewModel+AsyncBlock.swift
+++ b/WordPress/Classes/ViewRelated/Stats/SiteStatsViewModel+AsyncBlock.swift
@@ -35,7 +35,7 @@ extension AsyncBlocksLoadable where CurrentStore: StatsStoreCacheable, RowType =
                        error: AsyncBlock<Value>) -> Value {
         let featureFlag: FeatureFlag = type == .insights ? .statsAsyncLoading : .statsAsyncLoadingDWMY
         let isFeatureEnabled = Feature.enabled(featureFlag)
-        let containsCachedData = checkingCache != nil ? checkingCache!() : currentStore.containsCachedData(for: blockType)
+        let containsCachedData = checkingCache?() ?? currentStore.containsCachedData(for: blockType)
 
         if containsCachedData || !isFeatureEnabled {
             return block()
--- a/WordPress/Classes/ViewRelated/Stats/SiteStatsViewModel+AsyncBlock.swift
+++ b/WordPress/Classes/ViewRelated/Stats/SiteStatsViewModel+AsyncBlock.swift
@@ -1,28 +1,43 @@
 protocol AsyncBlocksLoadable {
-    typealias CompletionBlock<Value> = () -> Value
+    typealias AsyncBlock<Value> = () -> Value
+    typealias CacheBlock = () -> Bool
 
     associatedtype CurrentStore
     associatedtype RowType
 
+    /// The current Store used
     var currentStore: CurrentStore { get }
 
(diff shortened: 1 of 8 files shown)


ScoutHarris:
Hey @danielebogo . I'm unable to run your branch. I keep getting this error:
`[!] Error installing GiphyCoreSDK`

I think it just needs updating with `develop`.

danielebogo:
Thanks @ScoutHarris ! I merged develop and also fixed the conflict! It should be ok now!

WordPress/Classes/ViewRelated/Stats/SiteStatsViewModel+AsyncBlock.swift
```diff
@@ -0,0 +1,53 @@
...
+
+extension AsyncBlocksLoadable where CurrentStore: StatsStoreCacheable, RowType == CurrentStore.StatsStoreType {
+    func blocks<Value>(for blockType: RowType,
+                       type: StatType,
+                       status: StoreFetchingStatus,
+                       checkingCache: CacheBlock? = nil,
+                       block: AsyncBlock<Value>,
+                       loading: AsyncBlock<Value>,
+                       error: AsyncBlock<Value>) -> Value {
+        let featureFlag: FeatureFlag = type == .insights ? .statsAsyncLoading : .statsAsyncLoadingDWMY
```
ScoutHarris: Since `statsAsyncLoading` will be removed soon, and is currently set to `true`, can we make this check _not_ dependent on that flag?

ScoutHarris: I failed to elaborate - I'm going to remove `statsAsyncLoading` this week, so if this logic already worked without it, that'd be super helpful!

danielebogo:
@ScoutHarris Just updated it! Would you mind to have another look?

ScoutHarris:
Hey @danielebogo . Thank you for adding the headers & lines. It looks much better! Just one minor quibble - the lines when loading are not the same height as the regular lines.

loading

SylvesterWilmott:
>Thank you for adding the headers & lines. It looks much better! Just one minor quibble - the lines when loading are not the same height as the regular lines.

The lines heights appear to be correct but the ghost animation lightens them giving that appearance

danielebogo:
Thanks @SylvesterWilmott ! @ScoutHarris @frosty I changed the UIView extension to be able to remove the ghost animation! Would you mind to have a final look?
Simulator Screen Shot - iPhone 11 - 2019-10-30 at 10 14 07

danielebogo:
Thanks everyone!

PR #12792, opened 2019-10-28
Implementing NSBatchUpdateRequest in PostService+RefreshStatus

### LINKED ISSUE #12573: Consider using NSBatchUpdateRequest in refreshPostStatus
The suggestion was brought up by @koke in this comment:
https://github.com/wordpress-mobile/WordPress-iOS/issues/12420#issuecomment-535341929

[Explanation on NSBatchUpdateRequest](https://www.bignerdranch.com/blog/new-in-core-data-and-ios-8-batch-updating/)

majobujnak:
Fixes #12573 

Updated the referenced method to use `NSBatchUpdateRequest` instead of `NSFetchRequest`. Since there isn't an option to use any logic during the update and the posts were being previously updated using logic in the `markAsFailedAndDraftIfNeeded` method, I've decided to implement this logic using predicates, which could be done pretty simply and neatly (`notFailedPredicate`).

To test:
(Taken from https://github.com/wordpress-mobile/WordPress-iOS/pull/12567)
1. Be offline
2. Create a post
3. Publish post
4. See the message: "Post will be published..."
5. Quit the app
6. return to the app turn online
7. See that the post is being auto uploaded

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/PostService+RefreshStatus.swift
+++ b/WordPress/Classes/Services/PostService+RefreshStatus.swift
@@ -14,12 +14,14 @@ extension PostService {
             let processingPredicate = NSPredicate(format: "remoteStatusNumber = %@", NSNumber(value: AbstractPostRemoteStatus.pushingMedia.rawValue))
             let pushingOrProcessingPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: [pushingPredicate, processingPredicate])
             let notFailedPredicate = NSPredicate(format: "remoteStatusNumber != %@ AND NOT (postID != nil AND postID > 0)", NSNumber(value: AbstractPostRemoteStatus.failed.rawValue))
-        
+            
             request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [pushingOrProcessingPredicate, notFailedPredicate])
-            request.propertiesToUpdate = ["remoteStatusNumber": NSNumber(value: AbstractPostRemoteStatus.failed.rawValue),
+            request.propertiesToUpdate = [
+                "remoteStatusNumber": NSNumber(value: AbstractPostRemoteStatus.failed.rawValue),
                 "status": NSString(string: BasePost.Status.draft.rawValue),
-                "dateModified": NSDate()]
-
+                "dateModified": NSDate()
+            ]
+            
             do {
                 try self.managedObjectContext.execute(request)
 
--- a/WordPress/Classes/Services/PostService+RefreshStatus.swift
+++ b/WordPress/Classes/Services/PostService+RefreshStatus.swift
@@ -9,22 +9,25 @@ extension PostService {
     ///
     func refreshPostStatus(onCompletion: (() -> Void)? = nil, onError: ((Error) -> Void)? = nil) {
         self.managedObjectContext.perform {
-            let fetch = NSFetchRequest<Post>(entityName: Post.classNameWithoutNamespaces())
+            let request = NSBatchUpdateRequest(entityName: Post.classNameWithoutNamespaces())
             let pushingPredicate = NSPredicate(format: "remoteStatusNumber = %@", NSNumber(value: AbstractPostRemoteStatus.pushing.rawValue))
             let processingPredicate = NSPredicate(format: "remoteStatusNumber = %@", NSNumber(value: AbstractPostRemoteStatus.pushingMedia.rawValue))
-            fetch.predicate = NSCompoundPredicate(orPredicateWithSubpredicates: [pushingPredicate, processingPredicate])
+            let pushingOrProcessingPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: [pushingPredicate, processingPredicate])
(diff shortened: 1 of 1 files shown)


WordPress/Classes/Services/PostService+RefreshStatus.swift
```diff
@@ -9,22 +9,24 @@ extension PostService {
     ///
     func refreshPostStatus(onCompletion: (() -> Void)? = nil, onError: ((Error) -> Void)? = nil) {
         self.managedObjectContext.perform {
-            let fetch = NSFetchRequest<Post>(entityName: Post.classNameWithoutNamespaces())
+            let request = NSBatchUpdateRequest(entityName: Post.classNameWithoutNamespaces())
             let pushingPredicate = NSPredicate(format: "remoteStatusNumber = %@", NSNumber(value: AbstractPostRemoteStatus.pushing.rawValue))
             let processingPredicate = NSPredicate(format: "remoteStatusNumber = %@", NSNumber(value: AbstractPostRemoteStatus.pushingMedia.rawValue))
-            fetch.predicate = NSCompoundPredicate(orPredicateWithSubpredicates: [pushingPredicate, processingPredicate])
+            let pushingOrProcessingPredicate = NSCompoundPredicate(orPredicateWithSubpredicates: [pushingPredicate, processingPredicate])
+            let notFailedPredicate = NSPredicate(format: "remoteStatusNumber != %@ AND NOT (postID != nil AND postID > 0)", NSNumber(value: AbstractPostRemoteStatus.failed.rawValue))
```
yaelirub: Why are we checking for NOT (postID != nil AND postID > 0) here ?

WordPress/Classes/Services/PostService+RefreshStatus.swift
```diff
@@ -9,22 +9,24 @@ extension PostService {
...
+            request.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [pushingOrProcessingPredicate, notFailedPredicate])
+            request.propertiesToUpdate = [
+                "remoteStatusNumber": NSNumber(value: AbstractPostRemoteStatus.failed.rawValue),
+                "status": NSString(string: BasePost.Status.draft.rawValue),
```
yaelirub: Why are we updating to draft?

WordPress/Classes/Services/PostService+RefreshStatus.swift
```diff
@@ -9,22 +9,24 @@ extension PostService {
...
+                "dateModified": NSDate()
+            ]
             do {
-                let postsPushing = try self.managedObjectContext.fetch(fetch)
-                for post in postsPushing {
-                    self.markAsFailedAndDraftIfNeeded(post: post)
```
yaelirub: Is there nothing in the logic of this method that needs to be handled in a refresh state?

WordPress/Classes/Services/PostService+RefreshStatus.swift
majo-bujnak: this is a representation of the `post.hasRemote()` logic from `markAsFailedAndDraftIfNeeded` method

WordPress/Classes/Services/PostService+RefreshStatus.swift
majo-bujnak: in the `markAsFailedAndDraftIfNeeded` method, if the post does not have remote, its `status` is set to draft and `dateModified` to new date

WordPress/Classes/Services/PostService+RefreshStatus.swift
majo-bujnak: the logic from this method contained a guard for a post not being failed and not having a remote, which I've represented as `notFailedPredicate` on line 16

WordPress/Classes/Services/PostService+RefreshStatus.swift
yaelirub: That happens only if the post is a page

majo-bujnak: I now see that there is a different implementation of the `markAsFailedAndDraftIfNeeded` method. this is the implementation I have seen when I've submitted this commit:

markAsFailedAndDraftIfNeeded

[According to Apple](https://developer.apple.com/library/archive/featuredarticles/CoreData_Batch_Guide/BatchUpdates/BatchUpdates.html) however, NSBatchUpdateRequest can not contain any logic beside the predicate, what this means I think is that the posts would have to be updated in two batches now -- one batch for the posts that are not failed (and set them to failed) and the other posts that do not have a remote and are a Page. would that be alright?

majo-bujnak: hey @yaelirub, I've changed the current implementation so that it conforms to the `markAsFailedAndDraftIfNeeded()` method. it has to be done in two batches, which is probably still faster than the fetch. 

however, I've uncovered an unwanted behavior (at least I assume it's unwanted). the fetched entities were of type `Post: AbstractPost` and these entities were then checked whether they're of type `Page: AbstractPost`, which would never result in true, since pages would not be fetched at all. I've tested this. so I've changed the fetched entity to `AbstractPost`. please let me know if I should revert it back to `Post` in case I'm missing something.

PR #12798, opened 2019-10-29
Aztec: `shouldInteractWithURL` will return always false to avoid crashes

etoledom:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/12730

Following up from https://github.com/wordpress-mobile/WordPress-iOS/pull/12772.
I agree with @koke on https://github.com/wordpress-mobile/WordPress-iOS/pull/12772#issuecomment-547299425.

Since iOS 13.0 is so low on adoption, I believe is better to keep the simpler fix, regardless of the UX issue on 13.0, avoiding possible crashes.

I will do the gutenberg side changes next, but will probably go live on WPiOS 13.6.

To test:

Update release notes:


etoledom:
Thank you @SergioEstevao !
I have moved this PR to target `develop`. Since on `13.5` there's already a merge that will fix most of the crashes.

PR #12812, opened 2019-10-30
posts of self-hosted sites are not auto- uploaded if canceled

### LINKED ISSUE #12801: [Offline Posting] Drafts of self-hosted sites are always uploaded
## Expected behavior

The draft (with canceled intent) is not uploaded.

## Actual behavior

The draft (with canceled intent) is uploaded.

1. While offline, edit an existing draft post. 
3. Tap on X → Update Draft
4. In the Post List, tap _Cancel_. 
5. Go online.

The draft post is still getting uploaded. 

## Notes

This is because we call `PostCoordinator.autoSave()` but our `autoSave()` method also _uploads_ the post. But I think it's best if we can solve this within `PostAutoUploadInteractor`.

yaelirub:
Fixes #12801 

To test:

1. While offline, edit an existing draft post.
2. Tap on X → Update Draft
3. In the Post List, tap Cancel.
4. Go online.
5. See that the post is *not* auto-uploaded  
Update release notes:

### DIFF
--- a/WordPress/Classes/Services/PostAutoUploadInteractor.swift
+++ b/WordPress/Classes/Services/PostAutoUploadInteractor.swift
@@ -44,7 +44,7 @@ final class PostAutoUploadInteractor {
             return .upload
         } else {
             // because autosave call will end up uplaoding a post on self hosted we don't autosave in this case
-            return post.isHostedAtWPcom ? .autoSave : .nothing
+            return post.supportsWPComAPI ? .autoSave : .nothing
         }
     }
 
--- a/WordPress/Classes/Services/PostAutoUploadInteractor.swift
+++ b/WordPress/Classes/Services/PostAutoUploadInteractor.swift
@@ -43,7 +43,8 @@ final class PostAutoUploadInteractor {
         if post.isLocalDraft || post.shouldAttemptAutoUpload {
             return .upload
         } else {
-            return .autoSave
+            // because autosave call will end up uplaoding a post on self hosted we return nothing
+            return post.isHostedAtWPcom ? .autoSave : .nothing
         }
     }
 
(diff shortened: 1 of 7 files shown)


WordPress/Classes/Services/PostAutoUploadInteractor.swift
```diff
@@ -43,7 +43,8 @@ final class PostAutoUploadInteractor {
         if post.isLocalDraft || post.shouldAttemptAutoUpload {
             return .upload
         } else {
-            return .autoSave
+            // because autosave call will end up uplaoding a post on self hosted we don't autosave in this case
+            return post.isHostedAtWPcom ? .autoSave : .nothing
```
leandroalonso: `post.isHostedAtWPcom` will return `false` for AT sites.
In `develop` I'm able to `autoSave` content for an AT site. However, after this change, this doesn't happen anymore.

I did some investigation here and this seems to solve the issue:

```swift
return post.blog.supports(.wpComRESTAPI) ? .autoSave : .nothing
```

Would you mind to update the code?

The steps to reproduce the issue:

1. Use an AT site
2. While offline, edit a draft and then tap cancel
3. Go back online
4. The post is auto-saved
5. If I go to calypso I see an "A more recent revision of this post exists. Restore?" message

yaelirub: Excellent! Updating

yaelirub:
@leandroalonso , updated and ready for another look

yaelirub:
Ugh. Fixing the tests here seems to be tricky. In order to create an account for blog we need to have post, blog and account on the main context and  even when doing that the tests keep failing with" ```-[WordPressTest.PostCoordinatorFailedPostsFetcherTests testItReturnsPostsThatCanBeAutoUploadedOrAutoSaved] : failed: caught "Unresolved Core Data save error (Main Context)", "Non-optional property with a nil value on Post.blog"```

I'll have to figure out how to mock and update

leandroalonso:
@yaelirub I've refactored the tests to support setting the `supportsWPComAPI`.

Let me know what you think. :)

PR #12834, opened 2019-11-01
Try Gutenberg: Image fullscreen preview 

cameronvoell:
Implements https://github.com/wordpress-mobile/gutenberg-mobile/issues/1286

**PR is for testing a fullscreen preview, does not include gb-mobile release 1.16.0 

Gb-mobile - https://github.com/wordpress-mobile/gutenberg-mobile/pull/1527
Gutenberg - https://github.com/WordPress/gutenberg/pull/17109

To test:

1. Open article in the block editor.
2. Navigate to an image block with an image loaded, or add a new image block and add an image
3. Select the image block and notice the blue border shows around the image
4. Press the image while the image block is selected and notice the image will be viewed in fullscreen.
5. Press outside the boundaries of the image in fullscreen mode, or swipe the image down to exit the fullscreen image preview.

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -140,7 +140,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => '197cc2a7d7d6c1fa80fa7e3bc8eea37baa936b12'
+    gutenberg :commit => '0adfa6c3ea8d50a54719dfe60ec296f668393db8'
 
     ## Third party libraries
     ## =====================
--- a/Podfile
+++ b/Podfile
@@ -140,7 +140,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => 'e776a6c5dab552d0781f80e8c1298cd1c4786df2'
+    gutenberg :commit => '197cc2a7d7d6c1fa80fa7e3bc8eea37baa936b12'
 
     ## Third party libraries
     ## =====================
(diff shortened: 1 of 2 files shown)


iamthomasbishop:
Looking good @cameronvoell, nice work! I just have a couple bits of feedback:

* Sometimes the transition from the canvas to the image preview is abrupt, not using a cross-fade. It seems this usually happens on the first attempt after adding an image to canvas.
* Can we dismiss the preview on tap? If we were showing information/UI around the image then I'd propose the tap gesture would toggle visibility of that, but since we're not showing anything let's have tap dismiss the preview UI.
* It'd be great to allow omnidirectional drag to dismiss, but I don't think that's a blocker to ship v1.

cameronvoell:
> Sometimes the transition from the canvas to the image preview is abrupt, not using a cross-fade. It seems this usually happens on the first attempt after adding an image to canvas.

I'm still working on this. The first couple parameter tweaks etc. did not seem to help. See the video below. I agree it is not a good first animation.
first-transition-pops

iamthomasbishop:
@cameronvoell One tiny other thing I forgot to mention, is there a way we can remove the image outline (thick blue border) while the upload is progressing? I think the outline should only be applied once the image is successfully uploaded.

iamthomasbishop:
I also noticed in your last video above that the progress bar jumps to the bottom of the area when, mostly as the upload is finishing.

Separate issue, which I [created here](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1544): The icon used on the placeholder background could use some tweaks. This is especially jarring because there's some jumpiness in displaying the image.

cameronvoell:
Closed in favor of native iOS solution: https://github.com/wordpress-mobile/WordPress-iOS/pull/12979

PR #12843, opened 2019-11-04
Creates Media Editor Wrapper

### LINKED ISSUE #12784: Create Media Editor wrapper
We want to create a Media Editor wrapper that initially will:

1. Wrap `TOCropViewController` to provide Basics features: cropping, rotation, zoom in/out
2. Receive a `PHAsset` and output a `UIImage`

### Additional information

* `TOCropViewController`: https://github.com/TimOliver/TOCropViewController
* We should build this wrapper with componentization in mind. Please, avoid using any WP class/entity/helper — we want to easily extract it in the future.

leandroalonso:
Fixes #12784

## Project Organization

I've added the Media Editor as a framework inside the WPiOS project. This is due to:

- We will extract this library to its own repository (we aim it to be reusable)
- We want to maintain the library decoupled from the WPiOS project
- Yet, we want to take advantage of all the CI/hooks/action infrastructure the project already has without the overload of configuring thinks right now

## See in action

Using the `MediaEditor` is simple:

```swift
let mediaEditor = MediaEditor()
mediaEditor.edit(image, from: self) { image in
  // do something
}
```

To see it in action, in `GutenbergViewController` replace the `gutenbergDidRequestMediaFromDevicePicker` func for this one (and add `import MediaEditor` on the top of the file):

```swift
    func gutenbergDidRequestMediaFromDevicePicker(filter: WPMediaType, allowMultipleSelection: Bool, with callback: @escaping MediaPickerDidPickMediaCallback) {
        mediaPickerHelper.presentMediaPickerFullScreen(animated: true,
                                                       filter: filter,
                                                       dataSourceType: .device,
                                                       allowMultipleSelection: allowMultipleSelection,
                                                       callback: {(assets) in
                                                        guard let phAssets = assets as? [PHAsset] else {
                                                            callback(nil)
                                                            return
                                                        }
                                                        let asset = phAssets.first!

                                                        PHImageManager.default().requestImage(for: asset, targetSize: asset.pixelSize(), contentMode: .default, options: PHImageRequestOptions()) { (image, info) in
                                                            guard let image = image else {
                                                                return
                                                            }

                                                            let mediaEditor = MediaEditor()
                                                            mediaEditor.edit(image, from: self.navigationController) { image in
                                                                
                                                            }
                                                        }
        })
    }
```

Note that this will only display the crop screen. Your cropped image will not be uploaded.

## Testing

Overall, I'm looking for feedback on the `MediaEditor` API. You can see it in action following the steps above but testing it as a feature is not necessary for this PR.

-=-=-=-


WordPress/MediaEditor/MediaEditor.swift
```diff
@@ -0,0 +1,34 @@
...
+    public init(cropViewControllerFactory: @escaping (UIImage) -> TOCropViewController = TOCropViewController.init) {
+        self.cropViewControllerFactory = cropViewControllerFactory
+        super.init()
+    }
+
+    public func edit(_ image: UIImage, from viewController: UIViewController? = nil, callback: @escaping (UIImage?) -> ()) {
+        self.callback = callback
+        cropViewController = self.cropViewControllerFactory(image)
+        cropViewController?.delegate = self
+        viewController?.present(cropViewController!, animated: true)
```
yaelirub: Can you guard instead of force unwrap?

yaelirub:
Hey @leandroalonso , Wonderful work with the library integration.
A few observations on the available features:
1. Two icons (options) for rotation seem redundant.  
2. Editing the ratio seem to work only on the original so if you crop and then change the ratio it doesn't work as expected 
3. After changing the ratio I have to tap on the ratio icon twice to get the ratio action sheet again.

leandroalonso:
@yaelirub 

> Two icons (options) for rotation seem redundant.

We have the option to hide one of them if we want to. Perhaps, having both buttons could save the user some tappings?

@osullivanchris what do you think? Should we keep both buttons or just one? Here's a screenshot with both buttons (clockwise and counter-clockwise) and just counter-clockwise:

Screen Shot 2019-11-06 at 09 50 55 Screen Shot 2019-11-06 at 09 50 11 

> Editing the ratio seem to work only on the original so if you crop and then change the ratio it doesn't work as expected

The ratio will always be applied accordingly to the cropped area. If you cropped the image and applied the ratio, this ratio will respect the area you want to crop.

The Photos app and WhatsApp work the same way.

> After changing the ratio I have to tap on the ratio icon twice to get the ratio action sheet again.

When you set the ratio the icons changes it's color to blue — that means you have a ratio applied. When you resize the crop area, this ratio will be kept. If you tap it again, the ratio will be disabled.

WhatsApp/Photos also has this same behavior.

WordPress/MediaEditor/MediaEditor.swift
leandroalonso: If you look at the line `18` I just instantiated the cropViewController and `self.cropViewControllerFactory(image)` doesn't return an `Optional`.

I can use `guard` or `if` but it's a little bit strange:

```swift
cropViewController = self.cropViewControllerFactory(image)
cropViewController?.delegate = self
if let cropViewController = self.cropViewController {
    viewController?.present(cropViewController, animated: true)
}
```

or:

```swift
cropViewController = self.cropViewControllerFactory(image)
cropViewController?.delegate = self

guard let cropViewController = self.cropViewController else {
    return
}
viewController?.present(cropViewController, animated: true)
```

or another option:

```swift
let cropViewController = self.cropViewControllerFactory(image)
cropViewController.delegate = self
viewController?.present(cropViewController, animated: true)
self.cropViewController = cropViewController
```

What do you think?

osullivanchris:
> @osullivanchris what do you think? Should we keep both buttons or just one? Here's a screenshot with both buttons (clockwise and counter-clockwise) and just counter-clockwise:

I would just have one button, its taking me longer to parse the one with two, plus squeezing in 6 actions is making the touch targets quite tight. Is it an option to only show the clockwise version? The reason I ask is that the anticlockwise rotate icon and the undo icon are quite similar.

Can we change the color of the text actions or the icons if we wanted to?

leandroalonso:
@osullivanchris 

> Is it an option to only show the clockwise version?

Yes!

Screen Shot 2019-11-06 at 11 11 49

> Can we change the color of the text actions or the icons if we wanted to?

Also, yes! :)

osullivanchris:
> Also, yes! :)

Nice. I also realised in our own icon set we actually have icons for undo, rotate and crop. Is it possible to change the icons? Is there any risk in doing so?

I think we can leave 'Done' as yellow like the native photos app, but can we change 'Cancel' to white? Its similar but different to our blue, so probably best to just go simple with white.

leandroalonso:
> Nice. I also realised in our own icon set we actually have icons for undo, rotate and crop. Is it possible to change the icons? Is there any risk in doing so?

We have access to the actual button so we can change the icons. :)

> I think we can leave 'Done' as yellow like the native photos app, but can we change 'Cancel' to white? Its similar but different to our blue, so probably best to just go simple with white.

Would you mind if I do those changes in another task?

WordPress/MediaEditor/MediaEditor.swift
yaelirub: LGTM

yaelirub:
@leandroalonso :

> The ratio will always be applied accordingly to the cropped area. If you cropped the image and applied the ratio, this ratio will respect the area you want to crop.

I checked again on the device and it worked as expected. On the simulator for some reason, the ratio was applied to the original not the cropped. 

>When you set the ratio the icons changes it's color to blue — that means you have a ratio applied. When you resize the crop area, this ratio will be kept. If you tap it again, the ratio will be disabled.

Yep. You're right, whatsapp does that too, the photo app doesn't since it has an horizontal scroller for ratio selection.

Its all good by me though

PR #12858, opened 2019-11-05
Gutenberg/free media library

etoledom:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/720

This PR implements the iOS Bridge code needed for Stock Photos (Pexels) and https://github.com/wordpress-mobile/gutenberg-mobile/issues/1265

`gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/1539

pexels

**Note:**
- The "Free Photo Library" is the same as used on Aztec.
- I decided to make the Free Photos Library multi-selectable, even though we haven't activated this feature on iOS yet. I can keep it single selectable if we decide that this is better.

- There are some few small changes on the bridge media picker/upload methods, so it would be good to test all options on all media related blocks.

- I noticed that we could improve some things in the JS side, but opted to keep this PR focused on adopting the current implementation to avoid the need of Android changes.

**To test:**

Image Block:
- Add an Image block.
- Check that the "Free Photo Library" option is there.
- Open the Free Photo Library.
- Choose one (or many) images.
- Check that all images selected are added.
- Check that you can select only images in all other options.

Video Block:
- Add a video block.
- Check that "Free Photo Library" is NOT in the options.
- Check that you can only select videos in all other options.

Media&Text:
- Add an Media&Text block.
- Check that the "Free Photo Library" option is there.
- Open the Free Photo Library.
- Choose one (or many) images. (Multi-select is also possible on Android)
- Check that all images selected are added. (Extra images will be added as Image Block)
- Check that you can select images and videos in all other options.

Update release notes:


etoledom:
Thank you!

Issue #11496, opened 2019-04-18
Quick Start v2 VoiceOver: toast messages are not accessible


designsimply:
Steps to reproduce:

1. Go to Settings > General > Accessibility > VoiceOver and toggle the setting on.
1. If you already have a site with Quick Start setup, go to that site, double tap on "Customize Your Site", and skip to step 7.
1. Log out of the WPiOS app.
1. Go to WPiOS and double tap the "Sign up for WordPress.com" button.
1. Follow the prompts until you reach "Want a little help getting started?"
1. Double tap the "Yes, Help Me" button.
1. Double tap the "x" button.
1. Try to tap "Not now" or "Yes, show me" in the toast message that appears.
1. Double tap the toast message that says "Tap Your Site Icon to upload a new one."

Result: all the references to “Tap” should be read as “Double tap” by VoiceOver. Icons are not read by VoiceOver. Things like “Tap + to do something” is read as “Tap to do something” Changing its meaning completely. ([57s](https://cloudup.com/cg1o0H9PgvC))

<img src="https://user-images.githubusercontent.com/1119271/56328966-9f4cfc00-613e-11e9-9c0f-d5b9d3686fcf.png" width="270" alt="wpios-voiceover-toast-message">
<p style="clear:both;"><sup>Tested with WP Internal 12.2.0.20190412 on iPhone 6S iOS 12.2.</sup></p>

<p style="clear:both;"><sub>(internal reference: p77Llu-bQ6-p2 h/t @etoledom)</sub></p>

stevebaranski:
Given a description, specifying a `accessibilityLabel` is straightforward. I took some time to evaluate this issue, and I'd like to share my findings with regard to _Quick Start_.

- We'll likely need to amend the [`Waypoint` tuple](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Blog/QuickStartTours.swift#L4), and specify the value (i.e., the accessible description to be set) to use in each `QuickStartTour` implementor.
[...]

PR #12873, opened 2019-11-05
[VoiceOver] Announce Notices when they are shown

shiki:
Fixes part of #11496.

This adds 2 things: 

- d34c7af Announcement of Notices when they are shown
- d04b6f1 Grouping the title and message together so they are dictated once by VoiceOver

### Option A

Run through Quick Start and validate that whenever a Notice is shown, VoiceOver will announce the title and message. 

I noticed that VoiceOver will sometimes not dictate the Notice when you go navigate back to the Site page from a different page (e.g. Themes). VoiceOver will either: 

- Select the previously selected element
- Select the announced Notice

Not sure if it's worth it to chase the fix for that. It seems to be out of our control. 🤷‍♂ 

### Option B

1. Replace [`showNoInternetConnectionNotice`](https://github.com/wordpress-mobile/WordPress-iOS/blob/b96410ddc5700b50fcba763ddad6da17bcb78b00/WordPress/Classes/Utility/ReachabilityUtils%2BOnlineActions.swift#L46-L50) with this:
    ```swift
            let notice = Notice(title: "This is a title", message: message, actionTitle: "Run", cancelTitle: "Slide", tag: DefaultNoConnectionMessage.tag)
    ```
1. Navigate to Posts List 
2. Go offline
3. Pull to refresh (swipe down with three fingers in VoiceOver)

## Submitter Checklist

---

@frosty Can I bother you with this one? 🙂

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
+++ b/WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
@@ -249,6 +249,8 @@ class NoticePresenter: NSObject {
             DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + Animations.dismissDelay, execute: dismiss)
         })
 
+        UIAccessibility.post(notification: .layoutChanged, argument: noticeContainerView)
+
         return NoticePresentation(notice: notice, containerView: noticeContainerView)
     }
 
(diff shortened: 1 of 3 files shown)


shiki:
> Not sure what would be best to fix this – a longer timeout if VoiceOver is active? or a specific 'double tap to dismiss' action on the main body of the notice itself?

Right. It is a bit tricky. It also depends on the message. For example, the “No internet connection” may not be as useful to keep it shown in the UI. While the Quick Start messages are useful to be sticky.

I tried to find an article/document I read yesterday but couldn't find it anymore. What I read was, for Snackbars/Notices, since they are UI elements that disappear after a few seconds, the app should provide another way to access the actions. And we're doing that in most cases. For example, after uploading, we show the “Post uploaded...” notice with a _View_ action. Even if the user does not have time to reach the action, they can still access it in the Post List. 

> We could always add that in a followup PR, as your changes here are at least an improvement over what we had already :)

Thanks for the approval! It looks like I broke the UI tests. I guess I'll have to fix those first before I merge this. 😭

shiki:
The UI tests were fixed and reviewed in #12889. Merging...

PR #12564, opened 2019-09-26
Issue/11555 Site Creation: Space in the title/tagline doesn't appear

marosios:
Fixes #11555 
- added a workaround for the text fiel to be able to show the spaces (spaces need to be replaced by unicode space)

To test:
1. Start creating a new site
2. When entering the title, enter some name with spaces, spaces are now normally visible during typing

Update release notes:
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12886, opened 2019-11-06
Issue/12796 Fix domain registration failure due to spaces

### LINKED ISSUE #12796: Domain Registrations are failing with "Please enter a valid address" a for valid addresses
### Expected behavior

Should be able to register a domain in the WordPress mobile app using `United Kingdom` as the country and `HU5 3RJ` or `HU53RJ` as the Postal Code. 

### Actual behavior

Selecting UK country code, not able to register a domain using the postal code `HU5 3RJ` - with or without the space will result in `Please enter a valid postal code`

1. Have WordPress.com site with domain credit
2. In App, tap Register domain, filling out contact information
3. Register domain once all fields are filled out.

frosty:
Fixes #12796. A previous PR, #12564, introduced some changes to the way spaces were handled in the domain registration form. Unfortunately, this wasn't entirely successful and has resulted in form validation failures due to fields containing unicode non-breaking spaces instead of actual space characters.

This PR fixes that by ensuring text field values are sanitised before being passed back to the delegate. See the first commit for this change.

I also took the liberty to improve the UX of the domain registration form while I was in there. The second two commits enable auto capitalization on fields in the form and allows the Next button to tab to the next available editable field.

**To test:**

* Access domain registration for one of your sites. I did this by commenting out code at the start of `configureTableViewData` in `BlogDetailsViewController` so that the domains row is visible.
* Select a domain to get to the registration form.
* Enter the address `1200 Getty Center Dr, Los Angeles, CA 90049`, and ensure that no validation errors occur (without this patch, you'll get an error for address and city).
* While you're doing that, check you can use Next to move between fields, and your words are capitalised.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Cells/InlineEditableNameValueCell.swift
+++ b/WordPress/Classes/ViewRelated/Cells/InlineEditableNameValueCell.swift
@@ -13,10 +13,15 @@ import WordPressAuthenticator
 
 class InlineEditableNameValueCell: WPTableViewCell, NibReusable {
 
-    enum Const {
+    fileprivate enum Const {
         enum Color {
             static let nameText = UIColor.text
             static let valueText = UIColor.textSubtle
+        }
+
+        enum Text {
+            static let nonBreakingSpace = "\u{00a0}"
+            static let space = " "
         }
     }
 
@@ -64,7 +69,7 @@ class InlineEditableNameValueCell: WPTableViewCell, NibReusable {
     }
 
     @objc func textFieldDidChange(textField: UITextField) {
-        textField.text = textField.text?.replacingOccurrences(of: " ", with: "\u{00a0}")
+        textField.text = textField.text?.replacingOccurrences(of: Const.Text.space, with: Const.Text.nonBreakingSpace)
 
         let text = sanitizedText(for: textField)
         delegate?.inlineEditableNameValueCell?(self, valueTextFieldDidChange: text)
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Cells/InlineEditableNameValueCell.swift
```diff
@@ -61,19 +65,33 @@ class InlineEditableNameValueCell: WPTableViewCell, NibReusable {
...
-        textField.text = textField.text?.replacingOccurrences(of: "\u{00a0}", with: " ")
-        delegate?.inlineEditableNameValueCell?(self, valueTextFieldEditingDidEnd: textField)
+        let text = sanitizedText(for: textField)
+
+        textField.text = text
+        delegate?.inlineEditableNameValueCell?(self, valueTextFieldEditingDidEnd: text)
+    }
+
+    private func sanitizedText(for textField: UITextField) -> String {
+        return textField.text?.replacingOccurrences(of: "\u{00a0}", with: " ") ?? ""
```
danielebogo: Because the `\u{00a0}` is used even above, can we wrap it in a _Constant_ enum?

frosty: Good idea!

frosty:
@danielebogo I added a constant for the spaces. I re-tested the scrolling issue and it turns out it's a problem on devices without a home button, as the view tries to adjust for safe insets.

For now, I've removed the code that adjusts that inset, which fixes the scroll issue you were seeing. The side-effect is that now if you scroll to the bottom of the table view with the keyboard hidden, you'll see that the white background behind the bottom button doesn't extend behind the home indicator. I think that's an ok tradeoff for this PR – the way to fix that issue would, I think, be to redo the view hierarchy of this screen so that the footer isn't part of the tableview, but I didn't want to make such a big change here.

PR #12889, opened 2019-11-06
Update NoticeComponent to accommodate VoiceOver changes

shiki:
Fixes the failing UI tests caused by #12873. 

## Findings

In #12873, I changed the accessibility of `NoticeView` so that both the title and message are dictated as a single unit (d04b6f12abe9154a7bea0e7e64d817133d605cfb). This broke the editor UI tests. 

## Solution

This PR changes `EditorNoticeComponent` so that:

- 881a6c2 It will wait for the container of the Notice's title and message 
- ac4a04f It will validate based on the expected title and message

I'd love to have your opinion on the changes as I wasn't sure of how the UI test architecture connects all the classes together. 

## Testing

I believe testing that `EditorAztecTests` and `EditorGutenbergTests` pass should be good enough. 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
+++ b/WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
@@ -52,7 +52,7 @@ class NoticeView: UIView {
             configureDismissRecognizer()
         }
 
-        configureForVoiceOver()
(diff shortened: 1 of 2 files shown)


shiki:
Thank you, @rachelmcr!

PR #12891, opened 2019-11-06
[VoiceOver] Make WPImageViewController dismissible 

### LINKED ISSUE #12867: [VoiceOver] Reader: Cannot exit a Post View Image
Tap on a post's image. This will show an image preview. Nothing on this page is accessible. It does not seem to be possible to exit.

shiki:
Fixes #12867

## Findings

When `WPImageViewController` is presented by itself, not inside a `UINavigationController`, the VoiceOver escape gesture does not automatically handle the dismissal. 

## Solution

This adds 2 things:

- 93bbc6d Make the image accessible so the VoiceOver user can interact with it.
- 85458c3 React to the escape gesture and dismiss self

## Testing

1. Navigate to Reader and to a post with an image.
2. Turn on VoiceOver
3. Tap on the image. The `WPImageViewController` should be shown. 

Confirm that you can:

- Double-tap on the image to dismiss the view. 
- Use the escape gesture to dismiss the view (move two fingers back and forth three times quickly, making a “z”).

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/WPImageViewController.m
+++ b/WordPress/Classes/ViewRelated/Reader/WPImageViewController.m
@@ -152,6 +152,8 @@ static CGFloat const MinimumZoomScale = 0.1;
     [self setupActivityIndicator];
     [self layoutActivityIndicator];
 
+    [self setupAccessibility];
+
     [self loadImage];
 }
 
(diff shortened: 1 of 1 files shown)


shiki:
Thank you, @ScoutHarris!

Issue #12849, opened 2019-11-04
[Stats Widgets] Update Today widget to new design


ScoutHarris:
Update current Today widget to new design. 

Current widget:

<img width="400" alt="current_widget" src="https://user-images.githubusercontent.com/1816888/68164260-a2141080-ff19-11e9-8a79-3da258696046.png">

New design from Zeplin:

<img width="369" alt="today_widget" src="https://user-images.githubusercontent.com/1816888/68164114-3d58b600-ff19-11e9-97a9-5f335858c045.png">

PR #12893, opened 2019-11-06
[Stats Widgets] Update current Today widget to new design

ScoutHarris:
Ref #12849 

This updates the current Today widget to the new design. Specifically, the "compact" view, which is the only one currently displayed. 

Expanded view will be addressed in a future PR. To facilitate that, I've included some things in this PR that may seem unnecessary, but will be used later:
- Likes and Comments are now saved to the plist, but aren't displayed yet.
- The data (Views and Visitors) is inserted with a `TwoColumnRow` view. This will allow easily adding another data "row" when the view is expanded.
- `initRows` only inits one row. But it will be multiple when expanded, hence the plural instead of singular. Stay tuned!

To test:

---
- Start with a fresh install.
- Add the WP widget to the Today view.
- Verify the unconfigured view looks like this:

unconfigured_light

unconfigured_dark

---
- Log into the app and go to Stats.
- Select 'Today' in the nav bar and select 'Use this site'.
- Return to the widget.
- Verify the view looks like this. Things to note:
  - The widget display name is now "WORDPRESS TODAY".
  - The site url is displayed at the bottom.

configured_light

configured_dark

---
Update release notes:


ScoutHarris:
Hey @osullivanchris . Here are some hi res screenshots for you. I'm not sure about the dark mode button background color (the "Open WordPress" button). Let me know if you'd like any changes.

**Light Mode:**
light_unconfigured

light_configured

**Dark Mode:**
dark_unconfigured

dark_configured

WordPress/WordPressTodayWidget/TodayViewController.swift
```diff
@@ -8,48 +8,44 @@ class TodayViewController: UIViewController {
...
-    @IBOutlet var configureMeButton: UIButton!
+    @IBOutlet private weak var unconfiguredView: UIStackView!
+    @IBOutlet private weak var configureLabel: UILabel!
+    @IBOutlet private weak var configureButton: UIButton!
+
+    @IBOutlet private weak var configuredView: UIStackView!
+    @IBOutlet private weak var rowsStackView: UIStackView!
+    @IBOutlet private weak var separatorLine: UIView!
+    @IBOutlet private weak var siteNameLabel: UILabel!
+    @IBOutlet private weak var siteUrlLabel: UILabel!
```
danielebogo: Can me make these strong?

WordPress/WordPressTodayWidget/TodayViewController.swift
```diff
@@ -152,24 +171,51 @@ private extension TodayViewController {
         let data = TodayWidgetStats.loadSavedData()
         visitorCount = data.visitors
         viewCount = data.views
+        likeCount = data.likes
+        commentCount = data.comments
+    }
+
+    func initRows() {
+        guard let row = Bundle.main.loadNibNamed(Constants.rowNibName, owner: nil, options: nil)?.first as? TwoColumnRow else {
```
danielebogo: You can use `loadRootViewFromNib` from `Bundle+LoadFromNib`

WordPress/WordPressTodayWidget/TodayViewController.swift
```diff
@@ -152,24 +171,51 @@ private extension TodayViewController {
...
-    func changeTextColor() {
-        configureMeLabel.textColor = .text
-        siteNameLabel.textColor = .text
-        visitorsCountLabel.textColor = .text
-        viewsCountLabel.textColor = .text
-        visitorsLabel.textColor = .textSubtle
-        viewsLabel.textColor = .textSubtle
+    func displayString(for value: Int) -> String {
+        let numberFormatter = NumberFormatter()
+        numberFormatter.numberStyle = .decimal
```
danielebogo: Can the formatter be a `let` stored property with closure? Just to avoid to create a formatter every time the method is invoked.

WordPress/WordPressTodayWidget/TodayViewController.swift
```diff
@@ -186,4 +232,18 @@ private extension TodayViewController {
         return StatsServiceRemoteV2(wordPressComRestApi: wpApi, siteID: siteID.intValue, siteTimezone: timeZone)
     }
 
+    struct LocalizedText {
```
danielebogo: Nitpick, `enum` here is safer  than `struct`

WordPress/WordPressTodayWidget/TodayViewController.swift
```diff
@@ -186,4 +232,18 @@ private extension TodayViewController {
...
+        static let configure = NSLocalizedString("Display your site stats for today here. Configure in the WordPress app in your site stats.", comment: "Unconfigured stats today widget helper text")
+        static let openWordPress = NSLocalizedString("Open WordPress", comment: "Today widget button to launch WP app")
+        static let visitors = NSLocalizedString("Visitors", comment: "Stats Visitors Label")
+        static let views = NSLocalizedString("Views", comment: "Stats Views Label")
+    }
+
+    struct Constants {
```
danielebogo: Here too 😊

WordPress/WordPressTodayWidget/TodayViewController.swift
ScoutHarris: Well, I'd have to include that file in the widget's target. Which I could do. But I don't see a reason to add another file to the widget footprint for just one method when this works fine.

danielebogo: Ah got it! Was just to don't have duplication since there's an extension for it, but it's ok if you don't want to include it.

osullivanchris:
@ScoutHarris Hey! Again, this looks to me like its using a system material and should be a semi transparent background rather than a hard grey colour. 

In my files I found that using the 'thin' system material approximated the background of the box, and ultrathin on top of it for the button background, seemed to work well.

I am not sure if the naming convention is the same in the design components they provide as in the components/code. But what you have for the box background looks right, and I've gone one level lighter/thinner for the button background.

Here's how it looks on black like in Sylvester's mock, and on a phone background.

Screenshot 2019-11-07 at 17 32 22

ScoutHarris:
@frosty @osullivanchris 

> there's really no room for the .wordpress.com part of the site URL which makes me question the usefulness of having it there. I wonder if the view is compact if we should remove either the title or the URL and allow more space for whichever we keep. Any thoughts?

Just a couple comments. 

> I wonder if the view is compact

Compact vs expanded only affects the height, so this won't help.

> remove either the title or the URL

It is possible for a site to not have a title. If we elect to show only one value, maybe it should be the url so _something_ is shown? Or we could show the site name if it exists. If not, fallback to the url.

osullivanchris:
Sorry it just re-loaded when I added my comment and I saw @frosty's. I don't think its a big deal to remove one of those, I agree its better not to truncate especially twice.

The url seems a safe bet if there is sometimes not a title. The URL is also technically the thing that people are visiting, so its rational to me to have that one in there.

ScoutHarris:
@osullivanchris Can I make a suggestion?

There really is no need to have a button. It doesn't actually do anything as the entire widget is tappable. Maybe it would look better if we just remove it, and do something like this?

no_button

frosty:
> There really is no need to have a button. It doesn't actually do anything as the entire widget is tappable.

That's definitely true, but I think having the visible button there makes the whole thing _look_ more tappable?

danielebogo:
@osullivanchris @ScoutHarris @frosty Should the button follow one of these style? Or this is a specific/different case?
buttons

osullivanchris:
> @osullivanchris @ScoutHarris @frosty Should the button follow one of these style? Or this is a specific/different case?

I think this is definitely a different case, the widgets all look totally part of the native UI and having our own button styles would look out of place there

> There really is no need to have a button. It doesn't actually do anything as the entire widget is tappable.


> That's definitely true, but I think having the visible button there makes the whole thing look more tappable?

All the other widgets I've checked don't really have buttons. Just links/content. What about a divider? I think it makes it clear that there is an action there without needing the background. I also tried two further explorations with the text. 

In the second option, I was thinking its a bit weird to say 'Tap' on a touch device, a bit redundant. And the third option - I was wondering if we are deep linking to stats here or just opening the app? If we can deep link, perhaps we can make it clear that we're gonna help show you where to set up the widget? (if we're doing this maybe get a quick copy review). 

Let me know if you agree!

widget

ScoutHarris:
> I was wondering if we are deep linking to stats here or just opening the app? 

@osullivanchris we're just opening the app. Deep linking into stats requires a Site ID in order to know what stats to display. We don't have that yet, hence this view. 😄 

It sounds like we'll be going with the 2nd option.

Screen Shot 2019-11-08 at 11 07 57 AM

Thank you all for your feedback!

ScoutHarris:
@osullivanchris @danielebogo @frosty I believe I've covered all the feedback. I've updated the description with new screenshots. Ready for another review please! Thanks!

ScoutHarris:
Hey @osullivanchris . Final hi res screenshots for you. 

**Light Mode:**

unconfigured_light

configured_light


**Dark Mode:**

unconfigured_dark

configured_dark

osullivanchris:
@ScoutHarris thanks! as its a final screenshot I just did a double check against the mocks :) 

The only two things I see different are
- the mocks have a header background which the screenshots dont. However, again comparing to other widgets I see none with headers. So perhaps that's an outdated style. Let's leave it per the build.
- The dividers in the screenshots are full-width. In the mocks, for the ones with stats highlights (2-4 stats) the dividers are padded in from the left to align with the left-aligned content. Note that its just for those cards. Well this is a detail for sure anyway - if its a quick fix lets change - if not its definitely not a blocker.

Thanks for the ping!

ScoutHarris:
Hey @osullivanchris .

> the mocks have a header background which the screenshots dont. However, again comparing to other widgets I see none with headers. So perhaps that's an outdated style. Let's leave it per the build.

You are correct. There is no way to customize the header - it's iOS being iOS.

> The dividers in the screenshots are full-width. In the mocks, for the ones with stats highlights (2-4 stats) the dividers are padded in from the left to align with the left-aligned content. Note that its just for those cards. Well this is a detail for sure anyway - if its a quick fix lets change - if not its definitely not a blocker.

I was referencing Zeplin, which has them full width. Although I will say, left indenting looks a little odd... Anyway, if you want it so, I'd be happy to do so in my next PR of that's OK.

Thanks!

ScoutHarris:
> I'd be happy to do so in my next PR of that's OK.

And I just realized you have no idea what's next. 😄 I'm working expanding the widget, so I'm modifying the view anyway. I can make any adjustments with that.

PR #12894, opened 2019-11-06
Show Media Editor when choosing image from device in Gutenberg

leandroalonso:
Fixes #

To test:

Update release notes:


Podfile
```diff
@@ -156,6 +160,11 @@ target 'WordPress' do
     pod 'AlamofireNetworkActivityIndicator', '~> 2.3'
     pod 'FSInteractiveMap', :git => 'https://github.com/wordpress-mobile/FSInteractiveMap.git', :tag => '0.2.0'
 
+    ## Media Editor Third party libraries
+    ## =====================
+    ##
+    media_editor
```
jkmassel: Including this dependency in both places causes linker ambiguity – ideally we'd only include it once?

```
objc[95732]: Class TOActivityCroppedImageProvider is implemented in both [...snip...]/MediaEditor.framework/MediaEditor (0x10ff45278) and [...snip...]/WordPress.app/WordPress (0x10c256cb0). One of the two will be used. Which one is undefined.
```

leandroalonso:
I'm closing this PR because of some conversations @osullivanchris, @iamthomasbishop and I had.

We'll not focus on adding this flow to Gutenberg right now since the project will have multiple media selection in the next weeks and we haven't agreed on adding a new step in the current flow.

Issue #12866, opened 2019-11-05
[VoiceOver] Reader: Post View concerns


shiki:
This is for the page accessed when navigating to a post from _Followed Sites_ → 

## A: Cannot exit after tapping on an image ⚠️

Tap on a post's image. This will show an image preview. Nothing on this page is accessible. It does not seem to be possible to exit. 

## B: “Attachment. PNG. File” is dictated first. 

Open a post and the text “Attachment. PNG. File” will be dictated. My guess is that this is the featured image. The problem is this is still dictated even if there is no featured image. 

## C: The header cannot be navigated to when swiping

Please see [this video](https://drive.google.com/file/d/1-t3npWYakWyO9PeIvC_1c-LQkWNuEhAY/view?usp=sharing)

## D: The _More_ button in the header is not accessible

But even without VoiceOver, it looks like it cannot be tapped anyway.

shiki:
> B: The header cannot be navigated to when swiping

I tried to fix the VoiceOver navigation order but failed. As of now, the navigation can go like this:

1. Site name + URL
2. More Button
3. Post Title
4. Author + Date
5. Save button ❌ (not the post content)

Or like this:

1. View controller title 
2. Share button
3. The content's first paragraph ❌ (not the site name) 

I think I need to understand how the Reader detail's UI works some more to fix this.

dangermattic:
Thanks for reporting! 👍

PR #12901, opened 2019-11-07
Reader: Merge site name and URL for VoiceOver

shiki:
Refs #12866.  

This is just a minor improvement on the Reader Detail's header.

I removed the site URL from the accessible elements. It was too small as a touch target and the site name (plus URL) being dictated together should be good enough. 

Removed | Improved
--------|-------
IMG_0993        |       IMG_0992

## Testing

1. Turn VoiceOver on.
2. Navigate to a Reader post 
3. Validate that you can no longer navigate to the site URL using VoiceOver. 

## Reviewing 

## Submitter Checklist


shiki:
Thank you, @danielebogo!

PR #12351, opened 2019-08-21
Reader: Initialize attachments with default image data.

aerych:
Refs #12211 

This is a first pass at resolving some artifacting be hind images in the Reader on iOS 13.  It looks like NSTextAttachments that are initialized without default data, file reference, or image will show a placeholder graphic.  Adding a 1px x 1px white image as starting data seems to hide the placeholder. 

This is the first thing I thought of, and there might be a more elegant solution. Other ideas welcome. 

To test:
Confirm the artifacting by viewing the reader post detail with an iOS 13 development build. Look for any post with an attachment--image or media.
Switch to this branch and confirm that the artifacting no longer appear.

Update release notes:

- [x] If there are user facing changes, I have added an item to `RELEASE-NOTES.txt`.

PR #12903, opened 2019-11-07
[VoiceOver] Reader Detail: Hide “Attachment. PNG. File” dictation

shiki:
Refs #12866. 

This hides the dictated “Attachment. PNG. File“ message.

## Findings

The dictation was caused by the `ReaderPlaceholderAttachment`. The message gets dictated even for images. 

## Solution

For the life of me, I could not hide it from VoiceOver using `isAccessibilityElement`. I ended up using an empty `accessibilityLabel` instead. 

In addition, I also made the `CachedAnimatedImageView` used in `WPRichTextImage` accessible so when a user hovers over it, VoiceOver would announce that it's an image rather than be silent. 

## Testing

1. Turn VoiceOver on.
1. In Reader, navigate to a post with an image.
2. Hover over the first paragraph. Confirm that VoiceOver doesn't say ”Attachment. PNG. File” anymore and just proceeds to dictate the paragraph's content. 
4. Hover over an image. Confirm that VoiceOver will announce that it's an image.

Please also watch for regressions, particularly the one fixed by #12351. 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
@@ -15,6 +15,18 @@ class ReaderPlaceholderAttachment: NSTextAttachment {
     required init?(coder: NSCoder) {
         super.init(coder: coder)
     }
+
+    override var accessibilityLabel: String? {
+        get {
+            // Setting isAccessibilityElement to false does not seem to work for this
+            // `NSTextAttachment`. VoiceOver will still dictate “Attachment. PNG. File” which is
+            // really weird. Returning an empty label here so nothing will just be dictated at all.
+            return ""
+        }
+        set {
+            super.accessibilityLabel = newValue
+        }
+    }
 }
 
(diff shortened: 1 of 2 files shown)


shiki:
@aerych Can I bother you with this one? 🙂

paskalem:
😄of course you can

aerych:
@paskalem :( This sort of comment isn't very helpful.   It could be taken as you're answering for someone else, and its doesn't add any value to the thread.  It feels more like a random drive-by comment. 

We get very excited when new contributors take an interest in the project, and we'd like to be welcoming. If you've taken a real interest in WordPress for iOS, maybe have a quick peek at our [code of conduct](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/CODE-OF-CONDUCT.md), and our guide to [contributions](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/CONTRIBUTING.md).  :)

shiki:
Thank you, Eric! 🙏

Issue #12919, opened 2019-11-11
Change Media Editing project setup


leandroalonso:
The original setup of the Media Editing project has some downsides that we want to avoid:

* The dependency is dynamically linked, which causes the app startup time to increase
* Management of sub-targets in App Store Connect is far from ideal

The two possible solutions for the new setup are:

* Add the files into the main WPiOS project
* Or create it as a Pod dependency

PR #12924, opened 2019-11-11
Change Media Editing project setup

leandroalonso:
Implements #12919

### Why the changes?

* The original setup (embedded framework) caused it to be dynamically linked — and any of its dependencies. This slows down app startup time;
* Management of sub-targets in App Store Connect is hard.

The only downside from this setup is that the boundaries between WPiOS and the Media Editor are not enforced by the IDE. However, it will give us more dynamicity to develop/make changes in this phase (without the need to create a `Podspec`, reference it and so on).

### To test:

The green build should be enough.

Update release notes:


leandroalonso:
@shiki done in 9e96703e5d333b429af9e19c955fcaaad7d2c511

shiki:
I talked to Leandro. We missed moving the tests file. He'll take care of it.

leandroalonso:
@shiki Fixed in 6e21c07bbfb91aa5b274dfb3dac3379925783656

PR #12929, opened 2019-11-12
Update Aztec-Editor to 1.12 and add block editor paragraph alignment

mchowning:
This updates the AztecEditor to [version 1.12](https://github.com/wordpress-mobile/AztecEditor-iOS/releases/tag/1.12.0) and brings in the changes for paragraph alignment that needed that update.

To test:

* Smoke test of Aztec Editor
* Check paragraph alignment behavior in gutenberg editor (previously [this issue](https://github.com/wordpress-mobile/gutenberg-mobile/issues/1545) was observed and filed, but I'm actually not seeing that issue anymore, cc: @SergioEstevao )
* Check background color appearance in quote blocks

### TODO

- [ ] Update gutenberg ref after https://github.com/wordpress-mobile/gutenberg-mobile/pull/1489 is merged

### Update release notes


mchowning:
With @SergioEstevao 's [update to the AztecEditor](https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1229/files#diff-8d8015f3f021c0252c1569e4965dce80R278), I needed to have a default color for the blockquote background color. Although my change looked fine to me, I'm not especially confident that the color I picked is the best option. It's worth having a particularly skeptical look at those changes.

SergioEstevao:
@mchowning I hope you don't mind but I updated the colors values in your PR to set the quote and pre backgrounds. Here is how it looks now:

AztecPreQuote

iamthomasbishop:
@SergioEstevao Are we using system grays in Aztec? I would probably use a lighter gray on light mode, but I think dark mode looks fine.

I'm curious which specific colors we are using here for blockquote/pre – in light mode and dark mode?

SergioEstevao:
> I'm curious which specific colors we are using here for blockquote/pre – in light mode and dark mode?

We are using the new Muriel pallete 
so for quote we are using .shade(5)/.shade(90) - gray5 and gray90
and fore pre we are using .shade(10)/.shade(80)- gray10 and gray80

iamthomasbishop:
Ah, ok – the screenshot's colors looked slightly different from the current values on [Color Studio](https://color-studio.blog), so maybe it's just a color profile thing in the screenshot.

We can just use the same backgound color for both pre and quote. So in other words, `gray 5` in light mode and `gray 90` in dark mode.

image

SergioEstevao:
@iamthomasbishop update the color in pre to be the same in quote.

The issue with the screenshots it's because I have True Tone activated on my laptop. It messes the colors a bit.

PR #12930, opened 2019-11-12
Testing: Convert unit & UI tests to test plans

rachelmcr:
This PR converts our existing unit and UI tests to use the new test plans in Xcode 11. This doesn't change what tests are run or when, but it does remove the need for two build-for-testing steps on CircleCI. Both unit and UI test plans are linked to the WordPress scheme now (with unit tests as the default).

To test:

* Confirm all of the unit tests are run and passing on CircleCI.
* Confirm all of the UI tests are run on both iPhone and iPad and passing on CircleCI.
* In Xcode, confirm that when working with the WordPress scheme, unit tests are run by default and UI tests can be run by selecting the WordPressUITests test plan in the Test Navigator or under Product > Test Plan:

Screenshot 2019-11-12 18 51 57
Screenshot 2019-11-12 18 52 50

Update release notes:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,9 +39,6 @@ jobs:
             pod-install: true
       - ios/xcodebuild:
           command: build-for-testing
-          arguments: -workspace 'WordPress.xcworkspace' -scheme 'WordPressUITests' -configuration 'Debug' -sdk iphonesimulator -derivedDataPath DerivedData
-      - ios/xcodebuild:
-          command: build-for-testing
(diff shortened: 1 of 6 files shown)


rachelmcr:
> For some reason, the media export tests fail locally for me, but they're working on CI, so I think it's ok and can be chalked up to XC11 weirdness.

I bet I know what's causing that. I saw some media-related failures in a couple test runs caused by the picker not being fully loaded and have a solution for that; I'll get a PR up soon to try to head off any possible flakiness of that sort.

Thanks for the review!

PR #12931, opened 2019-11-12
Remove Muriel palette workaround for Xcode 11 GM issue

frosty:
This PR removes the temporary workaround we added for an issue with the Xcode 11 GM where colors weren't loaded correctly from asset catalogs on iOS < 12. This is no longer an issue.

**To test**

* In Xcode 11, install an iOS 11 simulator, or use an iOS 11 device.
* Build and run on both iOS 13 and iOS 11
* Ensure there are no views displaying that are coloured bright red. This is what we do as a fallback if no colors are able to be loaded from the asset catalog. It'll be pretty obvious.

Update release notes:


frosty:
Thanks!

Issue #12906, opened 2019-11-08
[VoiceOver] Notifications page concerns


shiki:
This is for the first page shown when tapping on the Notifications tab bar button. 

## A: Cells do not have the `.button` trait

I'm not sure if we need to add hints as well. 

- [x] Done in #12932

## B: The empty view may need to be improved

- [x] Done in #12943. 

Discussions are still ongoing at https://github.com/wordpress-mobile/WordPress-iOS/issues/11487

## C: _Mark Unread_, _Unapprove_, and other cell actions are not discoverable

Without VoiceOver, you can swipe left and right to see these actions. With VoiceOver, they do not seem to be discoverable or accessible at all. 

- [x] Invalid report

shiki:
> C: Mark Unread, Unapprove, and other cell actions are not discoverable

I found that I can do this after all by the swipe up/down actions. I couldn't do it before. Maybe one of the new PRs fixed it. 🤔

PR #12932, opened 2019-11-12
[VoiceOver] Enhancements for main Notifications page

shiki:
Refs #12906 

This enhances VoiceOver support for the Notifications page. 

### Cells as buttons

The cells now have the `.button` trait so that it's more obvious that they can be tapped. I also added hints (66a9c26) for the notification kinds that I know of. I couldn't find information on how to _create sample notifications_ the following:

- `.post`
- `.user`

### Headers 

Before | After 
--------|-------
IMG_1034        |       IMG_1033

I increased the accessible width of the table headers and added `.heading` traits on them. This makes them navigatable via Headings using [the rotor](https://support.apple.com/en-ca/guide/iphone/iph3e2e3a6d/ios) and also just simply give out more useful information than a normal label. 

## Testing

1. Turn VoiceOver on
2. Navigate to Notifications
3. Verify the changes have been implemented as described above. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -1075,6 +1075,8 @@ extension NotificationsViewController: WPTableViewHandlerDelegate {
             self?.cancelDeletionRequestForNoteWithID(note.objectID)
         }
 
+        cell.accessibilityHint = Self.accessibilityHint(for: note)
+
         cell.downloadIconWithURL(note.iconURL)
     }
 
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1110,8 +1112,26 @@ extension NotificationsViewController: WPTableViewHandlerDelegate {
             selectFirstNotificationIfAppropriate()
         }
     }
-}
 
+    private static func accessibilityHint(for note: Notification) -> String? {
```
danielebogo: Out of curiosity, why declare this as static?

shiki: Thanks for asking! 

I wanted to explicitly declare that this is a pure utility function and shouldn't rely on anything in `NotificationsViewController`. If it was an instance method, somebody (like me, especially) might be tempted to use one of the `NotificationsViewController`'s properties when they modify this in the future. It should also help in debugging because we'll know that this doesn't use any state aside from `Notification`. 

Do you have some suggestions on how to make this better though? How do you usually implement things like this? I thought about adding a function at the bottom of the file but I wasn't sure if that's an accepted way.

shiki:
Thank you, @danielebogo!

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
danielebogo: @shiki got it! Except for the protocol we use to make a type `Accessible` there's no specific rule for it. I think it depends mostly by the type and its design/context. So feel free to improve the `Accessible` protocol if you want or to find a better way to manage the accessibility in WP 😊

Issue #12920, opened 2019-11-11
Edit already published images in Aztec


leandroalonso:
Using Aztec:

1. Add an image to the post
2. Tap the image
3. The Action Sheet should show "Edit"

Using Aztec:

1. Add an image to the post
2. Tap the image
3. Tap "Edit" in the action sheet
4. Crop the image
5. Tap "Done"
6. The image should be replaced

Using Aztec:

1. Add an image to the post
2. Change the image: alignment, link, size, alt text and caption
2. Tap the image
3. Tap "Edit" in the action sheet
4. Crop the image
5. Tap "Done"
6. Check that the image was replaced and alignment, link, size alt text, and caption were maintained

PR #12933, opened 2019-11-12
Edit already published images in Aztec

leandroalonso:
Implements https://github.com/wordpress-mobile/WordPress-iOS/issues/12920

Depends on https://github.com/wordpress-mobile/WordPress-iOS/pull/12924

### Also related

* https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1233
* https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/225

#### Just cropping

1. Add an image to the post
2. Tap the image
3. Tap "Edit" in the action sheet
4. Crop the image
5. Tap "Done"
6. The image should be replaced

#### Cropping maintaining all the image properties

1. Add an image to the post
2. Change the image: alignment, link, size, alt text, and caption
2. Tap the image
3. Tap "Edit" in the action sheet
4. Crop the image
5. Tap "Done"
6. Check that the image was replaced and alignment, link, size alt text, and caption were maintained

#### Offline

1. Go offline
2. Create a new post with an image or edit an already published image
3. Check that the image is replaced and uploaded when the device is back online

### Design details

Please, ignore the design details right now. We'll have a review and a task dedicated to improving that.

Update release notes:

- [x] If it's feasible, I have added unit tests — unfortunately, `AztecPostViewController` is big and hard to test.


WordPress/Classes/Utility/MediaEditor/MediaEditor.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    public init(cropViewControllerFactory: @escaping (UIImage) -> TOCropViewController = TOCropViewController.init) {
+        self.cropViewControllerFactory = cropViewControllerFactory
+        super.init()
+    }
+
+    public func edit(_ image: UIImage, from viewController: UIViewController? = nil, onFinishEditing: @escaping (UIImage?) -> (), onCancel: (() -> ())? = nil) {
+        self.onFinishEditing = onFinishEditing
+        self.onCancel = onCancel
+        let cropViewController = self.cropViewControllerFactory(image)
```
yaelirub: You can remove the `self` here

WordPress/Classes/Utility/MediaEditor/MediaEditor.swift
```diff
@@ -0,0 +1,49 @@
...
+        self.cropViewController = cropViewController
+    }
+
+    public func dismiss(animated: Bool, completion: (() -> ())? = nil) {
+        cropViewController?.dismiss(animated: animated, completion: completion)
+    }
+
+    private func releaseCallbacks() {
+        onFinishEditing = nil
+        onFinishEditing = nil
```
yaelirub: Did you mean for the second one to be `onCancel = nil`

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -3407,3 +3420,27 @@ extension AztecPostViewController: PostEditorNavigationBarManagerDelegate {
...
+                  onFinishEditing: { image in
+                                if let image = image {
+                                    self.replace(attachment: imageAttachment, with: image)
+                                    self.mediaEditor.dismiss(animated: true)
+                                }
+        })
+    }
+
+    private func replace(attachment: ImageAttachment, with image: UIImage) {
+        let info = MediaAnalyticsInfo(origin: .editor(.deviceLibrary), selectionMethod: mediaSelectionMethod)
```
yaelirub: Should we pass the MediaSource to the replace function instead of hardcoding to .deviceLibrary here?

WordPress/WordPressTest/MediaEditorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    func testReturnsTheCroppedImage() {
+        let originalImage = UIImage()
+        let croppedImage = UIImage()
+        var returnedImage: UIImage?
+
+        mediaEditor.edit(originalImage, onFinishEditing: { croppedImage in
+        returnedImage = croppedImage
+
```
yaelirub: Extra line. Did you originally have `cropViewControllerMock.croppedImage = croppedImage` in the onFinishEditing?

WordPress/WordPressTest/MediaEditorTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        expect(self.cropViewControllerMock.didCallDismiss).to(beTrue())
+    }
+
+    func testPresentCropViewControllerFromAGivenUIViewController() {
+        let image = UIImage()
+        let viewControllerMock = UIViewControllerMock()
+
+        mediaEditor.edit(image, from: viewControllerMock, onFinishEditing: { _ in })
+
+        expect(viewControllerMock.didCallPresent).to(beTrue())
```
yaelirub: Is it sufficient or should we test the presented is TOCropViewController?

yaelirub:
Tested by described steps. Works as expected

WordPress/Classes/Utility/MediaEditor/MediaEditor.swift
leandroalonso: Done

WordPress/Classes/Utility/MediaEditor/MediaEditor.swift
leandroalonso: Whoops! Done

WordPress/WordPressTest/MediaEditorTests.swift
leandroalonso: Fixed!

> Did you originally have `cropViewControllerMock.croppedImage = croppedImage` in the onFinishEditing?
It was outside `onFinishEditing`. :)

WordPress/WordPressTest/MediaEditorTests.swift
leandroalonso: Refactored to test this behavior.

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
leandroalonso: This is a good question. For some reason, I forgot to update that.

So, since this image is already presented in the post, I can't tell from where it comes. But we know that if it's being inserted again, it came from the Media Editor. I changed to `.editor(.mediaEditor)`. :)

WordPress/WordPressTest/MediaEditor/MediaEditorTests.swift
```diff
@@ -6,71 +6,97 @@ import Nimble
...
         var returnedImage: UIImage?
-        let mediaEditor = MediaEditor.init(cropViewControllerFactory: TOCropViewControllerMock.init)
 
-        mediaEditor.edit(originalImage) { croppedImage in
+        mediaEditor.edit(originalImage, onFinishEditing: { croppedImage in
             returnedImage = croppedImage
-        }
-        (mediaEditor.cropViewController as? TOCropViewControllerMock)?.croppedImage = croppedImage
+        })
+        cropViewControllerMock.croppedImage = croppedImage
```
yaelirub: I don't think I understand this test. Questions:
1. Shouldn't this be `cropViewControllerMock.croppedImage = returnedImage` ?
2. Is it right that the mediaEditor.edit method is guaranteed to be on the main thread and that is why you are assigning value to  cropViewControllerMock.croppedImage outside of the onFinishEditing block?
3. Why are we assigning value to cropViewControllerMock.croppedImage if we are not using it in the test or did you mean to assert something else?

leandroalonso: I guess the syntax used was confusing. I changed it (1209d5755d97445f3e6e54f1a95a4a3fee5321f9), let me know what you think.

What this line is doing is simulating that the user has cropped the image and telling to the mock which image it should return.

PR #12934, opened 2019-11-12
Enable Preformatted block in Gutenberg.

SergioEstevao:
Enables the preformatted block in iOS for release 1.17.0

To test:
 - Start a post using Gutenberg
 - Add a preformatted block to the post
 - See if it works correctly and interacts correctly with other blocks.

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -140,7 +140,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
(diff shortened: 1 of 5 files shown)


koke:
> See if it works correctly and interacts correctly with other blocks

Everything else looked good, but I spotted an ugly bug. If you press backspace at the start of a paragraph block after preformatted, when they merge all the newlines in the preformatted block are gone:

Screen Recording 2019-11-13 at 13 36 48

SergioEstevao:
We detected this before it happens on the web too.

Issue #12918, opened 2019-11-11
Enhanced / Full Screen Commenting


frosty:
At the moment the comment input field is restricted to 5 lines of text. Editing a comment or composing a longer comment can, therefore, be fiddly and generally a bad experience. We want to improve this experience and remove the frustration from composing a longer comment.

**Goals:**

 - Allow a full-screen view for comments
 - Provide a better commenting experience

![](https://user-images.githubusercontent.com/912252/63276854-23a1be00-c29c-11e9-8b9a-04051956c486.png)

**Colors**

- Background color -> .basicBackground
- Divider -> .divider
- Text (even title) -> .text
- Placeholder text -> .textPlaceholder
- Expand/Collapse icon -> .listIcon
- Send button active -> .primary or .primaryButtonBackground
- Send button inactive -> .listSmallIcon

emilylaguna:
Subtasks:

- [ ] Update the UI for the existing `ReplyTextView`
  - Remove border from text field
  - Update the comment field font
  - Replace the 'Reply' with the new image from the design
    - Get the correct 'reply' icon

- [ ] Animate the reply button
   - When textfield gains focus:
     - Transition the button in from the right
   - When textfield loses focus: 
      - Transition the button out from the left
   - When the button is active:
    - Change tint color to: .primaryButtonBackground
  - When the button is inactive:
    - Change tint color to: .listSmallIcon

- [ ] Add ability to go fullscreen while commenting
  - Add a button to "enter fullscreen" with the gridicon: `chevron-up`
  - Utilize the EditViewController to allow for fullscreen commenting
     - Subclass the EditViewController to change the UI
       - Override the Cancel button to exit fullscreen instead
       - Add a button to "exit fullscreen" with the gridicon: `chevron-down`

- [ ] Implement fullscreen commenting in the following locations:
  - `CommentViewController`
  - `NotificationDetailsViewController`
  - `ReaderCommentsViewController`

PR #12935, opened 2019-11-13
12918 Full Screen Commenting

OurBigAdventure:
#12918 

Fixed issue flagged by Swift Lint that was preventing a commit
Updated UI for ReplyTextView to match new design for current functionality
Added expand capability to ReplyTextView to allow for full screen editing

Testing:
Nothing to test at this time, more to come on this PR


WordPress/Classes/ViewRelated/Comments/CommentViewController.m
```diff
@@ -660,6 +659,24 @@ - (void)textView:(UITextView *)textView didTypeWord:(NSString *)word
     [self.suggestionsTableView showSuggestionsForWord:word];
 }
 
+- (void)updateUIForExpandedReply
+{
+    [[self view] bringSubviewToFront:[self replyTextView]];
+    [UIView animateWithDuration:0.5 animations:^{
+        [[self navigationController] setNavigationBarHidden:TRUE];
```
frosty: It's usual in objective-c to use `YES` and `NO` instead of `TRUE` and `FALSE`.

frosty: There's a `setNavigationBarHidden:animated:` method you can use instead of wrapping this in an animation block (same applies to other usages of this method).

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
```diff
@@ -104,6 +98,16 @@ import WordPressShared.WPStyleGuide
...
     }
 
+    @objc open func collapseReplyTextView() {
+        isExpanded = false
+        self.headerView.isHidden = true
+        UIView.animate(withDuration: 0.5, animations: {
+            self.expandButton.isHidden = false
+            self.replyButton.isHidden = false
+            self.refreshInterface()
+        }) { _ in
```
frosty: We generally prefer not to use trailing closure syntax for cases like this that have multiple blocks – I think it's often  clearer and more consistent to name the last parameter. 🙂

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
```diff
@@ -104,6 +98,16 @@ import WordPressShared.WPStyleGuide
         textView.replace(newRange, withText: replacementText)
     }
 
+    @objc open func collapseReplyTextView() {
+        isExpanded = false
+        self.headerView.isHidden = true
+        UIView.animate(withDuration: 0.5, animations: {
```
frosty: You're using `0.5` in quite a few places. It would be preferable to define this as a constant somewhere rather than using the same literal value in various places, as it makes it harder to typo it in one place and easier to update if you want to tweak the entire animation speed.

WordPress/Classes/ViewRelated/Comments/CommentViewController.m
OurBigAdventure: I recently read a post where `YES` and `NO` were explained to not be the same as `TRUE` and `FALSE` in some cases.  Because this isn't a comparison that argument isn't valid.  I can swap those out.

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
OurBigAdventure: I agree!  In these animations I may not need the completion block so it may be removed before the PR is complete. :)

WordPress/Classes/ViewRelated/Comments/CommentViewController.m
```diff
@@ -660,6 +659,30 @@ - (void)textView:(UITextView *)textView didTypeWord:(NSString *)word
     [self.suggestionsTableView showSuggestionsForWord:word];
 }
 
+- (void)updateUIForExpandedReply
+{
+    [[self view] bringSubviewToFront:[self replyTextView]];
+    [UIView animateWithDuration:self.replyTextView.animationDuration animations:^{
+        [[self navigationController] setNavigationBarHidden:YES];
```
frosty: You switched some other cases to use `setNavigationBarHidden:animated:` but not here (or in the following method) – not sure if you missed these or not?

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -138,6 +139,17 @@ - (void)viewDidLoad
     [self configureReplyTextView];
     [self configureSuggestionsTableView];
     [self configureKeyboardGestureRecognizer];
+
+    NSDictionary *views         = @{
+        @"tableView"        : self.tableView,
+        @"postHeader"       : self.postHeaderWrapper,
+        @"replyTextView"    : self.replyTextView
+    };
+    self.verticalConstraint = [NSLayoutConstraint constraintsWithVisualFormat:@"V:|[postHeader][tableView][replyTextView]"
```
frosty: Is there a reason this isn't part of the `configureViewConstraints` method?

WordPress/Classes/ViewRelated/Comments/CommentViewController.m
```diff
@@ -660,6 +659,30 @@ - (void)textView:(UITextView *)textView didTypeWord:(NSString *)word
     [self.suggestionsTableView showSuggestionsForWord:word];
 }
 
+- (void)updateUIForExpandedReply
+{
+    [[self view] bringSubviewToFront:[self replyTextView]];
```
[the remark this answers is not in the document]
OurBigAdventure: Not a problem, I can switch these

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
OurBigAdventure: I was experimenting with the constraint and forgot to move it after.  It will be relocated with the next push.

WordPress/Classes/ViewRelated/Comments/CommentViewController.m
OurBigAdventure: I did miss those, that display had the fewest issues so it didn't get as much attention.  This will be corrected in the next push.

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
```diff
@@ -181,19 +189,32 @@ import Gridicons
...
+    @IBAction func toggleTextViewDisplay(_ sender: UIButton) {
+        if isExpanded {
+            collapseTextView()
+        } else {
+            expandTextView()
         }
     }
 
-    @IBAction func collapseTextView(_ sender: UIButton) {
+    fileprivate func expandTextView() {
```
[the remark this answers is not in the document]
OurBigAdventure: of course, a little whitespace never killed anyone :)

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
```diff
@@ -84,6 +84,8 @@ import Gridicons
         return textView.isFirstResponder
     }
 
+    @objc open var animationDuration: Double = 0.5
```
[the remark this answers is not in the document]
OurBigAdventure: yes, you're right.

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
```diff
@@ -367,6 +372,12 @@ import Gridicons
     fileprivate var viewSpacing: CGFloat = 8.0
     fileprivate var expandButtonExpandedYConstraint: NSLayoutConstraint!
     fileprivate var replyButtonExpandedYConstraint: NSLayoutConstraint!
+    fileprivate var safeAreaHeight: CGFloat {
+        if #available(iOS 11.0, *), let window = UIApplication.shared.keyWindow {
```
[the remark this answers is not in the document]
OurBigAdventure: I noticed that as well, this was one of those Xcode things that wouldn't go away until I agreed.  I'll make sure Xcode and I come to an agreement before the next push. :)

WordPress/Classes/ViewRelated/Notifications/ReplyTextView/ReplyTextView.swift
```diff
@@ -213,6 +214,7 @@ import Gridicons
                          },
                          completion: { _ in
                             self.expandButton.setImage(Gridicon.iconOfType(.chevronDown), for: .normal)
+                            expandButton.accessibilityLabel = NSLocalizedString("Reduce text area", comment: "Describing button to reduce the reply text editable area.")
```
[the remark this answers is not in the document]
OurBigAdventure: I like that, I wasn't sure how to word the description.  Thanks!

Issue #11487, opened 2019-04-17
NoResults view controller VoiceOver: suggested improvements for root view, multiline descriptions, and first element read


designsimply:
Steps to reproduce:

1. Go to Settings > General > Accessibility > VoiceOver and toggle the setting on.
1. Go to WPiOS > My Sites > [any site] > Blog Posts > Search, enter text until you see no results.
1. Are the contents of the screen announced properly? Try accessing the things on the screen.
1. Go to Trashed. Repeat step 3.
1. Go to Scheduled. Repeat step 3.
1. Go to Reader > Saved Posts. Repeat step 3. (Note the bookmark icon is not read aloud.)
1. Go to Notifications > Comments and Likes. Repeat step 3.
1. Go to Reader > My Likes. Repeat step 3.

Result: when using VoiceOver, screens with no results are difficult to work with. ([2m2s](https://cloudup.com/ciWHIWb6FXI))

Note: testing this is easier when using a newer site without very much content.

<img src="https://user-images.githubusercontent.com/1119271/56323176-4d00e080-6128-11e9-8216-b0a7063cc56a.png" width="200" alt="noresults-blog-posts-search"> <img src="https://user-images.githubusercontent.com/1119271/56323174-4d00e080-6128-11e9-80f4-3f475f534d74.png" width="200" alt="noresults-reader-saved-posts"> <img src="https://user-images.githubusercontent.com/1119271/56323173-4c684a00-6128-11e9-9106-cd872fe9e59b.png" width="200" alt="noresults-notifications-comments">
<p style="clear:both;"><sup>Tested with WP Internal 12.2.0.20190412 on iPhone 6S iOS 12.2.</sup></p>

Suggested improvements noted by @etoledom:

* The root view should be a whole accessibility element to avoid VoiceOver from reading elements behind it. This element is used in many areas of the app and can be super confusing for a VoiceOver user.
* The description is not read properly by VoiceOver when it’s multiline.
* It would be great if the no results title is the first element VoiceOver reads when a new screen is accessed. (Not sure if possible)

<p style="clear:both;"><sub>(internal reference: p77Llu-bQ6-p2 h/t @etoledom)</sub></p>

designsimply:
Adding to Groundskeeping for review because it has a `[Pri] High` label.

shiki:
> It would be great if the no results title is the first element VoiceOver reads when a new screen is accessed

[...]

PR #12943, opened 2019-11-13
[VoiceOver] NoResultsView enhancements

shiki:
Refs #11487 and #12906. 

This changes the `NoResultsViewController` so that the full view is accessible instead of just the title and subtitle as suggested in #11487. This makes the “no results” message easily discoverable by VoiceOver users. They do not need to have their finger travel far to _find_ the message.

The title and subtitle are now also dictated together. 

Before | After 
--------|-------
IMG_1041       |       IMG_1044
IMG_1042  | IMG_1043

Some attempts at announcing the “no results” message were done to no great results. More information about that starting at https://github.com/wordpress-mobile/WordPress-iOS/issues/11487#issuecomment-551319114 if you're interested. 🙂 

## View Changes

To make this work, I had to modify `NoResults.storyboard`:

- Made `noResultsView` cover the whole Safe Area
- Moved the `noResultsView` alignment and margins to the main stack view

image

I tried to create commits for every single constraint change. I hope that helps. 

## Reader → Saved Posts

The `ReaderSavedPostsViewController` usage of `NoResultsVC` was causing issues. Even before this change, it seems that everything under `NoResultsVC.view` was inaccessible. The effect is that VoiceOver would dictate “empty list” instead of the desired “No results” dictation. 

It took me a while but it turned out that the fix was to move the routine for making `NoResultsVC` a child of `ReaderSavedPostsVC` from `viewDidAppear` to `viewDidLoad` (8c452000d864acaf1a18cae17a8e788affcbc651). With this change, the `tableView` is no longer an accessible element and all the intended views underneath `NoResultsVC` are accessible. 

## Outstanding Issues

As described in #11487, VoiceOver ignores the _bookmark icon_ when dictating the subtitle in Reader → Saved Posts' no-results view. I'm not quite sure how to fix this yet. I'll fix it in a separate PR. 

## Testing

Testing involves using VoiceOver on the places where `NoResultsViewController` is used. The following steps is a suggestion: 

1. Turn VoiceOver on. 
2. Navigate to an empty tab in Notifications and validate that the changes described above work correctly. 
3. Navigate to Site → Post List. Enter a string in the search bar that will result in an empty result. Validate the changes described above. 
4. Turn the Internet connection off.
5. Navigate back to an empty tab in Notifications. The tab should show an “Unable to load this page right now.” message. Validate the changes described above. 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderSavedPostsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderSavedPostsViewController.swift
@@ -328,7 +328,6 @@ private extension ReaderSavedPostsViewController {
     struct NoResultsText {
         static let noResultsTitle = NSLocalizedString("No Saved Posts", comment: "Message displayed in Reader Saved Posts view if a user hasn't yet saved any posts.")
         static let subtitleFormat = NSLocalizedString("Tap [bookmark-outline] to save a post to your list.", comment: "A hint displayed in the Saved Posts section of the Reader. The '[bookmark-outline]' placeholder will be replaced by an icon at runtime – please leave that string intact.")
-        static let accessibilityLabel = NSLocalizedString("No posts saved – yet! Tap the Save Post button to save a post to your list.", comment: "Alternative accessibility text displayed to Voiceover users on the Reader Saved Posts screen.")
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Reader/ReaderSavedPostsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderSavedPostsViewController.swift
@@ -323,7 +323,6 @@ private extension ReaderSavedPostsViewController {
         messageText.replace("[bookmark-outline]", with: icon)
 
         noResultsViewController.configure(title: NoResultsText.noResultsTitle, attributedSubtitle: messageText)
-        noResultsViewController.view.accessibilityLabel = NoResultsText.accessibilityLabel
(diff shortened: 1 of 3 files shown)


shiki:
@ScoutHarris Can I bother you with this one? 🙂

shiki:
Marking this as not ready for review. I found some “Empty List” dictation bugs that are caused by this.

shiki:
Fixed the issue and updated the PR description. ☝️ This is now ready for review again. 🙂

WordPress/Classes/ViewRelated/Reader/ReaderSavedPostsViewController.swift
```diff
@@ -328,7 +323,6 @@ private extension ReaderSavedPostsViewController {
         messageText.replace("[bookmark-outline]", with: icon)
 
         noResultsViewController.configure(title: NoResultsText.noResultsTitle, attributedSubtitle: messageText)
-        noResultsViewController.view.accessibilityLabel = NoResultsText.accessibilityLabel
```
ScoutHarris: It looks like the `NoResultsText.accessibilityLabel` can be removed.

shiki: Good find! I missed that. Removed in 344c7b0e1ff1fcc255a2663543c2377539b1183a. Will merge once the build is green. Thank you, @ScoutHarris! 🙏

PR #12794, opened 2019-10-28
[Stats] remove old code

ScoutHarris:
Ref #10975
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/193
WPUI PR: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/46

The main point if this PR is to remove the `WordpressComStatsiOS.framework` (which accounts for the massive change count). 

However, when I did so, there were several warnings about duplicate methods. Here's how I addressed them:

---
- warnings: 
  - instance method 'flashRowAtIndexPath:scrollPosition:flashLength:completion:' conflicts with same method from another category
  - instance method 'deselectSelectedRowWithAnimation:' conflicts with same method from another category
- Fix: Removed `UITableView+Extensions` from the Share extension, let it use the same methods from WordPressUI.

---
- warning: instance method 'trimNewlines' conflicts with same method from another category
- Fix: Removed `trimNewlines` from WPiOS, let it use the same method from WordPressKit.

---
- warnings: 
  - instance method 'initWithMinLineHeight:maxLineHeight:lineBreakMode:alignment:' conflicts with same method from another category
  - instance method 'initWithMinLineHeight:lineBreakMode:alignment:' conflicts with same method from another category
- Fix: Removed `NSMutableParagraphStyle+Helpers` from WPiOS, let it use the same methods from WordPressKit.

---
- warning: instance method 'setValueIfNotNil' conflicts with same method from another category
- Fix: Removed `NSMutableDictionary+Helpers` from WPiOS, let it use the same method from WordPressKit.

---
- warning: instance method 'md5' conflicts with same method from another category
- Fix: This method exists in WordPressUI and WordPressKit. I didn't want to remove functionality from either, so I renamed the WordPressUI method to `md5Hash`.

---
To test:
- Clean + build the app.
- Verify there are no errors or warnings regarding duplicate methods.
- Verify Stats works as expected.

---
Update release notes:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #12580, opened 2019-09-30
Do not automatically retry upload after 3 failures

leandroalonso:
Fixes #12369

# To test

## Scenario 1: offline publishing

While offline:
1. Create and publish a post
2. Check that both the toast and the post card says: "Post will be published next time..."
3. Restart the app
4. Check that both the toast and the post card says: "Post couldn't be published. We'll try again later"
5. Restart the app and check the same behavior as above
6. Restart again and check that the post failed and that the message "Couldn't perform operation. Post not published" is shown

## Scenario 2: offline drafting

While offline:
1. Create and publish a post
2. Check that both the toast and the post card says: "Draft will be uploaded next time..."
3. Restart the app
4. Check that both the toast and the post card says: "Post couldn't be drafted. We'll try again later"
5. Restart the app and check the same behavior as above
6. Restart again and check that the post failed and that the message "Couldn't perform operation. Post not drafted" is shown

## Scenario 3: restart after 3 failures

1. While offline, add a post and make sure to reach the message "Couldn't perform operation. Post not published" (you can do that by restarting the app at least 3 times)
2. Click retry.
3. While offline, this should mark the post to be auto-uploaded later
4. If online, the bost should be uploaded right away

Update release notes:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.
- [x] If it's feasible, I have added unit tests.

PR #12972, opened 2019-11-19
Fix warnings in localize.py

loremattei:
This PR fixes two warnings issued by `Scritps/localize.py`:
1. https://github.com/wordpress-mobile/WordPress-iOS/pull/12794 removed `WordPressComStatsiOS`, so this PR removes it from the list of the source frameworks.
2. https://github.com/wordpress-mobile/WordPress-iOS/pull/12580 added this helper file https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressTest/I18n.swift that triggers an error in `genstrings` because it passes a not literal string as the argument. Since this is only used as a helper for some tests, this PR makes sure it's not analyzed by `genstrings`.

_Note_: this is an ugly fix for an ugly script that's overdue for an update to new technologies. Since we'll have to throw this one away in order to be able to make use of some of the new features with strings, I opted for the easiest (though horrible) fix.
 
### To test:
1. Checkout a test branch from `develop` on your local machine (don't push it).
2. Run `Scripts/localize.py`. This will update `localizable.strings`.
3. Commit the changes in your test branch.
4. Cherry pick the updated script from this branch to your test branch.
5. Run the new `Scripts/localize.py` and verify that the warnings are gone and that the branch is clean.
6. Delete the test branch.

Update release notes:

### DIFF
--- a/Scripts/localize.py
+++ b/Scripts/localize.py
@@ -134,9 +134,9 @@ def localize(path, language, include_pods_and_frameworks):
 
     # TODO: This is super ugly, we have to come up with a better way of doing it
     if include_pods_and_frameworks:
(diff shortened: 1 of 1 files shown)


loremattei:
Thanks for checking this!

PR #12974, opened 2019-11-19
Fix crash in Post List when ghost is visible

shiki:
Refs #12671. Co-authored with @leandroalonso. 

## Findings

Leandro and @koke realized that the crash log in Sentry always has a consistent `10` number of rows in the message. More information at p5T066-136-p2#comment-3736.

```
2019-11-18 15:01:09.559181-0700 WordPress[62059:2960374] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Invalid update: invalid number of rows in section 0. The number of rows contained in an existing section after the update (10) must be equal to the number of rows contained in that section before the update (10), plus or minus the number of rows inserted or deleted from that section (1 inserted, 0 deleted) and plus or minus the number of rows moved into or out of that section (0 moved in, 0 moved out).'
```

We deduced that this can happen in this scenario: 

1. The app is built using Xcode 11 and running on iOS 13.1
2. The user has no cached data on the device
3. The user navigates to the Post List → Drafts
4. The user taps on the plus (+) button which adds a post in the Drafts list while the ghost view is still visible. 

We made a unit test that simulates this. The test crashes without the fix:

https://github.com/wordpress-mobile/WordPress-iOS/blob/c2a416e4343e990c008d4f0524260e955154ae73/WordPress/WordPressTest/ViewRelated/Post/Controllers/PostListViewControllerTests.swift#L52-L83

### Stats

- All events related to crashes because of changing rows in the last 4 months: 6377
- Events related to crashing because of the PostListViewController in the last 4 months: 4979

## Solution

To avoid having `UITableView+Ghost` and `WPTableViewHandler` interacting with a single `UITableView` and potentially interrupting each other's row number calculation, we opted to create another `UITableView` for the ghosting, [`ghostableTableView`](https://github.com/wordpress-mobile/WordPress-iOS/blob/c2a416e4343e990c008d4f0524260e955154ae73/WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift#L347-L359). 

## Testing

1. Delete the app to make sure there is no cached data.
2. Run the app using Xcode 11 and on an iOS 13.1
2. Log in and quickly navigate to the Post List → Drafts
4. While the ghost view is visible, tap on the plus (+) button 

Confirm that the app no longer crashes. 

## Reviewing 

## Submitter Checklist


WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
```diff
@@ -787,26 +803,21 @@ class AbstractPostListViewController: UIViewController,
...
-        tableView.isScrollEnabled = false
+        ghostableTableView.isHidden = false
         noResultsViewController.view.isHidden = true
     }
 
-    func stopGhost() {
-        tableView.removeGhostContent()
-        tableView.isScrollEnabled = true
+    final func stopGhost() {
+        ghostableTableView.isHidden = true
```
frosty: I'm not that familiar with the ghosting functionality, but should we also be removing the ghost content when the ghost tableview is hidden? Does the animation continue (albeit hidden) if we don't?

WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
```diff
@@ -342,6 +344,20 @@ class AbstractPostListViewController: UIViewController,
             ])
     }
 
+    func configureGhostableTableView() {
+        view.addSubview(ghostableTableView)
+        ghostableTableView.isHidden = true
+
+        ghostableTableView.translatesAutoresizingMaskIntoConstraints = false
+        ghostableTableView.widthAnchor.constraint(equalTo: tableView.widthAnchor, multiplier: 1).isActive = true
```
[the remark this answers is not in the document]
leandroalonso: Done in f9a0ba18d2953749d3fdb1e49ac7f04b905cf377 :)

WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift
leandroalonso: Very good question.

I did a quick research and wasn't able to find that answer.

So, I'm assuming that the animation would continue. In 2935c7c83b3f26bbf38db1e1c4189ae01337bed5 I start/stop the animation to avoid that.

leandroalonso:
@frosty Ready for another 👀

shiki:
Thank you, @frosty!

PR #12976, opened 2019-11-19
Fixes suggestions service endpoint path

frosty:
While testing another PR I noticed that the suggestions service (used in CommentViewController in Notifications) has a broken endpoint path, which results in a lot of 404 requests when you type a comment:

```
2019-11-19 19:25:13:862 WordPress[31045:3833400] [Rest API] ! The specified path was not found. Please visit https://developer.wordpress.com/docs/ for valid paths.
2019-11-19 19:25:14:106 WordPress[31045:3833400] [Rest API] ! The specified path was not found. Please visit https://developer.wordpress.com/docs/ for valid paths.
2019-11-19 19:25:14:545 WordPress[31045:3833400] [Rest API] ! The specified path was not found. Please visit https://developer.wordpress.com/docs/ for valid paths.
```

Ideally, we should write a remote for this service which uses the remote methods for constructive a v1.1 endpoint path, but as a quick fix I just updated it in place for now.

**To test:**

* Open a comment in Notifications
* Type an @ symbol and see the suggestions UI appear

Screenshot 2019-11-19 at 19 52 08

Update release notes:

### DIFF
--- a/WordPress/Classes/Services/SuggestionService.m
+++ b/WordPress/Classes/Services/SuggestionService.m
@@ -59,7 +59,7 @@ NSString * const SuggestionListUpdatedNotification = @"SuggestionListUpdatedNoti
     // add this siteID to currently being requested list
     [self.siteIDsCurrentlyBeingRequested addObject:siteID];
     
-    NSString *suggestPath = @"v1.1/users/suggest";
+    NSString *suggestPath = @"rest/v1.1/users/suggest";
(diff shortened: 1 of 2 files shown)


frosty:
Thanks @leandroalonso – my main concern right now was just fixing the bug :) I don't even know if this is supposed to work for mentions! 

I just pushed up one more small tweak where I noticed we were creating a new suggestions service each time – this would mean it wasn't using the shared cache and would always return no suggestions.

frosty:
Thanks!

PR #12983, opened 2019-11-20
[Gutenberg] Image size improvements

etoledom:
Fixes n1 from https://github.com/wordpress-mobile/gutenberg-mobile/issues/1593

In this PR, we avoid using the PhotonAPI helper, which was aggressively stripping the image size properties from the URL.

The `downloadImage` method of `RCTImageURLLoader` is used by the JS function `Image.getSize()` to retrieve the original image and calculate its size.

I also took this opportunity to add a simple in-memory image cache to avoid fetching the same image multiple times.

To test:
- On a WPCom site:
- Add an image block.
- Add an image from any source.
- Change the image size to thumbnail.
- Check that the size of the image visually changes.

Update release notes:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
@@ -51,7 +51,7 @@ class EditorMediaUtility {
         return downloadImage(from: url, size: size, scale: scale, post: post, success: success, onFailure: failure)
     }
 
-    func downloadImage(from url: URL, size requestSize: CGSize, scale: CGFloat, post: AbstractPost, success: @escaping (UIImage) -> Void, onFailure failure: @escaping (Error) -> Void) -> ImageDownloader.Task {
+    func downloadImage(from url: URL, size requestSize: CGSize, scale: CGFloat, post: AbstractPost, allowPhotonAPI: Bool = true, success: @escaping (UIImage) -> Void, onFailure failure: @escaping (Error) -> Void) -> ImageDownloader.Task {
         var requestURL = url
         let imageMaxDimension = max(requestSize.width, requestSize.height)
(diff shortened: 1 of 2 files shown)


mchowning:
>I also took this opportunity to add a simple in-memory image cache to avoid fetching the same image multiple times.

👍  I have caching turned on for Android as well.

WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
```diff
@@ -70,10 +70,16 @@ class EditorMediaUtility {
             size.width = size.width * scale
             requestURL = WPImageURLHelper.imageURLWithSize(size, forImageURL: requestURL)
             request = URLRequest(url: requestURL)
-        } else {
+        } else if allowPhotonAPI {
```
SergioEstevao: @etoledom Can you clarify when allowPhotonAPI is set to true/false is there a related Gutenberg-Mobile PR where I can see the use of this argument?

etoledom: The previous behavior was to always fall back to the Photon API. With this change and setting `allowPhotonAPI = true` as default, all other instances continue working in the same way.

This PR uses `allowPhotonAPI = false` to fetch the image from the original URL directly and avoid the Photon resizing that affects the RN `Image.getSize` method.

etoledom: I'm giving this a second look trying to avoid introducing `allowPhotonAPI`.

etoledom: So, playing with the image scale we are able to get the desired result using the Photon Helper.
I believe it's ready for another 👀 @SergioEstevao 🙏

etoledom:
Thank you!

PR #12988, opened 2019-11-21
[Stats Refresh] Remove block loading Feature Flag

danielebogo:
Refs. #11852

This PR removes the Feature Flag for the block loading

## To test:
- Run this branch and openStats from selected site (ex. en.blog)
- Test Insights but specially DWMY and its Stats details
- Test Post stats
- Everything should work as before in **13.7**

PR submission checklist:


ScoutHarris:
Hey @danielebogo . I noticed that if you go to a Period view, access details for any stat, and return to the Period view, the chart `Likes` query never finishes. 

Screen Shot 2019-11-21 at 1 35 03 PM

However, I don't think this is new since it appears `develop` exhibits the same behavior.

danielebogo:
@ScoutHarris it's ready for another look! I also removed `viewIsVisible` in the table refresh because I don't think is needed anymore with the block loading. But let me know if that is required and I will revert that change!

danielebogo:
Thanks @ScoutHarris !!!

PR #12989, opened 2019-11-21
Fix crash with UIDocumentMenuViewController in WebKitViewController

### LINKED ISSUE #12882: NSGenericException: Your application has presented a UIDocumentMenuViewController (<UIDocumentMenuViewController: 0x1...
Sentry Issue: [WORDPRESS-IOS-5PZ](https://sentry.io/share/issue/95876ad2315a4d238aa3585a2e8d615a/)

```
NSGenericException: Your application has presented a UIDocumentMenuViewController (<UIDocumentMenuViewController: 0x111838e60>). In its current trait environment, the modalPresentationStyle of a UIDocumentMenuViewController with this style is UIModalPresentationPopover. You must provide location information for this popover through the view controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem.  If this information is not known when you present the view controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.
...
(11 additional frame(s) were not displayed)
```

**Update:** 

Steps to reproduce(Example):
1. Find a blog post in the reader that contains a URL(use google.com)
2. Click on the post and click on the url
3. From that Webview, navigate to wordpress.com and sign in
4. Try and create a new post
5. Add an image block
6. Click Upload to try an upload an image

yaelirub:
Fixes #12882 

In iOS 13 on iPhone devices, when displaying UIDocumentMenuViewController in a modally presented view (in this case web view) we need to explicitly define the location for the presentation of the popover or otherwise the app crashes. 

this PR fixes the crash by setting the WebKitViewController as the UIPopoverPresentationControllerDelegate when the terms mentioned above apply. 

It also adds a tap recognizer to WebKitViewController to be simultaneously recognized (since the webview handles taps too) so that we can display the popover from the tap location 

To test: (please test on iOS < 13*  as well to see the old behavior preserved):
  
1. Find a blog post in the reader that contains a URL(use google.com)
2. Click on the post and click on the url
3. From that Webview, navigate to wordpress.com and sign in
4. Try and create a new post
5. Add an image block
6. Click Upload to try an upload an image

In iOS 13 the popup should show up in the location of the tap 
IMG_5145 2

iOS 11.0
Simulator Screen Shot - iPhone 8 - 2019-11-21 at 15 22 34

PR submission checklist:


leandroalonso:
Reviewing 👀

WordPress/Classes/Extensions/UINavigationController+Helpers.swift
```diff
@@ -33,4 +33,17 @@ extension UINavigationController {
...
     }
+
+    /// Fixes a crash in iOS 13 (#12882) where presenting a UIDocumentMenuViewController in a webView
+    /// doesn't rautomattically ecognize the location for presenting the menu hence the crash.
+    /// The warning is probably a bug on iOS or on WebKit since replacing with UIDocumentPickerViewController doesn't prevent the crash.
+    ///
+    @objc override open func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        if #available(iOS 13, *), UIDevice.current.userInterfaceIdiom == .phone,
+            let webKitVC = topViewController as? WebKitViewController,
+            viewControllerToPresent is UIDocumentMenuViewController {
```
leandroalonso: Do we really need this `is UIDocumentMenuViewController`? I tried without it and the result was the same.

That way we can get rid of the warning.

WordPress/Classes/Extensions/UINavigationController+Helpers.swift
```diff
@@ -33,4 +33,17 @@ extension UINavigationController {
             }
         }
     }
+
+    /// Fixes a crash in iOS 13 (#12882) where presenting a UIDocumentMenuViewController in a webView
+    /// doesn't rautomattically ecognize the location for presenting the menu hence the crash.
+    /// The warning is probably a bug on iOS or on WebKit since replacing with UIDocumentPickerViewController doesn't prevent the crash.
```
leandroalonso: This is an iOS 13 bug. I hope this will be fixed by Apple in the near future, then we can remove that override. Do you think that would be useful information to have here?

WordPress/Classes/Extensions/UINavigationController+Helpers.swift
yaelirub: I guess I was concerned about presenting anything else that we might not want to present from the location of the tap. Since I am checking for WebKitVC at the top of the presentation stack and I don't believe we would want to display anything other than a popover on top of that. I removed the check for UIDocumentMenuViewController because the warning is annoying and it makes sense

WordPress/Classes/Extensions/UINavigationController+Helpers.swift
yaelirub: Ok. updated the comments

yaelirub:
@leandroalonso , updated

PR #12991, opened 2019-11-22
Update projects to swift 5

SergioEstevao:
Update projects to use Swift 5 has the baseline for Swift

To test:
 - Check that project builds and run correctly.

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
```diff
@@ -680,10 +680,10 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
         // Now that we have the image, create an aspect ratio constraint for
         // the featuredImageView
         let ratio = size.height / size.width
-        let constraint = NSLayoutConstraint(item: featuredImageView,
+        let constraint = NSLayoutConstraint(item: featuredImageView!,
```
koke: I don't understand these new force unwraps and don't see anything relevant in the [migration guide](https://swift.org/migration-guide-swift5/). `featuredImageView` is already defined as `CachedAnimatedImageView!`

WordPress/WordPressTest/PostTests.swift
```diff
@@ -301,11 +301,11 @@ class PostTests: XCTestCase {
         XCTAssertEqual(revision.statusForDisplay(), NSLocalizedString("Local changes", comment: "Local"))
 
         revision.status = .scheduled
-        let scheduledStatusDisplay = "\(Post.title(for: .scheduled))"
+        _ = "\(Post.title(for: .scheduled))"
         XCTAssertEqual(revision.statusForDisplay(), local)
 
         revision.status = .trash
-        let trashStatusDisplay = "\(Post.title(for: .trash))"
+        _ = "\(Post.title(for: .trash))"
```
koke: This doesn't look right? These two variables seem unused so we should delete those lines instead?

WordPress/Classes/ViewRelated/Me/PrivacySettingsViewController.swift
```diff
@@ -146,6 +146,7 @@ private class InfoCell: WPTableViewCellDefault {
         super.layoutSubviews()
         guard var imageFrame = imageView?.frame,
             let textLabel = textLabel,
+            let textLabelFont = textLabel.font,
```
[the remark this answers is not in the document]
SergioEstevao: The code further down was complaining that this was an optional (rightly so)so couldn't use it to set an attribute with it.

WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
SergioEstevao: I was also surprised by this one... if you remove the ! you will see a warning showing up.

WordPress/WordPressTest/PostTests.swift
SergioEstevao: I will try to remove it, but I think this call could have some expected side effect that the test is expecting?

koke: I think it might have been an oversight: https://github.com/wordpress-mobile/WordPress-iOS/pull/11705/files#diff-2aeb6e69ee88f42fb3a015b6bf5780edL305

WordPress/Classes/ViewRelated/Me/PrivacySettingsViewController.swift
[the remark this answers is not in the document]
koke: Ah, I missed that this was a `guard let` and not a simple `let` 🤦‍♂

Issue #12338, opened 2019-08-19
Crash on startup after updating iOS


iamgabrielma:
Reported on 2284689-zen

```
After updating WordPress for iOS, the app can no longer be started.
If you delete it and install it again, it will be forcibly terminated after 
login, and it will no longer be able to start.
```

Similar cases in the forums:

- https://ios.forums.wordpress.org/topic/wordpress-app-crashing
- https://ios.forums.wordpress.org/topic/app-crashing-after-update

This was reproduced in iPhone X with version 13.0, and iPhone 7. We could not reproduce it on 5s (iOS 12.4) or 6s . Here's a video of what @chaitanyamsv is experiencing on his phone:

https://cld.wthms.co/8FL1PD
<sup>Tested with iPhone 7 on iOS 12.2 right after upgrading to WPiOS 13.0.</sup>

(EDIT) Crash details: https://github.com/wordpress-mobile/WordPress-iOS/issues/12338#issuecomment-535896537

thehenrybyrd:
Seems to be the same crash as https://github.com/wordpress-mobile/WordPress-iOS/issues/12331

etoledom:
So, https://github.com/wordpress-mobile/WordPress-iOS/issues/12331 turned out to affect only apps running on Japanese.

This issue in particular seems to be related to this other ticket: https://github.com/wordpress-mobile/WordPress-iOS/issues/12028

aleone89:
Another report mentioned at hc-6166545

designsimply:
I tried testing with WPiOS 13.0 (app store version) on iPhone 6S iOS 12.4 and the app did not crash on startup in my test, so I'm looking for any details about how to trigger the crash that I might be missing when testing from my end.

designsimply:
Noting the details from https://ios.forums.wordpress.org/topic/app-crashing-after-update

<blockquote>Recently I Updated My WordPress App to Latest Version in iPhone 7 and After that Whenever i Open App it Just Crash Instantly. Like - I Clicked on WP App>> It Shows Dashboard for 0.5 Seconds>> Blackout for 0.5 Second and Closed without any Notification. I Tried to Restart my Device but no Luck.</blockquote>

designsimply:
Noting some details from 6166545-hc

>New iphone upgrade - WP App initiation just produces frozen blue screen with W in the middle. Computer is fine but I need to have both working.

Followed by:

[...]

PR #12995, opened 2019-11-22
Track Core Data startup errors to Sentry

leandroalonso:
Refs #12338

In https://github.com/wordpress-mobile/WordPress-iOS/pull/12569 we changed the Core Data startup error to a proper exception. However, we still unable to have more details about this issue. Mainly because:

1. Xcode reports don't give much information. They just state that an objc exception was thrown but no details;
2. No data is being sent to Sentry

This aims to fix number 2.

## To test

Add those lines in `ContextManager.m` right before `return _persistentStoreCoordinator;` inside `(NSPersistentStoreCoordinator *)persistentStoreCoordinator`:

```swift
NSError *errMsg = nil;
errMsg = [NSError errorWithDomain:@"aaa" code:1001 userInfo:@{ NSLocalizedDescriptionKey:@"aaa" }];
[errors addObject:errMsg];
errMsg = [NSError errorWithDomain:@"bbb" code:1001 userInfo:@{ NSLocalizedDescriptionKey:@"bbb" }];
[errors addObject:errMsg];
errMsg = [NSError errorWithDomain:@"ccc" code:1001 userInfo:@{ NSLocalizedDescriptionKey:@"ccc" }];
[errors addObject:errMsg];
[SentryStartupEvent.sharedInstance sendWithTitle:@"Test, ignore" errors:errors];

abort();
```

Run the app. It will crash.

After that go to [this link](https://sentry.io/organizations/a8c/events/?project=1438083) and make sure you can see the event in the list of events.

All the errors that happen until Core Data will be tracked and sent as breadcrumbs to Sentry. This would give us more insights into what's happening.

## Additional information

As of today, our tracking code needs a `CrashLoggingDataProvider` and we provide a `WPCrashLoggingProvider`. However, for this code to be initialized, we need Core Data. [You can see here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/Logging/WPCrashLoggingProvider.swift#L18).

One would assume that by fixing that we'd be able to track startup exceptions and/or crashes. Unfortunately not.

Sentry is not able to submit a [crash report that occurs on launch](https://docs.sentry.io/clients/cocoa/#testing-a-crash). My aim with that PR is to do that manually.

PR submission checklist:

- [x] I have considered adding unit tests where possible — since those are temporary changes to gain more info on the issue, I haven't spent time writing tests. If you think this is highly required please let me know.

### DIFF
--- a/WordPress/Classes/Utility/ContextManager.m
+++ b/WordPress/Classes/Utility/ContextManager.m
@@ -239,6 +239,7 @@ static ContextManager *_override;
     };
 
     NSError *error = nil;
+    NSMutableArray *errors = [NSMutableArray array];
     _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc]
                                    initWithManagedObjectModel:[self managedObjectModel]];
 
@@ -249,9 +250,8 @@ static ContextManager *_override;
                                                            error:&error]) {
         DDLogError(@"Error opening the database. %@\nDeleting the file and trying again", error);
 
-        if (error != nil) {
-            [SentryStartupEvent.sharedInstance sendWithTitle:@"Error opening the database." error:error];
-        }
+        [errors addObject:error];
+        error = nil;
 
         _migrationFailed = YES;
         
@@ -261,14 +261,16 @@ static ContextManager *_override;
                                                  error:&error];
 
         if (error != nil) {
-            [SentryStartupEvent.sharedInstance sendWithTitle:@"Error backing up the database" error:error];
+            [errors addObject:error];
+            error = nil;
         }
 
         // delete the sqlite file and try again
         [[NSFileManager defaultManager] removeItemAtPath:storeURL.path error:&error];
 
         if (error != nil) {
-            [SentryStartupEvent.sharedInstance sendWithTitle:@"Error removing failed migrated database" error:error];
+            [errors addObject:error];
+            error = nil;
         }
 
         if (![_persistentStoreCoordinator addPersistentStoreWithType:NSSQLiteStoreType
@@ -277,7 +279,8 @@ static ContextManager *_override;
                                                              options:nil
                                                                error:&error]) {
 
-            [SentryStartupEvent.sharedInstance sendWithTitle:@"Can't initialize Core Data stack" error:error];
+            [errors addObject:error];
+            [SentryStartupEvent.sharedInstance sendWithTitle:@"Can't initialize Core Data stack" errors:errors];
 
             @throw [NSException exceptionWithName:@"Can't initialize Core Data stack"
                                            reason:[error localizedDescription]
--- a/WordPress/Classes/Utility/ContextManager.m
+++ b/WordPress/Classes/Utility/ContextManager.m
@@ -4,6 +4,8 @@
 #import "WordPress-Swift.h"
 @import WordPressShared.WPAnalytics;
 
+#define SentryStartupEventAddError(event, error) [event addError:error file:__FILE__ function:__FUNCTION__ line:__LINE__]
+
 // MARK: - Static Variables
 //
 static ContextManager *_instance;
@@ -239,7 +241,7 @@ static ContextManager *_override;
     };
 
     NSError *error = nil;
-    NSMutableArray *errors = [NSMutableArray array];
+    SentryStartupEvent *startupEvent = [SentryStartupEvent new];
     _persistentStoreCoordinator = [[NSPersistentStoreCoordinator alloc]
                                    initWithManagedObjectModel:[self managedObjectModel]];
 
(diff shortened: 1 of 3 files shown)


leandroalonso:
Maybe it would be good to target the `13.7` version, so we can have this ASAP.

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -239,6 +239,7 @@ - (NSPersistentStoreCoordinator *)persistentStoreCoordinator
     };
 
     NSError *error = nil;
+    NSMutableArray *errors = [NSMutableArray array];
```
koke: Before this, the call to `migrateDataModelsIfNecessary` might try to do a migration. We should also log any error resulting from that attempt.

koke:
I played around with the code a bit, and maybe the approach in a2b66ef243e4a0ecaf148981d4211318eceeefbf would give a better output?

Screen Shot 2019-11-25 at 12 46 14

leandroalonso:
@koke This looks really good! Thank you for that, I cherry-picked your commit to this PR. :)

I was thinking, do you think we should target the current frozen branch or let's just keep it to `develop`?

koke:
It sounds like a good idea to ship it earlier. The code seems harmless and it would get us information sooner to try to fix this

leandroalonso:
@koke I change the target to `release/13.7`.

leandroalonso:
> The majority of the Sentry-specific code lives in Tracks in order to be useful across multiple projects. I think this class might make sense there, since tracking fatal startup issues could be valuable in any project.

You're right. My only concern here is that this class is completely coupled with this specific error we're having. The class doesn't work out of the box, it needs some manual changes as you can see in `ContextManager.m`. In that case, I'm not so sure it would be that valuable. :(

> There are some similar API methods within Tracks that I think could use a similar signature, perhaps with a `waitUntilCompleted` flag?

This one is a little bit more tricky. In order for us to have the `Client` working, we would need to [start](https://github.com/Automattic/Automattic-Tracks-iOS/blob/a6ac334bc2ffb0c4d66b2f85f1f989d93f1fc0bc/Automattic-Tracks-iOS/CrashLogging.swift#L31) the `CrashLogging`.  In `WordPressAppDelegate.swift` we do that by providing a `WPCrashLoggingProvider` — however, the property `currentUser` relies on Core Data, and if we use that the app will crash before being able to send the tracking information. :(

We probably could change that so the `WPCrashLoggingProvider` would be able to start without relying on the database but this would require a bigger change (that I fear might not be that simple to address a frozen branch).

I really thought about this — as you said — as a quick one-off thing that could give us some feedback on what's happening. Once we have enough data we'll probably remove this code.

Also, this works with the premise that a Core Data crash will happen and our normal Track tool would never be initialized.

Let me know what you think about all that @jkmassel. :)

jkmassel:
> the property currentUser relies on Core Data, and if we use that the app will crash before being able to send the tracking information

This is _mostly_ true, but we can actually initialize the crash logging system without that data – `currentUser` is optional for exactly this kind of case (we don't know which user it is, or a crash happens before the user is logged in). Because of that, we should be able to initialize the `dataProvider` without a user. Realistically, we could start the crash logging subsystem at the top of on `applicationWillFinishLaunching` and just call `setNeedsDataRefresh` once the core data stack is initialized in order to populate the user field.

If this is something we know we'll tear out once we fix immediate issues, I think it's ok to go in with a tightly coupled implementation, but if it's designed to be long-lived, I'd rather work these issues out early if we can – the only thing I think is a hard blocker is that we need to respect the user's crash logging opt-out, if present. The `dataProvider` has this information, and because it's stored in `NSUserDefaults`, it should be safe to retrieve at launch.

leandroalonso:
@jkmassel Thanks for all that information. :)

> (...) just call `setNeedsDataRefresh` once the core data stack is initialized in order to populate the user field.

This is good to know. I guess we can create an issue to tackle that, what do you think?

> If this is something we know we'll tear out once we fix immediate issues, I think it's ok to go in with a tightly coupled implementation, but if it's designed to be long-lived (which I think it should be – I can see this catching additional issues for us if we introduce new bugs down the road), I'd rather work these issues out early if we can – the only thing I think is a hard blocker is that we need to respect the user's crash logging opt-out, if present. The `dataProvider` has this information, and because it's stored in `NSUserDefaults`, it should be safe to retrieve at launch.

I made this change, now the tracking is only performed if the user has not opt-out. Ideally, the solution should be long-lived and generic — this one, unfortunately, is not. I'll keep an eye on it and I'll remove this code once we have enough data.

Just to give you more context, I'm targeting specifically the issue https://github.com/wordpress-mobile/WordPress-iOS/issues/12338 which has been happening for some time and we haven't much information yet.

leandroalonso:
I'm changing this to `develop` after talking to @jkmassel — since there's not much time left in this sprint and we don't want any trouble. :P

leandroalonso:
@jkmassel Absolutely! To be honest, once we have enough data we can remove it. Given the next release would be in 2 weeks, 2 months seems pretty reasonable. What do you think? I will create a reminder for me.

Issue #12997, opened 2019-11-23
Adding Tenor support


scihant:
In a recent PR ( #12769 ), Giphy was removed as a user-facing Gif import option, and their SDK was removed from the dependencies of the project.

This task covers integration of the Google's Tenor library to re-enabling the option to import Gif's from a third party Gif library.

Tenor doesn't offer any library for iOS, so we'll have to write our own wrapper and integrate it while taking advantage of reusing most of the Giphy components that are left in the app.

**Subtasks**

- Adding a wrapper for Tenor API #12999

PR #12998, opened 2019-11-23
Tenor integration

scihant:
This PR solves the issue of adding Gif support using Google's Tenor API ( #12997 )

I've created a simple wrapper to replace the Giphy object in the Giphy's old SDK. Since the response we get from the Tenor's API has a completely different structure, and there isn't 1-to-1 mapping between all the fields, I had to select suitable replacements among the fields that are available in the Tenor's API response. They might not be the best choices yet, that's why I haven't simplified the TenorMedia class further. I left some room for some experimentation if needed.

I've tried to use the existing Giphy related classes as much as possible. Therefore, there's a a lot of renaming and some relocating involved. That's why the PR is a bit long.

There's some force unwrapping and some unnecessary casting/bridging I'm gonna fix, but I wanted to get some feedback before doing so.

**To test:**

- Create an ApiKey by following the instructions in the [documentation](https://tenor.com/gifapi/documentation#quickstart).
- Add the key into the credentials file `.configure-files/wpcom_app_credentials` 

```
WPCOM_APP_ID=....
WPCOM_APP_SECRET=....
TENOR_APP_ID={The key you've just created}
```

- Rake dependencies
- Build and run, check it builds okay
- In the Media Library, tap + and select Tenor from the available options
- In Aztec, tap the + and then ... on the format bar and and select Tenor from the available options


scihant:
Looks like the tests failed because I'm using a newer version of XCode (11.2), which compiles with Swift 5.1.2. I had to update the Zendesk SDK to 3.0.3 (which adds support for Swift 5.1.2) to make the project compile in my mac, which caused the tests to fail here.

scihant:
@leandroalonso I can split the PR, but you won't be able to run anything in most of them because there won't be anything to test in the UI unless all tasks are finished. But you can review the code of course.

I'll keep the Zendesk change in my machine only.

scihant:
@leandroalonso I've opened up a smaller PR for a subtask #12999 .
I think you can close this one.

scihant:
Will be separated into  multiple PRs

leandroalonso:
@scihant Awesome! There's no problem if there isn't anything UI-related to test, don't worry. :) 

I will close this one.

PR #12999, opened 2019-11-23
Add a wrapper for Tenor API

scihant:
This PR is the first part of solving the issue of adding Gif support using Google's Tenor API ( #12997 )

I've created a simple wrapper to replace the Giphy object in the Giphy's old SDK. Since the response we get from the Tenor's API has a completely different structure and there isn't 1-to-1 mapping between all the fields, I had to select suitable replacements among the fields that are available in the Tenor's API response. They might not be the best choices yet, that's why I haven't simplified the TenorMedia class further. I left some room for some experimentation  when the feature becomes testable in the UI of the app.

This PR doesn't have any user-visible changes

**To test:**

- Create an ApiKey by following the instructions in the documentation.
- Add the key into the credentials file .configure-files/wpcom_app_credentials

WPCOM_APP_ID=....
WPCOM_APP_SECRET=....
TENOR_APP_ID={The key you've just created}

- Rake dependencies
- Build and run, check it builds okay


leandroalonso:
@scihant Thanks for opening this new PR.

I gave a glimpse of the code and it looks good, great work! I can start reviewing it right away but I think we can split it into even smaller PRs. Here's what I thought:

* One task related to the Client/Response
* Another task related to Pageable
* Another task for the Credential (including it into the scripts and so on)

This way I can focus on each part of your solution and we can discuss each different component of your solution in a single PR.

What do you think?

scihant:
Hey @leandroalonso 

Unlike my previous PR, which was really big, this one contains only a single, self-contained component of the task we can discuss. Think of it as a tiny network layer. It's the foundation on which the rest of the task will be built. I developed these parts together, and I believe that it's a good idea to review it as a whole. 

This way, you have the option to actually perform a search and look at the results in code at least. Remove even a single piece, and you won't be able to do that.

I could split it as you've suggested, but I think it would serve no other purpose than making things too granular and causing more merge commits.

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
```diff
@@ -0,0 +1,166 @@
...
+    enum MediaCodingKeys: String, CodingKey, CaseIterable {
+        case nanogif
+        case tinygif
+        case mediumgif
+        case gif
+    }
+
+    let id: String
+    let created: Date
+    let itemurl: String
```
leandroalonso: Do you think we can get rid of `created` and `itemurl`?

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
```diff
@@ -0,0 +1,166 @@
...
+
+        id = try container.decode(String.self, forKey: CodingKeys.id)
+
+        let createdDate = try container.decode(Float.self, forKey: CodingKeys.created)
+        created = Date(timeIntervalSince1970: Double(createdDate))
+
+        itemurl = try container.decode(String.self, forKey: CodingKeys.itemurl)
+        title = try container.decode(String.self, forKey: CodingKeys.title)
+
+        var mediaContainer = try container.nestedUnkeyedContainer(forKey: CodingKeys.media)
```
leandroalonso: Do you think we can move this to a method or add a comment to state what this logic is doing? In this case, we're iterating over all available media related to this gif, right?

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
```diff
@@ -0,0 +1,166 @@
...
+    var largeGif: TenorGif {
+        return gifs[.gif]!
+    }
+
+    var previewURL: URL {
+        return NSURL(string: previewGif.url)! as URL
+    }
+
+    var staticThumbnailURL: URL {
+        return NSURL(string: previewGif.preview)! as URL
```
leandroalonso: I would try to avoid using force unwraps. Can we make these properties Optionals?

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
```diff
@@ -0,0 +1,166 @@
...
+
+        return Int32(id) ?? 0
+    }
+
+    private func imageURL(with size: CGSize) -> URL {
+        return size == .zero ? previewURL : staticThumbnailURL
+    }
+
+    func cancelImageRequest(_ requestID: WPMediaRequestID) {
+        // Can't be canceled
```
leandroalonso: I guess this is something we can improve, to cancel the request of pending media in those cases:

* When the user changes the search query and there are media still being loaded
* When the user leaves the gif ViewController

What do you think?

We can add as a pending subtask in the Master Thread and work on that later.

WordPress/WordPressTest/Tenor/TenorTests.swift
```diff
@@ -0,0 +1,30 @@
...
+        XCTAssertNotNil(media.gifs[.nanogif], "Nano gif should exist")
+        XCTAssertNotNil(media.gifs[.tinygif], "Tiny gif should exist")
+        XCTAssertNotNil(media.gifs[.mediumgif], "Medium gif should exist")
+        XCTAssertNotNil(media.gifs[.gif], "Normal gif should exist")
+
+        let gif = media.gifs[.gif]!
+
+        XCTAssertTrue(gif.dims[0] > 0 && gif.dims[1] > 0, "Gifs should have non-zero dimensions")
+        XCTAssertNotNil(gif.url, "Gifs should have a url")
+        XCTAssertNotNil(gif.preview, "Gifs should have a preview url")
```
leandroalonso: There's a lot of things being tested here. :)

What do you think about splitting those checks in different tests with better descriptions?

We can break it into distinct tests that check for:

* The properties (`title`, `id`)
* The `gifs` property and it's childs: `dims`, `url`, `preview`
* Maybe the `previewGif`, `largeGif`, `previewURL` and `staticThumbnailURL would make good tests too

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
```diff
@@ -0,0 +1,166 @@
...
+
+// MARK: - Helpers needed by WPMediaAsset conformance
+
+extension TenorMedia {
+    var previewGif: TenorGif {
+        return gifs[.tinygif]!
+    }
+
+    var largeGif: TenorGif {
+        return gifs[.gif]!
```
leandroalonso: `gifs[.gif]` sounds a little bit odd. Maybe renaming `gifs` to `media` would be a good change? `media[.gif]` sounds better IMHO.

WordPress/Classes/ViewRelated/Media/Tenor/TenorClient.swift
```diff
@@ -0,0 +1,44 @@
...
+    case networkError
+    case wrongDataFormat
+    case wrongUrl
+}
+
+class TenorClient {
+    private let endPoint = "https://api.tenor.com/v1/search"
+    private let session = URLSession(configuration: .default)
+
+    // https://api.tenor.com/v1/search?q=excited&key=KEY&limit=5&pos=0
```
leandroalonso: I guess we can remove this comment.

WordPress/WordPressTest/Tenor/TenorTests.swift
```diff
@@ -27,4 +27,12 @@ class TenorTests: XCTestCase {
         XCTAssertNotNil(gif.url, "Gifs should have a url")
         XCTAssertNotNil(gif.preview, "Gifs should have a preview url")
     }
+
+    func testTonerResponsePaging() {
```
leandroalonso: We can rename this file to `TenorResponseTests` and create a new `TenorPageableTests` with this one.

WordPress/Classes/ViewRelated/Media/Tenor/TenorClient.swift
```diff
@@ -0,0 +1,44 @@
...
+
+typealias TenorResponseBlock = (Swift.Result<TenorResponse, Error>) -> ()
+
+enum TenorError: Error {
+    case networkError
+    case wrongDataFormat
+    case wrongUrl
+}
+
+class TenorClient {
```
leandroalonso: Since we just created this client this is a good choice for some fresh unit tests! We have OHHTTPStubs, what do you think of adding some tests to check the behavior of this client? In the cases of:

- Wrong/unavailable response
- Correct response
- Malformed URL

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
scihant: `created` is used for `WPMediaAsset` conformance, but `itemurl` seems to be a leftover. I'll remove it.

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
scihant: Yes. Most of this is boilerplate code well known to people familiar with the Codable protocol. The common practice here is to make the decoding in the init(from:) method (unless it is very complicated) because it is an initializer. I think adding a comment there would be a better solution if it isn't self-explanatory enough. What do you think?

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
scihant: That's how it was in the original implementation. Therefore I haven't looked at how exactly that method is used for cancellation in other contexts yet, assuming that it is actually used somewhere.

As you've said, I can work on that later, when we have a working version that can be debugged to see how things fit together more easily.

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
scihant: Yes, you're right, but having a field name media in a class that's also called a media would seem odd too.  I'll see if I can come up with a better name. In the final version, I'll have selected which two versions we need, and then I'll remove those fields and simplify the decoding logic.

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
scihant: Yes, that what it's doing. I'll try to make it easier to understand.

WordPress/Classes/ViewRelated/Media/Tenor/TenorResponse.swift
scihant: You're right. In normal circumstances these should be safe unwraps, because each item in the response should always have these fields. But we shouldn't rely on the integrity of the data coming from a third party, therefore eliminating the IUOs would be the best course of action. The optionals will propagate deeper into the code, but better safe than sorry.

PR #13005, opened 2019-11-26
Fix Google 2FA login

### LINKED ISSUE #13004: Login with Google 2FA account does not work
### Expected behavior
Attempting to login with a Google account that has 2FA enabled on the WordPress account should work.

### Actual behavior
Attempting to login with a Google account that has 2FA enabled on the WordPress account fails with the error:

google_2fa_fail

- Create a WP account with gmail address.
- On the WP account, enabled Two-Step Authentication.
- In the app:
  - Select Log In > Continue with Google.
  - Select the gmail address used above.
  - Get the error.

ScoutHarris:
Fixes #13004 
WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/157

This uses the WPAuth changes to fix the Google 2FA login flow.

To test:

---
The flow that is fixed:
- Select Log In > Continue with Google.
- Select a Google account that is already connected to WordPress, and has 2FA enabled.
- Verify you get prompted for a verification code, and can log in.

verification_prompt

---
Since a refactor was needed, sanity check these cases to be sure I didn't break anything:

- Select a Google account that is already connected to WordPress, but does _not_ have 2FA enabled. You should get logged in.
- Select a Google account that matches an existing WP account email, but is not yet connected to a WP account. You should get prompted to connect them.
- Select a Google account that does not match a WP account email. You should see this list of options.

what_do_i_do_now

---
PR submission checklist:


frosty:
I guess the above comments are technically more about the authenticator branch than this one, but as the testing steps are here I guess we can keep the discussion here too :)

ScoutHarris:
Regarding `when I tap on Continue with Google, I no longer get pushed to the 'waiting for Google to complete' screen` - 

Google _sign in_ never showed the loading view. Enabling the old login flow examples this.

google_log_in


It's happening on `develop` because it's erroneously showing the _sign up_ flow (which caused the 2FA bug to start with). So, it's actually displaying correctly, it's just Google signup vs signin have different flows.

PR #13012, opened 2019-11-27
Adds a debug menu and overriding feature flags

frosty:
This PR adds a Debug menu option to the bottom of App Settings (only displayed in debug builds), and allows you to toggle feature flags on and off at runtime.

ff1

Of course, depending on the implementation of feature flagged code, it's not guaranteed that toggling a flag while the app is already running will work correctly. But worst case, you should be able to restart the app after toggling a flag.

**To test**

* Build and run
* Domain credits is probably the easiest to test (or add your own simple feature flag!). Find a site that has a domain credit (or purchase a plan for a test site), and on the blog details screen observe that the domain credit banner is present.
* Navigate to Me > App Settings, and scroll to the bottom to Debug
* Toggle off the domain credits feature flag
* Go back to blog details and check the domain credit banner has gone
* Go back to the Debug screen and turn the feature flag back on
* Go back to blog details and check the banner is back

PR submission checklist:


WordPress/WordPressTest/FeatureFlagTests.swift
```diff
@@ -0,0 +1,108 @@
+//
+//  FeatureFlagTests.swift
+//  WordPressTest
+//
+//  Created by James Frost on 27/11/2019.
+//  Copyright © 2019 WordPress. All rights reserved.
+//
```
leandroalonso: I guess we can remove this header. :)

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -39,3 +47,98 @@ class Feature: NSObject {
         return feature.enabled
     }
 }
+
+
+extension FeatureFlag: CustomStringConvertible {
```
leandroalonso: Correct me if I'm wrong, but I understood that we need to conform to `CustomStringConvertible` because of `OverrideableFlag`, right?

In that case, wouldn't it make more sense to move the `description` to the `extension FeatureFlag: OverrideableFlag`?

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -39,3 +47,98 @@ class Feature: NSObject {
...
+            return false
+        default:
+            return true
+        }
+    }
+}
+
+/// Used to override values for feature flags at runtime in debug builds
+///
+struct FeatureFlagOverrideStore {
```
leandroalonso: What do you think about extracting the `FeatureFlagOverrideStore` to its own file?

koke:
> Of course, depending on the implementation of feature flagged code, it's not guaranteed that toggling a flag while the app is already running will work correctly. But worst case, you should be able to restart the app after toggling a flag

I think this is OK for a first iteration and I wouldn't want to over complicate things before we know how useful that would be, but if that becomes an issue I'd turn `FeatureFlag` into an actual `Store`, so other parts of the code can subscribe to changes

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
frosty: Sure, that makes sense 👍

WordPress/WordPressTest/FeatureFlagTests.swift
frosty: Whoops, nice spot!

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
frosty: Hmm, possibly? I left it in there as I felt it wasn't too big and obviously they're quite closely related. I think if we did extract it, I'd like to leave the `description` implementation in `FeatureFlag.swift`, as I think it's best to keep that near the initial definition of the FeatureFlag enum. Does that make sense? Which one would you prefer I change?

leandroalonso: We should definitely keep the `description` in `FeatureFlag.swift`!

I just thought about moving the `FeatureFlagOverrideStore` to another file to keep `FeatureFlag.swift` cleaner — just the flags and the descriptions.

But this is not a big deal, tbh. :)

frosty:
@leandroalonso Ready for another look!

frosty:
Thanks @leandroalonso!

PR #13020, opened 2019-11-28
Use new .org api client for Gutenberg networking

koke:
See https://github.com/wordpress-mobile/WordPressKit-iOS/pull/202 for description and testing instructions.

PR submission checklist:


WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift
```diff
@@ -36,27 +37,17 @@ struct GutenbergNetworkRequest {
...
+                        completion(.success(response))
+                    case .failure(let error):
+                        completion(.failure(error as NSError))
+                }
             }
-        }
-    }
-
-    private var selfHostedPath: String {
-        let removedEditContext = path.replacingOccurrences(of: "context=edit", with: "context=view")
```
pinarol: I know this is done just for being able to test but I'll drop a comment just not to forget. We can somehow leave requests that are already working with `context=view` out of this solution since the current solution is fragile.

SergioEstevao: @pinar @koke  What about we retry the request with `context=view` if the original one fails?

koke: I think I would rather avoid the `context=view` hack completely if we have authentication. It works for the image sizes but won't likely work for other requests and might cause some debugging headaches in the future.

SergioEstevao:
@etoledom and @pinarol can you test/review this again? I think it's ready to merge now.

loremattei:
Hey! I moved this PR to 14.6 as because 14.5 has been cut. If you want this to make it to 14.5, please feel free to ping me!

PR #13021, opened 2019-11-28
Remove entities with nil blog property

leandroalonso:
Refs #12028

Dependency: https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/227

As the main issue mentions, we're having a lot of crashes due to entities with `nil` blog property. We weren't able to determine WHY this is happening, but we were able to reproduce it by changing the `WordPress.sqlite` database.

Our attempt here is to sanitize Posts, Pages, Media, and Categories without a Blog.

Before diving into details, make sure you have https://sqlitebrowser.org/ installed on your computer.

## Testing

1. Run the app and detect where the `WordPress.sqlite` file is located (you can put a breakpoint in `ContextManager.m` or use [simsim](https://github.com/dsmelov/simsim).
2. Close the app
3. Open the database in SQLite Browser and run this query: `UPDATE ZBASEPOST SET ZBLOG = NULL` — now, all of your Posts and Pages will have `nil` Blog property
4. Close SQLite Browser and save the changes;
5. Open the app again
6. Check the `WordPress.sqlite` database again. The posts/pages without `Blog` were removed. Maybe you can have some entities without a `ZBLOG` — those are `ReaderPost` and they're not going to be deleted.
7. Load some posts, check some notifications: the app should work as it always worked. :)

## Additional info

This sanitization will happen once and every time the user updates the app. We added logging to keep track of how many users are affected by those orphaned entities.

## Reproducing the crash

Use `develop` for this.

1. First, look at your Notifications tab and find a comment notification. Take note on what Blog this post belong to;
2. Fresh install the app and login into your account;
3. Load the posts of the blog you received the notification in step 1 (do not open the Notifications tab)
4. Close the app;
5. Using the SQLite Browser, run this query: `UPDATE ZBASEPOST SET ZBLOG = NULL`
6. Open the app again, go to notifications. Tap on the notification and then to load the comments (you [can see where here](https://cld.wthms.co/LC0CE0)).
7. [The app will crash](https://cld.wthms.co/svrev5)

PR submission checklist:


koke:
> detect where the WordPress.sqlite file is located (you can put a breakpoint in ContextManager.m or use simsim.

Just a little tip for this :wink::

```
koke:~/ $ find `xcrun simctl get_app_container booted org.wordpress data`/Documents/WordPress.sqlite
/Users/koke/Library/Developer/CoreSimulator/Devices/C75644D6-5738-46D3-B1E7-9F56F8F2F23D/data/Containers/Data/Application/9E61054F-8009-48A4-A9D4-83C5D13A4394/Documents/WordPress.sqlite
```

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -71,6 +71,8 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
         // Restore a disassociated account prior to fixing tokens.
         AccountService(managedObjectContext: ContextManager.shared.mainContext).restoreDisassociatedAccountIfNecessary()
 
+        NullBlogPropertySanitizer().sanitize()
```
koke: I'm a bit concerned about potential race conditions with having this in the launch sequence, decoupled from Core Data initialization. It would still be possible that something else tries to save the context before this gets to do its thing.

It would be interesting to do this as part of `mainContext` initialization.

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
```diff
@@ -0,0 +1,66 @@
...
+            PostCategory.entityName()
+        ]
+
+        context.perform {
+            entityNamesWithRequiredBlogProperties.forEach { entityName in
+                let request = NSFetchRequest<NSManagedObject>(entityName: entityName)
+                let predicate = NSPredicate(format: "blog == NULL")
+                request.predicate = predicate
+
+                if let results = try? self.context.fetch(request), !results.isEmpty {
```
koke: Have you considered a [NSBatchDeleteRequest](https://www.avanderlee.com/swift/nsbatchdeleterequest-core-data/) instead? This seems like it could be relatively slow if there are many entities to fix, and it being in the launch path might end up causing a launch timeout.

leandroalonso: Very good point. The problem with the `NSBatchDeleteRequest` is that it will not use the validation rules we defined in Data Core thus not removing comments (for example). Although I believe this wouldn't be a problem this would leave the database in (another) inconsistent state.

One thing that I did was to move the place when we update the version that the sanitization happened (46df739af9f0f5e44b4ce2df701ce2cc0c46fc96). In a case where this fails or a timeout occurs, the next time it will not be executed — and the user wouldn't get locked out of the app.

We have two options here, both with pros/cons:

I guess we have a few options, let me know which one looks the better in your opinion:

- Use `NSBatchDeleteRequest` which is way quicker but with the risky of leaving the database in another consistent state;
- Keep the regular `delete` which is slower and might cause timeout with huge databases but will keep it consistent

@koke thoughts?

WordPress/Classes/System/WordPressAppDelegate.swift
leandroalonso: Done in d1fb327d8b30996755ef1360cb659b13298e2e81

shiki: Hmm, a problem with moving this in the `ContextManager.init` is the track events in [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/c8c932df4d32a7ee54145f91bcb994a0ae3093c6/WordPress/Classes/Services/NullBlogPropertySanitizer.swift#L48-L51) are no longer “tracked”. I don't see them in the log at all. I think we also need to ensure that Tracks is initialized first before `ContextManager`.

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
```diff
@@ -0,0 +1,66 @@
...
+/// set to NULL. They are currently inaccessible because they are not attached to any `Blog`. But
+/// leaving them in the database would cause a crash if Core Data tries to save new entities.
+///
+@objc class NullBlogPropertySanitizer: NSObject {
+    private let store: UserDefaults
+    private let key = "null-property-sanitization"
+
+    private let context: NSManagedObjectContext
+
+    @objc init(store: UserDefaults = UserDefaults.standard,
```
koke: I see that you are passing the `UserDefaults` here for testing, you might want to take a look at the existing `KeyValueDatabase` protocol which does a similar thing.

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
koke: > The problem with the NSBatchDeleteRequest is that it will not use the validation rules we defined in Data Core thus not removing comments (for example)

I'm not sure I understand which validation rules are you referring to, is this about the cascading deletes? So if I delete a post with a batch delete it won't get rid of the associated comments? That's pretty bad 😞

koke: I just realized that because this uses `context.perform`, the actually work is enqueued and out of the startup sequence, so this should not cause a launch timeout 😌 Optimizing this becomes less important now, but maybe we can add a `request.includesPropertyValues = false` to avoid loading all the data for objects that we are about to delete.

leandroalonso: > I'm not sure I understand which validation rules are you referring to, is this about the cascading deletes? So if I delete a post with a batch delete it won't get rid of the associated comments? That's pretty bad 😞

It's exactly that. :(

> Optimizing this becomes less important now, but maybe we can add a request.includesPropertyValues = false to avoid loading all the data for objects that we are about to delete.

Added in 9d3fcbdedfef7bafbcf0ab470cf94496c4fefd84

WordPress/Classes/System/WordPressAppDelegate.swift
leandroalonso: @shiki I was thinking that we can enqueue the values and track later once Tracks is initialized. I'd prefer to avoid changing any Tracks related code in this PR. What do you think?

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
leandroalonso: I tried that!

But since we're exposing that to ObjC I'm getting some weird errors when adding `@objc` in the `KeyValueDatabase`. :(

koke: Gotcha, I think what can work in that case is keeping the current initializer as Swift only, then adding a convenience `@objc init()` that calls the designated one with the default values, rather than using parameter defaults.

koke:
I think this is looking generally fine but I'm off until Tuesday so I won't be able to properly test until then, so don't hold on me if someone else approves

WordPress/Classes/System/WordPressAppDelegate.swift
shiki: We could probably also make the `PostCoordinator` usage of `ContextManager` to be `lazy` so it doesn't get initialized before everything else. And then force loading `ContextManager` ourselves in the AppDelegate right after Tracks is initialized. This seems risky though. 

> I was thinking that we can enqueue the values and track later once Tracks is initialized.

How do we do that?

shiki: Turns out that making `uploadsManager` `lazy` is enough to fix it. 😌 ef807792ff823cf747bfd7200126189aceca803c

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
shiki: Done from [ef80779 to 6057324](https://github.com/wordpress-mobile/WordPress-iOS/pull/13021/files/ef807792ff823cf747bfd7200126189aceca803c..605732455758ee30d00a82d9868953d007127bee).

shiki:
@koke Ready for another look. 🙂

WordPress/WordPressTest/NullBlogPropertySanitizerTests.swift
```diff
@@ -0,0 +1,69 @@
...
+        nullBlogPropertySanitizer.sanitize()
+
+        // Then
+        // The first invocation is in the _Given_ paragraph.
+        expect(self.keyValueDatabase.setValueForKeyInvocationCount).to(equal(1))
+    }
+}
+
+private class StubKeyValueDatabase: EphemeralKeyValueDatabase {
+    private(set) var setValueForKeyInvocationCount = 0
```
koke: I'm not sure why we need to override this class. Testing the number of times that the database is called seems like an implementation detail that the test should not care about: we are testing whether the version changes or not, I wouldn't really care if the implementation tried to write the existing value to the database.

The `lastSanitizationVersionNumber` is nice, but also something that could be done more generically for `KeyValueDatabase`. We already have a `bool(forKey:)` helper there, we could add a `string(forKey:)`, or even a generic one:

```swift
extension KeyValueDatabase {
    subscript<T>(key: String) -> T? {
        get {
            return object(forKey: key) as? T
        }
        set(newValue) {
            set(newValue, forKey: key)
        }
    }
}
```

Then you can just call that in the test:

```swift
expect(self.keyValueDatabase[NullBlogPropertySanitizer.lastSanitizationVersionNumber])
    .to(equal(currentBuildVersion))
```

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
```diff
@@ -0,0 +1,71 @@
...
+        self.store = store
+        self.context = context
+    }
+
+    @objc func sanitize() {
+        guard appWasUpdated() else {
+            return
+        }
+
+        self.store.set(self.currentBuildVersion(), forKey: Self.lastSanitizationVersionNumber)
```
koke:
```suggestion
        store.set(currentBuildVersion(), forKey: Self.lastSanitizationVersionNumber)
```

No need for the extra `self.`

WordPress/WordPressTest/NullBlogPropertySanitizerTests.swift
shiki: Thanks for the feedback!

> The lastSanitizationVersionNumber is nice, but also something that could be done more generically for KeyValueDatabase.

The intent was more about making the test readable. And also to hide the implementation detail (`lastSanitizationVersionNumber`). 

I did start with using `set(,:forKey)` and the lines got pretty long. Your `subscript` makes it a bit shorter. Thanks for that! TIL

Done in 595f5bba21e53a2238e39f06c6609befb0ec9a4d and d3d7e89c90254f21de2121c6fe0489c2134f0588.

> I'm not sure why we need to override this class. Testing the number of times that the database is called seems like an implementation detail that the test should not care about

You're right about the _implementation detail_ bit. In `testDoesntChangeVersionWhenSanitizationIsNotNeeded`, we are proving that the deletion routine in `sanitize()` is **never executed** unless the app was upgraded. Unfortunately, I couldn't find an easier way to test this. 

- We can't just check the value of `NullBlogPropertySanitizer.lastSanitizationVersionNumber` in `KeyValueDatabase` because it never changes during the test scenario. 
-  Passing a mocked `NSManagedObjectContext` and testing that the `context.perform { }` was never called seems more appropriate. But mocking that in Swift is probably a bit hairy. 
- Adding `Post` records with `NULL` `blog` values would be the best. We tried that but Core Data just fails and won't allow us to save the `Post`. 

With that said, I chose the `setValueForKeyInvocationCount` hack instead. It does bleed into the _"testing the implementation detail"_ side because we _had_ to know that the `set(,forKey:)` is only called once. I usually try to avoid doing testing like this. But considering the other options, I figured this was the less intrusive option. 😞 

I think I'll keep it for now. We are hoping for `NullBlogPropertySanitizer` to be temporary and iterating more may be too much at this point. 🙂

WordPress/Classes/Services/NullBlogPropertySanitizer.swift
shiki: Removed in 90f34bfffbcd47ce887c4bd41ea9c505c7d590ca.

shiki:
Thank you very much for reviewing this, @koke! 🙏

Podfile
```diff
@@ -11,13 +11,13 @@ workspace 'WordPress.xcworkspace'
...
     ## for production:
-    pod 'WordPressShared', '~> 1.8.9'
+    #pod 'WordPressShared', '~> 1.8.9'
 
     ## for development:
     # pod 'WordPressShared', :path => '../WordPress-iOS-Shared'
 
     ## while PR is in review:
-    # pod 'WordPressShared', :git => 'https://github.com/wordpress-mobile/WordPress-iOS-Shared', :branch => ''
+    pod 'WordPressShared', :git => 'https://github.com/wordpress-mobile/WordPress-iOS-Shared', :branch => 'issue/add-deleted-orphan-event'
```
koke: I just realized we probably should not have merged this until https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/227 was ready. I'm trying to update WordPressShared to the next beta version for some other new changes that are required, and now we're stuck pointing to a branch until that PR is resolved.

leandroalonso: Hi @koke, sorry about that.

Here we go: https://github.com/wordpress-mobile/WordPress-iOS/pull/13081

shiki: Gah! 🤦‍♂ That's my bad. I totally forgot about the pod. Sorry, everyone!

Issue #12790, opened 2019-10-28
Track usage in Media Editor


leandroalonso:
We want to track the usage of the Media Editor to:

1. Discover how many users are using it
2. Discover what operations the user is doing with the image

For that we will:

1. Fire `wpios_media_editor_show` when the Media Editor is shown
2. Fire `wpios_media_editor_used` with property `actions` (more details below) when the user edits the image

### Actions property

We will send the below actions:

1. `crop`: when the user cropped the image
2. `rotate`: when the user rotates the image
3. `crop, rotate`: when the user rotates and crop the image

### Additional information

The action `crop` could mean that the user resized the crop area OR that the user zoomed in/out the image OR both. Unfortunately, the library give us no detail about those operations.

### Funnel

We want to create a funnel to measure usage and check the `actions` to see what was most used.

PR #13031, opened 2019-12-03
Track usage of Media Editing

leandroalonso:
Ref #12790

Depends on #https://github.com/wordpress-mobile/WordPress-iOS/pull/12933

### Shown

1. Add an image to Aztec
2. Tap the image then tap "Edit"
3. Check in the log if there is a similar event to this one:

```
2019-12-03 09:50:26:755 WordPress[81680:1392001] 🔵 Tracked: media_editor_shown
```

### Used

1. Add an image to Aztec
2. Tap the image then tap "Edit"
3. Do some action with the image and tap "Done"
4. Check in the log if there is a similar event to this one:

```
2019-12-03 09:52:39:812 WordPress[81680:1392001] 🔵 Tracked: media_editor_used <actions: crop, rotate>
```

Make sure that `actions` contain only the actions you performed.

PR submission checklist:


Podfile
```diff
@@ -11,13 +11,13 @@ workspace 'WordPress.xcworkspace'
 ##
 def wordpress_shared
     ## for production:
-    pod 'WordPressShared', '~> 1.8.9-beta.1'
+    # pod 'WordPressShared', '~> 1.8.8'
```
yaelirub: You have an approved PR that updates to 1.8.10-beta-1 https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/225

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -2821,11 +2825,19 @@ extension AztecPostViewController {
 
         if showDefaultActions {
             if let imageAttachment = attachment as? ImageAttachment {
-                alertController.preferredAction = alertController.addActionWithTitle(MediaAttachmentActionSheet.editActionTitle,
+                alertController.preferredAction = alertController.addActionWithTitle(MediaAttachmentActionSheet.settingsActionTitle,
```
yaelirub: mmm... Are these updates to the action sheet relevant to tracking?

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -2821,11 +2825,19 @@ extension AztecPostViewController {
...
                                                                                      style: .default,
                                                                                      handler: { (action) in
                                                                                         self.displayDetails(forAttachment: imageAttachment)
                 })
+
+                if imageAttachment.isLoaded {
+                    alertController.addActionWithTitle(MediaAttachmentActionSheet.editActionTitle,
+                                                                                         style: .default,
+                                                                                         handler: { (action) in
+                                                                                            self.edit(imageAttachment)
```
yaelirub: You can remove the "( )" surrounding `action` and just be `action in`

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -3422,3 +3435,30 @@ extension AztecPostViewController: PostEditorNavigationBarManagerDelegate {
...
+//
+extension AztecPostViewController {
+    func edit(_ imageAttachment: ImageAttachment) {
+        guard let image = imageAttachment.image else {
+            return
+        }
+
+        WPAnalytics.track(.mediaEditorShown)
+
+        self.mediaEditor.edit(image, from: self,
```
yaelirub: can remove self here

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -3422,3 +3435,30 @@ extension AztecPostViewController: PostEditorNavigationBarManagerDelegate {
...
+                              onFinishEditing: { image, actions in
```
yaelirub: Out of curiosity, why do you put parentheses around a single parameter in a closure but not a round more than one?

yaelirub:
Hey @leandroalonso , great work! I tested and it work as described. 
A few questions:

1. For other trackings such as editor_photo_added for example, we add the blog_id, should we add that (or possibly other data) to the event?

2. If I open the media editor without editing anything and then tap 'Done`, the event `media_editor_used` gets fired and then the media is uploaded again. a couple of questions about that:
   2.1  Seems to me like we should either not call the event (cause the editor wasn't used) or create an 
          event "media_editor_closed". wdyt?
   2.2 Why are we uploading the media again if there was no change?
  
3. Should we add a property to `media_service_upload_started` or `media_service_upload_successful` events that shows the upload was due to edit?

Podfile
leandroalonso: Done in a03947d233d1714dd0775fae4013c1c0e399db59

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
leandroalonso: Done

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
leandroalonso: Hmmm, do I do that?

I always try to avoid using parentheses inside closures. If I'm doing that (and I don't even remember) I'm going crazy 🤣

leandroalonso:
1. Do you think we need that extra data in advance? I kept the same set of data as the others `addMedia` have in `AztecPostViewController`. What would be the usage of the specific blog ID in the event?
2. Very good point, I'll work on that: not upload the image and not firing the event.
3. Hmmm, this one is a little bit tricky. We will need to pass that information to `MediaCoordinator` and then to `MediaService`. Do you think that this is something we really need? If so, I'd be more comfortable working on that on another task.

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
yaelirub: 😝

yaelirub:
>Do you think we need that extra data in advance? I kept the same set of data as the others addMedia have in AztecPostViewController. What would be the usage of the specific blog ID in the event?

I believe the blog id is used to track in case something went wrong. Since we are tracking just for the sake of usability, maybe it's not necessary. Completely your call, I just thought I'd bring it up.

>Do you think that this is something we really need? If so, I'd be more comfortable working on that on another task.

My thought was that data will show a significant increase in media uploads and that might seem like more original medias are uploaded. Since we're intending on keeping a close eye on the usability data, I guess we can leave it out for now.

Issue #13030, opened 2019-12-02
[Stats Widgets] Add 'All Time' widget


ScoutHarris:
Design from Zeplin:

<img width="385" alt="all_time" src="https://user-images.githubusercontent.com/1816888/70001524-c2d08580-151a-11ea-9438-5733ebb4baf8.png">

PR #13034, opened 2019-12-04
[All Time Widget] Add default widget with supporting files and settings.

ScoutHarris:
Ref #13030 

This is step 1 of adding the 'All Time' stat widget. That is, adding a default widget with supporting files and settings.

To test:
The main purpose of this is to ensure supporting files/settings are correct.
- `Signing & Capabilities`:
  - Uses the Bundle Identifier `org.wordpress.WordPressAllTimeWidget`.  
  - Uses the `WordPress All Time Widget` Provisioning Profiles.
- `*.entitlements`: all the Entitlements looked exactly the same for all widgets & extensions, so I simply copied that implementation for this widget.
- `Info*.plist`s - the only difference between the three is the `Bundle display name`. 
- `Localizable.strings` - is added for Base. It's empty for now. I'll add content when the widget actually does something.

Please review anything you can think of related to widget settings.

In addition (of course), add the widget to your Today view. Since I've simply added the default widget, you'll be dazzled by the ubiquitous `Hello World` message.

default_widget

Footnote: I didn't see a way to "feature flag" a widget, so the feature branch `feature/13030-all_time_widget` will contain all widgety goodness until it's done.

PR submission checklist:


WordPress/WordPressAllTimeWidget/Tracks+AllTimeWidget.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    // MARK: - Private Helpers
+    fileprivate func trackExtensionEvent(_ event: ExtensionEvents, properties: [String: AnyObject]? = nil) {
+        track(event.rawValue, properties: properties)
+    }
+
+
+    // MARK: - Private Enums
+    fileprivate enum ExtensionEvents: String {
+        case Accessed           = "wpios_all_time_extension_accessed"
```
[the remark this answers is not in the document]
ScoutHarris: Ah, you caught me copying & pasting. 😄 I can do that. If you don't mind, since I've branched twice since this PR, I'll do it in the one I'm currently working on.

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -9186,6 +9191,7 @@
 		E1B34C091CCDFFCE00889709 /* Credentials */ = {
 			isa = PBXGroup;
 			children = (
+				29B97317FDCFA39411CA2CEA /* Resources */,
```
aerych: @ScoutHarris was this move intentional?  Seems like Resources shouldn't be a child of Credentials.

ScoutHarris: Gah! Nice catch! It's been fixed with https://github.com/wordpress-mobile/WordPress-iOS/pull/13177.

PR #13036, opened 2019-12-04
[All Time Widget] Add data handling

ScoutHarris:
Ref #13030 

This implements data handling for the 'All Time' widget.
- When Insights are accessed in the WP app, 'All Time' data is stored in `AllTimeData.plist`.
- The widget loads and saves data from/to the plist.
- The widget loads site information from `UserDefaults`, and uses it to fetch stat data.

The data is displayed in plain labels in the widget to facilitate testing this. The real UI will be implemented in a future PR.

To test:

---
- Go to Stats > Insights.
- Select `Widgets` in the nav bar, and then `Use this site`.
- Add the `WordPress All-Time` widget to your Today view.
- Verify the data displayed matches Insights > All-Time in the app.

insights

widget

---
- In the app, switch sites and `Use this site` again to change the site used for widgets.
- Verify the All Time widget updates accordingly.

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/AllTimeWidgetStats.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/AllTimeWidgetStats.swift
@@ -1,0 +1,73 @@
+import Foundation
+
+/// This struct contains data for the Insights All Time stats to be displayed in the corresponding widget.
+/// The data is stored in a plist for the widget to access.
+/// This file is shared with WordPressAllTimeWidget, which accesses the data when it is viewed.
+///
+
+struct AllTimeWidgetStats: Codable {
+    let views: Int
+    let visitors: Int
+    let posts: Int
+    let bestViews: Int
+
+    private enum CodingKeys: String, CodingKey {
+        case views
+        case visitors
+        case posts
+        case bestViews
+    }
+
+    init(views: Int, visitors: Int, posts: Int, bestViews: Int) {
+        self.views = views
+        self.visitors = visitors
+        self.posts = posts
+        self.bestViews = bestViews
+    }
+}
+
+extension AllTimeWidgetStats {
+
+    static func loadSavedData() -> AllTimeWidgetStats {
+        guard let sharedDataFileURL = dataFileURL,
+            FileManager.default.fileExists(atPath: sharedDataFileURL.path) == true else {
+                DDLogError("AllTimeWidgetStats: data file '\(dataFileName)' does not exist.")
+                return AllTimeWidgetStats(views: 0, visitors: 0, posts: 0, bestViews: 0)
+        }
+
+        let decoder = PropertyListDecoder()
+        do {
+            let data = try Data(contentsOf: sharedDataFileURL)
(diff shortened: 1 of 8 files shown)


WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/AllTimeWidgetStats.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+extension AllTimeWidgetStats {
+
+    static func loadSavedData() -> AllTimeWidgetStats {
+        guard let sharedDataFileURL = dataFileURL,
+            FileManager.default.fileExists(atPath: sharedDataFileURL.path) == true else {
+                DDLogError("AllTimeWidgetStats: data file '\(dataFileName)' does not exist.")
+                return AllTimeWidgetStats(views: 0, visitors: 0, posts: 0, bestViews: 0)
```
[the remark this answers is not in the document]
frosty: OR, we could make the normal parameters optionals with a default value of 0. That would also allow us to clean up a couple of places where we initialise with e.g. `viewsCount ?? 0`, I think?

WordPress/WordPressAllTimeWidget/AllTimeViewController.swift
```diff
@@ -1,30 +1,183 @@
...
+                // TODO: reload table here
+            }
+
+            completionHandler(NCUpdateResult.failed)
+            return
+        }
+
         tracks.trackExtensionAccessed()
-        completionHandler(NCUpdateResult.newData)
+        fetchData()
```
frosty: Should fetchData call the completionHandler on completion?

WordPress/WordPressAllTimeWidget/AllTimeViewController.swift
```diff
@@ -1,30 +1,183 @@
...
+        }
+
+        let wpApi = WordPressComRestApi(oAuthToken: oauthToken)
+        return StatsServiceRemoteV2(wordPressComRestApi: wpApi, siteID: siteID.intValue, siteTimezone: timeZone)
+    }
+
+    // MARK: - Helpers
+
+    func displayString(for value: Int) -> String {
+        return numberFormatter.string(from: NSNumber(value: value)) ?? "0"
```
frosty: Should we default to `noDataLabel` instead of 0 if we can't format a number? It just might better indicate that we couldn't load the value, rather than that there's actually data with a value of 0 there.

WordPress/WordPressAllTimeWidget/AllTimeViewController.swift
ScoutHarris: Wow, I missed that neither widget is doing that. Well spotted!

WordPress/WordPressAllTimeWidget/AllTimeViewController.swift
ScoutHarris: I elected to show the unformatted value instead. I think that's better than the `noDataLabel`.

ScoutHarris:
OK @frosty ready for another review please sir.

PR #13038, opened 2019-12-05
VoiceOver: add `accessibilityLabel` to Posts Lists' add & view buttons

### LINKED ISSUE #12872: Posts List: list view and add buttons are read incorrectly with VoiceOver
Steps to reproduce:

1. In the WordPress app, go to My Site > Blog Posts.
1. Swipe right to reach the buttons in the header for toggling compact and expanded view modes and for adding a new post.

**Result:** the compact list view button is read as "Button. Possibly menu." and should be something like "Switch to compact list view. Button." the expanded list view button is read as "Button." and should be something like "Switch to expanded list view. Button." and the add buton is read as "Button. Possibly add." and should be something like "Add new post. Button." ([5s](https://cloudup.com/cD6_Yuu34pZ))

posts-list-add-button
<sup>Tested with WPiOS 13.5.0.3 (App Store version) on iPhone 6S iOS 13.2.</sup>

<p style="clear:both;"><sub>/hat tip @etoledom for the report, internal reference: p77Llu-bQ6-p2</sub></p>

vphong:
Fixes #12872 - preliminary version

To test: VoiceOver on physical device

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostListViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostListViewController.swift
@@ -163,11 +163,15 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
     }
 
     func configureNavigationButtons() {
+        addButton.accessibilityLabel = "Add new post"
+        postsViewButtonItem.accessibilityLabel = isCompact ? "Switch to expanded list view" : "Switch to compact list view"
+
         navigationItem.rightBarButtonItems = [addButton, postsViewButtonItem]
     }
 
     @objc func togglePostsView() {
         isCompact.toggle()
+        postsViewButtonItem.accessibilityLabel = isCompact ? "Switch to expanded list view" : "Switch to compact list view"
 
         WPAppAnalytics.track(.postListToggleButtonPressed, withProperties: ["mode": isCompact ? Constants.compact: Constants.card])
     }


WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -163,11 +163,15 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
     }
 
     func configureNavigationButtons() {
+        addButton.accessibilityLabel = "Add new post"
```
shiki: I think a better place for this would be the `addButton`'s initializer in here:

https://github.com/wordpress-mobile/WordPress-iOS/blob/042e124c5f82516334a27380a4a30a68f0a0f8ce/WordPress/Classes/ViewRelated/Post/AbstractPostListViewController.swift#L131-L133

In this way, we can also fix the same problem in the Pages List. 

The label can also just be “Add”. There is enough context on the page that the user should be able to figure out that “Add” means adding a new post or adding a new page. You can read more about this in the [Guidelines for creating labels](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/iPhoneAccessibility/Making_Application_Accessible/Making_Application_Accessible.html#//apple_ref/doc/uid/TP40008785-CH102-SW20). 

Lastly, strings should be localized. Using `NSLocalizedString` should be enough. The translation will be handled during the release process. Please feel free to poke around in the codebase for some examples. 🙂 

- [ ] Move to the `addButton` initializer
- [ ] Use `NSLocalizedString`
- [ ] Shorten to `"Add"`

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -163,11 +163,15 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
...
+        postsViewButtonItem.accessibilityLabel = isCompact ? "Switch to expanded list view" : "Switch to compact list view"
```
shiki: It looks like you only need to do this once in `showCompactOrDefault()` which is called on `viewDidLoad` and whenever `isCompact` value is changed. The less code, the better. What do you think? 🙂 

As for the `accessibilityLabel`, this is a bit tricky because it's a toggle button. Changing the label can be confusing though. The user might think that it's a different button. My suggestion would be to have a static `accessibilityLabel`, `“List style”`, so the user will be able to identify that it's the same button. And then change the `accessibilityValue`. 

Layout Style | `accessibilityLabel` | `accessibilityValue`
--------|-------|---
Expanded        |       List style | expanded 
Compact | List style | compact

The label and the values will still be spoken by VoiceOver.

- [ ] Move to `showCompactOrDefault()`
- [ ] Use `accessibilityValue`

vphong: I did write that line twice, because I thought it was possible that the user might have left the app on a certain `compact` or `expanded` view, and they might want to return to their choice of view. So it calculates once on `viewDidLoad` and when the toggle is hit. Is this not necessary? 

I'll work on the other changes as well!

PR #13039, opened 2019-12-05
Fix search bar color

### LINKED ISSUE #13037: Search Bars: the color looks wrong for the search bar
Steps to reproduce:

1. Go to My Sites and check the color of the search bar.
1. Go to Blog Posts and check the color of the search bar.
1. Go to Reader > Manage and check the color of the search bar.
1. Switch from light to dark mode and repeat the steps above

**Result:** the search bar color is an odd grey in 13.8 beta.

IMG_3168 IMG_3170 IMG_3169 IMG_3171 IMG_3173 IMG_3174

leandroalonso:
Fixes #13037

| Light mode | Dark mode |
|   |   |

## To test

Use the app in both light and dark mode and check the search bar appearance.

## Additional info

It seems that, for some reason, Apple changed the default color for the `searchTextField`. This PR updates it with neutral with zero shade. This changes slightly the colors of the component, that's why I'm requesting a designer review.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+Search.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+Search.swift
@@ -9,10 +9,12 @@ extension WPStyleGuide {
         searchBar.autocorrectionType = .no
         searchBar.isTranslucent = false
         searchBar.barTintColor = .neutral(.shade10)
-        searchBar.searchTextField.backgroundColor = .basicBackground
         searchBar.layer.borderColor = UIColor.neutral(.shade10).cgColor
         searchBar.layer.borderWidth = 1.0
         searchBar.returnKeyType = .done
+        if #available(iOS 13.0, *) {
+            searchBar.searchTextField.backgroundColor = .basicBackground
+        }
     }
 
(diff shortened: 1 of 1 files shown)


mattmiklic:
This looks good overall, but I have one question. It looks like all of the colors specified for the search bars are from the "muriel" (custom) color palette. Are there default system colors for the search bars that we could use for iOS 13 and above, like we've done for most other components?

leandroalonso:
@mattmiklic I can change to `basicBackground` which uses the OS palette. The result:

| Light mode | Dark mode |
|   |   |

What do you think? The original or this one?

> Are there default system colors for the search bars that we could use for iOS 13 and above, like we've done for most other components?

@frosty has more context on the Dark Mode support. Can you help us with that question?

mattmiklic:
Thanks @leandroalonso; my vote would be to use the system default colors (for both light and dark) at least for iOS 13 and above since that matches how we've implemented the system colors to support dark mode.

leandroalonso:
@mattmiklic Awesome! I've changed to `basicBackground`.

leandroalonso:
@frosty Fixed!

Issue #13044, opened 2019-12-05
[Stats Widgets] Add 'Views This Week' widget


ScoutHarris:
Design from Zeplin:

<img width="367" alt="views_this_week" src="https://user-images.githubusercontent.com/1816888/70266001-d9642000-1758-11ea-9ad6-593f2a6c6c72.png">

PR #13045, opened 2019-12-05
[This Week Widget] Add default widget with supporting files and settings.

ScoutHarris:
Ref #13044 

This is step 1 of adding the 'This Week' stat widget. That is, adding a default widget with supporting files and settings.

To test:

---
The main purpose of this is to ensure supporting files/settings are correct.
- `Signing & Capabilities`:
  - Uses the Bundle Identifier `org.wordpress.WordPressThisWeekWidget`.  
  - Uses the `WordPress This Week Widget` Provisioning Profiles.
- `*.entitlements`: same as for the other widgets.
- `Info*.plist`s - the only difference between the three is the `Bundle display name`. 
- `Localizable.strings` - is added for Base, and it empty for now.

Please review anything you can think of related to widget settings.

---
Add the widget to your Today view. Since I've simply added the default widget, you'll just see a `Hello World` message.

default_widget

---
PR submission checklist:


ScoutHarris:
Hey @osullivanchris .

We may need to rename this widget.

If you'll recall, we added `WordPress` to all the widget names so they are grouped together in the widget list. This of course makes the widget names longer than originally designed.

For `WordPress Views This Week`, the length is OK for release. However, when we add `Internal` or `Alpha`, it gets truncated in the widget list.

release_name

internal_name


Please note the widget name is used in both the widget list and on the widget itself; it's not possible to have two different names. So whatever we decide on will be displayed in both places.

widget

frosty:
Or `WordPress This Week`?

osullivanchris:
AFK today but a quick reply to say I agree with @frosty's idea! But truncation could still be an issue with other languages.

If its still a problem in other languages, would we consider using **WP** instead of **WordPress** or is that not a done thing? Don't know if there are brand guidelines or something.

ScoutHarris:
> would we consider using WP instead of WordPress

I wondered the same - I'm not sure if we're "allowed" to do that. 

Since this widget is still a WIP, for this PR, I'll use `WordPress This Week`. An initial test shows it does fit.

Separately, I'll test with other languages and bring this back up if I find anything.

Thanks guys!

ScoutHarris:
@frosty I updated the widget names. Ready for a review please.

Issue #12953, opened 2019-11-15
[VoiceOver] Aztec Editor concerns


shiki:
## [A] There is an extra accessible “text field” element  

This is not the title and content text fields. It is also in between the title and content text fields in the navigation order. 

<img src="https://user-images.githubusercontent.com/198826/68908343-35162d00-0708-11ea-8ea8-9ac21088fca4.png" width="320">

## [B] “Insert {num}“ button doesn't have a hint

Something like “Inserts the selected media” can make it more obvious. 

- [x] Done in #13050 

## [C] “Insert” button should always be visible

It is more discoverable if it's there, disabled but accessible, and not suddenly appear without auditory notice. 

It can also make use of `.accessibilityValue` if it doesn't already. 

- [x] Done in #13050 

## [D] The _More_ button is dictated as “Possibly, More”. 

It should also be presented as a toggle and have the `.selected` trait when needed. A hint could also be useful here. 

We might also need to `post()` a [`.screenChanged` or `.layoutChanged` notification](https://developer.apple.com/documentation/uikit/uiaccessibility/notification/1620186-layoutchanged). 

- [x] Done in https://github.com/wordpress-mobile/AztecEditor-iOS/pull/1246

shiki:
> [A] There is an extra accessible “text field” element

This extra text field element is the rich content text field and is actually the correct text field. The _Share your story here..._ is just a label (placeholder) and shouldn't actually be an accessible element. 

The problem is when the rich content text is selected by VoiceOver, double-tapping it will **not activate it**. This cause seems to be because the content text field is **slightly covered** by the title text field. 

I've tried 2 things that might fix this:

1. Moving the title below the content in [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/2c8480032ef978549395f5bfa5340d932db77e68/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift#L715-L722). This causes the title to disappear though. But the content text field can be activated by VoiceOver.
[...]

PR #13050, opened 2019-12-06
[VoiceOver] Enhancements for Aztec editor

shiki:
Refs #12953

### Removed extra accessible “Title“

This accessible element is the title placeholder (`UILabel`) and can be confused with the title text field. 

### Made “Insert” button always visible when VoiceOver is turned on

Disabled if nothing is selected | Enabled when a media is selected
--------|-------
image        |       image

Making it visible could make it more discoverable and makes it easier to understand how to use the UI. If VoiceOver is turned off, the app will use the original behavior of hiding or showing the button. 

I also changed the `accessibilityLabel` of the button to `“Insert selected”` to make it more obvious **when** it will be enabled. 

### For: Removed extra accessible “Title“

1. Turn VoiceOver on.
2. Create a new post (Aztec)
3. Navigate by swiping right or left. Confirm that there is only one accessible “Title“ element and it can be activated

### For: Made “Insert” button always visible when VoiceOver is turned on

1. Turn VoiceOver on.
2. Create a new post (Aztec)
3. Navigate to the content text field so that the Insert Media button is enabled. 
4. Tap on the Insert Media button. 
5. Confirm that the Insert button (on the right) is shown but disabled. 
6. Select a media. 
7. Confirm that the Insert button is enabled and VoiceOver will speak “Insert selected, 1, button“. 

If possible, please do a regression test with VoiceOver turned off. The Insert button should still be hidden and shown if necessary. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -415,6 +415,13 @@ class AztecPostViewController: UIViewController, PostEditor {
         return insertItem
     }()
 
+    /// Indicates whether the `insertToolbarItem` is always shown in `FormatBarMode.media` mode.
+    ///
+    /// If true, the insertToolbarItem is enabled and disabled instead of shown and hidden.
+    private var alwaysDisplayInsertToolbarItem: Bool {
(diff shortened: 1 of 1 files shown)


shiki:
Thanks, @aerych!

PR #13051, opened 2019-12-06
Make placeholder images tinted with text color.

SergioEstevao:
This way the placeholder images will adapt to the current interface
mode on the device.

Fixes #https://github.com/wordpress-mobile/AztecEditor-iOS/issues/1236

**Note**: At this PR I'm not fixing the placeholders when the user switches from dark mode to light mode, while the editor is open. This will need a fix in the Aztec lib to force a refresh of the media attachments.

To test:
 - Start the app using dark mode

Scenario 1:

 - Open a new post
 - Insert a new media from device
 - Check that the placeholder image show properly while the final image is not ready

Scenario 2

 - Open a post with several images
 - Check that the placeholder images show properly while the final images are loading.

Run the same tests with image/video/files
Run the same tests with light mode to make sure all is ok

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
@@ -21,11 +21,8 @@ class EditorMediaUtility {
         default:
             icon = Gridicon.iconOfType(.attachment, withSize: size)
         }
-        if #available(iOS 13.0, *) {
-            if let color = tintColor {
-
-                icon = icon.withTintColor(color)
-            }
+        if #available(iOS 13.0, *), let color = tintColor {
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
```diff
@@ -21,7 +21,12 @@ class EditorMediaUtility {
         default:
             icon = Gridicon.iconOfType(.attachment, withSize: size)
         }
+        if #available(iOS 13.0, *) {
```
frosty: Nitpick, but I guess we could combine this into a single `if`:

```
if #available(iOS 13.0, *),
    let color = tintColor {
```

Issue #12960, opened 2019-11-15
Post Reblogging Master Issue


ScoutHarris:
Allow user to reblog a post from the Reader.

Initial discovery [post](https://wp.me/p77Llu-cLw-p2).
Design discovery [post](https://wp.me/paCL20-jC-p2).

- [x] Add feature flag. (#13055) 
- [x] Add reblog functionality. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13137)
  - [x] Reblog button in ReaderPostCardCell and ReaderDetailViewController (#13059)
  - [x] Reblog Action: Users with one available site (#13094)
  - [x] Post Reblogging - Users with no available sites and users with more than one available site. (#13124)
- [x] Ensure views are correct for Dark Mode.
- [x] Add Accessibility support.
- [x] Add Tracks. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13139)
- [x] Enable feature flag for internal.
- [x] Enable feature flag for all. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13361)
- [x] Remove feature flag.

#### Issues
- [x] From https://github.com/wordpress-mobile/WordPress-iOS/pull/13137#pullrequestreview-335217246:
We need to look at how we're incorporating images. Currently we're 'hotlinking' to the original image hosted on the source website – but this isn't the best practice as it uses the source website's bandwidth, and if the source website ever went away or reorganised their media URLs then the image would stop working. Some websites may even serve up a different image if sites try to link directly to their images. I think ideally, we'd need to re-upload the image to the user's site and use our own copy of it.
  - After a few discussions, it was determined that we can hotlink images through Photon (same as we do in the web)

PR #13059, opened 2019-12-09
Reblog button in ReaderPostCardCell and ReaderDetailViewController

Gio2018:
## This PR adds the reblog button in ReaderPostCardCell and ReaderDetailViewController

Ref: #12960

## Description
The goal of this PR is to add the reblog button in the reader, both in `ReaderPostCardCell` and in `ReaderDetailViewController`. 

The design specs have been applied (see extract from the design spec for `ReaderDetailViewController`)
reblog-action-location_option2

in `ReaderPostCardCel`l, the reblog button is placed between the save and the comment buttons.

Accessibility support has been added.

Unit tests are included (extended `ReaderPostCardCellTests.swift` and created `ReaderDetailViewControllerTests.swift`).

#### Analytics (tracks) are yet to be added.

##To test
- Build and run the app
- Tap on reader in the tab bar
- Navigate to one of your sites
- Find a post that supports actions
- Check that the reblog button is between the save and comment buttons
- Check that the reblog button highlights on tap (no action will be performed at this moment)
- Navigate to the detail view by tapping on the image or text of the post
- Check that the reblog button is present, as shown in the picture above
- Check that the reblog button highlights on tap (no action will be performed at this moment)
- Switch to dark mode and make sure that the reblog button has consistent appearance with the other buttons

##Notes
- Buttons have been added in the existing `StackViews` that contained the original action buttons
- The associated class is `PostMetaButton`
- Two methods have been added in `WPStyleGuide+Reader.swift`:
   - `applyReaderReblogActionButtonStyle(_ : , _ : )`: overall configuration for the reblog button
   - `applyReaderReblogActionButtonTitle(_ : , - : )`: label configuration for the reblog, allows to define an empty title (for the scenarios where we drop the labels)
  - The above methods are used in both `ReaderPostCardCell` and `ReaderDetailViewController` 
- Still in `WPStyleGuide+Reader.swift`,  `applyReaderSaveForLaterButtonTitles(_ : , _ : )` has been slightly modified to support empty title and it's used in both `ReaderPostCardCell` and `ReaderDetailViewController`


WordPress/Classes/ViewRelated/Reader/Reader.storyboard
```diff
@@ -346,8 +346,19 @@
...
                                             </button>
                                             <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="yq2-Pz-GyO">
-                                                <rect key="frame" x="48" y="0.0" width="192" height="50"/>
+                                                <rect key="frame" x="48" y="0.0" width="136" height="50"/>
                                             </view>
+                                            <button opaque="NO" contentMode="scaleToFill" horizontalHuggingPriority="750" horizontalCompressionResistancePriority="1000" contentHorizontalAlignment="right" contentVerticalAlignment="center" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="piB-Zd-RVQ" customClass="PostMetaButton">
+                                                <rect key="frame" x="200" y="13" width="40" height="24"/>
+                                                <constraints>
+                                                    <constraint firstAttribute="width" relation="greaterThanOrEqual" constant="30" id="Mum-BO-4Xd"/>
```
ScoutHarris: I noticed the other button width constraints have a priority of 750. I didn't see any spacing issues, but I suspect there was a reason. Maybe the reblog button constraint should follow suit?

button_constraints

WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
```diff
@@ -914,8 +925,12 @@ open class ReaderDetailViewController: UIViewController, UIViewControllerRestora
 
     fileprivate func configureLikeActionButton(_ animated: Bool = false) {
         likeButton.isEnabled = ReaderHelpers.isLoggedIn()
+        // as by design spec, only display like counts
+        let likeCount = post?.likeCount()?.intValue ?? 0
+        let shortTitle = likeCount > 0 ? "\(likeCount)" : ""
+
+        let title = FeatureFlag.postReblogging.enabled ? shortTitle : post!.likeCountForDisplay()
```
ScoutHarris: I know this is legacy, but we _really_ try not to force unwrap. Can you fix both of these?

`post!.likeCountForDisplay()`
`post!.isLiked`

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -231,6 +235,10 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
         WPStyleGuide.applyReaderSaveForLaterButtonStyle(saveForLaterButton)
     }
 
+    fileprivate func setupReblogActionButton() {
+        WPStyleGuide.applyReaderReblogActionButtonStyle(reblogActionButton)
```
ScoutHarris: It looks like this method only sets the style. Maybe instead of a separate method, this line can just be added to `ApplyStyles`? I know this is legacy, but I think the same can be done for `setupSaveForLaterButton` (i.e. setting the styles in `ApplyStyles` and nixing `setupSaveForLaterButton` and `setupReblogActionButton`).

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -501,6 +512,25 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
...
+        guard FeatureFlag.postReblogging.enabled else {
+            return false
+        }
+        // No post content -> nothing to reblog
+        guard contentProvider != nil else {
+            return false
+        }
+        // User not logged in -> User cannot reblog
+        guard loggedInActionVisibility.isEnabled else {
+            return false
```
ScoutHarris: Can we have just one `guard` statement for all these `false` conditions?

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
```diff
@@ -289,15 +289,40 @@ extension WPStyleGuide {
...
+
+        WPStyleGuide.applyReaderReblogActionButtonTitle(button, showTitle: showTitle)
+        WPStyleGuide.applyReaderActionButtonStyle(button)
+    }
+
+    /// Applies the reblog button title to the button passed as an argument
+    /// - Parameter button: the button to apply the title to
+    /// - Parameter showTitle: if set to true, will show the button label (default: true)
+    @objc public class func applyReaderReblogActionButtonTitle(_ button: UIButton, showTitle: Bool = true) {
+        let title = showTitle ? NSLocalizedString("Reblog", comment: "Text for the 'Reblog' button.") : ""
```
ScoutHarris: Should this fall back to "Reblog" instead of ""? It might be better to show _something_ in case the translation fails.

WordPress/WordPressTest/ReaderDetailViewControllerTests.swift
```diff
@@ -0,0 +1,36 @@
...
+        // will trigger viewDidLoad()
+        _ = viewController.view
+    }
+
+    override func tearDown() {
+        readerDetailViewController = nil
+        super.tearDown()
+    }
+
+    func testReblogButonMathchesExpectation() {
```
ScoutHarris: I'm assuming this should be `testReblogButtonMatchesExpectation` .

WordPress/WordPressTest/ReaderPostCardCellTests.swift
```diff
@@ -185,4 +187,36 @@ final class ReaderPostCardCellTests: XCTestCase {
         XCTAssertEqual(cell?.getVisitButtonForTesting().accessibilityLabel, String(format: "%@", TestConstants.visitLabel), "Incorrect accessibility label: Visit button"
     )
     }
+
+    func testReblogActionButonMathchesExpectation() {
```
ScoutHarris: I'm assuming this should be `testReblogActionButtonMatchesExpectation`.

ScoutHarris:
> Accessibility support has been added.
> No action is associated to the reblog button yet.
> Analytics (tracks) are yet to be added.

Thank you for specifying what is(not) included!

WordPress/Classes/ViewRelated/Reader/Reader.storyboard
Gio2018: Yes, it should be set as the others

WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift
Gio2018: Totally agree, even more so for the button title `post!likeCountForDisplay( )` which can also be set to nil, so it shouldn't be force unwrapped

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
Gio2018: Yeah they can and they should go in there, we reduce code lines and don't lose clarity

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
Gio2018: 👍

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
Gio2018: This method will set the title to an empty string if `showTitle` is false, which means we want the button to drop the label. `NSLocalizedString` should fall back to its key ("Reblog" in this case) if no translation is found. The method documentation in the comment is not very clear as it is, I'll add some clarifications

WordPress/WordPressTest/ReaderDetailViewControllerTests.swift
Gio2018: Yes, typo

WordPress/WordPressTest/ReaderPostCardCellTests.swift
Gio2018: Same

Gio2018:
> In addition to the various code comments:
>
> > Check that the reblog button highlights on tap
> This does happen on post details, but not in the reader.

Interesting catch. What happens is that, on the `UITableView` (subclass of `UIScrollView`) there is a default delay on the touches.

More precisely

> Because a scroll view has no scroll bars, it must know whether a touch signals an intent to scroll versus an intent to track a subview in the content. To make this determination, it temporarily intercepts a touch-down event by starting a timer and, before the timer fires, seeing if the touching finger makes any movement.

(extracted from Apple developer documentation: https://developer.apple.com/documentation/uikit/uiscrollview)

In fact, keeping the finger (or the pointer, in the simulator) for a little longer, will show the highlight.

The same thing happens for all the buttons in the cell.

This also does not seem to affect the responsiveness of the action itself, just the highlight of the button.

As the article points out, if we want we can override this behavior, by adding this line of code 

`tableView.delaysContentTouches = false`

in `ReaderStreamViewController.swift` (e.g. in the `setupTableView()` method).

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
ScoutHarris: My apologies! I misread what this method was doing.

WordPress/WordPressTest/ReaderDetailViewControllerTests.swift
```diff
@@ -27,7 +27,7 @@ class ReaderDetailViewControllerTests: XCTestCase {
         super.tearDown()
     }
 
-    func testReblogButonMathchesExpectation() {
+    func testReblogButtonMathchesExpectation() {
```
ScoutHarris: One more time - `Mathches` should be `Matches`.

WordPress/WordPressTest/ReaderPostCardCellTests.swift
```diff
@@ -188,7 +188,7 @@ final class ReaderPostCardCellTests: XCTestCase {
     )
     }
 
-    func testReblogActionButonMathchesExpectation() {
+    func testReblogActionButtonMathchesExpectation() {
```
ScoutHarris: One more time - `Mathches` should be `Matches`.

WordPress/WordPressTest/ReaderDetailViewControllerTests.swift
Gio2018: Sorry about that! Looks like I'm blind today

ScoutHarris:
> In fact, keeping the finger (or the pointer, in the simulator) for a little longer, will show the highlight.

Oh, that's interesting. And I see it now.

> if we want we can override this behavior

I don't think that's necessary. Once the reblog action is added, it won't even be noticed. Very much like the Comment action - you don't even notice the highlight because the Comment view is immediately shown. So, I think it's ok the way it is.  

Thanks!!

PR #13082, opened 2019-12-12
Testing Gutenberg after merging navigation feature

lukewalczak:
Fixes #

To test:

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -141,7 +141,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => '06898e77ae3cd8016e179cbb01cbfb4bbd014956'
+    gutenberg :commit => 'e2d1d9c3e7ba19cde8ab7c5a1e327215e8cfcc0b'
 
     ## Third party libraries
     ## =====================
--- a/Podfile
+++ b/Podfile
@@ -141,7 +141,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => 'e6ac5711989e9f69d63bf2fd151170330d008323'
+    gutenberg :commit => '06898e77ae3cd8016e179cbb01cbfb4bbd014956'
 
     ## Third party libraries
     ## =====================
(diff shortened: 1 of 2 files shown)


pinarol:
@iamthomasbishop Could you try to install the app with the peril link provided by this PR? It demonstrates how it looks after the merge of inner block enhancements. There are some things that look a bit odd to me but I can't be sure if that's intended. If you find some issues we'll need to fix those before holidays so the earlier we spot the better.

Screen Shot 2019-12-12 at 15 23 25

Screen Shot 2019-12-12 at 15 25 27

iamthomasbishop:
That's right, @pinarol – the Inline Toolbar does look like it has some extra spacing – looks like not just additional bottom spacing, but also and left/right spacing (16px?):

image

iamthomasbishop:
It is so great to see this in action – a long time coming! 😄 I have some quick feedback.

* As discussed earlier, there is an issue w/ the spacing of the Inline Toolbar.

* When I add any type of RichText block within an empty Group, the input isn't focused right away – this felt odd, and it doesn't happen on a non-empty Group block. 

* The margins/padding on the Columns block look a bit broken ([screenshot](https://cloudup.com/ctMkCU0kJdf)). Unsupported, I know, but I tried adding one on web 😄 

* Can we hide the Floating Toolbar on unsupported blocks like the Columns block? (see above screenshot)

* I know we had discussed the Appender spacing, but I think we might want to align to the same visual keyline as the child block borders ([screenshot](https://cloudup.com/cXshK3Os6xJ))

* Is it possible to remove the opacity-change effect when the breadcrumbs are tapped? Because nothing happens when I tap it (which is fine until we have the doc navigator), it might feel like a bug to users.

* It looks like there is extra padding applied on the Media & Text block now? ([screenshot](https://cloudup.com/cgzjtF03r2k))

That's all I've got right now, I am going to give it another spin tomorrow morning 😄 Thanks for all of the hard work on this thus far!

iamthomasbishop:
@pinarol @lukewalczak Updated my last comment to include which I think are blockers vs. non-blockers and roughly sorted by priority.

PR #13069, opened 2019-12-10
Media Editor Architecture Improvement

leandroalonso:
This PR improves the Media Editor architecture to:

* Make easier to add new capabilities in the future
* Make easier to deal with async images (next task)
* Make the code better and decoupled :)

## A little bit of information

The Media Editor is now modular, which means new capabilities can be added in the form of plugins! Yay!

Here's an image that shows that concept:

Artboard

This makes really easy for us or the community to grow the list of features of our Media Editor.

## To test

Nothing should have changed from this PR:

1. Using Aztec, add an image
2. Crop the image, check if the image was updated and the log of events is ok

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #13048, opened 2019-12-06
Editing async images


leandroalonso:
The Media Editor should be able to accept async images. This will cover some cases like:

* Images that we just have a small thumbnail and need to download the full quality image
* `PHAsset`s that we need to wait for the maximum size

## Acceptance criteria

1. User taps to edit an image we just have a thumbnail
2. A screen is shown with the thumbnail image and an activity indicator
3. Once the image is downloaded we show the crop screen
 
.

1. User taps to edit an image we just have a thumbnail
2. A screen is shown with the thumbnail image and an activity indicator
3. User tap cancel
4. We dismiss the screen and cancel the image request

## Design

![async_image](https://user-images.githubusercontent.com/7040243/70330737-e8140b00-181c-11ea-8346-8655d5c2bd1a.png)

* This activity indicator will probably change
* The crop buttons will not be displayed, only the Cancel

PR #13085, opened 2019-12-12
Edit async images

leandroalonso:
Refs #13048 

Depends on https://github.com/wordpress-mobile/WordPress-iOS/pull/13069

Allows the user to edit "async images", ie.: images that need to be downloaded to be manipulated.

Keep in mind that this PR **does not cover** the integration with Gutenberg.

## To test

First, [apply this patch](https://gist.github.com/leandroalonso/131eb194eb38d7c6ea0f2eb1d7574f50).

*Creating a new post*

1. Create a new post in Gutenberg
2. Add an image
3. Tap the image twice
4. Check that the full quality image will be downloaded and then the crop capabilities are shown

*Editing an image added from web*

1. Using the browser, create a new post using Gutenberg
2. Add an image, then change "Image Size" to"Thumbnail" (in the sidebar at the right)
3. Save or publish the post
4. Open it on the app
5. Double-tap the image
6. Checks that the full quality image is downloaded and can be cropped/resized — not the thumbnail

*iPhone, iPad, Portrait and Landscape*

1. Please, test it using iPhone or iPad. Portrait and landscape!

## More information

To handle editing and manipulating "async images" an `AsyncImage` protocol was created. By conforming to that, every entity can be an async asset. :)

Let's get our app as an example: we have `PHAsset`, Aztec's attachment, Gutenberg Media... all of that can be given to Media Editor — the only requirement is to conform to the protocol.

You can see an example at the given patch above. An extension for `PHAsset` will be added soon. The protocol is quite simple, it contains:

* A `thumb` optional UIImage — if given, this will be displayed right away (as we did in the Gutenberg example);
* A `thumbnail` method. If no `thumb` was given, this method will be called. It is unlikely that we won't have a thumbnail but who knows...
* A `full` method. This method will be called to start fetching the image, and once is done to call the callback.
* A `cancel` method. This is called when the user gave up on editing the media while the asset is being downloaded. Here we should cancel all the ongoing operations (let's save some network data)

PR submission checklist:


yaelirub:
@leandroalonso , Im reviewing this PR.
I noticed that I can get to the media editor by tapping on the image once.
I know that we're going to update the entry point but just wanted to make sure that's something you're aware of

WordPress/MediaEditor/MediaEditor.swift
```diff
@@ -53,17 +88,65 @@ public class MediaEditor: UINavigationController {
...
         capability.apply(styles: styles)
         currentCapability = capability
+
         pushViewController(capability.viewController, animated: false)
     }
+
+    private func prepareTransition() {
+        let transition: CATransition = CATransition()
+        transition.duration = 0.3
+        transition.type = CATransitionType.fade
```
yaelirub: you can remove CATransitionType

WordPress/MediaEditor/MediaEditor.swift
```diff
@@ -53,17 +88,65 @@ public class MediaEditor: UINavigationController {
         present(capability: capabilityEntity, with: image)
     }
 
+    private func cancel() {
```
yaelirub: Is this going to be called from anywhere else other than as the onCancel block for present?

leandroalonso: I don't think so. This might change in the future but this will be called only when the MediaEditor is dismissed.

leandroalonso:
@yaelirub I'm aware! This single tap is just for testing purposes.

I thought it was a double-tap, my bad.

WordPress/MediaEditor/MediaEditor.swift
```diff
@@ -21,30 +27,59 @@ public class MediaEditor: UINavigationController {
...
+        setupForAsync()
+
+        presentIfSingleImageAndCapability()
+    }
+
+    private func setupForAsync() {
+        if let thumb = asyncImage?.thumb {
+            hub.show(image: thumb)
+        } else {
+            asyncImage?.thumbnail(finishedRetrievingThumbnail: thumbnailAvailable)
```
diegoreymendez: Disclaimer: the following, though unlikely, is definitely possible.

This call starts an asynchronous download of the thumbnail.  The same is true for the `full()` call below.  Hence it's possible for the completion callbacks to be executed in a different order than you'd expect.

I believe the callbacks have no safeguards in place for triggering out of order.

Two simple possibilities to make this safer could be:

1. Not setting the thumbnail if the full image is already available, or
2. Queueing the downloads one after another (download thumbnail, and when done download full image).

I've no particular preference.

I realize this feels like a bit of an edge case, but since the extra work for a safeguard is not much, and since the bug would be annoying / hard to debug / hard to understand / noisy, I'd strongly suggest we try to prevent it.

WordPress/MediaEditor/MediaEditor.swift
```diff
@@ -53,17 +88,65 @@ public class MediaEditor: UINavigationController {
...
         )
         capability.apply(styles: styles)
         currentCapability = capability
+
         pushViewController(capability.viewController, animated: false)
     }
+
+    private func prepareTransition() {
+        let transition: CATransition = CATransition()
+        transition.duration = 0.3
```
diegoreymendez: Magic number: you could use something like `WPAnimationDurationDefault`, or have a custom constant for this class.

leandroalonso: Fixed in f1cdf6c20b116b69009cb2e69bdde0fc295bead1

WordPress/MediaEditor/MediaEditor.swift
leandroalonso: Great catch!

Fixed in 7b68415574545e0f64cfaeb8e75e8bd1f1a81089

leandroalonso:
@diegoreymendez Ready for another review!

WordPress/MediaEditor/MediaEditor.swift
diegoreymendez: Great commit.  Marking as resolved.

WordPress/MediaEditor/MediaEditor.swift
diegoreymendez: Marking as resolved.

PR #13094, opened 2019-12-13
Reblog Action: Users with one available site

Gio2018:
### This PR adds the reblog action to the app. It covers the use case of users with one available site.

Ref: #12960

### Description
The goal of this PR is to add the reblog action to the app, and connect it to the reblog buttons in either `ReaderPostCardCell` and `ReaderDetailViewController`.
At this moment, it only covers the use case of users who have only one available site. For this use case, when the `reblog` button is tapped, the block editor will open, and its editable content will consist of an image, a quote from the original post and a citation containing a link to the original post. Below is a short video that illustrates the behavior of the feature when tapping on the `reblog` button in the post detail, in the reader.

Dec-13-2019 17-15-34

This same behavior also happens when tapping on the `reblog` button in a cell of the post list (see sequence of screenshots below).

Screen Shot 2019-12-13 at 6 08 35 PM

Change log:

- Added a new `ReaderReblogAction.swift` file, containing 
  - a ReaderReblocAction type, with an execute(_ :, _ :) method, triggered by a UI action
  - a ReblogPresenter type, responsible of presenting the appropriate screens, depending on how many available sites are there
  - a ReblogFormatter type, responsible of formatting the content in a Gutenberg-ready html, which will trigger the block editor
  - an extension to `Post' that adds a method to prepare the content for reblogging
- Added @IBAction methods connected to the reblog buttons, in `ReaderPostCardCell.swift` and `ReaderDetailViewController.swift`, which will trigger ReaderReblogAction.execute(_:, _:)
- Added a new method `func readerCell(_ cell: ReaderPostCardCell, reblogActionForProvider provider: ReaderPostContentProvider)` in the `ReaderPostCardCellDelegate` protocol
- Added unit tests for the new objects

#### Note
1. The use cases of 
  - users with zero available sites
  - users with multiple (more than 1) available sites
are yet to come, and not included in this PR.

2. Analytics (tracks) are not yet included.

### To test:
- Make sure you have credentials for a test user that has only one available site.
  - if your test user has zero available  sites, or more than one available site, tapping on the `reblog` buttons will have no effect
- Update `wpcom_app_credentials` as needed.
- Build and run the app, and log in
- Select the reader in the tab bar
- From the `STREAMS` section, select the followed site(s) of your preference
- In the post list view, tap the `reblog` button on any post that has reblog enabled
- check that the content of the editor matches the description above
- tap on the ellipsis, and test that all the functionalities are enabled:
  - save as draft
  - switch to classic editor
  - switch to HTML mode
  - preview
  - post settings
- check that the post can be edited in both the block and the classic editor
- check that the label in the upper left matches the available site of your test user
- tap `Publish`, and check that that an action sheet containing the actions `Publish` and `Keep Editing` appears at the bottom of the screen
- Choose the `Publish` action
- Check that a `Post Published` toast, with the name title of the post and a `view` action appears at the bottom
- Tap that `view` action and confirm that it opens the viewer with the post
- Navigate to your site (either from the reader on my sites) and confirm that the post is present (pulldown to refresh might be needed)

PR submission checklist:

   - Unit tests have been added.

   - Since it's still a WIP feature, it shouldn't be necessary.

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
@@ -25,6 +25,7 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
     func readerCell(_ cell: ReaderPostCardCell, likeActionForProvider provider: ReaderPostContentProvider)
     func readerCell(_ cell: ReaderPostCardCell, menuActionForProvider provider: ReaderPostContentProvider, fromView sender: UIView)
     func readerCell(_ cell: ReaderPostCardCell, attributionActionForProvider provider: ReaderPostContentProvider)
+    func readerCell(_ cell: ReaderPostCardCell, reblogActionForProvider provider: ReaderPostContentProvider)
     func readerCellImageRequestAuthToken(_ cell: ReaderPostCardCell) -> String?
 }
 
@@ -507,6 +508,7 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
     }
 
     fileprivate func configureReblogActionButton() {
+        reblogActionButton.tag = CardAction.reblog.rawValue
         reblogActionButton.isHidden = !shouldShowReblogActionButton
     }
 
@@ -622,16 +624,18 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
     }
 
     @IBAction func didTapActionButton(_ sender: UIButton) {
-        if contentProvider == nil {
+        guard let contentProvider = self.contentProvider else {
             return
         }
 
         let tag = CardAction(rawValue: sender.tag)!
         switch tag {
         case .comment :
-            delegate?.readerCell(self, commentActionForProvider: contentProvider!)
+            delegate?.readerCell(self, commentActionForProvider: contentProvider)
         case .like :
-            delegate?.readerCell(self, likeActionForProvider: contentProvider!)
+            delegate?.readerCell(self, likeActionForProvider: contentProvider)
+        case .reblog:
+            delegate?.readerCell(self, reblogActionForProvider: contentProvider)
         }
     }
 
@@ -652,6 +656,7 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
     fileprivate enum CardAction: Int {
         case comment = 1
         case like
+        case reblog
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
@@ -624,11 +624,11 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
     }
 
     @IBAction func didTapActionButton(_ sender: UIButton) {
-        guard let contentProvider = self.contentProvider else {
+        guard let contentProvider = self.contentProvider,
+            let tag = CardAction(rawValue: sender.tag) else {
             return
         }
 
-        let tag = CardAction(rawValue: sender.tag)!
         switch tag {
         case .comment :
             delegate?.readerCell(self, commentActionForProvider: contentProvider)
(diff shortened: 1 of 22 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -622,16 +624,18 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
     }
 
     @IBAction func didTapActionButton(_ sender: UIButton) {
-        if contentProvider == nil {
+        guard let contentProvider = self.contentProvider else {
```
ScoutHarris: 👍 Thank you!

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
```diff
@@ -0,0 +1,127 @@
...
+    /// Presents the reblog screen(s)
+    func presentReblog(blogs: [Blog],
+                       readerPost: ReaderPost,
+                       origin: UIViewController) {
+
+        switch blogs.count {
+        case 0:
+            break
+        case 1:
+            let post = postService.createDraftPost(for: blogs[0])
```
ScoutHarris: Maybe `blogs.first` might be better?

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
```diff
@@ -0,0 +1,127 @@
...
+            content = ReblogFormatter.wordPressQuote(text: summary, citation: citation)
+        }
+        // insert the image on top of the content
+        if let image = readerPost.featuredImage {
+            content = ReblogFormatter.htmlImage(image: image) + content
+        }
+        self.content = content
+    }
+
+    private func update(with readerPost: ReaderPost) {
```
ScoutHarris: This `private` isn't necessary since it's in a `fileprivate extension`.

WordPress/WordPressTest/WordPressTest-Bridging-Header.h
```diff
@@ -6,3 +6,4 @@
 #import "TestContextManager.h"
 #import "TestingAppDelegate.h"
 #import <OHHTTPStubs/OHHTTPStubs.h>
+#import "BlogService.h"
```
ScoutHarris: If you add `@testable import WordPress` in `ReaderReblogActionTests`, this shouldn't be necessary.

ScoutHarris:
Hey @Gio2018 . This looks really nice!

I would still like @osullivanchris 's input (once screenshots/videos are added), and I'd like @frosty to give it a look.  But so far 👍 !

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -622,16 +624,18 @@ fileprivate func < <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
...
             return
         }
 
         let tag = CardAction(rawValue: sender.tag)!
```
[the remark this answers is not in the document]
Gio2018: Sure! I should have seen this

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
Gio2018: I know, the direct index does not look great. On the other hand, first returns an optional, which we can safely force unwrap in this case, but force unwrapping does not great too :) Which one do you prefer?

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
Gio2018: Oops.. yes

WordPress/WordPressTest/WordPressTest-Bridging-Header.h
Gio2018: That's right. Removed

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
ScoutHarris: Well, `blogs[0]` assumes the array is not empty, and this will crash if it is. It may be an edge case, but I think I'd be safer to ensure there is actually a blog here.

Gio2018: ok!

ScoutHarris:
I'm going to go ahead and merge this. If any changes are needed, they can be made later.

PR #13075, opened 2019-12-11
Adds some site creation UI improvements

emilylaguna:
Fixes #11686 subtask: Update the `SiteAssemblyContentView` UI

**Changes:**
- `AppImages.xcassets`
  - Added a new imageset `site-creation-loading` that is displayed while the user is waiting for their site to be created in both dark and light modes

- `SiteAssemblyContentView`:
  - Added 3 new properties:
    - `statusTitleLabel` and `statusSubtitleLabel`: Display the title and subtitle copy specified here: https://github.com/wordpress-mobile/WordPress-iOS/issues/11686#issuecomment-558860091
    - `statusImageView`: Displays the new `site-creation-loading` image

**To test:**
- Launch the app
- Go to the 'My Sites' tab
- Go to the 'Switch Sites' view
- Tap the + button to add a new site
- Tap the `Create WordPress.com site` option
- Tap any site option
- Tap Skip to skip the about
- Tap Skip to skip the basic information
- Search for any domain and select it
- Tap the 'Create Site' button

Example expected screenshots

| Device  | iOS Version | Mode | Screen |
|----------|----------|----------|----------|
| iPhone 5s  |  11.4 | N/A - Light | |
| iPad Air 2 | 12.4 | N/A - Light |  |
| iPhone 11 Pro | 13.2.2 | Light |  |
| iPhone 11 Pro | 13.2.2 | Dark |  |
| iPad Pro (12.9 inch) | 13.2.2 | Light |  |
| iPad Pro (12.9 inch) | 13.2.2 | Dark |  |

**Other notes:**
- I noticed there was the `TitleSubtitleHeader` class that already handles the title/subtitle flow which I initially tried to use, but in practice the constraints that were being added to it internally didn't match the design I was given so I reverted back to my own title/subtitle properties

**PR submission checklist:**
- [X] I have considered adding unit tests where possible.

- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13101, opened 2019-12-13
Site Creation -> Loading: Adds a new "rotating message:" class

### LINKED ISSUE #11686: Site Creation: Improve "creating new site" screen.
## Description:

When creating a site through our New Site Creation flow, the screen that tells the user the site is being created could use some design improvements.

## Image:

IMG_93155CCFF54F-1

## Matching Android Issue:

https://github.com/wordpress-mobile/WordPress-Android/issues/9854

emilylaguna:
Depends on: https://github.com/wordpress-mobile/WordPress-iOS/pull/13075
Fixes #11686 subtask: Create a separate view to manage the cycling/rotating messages

**Changes:**

Adds: `SiteCreationRotatingMessageView`
Adds: `SiteCreationRotatingMessageViewTests`

**Decisions**
- When approaching a solution to this issue I looked at the UI and the animation video and opted to use a UIStackView to manage the spacing between the icon and the label instead of having to do it manually with say just a string, this also allows us to add/remove new views in the future easier. 
- Initially I had the `SiteCreationRotatingMessageView` managing itself but when adding unit tests I decided to allow the class to be configured with an array of status messages, and the icon to display next to them
- The animation is a simple UIView alpha change 

**To test:**

Note: There are no UI changes in this pull request to view
- Launch the Xcode Project
- Hit Command-U or Product->Test

**PR submission checklist:**


WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
```diff
@@ -0,0 +1,189 @@
...
+        }
+    }
+
+    /// An array of status messages to rotating through
+    private(set) var statusMessages: [String]!
+
+    // MARK: - State Management
+
+    /// A timer that determines how long to show each message
+    internal var animationTimer: Timer!
```
leandroalonso: Out of curiosity, any special reason to explicit declare the `internal` access control?

I guess we can change this one and `visibleIndex` to `private(set)`. What do you think?

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
```diff
@@ -0,0 +1,189 @@
...
+        private static let statusDisplayDuration = TimeInterval(2)
+
+        //Adjust the total display time to take the fade into consideration
+        static var totalStatusDisplayDuration: TimeInterval {
+            return statusDisplayDuration + transitionAnimationDuration
+        }
+    }
+
+    /// An array of status messages to rotating through
+    private(set) var statusMessages: [String]!
```
leandroalonso: Since this is a property we're setting on the constructor we can remove the `!`
```suggestion
    private(set) var statusMessages: [String]
```

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
leandroalonso: This is a property that we change to `nil` at some point, it will be safer to change that to an Optional (and the code would still work without any other change).
```suggestion
    internal var animationTimer: Timer?
```

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
```diff
@@ -0,0 +1,189 @@
...
+            if nextIndex >= statusMessages.count {
+                nextIndex = 0
+            }
+
+            return nextIndex
+        }
+    }
+
+    // MARK: - View Properties
+    private var statusImageView: UIImageView!
```
leandroalonso: No need for the `!`

```suggestion
    private var statusImageView: UIImageView
```

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
```diff
@@ -0,0 +1,189 @@
...
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Public Methods
+    func startAnimating() {
+        stopAnimating()
+
+        self.animationTimer = Timer.scheduledTimer(timeInterval: Parameters.totalStatusDisplayDuration,
```
leandroalonso: We can remove the `self`.

```suggestion
       animationTimer = Timer.scheduledTimer(timeInterval: Parameters.totalStatusDisplayDuration,
```

There are other `self` that we could remove without any side effect,

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /// Updates the status label/accessiblity label with the provided text
+    /// - Parameter message: The text to be displayed
+    internal func updateStatus(message: String) {
+        let statusMessage = NSLocalizedString(message,
+                                              comment: "User-facing string, presented to reflect that site assembly is underway.")
+
+        self.statusLabel.text = statusMessage
+        self.statusLabel.accessibilityLabel = statusMessage
```
leandroalonso: Since this is a plain `UIlabel`, the `accessibilityLabel` will have the text content. I guess we can remove this line.

WordPress/WordPressTest/SiteCreation/SiteCreationRotatingMessageViewTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    override func tearDown() {
+        self.rotatingMessageView = nil
+
+        super.tearDown()
+    }
+
+    /// Make sure the view was properly configured with the provided status messages
+    func testSiteCreationRotatingMessageView_Initialize() {
+        XCTAssertEqual(rotatingMessageView?.statusMessages, testMessages)
```
leandroalonso: Maybe we can get rid of this test?

The test below checks if the messages are appearing on the `statusLabel` — which is what really matters to us. The `statusMessages` is just an implementation detail.

That makes sense?

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
emilylaguna: Oops, habits from writing Objective-C

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
emilylaguna: No specified reason other than an oversight on my part. I've changed both of those to `private(set)`

WordPress/Classes/ViewRelated/Site Creation/FinalAssembly/SiteCreationRotatingMessageView.swift
emilylaguna: Okay! I just mimicked it from the `SiteAssemblyContentView`

WordPress/WordPressTest/SiteCreation/SiteCreationRotatingMessageViewTests.swift
emilylaguna: Makes sense to me! My thinking was checking to see if the class was initialized correctly with the given status messages.

WordPress/WordPressTest/SiteCreation/SiteCreationRotatingMessageViewTests.swift
```diff
@@ -0,0 +1,73 @@
...
+        super.tearDown()
+    }
+
+    /// Test to make sure the statusLabel text and accesibility labels are being set correctly
+    func testSiteCreationRotatingMessageView_StatusUpdate() {
+        let message = "This is a test!"
+        rotatingMessageView?.updateStatus(message: message)
+
+        XCTAssertEqual(rotatingMessageView?.statusLabel.text, message)
+        XCTAssertEqual(rotatingMessageView?.statusLabel.accessibilityLabel, message)
```
leandroalonso: This line is failing — since we removed manually setting the `accessibilityLabel`. Let's remove it!

PR #13120, opened 2019-12-16
Fix retention cycles

### LINKED ISSUE #13097: Fix retention cycles
Due to extensive usage of callbacks, we're having problems with retention cycles.

This can be checked using Instruments (Allocations).

leandroalonso:
Fixes #13097

Depends on https://github.com/wordpress-mobile/WordPress-iOS/pull/13085

### To test

1. Open Instruments
2. Run "Allocations"
3. Run the app
4. Select the app under Your Simulator Device > Running Applications: WordPress
5. Tap "Record" (the button with a red circle)
6. In the bottom sidebar, type "mediaed"
7. Open and close the Media Editor multiple times
8. In Instruments, check that the MediaEditor classes appear and then disappear when it's closed

Plus: test typing "tocrop" and checking that the crop view controller is correctly deallocated.

PR submission checklist:


diegoreymendez:
@leandroalonso - In order to make this review easier I'm going to merge from `issue/media_editing_basics_phase_one` into this branch, so that I only see the changes introduces by this PR.

I apologize in advance for taking the liberty of doing this, but the idea is not to slow things down while you're AFK.

I hope that's fine @leandroalonso !

WordPress/MediaEditor/Capabilities/Crop/MediaEditorCropZoomRotate.swift
```diff
@@ -71,3 +65,13 @@ class MediaEditorCropZoomRotate: MediaEditorCapability {
         }
     }
 }
+
+extension MediaEditorCropZoomRotate: TOCropViewControllerDelegate {
+    func cropViewController(_ cropViewController: TOCropViewController, didFinishCancelled cancelled: Bool) {
+        onCancel()
```
diegoreymendez: I'm slightly confused by `onFinishEditing()` not being called here when `cancelled = false` but, again, if this is intentional, a quick explanation and marking this as "Resolved" will do for me.

leandroalonso: `onFinishEditing()` is called when the user made some change in the image and tap "Done".

In this case, the user is simply dismissing the crop view controller. That's why we call `onCancel()`.

In `AztecPostViewController.swift` there's only an `onFinishEditing` block, but we can also add an `onCancel`. We don't have it because if the user cancel we simply do nothing.

Does that make sense?

Issue #12971, opened 2019-11-19
[VoiceOver] Stats issues reported by user


shiki:
Reported on WordPress Slack.

> On the Insights tab, there are a lot of unlabeled buttons, many of them in my case even grayed out, which don‘t tell what they are.

> Also in Statistics, on the Days view, it would be awesome if each item in the bar chart would say what the number stands for. Like page impressions vs. visitors. I figured out that for each day, the first item is the number of page hits, the second is the number of individual visitors. Other than that, these views talk really nicely, like the description of the bar chart itself, the different table headers etc. below, all very nicely done.

Issue #11995, opened 2019-06-25
[Stats Refresh] Voiceover audit


frosty:
It's important that our new stats refresh is accessible for users who rely on VoiceOver. At the moment, it doesn't function as well as the previous version of stats. Let's do an audit, and note here any issues that need resolving. Note that these are not necessarily listed in priority order.

- [x] View More rows in stats tables are not read correctly
- [x] Reflect expanded / collapsed state of expandable rows
- [x] "No data for this period" rows are not read correctly
- [ ] Expanded rows are not accessible

### Date selector bar

- [x] Perhaps prefix the current period with text hinting that it is the currently displayed period. e.g. "Current period: 25 June 2019". Or as a hint: "25 June 2019. The current period displayed."
- [x] Date bar arrows just say 'button', but should describe their function.
- [x] The grey spacer row below date bar is highlightable / selectable. Would be good if we could skip it.

### Period Chart

- [ ] Difference label should hint that it's compared to the previous period.

### Insights

- [x] Heading for "Latest post summary" reads as "Latest P.O.S.T. summary". We may need to lowercase the accessibility description (#12201)
- [ ] Ordering of items in the latest post summary block is incorrect. It skips from the summary to View More, and then back to the items in the middle. "View More" is also read when VoiceOver summarises the entire block, in between the view count and the word "Views". (@frosty)
- [x] The Today block and the All Time Block have an invisible button at the bottom that gets highlighted – we should skip this.
- [ ] Annual site stats block isn't read correctly.
- [x] Most popular time block could improve the description of the most popular time – for me it reads "14, 5% of views". "2pm" would be better than 14 when reading it. (#12201)
- [x] #13123 Posting activity could provide a better description, even if it's just "graphic depicting days when posts were published over the last 3 months". When navigating through the graphic itself, we should improve the description of each day if possible (currently just says 'dimmed button'). Ideally: skip empty days, and note the number of posts on the days with posts. I don't know if we have this info available or not. 
- [ ] Posting Activity: Improve #13123. Possibly add a visual indicator of what date is currently being spoken. The red selection is not happening when you navigate using VoiceOver.
- [ ] Posting Activity: From #13123. The container's accessibilityLabel could describe what you can visually see. For example, “December 2019. High posting activity“. Or perhaps the total number of posts for the month?
- [ ] Reading order for Comments table is slightly off – table and rows are read before the tabs at the top. Same for Followers table (same style).
[...]

PR #13123, opened 2019-12-17
[VoiceOver] Improvements for Stats Posting Activity 

shiki:
Refs #12971, #11995.

This improves the accessibility of the [`PostingActivityMonth`](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Stats/Insights/Posting%20Activity/PostingActivityMonth.swift) component. 

Currently, VoiceOver users are able to access all the days within the component. This is a bit problematic if the user tries to navigate the page by swiping right. They would have to go through one hundred tiny UI elements before they can reach the bottom of the page. 

This is an attempt to make this better. 

Before | After 
--------|-------
 [before video (15s)](https://drive.google.com/file/d/12hC_VO-1BvJQikFvS4C3bXbZDBppuPF5/view?usp=sharing)       |       [after video (23s)](https://drive.google.com/file/d/12mFSze8u0pvlwhpkSsGw0Kc8WXFBMx-V/view?usp=sharing)
Everything inside is accessible | Only the container is accessible
_ | You can swipe up or down to access more data

I'm not sure about [preselecting the first day](https://github.com/wordpress-mobile/WordPress-iOS/blob/7ab48a2f01326bedb6e4d20012a91b109533b025/WordPress/Classes/ViewRelated/Stats/Insights/Posting%20Activity/PostingActivityMonth.swift#L195). But please let me know what you think. 🙂 

## Possible Enhancements

There are still more that can be done here. Some of my ideas are:

- The container's `accessibilityLabel` could describe what you can visually see. For example, “December 2019. High posting activity“. Or perhaps the total number of posts for the month? 
- Visual indicator of what date is currently being spoken. The **red** selection is not happening when you navigate using VoiceOver. 

## Testing

1. Turn VoiceOver on. 
2. Navigate to Stats → Insights.
3. Swipe right until you reach a month in the _Posting Activity_ section. 
4. When a month is selected, what do you think about what's being spoken by VoiceOver?
5. Swipe up or down to navigate through the dates with posts in them. 
6. What do you think about this navigation? 

Please do similar testing in Stats → Insights → Posting Activity.

## Reviewing 

## Submitter Checklist

- [ ] If it's feasible, I have added unit tests. 
- [ ] Add unit tests if the proposed changes are acceptable

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Insights/Posting Activity/PostingActivityMonth.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/Posting Activity/PostingActivityMonth.swift
@@ -157,6 +157,13 @@ private extension PostingActivityMonth {
 private class PostingActivityMonthAccessibilityElement: UIAccessibilityElement {
 
     private var events = [PostingStreakEvent]()
+    private var selectedIndex: Int = 0
+
+    private lazy var monthDayFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "MMMM d"
+        return formatter
+    }()
 
     override var accessibilityFrameInContainerSpace: CGRect {
         get {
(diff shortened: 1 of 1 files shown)


ScoutHarris:
Hey @shiki . This is sooooo much better! 👍  My only comment - 

> I'm not sure about preselecting the first day. But please let me know what you think.

My 2 cents - I like it better without. It felt like it was reading too much information when a month was selected. I disabled it, and I think it's clearer what's happening. 

What do you think?

shiki:
@ScoutHarris Thank you for reviewing! 

> My 2 cents - I like it better without. It felt like it was reading too much information when a month was selected.

Yeah, I'll remove it. I wasn't really sure about it. 

The only thing that bothers me is when: 

1. The user selects a month. “December 2019“ is spoken. 
2. The user swipes up. “December 3. 1 post“ is spoken.
3. The user selects a different month **and then** goes back to the December element. “December 2019. December 3. 1 post“ is spoken. 

This is because the `accessibilityValue` was previously set when the user swiped up the first time. I am not sure if it's possible to _reset_ it when the element is focused by VoiceOver. But _I think_ this is okay and would not be a surprise to the user since they initiated the change anyway. 🤔

ScoutHarris:
> But I think this is okay and would not be a surprise to the user since they initiated the change anyway.

I think that's acceptable.

PR #13124, opened 2019-12-17
Post Reblogging - Users with no available sites and users with more than one available site.

Gio2018:
### This PR extends the post reblogging feature, covering two additional use cases
- #### users with no available site
- #### users with more than one available site

Ref: #12960

### Description
This PR extends the reblog feature to all users, covering the two use cases listed above.
The site picker will be presented to users with more than one site when they tap on `reblog`, so they can choose the site they want to reblog the selected post.
Users with no available site will be presented a `NoResultsViewController` that will give them the opportunity to be redirected to `My Sites` where they can add a new site.
The flow reflects the design below:

version2

as discussed in the design discovery for this use case, the following subtitle has been added to NoResultsViewController

> Once you create a site, you can reblog content that you like to your own site

This PR also adds the selection of the editor (classic editor or block editor) based on the preference set by the user: the editor will be either classic or block, depending on what the user has set on the current site. It will always be possible to switch between the two, as long as the block editor is available.

The videos below illustrate how the feature works

#### (left) reblogging a post - (right) choose site

Publish      MultipleSites

### 

#### (left) choose editor - (right) no available site

SelectEditorType NoSites
### Change log
- `ReblogPresenter` now contains:
  - a `presentEditor(_ :, _ :, _ : )` method, that will present `EditPostViewController` injected with the reblog post content. Depending if the user has one or more sites, the blog picker will be displayed or not
  - a `presentNoSiteScene( _ : )` method, that will present the no sites flow
- `ReblogFormatter` has been partially refactored to allow formatting content for either Classic or Block editor
- Since `ReaderReblogAction.swift` was increasing in size, has been broken up into three files:
  - `ReaderReblogAction.swift`: contains the main Type (and will have an extension for analytics when it will be added)
  - `ReaderReblogAction+Formatter`: contains the formatter
  - `ReaderReblogAction+Presenter`: contains the presenter and the `Post` extension
- A boolean parameter `postIsReblogged` has been added to `PostEditor`. As a consequence, both `AztecPostViewController` and `GutenbergViewController` now implement this parameter, defaulting it to `false`. This parameter is used to determine if the post comes from a reblogging or not. If it does, the content will be kept unaltered when switching between sites, as opposed to a normal draft post, that (on Aztec) loses its media content, and strips media-related html (like ` Void)?
  -  dismissButtonHandler: (() -> Void)?
  - These are injected in the related action buttons. There is already a `NoResultsViewControllerDelegate` for this purpose, but since `ReaderReblogAction()` acts as `fire-and-forget`, conforming `ReaderReblogAction()` to `NoResultsViewControllerDelegate` would not work, as the reference would not hold (ReaderReblogAction() has already been deallocated when the `action` or `dismiss` buttons are tapped). Other options could be:
      - Hold a reference to `ReaderReblogAction()` in the UIViewController of origin (either `ReaderStreamViewController` (by means of `ReaderPostCellAction`) or `ReaderDetailViewController`
      - Conform the aforementioned  `UIViewControllers` to `NoResultsViewControllerDelegate`: these two already conform to that protocol and have logic for it, so it would be necessary to apply conditional logic
  - Since these two options seem more complicate, I choose the handlers, as they do not interfere with the existing logic and it's easy to understand what they do.

### To test:
1. Make sure you have two test users:
    - one with more available sites
    - one with no available sites
2. Build & Run
3. Log in with the one that has multiple sites (update `wpcom_app_credentials` as needed)
4. Go to `My Sites` and in the `Settings` turn on `Use Block Editor`
5. Go to the `Reader`, select a site of your preference from the `STREAMS` section and tap the `reblog` button on a post (alternatively, enter the post detail and tap the `reblog` button)
6. Check that the blog selector is displayed
7. Select a site (or hit cancel) and check that the block editor is shown
8. Recall the blog selector tapping on the blog name (upper left), and switch between available sites.
9. Check that the post content remains unaltered
10. Find the `switch to classic editor` in the `more` (ellipsis) menu in the upper right corner
11. Verify that the classic editor is displayed with the same content
12. Publish the post and verify that the `view` toast is displayed
13. Go back to `My Sites`, turn off `Use Block Editor`
14. Repeat step 5 and check that the classic editor is displayed now
15. Repeat steps 7-12 (or a subset of it). For the current site, the `switch to block editor` option should not be available
16. Switch to the user with no available sites (update `wpcom_app_credentials` as needed)
17. Go to the `Reader` and, either from the post list or the post detail, tap on `reblog`
18. Check that a `NoResultsViewController` is displayed, with
  - title: `No available sites`
  - subtitle: `Once you create a site, you can reblog content that you like to your own site`
  - action button label: `Manage sites`
19. Tap on `Manage sites` and check that you get redirected to `My Sites`
20. Check that `My Sites` displays the typical `NoResultsViewController` for users with no sites (see design)
21. Tap the `Add new site` button and check that the options for creating a new `WordPress.com` site or adding an existing one are available in the action sheet.

#### PR submission checklist:

  - Unit tests have been added.
  - Since it's a WIP feature, it should not be necessary.


WordPress/Classes/ViewRelated/Reader/ReaderReblogAction+Formatter.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    static func aztecImage(image: String) -> String {
+        let imageInHtml = htmlImage(image: image)
+        return embedInParagraph(html: imageInHtml)
+    }
+}
+
+
+// MARK: - Gutenberg formatter helpers
+extension ReblogFormatter {
```
ScoutHarris: Just a suggestion - if this were a `private` extension the methods wouldn't need to be explicitly `private`.

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction+Presenter.swift
```diff
@@ -0,0 +1,130 @@
+/// Presents the approptiate reblog scene, depending on the number of available sites
```
ScoutHarris: Nitpick - `approptiate` is misspelled.

WordPress/Classes/ViewRelated/Views/NoResultsViewController.swift
```diff
@@ -197,15 +201,15 @@ import Reachability
...
 
     /// Public method to show a 'Dismiss' button in the navigation bar in place of the 'Back' button.
-    ///
-    func showDismissButton() {
+    /// Accepts an optional title, if none is provided, will default to 'Dismiss'
+    func showDismissButton(title: String? = nil) {
         navigationItem.hidesBackButton = true
 
-        let dismissButton = UIBarButtonItem(title: NSLocalizedString("Dismiss", comment: "Dismiss button title."),
+        let dismissButton = UIBarButtonItem(title: title ?? NSLocalizedString("Dismiss", comment: "Dismiss button title."),
```
ScoutHarris: To reduce redundancy, maybe we can create the title once and use that in both places? 

Ex:
`let buttonTitle = title ?? NSLocalizedString("Dismiss", comment: "Dismiss button title.")`

Gio2018:
> Hey @Gio2018 . This is looking great!
> 
Thanks!!
> In addition to the specific code comments, I also noted these.
> 
> * An image placeholder is shown in the editor when the original post doesn't have an image.
> 
> no_image
> 
Yes, in fact I should have also checked that `featuredImage` is indeed a valid image url. The `isImageURL()`method in `WordPressShared` comes in handy for the purpose :) 
> * This message appears in the console log when switching editors via the options menu in the editor:
>   `!!! Attempted to create a revision of a revision`

This message is an info (DDlogInfo), and it is a result of calling `post = self.post.createRevision()` #line 442, #method createRevisionOfPost(_ : ) in `PostEditor+Publish.swift`.

This message means that, if a post already contains a revision to edit, the call will just return the same post (and pass it to the new editor) instead of creating a new one.

This message appears when switching between editors in all cases (reblogged post, new post, edit existing post), with the exception of the first switch in a new post.

This is because, for a new post, the original content will initially be empty, so when switching for the first time, the check
`if post.isLocalRevision, post.original?.postTitle == nil, post.original?.content == nil` (#line 405) will pass and the revision will be applied to the post.

The reblogged and existing posts, instead, already have an original content.

Since the three exclamation seemed a little too alarming for an info message, i took the liberty to change this one and the following one in the same method (`createRevision`, `AbstractPost.m`, #line 148 & 152)
> * Regarding these files: `ReaderReblogAction+Formatter` and `ReaderReblogAction+Presenter`:
>   The `+` in a filename usually indicates an extension of the root. However, I don't see that these are extending `ReaderReblogAction`. Maybe we can just remove the `+` from the names?
Yes those need to be changed. For consistency, I renamed ReblogFormatter to `ReaderReblogFormatter` and `ReblogPresenter` to `ReaderReblogPresenter`.

WordPress/Classes/ViewRelated/Views/NoResultsViewController.swift
Gio2018: Totally

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction+Presenter.swift
Gio2018: Thanks

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction+Formatter.swift
Gio2018: Done!

osullivanchris:
All the flows shown in the Gifs are looking good for me! Nice work. One small thing if not already covered - I think the 'Manage Sites' may have been my suggestion? If so, lets just make sure we have translations for that string or else find a similar string we already use. 

As far as I can see, this is a new flow otherwise made of all existing views. So I don't need to dig into UI details. But if I'm wrong, let me know! :)

Gio2018:
> All the flows shown in the Gifs are looking good for me! Nice work. One small thing if not already covered - I think the 'Manage Sites' may have been my suggestion? If so, lets just make sure we have translations for that string or else find a similar string we already use.
> 
> As far as I can see, this is a new flow otherwise made of all existing views. So I don't need to dig into UI details. But if I'm wrong, let me know! :)

Hi @osullivanchris! Yes the localizable string for "Manage Sites" has been included. Than

> All the flows shown in the Gifs are looking good for me! Nice work. One small thing if not already covered - I think the 'Manage Sites' may have been my suggestion? If so, lets just make sure we have translations for that string or else find a similar string we already use.
> 
> As far as I can see, this is a new flow otherwise made of all existing views. So I don't need to dig into UI details. But if I'm wrong, let me know! :)

Hi @osullivanchris ! I just checked and it looks like translations for "Manage Sites" are in there. Thanks!!

PR #13128, opened 2019-12-17
Post Scheduled Date Time Zone Conversions

### LINKED ISSUE #12992: Post Scheduling: post date doesn't match messaging
From release [testing](https://wp.me/p5T066-15Z-p2/#comment-3906):

Posts seem to be  scheduled using the device timezone not the site settings timezone. This is consistent with previous versions (I tested with 13.7 TestFlight beta), however, it doesn't match with the note displayed below the datetime picker.

To test:

<ol>
<li>Go to Settings > Timezone and set it to something a few hours later than the device timezone.</li>
<li>Schedule a new post for a few minutes in the future compared to the device time.</li>
<li>Notice that the Date and Time picker states the post will be published according to the site timezone.</li>
<li>Wait until the scheduled time passes according to the device timezone.</li>
<li>Observe that the post is published according to the device timezone not the site timezone.</li>
</ol>

postdate

Ref: https://github.com/wordpress-mobile/WordPress-iOS/pull/12820

### LINKED ISSUE #13080: Posts are getting published with the incorrect timezone on self-hosted sites
For a self-hosted site:

1. On the device, go to Settings > General> Date & Time and change the time zone to Tokyo.
1. On the web, go to My Sites > Settings > Timezone and change the time zone to Tokyo. 
1. Log out of the app.
1. Log in to a self-hosted site by entering your site address.
1. Start a new post and add a title and content.
1. Go to More > Post settings and back date the post to 3 days ago at noon.
1. In a browser, open the post in WP Admin and check to see if the date and time are correct.

**Result:** the timestamp for the published post is set to UTC and not to the site settings timezone as expected. This happens with both the Aztec and the block editor.

IMG_3195 IMG_3196 IMG_3197

<sup>(internal reference: p4a5px-2tM-p2 /hat tip @melindahelt)</sup>

### LINKED ISSUE #13132: Post Published date: make consistent with Post Scheduling
With [Post Scheduling Enhancements](https://github.com/wordpress-mobile/WordPress-iOS/pull/12820), the scheduled/published date is now displayed in a new view. However, the `Post Settings` date format does not match that on the `Publish` view. It'd be nice if `Post Settings` adopted the format from `Publish`.

post_settings

post_scheduling

bjtitus:
Fixes #12992
Fixes #13080 
Fixes #13132

This adds date formatters which consider the site's time zone instead of relying on device time zone settings. After this change, times shown in the Settings screen will show the time according to the site's time zone.

<details>
  <summary>Example Post (w/ screenshots)</summary>

Here are some screenshots illustrating the same scheduled post (Scheduled for 10:00AM Central Standard Time, 9:00AM in my local Mountain Standard Time) across the web and iOS products after this change:

Screen Shot 2019-12-17 at 8 31 22 AM

Screen Shot 2019-12-17 at 8 31 04 AM

Screen Shot 2019-12-17 at 8 31 04 AM&nbsp;&nbsp;Screen Shot 2019-12-17 at 8 31 04 AM

</details>

Here are the key changes:

- Added `testDisplayDate` method to test whether the display date is real. Verified this by changing the date formatter to use the incorrect time zone and the test was able to catch it.
- Adds `SiteDateFormatters` to pull blog's time zone from `BlogService`
- Adds new Date Formatter for use in `PostSettingsViewController` **this will change the formatting behavior for that cell regardless of the feature flag**
- Sets the time picker to the current time of the day in the site's time zone

## Reviewing Other Platforms

I checked this behavior against the web client behavior and there are a few things I wanted to point out:

- Typically, the Gutenberg editor shows the date in the site's time zone. One exception to this is when the site's time zone is changed _after_ a post is scheduled. In that case, the timestamp from the original time zone is still shown. I believe this is a bug and could lead to some user confusion since the timestamp is neither in the site time zone or the device time zone. iOS handles this appropriately by showing the date in the site's current time zone.

- The list of posts on the web displays the time in the device time zone. This is consistent with the iOS post list as well, although web uses relative date formats so it's a little clearer when the post is _about_ to be published (i.e. "in 5 minutes").

## Testing

- Set the Site Time Zone to a different Time Zone than the device
- Schedule a post
- Check that the displayed scheduled time matches the site's time zone and is published on that date

## Future Considerations

- The new `SiteDateFormatters` struct could be augmented to automatically cache the date formatters. These settings screens don't have a lot of performance concerns, but caching the formatters might be desirable in the future. The `ActivityDateFormatting` struct is very similar and a more unified and updated date approach
- Gutenberg uses the date and time format set in site settings. I think it would be valuable to use this on mobile somewhere, but it might not fit in the cell. It should be possible to convert the [PHP date format strings](https://www.php.net/manual/en/function.date.php) to [Unicode date format strings](https://www.unicode.org/reports/tr35/tr35-dates.html#Date_Format_Patterns). This would probably be a bigger task to consider for a larger date overhaul.

PR submission checklist:

- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ This is a smaller bug fix for a feature that was not released to production.


ScoutHarris:
Hey @bjtitus . Regarding your comment about release notes - 

> This is a smaller bug fix for a feature that was not released to production.

#13080 is actually a live bug, and if you've fixed that here maybe we could add a release note specifically for that being fixed?

bjtitus:
@ScoutHarris So I went back and checked version 13.4.0 so I think you're right that the incorrect timestamp issue may go back a ways. What do you think of this for the Release Notes?

```
* Fixed a bug that displayed incorrect time stamps for scheduled posts.
```

ScoutHarris:
> What do you think of this for the Release Notes?

That'll work!

WordPress/Classes/Utility/SiteDateFormatters.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  DateFormatters.swift
+//  WordPress
+//
+//  Created by Brandon Titus on 12/16/19.
+//  Copyright © 2019 WordPress. All rights reserved.
+//
```
ScoutHarris: Could you remove this comment block please? We don't normally leave them.

WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
```diff
@@ -156,9 +165,18 @@ private struct DateAndTimeRow: ImmuTableRow {
         let footerText: String?
 
         if let date = viewModel.date {
-            let publishedOnString = date.longStringWithTime()
-            let offsetLabel = viewModel.timeZone?.label ?? NSLocalizedString("Unknown UTC Offset", comment: "Unknown UTC offset label")
-            footerText = String.localizedStringWithFormat("Post will be published on %@ in your site timezone (%@)", publishedOnString, offsetLabel)
+            let publishedOnString = viewModel.dateTimeFormatter.string(from: date)
+
+            let nameStyle: NSTimeZone.NameStyle = viewModel.timeZone.isDaylightSavingTime(for: date) ? .daylightSaving : .standard
```
ScoutHarris: The display format of the timezone has changed. As originally designed and implemented:

Screen Shot 2019-12-17 at 1 21 16 PM

If we revert back to that display, checking `isDaylightSavingTime` should not be necessary.

bjtitus: I thought this format may match the "Site Timezone" setting a little better. i.e. if you set to `UTC-7` it will reflect that as `GMT-07:00` but if you use `Chicago` it will properly reflect that the timezone is in use and would use whatever rules apply to that zone.

bjtitus:
@ScoutHarris Thanks for the feedback! I addressed all of those requests. Should be ready for final review!

PR #13134, opened 2019-12-18
Edit already published images in Gutenberg

### LINKED ISSUE #13100: Edit already published images in Gutenberg
Using Gutenberg:

1. Add an image to the post
2. Edit the image
3. Tap "Done"
4. The image should be uploaded

leandroalonso:
Depends on https://github.com/wordpress-mobile/WordPress-iOS/pull/13085

Fixes #13100

This PRs adds the bridge methods to show the Media Editor from Gutenberg.

### To test

1. In `GutenbergViewController.swift` replace the contents of `gutenbergDidRequestFullscreenImage` for this single line: `gutenbergDidRequestMediaEditor(with: mediaUrl, callback: { _ in })` like this:
```swift
    func gutenbergDidRequestFullscreenImage(with mediaUrl: URL) {
        gutenbergDidRequestMediaEditor(with: mediaUrl, callback: { _ in })
    }
```
2. Run the app, open the block editor and double-tap an image
3. The Media Editor should appear

### Additional information

@etoledom This PRs seems big but it's because it depends on another one. The Gutenberg integration is covered on those files:

* `GutenbergViewController.swift`
* `GutenbergMediaInserterHelper.swift`
* `GutenbergMediaEditorImage.swift`

PR submission checklist:


WordPress/Classes/ViewRelated/Gutenberg/Utils/GutenbergMediaEditorImage.swift
```diff
@@ -0,0 +1,65 @@
...
+    private var tasks: [URLSessionDataTask] = []
+
+    private var originalURL: URL
+
+    private var fullQualityURL: URL? {
+        guard var urlComponents = URLComponents(url: originalURL, resolvingAgainstBaseURL: false) else {
+            return nil
+        }
+
+        urlComponents.query = nil
```
etoledom: This will only work for WPCom.
On Self-Hosted, sized urls can be of this kind: `https://testingsiteetoledom.mystagingwebsite.com/wp-content/uploads/2019/11/beautiful-blur-bright-326055-1-150x150.jpg`.

I'm updating some Gutenberg code so we can get the original URL directly from the bridge delegate 👍

leandroalonso: That's good to know!

Thanks @etoledom!

diegoreymendez:
Considering this includes changes from #13085 I'll wait until that PR is up to review this PR's diff.

WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
```diff
@@ -3498,8 +3485,6 @@ extension AztecPostViewController {
             return
```
diegoreymendez: The reason why this code is here, is no longer intuitive IMHO: https://github.com/wordpress-mobile/WordPress-iOS/pull/13134/files#diff-705f087506da53949ae67cbac0ed8d98L3497-L3499

Two questions:
1. Should this be removed?
2. If not - any chance we can add a note as to why this would make the editor bail out?

leandroalonso: > Should this be removed?

If no actions were made in the image we do not replace it. Otherwise, we will duplicate and re-upload the image. More about this [here](https://github.com/wordpress-mobile/WordPress-iOS/commit/6f1b0f666c7848df6cf7d62e0be13d3b68b5e641).

> If not - any chance we can add a note as to why this would make the image replacement bail out?

Sure!

leandroalonso: Done in 9f502ffdaee34baf4bd4da70cb9ea6034fd1055d

PR #13137, opened 2019-12-18
Add Post Reblogging from Reader

ScoutHarris:
Ref #12960 

Ref PRs:
#13055
#13059
#13094
#13124
#13139
#13164

This encompasses all @Gio2018 's changes that add post reblogging from the Reader list and Reader detail. 

**Reader list:**

reader_list

**Reader detail:**

reader_detail

---
To test:
The reblogging behavior should be the same when initiated from either location, so for the following scenarios (for different site counts) please test from both.

---
**Account with multiple sites**
- When reblog is selected:
  - The site selector is displayed.
  - When a site is selected, the post editor is then shown.
  - The site can be changed from the nav bar.

multiple_sites

---
**Account with one site**
- When reblog is selected:
  - The post editor is displayed immediately.

one_site

---
**Account with no sites**
- When reblog is selected:
  - The `No available sites` view is displayed.
  - Tapping `Manage Sites` displays the site creation flow.

no_sites

---
**Common Behavior**
- While editing the reblogged post: 
  - Change the editor from the options menu. 
  - The post content should update accordingly.

post_options

- After reblogging a post:
  - A toast message is displayed.
  - Selecting `View` on the toast message shows the view post flow.

toast

---
PR submission checklist:


Gio2018:
> Everything looks good to me, there's just one thing I don't have clarity on.
> 
> In the example with one site, when the icon is tapped it opens a full-screen view from the bottom with an X. That is fine. In the example with multiple sites, on top of this it displays the site picker.
> 
> However I can't see exactly in what mode the site picker is displayed. Unfortunately in the GIF a black bar from your emulator appears over the area I'm trying to see! But it looks like the site picker appears sort of like a modal, sort of like a view, but only going up as far as the nav bar; not covering the status bar. It might just be because I can't see it correctly, but I'd like to know what kind of view we are using here; and if its something we already do in the app. Thanks!

Hi @osullivanchris , the site picker displays in the same mode as when you tap on the site name, on the nav bar, to change it. It actually uses the same `editor.displayBlogSelector()` method, which presents the site picker modally.

The image below contains a screenshot with the site picker


Simulator Screen Shot - iPhone 11 Pro Max - 2019-12-19 at 09 48 31

PR #13139, opened 2019-12-18
Analytics tracks for Post Reblogging

Gio2018:
### This PR adds analytics tracks to the Post Reblogging feature

Ref #12960

### To test:
- Build and run the app
- Go to the reader
- Tap on the `reblog` button in the post list
- check that the console contains the following log:
  - `🔵 Tracked: reader_article_reblogged <blog_id: [some_id], post_id: [some_id]>`
- Go back to the post list
- Tap on a post to enter the post detail
- Tap on the `reblog` button
- Check that the console contains the following log:
  - `🔵 Tracked: reader_article_detail_reblogged <blog_id: [some_id], post_id: [some_id]>`

### change log
- added an `OriginType` enum in `ReaderReblogAction` to distinguish between detail and list
- added an extension to `ReaderReblogAction` containing a `trackReblog(_ :, _ : )` method that will execute the track
- `changed the `execute( )` method to take `OriginType` as an additional parameter
### Note: 
there was a missing `break` in the case of the newly added value `WPAnalyticsStatReaderArticleDetailReblogged`, and since it's Objective-C, was falling through to the next value. This PR fixes that as well.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
@@ -1366,6 +1366,7 @@ NSString *const TracksUserDefaultsLoggedInUserIDKey = @"TracksLoggedInUserID";
             break;
         case WPAnalyticsStatReaderArticleDetailReblogged:
             eventName = @"reader_article_detail_reblogged";
+            break;
         case WPAnalyticsStatReaderArticleOpened:
             eventName = @"reader_article_opened";
             break;
--- a/WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
@@ -1,5 +1,9 @@
 /// Encapsulates a command to reblog a post
 class ReaderReblogAction {
+    // tells if the origin is the reader list or detail, for analytics purposes
+    enum OriginType {
+        case list, detail
+    }
 
     private let blogService: BlogService
     private let presenter: ReaderReblogPresenter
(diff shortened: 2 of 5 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
```diff
@@ -17,9 +21,27 @@ class ReaderReblogAction {
...
                                 readerPost: readerPost,
                                 origin: origin)
     }
 }
+
+// MARK: - Analytics
+extension ReaderReblogAction {
+    fileprivate func trackReblog(readerPost: ReaderPost, originType: OriginType) {
+
+
```
ScoutHarris: Can you remove the extra empty line?

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
```diff
@@ -17,9 +21,27 @@ class ReaderReblogAction {
...
         presenter.presentReblog(blogService: blogService,
                                 readerPost: readerPost,
                                 origin: origin)
     }
 }
+
+// MARK: - Analytics
+extension ReaderReblogAction {
+    fileprivate func trackReblog(readerPost: ReaderPost, originType: OriginType) {
```
ScoutHarris: Just `private` should be sufficient.

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
```diff
@@ -1,5 +1,9 @@
 /// Encapsulates a command to reblog a post
 class ReaderReblogAction {
+    // tells if the origin is the reader list or detail, for analytics purposes
+    enum OriginType {
```
ScoutHarris: This name is a little vague. I wonder if something like `ReblogOrigin` or `ReblogSource` would be more obvious? Especially when it's actually used. For example, without context, this `originType` is a bit confusing.

`action.execute(readerPost: readerPost!, origin: controller, originType: .list)`

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
```diff
@@ -1,5 +1,9 @@
...
+        case list, detail
```
ScoutHarris: This is just a style preference, but we tend to declare the cases one per line, like so:
```
case list
case detail
```

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
```diff
@@ -1366,6 +1366,7 @@ + (TracksEventPair *)eventPairForStat:(WPAnalyticsStat)stat
             break;
         case WPAnalyticsStatReaderArticleDetailReblogged:
             eventName = @"reader_article_detail_reblogged";
+            break;
```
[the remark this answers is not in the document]
Gio2018: I was seeing twice "reader_article_opened" and thought what the?.. :)

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
Gio2018: Yes

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
Gio2018: Yup

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
Gio2018: Yes, ReblogSource has more context for sure

WordPress/Classes/ViewRelated/Reader/ReaderReblogAction.swift
Gio2018: 👍

PR #11495, opened 2019-04-18
[Stats Refresh] Period stats: restart queries when period or date changes

ScoutHarris:
Fixes #11353 

When a user performs any of these actions, the currently running queries will be cancelled and new queries run.
- Select another period.
- Select another date.
- Pull to refresh.
(Basically, any action that triggers `SiteStatsPeriodTableViewController.refreshData`.)

To test:
- Go to Stats > Period.
- Perform any of the above actions in quick succession.
- If any queries were still running, you'll see messages like this in the console:
`Error fetching posts: Optional("cancelled")`

PR #13140, opened 2019-12-19
Stats: force refresh when Period changes

### LINKED ISSUE #13138: Stats screens show stale data (often zero)
**Screenshot of the issue:** 

**Video of the issue**
RPReplay_Final1576705697 MP4

****Steps to reproduce****
- Delete and reinstall the app (13.8 or above)
- Log into an account with many sites
- Choose "stats", then remain on "insights" until it loads. Then tap through the other tabs. 
- Sometimes, the number won't change between two tabs – so "days" and "weeks" will show the same data. Other times, a tab will just show "0" at the top. 
- It may be necessary to try this on a few sites before it works.
- It _may_ also be related to persistence – I noticed that if the device froze very briefly when switching tabs (due to core data save?) the issue seemed to happen more.

**Workaround**
Pull-to-refresh seems to consistently show the correct values.

ScoutHarris:
Fixes #13138 

This change forces the period queries to be cancelled and restarted when period data is refreshed. 

I referred to the PR that fixed this problem initially (https://github.com/wordpress-mobile/WordPress-iOS/pull/11495), and it appears somewhere along the way the `forceRefresh` flags got swapped. I simply swapped them back.

To test:
- Got to Stats > Periods.
- Quickly switch between period tabs.
- Verify the tab you end up on displays the correct data.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodViewModel.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodViewModel.swift
@@ -56,7 +56,7 @@ class SiteStatsPeriodViewModel: Observable {
         periodReceipt = store.query(.periods(date: lastRequestedDate, period: lastRequestedPeriod))
         store.actionDispatcher.dispatch(PeriodAction.refreshPeriodOverviewData(date: lastRequestedDate,
                                                                                period: lastRequestedPeriod,
-                                                                               forceRefresh: false))
+                                                                               forceRefresh: true))
     }
 
     func isFetchingChart() -> Bool {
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodViewModel.swift
```diff
@@ -56,7 +56,7 @@ class SiteStatsPeriodViewModel: Observable {
         periodReceipt = store.query(.periods(date: lastRequestedDate, period: lastRequestedPeriod))
         store.actionDispatcher.dispatch(PeriodAction.refreshPeriodOverviewData(date: lastRequestedDate,
                                                                                period: lastRequestedPeriod,
-                                                                               forceRefresh: true))
+                                                                               forceRefresh: false))
```
danielebogo: This needs to be true since it forces to update the siteId for the `statsServiceRemote`

ScoutHarris:
Thanks @danielebogo !

@jkmassel - ready to go!

PR #13148, opened 2019-12-19
[This Week Widget] Add functionality

ScoutHarris:
Ref #13044

This adds full functionality to the This Week widget.

There are a few outstanding UI issues, which will be addressed in a follow-up PR:
- The footer jumps when expanding the widget.
- When the expanded widget is refreshed (ex: scrolling away from then back to Today):
  - The first two rows show the empty view while being refreshed.
  - The footer jumps.

To test:

---
- Start with a fresh install, or be logged out of the app.
- Add the `WordPress This Week` widget to your Today view.
  - Verify the unconfigured view is displayed.
  - Verify the widget cannot be expanded.
  - Verify tapping the widget launches the app.

unconfigured

---
- Log into the app.
- Go to Stats > Widgets > `Use this site`.
  - Verify the This Week widget displays data for the selected site, 2 days in compact and 7 in expanded.
  - Verify the stats match the Days' chart data for each day.
  - Verify the widget can be expanded/collapsed.
  - Verify the footer is _not_ shown in compact, _is_ shown in expanded.

app_stats

compact

expanded

---
- Return to the app, and select a different site.
- Go to Stats > Widgets > `Use this site`.
- Verify the This Week widget updates accordingly.

---
PR submission checklist:


WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetDifferenceCell.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    func configureLabels(day: ThisWeekWidgetDay?, isToday: Bool) {
+        guard let day = day else {
+            initializeLabels()
+            return
+        }
+
+        dataLabel.text = numberFormatter.string(from: NSNumber(value: day.viewsCount)) ?? String(day.viewsCount)
+
+        differenceLabel.text = String.localizedStringWithFormat(Constants.percentFormat,
```
bjtitus: Is it worth using `NumberFormatter` for this?

Something like:

```swift
let numberFormatter = NumberFormatter()
numberFormatter.numberStyle = .percent
numberFormatter.positivePrefix = "+"
numberFormatter.string(for: day.dailyChangePercent)
```

`dailyChangePercent` would remain as the decimal version and the Number Formatter would that piece of the conversion. Rounding would also be taken care of.

In some of the European languages, this adds a space between the number and percent and I think RTL languages would be handled a bit better than the format string.

| English | French | Arabic |
| `75%`  | `75 %` | `٧٥٪` | 

Someone with more language experience might have thoughts on this.

osullivanchris:
Looks good! Only thing that Sylvester spotted which I can see here now too, is that the dividers stand out a little next to widgets from other apps. I don't know how the colors / materials are defined as these widgets are new to me. But similar to the backgrounds it looks like its a system material rather than a color that we should be using for the dividers.

Image from iOS

WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetDifferenceCell.swift
frosty: Sounds like a good idea!

frosty:
Thanks @osullivanchris. Looks like the Shortcuts widget there might be using a system `VisualEffectView` with a vibrancy effect.

osullivanchris:
> @osullivanchris any ideas?

From 0 -> 0 I would have 0% in grey
From 0 -> 4 I would have it green but not sure what the number should be. Could we fake 0 being 1, and apply a percentage based on 1? So 0 to 4 would be the same % as going from 1 to 4? ∞% would also seem ok to me because its factually correct, even though its unusual to see it there.

I imagine with all the test things we create we'll see this 0 ∞% scenario more than real users.

ScoutHarris:
Thanks @osullivanchris @frosty .

> the dividers stand out a little next to widgets from other apps.

All the widgets use the same divider, so I'll look at that separately and fix it for all.

> The second divider was missing
> The footer wasn't positioned correctly.

I haven't seen these, but they'd be directly related to showing/hiding the footer, so I'll lump those in with the Known Issues and address them in the next PR.

> what the correct behaviour should be for the % display if we go from 0 the previous day to some non-zero number the following day

Regarding the colors and numbers, I emulated the behavior in the app when we show this difference. So ☝️ is consistent with the app. Specifically for the colors, the behavior is:
- Today is always grey. The difference doesn't matter. And no other scenario shows grey.
- For previous days:
  - Greater than or equal to 0 is green.
  - Less than 0 is red.

> Could we fake 0 being 1

I'd really rather not fake data.

> Android shows ∞%

I guess they got a different design. :smile: 

I can change this behavior, but I'd prefer to do it in both the app and the widget to be consistent.

ScoutHarris:
Hey @bjtitus . 

> Is it worth using NumberFormatter for this?

Thanks for this. Much better! 

Would you mind giving this a review? As noted above, I'll address the UI issues separately. So the only change I made was to support using the formatter.

ScoutHarris:
Thanks @bjtitus !

osullivanchris:
> All the widgets use the same divider, so I'll look at that separately and fix it for all.

sounds good to me!

> I'd really rather not fake data.

When you put it like that, seems fair!

> I guess they got a different design. 😄

Lol! You seem to know more about it than me so I'm glad to take your decision. Thanks!

PR #13152, opened 2019-12-20
Notifications Refresh

### LINKED ISSUE #13151: Notifications: Adjust Metrics
### Details:

- [x] Adjust **TableViewCell's** horizontal + vertical spacing
- [x] Adjust **TableViewHeader** Metrics
- [x] **Users** + **Post Titles** + **Comments** should now be bold
- [ ] Review Fonts

---

Mockup

strings

jleandroperez:
### Details:
In this PR we're:

1. Mapping a (patched up) version of WordPressShared, which allows us to get scaled fonts with the actual specified traits, rather than always regular.
2. Adjusting Notifications Font Mapping (**bold** for Users, Comments and Posts)
3. Slightly adjusting the Notifications TableView Cell
4. Adjusting the Notifications TableView Header's height

Closes #13151

**Note:** New font size is pending!

### To test:
1. Log into your WordPress.com account
2. Press over the Notifications Tab

- [ ] Verify that Users / Post Titles / Comments are now displayed in bold
- [ ] Verify that the new metrics match the mockup 

PR submission checklist:

---

### Screenshot:

**Yes**, blur on purpose, here and there. But the (new) metrics can be appreciated!

Notifications


jleandroperez:
@frosty Thank you for taking a look!!!

1. Metrics updated (may I bug you to take a second look?)
2. WordPress Shared updated as well (dynamic type should be better, if not completely fixed now)

Only pending I've got is the Header bug you've spotted. I couldn't repro on my end, what simulator / iOS release were you using?

Thank you sir!!

osullivanchris:
Finally taking the time to properly review what @jleandroperez  started here. This is really cool. Thanks for jumping on this and making the iOS version so fast I didn't even have a design yet :D 

Generally the changes you have made are an improvement! Really liking the direction. But I have a few follow ups and additional things to consider on iOS

1. Typography is looking really nice, great job. Agree with @frosty  that it needs spacing between the heading and the comment. I think we should add 4px (@1x) here.

2. In your point 4 you mentioned _"Adjusting the Notifications TableView Header's height"_. Looks good. But I would actually make it 16px above and only 8px below. Currently it looks like 16px both sides. If its closer to the section below it will be more clear that it applies to the list below it. Current and suggested in the below image, hopefully you can see what I mean
Screenshot 2020-01-10 at 17 15 00

3. Not introduced by what you've done but a potential improvement. It appears that the TableView header uses the same background color as the unread message. To me, the TableView header should sit on the standard page background color, while the highlighted row should be a different blue. What do you think?

4. Also on the TableView header, you've changed the text to a heavier weight. Its inconsistent with what we're using for those headers in the rest of the app, so I think we should leave this as the previous regular font weight.

5. We are using dividers that have padding on the left hand side. I believe it makes the colored rows look a little broken. I think the row would look more complete if this line extended to the edge
Screenshot 2020-01-10 at 17 17 46

As mentioned before I originally had just designed this for Android only. I can do some mocks on iOS too if you think it would be useful. But I think you've made a really nice start so I'm happy to tweak as you build if that works for you. Please keep me updated when you get a chance to work on it further.

loremattei:
Hey! I moved this to 14.1 because 14.0 has been cut. Please, ping me if you want it to make it to 14.0 😀

jleandroperez:
@frosty Hey there!! Few notes:

1. Unit Tests fixed (I didn't update OCMock by hand, `pod install` just did that... so I've locked the last stable version there)

2. Header bug should not be happening anymore. Turns out... we were setting the Ratings UI text **after** the layout cycle.

Ready for another look James!!!

---

@osullivanchris Apologies it took me a while to get back to you, got deep lost in a huge refactor elsewhere.  Few notes for you sir:

1. Spacing between labels set as specified (2 points, yields 4px in @1x, etc etc etc)

2. Adjusted metrics in the Section Headers

3. Happy to adjust the colors! buuut (as far as I understand) the current Header BG color is the same standard background we use. May I bug you with a specific location anywhere in the app, where the color we wanna use is currently wired?

4. I'm afraid we didn't really change the Header's text weight. _However_ we did fix a bug with Fonts treatment (fonts weren't properly adapting to the device settings). This could potentially explain any discrepancies. 
  As of now, the Section Title font it's set to `caption1` + `semibold`. Should we make it regular perhaps?

5. Separators updated as requested!!

Attaching a screenshot with all the changes below. Ready for another look Chris, thanks in advance!!!

---

Screenshot

osullivanchris:
Looks good @jleandroperez  thanks for coming back to me. No rush, its not my primary focus either. So 1, 2, 5 are resolved. 

On 3: 
Looking into it - you're right that is always our background color. I think whats up is that the highlight blue should be more of a blue, rather than the same background color. On Android you can see the blue is clearly different from the bg. Do we have a blue color defined like this already on iOS?
Screenshot_20200121-153929


On 4:
I don't want to specify anything particularly to this case - so it should just be the same as the other places we use those section headers. If you can check now that you've fixed it that it matches how it looks on the My Site page, I'm happy as long as they look the same.

Otherwise everything look great! One other idea - do you want to kill the little clock icon? It doesn't add much value, we don't have it on Android, and it might be easier to scan the actual information down the left hand side if it wasn't there!

jleandroperez:
@osullivanchris Clock is gone, loved that change (thank you!!)

### 3. Shades of Blue

We've got the (Muriel I think?) palette there, which already defines Light / Dark modes. Which one should we try for "Unread Notes"?

Screen Shot 2020-01-23 at 10 57 01 AM

### 4. Header BG

I'm afraid I've been thru the app a few times and couldn't really spot other usages of Section Headers.

In the **Me** tab, the headers themselves match the Table's BG color (and this is the only spot in which this component is used), so I'm afraid I'm a bit lost there. Happy to wire any colors you pick, but not sure which one it should be.

(FWIW I think it looks good, screenshots pasted below!)

Thank you!

---

Light
Dark

frosty:
@mattmiklic Are you able to help us out here with color selection?

WordPress/WordPressTest/SiteCreation/SiteSegmentsCellTests.swift
```diff
@@ -45,7 +45,18 @@ final class SiteSegmentsCellTests: XCTestCase {
     }
 
     func testCellTitleIsTheCorrectFont() {
-        XCTAssertEqual(cell?.title.font, WPStyleGuide.fontForTextStyle(.body, fontWeight: .semibold))
+        // SiteSegmentsCell's UILabel has `title.adjustsFontForContentSizeCategory = true`.
+        //  We cannot relaly assert on the Label's font, since it's bond to be changed.
+        //
+        //  XCTAssertEqual(cell?.title.font, WPStyleGuide.fontForTextStyle(.body, fontWeight: .semibold))
+
+        let expectedFont = WPStyleGuide.fontForTextStyle(.body, fontWeight: .semibold)
```
bjtitus: Does it make more sense to be able to pass in the trait collection to `WPStyleGuide.fontForTextStyle` for testing purposes?

Would end up with something like this:
```swift
WPStyleGuide.fontForTextStyle(.body, fontWeight: .semibold, compatibleWith: cell?.title.traitCollection)
```

jleandroperez: @bjtitus Actually, now that you mention it, we might be able to use [this API](https://developer.apple.com/documentation/uikit/uitraitcollection/3238082-performascurrent) to prettify things up a tiny bit.

That being said, personally speaking, IMO, I would remove the test entirely. Unit Testing that a font is set by our UITableViewCell method overrides isn't really valuable (again, IMO!).

Thank you sir!!

frosty:
Looking at this again, regarding the unread notification colors, I think we specifically moved away from blue on iOS 13 so that we could use system grays.

osullivanchris:
re; 3 I chatted with @mattmiklic  and although @frosty is correct 
>I think we specifically moved away from blue on iOS 13 so that we could use system grays.

we think that applies mostly to backgrounds and general UI. But something like this highlight color is a fair place to have special treatment so that it stands out.

We had a look on dark mode and the difference between the highlighted cell color and the page background color is different, so there is no action required on dark mode.

On light mode, we can keep the page background color, but can we change the cell highlight color. I tried a couple, but blue 0 in what you've sent looks too similar to the page background, blue 5 looks too different. I feel like we need something in between to match what we're doing on Android. Some of the issue could just be from working from screenshots. Do you have a magic color in between blue0 and blue5 @mattmiklic ?


re; 4. Apologies. I've been working on My Site / Me so long I said the wrong name. I meant as long as its the same header style visually and in code as the **My Site** page its all good
IMG_6515

mattmiklic:
> On light mode, we can keep the page background color, but can we change the cell highlight color. I tried a couple, but blue 0 in what you've sent looks too similar to the page background, blue 5 looks too different. I feel like we need something in between to match what we're doing on Android. Some of the issue could just be from working from screenshots. Do you have a magic color in between blue0 and blue5 @mattmiklic ?

I'm actually now wondering if the best solution would be to make the entire page background of notifications white. This would let us use the same blue for unread notifications you've implemented in Calypso. In dark mode the background could then be pure black.

jleandroperez:
Gentlemen, I've just opened Issue #13282 to track the pending Color updates. Let's move the fun over there, so that we can merge the current layout changes separately! 😄 

**Thanks everyone!!!**

cc @mattmiklic @osullivanchris

PR #13155, opened 2019-12-23
Fixes crash when the App is launched by certain notifications after being inactive.

### LINKED ISSUE #12669: EXC_BREAKPOINT: release
Sentry Issue: [WORDPRESS-IOS-1PJD](https://sentry.io/share/issue/4bb6106bf01f4a4c93c7e52cccceae36/)

```
EXC_BREAKPOINT: release
  ?, in <redacted>
  ?, in <redacted>
  ?, in <redacted>
  ?, in <redacted>
  ?, in <redacted>
...
(43 additional frame(s) were not displayed)
```
Over the past 14 days: 

~1.7k events
~1.3k affected users

diegoreymendez:
## Description:

Fixes crash when the App is launched by certain notifications after being inactive.

Fixes #12669 

## Before Testing:

You'll need to enable push notifications for our development environment.  This can be done by logging into your WordPress Sandbox, and running these commands:

```bash
cd /home/wpcom/public_html/bin/mobile-push-notification
svn up
php push-notifications-server.php -type=push -qi=10 -app_id=5
```

## Reproducing the original issue:

Just run the testing steps below in our `develop` branch and you should see an exception being raised.

## Testing:

1. Make sure push notifications are running for our dev environment (see "Before Testing").
2. Run the App.
3. Immediately send it to brackground.
4. Comment with another user in one of your test blogs so that a notification is sent to your device.
5. Tap on the notification.

The notification should be opened correctly, and the App should not crash.

## PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 * Post Settings: Added a new Calendar picker to select a Post's publish date
 * Fixed bugs with the "Save as Draft" action extension's navigation bar colors and iPad sizing in iOS 13.
 * Fixes appearance issues with navigation bar colors when logged out of the app.
+* Fixed a bug that was causing the App to crash when the user tapped on certain notifications.
 
 13.9
 -----
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -654,7 +654,7 @@ extension NotificationsViewController {
         // Before trying to show the details of a notification, we need to make sure the view is loaded.
         //
         // Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/12669#issuecomment-561579415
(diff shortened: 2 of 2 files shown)


diegoreymendez:
> Why is that view not loaded? Right before we attempt to show a specific note, we tell the tab bar controller to switch to the notification list. Is UIKit ignoring that until the app becomes active? Should we delay showing the note until the view is loaded normally instead of force loading it?

My initial take was that it's probably just views loading lazily when possible.  I tried analyzing it more today but haven't reached a conclusion.

That said, I'm not sure it's worth it to spend more time on this, considering that even if we did something else to force load the views earlier, I'd still suggest we keep this change since it's a pretty good safeguard to avoid this crash from coming up again.

diegoreymendez:
> Otherwise the crash would have been there... for years!

Do you think it would be worth it to keep digging more?  I'm not sure it'd be efficient to look for an alternative solution when this one does the trick... it could be just that Apple changed how / when views are loaded and we'd be chasing ghosts.

If we feel there's more to gain from looking deeper I don't mind to keep going, but my modest opinion is that digging deeper will just yield diminished returns as the issue is already solved by this PR.

PR #13164, opened 2019-12-30
Post Reblogging changes: reblog when not logged in, fixed labels in portrait mode, fixed editor type when picking a site.

Gio2018:
## This PR addressed the changes requested in the final PR  #13137 for post reblogging 

Master Issue - Ref: #12960

### Description
This PR addresses the following changes:

> I don't think we need to include the check for loggedIn here. Likes and comments require a login because they have to be associated with a dotcom account, but I can't see a reason a user with a self-hosted site (so they won't be logged into WordPress.com) couldn't use reblogging.

- the check for logged in users has been removed

> On the post list in the Reader, rotate to landscape and back and the labels show (but don't fit). This may have been a pre-existing issue but it's more obvious now.

- we currently configure the button labels in `traitCollectionDidChange(_ : )`, to detect orientation changes. When the method is called, we compare `superview?.frame.width` with a threshold value of `480`, to decide if we want to display labels or not (a custom comparison operator is implemented in `ReaderPostCardCell.swift` to handle the case of `superview` being `nil`). What happens is that, when `traitCollectionDidChange(_ : )` is called, there is no guarantee that `superview` has already adjusted its frame, so labels could be displayed when we rotate back to portrait from landscape.

  - From the tests I've done, looks like if we adjust the layout when `UIDevice.orientationDidChangeNotification` gets posted, instead of when `traitCollectionDidChange(_ : )` gets called, the problem does not occur.

  - Another option could be to still use `traitCollectionDidChange(_ : )` and  check the horizontal size class after the rotation (instead of the size of `superview`, and display the labels only if it is regular. With this solution:
    - It's guaranteed that we capture the size class after the rotation
    - labels would not show on iPhones when rotating back to portrait (the destination size class will be compact), as intended
    - The code would simplify (no need for custom comparison anymore, nor the magic number of `480`)
    - On the other hand, there would be no 1:1 match with `480`, so on some devices (e.g. iPhone 11 Pro) labels will not show up in landscape (cause the horizontal size class is still compact, even if it's larger that `480`), but this might be acceptable since we have already removed the labels in the detail view?
    - Labels will show in all devices with regular horizontal size class (e.g. iPads portrait or landscape, iPhone 11 Pro Max landscape, etc)

This PR contains the first solution, as it's consistent with the current experience, but it's pretty easy to switch to the second one if needed.

> For users with multiple sites, we don't seem to be respecting their site's setting for whether to use the block editor. I think it looks like we're defaulting to the setting of the the last used site (which may be Aztec) and we display the editor before we display the blog picker list? It would be nice when the user picks a site if we can ensure we're using their editor of choice.

 - The behavior has been changed. We don't default anymore to the last used setting, and the editor will now respect the user setting for that particular site, and the editor will be presented after a selection has been made. 
  - `BlogSelectorViewController` is initialized with no predefined site, and once a site is picked, the `successHandler` will instantiate and present the right editor
  - Some Notes:
    - In order to assign `successHandler` at a later time, and initialize `BlogSelectorViewController` with no predefined `objectID`, the related parameters in the initializer were set to `nullable`
    - the parameter `displaysPrimaryBlogOnTop` was not used so I removed it
    - `openWithBlogSelector` in `EditPostViewController` was not needed anymore, and removed
See animated gif below

reblog_editor

### To test:

For all use cases:

1. Find a test user with a self hosted site (not logged in to WordPress.com) and verify that the reblog button is visible

2.1 Run the app on any iPhone (simulator or device). On smaller iPhones (e.g. SE) the problem was more obvious
2.2 Rotate to landscape, then rotate back to portrait
2.3 Verify that labels don't show

3.1 Find a user with at least two sites
3.2 For each site, go to `My Sites`, then `settings` and make sure that there's at least one site with `Use block editor` enabled and at least one site with `Use block editor` disabled
3.3 Go to `Reader` and choose any site(s) of your preference
3.4 Tap the `reblog` button and verify that the site picker is presented
3.5. Select a site that has the block editor enabled, and verify that the block editor is presented
3.6 Discard the post
3.7 Tap `reblog` again, select a site that has the block editor disabled and verify that the classic editor is presented

PR submission checklist:


ScoutHarris:
Hey @Gio2018 . I'm still reviewing this, but I did notice one thing.

> the check for logged in users has been removed

It needs to be removed from `ReaderDetailViewController` as well.

WordPress/Classes/ViewRelated/Reader/ReaderReblogPresenter.swift
```diff
@@ -42,34 +42,90 @@ class ReaderReblogPresenter {
...
     }
+}
+
 
+// MARK: - No Sites
+private extension ReaderReblogPresenter {
     /// presents the no sites screen, with related actions
-    private func presentNoSitesScene(origin: UIViewController) {
+    func presentNoSitesScene(origin: UIViewController) {
+        // instantiate & configure
```
frosty: Super nitpicky, but I'd say the comments in here aren't really necessary.

WordPress/Classes/ViewRelated/Reader/ReaderReblogPresenter.swift
```diff
@@ -42,34 +42,90 @@ class ReaderReblogPresenter {
...
+        let successHandler: BlogSelectorSuccessHandler = { selectedObjectID in
+            guard let newBlog = blogService.managedObjectContext.object(with: selectedObjectID) as? Blog else {
+                return
+            }
+            navigationController.dismiss(animated: true) {
+                self.presentEditor(with: readerPost, blog: newBlog, origin: origin)
+            }
+        }
+        selectorViewController.successHandler = successHandler
+        // present
```
frosty: Super nitpicky, but I'd say the comments in here aren't necessary.

Gio2018:
Using traitCollectionDidChange addresses the problem seen on iPad
Traits

Issue #13102, opened 2019-12-13
Show multiple selected images


leandroalonso:
Using the Media Editor, the user might select more than one image and we want to show all the images in the Media Editor's Hub.

### Acceptance criteria

1. Given multiple images were given
2. Show all the images in the Hub

.

1. Given multiple images were given
2. Show all the images in the Hub
3. When the user taps one image, shows it

.

1. Given multiple images were given
2. Show all the images in the Hub
3. The user can swipe between the images

PR #13171, opened 2020-01-03
Show multiple images

leandroalonso:
Ref #13102

This PR adds the capability to show multiple images and to navigate between them on the Media Editor. Editing a selected image and tapping Done will be addressed in a future PR.

### To test

First, in `GutenbergViewController.swift` replace the function `gutenbergDidRequestFullscreenImage` by this one:

```swift
    func gutenbergDidRequestFullscreenImage(with mediaUrl: URL) {
        gutenbergDidRequestMediaEditor(with: mediaUrl, callback: { _ in })
    }
```

And inside `gutenbergDidRequestMediaEditor` give the `image` multiple times to the MediaEditor, like this (line 447):

```swift
let mediaEditor = WPMediaEditor([image, image, image, image, image, image, image, image, image, image])
```

You can test the same thing in Aztec, just pass the same image multiple times.

#### Swipe between the images

You can swipe between the images, the selected thumbnail (at the bottom) should be updated accordingly.

#### Navigate using the thumbnails

Tapping one thumbnail should display the image at the top. If the image isn't already loaded an activity indicator should appear.

#### Portrait and landscape, iPhone and iPad

If you rotate the device the interface should update accordingly.

PR submission checklist:

### DIFF
--- a/WordPress/MediaEditor/MediaEditor.swift
+++ b/WordPress/MediaEditor/MediaEditor.swift
@@ -45,6 +45,12 @@ public class MediaEditor: UINavigationController {
         setup()
     }
 
+    init(_ asyncImages: [AsyncImage]) {
+        self.asyncImages = asyncImages
+        super.init(rootViewController: hub)
+        setup()
+    }
+
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
     }
@@ -75,9 +81,15 @@ public class MediaEditor: UINavigationController {
 
         hub.apply(styles: styles)
 
+        isMultipleImages() ? hub.showThumbsToolbar() : hub.hideThumbsToolbar()
+
         setupForAsync()
 
         presentIfSingleImageAndCapability()
+    }
+
+    private func isMultipleImages() -> Bool {
+        return asyncImages.count > 1 || images.count > 1
     }
 
     private func setupForAsync() {
--- a/WordPress/MediaEditor/MediaEditor.swift
+++ b/WordPress/MediaEditor/MediaEditor.swift
@@ -24,7 +24,13 @@ public class MediaEditor: UINavigationController {
 
     var actions: [MediaEditorOperation] = []
 
+    var isSingleImageAndCapability: Bool {
+        return (asyncImages.count == 1 || images.count == 1) && Self.capabilities.count == 1
+    }
+
     private(set) var currentCapability: MediaEditorCapability?
+
+    private(set) var selectedImageIndex = 0
 
     public var styles: MediaEditorStyles = [:] {
         didSet {
@@ -96,23 +102,27 @@ public class MediaEditor: UINavigationController {
         presentIfSingleImageAndCapability()
     }
 
-    private func isMultipleImages() -> Bool {
-        return asyncImages.count > 1 || images.count > 1
-    }
-
     private func setupForAsync() {
-        if let thumb = asyncImages.first?.thumb {
-            hub.show(image: thumb)
-        } else {
-            asyncImages.first?.thumbnail(finishedRetrievingThumbnail: thumbnailAvailable)
+        asyncImages.enumerated().forEach { offset, asyncImage in
+            if let thumb = asyncImage.thumb {
+                thumbnailAvailable(thumb, offset: offset)
+            } else {
+                asyncImage.thumbnail(finishedRetrievingThumbnail: { [weak self] thumb in
+                    self?.thumbnailAvailable(thumb, offset: offset)
+                })
+            }
         }
 
-        showActivityIndicator()
-        asyncImages.first?.full(finishedRetrievingFullImage: fullImageAvailable)
+        if isSingleImageAndCapability {
+            showActivityIndicator()
+            asyncImages.first?.full(finishedRetrievingFullImage: { [weak self] image in
+                self?.fullImageAvailable(image, offset: 0)
+            })
+        }
     }
 
     func presentIfSingleImageAndCapability() {
-        guard let image = images.first, images.count == 1, Self.capabilities.count == 1, let capabilityEntity = Self.capabilities.first else {
+        guard isSingleImageAndCapability, let image = images.first, let capabilityEntity = Self.capabilities.first else {
             return
         }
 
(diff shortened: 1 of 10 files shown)


WordPress/MediaEditor/MediaEditorHub.swift
```diff
@@ -78,16 +117,128 @@ class MediaEditorHub: UIViewController {
...
+        selectedThumbIndex = indexPath.row
+        imagesCollectionView.scrollToItem(at: indexPath, at: .right, animated: true)
+    }
+
+    func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        guard scrollView == imagesCollectionView, isUserScrolling else {
+            return
+        }
+
+        let index = Int(round(scrollView.bounds.origin.x / imagesCollectionView.frame.width))
```
diegoreymendez: I'm not sure if I'm reading this right, but if I am, you can use this instead https://developer.apple.com/documentation/uikit/uicollectionview/1618020-indexpathsforvisibleitems

```swift
thumbsCollectionView.indexPathsForVisibleItems[0]
```

WordPress/MediaEditor/MediaEditorHub.swift
```diff
@@ -78,16 +117,128 @@ class MediaEditorHub: UIViewController {
...
+
+// MARK: - UICollectionViewDataSource
+
+extension MediaEditorHub: UICollectionViewDataSource {
+    func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
+        return numberOfThumbs
+    }
+
+    func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        // Thumbs Collection View
```
diegoreymendez: Suggestion: Whenever I see comments like this one and the one below (`// Images Collection View`), I tend to believe code can be separated even further, in order to simplify things and improve readability.

Something like:

```swift
func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {

    if collectionView == thumbsCollectionView {
        return cellForThumbsCollectionView(...)
    } else {
        return cellForImagesCollectionView(...)
    }
}
```

WordPress/MediaEditor/MediaEditor.swift
```diff
@@ -131,29 +164,29 @@ public class MediaEditor: UINavigationController {
         view.layer.add(transition, forKey: nil)
     }
 
-    private func thumbnailAvailable(_ thumb: UIImage?) {
-        guard let thumb = thumb, image == nil else {
```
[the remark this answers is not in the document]
leandroalonso: It is! This logic only worked for single images, and now we switched for the possibility of many images.

[This line](https://github.com/wordpress-mobile/WordPress-iOS/pull/13171/files#diff-1e86b82ea7f4582213468a01b6c12593R82) in `MediaEditorHub` makes sure that a thumb does not replace a full quality image.

Also, we have a unit test for this: https://github.com/wordpress-mobile/WordPress-iOS/pull/13171/files#diff-c0ddf5f907148e014521b0092d34df4bR184

WordPress/MediaEditor/MediaEditorHub.swift
leandroalonso: Here we're calculating which thumb should be selected based on the scroll in the `imagesCollectionView`.

You can quickly test this by slowing swiping to the next image. When you're seeing more than 50% of the next cell, the corresponding thumb is automatically selected.

But this code is a little bit confusing. I will refactor it to make it more meaningful.

leandroalonso: Oh, now that I'm reading again I guess your suggestion was to use:

```swift
thumbsCollectionView.indexPathsForVisibleItems[0]
```

However, I tried it and it doesn't work as I intended. To move to the correct thumbnail I'll need to check the direction of the scroll and check the `indexPathsForVisibleItems`.

WordPress/MediaEditor/MediaEditor.swift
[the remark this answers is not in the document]
diegoreymendez: Thanks!  That makes sense.

WordPress/MediaEditor/MediaEditorHub.swift
leandroalonso: Changed in a536e331ebf8d94e1c327451a6e5655405c42a3b.

Let me know what you think.

WordPress/MediaEditor/MediaEditorHub.swift
leandroalonso: Done in 59c5776e7bd784c6e543e575f2e261dabb388217

WordPress/MediaEditor/MediaEditorHub.swift
diegoreymendez: Looks fine to me.

WordPress/MediaEditor/MediaEditorHub.swift
```diff
@@ -78,16 +117,134 @@ class MediaEditorHub: UIViewController {
...
+        if collectionView == thumbsCollectionView {
+            return cellForThumbsCollectionView(cellForItemAt: indexPath)
+        }
+
+        return cellForImagesCollectionView(cellForItemAt: indexPath)
+    }
+
+    private func cellForThumbsCollectionView(cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let cell = thumbsCollectionView.dequeueReusableCell(withReuseIdentifier: Constants.thumbCellIdentifier, for: indexPath) as? MediaEditorThumbCell else {
+            return UICollectionViewCell()
```
diegoreymendez: This case happening would likely mean that the identifier provided above is not right, or it doesn't match the one we used for registering the cells.  Either way, this case triggering would signal something's off.

As such I'd suggest to log an error here, before returning.

If you use this method you can log an error with Sentry without that error killing the App or interrupting the user:

https://github.com/Automattic/Automattic-Tracks-iOS/blob/b8e747a37cf5731c77ba7832e3ef54f540740e60/Automattic-Tracks-iOS/CrashLogging.swift#L110

WordPress/MediaEditor/MediaEditorHub.swift
```diff
@@ -78,16 +117,134 @@ class MediaEditorHub: UIViewController {
...
+
+        cell.thumbImageView.image = availableThumbs[indexPath.row]
+        indexPath.row == selectedThumbIndex ? cell.showBorder(color: selectedColor) : cell.hideBorder()
+
+        return cell
+    }
+
+    private func cellForImagesCollectionView(cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let cell = imagesCollectionView.dequeueReusableCell(withReuseIdentifier: Constants.imageCellIdentifier, for: indexPath) as? MediaEditorImageCell else {
+            return UICollectionViewCell()
```
diegoreymendez: Same comment as above.

WordPress/MediaEditor/MediaEditorHub.swift
leandroalonso: If we return a `UICollectionViewCell()` directly (without using `dequeueReusableCell`) the app will crash anyway (and the crash will be reported to Sentry).

I refactored those methods to make it simpler and close to other examples we have in the app. :)

WordPress/MediaEditor/MediaEditorHub.swift
leandroalonso: Same reply as above 😄

leandroalonso:
@diegoreymendez since the comments you made also apply to the next PR I'm merging this one. If you have anything else feel free to point on the new one. :)

Hope that's not a problem!

PR #13178, opened 2020-01-07
[VoiceOver] Stats Overview Improvements

### LINKED ISSUE #12971: [VoiceOver] Stats issues reported by user

shiki:
Fixes #12971. Refs #11995.

This prioritizes fixing the user-reported issue #12971. But I also included some little things along the way. 

### Stacked Bar Chart Labels

The user reported that it is not easy to identify Visitors from Views. This adds labels to the bar chart items. 

The label only applies in stacked bar charts. This is the default behavior from the Charts library. See `BarChartRenderer.createAccessibleElement` for that. 

Before | After 
--------|-------
image        |       image

### Tab Buttons

VoiceOver reads these buttons without pause. Which is quite odd when you hear it. This PR improves that by making use of [`accessibilityLabel` and `accessibilityValue`](https://github.com/wordpress-mobile/WordPress-iOS/blob/4d966872a41140e9a3d483b6ebe369dfbe50f399/WordPress/Classes/ViewRelated/System/FilterTabBar.swift#L11-L13) for pausing. I also added hints. 

Before | After 
--------|-------
image        |       image

### Decimal Values

The Visitors values are oddly read with a “point zero” by VoiceOver. This was because the Charts library's default formatter uses a single decimal point. I added a [custom formatter](https://github.com/wordpress-mobile/WordPress-iOS/blob/4d966872a41140e9a3d483b6ebe369dfbe50f399/WordPress/Classes/ViewRelated/Stats/Charts/PeriodChart.swift#L56-L64) to fix this.

Before | After 
--------|-------
image        |       image

### Week Range

VoiceOver doesn't read the dash in the week range label very well. It just seems to ignore it and so the dates are read without pausing. I changed it so that the dash is now a `“to”`. Hopefully, this is more understandable when listened to.

Before | After 
--------|-------
image        |       image

## Outstanding Issues

There is a bug where the bar chart items are not selectable when you choose a different dimension (e.g. Views, Visitors, etc).

1. Open Stats → Days. One of the dimension (e.g. Views) is selected.
2. Slide finger over the bar chart. The hovered bar is read by VoiceOver.
3. Tap on another (e.g. Visitors)
4. Slide finger over the bar chart. VoiceOver reads “Bar chart depicting views…” instead of the bar item's label.

## Testing

1. Turn VoiceOver on.
2. Open Stats → Weeks tab.
3. The Views tab in the bar chart should be selected. If not, select it and select a different tab and go back to avoid the outstanding issue described above.
4. Confirm that a bar chart item's label is read before the value. For example, “Views, Jun 2019: 325”. 
5. Select the Visitors, Likes, or Comments tab but don't activate them yet. 
6. Confirm that there is a pause between the title and the value when VoiceOver reads them. 
7. Confirm that the hint is read after a slight pause. 
8. Navigate to the Visitors tab. 
9. Confirm that no decimal value is read when hovering over the bar items.
10. Select Stats → Weeks again. 
11. Confirm that hovering over a bar item will read “{date} **to** {date}". 

## Reviewing 

- What do you think of the tab button hints? Do they correctly describe what will happen when the buttons are tapped? 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Charts/Charts+AxisFormatters.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Charts/Charts+AxisFormatters.swift
@@ -61,7 +61,7 @@ class HorizontalAxisFormatter: IAxisValueFormatter {
             return ""
         }
 
-        return "\(formatter.string(from: weekStart)) – \(formatter.string(from: weekEnd))"
+        return "\(formatter.string(from: weekStart)) to \(formatter.string(from: weekEnd))"
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Stats/Charts/Charts+Support.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Charts/Charts+Support.swift
@@ -4,15 +4,16 @@ import Charts
 // MARK: - Charts extensions
 
 extension BarChartData {
-    convenience init(entries: [BarChartDataEntry]) {
-        let dataSet = BarChartDataSet(values: entries)
+    convenience init(entries: [BarChartDataEntry], valueFormatter: IValueFormatter? = nil) {
+        let dataSet = BarChartDataSet(values: entries, label: nil, valueFormatter: valueFormatter)
         self.init(dataSets: [dataSet])
     }
 }
 
 extension BarChartDataSet {
-    convenience init(values: [BarChartDataEntry]) {
-        self.init(values: values, label: nil)
+    convenience init(values: [BarChartDataEntry], label: String?, valueFormatter: IValueFormatter?) {
+        self.init(values: values, label: label)
+        self.valueFormatter = valueFormatter
     }
 }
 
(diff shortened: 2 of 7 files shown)


shiki:
Hi @ScoutHarris! Do you have time for this one? 🙂

ScoutHarris:
> What do you think of the tab button hints? Do they correctly describe what will happen when the buttons are tapped?

It all sounds good to me. They make sense and are clear. 👍

PR #13195, opened 2020-01-07
Stats: restore selected site ID when widget Stats modal is dismissed

### LINKED ISSUE #13194: Stats Widgets: incorrect stats displayed after modal dismissed
From https://wp.me/p5T066-18g/#comment-4107:

The wrong stats are displayed after tapping "Done" if you were previously on the Stats screen for a site other than the one the widget is configured to display. See video: <a href="https://cloudup.com/cFkRqW7iNaJ" rel="nofollow ugc">1m9s</a>.

Steps to reproduce:

<ol>
<li>Swipe right over the Home or Lock screen, scroll down, tap Edit, add a stats widget.</li>
<li>Go to Blog A in the app > Stats > Widgets and tap "Use this site."</li>
<li>Go to My Sites > Switch Site > select a different site > Stats.</li>
<li>Close the app, swipe right until you see the screen with the widgets.</li>
<li>Tap a widget.</li>
<li>Note that the stats are correct for the site the widget is configured to show.</li>
<li>Tap Done.</li>
<li>Note that the stats for the configured site are still showing.</li>
<li>Tap the back button at top left the re-open Stats.</li>
<li>Note that the correct stats for the site that is open in the app are now showing.</li>
</ol>

ScoutHarris:
Fixes #13194 

The problem was that the site ID used for stats was being overwritten by the site ID from the widget. To resolve this, the selected site ID is restored when the widget's Stats modal is dismissed. The Stats view is then refreshed for the correct site ID.

To test:

- Add the `Today` and/or the `All-Time` widget to the Today view.
- In the app:
  - Select a site. Go to Stats > Widgets > `Use this site`.
  - Select a different site, and go to Stats.
- Close the app.
- Return to the Today view and tap a widget.
- The app is launched, and a modal is displayed with the Stats for the widget site.
- Tap `Done` to close the modal.
- Verify the Stats view then displays the Stats for the selected site (not the widget site).

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 * Fixed bugs with the "Save as Draft" action extension's navigation bar colors and iPad sizing in iOS 13.
 * Fixes appearance issues with navigation bar colors when logged out of the app.
 * Fixed a bug that was causing the App to crash when the user tapped on certain notifications.
+* Stats fixed issue that could cause incorrect Stats to be displayed when viewing Stats from a widget.
 
 13.9
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,7 +6,7 @@
 * Fixed bugs with the "Save as Draft" action extension's navigation bar colors and iPad sizing in iOS 13.
 * Fixes appearance issues with navigation bar colors when logged out of the app.
 * Fixed a bug that was causing the App to crash when the user tapped on certain notifications.
-* Stats fixed issue that could cause incorrect Stats to be displayed when viewing Stats from a widget.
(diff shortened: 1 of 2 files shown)


bjtitus:
:shipit: 

Seems to behave as expected.

Out of scope: I wonder if we could remove the flash of the old stats by triggering a query refresh earlier. Also still trying to wrap my head around the singleton usage. I feel like the `siteID` should be set for each view controller instance, but I need to investigate a little further to understand the whole `Store` system.

ScoutHarris:
Thanks @bjtitus .

> I wonder if we could remove the flash of the old stats by triggering a query refresh earlier.

Yea, I wasn't real happy with it showing the widget stats then the selected site stats, but I wanted to get a fix out in the next release. I'll look at this a bit closer separately.

PR #13202, opened 2020-01-08
[VoiceOver] Various improvements for Stats

shiki:
Refs #11995.

This is a collection of unrelated improvements for Stats. I just made a pass and improved as much as I can. 🙂 

### [A] Header navigation order

23ad69c20a406196666234a8ea3bef1fc0c8ccb1 Changed the order so that when swiping right, the date and the timezone are accessed first. 

Before | After 
--------|-------
image        |       image

### [B] Header titles as `.header`

b23542d5c88080ed648a96cae8f616aa36dc9c2c Headers are now read with “Heading” by VoiceOver. 

Before | After 
--------|-------
image        |       image

### [C] Insights section header navigation order

4f61ee112b754ef8cc0fc6f0851905485c4bcf00 Fixed the swiping order of the section header so that the title is accessed first before the Manage Insights button. 

Before | After 
--------|-------
image        |       image

### [D] View More button accessibility labels

Fixed the labels of some of the View More buttons. They were previously read as just “Button”. 

- f0c179f63c7e659d9117188a9f24b007696ef30d
- d37c7d68e11dc72e721b38fd7f0374bc4df887f8
- b58c85f2c4c96b9db014a878ecf3005981c3cf3a

Before | After 
--------|-------
image        |       image

### [E] Insights → Posting Activity navigation order

The navigation order is a bit odd. I have a hard time describing it so I made videos. 😄 Fixed in 68ea094456cfffb41d706bce9f1da015f19caa71.

Before | After 
--------|-------
[video](https://drive.google.com/file/d/1__uyiBtAD4uye5YQ4Ys6Fuj1ebHS7iNd/view?usp=sharing)     |       [video](https://drive.google.com/file/d/1uZ6lQKsbJanPi8a0kz5my-hSx0JvFzs1/view?usp=sharing)

### [F] Period header week range label

964576271d2064b41a83c5628ab7e00b50078271 The dash in the period header week range isn't read by VoiceOver. I replaced it with “to”. 

Before | After 
--------|-------
        |       

### [G] World map label

20ef1c10814c118114462f30d9e7e075e5a87f99 The world map is read by VoiceOver using the minimum and maximum views. 

I pondered about making VoiceOver describe every little detail on the map. But I think describing what is “on that area of the screen” is good enough. Especially because we have the list of countries and views below the world map. 

I also thought about making the map inaccessible. I concluded that it's probably a bad idea because it covers a significant area of the screen. We might just make the user wonder more about what's on there.

Before | After 
--------|-------
image |       image

## Testing

1. Turn VoiceOver on.
2. Navigate to Stats → Weeks. 
3. Confirm that:
    - [ ] **[A]** Navigating by swiping right would make VoiceOver navigate to the Date and Timezone first before the Back and Forward buttons
    - [ ] **[B]** The Posts and Pages and other section titles are read with “Heading”. 
    - [ ] **[F]** The week's date range is read with a “to”. For example, “January 6th to January 12th”. 
    - [ ] **[G]** The country map is read as “World map showing views by country.”. What do you think about this label? Is it descriptive enough? 
3. Navigate to Stats → Insights.
4. Confirm that:
    - [ ] **[C]** Navigating by swiping right would make VoiceOver navigate and read a section's title first before the Manage Insight button.
    - [ ] **[D]** All the View More buttons are read as “View More. Button” by VoiceOver. 
    - [ ] **[E]** Navigating by swiping right in the Posting Activity would navigate in this order: 
        1. Posting Activity title
        2. Manage Insight button
        3. First Month
        4. Second Month
        4. Third Month
        4. View More button

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Shared Views/Date Chooser/SiteStatsTableHeaderView.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Shared Views/Date Chooser/SiteStatsTableHeaderView.swift
@@ -94,10 +94,7 @@ class SiteStatsTableHeaderView: UITableViewHeaderFooterView, NibLoadable, Access
     }
 
     func prepareForVoiceOver() {
-        if let period = dateLabel.text {
-            let localizedLabel = NSLocalizedString("Current period: %@", comment: "Period Accessibility label. Prefix the current selected period. Ex. Current period: 2019")
-            dateLabel.accessibilityLabel = .localizedStringWithFormat(localizedLabel, period)
-        }
+        dateLabel.accessibilityLabel = displayDateAccessibilityLabel()
 
         backButton.accessibilityLabel = NSLocalizedString("Previous period", comment: "Accessibility label")
         backButton.accessibilityHint = NSLocalizedString("Tap to select the previous period", comment: "Accessibility hint")
@@ -158,6 +155,23 @@ private extension SiteStatsTableHeaderView {
             return "\(from)"
         }
     }
+
+    func displayDateAccessibilityLabel() -> String? {
+        guard let components = displayDateComponents() else {
+            return nil
+        }
+
+        let (from, to) = components
+
+        if let to = to {
+            let format = NSLocalizedString("Current period: %@ to %@", comment: "Week Period Accessibility label. Prefix the current selected period. Ex. Current period: Jan 6 to Jan 12.")
+            return .localizedStringWithFormat(format, from, to)
+        } else {
+            let format = NSLocalizedString("Current period: %@", comment: "Period Accessibility label. Prefix the current selected period. Ex. Current period: 2019")
+            return .localizedStringWithFormat(format, from)
+        }
+    }
+
     /// Returns the formatted dates for the current period.
     func displayDateComponents() -> (from: String, to: String?)? {
         guard let date = date, let period = period else {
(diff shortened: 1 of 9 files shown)


WordPress/Classes/ViewRelated/Stats/Shared Views/Date Chooser/SiteStatsTableHeaderView.swift
```diff
@@ -139,6 +143,37 @@ private extension SiteStatsTableHeaderView {
     }
 
     func displayDate() -> String? {
+        guard let components = displayDateComponents() else {
+            return nil
+        }
+
+        let (from, to) = components
```
ScoutHarris: This is a bit nitpicky, but can we give these `from` and `to` vars more descriptive names? Like maybe `fromDate` and `toDate` or something like that?

ScoutHarris:
Hey @shiki . Thanks for the detail explanations!

> With all that said, the button should say that it's there but nothing will happen when they tap on it. In your example, the button should say “dimmed”. But it doesn't. 😄 I'll create a separate PR for that to not overload this one.

:thumbsup:

> I believe that the “View More” label is already obvious enough to inform the user about what would happen if they activate the button. The “Tap for more detail” also sounds similar to “View more” IMO and is probably just repetitive. The “Tap” word is also not necessary and not recommended by Apple.

Sounds reasonable. I'd vote to remove the hints.

shiki:
Merging this. The additional changes discussed here will be covered in https://github.com/wordpress-mobile/WordPress-iOS/pull/13210. 

Thank you, @ScoutHarris!

PR #13203, opened 2020-01-08
Implements tests for getting the user agent in WKWebView.

### LINKED ISSUE #13184: Add tests in WPUserAgentTests for WKWebView.
`WPUserAgentTests` instantiates a `UIWebView` to retrieve the current user agent.  Implement the same test for `WKWebView` without replacing the existing code.  This will be useful for a smooth transition.

diegoreymendez:
Fixes #13184 

Implements side by side testing of the User Agent

## To test:

Run the unit tests and make sure they work properly.

Check the code and make sure the changes make sense.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/WKWebView+SynchJSEvaluation.swift
+++ b/WordPress/Classes/Extensions/WKWebView+SynchJSEvaluation.swift
@@ -1,0 +1,64 @@
+import AutomatticTracks
+import WebKit
+
+/// This extension contains a hack used to make it possible to
+/// evaluate JavaScript code synchronously.
+///
+extension WKWebView {
+
+    @objc
+    func stringByEvaluatingJavaScript(fromString javascriptString: String) -> String {
+        var result = ""
+        var finished = false
+
+        evaluateJavaScript(javascriptString) { (output, error) in
+            defer {
+                finished = true
+            }
+
+            guard error == nil else {
+                CrashLogging.logMessage(
+                    "Failed to evaluate JavaScript",
+                    properties: ["error": error!, "evaluatedString": javascriptString],
+                    level: .error)
+                return
+            }
+
+            guard let output = output as? String else {
+                return
+            }
+
+            result = output
+        }
+
+        while (!finished) {
+            RunLoop.current.run(mode: .default, before: NSDate.distantFuture)
+        }
+
+        return result
+    }
+    /*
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Extensions/WKWebView+SynchJSEvaluation.swift
```diff
@@ -0,0 +1,47 @@
+import AutomatticTracks
+import WebKit
+
+/// This extension contains a hack used to make it possible to
+/// evaluate JavaScript code synchronously.
+///
+extension WKWebView {
+
+    /// Synchornous evaluation of JavaScript in WKWebView.
```
leandroalonso: Small typo here

```suggestion
    /// Synchronous evaluation of JavaScript in WKWebView.
```

WordPress/Classes/Extensions/WKWebView+SynchJSEvaluation.swift
```diff
@@ -0,0 +1,47 @@
...
+
+            guard let output = output as? String else {
+                return
+            }
+
+            result = output
+        }
+
+        while !finished {
+            RunLoop.current.run(mode: .default, before: NSDate.distantFuture)
```
leandroalonso:
```suggestion
            RunLoop.current.run(mode: .default, before: .distantFuture)
```

WordPress/Classes/Extensions/WKWebView+SynchJSEvaluation.swift
```diff
@@ -0,0 +1,47 @@
...
+        var result = ""
+        var finished = false
+
+        evaluateJavaScript(javascriptString) { (output, error) in
+            defer {
+                finished = true
+            }
+
+            guard error == nil else {
+                CrashLogging.logMessage(
```
leandroalonso: Just out of curiosity: we're logging here as a crash. But this will not cause a crash, will it?

diegoreymendez: We're not exactly logging here as a "crash" but as an error.  I believe Sentry should show this as a non crashing error.

That said the reason why logging this is important is because we don't want this to fail silently if there are issues evaluating the Javascript.  This could happen eventually due to typos in the Javascript string, or due to things like changing APIs.

Issue #13197, opened 2020-01-08
Post Preview Enhancements Master Issue


bjtitus:
Master Thread: https://wp.me/paCL20-dF-p2
Design: https://wp.me/paCL20-aw

![device_previews_ios-1](https://user-images.githubusercontent.com/3250/71990909-794a1900-31f1-11ea-9a1d-cc315574888d.png)

- [x] Add feature flag. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13206)
- [x] Add new post preview UI.
  - [x] Replace UIWebView with WKWebView (existing: https://github.com/wordpress-mobile/WordPress-iOS/issues/13180) (https://github.com/wordpress-mobile/WordPress-iOS/pull/13206)
  - [x] Add new toolbar with controls. (#13245)
  - [x] Add Desktop / Mobile preview options (#13245)
- [x] Verify redirect behavior (`wpios_redirect` after logging in) matches previous implementation.
- [x] Ensure views are correct for Dark Mode.
- [x] Add Accessibility support. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13292)
- [x] Add Tracks. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13281)
- [x] Enable feature flag for all. (#13334)
- [x] Remove feature flag. (#13587)

bjtitus:
Here's a summary explaining some of the [logging issues](https://github.com/wordpress-mobile/WordPress-iOS/pull/13226#pullrequestreview-342757084) @ScoutHarris noticed with some of these changes that have since been mitigated or are otherwise accounted for:

> 1. This appears in the console:

```
WordPress.PreviewWebKitViewController Error Loading [Error Domain=WebKitErrorDomain Code=102 "Frame load interrupted" UserInfo={_WKRecoveryAttempterErrorKey=<WKReloadFrameErrorRecoveryAttempter: 0x600001e39e40>, NSErrorFailingURLStringKey=https://wordpress.com/?wpios_redirect=https://thescoutharris.wordpress.com/2019/12/17/comments-and-collatz-conundrum-2/, NSErrorFailingURLKey=https://wordpress.com/?wpios_redirect=https://thescoutharris.wordpress.com/2019/12/17/comments-and-collatz-conundrum-2/, NSLocalizedDescription=Frame load interrupted}]
```

[...]

PR #13206, opened 2020-01-08
[Post Preview] Adds NewPostPreviewViewController for redesign work

### LINKED ISSUE #13180: Migrate PostPreviewViewController to WKWebView
Tasks:

- [ ] Replace UIWebView and any related code with WKWebView.
- [ ] Replace UIWebViewDelegate implementation.

bjtitus:
Fixes #13180 (once `postPreview` feature flag is disabled)

Part of #13197

This is intended to be a rough copy of the existing class but will change as additional functionality is added (such as Desktop/Mobile previews)

To test:

- Preview a post from Edit, Post List, or other screen
- Check that new view controller is used

PR submission checklist:

- [x] ~I have considered adding unit tests where possible.~ I think the `NewPostPreview` will undergo some changes and tests should be written once it is complete.

- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ This can be done once the feature flag is released.


WordPress/Classes/ViewRelated/Post/NewPostPreviewViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+
+    private func setupWebView() {
+        webView = WKWebView(frame: view.bounds)
+        webView.navigationDelegate = self
+        view.addSubview(webView)
+        view.pinSubviewToAllEdges(webView)
+    }
+
+    @objc private func dismissPreview() {
+        onClose?()
```
frosty: This is a slightly different behaviour to the existing view controller (not sure if that matters?). The previous one only called dismiss on the presenting view controller if there was no `onClose` block provided.

WordPress/Classes/ViewRelated/Post/NewPostPreviewViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+        view.pinSubviewToAllEdges(controller.view)
+        noResultsViewController?.didMove(toParent: self)
+    }
+}
+
+extension NewPostPreviewViewController: PostPreviewGeneratorDelegate {
+    func preview(_ generator: PostPreviewGenerator, attemptRequest request: URLRequest) {
+        startLoadAnimation()
+        var newRequest = request
+        newRequest.url = URL(string: "http://sdfsldfjlwefj.sdlfj")
```
frosty: Is this intentional? The existing implementation just passes the request to the webView.

WordPress/Classes/ViewRelated/Post/NewPostPreviewViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        if navigationAction.request.url?.absoluteString == post.permaLink {
+            // Always allow loading the preview
+            decisionHandler(.allow)
+            return
+        }
+
+        if navigationAction.navigationType == .linkActivated {
+            UIApplication.shared.open(navigationAction.request.url!, options: [:], completionHandler: nil)
```
frosty: Should we unwrap the `url` here to be safe?

WordPress/Classes/ViewRelated/Post/NewPostPreviewViewController.swift
```diff
@@ -0,0 +1,252 @@
...
+            decisionHandler(.cancel)
+            return
+        }
+
```
frosty: We're missing the `formSubmitted` case from the original implementation.

WordPress/Classes/ViewRelated/Post/NewPostPreviewViewController.swift
bjtitus: Oops, accidentally pushed this when testing NoResults. 🤦‍♂

bjtitus: Oops, accidentally pushed this after testing NoResults 🤦‍♂

bjtitus:
@frosty Thanks for taking a look! I did a little extra clean up and fixed those issues you had mentioned.

Should be ready for review now.

bjtitus:
@frosty I went ahead and changed all of the presentations to modals and now always show the done button.

The next set of changes will use WebKitViewController anyway so that class will end handling all of the `navigationItem` set up.

WordPress/Classes/ViewRelated/Post/NewPostPreviewViewController.swift
```diff
@@ -47,6 +44,7 @@ class NewPostPreviewViewController: UIViewController {
     }()
 
     private lazy var doneBarButtonItem: UIBarButtonItem = {
+//        let buttonItem = UIBarButtonItem(barButtonSystemItem: UIBarButtonItem.SystemItem.done, target: self, action: #selector(dismissPreview))
```
frosty: Can we just remove this line?

PR #13213, opened 2020-01-09
Migrate WPUserAgent to WKWebView

### LINKED ISSUE #13186: Migrate WPUserAgent to WKWebView.
Migrate `WPUserAgent` to use `WKWebView`.

diegoreymendez:
Fixes #13186 

Migrates `WPUserAgent` to `WKWebView`.  Also adds a useful `WKWebView` extension to abstract `WPUserAgent` from the details of how to retrieve the user agent.

## Dependencies:

Make sure #13184 is merged into `develop` and that this branch is updated from `develop` before assigning this PR for review.

## To test:

Running the unit tests should be enough since they ensure that the user agent hasn't changed from what it was before.

## PR submission checklist:


WordPress/Classes/Extensions/WKWebView+UserAgent.swift
```diff
@@ -0,0 +1,22 @@
+import AutomatticTracks
```
leandroalonso: I guess we can remove this import.

Issue #12702, opened 2019-10-15
Stats Widgets Master Issue


ScoutHarris:
Add new Stats Widgets ([#](https://wp.me/paCL20-ee-p2)).

---
#### 14.2:
- [x] Stats: when viewed in a modal, there is a grey bar at the bottom of the view (#13301)
- [x] Update Tracks (#13272)
- [x] Stats Widgets: Update only if data has changed. (#13331)
- [x] Stats Widgets: show failed loading view (#13333)

#### 14.1:
- [x] Stats Widgets: reload table when stats are updated. (#13232)
- [x] Add new _Views This Week_ stat widget. (#13044)
- [x] Stats Widgets: fix site url UI issues (#13243)
- [x] Stats Widgets: localize display names (#13247)
- [x] Stats Widgets: clear stored data when site is changed (#13254)
- [x] Stats Widgets: show no connection view. (#13271)

#### 14.0:
- [x] Stats Widgets: incorrect stats displayed after modal dismissed (#13194)
- [x] Stats Widgets: Abbreviate large values (#13196)
- [x] Stats Widgets: update separator line color (#13207)
- [x] Stats: add accessibility label for abbreviated stats (#13214)

#### 13.9:
- [x] Change nav bar button from `Today` to `Widgets`. (included in #13036)
- [x] Add new _All Time_ stat widget. (#13030)

#### 13.7:
- [x] [Stats] Remove old WPKit stats code (#12848)
- [x] [Stats Refresh] Remove feature flag & legacy code (#10975)
- [x] Update Today widget to new design (#12849)

#### 13.6:
- [x] Update current Today widget to use new Stats. (#12756)
- [x] [Stats Refresh] Remove feature flag (#12771)
- [x] [Stats] remove old code from WPiOS (#12794)

---
![all_widgets_states_ios-2](https://user-images.githubusercontent.com/1816888/66867636-ef274700-ef58-11e9-9bb0-d2c9bd5736be.png)

ScoutHarris:
Hey @osullivanchris .

Can we revisit the designs for these? Specifically, the grey background.

In iOS13, the widgets are even more transparent. Which makes the new grey background _very obvious_. 

<img width="400" alt="new_ios13_light" src="https://user-images.githubusercontent.com/1816888/68259735-e83fa180-fff7-11e9-8baa-1dadbfb6a274.png">

[...]

PR #13214, opened 2020-01-10
Stats: add accessibility label for abbreviated stats

ScoutHarris:
Ref: #12702, #11995

This change adds an accessibility label to abbreviated stats so that the unit name is read instead of the abbreviated letter. For example, “54.8M” is now read as “54 point 8 million” instead of "54 point 8 m".

The accessibility label is not yet utilized _everywhere_ an abbreviated value is displayed, but does hit most. Specifically:
- `StatsTotalRow` - used by all non-specialized stat cards.
- `StatsTwoColumnRow` - used by Insights two-column stat cards. (`Today`, `All-Time`, `Follower Totals`, `This Year`)
- `WidgetTwoColumnCell` - used by the `Today` and `All-Time` widgets.

Also, as an added bonus, the abbreviated strings are now translated.

To test:

Prereqs:
- Run the app & enable VoiceOver.
- Select a high volume site and go to Stats. Hogwarts or en.blog are usually good for this.

---
Stats Total Row:
- Select a stat row on any non-specialized card with an abbreviated value. 
- Verify the value is read correctly.

stat_row_abbrev

- Select a stat row on any non-specialized with an unabbreviated value.
- Verify the value is read correctly.

stat_row_unabbrev

---
Stats Two Column:
- Go to Insights.
- Select a two-column stat card.
- Verify both abbreviated and unabbreviated values are read correctly.

stats_two_column

---
Widgets Two Column:
- On the high volume site Stats, select `Widgets` on the nav bar, then `Use this site`.
- Go to your device's Today view and add the `Today` and/or `All-Time` widget(s).
- Select a data row. (Note widgets are tables, so the rows are read separately.)
- Verify the values are read correctly.

widget_two_column

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
@@ -1,8 +1,8 @@
 import Foundation
 
 fileprivate struct Unit {
-    let abbreviation: String
-    let name: String
+    let abbreviationFormat: String
+    let accessibilityLabelFormat: String
 }
 
 extension Double {
@@ -20,23 +20,36 @@ extension Double {
     }
 
     private var decimalFormatter: NumberFormatter {
-         get {
-             let formatter = Double._decimalFormatter
-             // Show at least one digit after the decimal
-             formatter.minimumFractionDigits = 1
-             return formatter
-         }
-     }
+        get {
+            let formatter = Double._decimalFormatter
+            // Show at least one digit after the decimal
+            formatter.minimumFractionDigits = 1
+            return formatter
+        }
+    }
 
     private static var _units: [Unit] {
         get {
             var unitsArray: [Unit] = []
-            unitsArray.append(Unit(abbreviation: "K", name: "thousand"))
-            unitsArray.append(Unit(abbreviation: "M", name: "million"))
-            unitsArray.append(Unit(abbreviation: "B", name: "billion"))
-            unitsArray.append(Unit(abbreviation: "T", name: "trillion"))
-            unitsArray.append(Unit(abbreviation: "P", name: "quadrillion"))
-            unitsArray.append(Unit(abbreviation: "E", name: "quintillion"))
+
+            unitsArray.append(Unit(abbreviationFormat: NSLocalizedString("%@K", comment: "Label displaying value in thousands. Ex: 66.6K."),
+                                   accessibilityLabelFormat: NSLocalizedString("%@ thousand", comment: "Accessibility label for value in thousands. Ex: 66.6 thousand.")))
+
+            unitsArray.append(Unit(abbreviationFormat: NSLocalizedString("%@M", comment: "Label displaying value in millions. Ex: 66.6M."),
+                                   accessibilityLabelFormat: NSLocalizedString("%@ million", comment: "Accessibility label for value in millions. Ex: 66.6 million.")))
+
+            unitsArray.append(Unit(abbreviationFormat: NSLocalizedString("%@B", comment: "Label displaying value in billions. Ex: 66.6B."),
+                                   accessibilityLabelFormat: NSLocalizedString("%@ billion", comment: "Accessibility label for value in billions. Ex: 66.6 billion.")))
+
+            unitsArray.append(Unit(abbreviationFormat: NSLocalizedString("%@T", comment: "Label displaying value in trillions. Ex: 66.6T."),
+                                   accessibilityLabelFormat: NSLocalizedString("%@ trillion", comment: "Accessibility label for value in trillions. Ex: 66.6 trillion.")))
+
+            unitsArray.append(Unit(abbreviationFormat: NSLocalizedString("%@P", comment: "Label displaying value in quadrillions. Ex: 66.6P."),
+                                   accessibilityLabelFormat: NSLocalizedString("%@ quadrillion", comment: "Accessibility label for value in quadrillion. Ex: 66.6 quadrillion.")))
+
+            unitsArray.append(Unit(abbreviationFormat: NSLocalizedString("%@E", comment: "Label displaying value in quintillions. Ex: 66.6E."),
+                                   accessibilityLabelFormat: NSLocalizedString("%@ quintillion", comment: "Accessibility label for value in quintillions. Ex: 66.6 quintillion.")))
+
             return unitsArray
         }
     }
@@ -90,9 +103,8 @@ extension Double {
         roundedNum = self < 0 ? -roundedNum : roundedNum
         let formattedValue = roundedNum.formatWithFractions()
 
-        let format = NSLocalizedString("%@%@", comment: "Label displaying abbreviated value. The first parameter is the value, the second is the unit. Ex: 55.5M, 66.6K.")
-        let formattedString = String.localizedStringWithFormat(format, formattedValue, unit.abbreviation)
-        formattedString.accessibilityLabel = String.localizedStringWithFormat(format, formattedValue, unit.name)
+        let formattedString = String.localizedStringWithFormat(unit.abbreviationFormat, formattedValue)
+        formattedString.accessibilityLabel = String.localizedStringWithFormat(unit.accessibilityLabelFormat, formattedValue)
 
         return formattedString
     }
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
```diff
@@ -13,6 +19,34 @@ extension Double {
         }
     }
 
+    private var decimalFormatter: NumberFormatter {
+         get {
+             let formatter = Double._decimalFormatter
+             // Show at least one digit after the decimal
+             formatter.minimumFractionDigits = 1
+             return formatter
```
shiki: Nitpick: Just noting that these lines are indented differently compared to the contents of `numberFormatter` above. 😄

shiki: If I understand correctly, the purpose of the `_decimalFormatter` is to cache it since it is a heavy object. Another way to do this so that there is no additional `static var` that is potentially **accessible by other routines** is to make it local using a `struct`:

```swift 
    private var decimalFormatter: NumberFormatter {
         get {
            struct Cache {
                static let formatter: NumberFormatter = {
                    let formatter = NumberFormatter()
                    formatter.minimumFractionDigits = 1
                    return formatter
                }()
            }

            return Cache.formatter
         }
     }

```

Not important for this PR, just thought of mentioning my ideas. 🙂

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
```diff
@@ -13,6 +19,34 @@ extension Double {
...
+
+    private static var _units: [Unit] {
+        get {
+            var unitsArray: [Unit] = []
+            unitsArray.append(Unit(abbreviation: "K", name: "thousand"))
+            unitsArray.append(Unit(abbreviation: "M", name: "million"))
+            unitsArray.append(Unit(abbreviation: "B", name: "billion"))
+            unitsArray.append(Unit(abbreviation: "T", name: "trillion"))
+            unitsArray.append(Unit(abbreviation: "P", name: "quadrillion"))
+            unitsArray.append(Unit(abbreviation: "E", name: "quintillion"))
```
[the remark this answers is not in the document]
bjtitus: Don't want to throw too much of a wrench in things, but aren't the large number groupings also different in many Locales? ([Indian](https://en.wikipedia.org/wiki/Indian_numbering_system), [Chinese](https://resources.allsetlearning.com/chinese/grammar/Big_numbers_in_Chinese)).

`NSNumberFormatter` has some handlers for some of this ([`secondaryGroupingSize`](https://developer.apple.com/documentation/foundation/numberformatter/1413348-secondarygroupingsize#)) but it's not really built in to the format strings at all.

I tried to play with augmenting `NumberFormatter` to do more of the work, which I think slightly simplifies the code but doesn't really solve this problem you're mentioning (note that the rounding change needs a bit of finagling for the higher numbers). Maybe setting the `positiveSuffix` / `positivePrefix` appropriately might be cleaner than swapping format string ordering? 🤷‍♂ 

```swift
extension Double {
        
    var abbreviated: String? {
        let numberFormatter = NumberFormatter()
        numberFormatter.locale = Locale(identifier: "ar_SA")
                
        // Convert to log1000
        guard self != 0 else { return "0" }
        let exp = Double(floor(log10(self) / log10(1000.0)))
        let mult = Double(pow(0.1, (exp * exp) + 2))
        
        let units: [String] = ["K", "M", "B", "T", "P", "E"]
        
        numberFormatter.multiplier = NSNumber(value: mult)//NSNumber(value: rounding/1000)
        numberFormatter.positiveFormat = "#,##0.#"
        numberFormatter.positiveSuffix = units[units.index(0, offsetBy: Int(exp) - 1)]
        
        return numberFormatter.string(for: self)
    }
}
```

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
```diff
@@ -1,8 +1,14 @@
...
 
+fileprivate struct Unit {
+    let abbreviation: String
+    let name: String
+}
+
 extension Double {
 
     private static var _numberFormatter: NumberFormatter = NumberFormatter()
+    private static var _decimalFormatter: NumberFormatter = NumberFormatter()
```
shiki: Is there a reason why we use `var` instead of `let`? Do we change these somewhere? I think I missed it. 🤔

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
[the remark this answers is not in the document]
shiki: @bjtitus Thanks for jumping in! 

> but doesn't really solve this problem you're mentioning

Not sure I understand. Would you mind pointing me to which one isn't solved by your code? 😄 Sorry, I realized I wrote a very long message in https://github.com/wordpress-mobile/WordPress-iOS/pull/13214#r365284167. 

> but aren't the large number groupings also different in many Locales? (Indian, Chinese)



OMG. This got much more complicated. 😱 I really wish `NSNumberFormatter` handles all of these for us. 

I think we can consider this out of scope for now and stick with the `["K", "M", "B", "T", "P", "E"]`groupings. What do you think, @bjtitus? 

> Maybe setting the positiveSuffix / positivePrefix appropriately might be cleaner than swapping format string ordering?

This looks like a decent solution. However,  it doesn't seem to solve the possible spacing between the values. I think this is the _unsolved problem_ that you mentioned.

Also, I think this would require us to check if the device's language is RTL. Which can be another point of failure. I'd still probably prefer to leave the decision to the language translators.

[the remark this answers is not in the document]
bjtitus: @shiki 

> Would you mind pointing me to which one isn't solved by your code?

Sure! I was referring to the fact that my code didn’t include the RTL switching you had mentioned:

> It turns out that the abbreviation should be on the left of the number for Arabic.

The prefix/suffix solution I mentioned was just a possible “easier” solution.

> it doesn't seem to solve the possible spacing between the values.

Yeah, I played around with this and thought the locale formatting might properly space things out but I can’t figure that one out.

> I think we can consider this out of scope for now and stick with the ["K", "M", "B", "T", "P", "E"]groupings.

I completely agree. I think we should probably be cautious going too far down this rabbit hole.

> I really wish NSNumberFormatter handles all of these for us.

🧚🏼‍♀️ They do this with Date and Measurement formatters. Seems like there could be a short/long formatting option added for simple counting units. I’ll file a feedback request.

ScoutHarris:
@shiki @bjtitus thank you for all the feedback!

I'll look into using `NumberFormatter` to calculate the values separately since it's a significant change.

I did implement @shiki 's suggestions for the string format and using `Cache` instead of static vars. 

@shiki ready for another review please sir. Thank you!

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
```diff
@@ -32,31 +82,44 @@ extension Double {
...
-            return "\(sign)\(roundedNum)\(units[exp-1])"
+            roundedNum = unsignedRoundedNum
+            unit = units[exp-1]
         }
+
+        roundedNum = self < 0 ? -roundedNum : roundedNum
+        let formattedValue = roundedNum.formatWithFractions()
+
+        let formattedString = String.localizedStringWithFormat(unit.abbreviationFormat, formattedValue)
+        formattedString.accessibilityLabel = String.localizedStringWithFormat(unit.accessibilityLabelFormat, formattedValue)
```
shiki: This is really cool. I didn't know that `String` has an `accessibilityLabel`. It's like Inception. 😄

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
```diff
@@ -32,31 +82,44 @@ extension Double {
...
 
-        if roundedNum == 1000.0 {
-            return "\(sign)\(1)\(units[exp])"
+        if unsignedRoundedNum == 1000.0 {
+            roundedNum = 1
+            unit = units[exp]
         } else {
-            return "\(sign)\(roundedNum)\(units[exp-1])"
+            roundedNum = unsignedRoundedNum
+            unit = units[exp-1]
```
shiki: It's probably impossible that `exp-1` will be out of bounds. But do you think we should still handle the possibility that it will in order to avoid a crash?

ScoutHarris: Ah yea. Nice catch. Tis done!

ScoutHarris:
Thanks @shiki !

PR #13227, opened 2020-01-13
Add Media Editor to Gutenberg and Aztec flows

leandroalonso:
Fixes #13218

### Gutenberg

As for now, Gutenberg doesn't support multiple images. However, the flow is already implemented and you can test. In order to do so, in the method `gutenbergDidRequestMediaFromDevicePicker` change `allowMultipleSelection` to `true`.

1. Create a new post using Gutenberg
2. Insert a block image
3. Tap "Add image" > "Choose from device"
4. Select images (only the first selected will be uploaded)
5. Tap "Preview X"
6. Edit the first image
7. Tap Done
8. Check that the image was uploaded correctly

### Aztec

1. Create a new post using Aztec
2. Tap "+"
3. Tap 
4. Select as many images as you want
5. Tap "Preview X"
6. Edit the images and tap Done
7. All the images should be correctly uploaded

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #13098, opened 2019-12-13
Show message when image could not be retrieved


leandroalonso:
When we're loading an image we show this screen:

![Simulator Screen Shot - iPhone 11 - 2019-12-12 at 19 05 34](https://user-images.githubusercontent.com/7040243/70826578-0e5e1b80-1dc6-11ea-9788-c64a77fde215.png)

However, this process might fail and we need to give feedback to the user.

leandroalonso:
@mbshakti Can you take a look at this one?

PR #13230, opened 2020-01-14
Show message when image fails to load

leandroalonso:
Ref #13098

Review first: https://github.com/wordpress-mobile/WordPress-iOS/pull/13227

#### Single image

1. In `PHAsset+AsyncImage.swift` inside the method `full` change `finishedRetrievingFullImage(image)` to `finishedRetrievingFullImage(nil)` to simulate an image failing to load
2. Create a new post using Aztec
3. Tap "+" and then 
4. Select a single image, tap "Preview 1"
5. Check that a message saying that the media couldn't be loaded appears right away

Put a breakpoint inside the `full()` method in `PHAsset+AsyncImage.swift` and make sure it's hit when tapping retry icon.

#### Multiple image

1. In `PHAsset+AsyncImage.swift` inside the method `full` change `finishedRetrievingFullImage(image)` to `finishedRetrievingFullImage(nil)` to simulate an image failing to load
2. Create a new post using Aztec
3. Tap "+" and then 
4. Select multiple images, tap "Preview X"
5. Tap the crop button
6. Check that a message saying that the media couldn't be loaded appeared

Put a breakpoint inside the `full()` method in `PHAsset+AsyncImage.swift` and make sure it's hit when tapping retry icon.

PR submission checklist:


mbshakti:
Overall, looks great. Two comments:

(1) Since the user can't crop when the image fails to load, can we disable the crop button for this view? You mentioned that technically the crop button still works (user can still tap the crop button again to try to load the image), but I think we can just disable it since the retry button works; the user can just tap the retry button if they want to try reload the image. 

(2) Can we change the copy to: "Failed to insert media. Please tap to retry."

leandroalonso:
@mbshakti 

1) Right now we only have the crop button, but there might be more buttons in the near future. That means we would block the user from trying to tap into something else (other than crop, for example). What do you think?

2) Do you mean changing the text to "Failed to insert media. Please tap to retry."? If that's the case, the Media Editor is trying to retrieve the image so the user can manipulate it somehow — instead of trying to insert it. What about "We couldn't retrieve this media. Please tap to retry."? I'm using this tone because it was the one we used through the [Offline Support project strings](https://github.com/wordpress-mobile/WordPress-Android/pull/10542).

mbshakti:
> Right now we only have the crop button, but there might be more buttons in the near future. That means we would block the user from trying to tap into something else (other than crop, for example

Would those buttons also not work like the crop button doesn't work here? We should disable any button that doesn't work. 

Re the copy, that sounds good!

leandroalonso:
> Would those buttons also not work like the crop button doesn't work here? We should disable any button that doesn't work.

All of them would still work, they will retry fetching the image again. That's why I think we might let them enabled, the user will have the option to tap to retry again or tap in any of the available capabilities too. Otherwise, let's say that the user tapped crop, it failed, and then he changed his mind and tapped "filters". By disabling the buttons that wouldn't be possible.

JFYI, the buttons are disabled while the image is _being_ fetched.

> Re the copy, that sounds good!

Updated!

leandroalonso:
@mbshakti I'll merge this PR but we still have to figure out if we're going to keep the buttons enabled or not.

Since there's another PR on the way, we can do this change there. Hope that's reasonable for you.

leandroalonso:
@mbshakti Yes!

PR #13243, opened 2020-01-17
Stats Widgets: fix site url UI issues

ScoutHarris:
Fixes #n/a
Ref #12702 

The site url is now displayed with a `UITableViewCell` instead of `UITableViewHeaderFooterView`, and is added to the table as simply another row instead of a footer.

To test:
- In the app, select a site. Go to Stats > Widgets > `Use this site`.
- Add all the Stats widgets to the Today view.

**For `All-Time` and `Today`:**
- Expand the widgets.
- Scroll away from the Today view, wait a few seconds, then scroll back.
- Verify the site url doesn't bounce.

Before | After
--------|-------
reload_after        |      two_col_reload_after

**For `This Week`:**
- Expand the widget.
- Verify the site url appears at the bottom of the table (instead of appearing at the top and jumping down).

Before | After
--------|-------
this_week_before        |       this_week_after

PR submission checklist:


ScoutHarris:
Thanks @bjtitus .

> I was hoping there would be a nicer way to remove the height calculation code (expandedHeight). I was able to achieve very similar results by setting preferredContentSize = tableView.sizeThatFits(maxSize) but the animations weren't quite as smooth. The demo project also seems to recommended doing the manual height calculation.

Indeed. I'm not a fan of manually calculating the height (it did cause some pain points, like what this PR fixes), and tried other solutions. But, yea, all the examples and docs do it this way.

PR #13254, opened 2020-01-21
Stats Widgets: clear stored data when site is changed

ScoutHarris:
Ref #12702 

When changing the site used for widgets, the stored data for the old site is now purged.

To test:
The easiest way to test this is to prevent the widgets from fetching data for the new site.

- Add the 3 WordPress widgets to the Today view.
- Return to the app and change the site used for widgets.
- Disable your internet connection.
- Return to the Today view.
- Verify all the widgets show dashes (because no data was loaded from the plists and new data cannot be fetched).

**Note:** I'm currently working on a "no connection" view, so showing dashes in this state is temporary. This is just the easiest way to verify the data is purged.

no_data

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/AllTimeWidgetStats.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/AllTimeWidgetStats.swift
@@ -21,11 +21,11 @@ struct AllTimeWidgetStats: Codable {
 
 extension AllTimeWidgetStats {
 
-    static func loadSavedData() -> AllTimeWidgetStats {
+    static func loadSavedData() -> AllTimeWidgetStats? {
         guard let sharedDataFileURL = dataFileURL,
             FileManager.default.fileExists(atPath: sharedDataFileURL.path) == true else {
                 DDLogError("AllTimeWidgetStats: data file '\(dataFileName)' does not exist.")
-                return AllTimeWidgetStats()
+                return nil
         }
 
         let decoder = PropertyListDecoder()
@@ -34,7 +34,7 @@ extension AllTimeWidgetStats {
             return try decoder.decode(AllTimeWidgetStats.self, from: data)
         } catch {
             DDLogError("Failed loading AllTimeWidgetStats data: \(error.localizedDescription)")
-            return AllTimeWidgetStats()
+            return nil
         }
     }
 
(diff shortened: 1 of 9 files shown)


WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/AllTimeWidgetStats.swift
```diff
@@ -34,7 +34,20 @@ extension AllTimeWidgetStats {
...
         } catch {
             DDLogError("Failed loading AllTimeWidgetStats data: \(error.localizedDescription)")
-            return AllTimeWidgetStats()
+            return nil
+        }
+    }
+
+    static func clearSavedData() {
+        guard let dataFileURL = AllTimeWidgetStats.dataFileURL else {
+            return
```
bjtitus: Is it worth adding any kind of logging here to track when the file URL was empty?

WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/ThisWeekWidgetStats.swift
```diff
@@ -45,7 +45,20 @@ extension ThisWeekWidgetStats {
...
         } catch {
             DDLogError("Failed loading ThisWeekWidgetStats data: \(error.localizedDescription)")
-            return ThisWeekWidgetStats()
+            return nil
+        }
+    }
+
+    static func clearSavedData() {
+        guard let dataFileURL = ThisWeekWidgetStats.dataFileURL else {
+            return
```
bjtitus: Is it worth logging this so we have some idea when this URL is empty?

WordPress/Classes/ViewRelated/Stats/Today Widgets/Data/TodayWidgetStats.swift
```diff
@@ -21,20 +21,33 @@ struct TodayWidgetStats: Codable {
...
-            DDLogError("Failed loading TodayWidgetStats data: \(error.localizedDescription)")
-            return TodayWidgetStats()
+            DDLogError("TodayWidgetStats: Failed loading data: \(error.localizedDescription)")
+            return nil
+        }
+    }
+
+    static func clearSavedData() {
+        guard let dataFileURL = TodayWidgetStats.dataFileURL else {
+            return
```
bjtitus: Is it worth logging here so we know when this URL is missing?

ScoutHarris:
Hey @bjtitus .

> Is it worth logging here so we know when this URL is missing?

Sure, I can do that. I'll add those in my next PR to catch all the places that check the `dataFileURL`. Thanks!

PR #13262, opened 2020-01-22
Make sure we are not checking for failed media on an empty array.

SergioEstevao:
This PR makes the media upload notifications to show again.

I think that the behaviour of `allSatisfy` may have changed between
Swift versions because it looks this was working correctly before.

To test:
 - Open the app and go to a media library of a site
 - Press on the + button and multiple media
 - Wait for the uploads to finish
 - See if the notification/toast is showed correctly

PR submission checklist:


loremattei:
Hey! I moved this to 14.2 because 14.1 has been cut. Please, ping me if you want it to make it to 14.1 😀

PR #13265, opened 2020-01-22
Fix #13258: Post time string uses site time zone

### LINKED ISSUE #13258: Posts List uses device timezone while Post Settings uses site timezone
With [Post Scheduling Enhancements](https://github.com/wordpress-mobile/WordPress-iOS/pull/12820), the scheduled/published date is now displayed in a new view using the site timezone. However, the Posts List still shows the date/time in the device timezone.

Steps to reproduce using a scheduled post:

1. On the device, set the device timezone to Denver.
1. In the app in My Site > settings, set the site timezone to UTC.
1. Go to Blog Posts > Add New.
1. Add a title and content.
1. Go to More > Post Settings > Publish to see Date and Time.
1. Tap to modify and choose a time a few minutes in the future.
1. Note that the post scheduling view and the post settings view both show the site timezone (UTC in this example).
1. Schedule the post.
1. Go to Blog Posts > Schedule and refresh.

**Result:** site timezone is shown in post settings while the device timezone is shown in the Posts List, which can be fairly confusing. ([2m22s](https://cloudup.com/cinor7JfJge))

site-timezone-used-in-post-settings-wpios-14 0 0 0-testflight-beta device-timezone-used-in-post-list-wpios-14 0 0 0-testflight-beta
<sup>Tested with WPiOS 14.0.0.0 TestFlight beta on iPhone 6S iOS 13.3.</sup>

Steps to reproduce using a backdated post:

1. On the device, set the device timezone to Denver.
1. In the app in My Site > settings, set the site timezone to UTC.
1. Go to Blog Posts and open any post for editing.
1. Go to More > Post Settings > Publish to see Date and Time.
1. Back date the post to a time that is one date in the site timezone and another date in the device timezone (i.e. Jan 15 5:00am UTC is Jan 14 11:00pm MST).
1. Update the post.
1. Go to Blog Posts > Published and refresh.

**Result:** Post Settings shows the date of the site timezone but the Published Posts List shows the date of the device timezone, which is very confusing when the time selected happens to span two different dates. ([1m32s](https://cloudup.com/cvKAT0bDREn))

post-settings-utc-example post-list-device-mst-example
<sup>Tested with WPiOS 14.0.0.0 TestFlight beta on iPhone 6S iOS 13.3.</sup>

bjtitus:
Fixes #13258 where the Post list displayed a time string which did not take into account the site time zone, unlike Post Settings.

- Uses [new version](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/240) of WordPressShared which includes overrides for time zones.
- Adds a new `displayDate()` helper which duplicates the logic from `dateStringForDisplay()` with new time zones passed in.

### To test

Follow instructions in #13258 or:

- Set a site time zone two hours away from your own.
- Schedule a post for a time of your choosing.
- Note that the Post List will show a time different than the time in Post Settings.

Note that Scheduling always exhibits this behavior, Drafts and Published will exhibit this behavior only if the elapsed time is greater than 7 days because the relative time formatter should be correct without any time zone adjustment.

PR submission checklist:

- [x] **I have considered adding unit tests where possible.** Unit tests added to WordPressShared

- [x] **I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.** Feature was released in 14.0 so the fix doesn't need an addition.

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -20,7 +20,6 @@
 * Stats: fixed issue that could cause incorrect Stats to be displayed when viewing Stats from a widget.
 * Stats Today widgets: large numbers are now abbreviated.
 * Fixed a bug where files imported from other apps were being renamed to a random name.
-* Fixed a bug where the Post might show a different date in lists than in the Post Settings screen.
 
 13.9
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -20,6 +20,7 @@
 * Stats: fixed issue that could cause incorrect Stats to be displayed when viewing Stats from a widget.
 * Stats Today widgets: large numbers are now abbreviated.
 * Fixed a bug where files imported from other apps were being renamed to a random name.
+* Fixed a bug where the Post might show a different date in lists than in the Post Settings screen.
 
 13.9
 -----
--- a/WordPress/Classes/Extensions/AbstractPost+Dates.swift
+++ b/WordPress/Classes/Extensions/AbstractPost+Dates.swift
@@ -12,14 +12,21 @@ extension AbstractPost {
         let blogService = BlogService(managedObjectContext: context)
         let timeZone = blogService.timeZone(for: blog)
 
+        // Unpublished post shows relative or date string
         if originalIsDraft() || status == .pending {
             return dateModified?.mediumString(timeZone: timeZone)
-        } else if isScheduled() {
-            return dateCreated?.mediumStringWithTime(timeZone: timeZone)
-        } else if shouldPublishImmediately() {
-            return NSLocalizedString("Publish Immediately", comment: "A short phrase indicating a post is due to be immedately published.")
-        } else {
-            return dateCreated?.mediumString(timeZone: timeZone)
         }
+
+        // Scheduled Post shows date with time to be clear about when it goes live
+        if isScheduled() {
+            return dateCreated?.mediumStringWithTime(timeZone: timeZone)
+        }
+
+        // Publish Immediately shows hard coded string
+        if shouldPublishImmediately() {
+            return NSLocalizedString("Publish Immediately", comment: "A short phrase indicating a post is due to be immedately published.")
+        }
+
+        return dateCreated?.mediumString(timeZone: timeZone)
     }
 }
--- a/WordPress/Classes/Extensions/AbstractPost+Dates.swift
+++ b/WordPress/Classes/Extensions/AbstractPost+Dates.swift
@@ -1,0 +1,25 @@
+import Foundation
+
+extension AbstractPost {
+
+    /// The string to display to the user representing a post's date.
+    ///
+    /// - **Scheduled**: Displays time + date
+    /// - **Immediately**: Displays "Publish Immediately" string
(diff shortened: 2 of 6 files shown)


RELEASE-NOTES.txt
```diff
@@ -20,6 +20,7 @@
 * Stats: fixed issue that could cause incorrect Stats to be displayed when viewing Stats from a widget.
 * Stats Today widgets: large numbers are now abbreviated.
 * Fixed a bug where files imported from other apps were being renamed to a random name.
+* Fixed a bug where the Post might show a different date in lists than in the Post Settings screen.
```
ScoutHarris: I don't think this is necessary. The difference only occurred with the new Post Scheduling, right? Which isn't public yet.

WordPress/Classes/Extensions/AbstractPost+Dates.swift
```diff
@@ -0,0 +1,25 @@
...
+    /// - **Immediately**: Displays "Publish Immediately" string
+    /// - **Published or Draft**: Shows relative date when < 7 days
+    public func displayDate() -> String? {
+        let context = managedObjectContext ?? ContextManager.sharedInstance().mainContext
+        let blogService = BlogService(managedObjectContext: context)
+        let timeZone = blogService.timeZone(for: blog)
+
+        if originalIsDraft() || status == .pending {
+            return dateModified?.mediumString(timeZone: timeZone)
+        } else if isScheduled() {
```
ScoutHarris: Since each `if` has a `return`, the `else`s aren't really necessary.

RELEASE-NOTES.txt
bjtitus: @ScoutHarris Ah yes, you're right. 14.0 would be the first release with the feature enabled.

WordPress/Classes/Extensions/AbstractPost+Dates.swift
bjtitus: I often leave the `else` in to make it clear that there isn't one scope that isn't actually returning (easier to skim). I went ahead and removed and added some comments explaining the intent of each block.

ScoutHarris:
Hey @pinarol @loremattei .

Can you please include this in the next 14.0 release? It also needs WordPressShared 1.8.12-beta.2 to go with it ([related PR](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/240)).

Thanks!!

bjtitus:
The failing unit test here is due to a change in `UIFont` scaling: https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/237.

The test needs to be updated to take into account the trait collection of the label since it is auto-adjusted for the trait collections. Something like:

```swift
func testCellTitleIsTheCorrectFont() {
    XCTAssertEqual(cell?.title.font, WPStyleGuide.fontForTextStyle(.body, fontWeight: .semibold, compatibleWith: cell?.title.traitCollection))
}
```

I'll have to make a change to `WordPress-iOS-Shared` to fix the test issue. I was able to verify that the `SiteSegmentsCell` properly adjusts to font changes as expected.

To test:
* Create a new site
* Check the font for the cells displayed under "Tell us what kind of site you'd like to make"

PR #13266, opened 2020-01-23
Today Widget: show no connection view

ScoutHarris:
Ref #12702

For the _Today_ widget, a `No network available` message is displayed if:
- There is no network connection.
- There is no saved data.

If there is saved data, it will be displayed.

To test:

Side note: In the simulator, the widget was never notified when the network connection was re-established. I suggest testing on a real device as the reachability notifications work as expected. 

Since `No network available` will only be displayed if there is no saved data, the data needs to be purged, and then the network disabled.

---
- Add the _Today_ widget to the Today view.
- Return to the app and change the site used for widgets (so the saved data is purged).
- Disable internet connection.
- Return to the Today view.
- Verify the no connection view is displayed.

no_connection

---
- Enable internet connection.
- Verify the data loads.

loaded

---
- Disable internet connection again.
- Verify the data is still displayed.

---
PR submission checklist:

### DIFF
--- a/WordPress/WordPressTodayWidget/TodayViewController.swift
+++ b/WordPress/WordPressTodayWidget/TodayViewController.swift
@@ -86,10 +86,25 @@ class TodayViewController: UIViewController {
 
         coordinator.animate(alongsideTransition: { _ in
             self.tableView.performBatchUpdates({
-                let lastDataRowIndexPath = [IndexPath(row: 1, section: 0)]
+
+                var indexPathsToInsert = [IndexPath]()
+                var indexPathsToDelete = [IndexPath]()
+
+                // If the first row was unconfigured, then rows are just being added.
+                // Otherwise, a data row is being inserted/deleted.
+                if self.tableView.visibleCells.first is WidgetUnconfiguredCell {
+                    let indexRange = (1..<updatedRowCount)
+                    let indexPaths = indexRange.map({ return IndexPath(row: $0, section: 0) })
+                    indexPathsToInsert.append(contentsOf: indexPaths)
+                } else {
+                    let lastDataRowIndexPath = IndexPath(row: 1, section: 0)
+                    indexPathsToInsert.append(lastDataRowIndexPath)
+                    indexPathsToDelete.append(lastDataRowIndexPath)
+                }
+
                 updatedRowCount > self.minRowsToDisplay() ?
-                    self.tableView.insertRows(at: lastDataRowIndexPath, with: .fade) :
-                    self.tableView.deleteRows(at: lastDataRowIndexPath, with: .fade)
+                    self.tableView.insertRows(at: indexPathsToInsert, with: .fade) :
+                    self.tableView.deleteRows(at: indexPathsToDelete, with: .fade)
             })
         })
     }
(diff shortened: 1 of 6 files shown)


ScoutHarris:
Hey @osullivanchris .

I implemented this view as designed. However, I also implemented it so that the widget will automatically update when the connection has been re-established. Which sort of makes the `Retry` action superfluous. Tapping it accomplishes nothing since it's only displayed when there is actually no connection. Once a connection is established, the view automatically updates with data. 

Any thoughts on redesigning this view? (For consideration after this PR, not for it.)

osullivanchris:
Sure. Could we just add a subheading to it, and remove the area with the button so it looks something like this? I also tried sketching it with the subheading in the button area below the divider, but it looked weird to have the information separate.

Screenshot 2020-01-23 at 10 13 30

Also just wondering in what cases we show this. I'm sure you've thought about it more than me. But can we store the information? The situation I'm thinking of is:
- At 10am the stats for today shows 99 views
- At 10:05am I lose internet connection
- If I view the widget at 10:06am, do we show 99 views? Pro: its some information rather than nothing. Con: its out of date
- And similarly if I'm still offline at 2:00PM, or 10:00AM the following day, does it behave the same? 

Thanks!

ScoutHarris:
Hey @osullivanchris .

Thanks for the quick response!

> Could we just add a subheading to it

Sure. I'll take a look at it today. I figured the new layout would be different, so I'll make those changes in a separate PR.

> Also just wondering in what cases we show this. 
> can we store the information?

It's pretty straightforward:
- When stats are accessed in the app for the site used for widgets, the data is stored.
- When a widget is accessed, the stored data is loaded and displayed, then new data is fetched and the widget is updated. 
- This newly fetched data is then stored.

The behavior for no connection:
- If there is no stored data, the no connection view is displayed.
- If there is stored data, the data is displayed.
- Once a connection is re-established, data is fetched and the new data is displayed.

There are two ways widget data can be updated:
- By viewing stats for that site in the app.
- By viewing the widget.

To address your scenario:

> If I view the widget at 10:06am, do we show 99 views?

Yes, we will show the stored data.

> Pro: its some information rather than nothing. Con: its out of date

Agreed. I figured this scenario was better than showing the no connection view when we do have _something_ to show. In addition, this behavior matches the app - that is, we show stored data until new data can be fetched.

> And similarly if I'm still offline at 2:00PM, or 10:00AM the following day, does it behave the same?

Yes. Again, this matches the app. We don't show any indication stats are outdated in the app, but we could think about doing so in the widget if we wanted. I'm not sure where we'd put it, but maybe a 'last updated' timestamp, or a small 'connect your internet yo!' message somewhere.

osullivanchris:
@ScoutHarris thanks for such a thorough explanation! You should capture that somewhere its like a wiki / FAQ :) Totally understand

As I read through the only open question I had was what you arrived on

> I'm not sure where we'd put it, but maybe a 'last updated' timestamp, or a small 'connect your internet yo!' message somewhere.

that's exactly where my line of reasoning brought me too. The reason I gave two time examples was it seems particularly false information when it says "today" but it was the previous day. A few hours later - not so big a deal. Might be more an issue with the label "Today" rather than if we said "January 23" for example. Then if you looked on January 24th you'd know.

In any case its someone without a connection and the expectation may be for stored data. I don't think its a big issue. But we could capture it as another idea/potential enhancement task

WordPress/WordPressTodayWidget/TodayViewController.swift
```diff
@@ -330,7 +391,17 @@ private extension TodayViewController {
 
     func expandedHeight() -> CGFloat {
         var height: CGFloat = 0
-        let dataRowHeight = tableView.rectForRow(at: IndexPath(row: 0, section: 0)).height
+        var dataRowHeight: CGFloat = 0
```
bjtitus: Can this be made immutable since it doesn't change after setting the first time?

ScoutHarris: Done!

ScoutHarris:
Hey @bjtitus . Ready for another review please!

PR #13268, opened 2020-01-23
Fix UI unit tests

SergioEstevao:
This PR changes the accessibility label for the more button on the Post View Controller and also adds an accessibility identifier to make the UI tests more robust.

To test:
 - Run the UI tests
 - Check that they all pass.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditorNavigationBarManager.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditorNavigationBarManager.swift
@@ -37,7 +37,8 @@ class PostEditorNavigationBarManager {
         let button = UIButton(type: .system)
         button.setImage(image, for: .normal)
         button.frame = CGRect(origin: .zero, size: image.size)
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Post/PostEditorNavigationBarManager.swift
```diff
@@ -37,7 +37,8 @@ class PostEditorNavigationBarManager {
         let button = UIButton(type: .system)
         button.setImage(image, for: .normal)
         button.frame = CGRect(origin: .zero, size: image.size)
-        button.accessibilityLabel = NSLocalizedString("More", comment: "Action button to display more available options")
+        button.accessibilityLabel = NSLocalizedString("More Post Options", comment: "Action button to display more available options")
```
rachelmcr: We'll probably need a label that doesn't specifically mention "Post" here, since this button is also used in the page editor.

SergioEstevao:
@rachelmcr I updated the PR with your suggestions.

Issue #13271, opened 2020-01-23
Stats Widgets: show no connection view


ScoutHarris:
For all the Stats widgets, show a `No network available` message if:
- There is no network connection.
- There is no saved data.

Tasks:
- [x] _Today_: Update view (previously added with https://github.com/wordpress-mobile/WordPress-iOS/pull/13266) (#13273)
- [x] _All-Time_: show view (#13283)
- [ ] _This Week_: show view (#13284)

Design:
<img width="400" alt="no_network_available" src="https://user-images.githubusercontent.com/1816888/73008631-2d3fcc80-3dcc-11ea-9deb-d332f5bb538b.png">

PR #13273, opened 2020-01-23
Today Widget: update no connection view

ScoutHarris:
Ref #13271

This updates the no connection view in the _Today_ widget added in #13266. It is displayed when:
- There is no network connection.
- There is no saved data.

If there is saved data, it will be displayed.

To test:

Side note: In the simulator, the widget was never notified when the network connection was re-established. I suggest testing on a real device as the reachability notifications work as expected. 

Since `No network available` will only be displayed if there is no saved data, the data needs to be purged, and then the network disabled.

---
- Add the _Today_ widget to the Today view.
- Return to the app and change the site used for widgets (so the saved data is purged).
- Disable internet connection.
- Return to the Today view.
- Verify the new no connection view is displayed.

no_network

---
- Enable internet connection.
- Verify the data loads.

loaded

---
- Disable internet connection again.
- Verify the data is still displayed.

---
PR submission checklist:


ScoutHarris:
Hey @osullivanchris .

I adjusted the layout, and updated the screenshot in the description. See what you think.

ScoutHarris:
Thanks @osullivanchris !

@bjtitus ready for a review when you get a chance please.

ScoutHarris:
Thanks @bjtitus !

PR #13274, opened 2020-01-23
Fixed improper spacing on notice for saving drafts offline

### LINKED ISSUE #12799: When exiting the editor, Notices cover the tab bar
## Expected behavior

The Notice is displayed above the tab bar.

## Actual behavior

The Notice is displayed on top of the tab bar, **covering** it. 

1. While offline, create a post. 
2. Tap X → Save Draft

A Notice saying something like “We'll save your draft...”. should be shown. 

Tested on iPhone XS, iOS 12.4.1, WPiOS `develop` (d2d9ada).

kduncanwelke:
Removed check for tabBarController.presentedViewController being nil. When saving a draft in offline mode the editing view is dismissed - this causes the check for the presentedViewController to return nil, thus placing the notice incorrectly, without the proper space for the tabBar. Other notices continue to display as normal (removing the nil does not appear to have had negative effects).

Fixes #12799

To test:

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Notices/UntouchableWindow.swift
+++ b/WordPress/Classes/ViewRelated/System/Notices/UntouchableWindow.swift
@@ -30,9 +30,12 @@ class UntouchableViewController: UIViewController {
     }
 
     var offsetOnscreen: CGFloat {
+        // removed check for tabBarController.presentedViewController being nil here
+        // when saving a draft in offline mode the editing view is dismissed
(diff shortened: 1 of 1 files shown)


frosty:
@kduncanwelke Thanks for making another contribution, and huge apologies for the delay reviewing this. We'll take a look at it shortly!

leandroalonso (approved):
Hi @kduncanwelke, thanks for the fix!

Can you please merge `develop` here so the CI run the tests?

PR #13287, opened 2020-01-27
Merge release 14.0 into master

loremattei:
This PR aims to merge release 14.0 into master.
We're slightly breaking GitFlow here in order to keep `master` in a good shape: https://github.com/wordpress-mobile/WordPress-iOS/pull/13265 was merged in the `release/14.0` branch with a failing test. 
The code was actually good and the test needed to be updated, but the PR broke the tests for the branch. 
I've cherry-picked the fix from `develop` here, so `master` will have a commit after the `14.0` tag, but the tests will succeed. 
Since the cherry-picked commit is already in `develop` and it will be part of `14.1`, `master` will get back to normal status after the `14.1` release.


Podfile
```diff
@@ -11,14 +11,14 @@ workspace 'WordPress.xcworkspace'
 ##
 def wordpress_shared
     ## for production:
-    pod 'WordPressShared', '1.8.10'
+    pod 'WordPressShared', '1.8.12-beta.2'
```
jkmassel: It seems odd that we have a beta version here?

loremattei: Yep! It is. We had a very late fix merged on Friday that was approved this way. A beta 3 was already released, so the clean way would had been to branch from 1.8.11, reapply the patch and release a new version, but this way we had very little testing time, so decided to move this ahead with the beta for this time.

PR #13310, opened 2020-01-29
Add RN patch for RCTShadowView.m to prevent crash in nested group block

dratwas:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1747

There is an issue on iOS with nested views.

>Searching for the ancestor view while measuring a shadow view layout relative to the ancestor is limited to 30. I couldn't find the reason nor commit where it was introduced (because I have no access to internal facebook commits).
I think we should keep looking for an ancestor view until the shadow view has the superview property.
https://github.com/facebook/react-native/pull/26986

This is fixed here https://github.com/facebook/react-native/pull/26986
It's merged to master and will land in 0.62 version of `react-native`, however it is a blocker for us since we can not nest more than 3 group blocks because of that. We could cherry-pick this commit to our fork but as I can see we do not use our fork anymore.

To fix that on our side we used a patch mechanism. The mechanism is very simple. I created a patch that is in the `patches` directory and contains my changes. The patch is applied in `post_install` hook of cocoapods. 

To test:
 - install dependencies
- run app and open gutenberg editor
- add at least 4 nested group blocks
- the app shouldn't crash


lukewalczak:
Tested manually in WordPress-iOS app - nested 5 groups:

nesting_groups

PR #13323, opened 2020-01-29
Stats: fix background color when viewed in a modal

### LINKED ISSUE #13301: Stats: when viewed in a modal, there is a grey bar at the bottom of the view
### Expected behavior
When viewing Stats in a modal, the bottom area should be consistent with app colors.

### Actual behavior
When viewing Stats in a modal, the bottom area is a light grey, in both light and dark modes.

AFAIK, the only way to view Stats in a modal is to open Stats from a Today widget, so:
- In the app, go to Stats > Widgets > Use this site.
- Add a Stats widget to the device's Today view.
- Tap a widget. The app will launch, displaying the widget site's stats in a modal.
- Notice the area at the bottom.

dark

light

ScoutHarris:
Fixes #13301 

To test:

Note: This appears to have been an issue only on devices without a home button. Like iPhone X* and iPhone 11*.

- Run the app.
- Go to Stats > Widgets > Use this site.
- Add any Stat widget to the Today view.
- Tap on the widget. The app will launch, displaying the Stats in a modal.
- Verify the bottom of the view is extended.

Before | After
--------|-------
before_light        |       after_light
before_dark        |       after_dark

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/SiteStatsDashboard.storyboard
+++ b/WordPress/Classes/ViewRelated/Stats/SiteStatsDashboard.storyboard
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="14490.70" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="Tqa-zb-Liz">
-    <device id="retina4_7" orientation="portrait">
-        <adaptation id="fullscreen"/>
-    </device>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="15705" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="Tqa-zb-Liz">
+    <device id="retina4_7" orientation="portrait" appearance="light"/>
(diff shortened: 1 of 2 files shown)


bjtitus:
Would it make more sense to extend this view underneath the home indicator? Most other scroll views (Contacts.app, Settings.app) do this so that the content is visible underneath:

| PR | Extended |
|--|--|
| Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-29 at 17 57 57 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-01-29 at 17 55 02 |

Changing the bottom constraint of in the container view in `SiteStatsDashboard` to the bottom of the superview instead of the safe area accomplishes this. I made [a small patch](https://github.com/wordpress-mobile/WordPress-iOS/files/4131654/SiteDashboardBottomConstraint.patch.zip) showing that change.

ScoutHarris:
Hey @bjtitus .

> Would it make more sense to extend this view underneath the home indicator?
> Changing the bottom constraint of the container view in SiteStatsDashboard to the bottom of the superview instead of the safe area

I don't think so. I tried that, and the text is covered by the bar the bottom, and can be cut off on the sides.

Screen Shot 2020-01-29 at 6 22 14 PM

Screen Shot 2020-01-29 at 6 22 35 PM

Perusing the app, there are several places we overcome this issue by setting the background color, and not extending the view. So I think my approach is:
1) More consistent.
2) Looks better.

bjtitus:
@ScoutHarris Out of curiosity, where else do we do this? Notification Settings and Create Site both seem to use the behavior I was suggesting.

ScoutHarris:
Hey @bjtitus .

> Notification Settings and Create Site both seem to use the behavior I was suggesting.

> inconsistent with the platform and Apple's guidelines for developers

Fair enough. I've removed the background color and modified the storyboard, and updated the PR description. See what you think now.

PR #13333, opened 2020-01-30
Stats Widgets: show failed loading view

ScoutHarris:
Ref #12702 

For all the Stats widgets, if fetching data fails, a `Couldn't load data` view is displayed.

To test:

A bit of hackery is needed to force data fetching to fail. In the `retrieveSiteConfiguration` method in all three VCs, you can replace this line:

```
siteID = sharedDefaults.object(forKey: WPStatsTodayWidgetUserDefaultsSiteIdKey) as? NSNumber
```

With this magic:

```
if loadingFailed {
    siteID = sharedDefaults.object(forKey: WPStatsTodayWidgetUserDefaultsSiteIdKey) as? NSNumber
} else {
    siteID = 9999999999999999
}
```

That will cause the first attempt to fail by setting an invalid `siteID`, and the second attempt to succeed by setting the correct `siteID`.

Now, to actually test:
- In the app, go to Stats > Widgets > Use this site.
- Add the Stats widgets to the device's Today view.
- Initially, stored data will be displayed. But once the data fetching is attempted and fails, the failure view will be displayed.

fail

- Tap the widgets.
- They should load data successfully and update accordingly.

loaded

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
@@ -9,9 +9,9 @@ enum WidgetType {
     var configureLabelFont: UIFont {
         switch self {
         case .loadingFailed:
-            return UIFont.systemFont(ofSize: UIFont.preferredFont(forTextStyle: .headline).pointSize)
+            return WidgetStyles.headlineFont
         default:
-            return UIFont.systemFont(ofSize: UIFont.preferredFont(forTextStyle: .footnote).pointSize)
+            return WidgetStyles.footnoteNote
         }
     }
 }
--- a/WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
@@ -4,6 +4,16 @@ enum WidgetType {
     case today
     case allTime
     case thisWeek
+    case loadingFailed
+
+    var configureLabelFont: UIFont {
+        switch self {
+        case .loadingFailed:
+            return UIFont.systemFont(ofSize: UIFont.preferredFont(forTextStyle: .headline).pointSize)
+        default:
+            return UIFont.systemFont(ofSize: UIFont.preferredFont(forTextStyle: .footnote).pointSize)
+        }
+    }
 }
 
 class WidgetUnconfiguredCell: UITableViewCell {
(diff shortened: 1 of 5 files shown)


osullivanchris:
Hey @ScoutHarris what would you like me to review here? Just want to be sure I know what's new/needing input.

IIRC last time we looked at offline states they just automatically updated when connection returned, there was no Retry button. Is the Retry button new, and is that what you need me to check?

WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
```diff
@@ -4,6 +4,16 @@ enum WidgetType {
     case today
     case allTime
     case thisWeek
+    case loadingFailed
+
+    var configureLabelFont: UIFont {
+        switch self {
+        case .loadingFailed:
+            return UIFont.systemFont(ofSize: UIFont.preferredFont(forTextStyle: .headline).pointSize)
```
mindgraffiti: Do the font configs belong in `WPStyleGuide+Stats`?

WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
```diff
@@ -4,6 +4,16 @@ enum WidgetType {
...
+        default:
+            return UIFont.systemFont(ofSize: UIFont.preferredFont(forTextStyle: .footnote).pointSize)
```
mindgraffiti: same here?

WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
```diff
@@ -45,10 +55,13 @@ private extension WidgetUnconfiguredCell {
                 return LocalizedText.configureAllTime
             case .thisWeek:
                 return LocalizedText.configureThisWeek
+            case .loadingFailed:
+                return LocalizedText.loadingFailed
```
mindgraffiti: Oooh this is nice! I like this naming convention. I've always done `Constants.loadingFailed`

WordPress/Classes/ViewRelated/Stats/Today Widgets/Shared Views/WidgetUnconfiguredCell.swift
ScoutHarris: Nice catch. `WPStyleGuide+Stats` is for the app Stats, but there is a `WidgetStyles` specifically for the Widgets. So I moved the font definitions there.

ScoutHarris:
Hey @osullivanchris .

My apologies for not being clear.

This is a new view and behavior, so I just wanted to be sure it looked and behaved as expected.

> IIRC last time we looked at offline states they just automatically updated when connection returned, there was no Retry button. 

This isn't really an offline state per se. We still have the 'no connection' view for that. This is specifically for the off chance that something goes wrong when the widget attempts to fetch new data. In this case, the widget can't update itself because there's no clear failure - all it knows is the network call failed. So the Retry is to:
1. Notify the user something went wrong (because it won't be updated automatically).
2. Let them manually attempt a refresh.

To note, when you scroll away from and back to the Today view, the widget will attempt to load data again, but because the widget is recreated. While the widget persists it won't update automatically.

ScoutHarris:
Hey @osullivanchris . I'm going to go ahead and merge this. If any changes are needed, I can do a follow-up PR. Thanks!

osullivanchris:
Looks fine to me, sorry I didn't respond in time!

PR #13350, opened 2020-01-31
[Media Editing] Load iCloud Photos

leandroalonso:
Fixes https://github.com/wordpress-mobile/MediaEditor-iOS/issues/10

### To test

1. Try to edit any media that only exists in iCloud (and it's not stored in your device)
2. Check that the image loads correctly

PR submission checklist:


leandroalonso:
@frosty when you select multiple images and tap the "Crop" icon the images aren't loading?

frosty:
So with a single image, it loads as soon as I tap 'Preview 1'. With multiple images, when I tap 'Preview X' I still see the blurry versions. If I tap the Crop button on one of them then that image loads, although I'm still having the issue with the bottom toolbar so I'm unable to change or confirm the crop. I recorded a video for you: https://cloudup.com/cwqv7XPSgDb

leandroalonso:
Thank you so much for the video, @frosty! 

Apparently, the `actionBar` always appears when the library is updated: like when you took screenshots or when an iCloud photo is loaded (the case on that video).

I've addresses two things:

* The `actionBar` will not appear anymore. We hide it before showing the MediaEditor and then show it again (if needed);
* The Media Editor will update the image with a slightly better quality one (based on device screen size) even if you don't tap the "Crop" icon.

Can you please test again? 🙏

PR #13355, opened 2020-02-03
Sentry logging: Log filename instead of absolute path

frosty:
This is a small tweak to our sentry logging so that we use the filename of a crash location instead of the absolute path (this was resulting in logging the absolute path to the file for whoever happened to be the release manager for a given release). This should help Sentry better group together these crashes.

**To test:**

- To test, I edited the `persistentStoreCoordinator` method in `ContextManager`, and added the following at line 248:

```
error = [NSError errorWithDomain:NSSQLiteErrorDomain code:1 userInfo:nil];
SentryStartupEventAddError(startupEvent, error);
```

- I then inserted a breakpoint in the method modified by this PR, to check the incoming filename was logged correctly.

PR submission checklist:

- [ x I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.


WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
```diff
@@ -9,7 +9,8 @@ startup time. This will block the thread. Do not use unless you're sure.
     private var errors = [String]()
 
     func add(error: NSError, file: String = #file, function: String = #function, line: UInt = #line) {
-        errors.append("\(function) (\(file):\(line)) \(error.localizedDescription) | userInfo: \(error.userInfo)")
+        let filename = (file as NSString).lastPathComponent
```
koke: Since we're only using this for one thing, it should be OK, but this is stripping too much information unless we ensure all our filenames are unique.

Also, the `userInfo` will contain a `NSFilePath`, with a device-specific UUID, so this won't likely help with grouping on sentry.

I think the deeper problem is that we are setting `event.message = "\(title): \(lastError)"`, and sentry isn't showing the breadcrumbs with all the error details. Otherwise, we should set a generic title and leave the error details outside of it

frosty:
@koke Sorry for the delay here... do you think this is any better?

@jkmassel Would you mind taking a look at this too?

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
```diff
@@ -26,11 +34,11 @@ startup time. This will block the thread. Do not use unless you're sure.
...
         let semaphore = DispatchSemaphore(value: 0)
         let event = Event(level: .debug)
-        let lastError = errors.removeLast()
-        event.message = "\(title): \(lastError)"
+        event.message = title
         for error in errors {
             let breadcrumb = Breadcrumb(level: .debug, category: "Startup")
-            breadcrumb.message = error
+            breadcrumb.message = error.0
+            breadcrumb.data = error.1
```
koke: You can make this into named variables as you iterate for a bit more clarity:

Screen Shot 2020-02-11 at 09 53 26

koke:
The code looks better, do you have a sample Sentry crash with the new format to see how it would look?

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
frosty: Ah nice, thank you 👍

frosty:
Sure, here's an example issue: https://sentry.io/share/issue/6dc6d59759f44099b7acea5b53e5f0a7/

It would be nice if it was slightly better organised, but I don't see any other usable properties in the event classes than putting this in the breadcrumbs. I don't even see a way to provide a separate title and message, which is why the message itself doesn't contain much information.

koke:
> It would be nice if it was slightly better organised, but I don't see any other usable properties in the event classes than putting this in the breadcrumbs. I don't even see a way to provide a separate title and message, which is why the message itself doesn't contain much information.

You can use the `extra` property:

```swift
event.extra = errors.enumerated().reduce(into: [String: Any](), { (result, arg1) in
    let (index, error) = arg1
    let (message, info) = error
    result["Error \(index + 1)"] = [
        "Method": message,
        "Error": info.description,
    ]
})
```

This generates something like this in Sentry:

Screen Shot 2020-02-11 at 17 15 20

https://sentry.io/share/issue/6dc6d59759f44099b7acea5b53e5f0a7/

frosty:
@koke Thanks, that's much better! I've just pushed that change now.

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
```diff
@@ -6,10 +6,19 @@ WARNING: This class was created to track events of failures during
...
+    private typealias UserInfo = [String: Any]
+    private typealias ErrorWithInfo = (String, UserInfo)
+
+    private var errors = [ErrorWithInfo]()
 
     func add(error: NSError, file: String = #file, function: String = #function, line: UInt = #line) {
-        errors.append("\(function) (\(file):\(line)) \(error.localizedDescription) | userInfo: \(error.userInfo)")
+        let filename = (file as NSString).lastPathComponent
+        let info: UserInfo = [
+            "Description": error.description,   // Intentionally not using `localizedDescription`, as it helps to have error in Sentry in a single language
```
koke: To clarify, `description` contains a description of the error object. Something like this:

> Error Domain=NSURLErrorDomain Code=-1004 "No se ha podido establecer conexión con el servidor." UserInfo={NSUnderlyingError=0x2832203f0 {Error Domain=kCFErrorDomainCFNetwork Code=-1004 "(null)" UserInfo={_kCFStreamErrorCodeKey=61, _kCFStreamErrorDomainKey=1}}, NSErrorFailingURLStringKey=http://192.168.1.38:8081/status, NSErrorFailingURLKey=http://192.168.1.38:8081/status, _kCFStreamErrorDomainKey=1, _kCFStreamErrorCodeKey=61, NSLocalizedDescription=No se ha podido establecer conexión con el servidor.}

It's not an English version of the `localizedDescription` method, and the human-readable messages in there will still be localized.

I suggested the other code snippet trying to avoid changing the current code too much, but if we're going with this solution, I'd probably store the error directly here instead of the `UserInfo` type.

Perhaps instead of `ErrorWithInfo` we can store an array of `ErrorWithCaller = (Error, String)`, or make it a proper struct so the enumeration code looks a bit cleaner.

frosty:
Okay, I've taken another attempt – how's this looking?

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
```diff
@@ -25,13 +34,20 @@ startup time. This will block the thread. Do not use unless you're sure.
...
+
+        event.extra = errors.enumerated().reduce(into: [String: Any](), { (result, arg1) in
+            let (index, errorWithCaller) = arg1
+            let error = errorWithCaller.error
+            result["Error \(index + 1)"] = [
+                "Method": errorWithCaller.caller,
+                "Domain": error.domain,
+                "Code": error.code,
+                "Description": error.localizedDescription,
+                "User Info": error.userInfo.description
```
koke: This is fine and might actually help reading reports more easily, but just in case you didn't know, you are not limited to strings for those fields, you could have `error.userInfo` directly and sentry would show the right thing:

Screen Shot 2020-02-14 at 12 50 21

frosty: Hmm, I tried that and Sentry's SDK complained that it couldn't be serialised to JSON 🤔

koke: Ah yes, I think I saw some of that if any of the data types in userInfo wasn't serializable, `description` is probably safer 😄

PR #13357, opened 2020-02-03
Removes the postPreview feature flag.

### LINKED ISSUE #13180: Migrate PostPreviewViewController to WKWebView

diegoreymendez:
Closes #13180 (by superseding it).

This PR shouldn't change the existing logic, since the feature flag is set to `true` at all times.  There's just one exception for which I'll add a comment in the code for the reviewer.

## PR submission checklist:


WordPress/Classes/ViewRelated/Post/PostNoticeNavigationCoordinator.swift
```diff
@@ -23,23 +23,13 @@ class PostNoticeNavigationCoordinator {
                 return
         }
 
-        if FeatureFlag.postPreview.enabled {
-            let controller = PostPreviewViewController(post: page)
```
diegoreymendez: @bjtitus - I think this was a bug in the original code, since we were showing `PostPreviewViewController` when `FeatureFlag.postPreview.enabled`.  So I just left the `else` condition instead.

Can you double-check that the change is fine?

bjtitus:
I think we may want to wait on merging this until 14.2 is frozen with the feature flag enabled but still available and then remove the flag in 14.3. Just so that there is a way to turn off the feature. @frosty Does that sound right?

frosty:
@diegoreymendez I agree with Brandon, would it be possible for us to leave this feature flag in place for a little longer? The feature hasn't been released yet, and typically we'd leave the feature flag in place for a couple of releases just in case we needed to roll it back for some reason.

Is there a reason we need to remove this now?

diegoreymendez:
@frosty - There's no urgency per-se other than this being a blocker for me to wrap up the `WKWebView` migration project, since there's no sense in migrating a VC that's soon to be deleted.

My line of thinking for this PR was that if we want to recover the old code we always have git to recover it... but I imagine you prefer it to be more handy with the feature flag, correct?

From a practical point of view I could:

1. Merge this and finish the migration project.
2. Wait for this to be merged to finish the migration project.
3. Just consider the migration project finished and agree that this it out of scope.

I'll think about it more, but for the time being I think it's fine not to merge this.

Let me know your thoughts, and we can proceed to close this PR if we all agree!

diegoreymendez:
Yeah, I just went ahead and marked `PostPreviewViewController` as being out of scope for the migration: https://github.com/wordpress-mobile/WordPress-iOS/issues/13180

I'm closing this PR for now.

PS: @bjtitus - if you haven't already, you should consider switching the reverted IF condition I marked in this PR, in the live code.

PR #13367, opened 2020-02-04
Enable Group block in production

jbinda:
Updates Gutenberg so the `Group` block is available in production.

`Gutenberg PR`:  https://github.com/WordPress/gutenberg/pull/20004

To test:
- Run the app without metro running
- Start a new post
- Verify Group block is visible in the block picker
- Select the `Group` block
- Verify it is working as expected:
  - Try to break it
  - Use try portrait & landscape modes 
  - Try nesting it more than expected
  - Check if the navigation button on `FloatingToolbar` works
  - Check if `AppenderButton` is visible when `Group` is selected
  - Check if `AppenderButton` is show as placeholder when `Group` is unselected
  - Check if `AppenderButton` open the `Inserter` after press (when `Group` block is empty and when it already has some nested blocks)
  - Move the block between other ones inside a `Group` and check it works correctly
  - Move the whole `Group` block between another `Group` blocks in RootList
  - Try to use `Ungroup` button on `Toolbar`
  - Remove the block
  - Add one again
 - Use DarkMode
- Save as draft
- Open your draft and check the Group block is shown correctly

PR submission checklist:


jbinda:
Closed because it was opened just for testing purpose

Issue #13318, opened 2020-01-29
Information Architecture: Quick Actions


bjtitus:
Design: https://wp.me/paCL20-ew-p2
Latest Updates: https://wp.me/pauD4L-nx-p2

![quick_actions-2](https://user-images.githubusercontent.com/3250/73396146-0850cb00-429e-11ea-8e2f-66d4ccfed628.png)

- [x] Add feature flag. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13373)
- [x] Add new My Site Header. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13373)
  - [x] Add Spotlight view for quick start experience (https://github.com/wordpress-mobile/WordPress-iOS/pull/13416)
  - [x] ~Add Drop Target API~ (https://github.com/wordpress-mobile/WordPress-iOS/pull/13409)
  - [x] ~Verify removal of cells in Publish section~ _We don't want to remove these cells at this time_ (https://github.com/wordpress-mobile/WordPress-iOS/pull/13373#issuecomment-582852690)
- [x] Add new Quick Action buttons. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13373)
- [x] Ensure views are correct for Dark Mode.
- [x] Add Accessibility support. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13461)
- [x] Add Tracks. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13461)
- [ ] Enable feature flag for internal.
- [ ] Enable feature flag for all.
- [ ] Remove feature flag.

PR #13373, opened 2020-02-05
Information Architecture Quick Actions implementation

bjtitus:
This adds a Feature Flag (`quickActions`) and initial implementation for a new Quick Actions design on the Blog Details page. See https://github.com/wordpress-mobile/WordPress-iOS/issues/13318 for designs.

  

Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-04 at 21 00 25

Known issues:

Missing:
- The Spotlight view for quick start experience (which highlights the site icon) is missing
- The Drop Target API is unimplemented so you cannot drag and drop an image onto the site icon

Both tasks have been added to the issue to be completed before enabling the feature flag for release: https://github.com/wordpress-mobile/WordPress-iOS/issues/13318

To test:

- View a Site's details page

PR submission checklist:

- [x] ~I have considered adding unit tests where possible.~ Will follow up in future PR
- [x] ~I have considered adding accessibility improvements for my changes.~ Did check some basic accessibility but will also follow up with future PG.
- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~


bjtitus:
> On iPad, there's no padding on the left of the site title and URL

Sorry about that. Thanks for catching it! I decided to just constrain these to the `layoutMarginsGuide` so that the spacing will adjust using the safe area where the notch exists. 

This does mean the alignment may seem a bit odd in Landscape on those devices. It does seem a bit odd at first glance but the labels do line up with the other elements according to the safe area. Here are some examples:

Screen Shot 2020-02-05 at 9 51 37 PM
Screen Shot 2020-02-05 at 9 51 46 PM

> Regarding the constraint warnings ... I'm still seeing them on app launch

Looks like this is the `UIView-Encapsulated-Layout-Height` zero height constraint that UITableView adds before the layout pass has happened. I've reduced the priority of one of the stack view vertical constraints so the false zero height shouldn't lead to the breakage anymore. I'm not seeing any more auto layout warnings on switching between sites. Let me know if you're seeing anything else.

frosty:
> @osullivanchris can you confirm whether the items should remain in the list with the new buttons added, meaning the items would be duplicated?

For now, the items should remain both in the list and as quick action buttons 👍 Thanks for checking!

osullivanchris:
Hey @bjtitus ! Thanks for the ping. I chatted to @frosty about the above truncation issue he mentioned. There's a couple of things I'll throw in

- It does look wrong to me when the layout only has left padding as mentioned. Is the safe area to account for the hardware notch? Even if the spacing cannot be even left and right, I think we should add a bit of padding to the right also, it will balance things a bit
- The icon colours should be the same colours as the icons in the list items
- More truncation issues could occur on landscape/tablet but I haven't come up with a good resolution yet 
- One thing discussed with @frosty was that we could revert to a smaller header on iPhone landscape where vertical space is at such a premium. But given such an issue only arises on landscape and with certain models of phone, I would only consider the below if it is easy to do (the big new header would remain on portrait and on tablet)
Screenshot 2020-02-06 at 12 48 42
- This is a very quick rough mock, but you can also see where I'm suggesting we add at least _some_ right padding to balance things out (in turqoiuse). I also think if it helps, the padding between the Quick Actions (yellow) can just be distributed evenly rather than having to conform to a set number like 8pt, 16pt or whatever...if that helps at all!

bjtitus:
Thanks for all of the feedback @osullivanchris!

> Is the safe area to account for the hardware notch?

Yes, the safe area accounts for the notch in landscape. The same safe area is also applied to the side _without_ the notch (that's why the back button sits so far from the left side), which leads it to look a little strange on that side.

> The icon colours should be the same colours as the icons in the list items

Changed this out so it should match exactly now.

> I'm suggesting we add at least _some_ right padding to balance things out

We chatted about this over slack but here are the screenshots with some padding added to the right and reduced padding on the left:

Screen Shot 2020-02-06 at 9 40 35 AM
Screen Shot 2020-02-06 at 9 40 46 AM

> we could revert to a smaller header on iPhone landscape where vertical space is at such a premium

I think this is definitely possible since this would happen when a compact height is used. I'm not sure how useful it really is since all of the content is in a scroll view. It wouldn't be _trivial_ to implement but we could definitely give it a shot if it would make navigation better.

bjtitus:
@frosty @osullivanchris

> is it possible to give the labels a little bit more margin on either side

I adjusted the spacing a bit to allow the label to expand. This won't always work when the width gets too small, but it helps in the portrait iPhone mode for German, at least.

ScoutHarris:
Hey @bjtitus .

I'm still seeing one constraint warning, when logging in/switching sites (still on the iPhone Pro Max simulator).

Let me know if prefer to fix it on this PR or another, and I'll review accordingly.

```
2020-02-06 15:08:58.817325-0700 WordPress[59663:1320919] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x6000032857c0 UIStackView:0x7fdf85c3a4d0.leading == UILayoutGuide:0x600002bb6140'UIViewSafeAreaLayoutGuide'.leading   (active)>",
    "<NSLayoutConstraint:0x600003284050 H:|-(>=8)-[UIStackView:0x7fdf85c3a4d0]   (active, names: '|':WordPress.NewBlogDetailHeaderView:0x7fdf85c2c0c0 )>",
    "<NSLayoutConstraint:0x600003285810 'UIViewSafeAreaLayoutGuide-left' H:|-(0)-[UILayoutGuide:0x600002bb6140'UIViewSafeAreaLayoutGuide'](LTR)   (active, names: '|':WordPress.NewBlogDetailHeaderView:0x7fdf85c2c0c0 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x6000032857c0 UIStackView:0x7fdf85c3a4d0.leading == UILayoutGuide:0x600002bb6140'UIViewSafeAreaLayoutGuide'.leading   (active)>
```

bjtitus:
@ScoutHarris I just pushed a fix for that particular issue. I do see some other autolayout warnings at launch on iPad but they don't appear to be the same stack views from the the header. This should be ready for another look if you get a chance.

osullivanchris:
Thanks for making the tweaks, things are looking much better now! 

> I think this is definitely possible since this would happen when a compact height is used and we can swap our layout when that trait collection changes. I'm not sure how useful it really is since all of the content is in a scroll view anyway. It's not exactly easy but we could definitely give it a shot if it would make navigation better.

I'm feeling like there's not enough value to make this change just for one device and orientation. So maybe we should just leave it.

PR #13386, opened 2020-02-05
Add localization doc

shiki:
This adds a short guide for how to localize strings. Most of the information was taken from @mindgraffiti and @loremattei's work. 

## Reviewing 

1. View the file [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/ee60ffa8f2ba0ada3abf9b98ff44ad59e520f3fd/docs/localization.md).
2. Please feel free to scrutinize. 😄 
3. Please check that the examples are correct.

## Submitter Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,6 +20,7 @@
 
 ## Others 
 
+- [Localization](localization.md)
 - [Feature Flags](feature-flags.md)
 - [Formattable Content](formattable-content.md)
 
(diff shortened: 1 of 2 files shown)


shiki:
> Looks good! The only thing I would suggest adding is the reason why we don't use iOS pluralizations (it breaks glotpress).

Thanks! Added in 18ed6cda35ea167a1fdda595d9b1ecb2b7d0bc5b. 

> Thanks for writing such beautiful documentation!

Thank you! 😊 

---

Thank you for reviewing @mindgraffiti. 🙏 Just waiting for the green build before merging.

PR #13389, opened 2020-02-06
Migrates SharingAuthorizationWebViewController to Swift.

diegoreymendez:
Migrates SharingAuthorizationWebViewController to Swift.

## Important:

As discussed privately with @leandroalonso I'm momentarily adding comments with the previous ObjC code.  This is meant to provide an easier reference to evaluate the migration.

## To test:

Blog List > Pick a test blog > Sharing

Try connecting to each of the services listed there.

## PR submission checklist:


WordPress/Classes/ViewRelated/Blog/SharingAuthorizationWebViewController.swift
```diff
@@ -0,0 +1,503 @@
...
+//@property (nonatomic, strong) PublicizeService *publicizer;
+    private let publicizer: PublicizeService
+
+
+//@property (nonatomic, strong) NSMutableArray *hosts;
+    private var hosts = [String]()
+
+// From old header...
+//@property (nonatomic, weak) id<SharingAuthorizationDelegate> delegate;
+    private let delegate: SharingAuthorizationDelegate
```
leandroalonso: I guess the delegate should be a `weak` var. Otherwise, we might end with cycle references.

diegoreymendez: Changed!

PR #13390, opened 2020-02-06
Fix media library pick while offline

### LINKED ISSUE #11426: Offline Support: Can't bring up Media Library issues while offline.
While editing a post offline, you cannot choose photos from the Media Library. Bringing up the Media Library shows an alert that, when tapped, immediately dismisses it.

SergioEstevao:
Fixes #11426 

To test:
 - Start the app
 - Select a site where you opened the site Media Library before
 - Start a new post ( test with GB and Aztec)
 - Go offline
 - Tap on add media 
 - Select the option to pick from the site MEdia Library
 - Check if you are able to pick a cached media object and insert on the post

 - Start the app
 - Select a site where you **never** open the site Media Library on the device
 - Start a new post ( test with GB and Aztec)
 - Go offline
 - Tap on add media 
 - Select the option to pick from the site MEdia Library
 - Check that you see an alert that says that you are unable to sync with the server

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
+++ b/WordPress/Classes/ViewRelated/Aztec/ViewControllers/AztecPostViewController.swift
@@ -365,7 +365,9 @@ class AztecPostViewController: UIViewController, PostEditor {
     /// Media Library Data Source
     ///
     lazy var mediaLibraryDataSource: MediaLibraryPickerDataSource = {
-        return MediaLibraryPickerDataSource(post: self.post)
+        let dataSource = MediaLibraryPickerDataSource(post: self.post)
+        dataSource.ignoreSyncErrors = true
+        return dataSource
     }()
 
     /// Device Photo Library Data Source
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
@@ -20,7 +20,9 @@ class GutenbergMediaPickerHelper: NSObject {
     /// Media Library Data Source
     ///
     fileprivate lazy var mediaLibraryDataSource: MediaLibraryPickerDataSource = {
-        return MediaLibraryPickerDataSource(post: self.post)
+        let dataSource = MediaLibraryPickerDataSource(post: self.post)
+        dataSource.ignoreSyncErrors = true
+        return dataSource
     }()
 
     /// Device Photo Library Data Source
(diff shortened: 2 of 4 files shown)


diegoreymendez:
The first test case, when adding an image from a site with a cached media library, I'm getting stuck here in Gutenberg:

Screen Shot 2020-02-06 at 3 00 57 PM

Also, as an additional note... as part of our [Offline Principles](https://automattic.design/2019/06/21/offline-principles/) we're trying to avoid blocking alerts as much as possible, since they often require the user to take action often without providing choice or value - for example in cases where there's only 1 button for the user to tap, like in this case.

Screen Shot 2020-02-06 at 3 11 45 PM

I believe that this is a message that could be shown inline as part of a no-results view, providing a much better offline experience.  The example below is one that I feel would fit very well in this case:

Screen Shot 2020-02-06 at 3 17 06 PM

Maybe you can share your thoughts on this additional point @iamthomasbishop, @osullivanchris ?

etoledom:
> I believe that this is a message that could be shown inline as part of a no-results view, providing a much better offline experience.

In most cases, there will be cached images to select even if there is no internet connection, so a no-results view won't be displayed.

I agree that there are better ways than showing the alert. I was thinking about a banner at the top or using the bottom notices. Wdyt @diegoreymendez ?

Anyway I believe this PR is meant to solve the inability to choose an image without connection (when there are catched) as a small patch, and all these extra improvements we might want to work on them separately, but as @SergioEstevao prefer.

etoledom:
> The first test case, when adding an image from a site with a cached media library, I'm getting stuck here in Gutenberg:

With the latest bundle, the image will refresh and re-load when the connection is back (it might take a bit since the network retry delay increases on each try):

https://cldup.com/FS3IpJ_LKP.mov

SergioEstevao:
@diegoreymendez @etoledom , I added to Gutenberg editor the "fancy" no results view, and removed the explicit alert for connection error like Diego suggested.

I will leave for a future PR the improvement in Gutenberg to show a better placeholder image when no network is available.

diegoreymendez:
> I agree that there are better ways than showing the alert. I was thinking about a banner at the top or using the bottom notices. Wdyt @diegoreymendez ?

I think that for consistency with the rest of the App a no results view would be better in this case.

osullivanchris:
Thanks for the ping @diegoreymendez ! Only circling back now

> I believe that this is a message that could be shown inline as part of a no-results view, providing a much better offline experience. The example below is one that I feel would fit very well in this case:

Looks good to me. I tested this on Android and we are doing close to this. Although I think it should say offline or something rather than telling me I have no media (when I might)
Screenshot_20200207-181454

> I think that for consistency with the rest of the App a no results view would be better in this case.
Seems fine to me. I'd expect an alert if something happened or something changed. But I'm probably already offline, and no event occurred. So displaying it in the page seems fine to me.

Its cool that we are caching them that it does work if I've viewed before though, that's nice :)

PR #13422, opened 2020-02-11
Experiment with workflow names

mokagio:
This PR is here just to see how GitHub renders workflows with different name configurations.


mokagio:
Damn. Another PR on the wrong repo 😞 Sorry for the noise.

Issue #13366, opened 2020-02-04
Migrate from self-hosted Peril to Danger on GitHub Actions


loremattei:
We've been using [Peril](https://github.com/danger/peril) to support PR reviews for a couple of years now and it has become a key part of our process. 

Peril is basically a self-hosted version of [Danger](https://github.com/danger/danger).

Since GitHub released Actions, the development of Peril has stalled and the authors and maintainers have focused on improving Danger and making it work flawlessly with GH Actions.

We currently host Peril on a free Heroku account and with the  projects we've now hooked we're reaching its limits. 

Given the above, it would be interesting for us to investigate the migration from Peril to Danger on GitHub Actions. 
Our rules are Danger's rules and they are already stored in a [shared repository](https://github.com/automattic/peril-settings/), so it should be possible to run mixed environments and to migrate on a phased roll out.

mokagio:
Here's an update on the state of this. We decided to **pause the migration effort, because of time and return of investment considerations**.

Danger, whether the Ruby or the Javascript implementation, works great for PRs. We were able to replicate the centralized nature of the Peril configuration by defining project specific `Dangerfile`s that imported and run the granular checks they needed, see [here](https://github.com/Automattic/danger-settings).

Unfortunately, I wasn't able to replicate Peril's behavior for `issues` and `status` webhooks in the Danger + GitHub Actions setup.

- There is [a bug](https://github.com/danger/danger-js/issues/999) that makes Danger when running on a workflow triggered by an [`issues` event](https://help.github.com/en/actions/reference/events-that-trigger-workflows#issues-event-issues).
- When a GitHub Actions workflow runs as a result of a [`status` event](https://help.github.com/en/actions/reference/events-that-trigger-workflows#status-event-status), there is no PR attached to it, and [Danger only expose its GitHub API interface](https://github.com/danger/danger-js/issues/1048). We'd have to add extra code to manipulate the status value and post on the PR like [we currently do with Peril](https://github.com/Automattic/peril-settings/blob/23332976e0e0eb59cd25b2aabc946e4eee7ae234/org/pr/installable-build.ts).

We _could_ adopt Danger only for PR checks and keep Peril for the other ones, but this would mean maintaining two setups instead of one, for no real gain. Danger even runs slower than Peril, because of the GitHub Actions CI boot time.

[...]

PR #13423, opened 2020-02-12
Set up GitHub workflows to run basic checks and validate PR labels via Danger

mokagio:
This is the first step to implement https://github.com/wordpress-mobile/WordPress-iOS/issues/13366.

This adds a GitHub workflows that run the [shared Peril settings](https://github.com/Automattic/peril-settings) to:
-  [validate the PR's labels](https://github.com/Automattic/peril-settings/blob/f8b40cef2adae13967ddd55c3fd917c3b0f200b6/org/pr/labels.ts)
- [preform consistency checks on the iOS code](https://github.com/Automattic/peril-settings/blob/f8b40cef2adae13967ddd55c3fd917c3b0f200b6/org/pr/ios-macos.ts)

screencapture-github-mokagio-WordPress-iOS-pull-6-2020-02-11-14_42_08

This setup uses the [Danger GitHub Action](https://github.com/marketplace/actions/danger-js-action). I like it because we don't have to add anything to our repo in order to run Danger, but it comes with a time overhead on CI because the action needs to be fetched and built. 

Using the action is fine for now, I think, but we might want to do a performance comparison with a local Danger setup later on. It might be worth takin on the cost of maintaining a local setup if we can get a noticeably faster feedback loop for those opening the PR.

The code here is a squash of the work done on https://github.com/mokagio/WordPress-iOS/pull/6. To test how it behaves, visit that PR and:

- Add/remove labels
- Push commits that would hit one of the Danger rules

There's more info in that PR, as well as screenshots of the tests that I run.

### DIFF
--- a/.github/workflows/danger-pr-labels.yml
+++ b/.github/workflows/danger-pr-labels.yml
@@ -1,0 +1,18 @@
+name: Validated PR Labels
+on:
+  pull_request:
+    # The restrictions on the types of pull_request event that can trigger this
+    # workflow is intentional. We only want to run label related checks on a
+    # new or reopened PR, or when its labels change.
+    types: [opened, reopened, labeled, unlabeled]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate PR Labels
+        uses: danger/danger-js@9.1.8
+        with:
+          args: "--dangerfile Automattic/peril-settings/org/pr/label.ts"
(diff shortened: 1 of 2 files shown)


mokagio:
You may notice that neither of the workflows I add in this diff run. I'm pretty sure that's because "only repository collaborators may create workflows".

Screen Shot 2020-02-12 at 11 25 59

PR #13424, opened 2020-02-12
Better error-message logic for post uploads (for page reusability).

diegoreymendez:
This PR is the first step towards leveraging our Offline Post messages for use in Offline Pages.

Changes:

- Relocated offline post messaging (so that the code is cleaner and it's easier to reuse our existing logic for pages).
- Implemented offline page messaging.  This is still not wired / no need to test this.
- Restructures some unit tests so we will be able to extend them for pages easily.
- Reorganizes some unit tests to add cases we weren't testing in offline posting.

### To test:

While offline, try a combination of saving a draft, publishing a post, private post, scheduling, submitting for review.

This should be mostly covered by the proposed unit tests, so make sure to dedicate some time to reviewing the unit tests.

### PR submission checklist:


diegoreymendez:
@leandroalonso - No need to do a full review... but can I get your opinion on the approach used in this PR?

Don't worry though... I'm still planning to add quite a few unit tests before submitting this PR, plus test it manually.

The idea behind this PR is to be able to simplify how we get the messaging for both pages and posts using mostly the same logic.

This PR also tends to group most of the message-selection logic inside `PostAutoUploadMessages`.

leandroalonso:
@diegoreymendez Sure! I'll take a look!

WordPress/Classes/ViewRelated/Post/PageAutoUploadMessageProvider.swift
```diff
@@ -0,0 +1,73 @@
+import Foundation
+
+class PageAutoUploadMessageProvider: AutoUploadMessageProvider {
+
+    // MARK: - Cancelling
+
+    static let changesWillNotBePublished = NSLocalizedString("We won't publish these changes.",
```
leandroalonso: I liked this approach a lot!

One small suggestion, some of the messages are very similar. Eg.: _"We won't publish these changes."_

I wonder if we should just extend the `AutoUploadMessageProvider` and provide that as a default. This will drastically reduce the size of `PageAutoUploadMessageProvider`  and `PostAutoUploadMessageProvider`.

What do you think?

diegoreymendez: The thing is that the messages are the same only coincidentally.  I think this is one of those cases where redundancy makes sense.

More to the point on why I think this is good redundancy: if the message changed to be different across posts and pages ("We won't publish the changes in this post.") we'd be forced to modify the class structure instead of just changing the string.

diegoreymendez: Additionally, although unlikely, this gives the opportunity for translations to use a specific wording.

leandroalonso: Hmm, I agree.

WordPress/WordPressTest/ViewRelated/Post/Utils/PostNoticeViewModelTests.swift
```diff
@@ -24,114 +64,166 @@ class PostNoticeViewModelTests: XCTestCase {
...
-    func testNoticesToBeShownAfterFailingToUploadPosts() {
-        struct Expectation {
-            let scenario: String
-            let post: Post
-            let isInternetReachable: Bool
-            let title: String
-            let actionTitle: String
+    // MARK: - Expectations
+
+    private func verify(_ scenario: Scenario) -> Nimble.ToSucceedResult {
```
leandroalonso: I had a veeery long discussion with @shiki in the past about Parameterized Testing here.

My argument at the time was that I preferred a lot of unit tests/code (aka `func`) but easy to read versus saving lines of code in the favor of abstraction — which IMHO makes it harder to read.

Now I'm more worried because we have more lines of code just to do the asserts and different entities. 😬 Someone in the future will probably be worried by the functions and entities in this test, it's not as simple as just reading a single test.

Anyway, I don't think we need to refactor the entire thing. But keep that in mind.

I prefer clear tests with more lines of code than abstracting it (which I believe shouldn't be in unit tests at all).

diegoreymendez: Just for context, we discussed this and agreed to move ahead with the merge.

For extra context: I do agree that it feels like we have a framework just for these tests, and it would be better to simplify a bit.  That said, the main reason I introduced these changes was so that I don't have to add the page tests into the single method we had for testing.

shiki: Oof. Yeah, the `Expectation` was an experiment, really. I had planned for the design to grow to be reusable as a parameterized test framework within the unit tests. Obviously, it wasn't reusable in `PostNoticeViewModelTests` since the other goal was to make it isolated. 😄 

But I'd agree that it probably wasn't the best way to do it. Perhaps discussions should have happened first. 🤷‍♂ 

Parameterized tests are great for testing a lot of similar things in one go. I'd agree that to a point, it does get a little bit tricky to read and understand. On the other hand, having individual methods for all of these would also make it difficult to identify that they're all related and testing just one path. 

I don't think @leandroalonso and I ever agreed on this. The agreement wasn't that important when you consider the overarching goal during the time this was created → finishing Offline Posting. At that time, any test was better than no test! 😄 

I'm all for killing this code if it comes down to it. My feelings will not be hurt. 😄

Issue #13319, opened 2020-01-29
Information Architecture: ‘Me’ menu move


bjtitus:
Design: https://wp.me/pauD4L-nx-p2

![ios-my-site-blueprint](https://user-images.githubusercontent.com/3250/73475610-f0388480-434d-11ea-9aac-10d449268f78.png)

Moves the ‘Me’ menu to the ‘My Site‘ screen.

- [x] Add feature flag.
- [x] Add the ‘Me‘ icon/gravatar to ‘My Site‘ nav bar
- [x] Add action to modally present (as `pageSheet`) the ‘Me‘ page
  - [x] This action needs to be called from the ‘Me‘ icon, as well as from any other part of the app that requires the ‘Me‘ page
    - [x] `WpTabBarController.showMeTab` callers: 
      - [x] Update `MeNavigationAction` for universal links (`Routes+Me.swift`)
      - [x] Update `openAppSettingsScreen()`, `openMeTab() `and `openSupportScreen()` in `SearchManager.swift`
      - [x] Update `handleSupportNotification` in `PushNotificationManager.swift`
- [x] Add the `Me` button to the `No Sites` screen
- [x] Remove the ‘Me‘ menu from the tab bar (`WpTabBarController.setViewControllers`)
  - [x] Ensure that re-arranging tabs does not require changing indexes in the app (and change them if it does)
- [x] Ensure views are correct for Dark Mode.
- [x] Add Accessibility support.
- [x] Add Tracks.
- [x] Add unit tests.
- [x] Enable feature flag for internal.
- [ ] Enable feature flag for all.
- [ ] Remove feature flag.

ScoutHarris:
@Gio2018 @bjtitus - I don't see it explicitly addressed here, so I just wanted to bring up the scenario where an account has no sites to be sure it's addressed. Since there is no site details, that Me button can't be displayed. On Android, the Me button in this scenario is displayed in the nav bar.

![android_no_sites](https://user-images.githubusercontent.com/1816888/74680529-23687980-517e-11ea-8faf-36443e1bbc05.png)

Gio2018:
> @Gio2018 @bjtitus - I don't see it explicitly listed here, so I just wanted to bring up the scenario where an account has no sites to be sure it's addressed. Since there is no site details, that Me button can't be displayed. On Android, the Me button in this scenario is displayed in the nav bar.
> 
[...]

PR #13445, opened 2020-02-12
Add Me Menu button to My Site Navigation Bar

Gio2018:
Ref #13319

Tapping the `Me` button added to the navigation bar, in `My Site` will present the `Me` scene modally
#### Change log
- Add feature flag for `Me Move`
- Add `ScenPresenter` and (concrete implementation) `MeScenePresenter`: responsible of presenting the `Me` scene, injected into `BlogDetailsViewController`
- Add new initializer for `BlogDetailsViewController` (with the above mentioned dependency injection)
- Add extension to `BlogDetailsViewController containing methods to make the new `Me` button
- Since I was working on the `Classes->ViewRelated->Blog` group, I took the occasion to reorganize files and folders in there

#### Notes

1. Known issue: The `Me` image size is not right yet
2. Known issue: when rotating to landscape, the top bars are misaligned in the split view
3. This PR does not remove the original `Me` scene in the tab bar (yet)
4. This PR does not include the gravatar in the `Me` button
5. This PR does not include unit tests

To test:
- Build/run
- Choose a site of your preference and go to `My Site` 
- Check that the `Me` button is visible
- Tap the `Me` button and verify that the `Me` scene is presented modally and works as expected.

PR submission checklist:

  - yet to be added
  - added localized strings where needed, TBD other changes.
  - Too early for that.


frosty:
@ScoutHarris Thanks for raising the iPad / split view Done button issue! I think that's something that we might've missed with design.

I'd be fine with us moving the Done button to the left hand side (I think it looks better than where it is currently with a split view) to get this merged, and we can investigate moving it to the right hand side of the detail pane in a separate PR.

Gio2018:
@ScoutHarris , @frosty , thanks for the feedback on the `Done` button! I agree that it looks kinda strange where it is now, and it can certainly be moved to the left hand side. I made some preliminary investigation on this, hoping to give some more context. Here are some findings:

1. `UISplitViewController` has no significant appearance by itself (https://developer.apple.com/documentation/uikit/uisplitviewcontroller) this means that we probably cannot add a `navigationItem` to it
2. We could include it in a container view, but this would probably look odd, and also, not recommended by Apple (still from https://developer.apple.com/documentation/uikit/uisplitviewcontroller)
3. I did some research on what apps do in similar situations, starting from WordPress-iOS: we have similar behavior in `Notifications`, where we have the settings button on the right, and in landscape or iPad, it stays on the right of the primary `UIViewController`. Same thing does the `Apple Mail` app, where the `Edit` button behaves as our `Done` button.
4. Should we decide to maintain the `Done` button where it is, a few things might help:
    1. Make the difference between the title and the button more prominent (e.g., using an icon like in `Notifications`, using a different color like in `Apple Mail` (not sure this is applicable as it likely conflicts with our style guide)
    2. Part of the problem could be that the title and `Done` are too close: this can easily be addressed choosing the `Default` width for the primary view instead of `Narrow`. We actually do that in `Notifications`
5. If we were to add the `Done` button to the far right, it would likely need to be added as a `navigationItem` to all the `UIViewControllers` in the navigation stack. 
    1. If one of them already has a right navigation item, the coexistence would need to be handled (this is not the case on the `Me` scene as of now, but needs to be taken into account)
6. the `Done` button can be moved to the left. In this case we should probably still use the `Default` width instead of the `Narrow`.

Please let me know your thoughts, I personally think that we could go either with 4.ii or 6 for now and then have a more detailed discussion with @osullivanchris about this.

ScoutHarris:
> I'd be fine with us moving the Done button to the left hand side (I think it looks better than where it is currently with a split view) to get this merged, and we can investigate moving it to the right hand side of the detail pane in a separate PR.

I'm ok with merging this PR as-is and addressing the Done button in a follow-up PR.

frosty:
Thanks @Gio2018, there are definitely complexities around adding it to the right hand side – does it only show at the top level, or does it also show if you navigate down into sublevels, for example? I think for now, let's move it to the left and we can discuss with Chris next week?

Gio2018:
> Thanks @Gio2018, there are definitely complexities around adding it to the right hand side – does it only show at the top level, or does it also show if you navigate down into sublevels, for example? I think for now, let's move it to the left and we can discuss with Chris next week?

Totally agree. Moved to the left for now.

ScoutHarris:
Well now the button in on the left for _all_ views. I'm not sure this was the desired result.

Screen Shot 2020-02-13 at 10 32 44 AM

Gio2018:
> Well now the button in on the left for _all_ views. I'm not sure this was the desired result.
> 
> Screen Shot 2020-02-13 at 10 32 44 AM



> Well now the button in on the left for _all_ views. I'm not sure this was the desired result.
> 
> Screen Shot 2020-02-13 at 10 32 44 AM

I thought about using something like positioning the `Done` button on the right or on the left depending on the horizontal size class (and capture the change in `traitCollectionDidChange`), but then would this be a little confusing for the users? If you'd agree, this could also be part of the discussion with @osullivanchris

ScoutHarris:
> If you'd agree, this could also be part of the discussion with @osullivanchris

Sure.

osullivanchris:
Just catching up on this :) Thanks for the pings and investigating so many options @Gio2018 super work!

I keep getting tripped up on minor iPad details. Mainly because I don't have an iPad so I can't look at these conventions :D So I'm ordering one today!

Although visually I agree that 'done' doesn't look correct in the split view, logically it is correct in that 'done' collapses the whole modal. I think if it was part of the right hand side it would imply it only relates to the deeper view.

Moving it to the left seems to resolve the issues that occured in the quickest manner. Happy to continue in a follow up task. My concerns with this solution:
- I used the right hand side for 'done' to reserve the left hand side for going back. In that way, if the user is 5 levels deep and want to go back, they can tap back to just go up one level. Or use 'done' to collapse the whole stack. Both of these options I believe are still available on iPad, but now that we have 'done' on the left - what happens when I go a level deeper on iPhone? Is 'done' replaced with 'back'?
- Related... the behaviour I guess is now different with notifications. I would say we should solve it the same way for both. Noting that on iPhone you can see the 'done' and 'back' working in combination if you go to notifications > settings > choose any list item
- I don't understand 4ii or what the difference would be. But I'm interested in understanding and discussing that further.

Thanks for highlighting - good quick fix but lets keep discussing to make sure we have a tight solution going forward that we can use for all cases.

Gio2018:
Thanks @osullivanchris for the feedback! Here are a few additional considerations:

> what happens when I go a level deeper on iPhone? Is 'done' replaced with 'back'?
The navigation would still work the same as before: in our split views, the navigation typically happens on the right, and it would not change with this implementation

##### new navigation
new_navigation

##### old navigation
old_navigation

> the behaviour I guess is now different with notifications. I would say we should solve it the same way for both. Noting that on iPhone you can see the 'done' and 'back' working in combination if you go to notifications > settings > choose any list item

Good point: in Notification -> Settings we don't use a `UISplitViewController`: the advantage here Is a cleaner ui, but to dismiss the view in landscape or on the iPad (where is displayed on top of the previous view but not full screen) we have to go back to the top level and then hit Done, instead of just dismissing with Done wherever we are in the navigation level. I agree that it would be better if we have consistent experience throughout the app.

> I don't understand 4ii or what the difference would be. But I'm interested in understanding and discussing that further.

I guess it's just a tiny visual difference: if you make the first column larger and put the `Done` to the right of that column (not to the far right then) it doesn't look so close to the title. It's probably not that effective anyway

osullivanchris:
Questioning from first principles, what do you think about just killing the 'Done' action, @Gio2018 ?

Looking at what @frosty raised on https://wp.me/pbArwn-7o, we may not need it all that much

- On iPhone, a user going back through the stack can just tap back several times as they would have done before it was a modal presentation. But as its a modal, if they really want to skip that, they can pull down to dismiss the whole stack, I believe?
- on iPad, if we go with a sheet style presentation, its a very common convention to tap the skrim/overlay area to dismiss the sheet. 

Knowing all that, I don't know if adding the 'Done' button is so necessary now.

Gio2018:
> Questioning from first principles, what do you think about just killing the 'Done' action, @Gio2018 ?
> 
> Looking at what @frosty raised on https://wp.me/pbArwn-7o-p2, we may not need it all that much
> 
> * On iPhone, a user going back through the stack can just tap back several times as they would have done before it was a modal presentation. But as its a modal, if they really want to skip that, they can pull down to dismiss the whole stack, I believe?
> * on iPad, if we go with a sheet style presentation, its a very common convention to tap the skrim/overlay area to dismiss the sheet.
> 
> Knowing all that, I don't know if adding the 'Done' button is so necessary now.
> 
> Edit: was reading that p2 when I replied. But looking again at the videos above, I have addressed iPad and iPhone portrait but not iPhone landscape. Perhaps there does need to be a way to solve this on iPhone landscape.
> 
> So my question for iPhone landscape is:
> do we use split view for anything other than top level tabs? My phone is not big enough to test. I'm thinking that as 'me' is no longer a top level tab, perhaps it should no longer have a split view. Not sure if that logic holds. But if it wasn't a split view it would just fall into a normal 'back' step which would always eventually return to the 'My Sites' screen

I think that if we go with the Option 3 discussed here https://wp.me/pbArwn-7o-p2, we can keep the `Done` button and do something similar at what we do for the notification settings. That way, done would give users another opportunity to dismiss. Basically we could do exactly what the notification settings are doing today. What do you think @osullivanchris ?

osullivanchris:
Yeah that makes sense to me, it lines up with what I was saying in a round about way :) Good to have consistency with notifications settings as well. Thanks for resolving @Gio2018

Issue #12916, opened 2019-11-10
Display interstitial after login if no sites


yaelirub:

TahminaMily:
the app will prompt the interstitial screen only once for the first time login and remember it for future login

rachelmcr:
Added in https://github.com/wordpress-mobile/WordPress-iOS/pull/13523.

PR #13459, opened 2020-02-13
Adds logic to determine when to show the post signup interstitial

### LINKED ISSUE #12877: Display new interstitial after signup
Display the screen we created in #12875 after sign up

emilylaguna:
Fixes #12877 and #12916

To test:
1. Launch the app
2. If you're logged in, log out
3. Login to an account that has no sites
3.a Sign up for a new account
4. You will be shown the epilogue views
5. Tap next
6. You should be displayed the interstitial
7. Relaunch the app, you should not be displayed the interstitial

PR submission checklist:


WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
```diff
@@ -210,12 +211,37 @@ - (void)updateSearchVisibility
 
 - (void)maybeShowNUX
```
leandroalonso: Is this method name correct? NUX is a concept or a specific VC?

WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
```diff
@@ -210,12 +211,37 @@ - (void)updateSearchVisibility
...
+    BOOL shouldDisplayPSI = [PostSignUpInterstitialViewController shouldDisplayWithNumberOfBlogs:blogCount];
+
+    if (!shouldDisplayPSI) {
+        return;
     }
+
+    [self showPostSignUpInterstitial];
+}
+
+#warning TODO: Change Name
```
leandroalonso: Shall we change the name and remove the `#warning`? :P

WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
```diff
@@ -210,12 +211,37 @@ - (void)updateSearchVisibility
...
+- (void)showPostSignUpInterstitial
+{
+    #warning TODO: Add logic to only display this once
```
leandroalonso: I guess this was already done and we can remove this line,

WordPress/Classes/ViewRelated/NUX/Post Signup Interstitial/PostSignUpInterstitialViewController.swift
```diff
@@ -46,6 +49,9 @@ class PostSignUpInterstitialViewController: UIViewController {
         view.backgroundColor = .listBackground
 
         configureI18N()
+
+        //Mark it as seen
+        PostSignUpInterstitialViewController.markAsSeen()
```
leandroalonso: This still reads very strangely to me.

If we aim to decouple the seen/not seen logic from the ViewController we can extract it to another entity. This entity might even be encapsulated inside this VC if you want to  (and it can be a struct, since there's no state at all).

I think my discomfort here comes from two things:

- `static`/`class` being used for the purpose of decoupling (which I believe it's not what static/class is intended for), you can see here that the static name has this class name, so they're related :P 
- It's very procedural
- We're breaking the single responsibility principle: the VC should only present data and here we're also handling a database / saved state.

This VC is quite simple, so I don't think it's a huge problem if we keep this at it is. Just my 2 cents for you to think about. :)

WordPress/Classes/ViewRelated/NUX/Post Signup Interstitial/PostSignUpInterstitialViewController.swift
```diff
@@ -78,3 +84,59 @@ class PostSignUpInterstitialViewController: UIViewController {
...
+    class var userDefaultsKey: String? {
+        get {
+            guard
+                let account = defaultWPComAccount(),
+                let userId = account.userID
+            else {
+                return nil
+            }
+
+            return String.init(format: Constants.userDefaultsKeyFormat, userId)
```
leandroalonso: We can remove the `init`

WordPress/Classes/ViewRelated/NUX/Post Signup Interstitial/PostSignUpInterstitialViewController.swift
```diff
@@ -78,3 +84,59 @@ class PostSignUpInterstitialViewController: UIViewController {
...
+extension PostSignUpInterstitialViewController {
+    /// Determines whether or not the PSI should be displayed for the logged in user
+    /// - Parameters:
+    ///   - numberOfBlogs: The number of blogs the account has
+    @objc class func shouldDisplay(numberOfBlogs: Int) -> Bool {
+        if !AccountHelper.isLoggedIn || hasSeenBefore() {
+            return false
+        }
+
+        return (numberOfBlogs == 0)
```
leandroalonso: No need for the parenthesis.

WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
emilylaguna: NUX is a concept of the "New User Experience", the more up to date naming is now onboarding if I remember correctly. 

I'm not sure I want to rename this method in this PR since I feel it's a bit out of scope of what I'm doing. Thoughts?

leandroalonso: I misunderstood then. Let's keep it.

WordPress/Classes/ViewRelated/Blog/BlogListViewController.m
emilylaguna: 👍I've just removed the warning, looking at the other code in the class using the word "show" is the correct terminology

WordPress/Classes/ViewRelated/NUX/Post Signup Interstitial/PostSignUpInterstitialViewController.swift
emilylaguna: After taking another look at this, this morning I agree. I'll take another pass at my implementation.

PR #13460, opened 2020-02-13
Leverage PostAutoUploadInteractor support for offline pages.

diegoreymendez:
This PR leverages the code in `PostAutoUploadInteractor` to also support pages.  This isn't wired though, since the interactor is used by the post coordinator, which doesn't support pages yet.

Additionally: this PR implements several unit tests for page messaging.  These tests required the interactor to work with pages, so they sort of validate the changes mentioned above.

### To test:

Run the unit tests and make sure that they succeed. Double check their definition.

### PR submission checklist:


diegoreymendez:
This PR is ready for review.

WordPress/Classes/Services/PostAutoUploadInteractor.swift
```diff
@@ -46,7 +46,6 @@ final class PostAutoUploadInteractor {
         guard post.isFailed,
             let status = post.status,
             !PostAutoUploadInteractor.disallowedStatuses.contains(status),
-            !(post is Page),
```
leandroalonso: This will put `develop` in an inconsistent state. With this remotion, I'm able to tap "Publish" in a Page and the apps tell me it will upload my draft when I'm back online.

When I go back online, the page is uploaded as a draft.

I wonder if we should put this under a FeatureFlag.

Thoughts?

Besides that, everything looks ok ✅

diegoreymendez: Actually that's unintended.  I will feature flag this check.

Good catch @leandroalonso !

diegoreymendez: This was resolved by:

1. Adding the feature flag.
2. Retargetting this PR to the feature branch so that it won't disrupt the work of other devs.

diegoreymendez:
@leandroalonso -  Ready for another round!

diegoreymendez:
Thanks @leandroalonso !

PR #13475, opened 2020-02-14
Add gravatar to the Me button

Gio2018:
#### Download the gravatar and add it to the new `Me` button in `My Site`. If gravatar is unavailable, fall back to the default `Me` icon (circled user).

Ref #13319

##### Change log
- Update BlogDetailsViewController+MeBarButton.swift, add UIBarButtonItem extension that downloads the gravatar and sets it to the custom view of the button
- Update CircularImageView.swift, add an optional tap animation to animate the tap on the image

##### Notes

1. The gravatar size is set to the default `GridIcon` size (24 x 24). TBD with @osullivanchris if this needs to be changed.
2. Known issue: when rotating to landscape, the top bars are misaligned in the split view
3. This PR does not remove the original `Me` scene in the tab bar (yet)
4. This PR does not include unit tests

To test:
1. Build/run and login with an account that has a gravatar image
2. Choose a site of your preference and go to `My Site`
3. Verify that the `Me` button contains the gravatar.
4. Tap the `Me` button and verify that the `Me` scene is presented modally and works as expected.
5. Repeat steps 1 and 2 with an account that has no gravatar, and verify that the default image (circled user) is displayed on the `Me` button in `My Site`.

PR submission checklist:


WordPress/Classes/ViewRelated/Views/CircularImageView.swift
```diff
@@ -34,4 +37,11 @@ class CircularImageView: UIImageView {
             layer.masksToBounds = shouldRoundCorners
         }
     }
+    /// Add the custom animation on tap.
+    override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        guard let animation = tapAnimation else {
```
bjtitus: Should `super` be called here? I'm not sure that we want to intentionally disable any handling UIImageView may do with touches (although it _probably_ doesn't do anything).

Gio2018: 👍

Gio2018:
> I wonder if the animation should follow the back button a little more closely: dimming to 0.5 opacity, staying there while the touch is still down, and then going back to 1 upon the touch ending.
> 
> That would probably require implementing a few more touch handling methods, which then might make the UITapGesture recognizer unnecessary. Not sure if any of that is worth doing at the moment but just thought I'd bring up the slight animation difference.

We could. In most of the cases, if a user does not keep the finger on the button, the animation would look the same, but if they do, there would be a difference.

Ideally I would still use the closure in `CircularImageView` for things that are related to the view, and keep the action of presenting the `Me` view in a separate object, but not sure how complex the code would become.
+@osullivanchris

Gio2018:
> Hey @Gio2018 . I noticed an issue. On a site with no gravatar:
> 
> * The default image showed as expected.
> * I navigated to Me > Update Profile Photo and selected one.
> * I expected the Me button to update, but it did not.

Yup. Looks like both `MeHeaderView.m` and `MyProfileHeaderView.swift` were missing a call to `updateGravatar`, which basically posts a notification that triggers the gravatar update on all the gravatars.

To this point, in both places where I add the call to `updateGravatar`, I found this note in the comments:

>         // Note:
>         // We need to update the image internal cache. Otherwise, any upcoming query to refresh the gravatar
>         // might return the cached (outdated) image, and the UI will end up in an inconsistent state.
>         //

Not sure what that means, but it might be worth investigating, I will do some more investigation and post something about it

osullivanchris:
@Gio2018 this sounds great. Confirming I'd expect the gravatar to be the same size for now. We might consider a different header & gravatar layout in future. Would love to see a screenshot with both the gravatar and with the fallback icon if you get a chance!

PR #13442, opened 2020-02-12
Fix 13436: Preview Publish dismiss Editor w/ Confirm

bjtitus:
Fixes #13436 by dismissing the editor after publishing and adding a confirmation dialog.
Fixes #13438 by taking the user straight to the post list so that post settings are not as easy to change while publishing.

It seems like there are a couple of options to fix the issue:
1. Dismiss both the preview and the editor and go back to the Post List. (Equivalent to using Publish Now)
2. Go back to the Editor with the post fully published and everything updated.
3. Go back to the Editor with the status changed to Published so that if you “Cancel” instead of hitting the additional Publish button in the navigation bar, the Publish would be cancelled. (equivalent of changing the Post's Status to Published in Post Settings)

The first option has the benefit of being a small code change and completely avoiding potential questions about how to handle cancelling out of the Editor or waiting on the Publish action to complete. I checked with @osullivanchris on this and we think dismissing both views is fine for this case.

| Before | After |
|--|--|
| <a href="https://user-images.githubusercontent.com/3250/74374527-d3636e80-4d9b-11ea-825a-d1937a766a31.gif"></a> | <a href="https://user-images.githubusercontent.com/3250/74373935-a9f61300-4d9a-11ea-8bc7-a230e08d3a76.gif"></a> |

To test:

- Open a post preview via the editor
- Press the Publish button in the bottom left
- Notice new confirmation dialog
- Editor should dismiss along with the preview upon publishing.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ Not yet released

PR #13476, opened 2020-02-17
Install Danger locally for better performances on CI

mokagio:
https://github.com/wordpress-mobile/WordPress-iOS/pull/13423 added two [GitHub workflows](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions) to run static checks on PRs via [Danger](https://danger.systems/js/).

To keep the setup simple, I decided to use the [Danger GitHub Action](https://github.com/marketplace/actions/danger-js-action), but the time it takes to run on CI is too much, more than a minute.

This diff switches the setup to a local installation of Danger, which runs on CI in less than 30 seconds –after the first run, when the cache has been build.

Before | After | After, with cache
--- | --- | ---
Screen Shot 2020-02-17 at 11 17 21 | Screen Shot 2020-02-17 at 11 15 26 | Screen Shot 2020-02-17 at 11 18 37

This approach adds a bit of overhead on the repo: another dependency manager. Devs don't need to interact with it though, only CI does. I think it's an acceptable tradeoff to deliver feedback on the PR faster.

### How to test

Because this Danger setup runs on [pull request events](https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-event-pull_request), you're better off testing this changes in a new PR. It will keep this PR cleaner. Feel free to use https://github.com/mokagio/WordPress-iOS/pull/9 as a base for it, the code is the same, just less tidy.

The test strategy should be the same as #13442:

- Add and remove labels to verify Danger leaves comments when the PR has no label.
- Change the codebase in a way that will trigger a warning or failure in the consistency checks.


.github/workflows/danger-pr-labels.yml
```diff
@@ -10,9 +10,32 @@ jobs:
...
         with:
-          args: "--dangerfile Automattic/peril-settings/org/pr/label.ts"
+          path: node_modules
+          key: ${{ runner.os }}-node_modules-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-node_modules
+
+      - name: Yarn Install
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
+        # frozen-lockfile will make the build fail if the lockfile is not there
+        run: yarn install --frozen-lockfile
```
mokagio: Only the output of `yarn install` is cached. I thought about making the `npm` command to install Yarn be a local one, to cache that output as well, but GitHub has already optimized `npm` runs and put in place their own caching for package used as much as Yarn.

Screen Shot 2020-02-17 at 11 03 38

.github/workflows/danger-prs.yml
```diff
@@ -2,17 +2,40 @@ name: iOS Consistency Checks
 on:
   pull_request:
     # The restrictions on the types of pull_request event that can trigger this
-    # workflow is intentional. We only want to run these sanity checks when the
```
mokagio: More inclusive language. Thanks again @jkmassel for pointing this out.

package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "devDependencies": {
+    "danger": "^9.2.10",
+    "typescript": "^3.7.5"
```
mokagio: Notice that, on top of Danger, we have TypeScript as well. That's because [our `Dangerfile`s are written in TypeScript](https://github.com/Automattic/peril-settings/tree/5d22258ca46c3649f2cb863cfac2d16f12c4cba0), so Danger needs the TypeScript transpiler to be in the `node_modules/` folder in order to process them.

.github/workflows/danger-prs.yml
jkmassel: Nice catch! 👍

.github/workflows/danger-pr-labels.yml
jkmassel: Given the fact that this is 1 second of the CI time, this seems like a sensible decision!

.github/workflows/danger-pr-labels.yml
```diff
@@ -10,9 +10,32 @@ jobs:
...
+
+      - name: Validate PR Labels
```
jkmassel: This differs from the job name which is "Validated PR Labels". IMHO saying "Validate" is better as it's a present-tense verb. Can we use it everywhere? :)

jkmassel:
Note for future – this PR was tested using https://github.com/wordpress-mobile/WordPress-iOS/pull/13494.

.github/workflows/danger-pr-labels.yml
mokagio: Great suggestion. I do like the present-tense verb approach ❤️ . I'm onto it. 👍 .

mokagio: Oh :o. Just seen that past-tense "Validate_d_" in the code. Not intentional, just a typo. Good catch 😄

mokagio:
> One potential area for improvement for you to consider (and sorry if you've addressed this elsewhere) – while it's nice to run these in parallel, a lot of the overhead work (checking out the repo, setting up node, etc) is repeated for each job. **Running all the checks in the same job** would slightly increase run time, but given the degree to which you've optimized things, would that be worth it? Part of the benefit here would be grouping feedback from all of the actions into the same GitHub comment (in the same way that Peril does now). WDYT about this? 
(_emphasis mine_)

Thanks @jkmassel, that's a great suggestion. 

**TL;DR:** 👉 I like the idea and will try it out in a dedicated PR.

---

I ended up with this setup because I let the events that might trigger a workflow build drive which Danger file to run.

We run the check for the labels only when the labels can change:

https://github.com/wordpress-mobile/WordPress-iOS/blob/ca1f915437ffe8ffe8a7fe250b2bebfd07f621e2/.github/workflows/danger-pr-labels.yml#L2-L7

In the same fashion, we run the check on the code only when the code can change, not when the labels change:

https://github.com/wordpress-mobile/WordPress-iOS/blob/ca1f915437ffe8ffe8a7fe250b2bebfd07f621e2/.github/workflows/danger-prs.yml#L2-L7

I think this is less wasteful, at least from a certain point of view. Here's a PR life-cycle example

Event | Danger runs
--- | ---
open | labels validation, consistency checks
add label | labels validation
push new commits | consistency checks
push new commits | consistency checks
reviewed with change request | none
push new commits to address feedback | consistency checks
reviewed with approval | none
merge | none

In such a case, I think we'd save time by not running the label validation on every change to the PR.

But ☝️ the point you raise is totally legit. Actually, **given that the time it takes to run Danger is very little compared to the rest of the workflow setup and that those multiple comments are very noisy, I think the trade-off of running Danger checks that will produce the same result because nothing has changed is worth ends up being worth it**. If anything, it's definitely less usage on CI, not that we pay for it, but that's not an excuse to waste resources 🔋 🌏 ♻️ .

There's only one little thing I'm not sure about. I don't know if I can tell Danger to run more than one `Dangerfile` with the `--dangerfile` option. I'm going to look into it, then submit a dedicated PR.

mokagio:
@jkmassel yeah. I don't have merge rights though, so you'll have to merge this for me :pray:

PR #13498, opened 2020-02-19
Add navigation to the new Me Scene from Spotlight Search and Universal links.

Gio2018:
#### Replaces the existing calls to the `Me` tab with the new modal presentation, from Spotlight Search and Universal links. Also adds the Me button to My Sites if no site is available, and removes the `Me` tab from the app.

Ref #13319

#### Change log
- Add `WPTabBarController+MeNavigation.swift`, replace the old calls with new calls to the new `Me` scene
- Update `MeViewController.swift`, add method `refreshDetailView()` that fixes a layout bug when rotating an iPhone from compact to regular
- Update `BlogDetailsViewController+MeBarButton.swift`, add accessibility to the `Me` gravatar button
- Update BlogListViewController+MeBarButton.swift, renamed to UIBarButtonItem+MeBarButton.swift as now contains extension from blog list and blog details
- Update BlogListViewController.h, added scenePresenter property
- Update BlogListViewController.m, added logic in setAddSiteBarButtonItem to add the Me button if no sites are available
- Add unit tests for the `Me` gravatar button in `My Site` (`BlogDetailsViewController+MeButtonTests.swift`)
- Fixed crash in unit tests
- Rename `scenePresenter` to `meScenePresenter` in `BlogLis`t, `BlogDetail` and `WPtabController`
- Improved dependency injection of `meScenePresenter` in `BlogList` and `BlogDetails`
- Update `WPTabBarController.m`, remove the `Me` tab
- As discussed with @bjtitus , update `NewBlogDetailHeaderView.swift`, make `delegate` a `weak var` to prevent leaks
- Update UI tests to reflect the new UI

To test:

Universal links
- With the app installed on the simulator, open terminal and run any of the following commands
  - `xcrun simctl openurl booted https://wordpress.com/me`
  - `xcrun simctl openurl booted https://wordpress.com/me/account`
  - verify that the corresponding screen opens (Me or Account Settings)

Spotlight
- From the home screen, swipe down to show the Spotlight Search bar
- Search for WordPress
- In the search results for the WordPress app, tap `Me` and verify that the new `Me` scene opens
- If `Me` doesn't show, open the app and access the screen a few times so it gets indexed. You can do the same with `Me->App Settings` and all the other sub levels.

No Sites
- Log in with an account that has no sites
- Verify that the `Me` button appears in the top right and works as expected.

Me tab removed
- Ensure that the `Me` tab does not appear in the tab bar
- Tap on the various tabs and verify that they behave as expected
- Trigger the tab navigation remotely, using the commands below and verify that they show the corresponding tab
  - `xcrun simctl openurl booted https://wordpress.com/read` (reader)
  - `xcrun simctl openurl booted https://wordpress.com/notifications` (notifications)

PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.h
```diff
@@ -2,9 +2,12 @@
 
 @class Blog;
 
+@protocol ScenePresenter;
+
 @interface BlogListViewController : UIViewController
 
 @property (nonatomic, strong) Blog *selectedBlog;
+@property (nonatomic, strong) id<ScenePresenter> scenePresenter;
```
bjtitus: This name seems kind of generic. It seems like there are a lot of things that are presented by `BlogList`, `WPTabBarController` and `BlogDetails`. Would it make more sense to specify that this as the "Me" presenter, or would the scene presenter be more generic for other presentations?

Gio2018: Thanks for bringing this up @bjtitus ! It's a very good point.

The type is generic as it was intended to be, but there's a bit of misuse.
- The name of the property is generic, but in reality it's used specifically for the `Me` button
- Plus, there's a bit of duplication and injection is not properly used. A few things can be done to improve it:
  - leave the type generic but name the parameters meScenePresenter as you say (and do this in all the three cases: BlogList, BlogDetails, TabBarController)

  - In Addition:
    - Add a new initializer to BlogListViewController and inject the dependency (as I already did for BlogDetailsViewController)
    - Because the `ScenePresenter` instance is always a `MeScenePresenter` instance for all the three cases (BlogList, BlogDetail and TabBarController), and there would not be more than one presentation at a time, then we can use `WPTabBarController` at the "outer scope" and inject the same instance from there into BlogList and from BlogList into BlogDetails. It's not super ideal cause WPTabBarController is a singleton, but it's an alternative of having an instance of `MeScenePresenter` in each one of them. What do you think?

bjtitus: It's really the just the name I'm wondering about. Since the property is only called from the locations presenting the Me screen, it seems like it could be more specific.

As far as the injection stuff goes:

> Add a new initializer to BlogListViewController and inject the dependency

I think this might make sense as a follow up to clean this controller up. It might be better as a separate issue, though, so it isn't intertwined with the iA work.

> we can use WPTabBarController as the "outer scope" and inject the same instance from there into BlogList and from BlogList into BlogDetails

This could make sense. Any issue to remove the WPTabBarController singleton would probably be a good place to do that.

Gio2018:
> Hey @Gio2018 . This looks good. The only thing I noticed is, as CircleCI indicates, `DomainsServiceTests` is failing. They do pass on `develop`. If it helps, I only see them fail when running all tests via `WordPressTest`. Running only `DomainsServiceTests` works fine. 🤷‍♀

I fixed the unit tests, but UI tests are still failing. I ran them on develop (locally) and they seem to fail there too

WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/NewBlogDetailHeaderView.swift
```diff
@@ -1,6 +1,6 @@
 class NewBlogDetailHeaderView: UIView {
 
-    @objc var delegate: BlogDetailHeaderViewDelegate?
+    @objc weak var delegate: BlogDetailHeaderViewDelegate?
```
bjtitus: 🙇

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -52,6 +52,16 @@ - (instancetype)init
     return self;
 }
 
+- (id)initWithMeScenePresenter:(id<ScenePresenter>)meScenePresenter
```
bjtitus: 👏

WordPress/Classes/ViewRelated/System/WPTabBarController.m
```diff
@@ -476,7 +483,12 @@ - (void)showTabForIndex:(NSInteger)tabIndex
 - (NSInteger)adjustedTabIndex:(NSInteger)tabIndex toTabType:(BOOL)toTabType {
     //TODO: Remove this change once `floatingCreateButton` feature flag is enabled
     if ([Feature enabled:FeatureFlagFloatingCreateButton] && tabIndex > WPTabReader) {
-        return tabIndex + (toTabType ? -1 : 1); // Adjust the index if we are hiding the new post button
+        if ([Feature enabled:FeatureFlagMeMove]) {
```
bjtitus: I wonder if this should be separate. It might not be out of the question for one of these features to ship without the other.

```swift
NSInteger tabOffset = 0;
if ([Feature enabled:FeatureFlagFloatingCreateButton] && tabIndex > WPTabReader) {
    tabOffset += 1;
}
if ([Feature enabled:FeatureFlagMeMove] && tabIndex > WPTabNewPost) {
    tabOffset += 1;
}
return tabIndex + (toTabType ? -tabOffset : tabOffset);
```

_I didn't actually test this, it's just an example._

Gio2018: I think they will likely be shipped together but yeah, this reads a lot better. I just tested it and seems to work

WordPress/WordPressUITests/Screens/TabNavComponent.swift
```diff
@@ -15,12 +14,14 @@ class TabNavComponent: BaseScreen {
...
         notificationsTabButton = tabBars.buttons["notificationsTabButton"]
-        meTabButton = tabBars.buttons["meTabButton"]
-        super.init(element: meTabButton)
+        super.init(element: mySitesTabButton)
     }
 
     func gotoMeScreen() -> MeTabScreen {
-        meTabButton.tap()
+        gotoMySitesScreen()
+        app.cells["infocusphotographers.com"].tap()
```
bjtitus: Might be better to use `WPUITestCredentials.testWPcomSitePrimaryAddress` here.

WordPress/WordPressUITests/Screens/TabNavComponent.swift
```diff
@@ -15,12 +14,14 @@ class TabNavComponent: BaseScreen {
...
     }
 
     func gotoMeScreen() -> MeTabScreen {
-        meTabButton.tap()
+        gotoMySitesScreen()
+        app.cells[WPUITestCredentials.testWPcomSitePrimaryAddress].tap()
+        let meButton = app.navigationBars.buttons["meBarButton"]
+        meButton.tap()
         return MeTabScreen()
     }
```
rachelmcr: We should move the method that opens the Me screen out of `TabNavComponent` (which represents the bottom navigation bar in the app) and into `MySiteScreen` (since the button to open `Me` is now an element only on that screen). That way, our page objects only include elements and actions that are on that screen (or, in the case of the bottom nav bar, in that component) and the movement through different screens the app is more transparent in the tests.

That said, as long as the tests are passing like this, I'm ok with that change being done in a followup PR.

WordPress/WordPressUITests/Tests/EditorAztecTests.swift
```diff
@@ -18,6 +19,7 @@ class EditorAztecTests: XCTestCase {
             EditorFlow.returnToMainEditorScreen()
             editorScreen.closeEditor()
         }
+        LoginFlow.logoutIfNeeded()
```
rachelmcr: Was there a problem with the tests that this solves? We originally decided not to log out between each editor test because it shouldn't be needed and the extra login/logout steps slow down the tests.

WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
```diff
@@ -18,6 +19,7 @@ class EditorGutenbergTests: XCTestCase {
             EditorFlow.returnToMainEditorScreen()
             editorScreen.closeEditor()
         }
+        LoginFlow.logoutIfNeeded()
```
rachelmcr: Same as above.

WordPress/WordPressUITests/Screens/TabNavComponent.swift
Gio2018: This makes sense, and I would do another PR (even right after this) so it's specifically focused on UI Test refactoring.

WordPress/WordPressUITests/Tests/EditorAztecTests.swift
Gio2018: I noticed, at least locally, that some tests were failing if the user was already logged in.

WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
Gio2018: same as above

WordPress/WordPressUITests/Tests/EditorAztecTests.swift
rachelmcr: Hm, that's odd! I'm ok with leaving this change, then, and I opened https://github.com/wordpress-mobile/WordPress-iOS/issues/13547 to follow up on this.

WordPress/WordPressUITests/Screens/TabNavComponent.swift
rachelmcr: Perfect, thank you!

PR #13500, opened 2020-02-19
Add README with naming conventions in .github/workflows folder

mokagio:
Codifies @jkmassel naming suggestions from https://github.com/wordpress-mobile/WordPress-iOS/pull/13476#discussion_r380899847 so we can keep following them in the future.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,28 @@
+## Workflows Naming Convention
+
+Whenever possible, follow this naming convention for workflows:
+
+```yaml
+# Use a short descriptive name for the whole workflow
+name: Validate PR Labels
+on: ...
+
+jobs:
+  # Use a present-tense verb as the name for the job, aligned with the workflow name
+  validate:
+    steps:
+      # Omit names for steps using an action
+      - uses: actions/checkout@v2
+
+      ...
+
+      # Keep using short present-tense names for steps that set up the stage
+      - name: Install Yarn
+        run: npm install -g yarn
+
+      ...
+
+      # Use the same present-tense name as the job for its core step, if any
+      - name: Validate
+        run: ...
+```


.github/workflows/README.md
```diff
@@ -0,0 +1,28 @@
+## Workflows Naming Convention
+
+Whenever possible, follow this naming convention for workflows:
+
+```yaml
+# Use a short descriptive name for the whole workflow
+name: Validate PR Labels
```
jkmassel: I'd add that it should work nicely with the job names.

When it's displayed in the list of checks, it'd be nice for it to look something like:

`Validate / PR Labels`
`Validate / Localizations`
...etc

.github/workflows/README.md
```diff
@@ -0,0 +1,28 @@
...
+on: ...
+
+jobs:
+  # Use a present-tense verb as the name for the job, aligned with the workflow name
+  validate:
+    steps:
+      # Omit names for steps using an action
```
jkmassel: Interesting! Why is this one important?

.github/workflows/README.md
mokagio: Nice. Yeah, that would work better for workflows with more than one step that is not setup, like in your example.

In such case then it would also make sense to have an actual `name` for a job. My suggestion is to omit it, but then it would do

`validate / PR Labels`
`validate / Localizations`
...etc

That doesn't look as nice...

.github/workflows/README.md
mokagio: Good question, I should have added my rationale. To me, it seems unnecessary to add a custom `name` for something that already has one. Third party actions already have a name, like `actions/checkout` or `danger/danger-js`.

Having said all that, I can see the case for a custom `name` when the action takes arguments and the value of those arguments is useful to know when scanning the workflow log, where with "workflow log" I mean this

Screen Shot 2020-02-21 at 20 31 02

mokagio:
Hey @jkmassel, thanks for the thoughtful comments.

> This is something that _might_ not be fully set in stone yet

Absolutely. I think that's why it's cool to have a document/PR where people can discuss about the conventions to use.

Something just occurred to me though... I wonder how much sense it makes for a document like this one to live in the WordPress-iOS repo. Seems like something more generic that should be standalone.

Anyways, I might have jumped ahead of myself here. There's also a case for first getting a few successful Actions setups under our belt, and only after take a look at how they're written and draw some guidelines from there. I can close this for now, if it's too much overhead.

mokagio:
Gonna close this, it's stale.

If we'll want to have a standard, it shouldn't live in WordPress but somewhere shared.

PR #13516, opened 2020-02-20
Fix crash in Post List Action Sheet

### LINKED ISSUE #13514: Cannot handle unexpected button for post action sheet: retry. This is a configuration error.
The `PostActionSheet` is not handling the `retry` option: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Post/PostActionSheet.swift#L64

We can refactor this part to avoid a similar crash from happening.

Sentry: https://sentry.io/share/issue/08f91162838940aaa975c800b35b6f3d/

leandroalonso:
Fixes #13514

### To test

I wasn't able to reproduce the case where the "Retry" action appears in the Action Sheet (normally is the 2nd option in the Post Card).

However, I'm sure that his change will fix the crash and the option will appear.

Also, I made the `switch` exhaustive instead of using a `default`.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostActionSheet.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostActionSheet.swift
@@ -65,9 +65,11 @@ class PostActionSheet {
                         self?.interactivePostViewDelegate?.retry(post)
                     }
                 case .edit:
-                    break // Not available in Action Sheet
+                    actionSheetController.addDefaultActionWithTitle(Titles.edit) { [weak self] _ in
+                        self?.interactivePostViewDelegate?.edit(post)
+                    }
                 case .more:
-                    break // Not available in Action Sheet
+                    CrashLogging.logMessage("Cannot handle unexpected button for post action sheet: \(button). This is a configuration error.", level: .error)
                 }
         }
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Post/PostActionSheet.swift
```diff
@@ -60,8 +60,14 @@ class PostActionSheet {
...
                         self?.interactivePostViewDelegate?.cancelAutoUpload(post)
                     }
-                default:
-                    CrashLogging.logMessage("Cannot handle unexpected button for post action sheet: \(button). This is a configuration error.", level: .error)
+                case .retry:
+                    actionSheetController.addDefaultActionWithTitle(Titles.retry) { [weak self] _ in
+                        self?.interactivePostViewDelegate?.retry(post)
+                    }
+                case .edit:
+                    break // Not available in Action Sheet
```
diegoreymendez: This feels like we're failing to decide if we support an "edit" action or not.

To put it in a different light, I think that here we are either:

- Silently ignoring an error for an invalid scenario, or;
- Missing proper support for a valid scenario.

My improvement suggestion here would be to either:

1. Implement the logic for the edit action and let the UI decide whether it's available or not, or;
2. Recording an error here through `CrashLogging.logError()` or `CrashLogging.logMessage()`.  The good thing about these methods is that they will log an error without having any visible effect for the user, while at the same time letting us know that a case we thought was invalid is not.

I think that option 1 may be best for the `.edit` case since the `InteractivePostViewDelegate` gives us a handy implementation for this case.

Option 2 could be better for the `.more` case since I'm not sure we have a handy implementation of it.

My 2 cents.

leandroalonso: Done in 2766441263b7490746e0592a32bf5963b11589e3

PR #13548, opened 2020-02-25
Remove UISplitViewController from the new Me screen

Gio2018:
#### The new Me screen does not use `UISplitViewController` anymore.
- Default modal presentation style for iPhone
- Form sheet presentation style for iPad
- is the app is visualized in a portion of a split screen on iPad with compact horizontal size class, the modal presentation will be the same as an iPhone
- State restoration reflects what was in place with the previous Me tab screen (only at the top level). If we feel that we need to restore other levels, we can handle it in a separate PR.
- The `Done` button is visible at the root level and the first sub level of navigation.
- Took the occasion to reorganize files in Classes->ViewRelated->Me since I was working on those files
##### Default
iPhoneLandscape
 ##### Form Sheet
iPadSplitScreen

Ref #13319

To test:
- Go to `My Site`
- tap the `Me` button

on iPhone with horizontal regular size class (e.g. 11 Pro Max)
- check that the view in portrait and landscape do not have split view
- check that the done button is visible and actionable at the root and first sub level

on iPad
- check that the presentation style is form sheet, and the done button is visible and actionable at root and first sub level

State restoration - on any device
- Tap the `Me` button to open the `Me` screen
- Go to the device or simulator home screen
- Stop the app from Xcode
- Relaunch from Xcode and verify that the `Me` screen shows up

PR submission checklist:


WordPress/Classes/ViewRelated/Me/Me Main/Presenter/MeScenePresenter.swift
```diff
@@ -42,20 +43,23 @@ private extension MeScenePresenter {
...
         let meController = makeMeViewController()
         let navigationController = UINavigationController(rootViewController: meController)
-        meController.navigationItem.leftBarButtonItem = makeDoneButton()
+        navigationController.restorationIdentifier = Restorer.Identifier.navigationController.rawValue
+        meController.navigationItem.rightBarButtonItem = makeDoneButton()
+        // present in formSheet on iPad, default on iPhone
+        if WPDeviceIdentification.isiPad() {
+            navigationController.modalPresentationStyle = .formSheet
+            navigationController.modalTransitionStyle = .coverVertical
+        }
```
bjtitus: From what I can tell, this check isn't necessary since `.formSheet` automatically adjusts depending on the trait environment.

```suggestion
        navigationController.modalPresentationStyle = .formSheet
        navigationController.modalTransitionStyle = .coverVertical
```

WordPress/Classes/ViewRelated/Me/Me Main/MeViewController+UIViewControllerRestoration.swift
```diff
@@ -0,0 +1,36 @@
...
+    override func decodeRestorableState(with coder: NSCoder) {
+        super.decodeRestorableState(with: coder)
+        if FeatureFlag.meMove.enabled {
+            // needs to be done after self has been initialized, so we do it in this method
+            let doneButton = UIBarButtonItem(target: self, action: #selector(dismissHandler))
+            navigationItem.rightBarButtonItem = doneButton
+            if WPDeviceIdentification.isiPad() {
+                navigationController?.modalPresentationStyle = .formSheet
+                navigationController?.modalTransitionStyle = .coverVertical
+            }
```
bjtitus: From what I can tell, this check isn't necessary since `.formSheet` automatically adjusts depending on the trait environment.

> In a vertically regular, but horizontally compact environment, this option displays a sheet interface, where a portion of the underlying content remains visible near the top of the screen.
> 
> In a vertically compact environment, this option is essentially the same as UIModalPresentationStyle.fullScreen

This seems to mean Portrait iPhone would get the `.pageSheet` interface (the existing default for `.automatic`) while Landscape iPhone gets `.fullscreen`.

## After

<a href="https://user-images.githubusercontent.com/3250/75374692-fad13580-5889-11ea-8acd-76fcb0415758.png"></a> <a href="https://user-images.githubusercontent.com/3250/75374700-fdcc2600-5889-11ea-8a41-1f36d3df6a08.png"></a>


```suggestion
            navigationController.modalPresentationStyle = .formSheet
            navigationController.modalTransitionStyle = .coverVertical
```

Gio2018: > From what I can tell, this check isn't necessary since `.formSheet` automatically adjusts depending on the trait environment.
> 
> > In a vertically regular, but horizontally compact environment, this option displays a sheet interface, where a portion of the underlying content remains visible near the top of the screen.
> > In a vertically compact environment, this option is essentially the same as UIModalPresentationStyle.fullScreen
> 
> ###### from [`UIViewController.modalPresentationStyle`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621355-modalpresentationstyle#)
> This seems to mean Portrait iPhone would get the `.pageSheet` interface (the existing default for `.automatic`) while Landscape iPhone gets `.fullscreen`.
> 
> ## After
>  

This seems to be working as expected in iOS 13, but not on older version (Apple documentation should highlight the difference..). Here are a screenshots on iPhone 8 Plus, iOS 11.4 and 12.4
# 
Screen Shot 2020-02-26 at 3 05 10 PM
# 
Screen Shot 2020-02-26 at 3 05 24 PM

Gio2018:
> This looks great! I left one small code comment, and I ran into two functional issues:
> 
> * After navigating into App Settings > About WordPress for iOS > Source Code or Acknowledgements, there's something up with the transition as the bottom toolbar area stays visible when you navigate back. Perhaps we should present these webviews modally?
> * I've not been able to recreate it yet, but I ran into a situation where tapping the gravatar toolbar icon in the blog details list did nothing. It presented the Me view controller, the first time, but on subsequent attempts I could tap it many times and nothing happened. I'll keep trying to recreate.

Good catch!! The changes on `ZendeskUtils.swift` should address this.

WordPress/Classes/ViewRelated/Me/Me Main/MeViewController+UIViewControllerRestoration.swift
bjtitus: Yep! That definitely looks like the case. 😞

Gio2018:
> * After navigating into App Settings > About WordPress for iOS > Source Code or Acknowledgements, there's something up with the transition as the bottom toolbar area stays visible when you navigate back. Perhaps we should present these webviews modally?

Sorry I totally overlooked this, will include it in the next PR!!

PR #13544, opened 2020-02-24
Adds show / hide animations for Create Button

bjtitus:
#13320 

This adds the hide/show functionality to the Floating Create Button and repositions the button to sit over the primary view controller in the split view controller.

<a href="https://user-images.githubusercontent.com/3250/75187309-0cd69b00-5707-11ea-8ce1-5174f09a0d41.gif"></a>

iPadFAB

To test:

- View Blog Details, Post or Site Pages list views to see the create button
- All other views should hide the button

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ Not released

Issue #13320, opened 2020-01-29
Information Architecture: Create Button


bjtitus:
Design: https://wp.me/pauD4L-wu-p2
Latest Updates: https://wp.me/pauD4L-nx-p2

![ios-my-site-blueprint](https://user-images.githubusercontent.com/3250/73475890-781e8e80-434e-11ea-9385-8ea783d34778.png)

- [x] Add feature flag. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13474)
- [x] Add new Create Button (https://github.com/wordpress-mobile/WordPress-iOS/pull/13474)
- [x] Show/Hide Create Button. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13555)
- [x] Add alert sheet to select Post or Page.
- [x] Ensure deep links are using new presentations. (The code for presenting a new post has not changed) (https://github.com/wordpress-mobile/WordPress-iOS/pull/13707)
- [x] Ensure views are correct for Dark Mode.
- [x] Move create button off of Tab Bar view controller and into BlogList or its navigation controller. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13681)
- [x] Add Accessibility support.
  - [x] Disable animation if Reduce Motion is enabled (https://github.com/wordpress-mobile/WordPress-iOS/pull/13555#issuecomment-597044164) (https://github.com/wordpress-mobile/WordPress-iOS/pull/13681)
- [x] Add Tracks. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13705)
- [x] Fix layout issue after logging in to app (button jumps from left to right) (https://github.com/wordpress-mobile/WordPress-iOS/pull/13767)
- [x] Enable feature flag for internal. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13777)
- [x] Enable feature flag for all.
- [ ] Remove feature flag.

bjtitus:
@osullivanchris It came up in a [PR discussion](https://github.com/wordpress-mobile/WordPress-iOS/pull/13474#issuecomment-589910190) that there are some cases where controls are obscured by the Floating Action Button:

[...]

PR #13555, opened 2020-02-26
Add hide/show animation to floating action button

bjtitus:
#13320 

This adds the hide/show functionality to the Create Floating Action Button and repositions the button to sit over the primary view controller in the split view controller.

We've decided to only show the Create button on the Blog Detail screen. Because of that, the implementation has changed to simply show/hide the button when that screen is appearing or disappearing. This simplifies some of the previously complex (and fragile) logic in #13544.

<a href="https://user-images.githubusercontent.com/3250/75308936-e69b2300-580c-11ea-86b6-fcdf1da65505.gif"></a>

iPhoneLandscape

To test:

- View Blog Details page to see the create button
- All other views should hide the button

PR submission checklist:

- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ Not yet released

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
@@ -364,7 +364,9 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
-    [[WPTabBarController sharedInstance] showCreateButton];
+    if ([self.tabBarController isKindOfClass:[WPTabBarController class]]) {
+        [(WPTabBarController *)self.tabBarController showCreateButton];
+    }
     [self createUserActivity];
     [self startAlertTimer];
 }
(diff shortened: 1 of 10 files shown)


bjtitus:
@osullivanchris Thanks for going over this with me yesterday! If you get a chance, can you check build 23613 and see if this works how you expect? https://install.appcenter.ms/orgs/automattic/apps/WPiOS-One-Offs/

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -364,6 +364,9 @@ - (void)viewWillAppear:(BOOL)animated
 - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
+    if ([self.tabBarController isKindOfClass:[WPTabBarController class]]) {
+        [(WPTabBarController *)self.tabBarController showCreateButton];
```
Gio2018: Just a thought/question (does not need to be addressed now): if the `create button` only pertains to `BlogDetailsViewController`, should `WPTabBarController` still handle or be aware of it?

bjtitus: > if the create button only pertains to BlogDetailsViewController, should WPTabBarController still handle or be aware of it?

It definitely could move into BlogDetailsViewController, but the button would still need to be added to the WPTabBarController's view. In my opinion, it seems like WPTabBarController should hold some kind of reference to it because of that.

The view setup code could be moved out of the TabBarController extension if that seems better, though.

bjtitus:
@osullivanchris I'll have you check a build containing the hide/show animation and the action sheet later on so you only have to check one build.

osullivanchris:
> @osullivanchris I'll have you check a build containing the hide/show animation and the action sheet later on so you only have to check one build.

Ok sure. From the gifs provided and slack conversations, all looks good so far!

frosty:
@bjtitus Could we make a small followup to this and disable the animation if a user has Reduce Motion enabled? (checking for `UIAccessibilityIsReduceMotionEnabled`)

PR #13556, opened 2020-02-26
Add GitHub workflows to check PR diff size and milestone

mokagio:
Follows up on the https://github.com/wordpress-mobile/WordPress-iOS/issues/13366 work by migrating the [diff size](https://github.com/Automattic/peril-settings/blob/5d22258ca46c3649f2cb863cfac2d16f12c4cba0/org/pr/diff-size.ts)) and [milestone](https://github.com/Automattic/peril-settings/blob/5d22258ca46c3649f2cb863cfac2d16f12c4cba0/org/pr/milestone.ts) checks to Danger, adding them to the existing GitHub Actions workflow.

To test, please refer to the instructions on the original PRs: https://github.com/mokagio/WordPress-iOS/pull/22 for diff size and https://github.com/mokagio/WordPress-iOS/pull/27 for milestone.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -45,3 +45,9 @@ jobs:
           yarn run danger ci \
             --dangerfile Automattic/peril-settings/org/pr/ios-macos.ts \
             --id consistency_checks
+
+      - name: Check Diff Size
+        run: |
+          yarn run danger ci \
+            --dangerfile Automattic/peril-settings/org/pr/diff-size.ts \
+            --id diff_size
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -5,12 +5,23 @@ on:
     # when the labels change, not only when a PR is opened/reopened or changes
     # are pushed to it.
     types: [opened, reopened, synchronize, labeled, unlabeled]
+  # In order to validate the milestone, we need to listen for the event on the
+  # issue, as there's no event for milestones in PRs –GitHub PRs are a special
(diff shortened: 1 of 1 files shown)


.github/workflows/danger.yml
```diff
@@ -5,12 +5,23 @@ on:
     # when the labels change, not only when a PR is opened/reopened or changes
     # are pushed to it.
     types: [opened, reopened, synchronize, labeled, unlabeled]
+  # In order to validate the milestone, we need to listen for the event on the
+  # issue, as there's no event for milestones in PRs –GitHub PRs are a special
+  # kind of issue. Un
```
jkmassel: Is there a typo at the end of this comment?

.github/workflows/danger.yml
```diff
@@ -5,12 +5,23 @@ on:
...
+  issues:
+    types: [milestoned, demilestoned]
```
jkmassel: It seems the triggers for milestones don't run correctly:

- If I open a new PR with no milestone and no labels, I see warnings about both
- If I then add a milestone, the warning doesn't go away
- If I then add a label, both warnings go away

It turns out this is an issue with Peril as well – it has the same bug. Is it potentially related to the `if` condition below?

mokagio:
Closing in favor of https://github.com/wordpress-mobile/WordPress-iOS/pull/14187

.github/workflows/danger.yml
mokagio: I understood why that happened in #14336.

Issue #13413, opened 2020-02-10
Unified Auth: new Login Epilogue


ScoutHarris:
Implement new Login Epilogue.

Refs:
- [Unified Auth Design](https://wp.me/pbArwn-5X)

Tasks:
- [x] Update 'My Sites' list. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13511)
- [x] Update User Info section. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13543)
- [x] Remove 'Connect another site' button, add link. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13561)
- [x] Update button view. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13570)
- [x] Update iPad layout. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13860)
- [x] Accessibility. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13914)
- [x] User Info - adjust is the amount of extra space between the Gravatar and the display name (https://github.com/wordpress-mobile/WordPress-iOS/pull/13543#issuecomment-596603280) (https://github.com/wordpress-mobile/WordPress-iOS/pull/13699)
- [x] My Sites list - If the user has no sites drop the "My Sites" header (https://github.com/wordpress-mobile/WordPress-iOS/pull/13561#pullrequestreview-371341140) (https://github.com/wordpress-mobile/WordPress-iOS/pull/13699)

Pre-existing issues:
- [x] Login Epilogue: adjust table cells for dynamic text (https://github.com/wordpress-mobile/WordPress-iOS/pull/13814)
- [x] Login: Username & display name never load on login epilogue for certain self-hosted users (https://github.com/wordpress-mobile/WordPress-iOS/pull/13917)

---
Designs from Zeplin:

![Epilogue States](https://user-images.githubusercontent.com/1816888/75069920-73fa1280-54af-11ea-9f80-a3d746404670.png)

![Device Examples — Epilogue](https://user-images.githubusercontent.com/1816888/75069857-5167f980-54af-11ea-9fe8-43606a7029cd.png)

PR #13570, opened 2020-02-27
Login Epilogue: updated button panel view

ScoutHarris:
Ref #13413 

This updates the button panel display (the view behind the Done button). 
- The button panel is the full width of the screen on all devices.
- Removed the shadow that appeared above the button panel when the table was scrollable.
- When the table is scrollable, the button panel is grey/black with a line at the top.
- When the table is not scrollable, the button panel color matches the table, and there is no line.

Not yet updated:
- iPad specific layout, including the Done button width matching the table width.
- Accessibility.

To test:
- Log into the app.
- Verify the button panel behavior is as described above (scrollable vs not).
- Verify the button panel looks like the Zeplin design, or that on #13413 .

|  short_list_light |  short_list_dark |
|--------|-------|
|  long_list_light |  long_list_dark |

- On an iPad, verify the button panel is the full width of the view (it's the same on iPhone, but the difference is only visible on iPad).

| ipad_portrait | ipad_landscape  |
|--------|-------|

PR submission checklist:


WordPress/Classes/ViewRelated/NUX/LoginEpilogueViewController.swift
```diff
@@ -7,13 +7,14 @@ import WordPressAuthenticator
 //
 class LoginEpilogueViewController: UIViewController {
 
-    /// Button's Container View.
+    /// Button Container View.
     ///
     @IBOutlet var buttonPanel: UIView!
 
-    /// Separator: to be displayed above the actual buttons.
+    /// Line displayed atop the buttonPanel when the table is scrollable.
```
mindgraffiti: TIL atop. What a beautiful word!

PR #13572, opened 2020-02-27
Enlarge tappable area behind the `Me` gravatar button

Gio2018:
##### Changes in this PR

- Add a larger transparent view behind the gravatar `Me` button to enlarge the tappable area. To keep the same position of the gravatar, this is vertically centered in the transparent view, and aligned to the right (trailing margin set to 0). The size of the transparent view is 44 x 44

- Add a white border to the gravatar

borderGravatar

- Present modally the web views from `AboutViewController()

acknowledgements

Ref #13319

To test:

- go to `My Site`
- Check that the `Me` gravatar button: 
  - works as expected
  - has a comfortable tap area
 - Go to App Settings -> About
  - check that the web views (e.g. Acknowledgement) are presented modally and no bottom bar persists in the `Me` screen after they are dismissed. 
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
@@ -36,7 +36,7 @@ extension BlogListViewController {
 private extension UIBarButtonItem {
     /// gravatar configuration parameters
     struct GravatarConfiguration {
-        static let radius: CGFloat = 24.0
+        static let radius: CGFloat = 32
         static let tappableWidth: CGFloat = 44
         static let fallBackImage = Gridicon.iconOfType(.user)
     }
--- a/WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
@@ -37,63 +37,92 @@ private extension UIBarButtonItem {
     /// gravatar configuration parameters
     struct GravatarConfiguration {
         static let radius: CGFloat = 24.0
-        static let fallBackImage = Gridicon.iconOfType(.userCircle)
+        static let tappableWidth: CGFloat = 44
+        static let fallBackImage = Gridicon.iconOfType(.user)
     }
 
     /// Assign the gravatar CircularImageView to the customView property and attach the passed target/action.
-    /// If email is nil, fall back to the gravatar icon. Adds `Me` accessibility traits to the button
     convenience init(email: String?, style: UIBarButtonItem.Style = .plain, target: Any?, action: Selector?) {
-        guard let email = email else {
-            self.init(image: GravatarConfiguration.fallBackImage,
-                      style: style,
-                      target: target,
-                      action: action)
-            makeMeButtonAccessible()
-            return
-        }
         self.init()
         makeMeButtonAccessible()
-
-        customView = makeGravatarView(with: email)
+        customView = makeGravatarTappableView(with: email)
         addTapToCustomView(target: target, action: action)
     }
 
     /// Create the gravatar CircluarImageView with a fade animation on tap.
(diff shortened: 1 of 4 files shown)


WordPress/WordPressTest/BlogDetailsViewController+MeButtonTests.swift
```diff
@@ -66,13 +66,21 @@ class BlogDetailsViewControllerTests: XCTestCase {
...
+        guard controller.navigationItem.rightBarButtonItem != nil else {
             XCTFail("Me Button not installed")
             return
         }
         scenePresenter?.presentExpectation = expectation(description: "Me was presented")
         // When
-        _ = meButton.target?.perform(meButton.action)
+
+        guard let target = blogDetailsViewController?.target(forAction: #selector(BlogDetailsViewController.presentHandler), withSender: blogDetailsViewController) else {
+            XCTFail("Cippa")
```
ScoutHarris: I'm thinking this was meant to be replaced with a meaningful fail message. 😄

Gio2018: oooooppsssss yes it was :)

Gio2018:
>I think we can get away with 32 visually without it standing out too much. Not sure how hard it is to tweak the code though?
Not hard at all! Just replacing 24 with 32!
> So could we increase the image to 32p visually, and the tappable area could remain @ 44 as described?

Totally. Note that on smaller iPhones (SE, 6s, X and 11 Pro), it does take the whole height of the top bar (but does not alter the size of the bar itself)

gravatar32Toolbar

I pushed the change that resizes it to 32.
If you wanna play with the gravatar size yourself, in Xcode, find the file `UIBarButtonItem+MeBarButton.swift` (if you don't find it in the project navigator, hit cmd + shift + O and type the name in the search bar), go to line 39 and find this statement:
`static let radius: CGFloat = 32`

that is the size of the gravatar, as you change it, you will see different sizes of it after you build.

Gio2018:
> > Totally. Note that on smaller iPhones (SE, 6s, X and 11 Pro), it does take the whole height of the top bar (but does not alter the size of the bar itself)
> 
> Do you think we should make it a different size on smaller iPhones? The screenshot is on landscape - if its just on landscape I'm not worried. But if its squished on portrait on smaller iPhones perhaps we could make the image smaller just for those?

Here are iPhoneX, iPhone 11 Pro Max and iPhone SE in portrait, they all look fine and quite similar in portrait (SE is larger, but expected cause it's lower resolution, and it's the smallest iPhone that we support), so I think we could use 32

iPhoneX
iPhoneX


iPhone 11 Pro Max 
iPhone11ProMax

 
iPhone SE
iPhoneSE


> 
> > I pushed the change that resizes it to 32.
> 
> Thanks!! That's great. By the way - if the user does not have a gravatar we display the icon - will that icon be 24 or 32?

The icon will have the same size of the gravatar, so 32

ScoutHarris:
> I pushed the change that resizes it to 32

Oh that made a big difference, especially on iPad. It's not teeny tiny anymore! 😄 Nice!

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
```diff
@@ -36,64 +36,93 @@ extension BlogListViewController {
...
+    /// adds a 'tap' action to customView
+    func addTapToCustomView(target: Any?, action: Selector?) {
+        let tapRecognizer = UITapGestureRecognizer(target: target, action: action)
+        customView?.addGestureRecognizer(tapRecognizer)
+    }
 
-        gravatarImageView.translatesAutoresizingMaskIntoConstraints = false
-        NSLayoutConstraint(item: gravatarImageView,
+    /// embeds a view in a larger tappable area, vertically centered and aligned to the right
+    func embedInTappableArea(_ imageView: UIImageView) -> UIView {
```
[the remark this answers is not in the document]
Gio2018: I love this!!! I'll do a separate PR to avoid blocking the feature

PR #13587, opened 2020-03-04
Removes the postPreview feature flag.

diegoreymendez:
This PR shouldn't change the existing logic, since the feature flag is set to `true` at all times.  There's just one exception for which I'll add a comment in the code for the reviewer.

## Context / References:

Second attempt at removing the `postPreview` feature flag.  First attempt was here: https://github.com/wordpress-mobile/WordPress-iOS/pull/13357

## Testing:

It's important for all test scenarios below that you don't use an atomic private site.  Atomic site previews are currently failing for private sites since we're missing some necessary cookies.

### Test 1: Spotlight Search

Go to [this method](https://github.com/wordpress-mobile/WordPress-iOS/blob/9d533e2acc8fa73fb83ac83ce4b418d9c9b310e6/WordPress/Classes/Utility/Spotlight/SearchManager.swift#L411) and edit it to make sure it returns immediately after calling the `onFailure` block.

Insert at the first and second lines:

```swift
onFailure()
return
```

Then:

1. Get the name of a published post in WPcom.
2. Go to make a Spotlight search and insert the name of the post.
3. When the result comes up tap on it.

Make sure that:

- [This method](https://github.com/wordpress-mobile/WordPress-iOS/blob/508b8b9d76af23aa42a7d05384061d84ee925ee7/WordPress/Classes/Utility/Spotlight/SearchManager.swift#L474-L484) is called.
- The preview looks good.

### Test 2: Posts & Pages list.

This test should be executed for both posts and pages.

1. Go to the list of posts / pages.
2. Select the published filter.
3. Tap "View" to preview the post.

Make sure the preview looks good.

### Test 3: Editor Preview.

This test should be executed for both posts and pages.

1. Edit an existing published post or page.
2. Select the option to preview the post / page.

Make sure the preview looks good.

### Test 4: New-Post Preview.

1. Create a new post.
2. Publish it.
3. As soon as it's published a notice should come up saying "Post Published".  Tap on the "View" action in the notice.
4. The Post Epilogue screen will come up, tap on "View Post".

Make sure the preview looks good.

### Test 5: New-Page Preview.

1. Create a new page.
2. Publish it.
3. As soon as it's published a notice should come up saying "Page Published".  Tap on the "View" action in the notice.

Make sure the preview looks good.

## PR submission checklist:


WordPress/Classes/ViewRelated/Post/PostNoticeNavigationCoordinator.swift
```diff
@@ -23,23 +23,13 @@ class PostNoticeNavigationCoordinator {
                 return
         }
 
-        if FeatureFlag.postPreview.enabled {
-            let controller = PostPreviewViewController(post: page)
-            controller.navigationItem.title = NSLocalizedString("View", comment: "Verb. The screen title shown when viewing a post inside the app.")
-            controller.hidesBottomBarWhenPushed = true
+        let controller = PreviewWebKitViewController(post: page)
```
diegoreymendez: I think this was a bug in the original code, since we were showing PostPreviewViewController when FeatureFlag.postPreview.enabled. So I just left the else condition instead.

PR #13596, opened 2020-03-04
WordPressComSyncService: Synchronizing Settings

### LINKED ISSUE #13594: Signup Broken
### Details:
1. Fresh install the app
2. Signup for a WordPress.com account
3. Enter your email
4. Press over the Magic Link email

As a result, the Signup Epilogue will remain stuck with the following error in the console:

```
Tried to apply a change to nonexistent settings
```

Screen Shot 2020-03-04 at 12 31 03 PM copy

jleandroperez:
### Details:
In this PR we're adding new API to **AccountSettingsService**, that deals with Display Name changes, and wiring it in the Signup Epilogue. Why:

- Former mechanism involved using `AccountSettingsChange`, meant to keep Core Data in consistent state, in the scenario of a REST call failure
- The Signup Epilogue **isn't tied up** (AKA does not populate) its fields from Core Data
- The **AccountSettings** are not synchronized in the Epilogue
- And furthermore, the Epilogue **does not need** the **AccountSettings**

For the reasons above, it makes 1000% sense to mimic the same mechanisms used in the Epilogue to (A) Change Passwords or (B) Change Username, without going thru `AccountSettingsChange` (which is not really needed here).

Fixes #13594

cc @diegoreymendez WDYT? (Thanks in advance!!

### To test:

1. Fresh install the app
2. Signup for a new WordPress.com account (with your email)
3. Click over the Magic Link

- [x] Verify you get the Signup Epilogue right after the Magic Link is processed
- [x] Verify that you can set a custom password
- [x] Verify that you can change the Display Name
- [x] Verify that pressing over **Continue** effectively dismisses the Epilogue

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/AccountSettingsService.swift
+++ b/WordPress/Classes/Services/AccountSettingsService.swift
@@ -127,6 +127,17 @@ class AccountSettingsService {
         }
     }
 
+    func updateDisplayName(_ displayName: String, finished: ((Bool, Error?) -> ())? = nil) {
+        remote.updateSetting(.displayName(displayName), success: {
+            finished?(true, nil)
+        }) { error in
+            DDLogError("Error saving account settings change \(error)")
+            NotificationCenter.default.post(name: NSNotification.Name.AccountSettingsServiceChangeSaveFailed, object: self, userInfo: [NSUnderlyingErrorKey: error])
+
+            finished?(false, error)
+        }
+    }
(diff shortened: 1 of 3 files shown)


jleandroperez:
Thank you Diego!!!

diegoreymendez:
Awesome work @jleandroperez !

PR #13615, opened 2020-03-10
Fixes the split view for the existing Me Tab

Gio2018:
Fixes (Issue not yet opened)

To test:
- Disable the feature flag for `Me`
- Build/run
- Go to the `Me` tab
- check that the sub levels of the navigation display correctly on iPad (portrait and landscape) or iPhone 11 Pro Max landscape, as well as any iPhone in portrait
- Repeat the tests with the feature flag enabled.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Me/Me Main/MeViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/Me Main/MeViewController.swift
@@ -74,7 +74,7 @@ class MeViewController: UITableViewController {
     //TODO: Remove when the new Me scene is shipped
     override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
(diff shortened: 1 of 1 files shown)


frosty:
Oh, I just noticed that the MeViewController cells show the disclosure arrow when first rotated into landscape – I would expect the arrow to not be shown then: (iPhone XR)

Screenshot 2020-03-10 at 16 32 16

If I rotate back to portrait and then back to landscape, the arrows disappear.

Gio2018:
> Oh, I just noticed that the MeViewController cells show the disclosure arrow when first rotated into landscape – I would expect the arrow to not be shown then: (iPhone XR)

Thanks for catching this! There was a wrong feature flag check in `traitCollectionDidChange()` and `accountDidChange()`, just pushed the commit for it.

Gio2018:
@loremattei just pinging you here to give you a heads-up that this PR fixes a bug found in 14.4, thanks!!!

PR #13626, opened 2020-03-11
Add missing welcome notification indicator 

### LINKED ISSUE #13593: Missing Tabbar notification indicator (dot) on initial welcome notification
### Expected behavior
After signup, I expect to see a notification indicator for the welcome notification

### Actual behavior
Indicator not showing even though the welcome notification exists

1. Complete signup process
2. See no notification indicator 
3. Go to notifications and see the welcome notification

yaelirub:
Fixes #13593 

As of iOS 8, the application must register for user notifications using -[UIApplication registerUserNotificationSettings:] before being able to set the icon badge which we are relying on when setting the notification tab bar item's to it's unread state. 

Since we don't prompt the user to allow notification (registerUserNotificationSettings) right after signup, we need to show the "unread notification" icon in a different way.

This PR, adds a user defaults key named "welcomeNotificationSeen" and sets it to false right after signup (when the user taps continue on the signup epilogue. 

`updateNotificationBadgeVisibility` in WPTabController, in addition to checking the `applicationIconBadgeNumber` is now also looking for the value of the new UserDefaults key to determine if to show the unread notification icon.

The value will be set to true (marks the notification as read) only when the user opens the notification controller. 

To test:
1. Go through Signup for WordPress.com 
2. See signup epilogue screen
3. tap continue
4. Skip or don't skip site creation
5. See dot in notification tabbar item 
6. See that dot persists till you tap on the notification tab bar item

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/System/WPTabBarController.m
+++ b/WordPress/Classes/ViewRelated/System/WPTabBarController.m
@@ -70,6 +70,8 @@ static CGFloat const WPTabBarIconSize = 32.0f;
 
 @property (nonatomic, strong) CreateButtonCoordinator *createButtonCoordinator;
 
+@property (nonatomic, assign) BOOL isFromSignup;
+
 @end
 
 @implementation WPTabBarController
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/System/WPTabBarController.m
```diff
@@ -913,7 +913,12 @@ - (void)updateNotificationBadgeVisibility
     // Discount Zendesk unread notifications when determining if we need to show the notificationsTabBarImageUnread.
     NSInteger count = [[UIApplication sharedApplication] applicationIconBadgeNumber] - [ZendeskUtils unreadNotificationsCount];
     UITabBarItem *notificationsTabBarItem = self.notificationsNavigationController.tabBarItem;
-    if (count > 0) {
+
+    NSUserDefaults *standardUserDefaults = [NSUserDefaults standardUserDefaults];
```
leandroalonso: What about extracting that logic to a private method? `welcomeNotificationSeen` may even be the name of the method.

PR #13629, opened 2020-03-11
Add tracks without the need of updating WordPressShared

leandroalonso:
### Goal

To decentralize the `WPAnalytics` from `WordPressShared`. Nowadays every single time we need to add a new track event you must change the `WordPressStat` in `WordPressShared`. That creates a lot of bureaucracy.

### What have been done so far

I haven't changed the way `WordPressStat` (the current "global" enum) works. Neither have I changed the existent tracks (except Media Editor).

My idea is to do a non-destructive refactor. This is the first step — which allows us to add events from the `WPiOS` project.

### To test

I migrated the events of showing and using the Media Editor.

1. Open the Media Editor
2. Check that the event is correct fired

### Usage

```swift
WPAnalytics.track(.mediaEditorUsed, properties: ["actions": actions.description])
```

In Obj-C:

```
[WPAnalytics trackEvent:WPAnalyticsEventMediaEditorUsed];
```


leandroalonso:
@diegoreymendez updated. Can you please review it again?

leandroalonso:
@diegoreymendez available again! Now supporting Obj-C! 😎

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
```diff
@@ -0,0 +1,71 @@
...
+    var value: String {
+        switch self {
+        case .mediaEditorShown:
+            return "media_editor_shown"
+        case .mediaEditorUsed:
+            return "media_editor_used"
+        }
+    }
+
+    var properties: [AnyHashable: Any]? {
```
diegoreymendez: Is this being used?

PR #13648, opened 2020-03-13
Notification Layout Issues

### LINKED ISSUE #13646: Notifications: Layout Issues
### Details:
We've received reports of Comment-Y Notifications getting cut off. Opening this issue to properly track it.

The bug simply involves the Comment's payload getting cut off. In the screenshot pasted below, there's an extra line not being displayed

jleandroperez:
### Description:
In this PR we're working around an Autolayout-Engine-Y (alledged) bug, which results in text getting cut off.

**PLUS** since I was in the neighborhood, I'm patching up the Comment Notification Actions spacing, since current layout yields cutoff text in smaller devices.

@diegoreymendez May I bug you with a hopefully quick one?

### Details:
After literally... hours of debugging, i've managed to pinpoint exactly what's going on.

0. **NoteBlockCommentTableViewCell** is the exact component that's getting cutoff
1. During cell initialization (before there's a superview) we set all of its properties
2. Whenever **NoteBlockCommentTableViewCell.isApproved** runs, we were executing the method **refreshSeparators**
3. The above lead, in a specific flow, into the execution of **NoteBlockTableViewCell.readableSeparatorInsets** without a proper superview
4. This translates into **UITableViewCell.readableContentGuide** being executed

The above should not be a problem. In theory. In practise, what actually happens is that the **intrinsicContentSize** for **RichTextView** is miscalculated (becuse the width constraints are unknown).

This invalid value shouldn't be bugging us, but **UIKit appears to be caching it**. And this is, precisely, the source of our bug.

In this PR we're simply deferring the execution of this entire sequence, up until `wilDisplayCell`, when there's an actual superview, and calculations won't be off.

Fixes #13646

### To test:
1. Checkout commit ff48e233f9
2. Launch WordPress into an **iPhone SE** Simulator
3. Open the Notifications Tab
4. Press over the only notification you've got
5. Verify that the text gets cut off (should end in `Thanks @wwwww for`)
6. Checkout commit d71fcb5f1f (HEAD of this branch)
7. Reopen the same notification as in (5)

- [x] Verify that the note now ends with `Thanks @wwwww for callint it out 👍`

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/NotificationSyncMediator.swift
+++ b/WordPress/Classes/Services/NotificationSyncMediator.swift
@@ -98,7 +98,14 @@ class NotificationSyncMediator {
                 return
             }
 
-            self.deleteLocalMissingNotes(from: remoteHashes) {
+            let hackedHashes = [
+                RemoteNotification(document: [
+                    "id": 1234,
+                    "note_hash": 9999999
+                    ] as [String: AnyObject] )!
+            ]
+
+            self.deleteLocalMissingNotes(from: hackedHashes) {
 
                 self.determineUpdatedNotes(with: remoteHashes) { outdatedNoteIds in
                     guard outdatedNoteIds.isEmpty == false else {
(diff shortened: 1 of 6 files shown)


diegoreymendez:
Just a note in case someone wants to re-test this: I could not reproduce the original issue using iOS 11.4.  I could, however, reproduce it using. iOS 13.4.

jleandroperez:
Thank you Diegoooo!!!

PR #13673, opened 2020-03-17
Implements AtomicAuthenticationService in WPiOS.

### LINKED ISSUE #13563: Implement class AtomicAuthenticationService.
This issue depends on: https://github.com/wordpress-mobile/WordPressKit-iOS/issues/216

Implement class `AtomicAuthenticationService` with a method to return a `HTTPCookie` object.

It will call `AtomicAuthenticationRemoteService` and map the resulting object to a `HTTPCookie` object and return it.

diegoreymendez:
Fixes #13563 

Implements a class that provides a service to retrieve Atomic Authentication Cookies from remote.

## To test:

This can't be tested as the service isn't wired in WPiOS yet.  We'll provide manual testing steps once it's integrated.

For now make sure the unit test provided makes sense.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/AtomicAuthenticationService.swift
+++ b/WordPress/Classes/Services/AtomicAuthenticationService.swift
@@ -1,0 +1,20 @@
+import Foundation
+import CocoaLumberjack
+import WordPressKit
+
+class AtomicAuthenticationService {
+    let remote: AtomicAuthenticationServiceRemote
+    fileprivate let context = ContextManager.sharedInstance().mainContext
+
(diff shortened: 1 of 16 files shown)


WordPress/Classes/Services/AtomicAuthenticationService.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+import CocoaLumberjack
```
leandroalonso: Do we need to import that here?

diegoreymendez: Removed.

PR #13682, opened 2020-03-18
Update Gridicons to 1.0

frosty:
Updates to use the new 1.0 version of the Gridicons framework. Yes, there are a lot of changes here! I don't expect you to look through every line of code change though :)

**To test**

* Skim through the code changes, pick a few areas and make sure the changes look okay
* Build and run, and do a smoke test through the app checking icons look as you expect

PR submission checklist:


Podfile
```diff
@@ -54,7 +55,7 @@ def shared_with_all_pods
     wordpress_shared
     pod 'CocoaLumberjack', '3.5.2'
     pod 'FormatterKit/TimeIntervalFormatter', '1.8.2'
-    pod 'NSObject-SafeExpectations', '0.0.4'
+    pod 'NSObject-SafeExpectations', '~> 0.0.4'
```
Gio2018: This is ok, but why do we need to explicitly add ~>? This should be needed if we want to automatically update to the next patch when we update pods, and since 0.0.4 is the latest version, this should really be needed here? (same for the other parts where ~> was added)

WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
```diff
@@ -9,7 +9,7 @@ import Gridicons
     }
 
     var button: FloatingActionButton = {
-        let button = FloatingActionButton(image: Gridicon.iconOfType(.create))
```
Gio2018: As @mindgraffiti pointed out, line 73 and 78 of this file still have the old syntax, and need to be updated.

Podfile
frosty: I think this should be safe as it should only update to 'compatible' versions (0.0.x) and it make make the process smoother when you need to pull in new versions of our own pods which may rely on the same libraries.

WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
frosty: Thanks!

frosty:
Thanks both for your reviews!

PR #13707, opened 2020-03-23
Add deep link for Page editor

bjtitus:
Adds the following Deep Link routes:

`wordpress://newpage` following the same URL query parameters as the `newpost` URL but without`tags`:
* `content` at minimum
* `title`

`https://apps.wordpress.com/get#/page`
`https:/apps.wordpress.com/get#/page/<domain>`

Part of #13320

## To test
- Open `https://apps.wordpress.com/get#/page` as a link (I used Messages to send myself the link and open the app. All other copies of the app must be uninstalled since I believe it is random which copy of the app is chosen to open the URL)
- Open `wpdebug://newpage?content=test` as a link (also used Messages for this one after configuring the app to register for the wordpress scheme)

PR submission checklist:

- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ Unreleased, although this could theoretically be released without enabling the Create button.

### DIFF
--- a/WordPress/Classes/System/WordPressAppDelegate+openURL.swift
+++ b/WordPress/Classes/System/WordPressAppDelegate+openURL.swift
@@ -137,7 +137,7 @@ import AutomatticTracks
             return false
         }
 
-        // Should more formats be accepted be accepted in the future, this line would have to be expanded to accomodate it.
+        // Should more formats be accepted in the future, this line would have to be expanded to accomodate it.
         let contentEscaped = contentRaw.escapeHtmlNamedEntities()
 
         let post = PostService(managedObjectContext: context).createDraftPost(for: blog)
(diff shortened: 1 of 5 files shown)


WordPress/Classes/System/WordPressAppDelegate+openURL.swift
```diff
@@ -153,6 +155,35 @@ import AutomatticTracks
...
+
+        let title = params.value(of: NewPostKey.title)
+
+        let context = ContextManager.sharedInstance().mainContext
+        let blogService = BlogService(managedObjectContext: context)
+        guard let blog = blogService.lastUsedOrFirstBlog() else {
+            return false
+        }
+
+        // Should more formats be accepted be accepted in the future, this line would have to be expanded to accomodate it.
```
ScoutHarris: Dupe 'be accepted' here.

bjtitus:
@ScoutHarris Thanks! For some reason that wasn't happening with the app's URL scheme for me. The editor was being presented twice in a row.

PR #13712, opened 2020-03-24
Adapt WebViewAuthenticator for atomic

### LINKED ISSUE #13566: Changes to WebViewAuthenticator for Atomic Private Sites.
We need to modify several things within `WebViewAuthenticator`:

1. Check if the atomic cookie is present (for atomic private sites).  We can probably modify this verification slightly to behave differently for atomic sites: https://github.com/wordpress-mobile/WordPress-iOS/blob/34c4e7db5b551173cb69978b72d3fb9f4e426f00/WordPress/Classes/Utility/WebViewAuthenticator.swift#L72-L81
2. Consider modifying this method to return using a completion block instead.  Right now this method can authenticate and redirect the user to the desired URL, but the atomic authentication does not redirect, so we'll need to perform two requests: https://github.com/wordpress-mobile/WordPress-iOS/blob/34c4e7db5b551173cb69978b72d3fb9f4e426f00/WordPress/Classes/Utility/WebViewAuthenticator.swift#L131-L137
3. Modify this method to [request the auth cookie for atomic sites](https://github.com/wordpress-mobile/WordPress-iOS/issues/13563), and then request the original URL.   We'll also need this to return using a completion block: https://github.com/wordpress-mobile/WordPress-iOS/blob/34c4e7db5b551173cb69978b72d3fb9f4e426f00/WordPress/Classes/Utility/WebViewAuthenticator.swift#L118-L127

## Dependencies:

https://github.com/wordpress-mobile/WordPress-iOS/issues/13562 - To be able to tell when a site is atomic
https://github.com/wordpress-mobile/WordPress-iOS/issues/13563 - To be able to request the atomic auth cookie

diegoreymendez:
Fixes #13566 

This PR implements new code to request special auth cookies for Atomic Private Sites.

This PR also modularizes and simplifies the WebViewAuthenticator code, so that it doesn't need to know the details of how to request authorization cookies.

We're no longer requesting the cookies directly through the web view but instead using ephemeral URL Sessions within services.  This is good in theory because it allows our code to request those authorization cookies regardless of if we are using a web view.

### Test 1:

1. Create an atomic private site.
2. Make sure it's private.
3. Preview the site, a page and a post.

Make sure you can preview the content just fine.
Do the same test for post and page previews.

### Test 2:

Same but with a WP.com site.
Do the same test for post and page previews.

### Test 3:

Same but with a self-hosted site connected through XMLRPC.
Do the same test for post and page previews.

### Test 4:

Same but with a self-hosted site connected through Jetpack.
Do the same test for post and page previews.

### Test 5:

1. Create a self hosted site.
2. Connect in the app through XMLRPC.
3. Go to the site's dashboard in the app and tap "Stats".
4. Complete the Jetpack installation and connection.

Make sure it works fine.

### Test 6:

Follow a private site.
Open posts from it in the reader.

## PR submission checklist:


diegoreymendez:
@koke - I'm not sure I've covered this scenario: https://github.com/wordpress-mobile/WordPress-iOS/pull/13712/files#diff-8c7121e5f4cf2c659742cb79bce89011R291

Can you give me your thoughts on it?

diegoreymendez:
I had to fix some broken unit tests which weren't trivial.  The PR is ready for review now, apologies for the extra noise.

WordPress/Classes/Services/AuthenticationService.swift
```diff
@@ -0,0 +1,190 @@
...
+        var request = URLRequest(url: loginURL)
+
+        request.httpMethod = "POST"
+        request.httpBody = body(withParameters: [
+            "log": username,
+            "pwd": password,
+            "rememberme": "true"
+        ])
+
+        let task = session.dataTask(with: request) { data, response, error in
```
aerych: It looks like the task code could be factored into its own method accepting a URL, success, and failure blocks if we want to reduce the duplication in the wpcom and selfhosted methods.

WordPress/Classes/Services/AtomicAuthenticationService.swift
```diff
@@ -1,20 +1,59 @@
...
 class AtomicAuthenticationService {
+
     let remote: AtomicAuthenticationServiceRemote
     fileprivate let context = ContextManager.sharedInstance().mainContext
 
     init(remote: AtomicAuthenticationServiceRemote) {
         self.remote = remote
     }
 
+    init(account: WPAccount) {
```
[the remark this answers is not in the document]
diegoreymendez: Done.

WordPress/Classes/Services/AuthenticationService.swift
diegoreymendez: Been playing a bit with this.

The thing I'd like to avoid is having a single parametrized-method to configure the parameters for the different calls (which is kind of what we had before), since parameter configuration IMHO is specific to each call and not really shared.

With that in mind... how does this look to you?

```swift
private func requestAuthCookies(
        from url: URL,
        headers: [String: String],
        parameters: [String: String],
        success: @escaping (_ cookies: [HTTPCookie]) -> Void,
        failure: @escaping (Error) -> Void) {

        // We don't want these cookies persisted in other sessions
        let session = URLSession(configuration: .ephemeral)
        var request = URLRequest(url: url)

        request.httpMethod = "POST"
        request.httpBody = body(withParameters: parameters)
        
        headers.forEach { (key, value) in
            request.setValue(value, forHTTPHeaderField: key)
        }

        let task = session.dataTask(with: request) { data, response, error in
            if let error = error {
                failure(error)
                return
            }

            // The following code is a bit complicated to read, apologies.
            // We're retrieving all cookies from the "Set-Cookie" header manually, and combining
            // those cookies with the ones from the current session.  The reason behind this is that
            // iOS's URLSession processes the cookies from such header before this callback is executed,
            // whereas OHTTPStubs.framework doesn't (the cookies are left in the header fields of
            // the response).  The only way to combine both is to just add them together here manually.
            //
            // To know if you can remove this, you'll have to test this code live and in our unit tests
            // and compare the session cookies.
            let responseCookies = self.cookies(from: response, loginURL: url)
            let cookies = (session.configuration.httpCookieStorage?.cookies ?? [HTTPCookie]()) + responseCookies
            success(cookies)
        }

        task.resume()
    }
```

Which then can be called like this:

```swift
    func getAuthCookiesForWPCom(
        username: String,
        authToken: String,
        success: @escaping (_ cookies: [HTTPCookie]) -> Void,
        failure: @escaping (Error) -> Void) {
        
        let loginURL = URL(string: AuthenticationService.wpComLoginEndpoint)!
        let headers = [
            "Authorization": "Bearer \(authToken)"
        ]
        let parameters = [
            "log": username,
            "rememberme": "true"
        ]

        requestAuthCookies(
            from: loginURL,
            headers: headers,
            parameters: parameters,
            success: success,
            failure: failure)
    }
```

aerych: Oh! You went farther than I was thinking. I was just thinking about the task construction.  I agree it would be nice to not over-parameterize things. All I was thinking of was something like

```
func startDataTask(for request: URLRequest,
                        url: URL,
                        success: @escaping (_ cookies: [HTTPCookie]) -> Void,
                        failure: @escaping (Error) -> Void) {
        let session = URLSession(configuration: .ephemeral)
        let task = session.dataTask(with: request) { data, response, error in
            if let error = error {
                failure(error)
                return
            }

            // The following code is a bit complicated to read, apologies.
            // We're retrieving all cookies from the "Set-Cookie" header manually, and combining
            // those cookies with the ones from the current session.  The reason behind this is that
            // iOS's URLSession processes the cookies from such header before this callback is executed,
            // whereas OHTTPStubs.framework doesn't (the cookies are left in the header fields of
            // the response).  The only way to combine both is to just add them together here manually.
            //
            // To know if you can remove this, you'll have to test this code live and in our unit tests
            // and compare the session cookies.
            let responseCookies = self.cookies(from: response, loginURL: url)
            let cookies = (session.configuration.httpCookieStorage?.cookies ?? [HTTPCookie]()) + responseCookies
            success(cookies)
        }

        task.resume()
    }
```

diegoreymendez:
@aerych - Took a while, but this is ready for another look.

WordPress/Classes/Services/AuthenticationService.swift
diegoreymendez: I implemented a solution, pls check it out.

diegoreymendez: Oh my, sorry that last message was sent before I read yours! 😄

WordPress/Classes/Utility/WebViewAuthenticator.swift
```diff
@@ -70,154 +98,128 @@ class WebViewAuthenticator: NSObject {
     ///     authentication, or a request for the original URL.
     ///
     @objc func request(url: URL, cookieJar: CookieJar, completion: @escaping (URLRequest) -> Void) {
-        cookieJar.hasCookie(url: loginURL, username: username) { [weak self] (hasCookie) in
-            guard let authenticator = self else {
-                return
-            }
-
-            let request = authenticator.request(url: url, authenticated: !hasCookie)
+        func done() {
```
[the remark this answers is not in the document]
diegoreymendez: Thanks!  This was indeed old code.

PR #13726, opened 2020-03-25
Fix/post scheduling hotfix

### LINKED ISSUE #13654: Cannot change a draft to a use a date in the past without publishing it
Steps to reproduce:

1. Go to My Site > Blog Posts > + to start a new draft.
1. Add a title and content.
1. Go to More > Post Settings > Publish. 
1. Change to a date in the past.
1. Go back to the edit view.
1. Tap the X button at top left.
1. Tap "Update Post".

**Result:** the post is published when I thought a draft would be saved. ([2m21s](https://cloudup.com/cbvtcS95kbq))

Screen Shot 2020-03-13 at Mar 13 4 35 55 PM Screen Shot 2020-03-13 at Mar 13 4 36 07 PM
<sup>Tested with WPiOS 14.4.0.0 TestFlight beta on iPhone 11 iOS 13.3.1.</sup>

### LINKED ISSUE #13655: "Move to Draft" removes date and time from post settings
Steps to reproduce:

1. Go to My Site > Blog Posts > Published.
1. Pick out a post that was published in the past and note the publish date.
1. Tap More > Move to Draft.
1. Go to Drafts in the Posts Lists.
1. Open the post.
1. Tap More > Post Settings.
1. Notice that the publish date is lost.

**Result:** if I move a post from published to draft, the post date is lost. The expected behavior is that the post date should be kept and only the state should change. ([42](https://cloudup.com/clGyL27joGW))

Screen Shot 2020-03-13 at Mar 13 4 59 16 PM
<sup>Tested with WPiOS 14.4.0.0 TestFlight beta on iPhone 11 iOS 13.3.1.</sup>

### LINKED ISSUE #13710: Cannot back date a draft
### Expected behavior

I expect to be able to back date a new post by selecting a publish date and time in the past.

### Actual behavior

I am able to select a date and time in the past in the publish calendar but if I select a date and time in the past it doesn't get saved.

1. Go to My Site > Blog Posts > +.
1. Add a title and some content.
1. Go to More > Post Settings > Publish.
1. Change the date to something in the past.
1. Try to exit the post with the post backdated.

Result: I am allowed to select a date in the past using the calendar but the date is not reflected on the Post Settings > Publish screen or saved as part of the post. ([1m10s](https://cloudup.com/cJtXr0zWuxc))

##### Tested with WPiOS 14.5.0.0 TestFlight beta on iPhone 11 iOS 13.3.1.

Aside: For reference/comparison, here's how the flow looked in 14.3.1 where it was very easy to mistakenly publish a draft using the same steps ([1m22s](https://cloudup.com/cpjrEXZA7No)).

bjtitus:
Fixes #13655
Fixes #13654
Fixes #13710

Fixes the following issues:

* "Publish Immediately" logic being shown when Draft was back scheduled. (seems to be the same issue seen in #13710)
* Post button (in the upper right of the editor) displaying the wrong action after changing the publish date.
* Adds a simple confirmation dialog so it is extra clear when a Post moves from a non-published status to Published.

### Back Scheduled Draft

A draft should be able to be scheduled in the past. Previously, "Immediately" was shown as the selected Date/Time. See https://github.com/wordpress-mobile/WordPress-iOS/issues/13710 for more details.

| Before | After |
|--|--|
| <a href="https://user-images.githubusercontent.com/3250/77583593-fa3ab780-6ea6-11ea-94b6-88c20563d2d7.gif"></a> | <a href="https://user-images.githubusercontent.com/3250/77567638-1598c900-6e8d-11ea-905f-53f346ef51e4.gif"></a> |

### Confirmation Dialog

A confirmation dialog should show any time the post is Published from a non-published status. This is a temporary solution for https://github.com/wordpress-mobile/WordPress-iOS/issues/13654 to make it more obvious to users when a post will move to "Published". _Note that this happens _any time_ a post moves to Published from another status, not just when backdating a post._

<a href="https://user-images.githubusercontent.com/3250/77569847-77a6fd80-6e90-11ea-9815-bb8c86f3110c.gif"></a>

### Move to Draft

Move an existing Published post to Draft. The Published date in Post Settings should be retained. See https://github.com/wordpress-mobile/WordPress-iOS/issues/13655 for more details.

## Known Issues

* "Publish Immediately" on an already published post changes the published date to the current date and time but does not show "Immediately".
* Scheduling a Post and then Publishing it seems to show the wrong "Action" button. Shouldn't the button say "Publish" since the status is changing from "Scheduled" to "Published"? The Publish confirmation is also not shown in this case.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/AbstractPost.m
+++ b/WordPress/Classes/Models/AbstractPost.m
@@ -420,7 +420,7 @@
 
 - (BOOL)shouldPublishImmediately
 {
-    return [self originalIsDraft] && ![self hasFuturePublishDate];
+    return [self originalIsDraft] && self.dateCreated == nil && ![self hasFuturePublishDate];
 }
 
 - (NSString *)authorNameForDisplay
--- a/WordPress/Classes/Models/AbstractPost.m
+++ b/WordPress/Classes/Models/AbstractPost.m
@@ -420,7 +420,7 @@
 
 - (BOOL)shouldPublishImmediately
 {
-    return [self dateCreatedIsNilOrEqualToDateModified];
+    return [self originalIsDraft] && [self dateCreatedIsNilOrEqualToDateModified];
 }
 
 - (NSString *)authorNameForDisplay
--- a/WordPress/Classes/Models/AbstractPost.m
+++ b/WordPress/Classes/Models/AbstractPost.m
@@ -420,7 +420,7 @@
 
 - (BOOL)shouldPublishImmediately
 {
-    return [self originalIsDraft] && self.dateCreated == nil && ![self hasFuturePublishDate];
+    return [self dateCreatedIsNilOrEqualToDateModified];
 }
 
 - (NSString *)authorNameForDisplay
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -111,7 +111,7 @@ extension PostEditor where Self: UIViewController {
             !UserDefaults.standard.asyncPromoWasDisplayed {
             promoBlock()
         } else if action.isAsync,
-            let postStatus = self.post.status,
+            let postStatus = self.post.original?.status ?? self.post.status,
             ![.publish, .publishPrivate].contains(postStatus) {
             // Only display confirmation alert for unpublished posts
             displayPublishConfirmationAlert(for: action, onPublish: publishBlock)
--- a/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
@@ -70,6 +70,11 @@ struct PublishSettingsViewModel {
     mutating func setDate(_ date: Date?) {
         if let date = date {
             post.dateCreated = date
+            if post.hasFuturePublishDate() {
+                post.status = .scheduled
+            } else {
+                post.status = .publish
+            }
         } else {
             post.publishImmediately()
         }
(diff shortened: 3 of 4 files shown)


ScoutHarris:
Hey @bjtitus . There's some possible oddities with 'Back Scheduled Draft'. 

---
- Start a new post.
- Go to publish date/time.
- Select the past.
- The Status says Published.

---
- Save a post as draft with past date.
- Go to Blog Posts > Drafts.
- On the draft, select Publish Now.
- On the post, go to Published and note the published date is the current date/time.

---
- Save a post as draft with past date.
- Go to Blog Posts > Drafts.
- Note the date displayed is the one selected in the past.
- Edit the draft, and go to Post Settings.
- Notice the date now says Immediately.

bjtitus:
I believe the first two are mainly choices with the current flow. I think there are a bunch of questions about what we should do in these cases now that we're taking a closer look.

### 1. Past date changes status to Published

Any selection from the Schedule calendar control sets the status to Published or Scheduled. Keeping the status as "Draft" would require you to manually select "Published" or "Scheduled" from the Post Settings screen. This was a conscious choice that seemed to mirror the previous functionality but could be changed to keep the Post status as whatever it was to begin with.

### 2. Publish Now

As far as I can tell, this is consistent with the previous logic (nothing has changed with that button's actions AFAIK). I interpret "Publish Now" to mean: change the date to "Now" and Publish.

### 3. "Publish Immediately" on a brand new + backdated post

Interestingly, this doesn't happen _after_ the post has been saved once and _then_ the date is changed. This appears to be a bug and requires some explanation about how "Immediately" actually works:

* Change publish date of the post -> `dateCreated` property is set on AbstractPost
* AbstractPost's `dateCreated` property is sent as `date` property to API (see https://developer.wordpress.org/rest-api/reference/posts/ for API reference)
* `date` property appears to be used to fill in the `modified` property on a brand new post
* "Publish Immediately" is shown whenever the `dateCreated` and `dateModified` properties are equal or `dateCreated` is `nil`.

There's no way to set a separate modified date, as it is read-only in the API. The only way I've figure out to handle this is to first upload the draft using the `dateModified` date and then update that same post with the proper `dateCreated` date for publishing.

I tested #3 on Build 14.1 (21544) with the Scheduled Feature Flag disabled and noticed the same exact behavior. Not sure if this was introduced when offline posting was added, or if it has always been like this.

ScoutHarris:
Hey @bjtitus . Thanks for the thorough explanations! Given that, I think this fulfills it's intent.

PR #13727, opened 2020-03-25
Refactor Comment action method signature to be compatible with Xcode 11.4

SergioEstevao:
This PR fixes the build problems when using Xcode 11.4 to build the project

To test:
 - Check the notification area of the app
 - Try to mark notification with the multiple actions available: like, spams, reply, etc..

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Notifications/Actions/ApproveComment.swift
+++ b/WordPress/Classes/Models/Notifications/Actions/ApproveComment.swift
@@ -18,8 +18,11 @@ class ApproveComment: DefaultNotificationActionCommand {
         return on ? .neutral(.shade30) : .primary
     }
 
-    override func execute<ObjectType: FormattableCommentContent>(context: ActionContext<ObjectType>) {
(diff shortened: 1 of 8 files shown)


bjtitus:
I made the same change yesterday to get this to compile, but I tested just now and it looks like the superclass method is called instead of these implementations now (so liking a notification doesn't work, for example).

An example of the override + cast in `LikeComment.swift`:

```swift
override func execute<ContentType: FormattableContent>(context: ActionContext<ContentType>) {
    if let block = context.block as? FormattableCommentContent {
        if on {
            removeLike(block: block)
        } else {
            like(block: block)
        }
    }
}
```

I think the protocol could potentially be made generic and have the class provide the ContentType as a type alias, but I didn't have time to dig into that more.

diegoreymendez:
Just noting I also saw the superclass's `execute` method called instead.

@bjtitus 's idea sounds like a good compromise until we can re-engineer this.

PR #13770, opened 2020-03-30
Add authentication support to image downloads for Atomic Private Sites.

### LINKED ISSUE #13571: Spike: make media downloads work for Atomic Private Sites.
We need to analyze throughout the code the places where we directly request media, and try to figure out how we can request the authentication cookie whenever needed for those requests.

This includes direct media downloads in the editors.

diegoreymendez:
Closes #13571 

This PR forces authentication for all our media downloads from WP sites.  In particular, this ensures we can make media requests to Atomic Private sites either directly or through Photon URLs (which are replaced by a proxy URL since Photon currently does not support Atomic Private Sites).

## Additional Info:

I had to change some calls to download images from `UIImageView+Networking.swift` to AlamofireImage.  The reason was that I was seeing a lot of issues that were NOT introduced by my code, but that were made more noticeable by it.

Please DM if you want to know the details, as the conversation around this happened in private.

As a side note, I'm not sure if it's just me but images are loading extremely fast after the switch, and the Apps feels a lot lighter and snappier.

## To test:

Try to make sure you test with a public site, a private site, an atomic private site, and a self-hosted site.

### Test 1:

1. Create a post with images from different sources (device media, WP library, etc).
2. Publish it.
3. Make sure the image looks fine.

### Test 2:

Open the media library for your  site and make sure it looks good.

### Test 3:

Set a site icon for your site and make sure it loads fine.

## PR submission checklist:


diegoreymendez:
Since this is such a huge PR I'll finish this up, and once I'm happy with the functionality, I'll try to start breaking it up into smaller PRs.

This will take a bit more time, but I hope to ensure that I can offer reasonable changes to review.  This PR changes some sensitive parts of our code that I want to avoid breaking.

Fyi @aerych 
cc @yaelirub

diegoreymendez:
This is the first PR spawned / broken-up from the changes here: https://github.com/wordpress-mobile/WordPress-iOS/pull/13813

WordPress/Classes/ViewRelated/Post/PrivateSiteURLProtocol.h
```diff
@@ -20,8 +20,4 @@
 + (void)registerPrivateSiteURLProtocol;
 + (void)unregisterPrivateSiteURLProtocol;
 
-+ (nonnull NSURLRequest *)requestForPrivateSiteFromURL:(nonnull NSURL *)url;
```
SergioEstevao, line 23: I think that we the removal of the uses of this methods we no longer need this class at all. 
You probably can remove it completely.

diegoreymendez: Agree!  Nice catch!

diegoreymendez:
I addressed some extra issues that @SergioEstevao reported via DM.  The reader was failing to show images for private sites due to some of the changes.  All is working now, and this is ready for another look.

SergioEstevao:
I followed the test plan with the following scenarios:

 - Public site hosted on WP.com
 - Private site hosted on WP.com
 - Self hosted site
 - Public Atomic site

I also did an extra test to check all of the above sites on the Reader.

All is working correctly! Great work @diegoreymendez

SergioEstevao:
Just found a small regression, when uploading images on the media library we used to see the following:

  - Local image with an overlay was displayed while the upload was being done
  - When the upload of the image finished, the local image was kept in place while the final image is downloading
 - When the remote image is available we show it

With this version when the upload finished we see a grey placeholder while the remote image is being dowloaded.

It's not a super big thing but it will be nice to have fixed.

diegoreymendez:
@SergioEstevao - This is  what I'm seeing in develop with a custom link conditioner configuration of:

Bandwidth: 1000 Kbps
Packet Loss: 0%
Delay: 200ms

It seems to me like the issue is present in develop as well?

uploadIssues

aerych:
> This is what I'm seeing in develop

Seeing the same in the current beta on a device fwiw.

diegoreymendez:
Thanks both!

PR #13772, opened 2020-03-30
Extract post visibility selection

leandroalonso:
This is a refactor to allow the Post Visibility selection View Controller to be used in other parts (Nudges View, I'm looking at you).

### To test

1. Create a new post
2. Go to Settings
3. Change the visibility

The post should be published with the visibility you specified.


WordPress/Classes/ViewRelated/Post/PostVisibilitySelectorViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    override init(style: UITableView.Style) {
+        super.init(style: style)
+    }
+}
+
+// MARK: - Abstract Post private extension
+
+private extension AbstractPost {
```
emilylaguna: There's duplicate logic located for this in: `PostSettingsViewController.m:1284`

Should we refactor that code to use this extension?

leandroalonso: Refactored in f2e47765f7ce42aff072afbccdc9a691fbf53962

PR #13809, opened 2020-04-03
Display Cancel button in "Other Apps" media browser

guarani:
## Description

Replaces existing approach with a simpler, more specific solution.
The new appearance selector only targets bar button items in document browser view controllers.
It also eliminates the need to reset the appearance of the items after media is chosen.

This fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2040

It also fixes this bug where after using the Site | Media library with the "Other Apps" source, subsequently opening a post shows blue – instead of white – navigation bar items (such as the "Publish" button).

## Steps to Test

On a device with light mode active:
1. Create a new post/page, and in the block editor add an image using the "Other Apps" source option
2. Ensure the Cancel button displays on the document browser modal sheet 
3. Repeat steps 1 and 2 for the Classic Editor
4. Go to site media (Media option under Publish section) on site screen and add media, again using the "Other Apps" source option
5. Again, ensure the Cancel button displays on the document browser modal sheet 
6. Toggle between the two tabs of the document picker, "Recents" and "Browse" – notice that the Cancel button does not lose its color

I have tested on both Light and Dark Mode (iOS 13.3.1) as well as iOS 12.4.5 (Light Mode).

## PR Submission Checklist


guarani:
I'm not sure if the color I used for the "Cancel" (and "Browse") buttons is correct – previously they were white and were hidden against the white navigation bar background.

iamthomasbishop:
I think the actions on the navigation bar should probably be our primary blue in light mode. @mattmiklic what do you think?

guarani:
> I'd recommend Blue 50 in light mode and Blue 30 in dark mode for these links.

Thanks @mattmiklic, I've update to use these colors.

PR #13817, opened 2020-04-06
Atomic Private Sites: add support for downloading the site icon.

diegoreymendez:
This PR adds atomic-private support to site icons.

This is the second part of a larger change (#13770).

## To test:

1. Create an Atomic Private Site following the instructions from p58i-8xH-p2
2. Remember to make sure the site is indeed, well, private 😄 
3. Set a site icon for the site by going into WP-Admin > Appearance > Customize > Site Identity
4. Make sure the site icon is shown in WPiOS.

## PR submission checklist:


WordPress/Classes/Models/Blog.m
```diff
@@ -331,10 +331,15 @@ - (NSString *)postFormatTextFromSlug:(NSString *)postFormatSlug
     return formatText;
 }
 
-// WP.COM private blog.
```
aerych, line 334: Would it make sense to add a comment explaining the use of `isPrivate` and how it differs from `isPrivateAtWPCom`?

WordPress/Classes/Models/Comment.m
```diff
@@ -102,8 +102,8 @@ - (NSDate *)dateCreated
 
 - (BOOL)isPrivateContent
 {
-    if ([self.post respondsToSelector:@selector(isPrivate)]) {
-        return (BOOL)[self.post performSelector:@selector(isPrivate)];
+    if ([self.post respondsToSelector:@selector(isPrivateAtWPCom)]) {
+        return (BOOL)[self.post performSelector:@selector(isPrivateAtWPCom)];
```
aerych, line 106: I've looked at BasePost, AbstractPost and ReaderPost, and ran `ack isPrivateAtWPCom`.  I'm not finding any where that post implements `isPrivateATWPCom`.   Was this supposed to be modified to check `self.post.blog` or, given the change to the AbstractPost's `isPrivate` method, should this change be rolled back?

WordPress/Classes/Models/AbstractPost.m
```diff
@@ -477,7 +477,7 @@ - (BOOL)supportsStats
 
 - (BOOL)isPrivate
 {
-    return self.blog.isPrivate;
+    return self.blog.isPrivateAtWPCom;
```
aerych, line 480: When this was a pass-thru call to the blog it was clear just by the method names what was intended. Now there is a bit of a naming mismatch and I'm wondering about future confusion.  Maybe add a comment to clarify things?

diegoreymendez:
@aerych - Thanks for the thorough review!  I'll try to clear the confusions and fix anything I missed around this.

WordPress/Classes/Models/Blog.m
diegoreymendez: I added some comments to clarify the purpose of both methods.

Btw, I think checking if something is hosted at WP.com is a code smell, and should prolly be replaced with `isAccessibleThroughWPCom`... but  it goes far beyond the scope of this PR.

WordPress/Classes/Models/Comment.m
diegoreymendez: I think this was a mistake in the auto-renaming tool in Xcode.  Nice catch!

WordPress/Classes/Models/AbstractPost.m
diegoreymendez: So many thoughts around this.  I'll share them with you so you can evaluate what I did here - please bear with me for a second since this isn't straightforward.

### Why I went for a rename:

At first I was using `blog.isPrivate`.   But I soon found out that it wasn't working at all for Jetpack sites - hence Atomic Private sites.

The reason is that (the original) `blog.isPrivate` was also checking that the blog was hosted at WPcom.

This was a code smell IMHO since the name `isPrivate` doesn't convey that additional check.

### The two alternatives:

So I needed to either change the logic in that method for the whole App (risking breakage), or renaming it, and creating a cleaner `isPrivate` method that only checks if a blog is private, regardless of where it's hosted.

I opted for the second solution, as it would ensure the rest of the App would still keep working as it has been so far, and I would still get a new cleaner method for Atomic Private  sites.

### A possible third alternative:

As I mentioned above, my hunch is that we probably shouldn't be checking if a blog is hosted in WPcom, but  rather if it's accessible through WPcom.

I didn't want to dive too deep into this though, as it's easy to get lost in the rabbit hole.

### Going back to your review!

So that's the background.

I agree with what you're saying here.  Maybe what I can do is rename this method too so that it's also `isPrivateAtWPCom`, and then create a cleaner `isPrivate` method that redirects to `blog.isPrivate`.

Thoughts?

WordPress/Classes/Models/Comment.m
aerych: Silly Xcode :)

WordPress/Classes/Models/AbstractPost.m
aerych: Thanks for the background! 
That works for me. It would avoid any confusion that way :)  I had started to suggest renaming but I was worried there would be a massive number of places that would need to be updated. 😬  
If it's a chore I think a comment would probably suffice. Up to you!

Podfile
```diff
@@ -34,12 +34,13 @@ end
...
 def wordpress_ui
     ## for production:
-    pod 'WordPressUI', '~> 1.5.2'
+    pod 'WordPressUI', '~> 1.5.3-beta.1'
 
     ## for development:
     #pod 'WordPressUI', :path => '../WordPressUI-iOS'
     ## while PR is in review:
     #pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :branch => ''
+    #pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :commit => '71f32a3300b4c630b41ba7ae8101896f9d297606'
```
jleandroperez, line 43: Since this is already commented out, do we need to push it?

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
```diff
@@ -61,22 +61,60 @@ extension UIImageView {
...
+                    }
+
+                    self.removePlaceholderBorder()
+            },
+                failure: { error -> () in
+                    if let error = error {
+                        CrashLogging.logError(error)
+                    }
+            })
+        }
```
jleandroperez, line 96: 100000% sure this one needs to be a nested function? 

Just asking for sanity, i've not seen many of these around. Actually, first one in years!

WordPress/Classes/Models/Comment.m
diegoreymendez: I rolled back this change as a result of the rename below.

WordPress/Classes/Models/AbstractPost.m
diegoreymendez: Done!

Podfile
diegoreymendez: This is for convenience during development only.  It allows quickly replacing the commit number and enabling this line rather than having to look up the command formatting.

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
diegoreymendez: Need no, but I wasn't sure I wanted this method accessible from elsewhere.  The code around this logic needs a lot of cleanup.

I'm moving the logic to another method to resolve this comment.

diegoreymendez:
@aerych - This is ready for another look.  Can you double check the Comment.m rollback I did, since now the method does exist?

WordPress/Classes/ViewRelated/Gutenberg/EditorMediaUtility.swift
```diff
@@ -62,7 +62,7 @@ class EditorMediaUtility {
 
         if url.isFileURL {
             request = URLRequest(url: url)
-        } else if post.blog.isPrivate() && PrivateSiteURLProtocol.urlGoes(toWPComSite: url) {
+        } else if post.blog.isPrivateAtWPCom() && PrivateSiteURLProtocol.urlGoes(toWPComSite: url) {
```
aerych, line 65: No need to reference blog anymore :)

WordPress/Classes/ViewRelated/Post/PostPreviewGenerator.swift
```diff
@@ -87,7 +87,7 @@ private extension PostPreviewGenerator {
         case .draft, .publishPrivate, .pending, .scheduled, .publish:
             return true
         default:
-            return post.blog.isPrivate()
+            return post.blog.isPrivateAtWPCom()
```
aerych, line 90: No need to reference blog anymore here either :)

PR #13821, opened 2020-04-06
Adapt the Bottom Sheet for Landscape

### LINKED ISSUE #13801: Adapt the Bottom Sheet for Landscape

leandroalonso:
Fixes #13801 

### To test

1. Open the Nudges View
2. Go landscape

Also, open the Nudges view while in Landscape and alternate between Portrait/Landscape.

### Technical detail

The landscape code comes from @bjtitus `BottomSheetPresentationController`, all credits to him. :)

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
+++ b/WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
@@ -133,6 +133,7 @@ public class DrawerPresentationController: FancyAlertPresentationController {
     override public func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         coordinator.animate(alongsideTransition: { _ in
             self.presentedView?.frame = self.frameOfPresentedViewInContainerView
+            self.transition(to: self.currentPosition)
         }, completion: nil)
         super.viewWillTransition(to: size, with: coordinator)
     }
--- a/WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
+++ b/WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
@@ -47,6 +47,8 @@ private enum Constants {
     static let flickVelocity: CGFloat = 300
     static let bounceAmount: CGFloat = 0.01
 
+    static let maxWidthPercentage: CGFloat = 0.66 /// Used to constrain the width to a smaller size (instead of full width) when sheet is too wide
+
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
```diff
@@ -21,13 +21,27 @@ public enum DrawerHeight {
...
+
 public protocol DrawerPresentable: AnyObject {
     /// The height of the drawer when it's in the expanded position
     var expandedHeight: DrawerHeight { get }
 
     /// The height of the drawer when it's in the collapsed position
     var collapsedHeight: DrawerHeight { get }
 
+    /// The width of the Drawer in compact screen
+    var width: DrawerWidth { get }
```
emilylaguna, line 43: I am worried about this being too vague and could cause conflicts. What about prefixing it? 

`drawerWidth` ?

WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
```diff
@@ -113,10 +132,36 @@ public class DrawerPresentationController: FancyAlertPresentationController {
...
+                width = givenWidth
+            case .maxWidth:
+                break
+            }
+        }
+
+        /// If we constrain the width, this centers the view by applying the appropriate insets based on width
+        let leftInset: CGFloat = ((containerView.bounds.width - width) / 2)
+
+        return CGRect(x: leftInset, y: frame.origin.y, width: width, height: frame.height)
```
emilylaguna, line 156: I don't think we need to create a new CGRect here. We can just modify the values of `frame` directly.

```
/// If we constrain the width, this centers the view by applying the appropriate insets based on width
frame.origin.x = ((containerView.bounds.width - width) / 2)
frame.size.width = width

return frame
```

WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
leandroalonso: Done in f666e7c1888220008100034ac53c287bc19c598e

leandroalonso:
@emilylaguna ready for another 👀

Issue #13580, opened 2020-03-02
Information Architecture: Reader Navigation - Tabs Navigation


Gio2018:
Related issues

#13322
#13581
#13582

Specific Reader topics are fetched from the backend, synched in Core Data and used to populate the tab bar. Topics are also used to feed ReaderStreamViewController, which will be displayed in the container view below the buttons.

Specific topics are

- defaults
- teams
- lists

<img width="240" alt="Screen Shot 2020-03-31 at 3 45 37 PM" src="https://user-images.githubusercontent.com/34376330/78073584-ce1bac80-7366-11ea-8e34-a00c2283f27f.png">

Design: https://wp.me/pauD4L-Dx

- [x] fetch reader topics and use them to feed the tab bar
- [x] hide button bar for topics different from 'Following'
- [x] add ReaderStreamViewController to the container view
- [x] use the selected tab (topic) to feed ReaderStreamViewController
- [x] remove the 'manage' header from the 'Following' screen
- [x] ensure the 'Discover' screen keeps its header
- [x] ensure the navigation from ReaderStreamViewController to its sub levels works the same way it used to be
- [x] add saved posts
- [x] handle state restoration
- [x] check that cleanup actions in `ReaderTabViewModel` are performed only when necessary
- [x] Handle issues in poor network conditions
- [x] ensure universal links still work as expected
- [x] ensure spotlight search still works as expected
- [x] ensure the views display correctly in dark mode
- [x] Remove site header when a site is fetched from a filter action
- [x] hide filter and settings button when the tab bar is still loading
- [x] Handle airplane mode/no connection scenario of existing cached content
- [x] handle airplane mode/no connection scenario of no cached content
- [x] add the empty 'Following' tab for the self hosted site experience
- [ ] I~~mprove the loading experience using ghost~~ will be addressed in a future release
- [x] add/improve accessibility
- [x] add unit tests
- [ ] ~~add/update UI tests~~ current unit tests are passing with the new Reader. There will be improvements and added tests in future releases

PR #13840, opened 2020-04-07
Adds Reader Stream to the new Reader

Gio2018:
Ref #13580

##### Notes:
- the 'Saved' tab is not active at the moment. Tapping on 'Saved' should not produce any effect, except hiding the 'Filter' and 'Settings' (cog) button if you previously were on the 'Following' tab
- the 'Filter' button is not functional at the moment. If tapped, it will visually simulate the behavior of a filter being set (showing 'Phoebe's Photos as an example), and the 'x' button will reset the filter, but this is only a visual mock.
- the 'Settings' (cog) button is not functional at the moment.
- At the moment, it is expected that some UI tests might fail. The feature flag will be set to false before merging to `develop`

readerTabs2

To test:
##### Reader content
- build/run and navigate to the 'Reader' tab
- verify that the UI shows the tab bar on top, the 'Filter' and 'Settings' button when on the 'Following' tab
- Verify that the tab bar topics contain:
  - The default topics (Following, Discover, Likes, Saved)
  - Any of the following that are associated with your account:
    - Lists
    - Teams (e.g. the 'Automattic' team should appear on any Automattic account)
- Verify that the followed sites stream is displayed in the content area below the tab bar and buttons.
- Verify that the navigation to post details works as expected.
- Verify that 'Discover' and 'Likes' behave as expected

##### Account change
- After executing the above steps, log out and log in with a different account
- Verify that the content is properly updated, as well as the tab bar topics
- Optionally, repeat any of the steps of the 'Reader Content' test section

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderFollowedSitesStreamHeader.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 import Gridicons
 import WordPressShared.WPStyleGuide
-
+// TODO: - READERNAV - This (and the related xib) might need to be removed once the new Reader Tab Navigation is stable
```
ScoutHarris, line 4: 👍

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabViewModel.swift
```diff
@@ -53,8 +63,89 @@ extension ReaderTabViewModel {
         service.fetchReaderMenu(success: { [weak self] in
             self?.fetchTabBarItems(completion: completion)
             }, failure: { error in
-                DDLogError("Error syncing menu: \(String(describing: error))")
+                DDLogError(ReaderTopics.remoteFetchError + "\(String(describing: error))")
                 completion(nil)
         })
     }
+
+    private enum ReaderTopics {
```
ScoutHarris, line 71: Nice.

Gio2018:
> There's a couple of different divider styles here. They should all be the same. I think divider style 1 (purple) is correct and what we're using elsewhere. Divider style 2 (orange) is appearing at the bottom of the filter bar, then becomes the divider below the tabs when the filter bar is gone. Nice job animating the filter bar btw.
> 
> So I think that new divider marked 2, should take the same style as 1.
> 
> And while we're there, divider 3, separating the cog, is also a slightly different color /style. Could it also match 1?
> 
Thanks for catching this!! There were two problems:
- Dividers style mismatch: I set all the divider widths to `.hairlineBorderWidth`. and color to .lightGray (as it already was on the tab bar)
- Double bottom divider when the buttons are hidden: I just forgot to hide the bottom divider too, I have now added that.

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -129,19 +134,33 @@ extension ReaderTabView {
         WPStyleGuide.applyReaderSettingsButtonStyle(settingsButton)
     }
 
-    private func setupContainerView() {
+    private func addContentToContainerView() {
+        guard let controller = self.next as? UIViewController,
+            let readetTabItem = tabBar.items[tabBar.selectedIndex] as? ReaderTabItem,
```
ScoutHarris, line 139: Looks like `readetTabItem` is misspelled.

WordPress/Classes/ViewRelated/System/FilterTabBar.swift
```diff
@@ -457,7 +456,7 @@ class FilterTabBar: UIControl {
 
     private enum AppearanceMetrics {
         static let height: CGFloat = 46.0
-        static let bottomDividerHeight: CGFloat = 0.5
+        static let bottomDividerHeight: CGFloat = .hairlineBorderWidth
```
bjtitus, line 459: 👍

Gio2018:
> Hey @Gio2018 . This is awesome! 🎉

Thanks!!!

> 
> Besides one minor code comment, I did notice these:
> 
> * Toggling the feature via the Debug menu didn't do anything until I relaunched the app.

This is a case when it would not be easy to toggle the feature via debug menu, so I disabled it in there for now

> * I'm not sure if Teams are functional yet, but `Automattic` showed no posts, but the old Reader did.
> 
There was a loading issue that I investigated with @bjtitus , it should work in normal network conditions. There will be a future issue and PR in the project that further investigates this issue in slow network conditions.

PR #13830, opened 2020-04-06
Implement Tenor client API

ngoctr:
This PR implements Tenor client API as part of #13803 

## Testing
**Tenor API Key credential generation:**

1. Update your `.configure-files/wpcom_app_credentials` by appending a line  `TENOR_API_KEY=YOUR_TENOR_API_KEY`
2. Press Cmd+B (Build your project)
3. Verify that `Derived Sources/ApiCredentials.m` has the method for retrieving YOUR_TENOR_API_KEY above
```
+ (NSString *)tenorApiKey {
    return @"YOUR_TENOR_API_KEY";
}
``` 
**Functional Testing:**
This PR doesn't include any end-user feature for testing

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #13803, opened 2020-04-02
Tenor Implementation Master Issue


ScoutHarris:
Implement Tenor to replace Giphy for third-party gif import.

Refs:
- Giphy [removal PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/12769).
- Tenor [documentation](https://tenor.com/gifapi/documentation#quickstart).

Tasks:
- [x] Add Feature Flag. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13829)
- [x] Implement Tenor support.
  - [x] Tenor API credentials support (https://github.com/wordpress-mobile/WordPress-iOS/pull/13842)
  - [x] Implement Tenor client api (https://github.com/wordpress-mobile/WordPress-iOS/pull/13843)
  - [x] Tenor picker (https://github.com/wordpress-mobile/WordPress-iOS/pull/13861)
- [x] Add Tenor to Media Library options. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13907)
- [x] Add Tenor to Aztec editor media options. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13908)
- [x] Add Tenor to block editor media options. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13910)
- [x] Add accessibility support. (no changes required)
- [x] Add Tracks. [Related P2](https://iosp2.wordpress.com/2020/03/16/adding-tracks-forget-wordpress-shared/) on the new way of doing this. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13923)
- [x] Animating GIFs in Tenor Picker (https://github.com/wordpress-mobile/WordPress-iOS/pull/13941)
- [x] Enable feature, add to release notes. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14297) (Released in 15.1)
- [x] Mobile Guild P2 post. (https://wp.me/p7cLQ7-Ze)
- [x] Remove any remaining Giphy support/files. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14302)
- [ ] Remove Feature Flag.

frosty:
Some remaining issues I observed while testing the main PR:

* Issue in the block editor where I saw search results appear briefly and then vanish. For some reason it happens for me every time with the search term 'blathers'.
* There's no animation or indication of a gif in the block editor – the classic editor shows a GIF badge in the top left of an image.
[...]

PR #13842, opened 2020-04-08
Tenor API credentials support

ngoctr:
This PR is separate one from #13830 to address suggestions from @frosty, This is a part of #13803

## Testing
**Tenor API Key credential generation:**

1. Update your `.configure-files/wpcom_app_credentials` by appending a line  `TENOR_API_KEY=YOUR_TENOR_API_KEY`
2. Press Cmd+B (Build project)
3. Verify that `Derived Sources/ApiCredentials.m` has the method for retrieving YOUR_TENOR_API_KEY above
```
+ (NSString *)tenorApiKey {
    return @"YOUR_TENOR_API_KEY";
}
``` 
**Functional Testing:**
This PR doesn't include any end-user feature for testing

PR submission checklist:


WordPress/Credentials/gencredentials.rb
```diff
@@ -242,4 +254,4 @@ def print_class(client, secret, sentry, appcenter, giphy, google_client, google_
   end
 end
 
-print_class(client, secret, sentry, appcenter, giphy, google_client, google_scheme, google_login_server, debugging_key, zendesk_app_id, zendesk_url, zendesk_client_id)
+print_class(client, secret, sentry, appcenter, giphy, google_client, google_scheme, google_login_server, debugging_key, zendesk_app_id, zendesk_url, zendesk_client_id, tenor_key)
```
ScoutHarris, line 257: I think it would be helpful to add a check for `tenor_key` and print an error message like we do for the other credentials. For example:

```
  if google_client.nil?
    $stderr.puts "warning: Google Login Client ID not found"
  end
```

ngoctr: Thanks for spotting this @ScoutHarris. I have added the check for Tenor part

PR #13847, opened 2020-04-08
Changes the auth-proxy path.

### LINKED ISSUE #13811: Change path for the Atomic Proxy service.
@adamziel mentioned that the Atomic Proxy endpoint for Media requests should now be called with a query parameter (`path`), in order to resolve a caching issue when having the media path as part of  the endpoint path.

In other words we need to replicate the changes done in this PR: https://github.com/Automattic/wp-calypso/pull/40751

So instead of using the path:

`/sites/${ siteIdOrSlug }/atomic-auth-proxy/file${ mediaPath }`

We should use the path parameter like below:

`/sites/${ siteIdOrSlug }/atomic-auth-proxy/file?path=${ mediaPath }`

diegoreymendez:
Fixes #13811 

The old proxy endpoint was:

`/sites/${ siteIdOrSlug }/atomic-auth-proxy/file${ mediaPath }`

The new one is:

`/sites/${ siteIdOrSlug }/atomic-auth-proxy/file?path=${ mediaPath }`

## To Test:

1. Create an Atomic Private Site following the instructions from p58i-8xH-p2
2. Remember to make sure the site is indeed, well, private 😄 
3. Publish a post with an image through the web, or set the site's icon.
4. Before running WPiOS place a breakpoint here: https://github.com/wordpress-mobile/WordPress-iOS/blob/665162286a4bd8a518934d8a90d511f37fe7b1d4/WordPress/Classes/Networking/MediaRequestAuthenticator.swift#L229
5. Run the App and make sure that breakpoint is hit for your image.  Check that the URL matches the new endpoint path described above.
6. Make sure the image is shown.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Networking/MediaRequestAuthenticator.swift
+++ b/WordPress/Classes/Networking/MediaRequestAuthenticator.swift
@@ -214,11 +214,12 @@ class MediaRequestAuthenticator {
             return
         }
 
-        let contentPath = components.path[wpContentRange.lowerBound ..< components.path.endIndex]
+        let contentPath = String(components.path[wpContentRange.lowerBound ..< components.path.endIndex])
 
         components.scheme = secureHttpScheme
         components.host = wpComApiHost
-        components.path = "/wpcom/v2/sites/\(siteID)/atomic-auth-proxy/file\(contentPath)"
+        components.path = "/wpcom/v2/sites/\(siteID)/atomic-auth-proxy/file"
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Hey folks, since this is quite a small PR I'm merging... but if you still want to add a review, be my guest!  I'll make sure to address any concerns.

diegoreymendez:
Thanks @jleandroperez !

PR #13851, opened 2020-04-08
Prepublishing Nudges Accessibility Improvements

### LINKED ISSUE #13849: Make sure sheet has accessibility

leandroalonso:
Fixes #13849

This PR:

- Makes sure the Voice Over focus move to Prepublishing Nudges when it appears
- That all the options are accessible and the user can return to the main screen
- That the bottom sheet is dismissable

### To test

You will need a real device.

1. Enable Voice Over
2. Create a new post
3. Tap Publish
4. Navigate through the Prepublishing Nudges and try to find a dead-end

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/Prepublishing Nudge/PrepublishingHeaderView.swift
+++ b/WordPress/Classes/ViewRelated/Post/Prepublishing Nudge/PrepublishingHeaderView.swift
@@ -24,17 +24,10 @@ class PrepublishingHeaderView: UIView, NibLoadable {
 
     // MARK: - Close button
 
-    func hideCloseButton() {
-        closeButtonView.layer.opacity = 0
-        closeButtonView.isHidden = true
-        leadingConstraint.constant = Constants.leftRightInset
-        layoutIfNeeded()
-    }
-
-    func showCloseButton() {
-        closeButtonView.layer.opacity = 1
-        closeButtonView.isHidden = false
-        leadingConstraint.constant = 0
+    func toggleCloseButton(visible: Bool) {
+        closeButtonView.layer.opacity = visible ? 1 : 0
+        closeButtonView.isHidden = visible ? false : true
+        leadingConstraint.constant = visible ? 0 : Constants.leftRightInset
         layoutIfNeeded()
     }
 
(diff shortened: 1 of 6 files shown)


WordPress/Classes/ViewRelated/Post/Prepublishing Nudge/PrepublishingHeaderView.swift
```diff
@@ -53,8 +53,12 @@ class PrepublishingHeaderView: UIView, NibLoadable {
     }
 
     private func configureBackButton() {
-        backButtonView.isHidden = true
-        backButton.setImage(.gridicon(.chevronLeft, size: Constants.backButtonSize), for: .normal)
+        closeButtonView.isHidden = true
+        closeButton.setImage(.gridicon(.cross, size: Constants.backButtonSize), for: .normal)
+        closeButton.accessibilityHint = Constants.doubleTapToDismiss
```
emilylaguna, line 58: When I highlight the button using voice over it just says "Button", should we also set the `accessibilityLabel` to "Close" so when highlighted it'll read as "Close, button"

WordPress/Classes/ViewRelated/Post/Prepublishing Nudge/PrepublishingHeaderView.swift
```diff
@@ -22,24 +22,24 @@ class PrepublishingHeaderView: UIView, NibLoadable {
         blogTitleLabel.text = blog.title
     }
 
-    // MARK: - Back button
+    // MARK: - Close button
 
-    func hideBackButton() {
-        backButtonView.layer.opacity = 0
-        backButtonView.isHidden = true
+    func hideCloseButton() {
```
emilylaguna, line 27: Is there a reason for 2 different methods versus just using 1 that takes a `hidden` argument?

WordPress/Classes/ViewRelated/Post/Prepublishing Nudge/PrepublishingHeaderView.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Post/Prepublishing Nudge/PrepublishingHeaderView.swift
leandroalonso: Done!

PR #13832, opened 2020-04-07
Tenor picker support for Media & Aztec editor

ngoctr:
This PR implements Tenor GIF picker for Media Library options as well as Aztec editor media options as part of #13803

# Testing
**First steps**
1. Register your Tenor developer API here: https://tenor.com/gifapi
2. Check out this branch
3. Put your API Key into `.configure-files/wpcom_app_credentials` (TENOR_API_KEY line)
4. Run `rake dependencies`
5. Build and run

**Verify Tenor support in Media Library**
1. Go to Site -> Media
2. Tap the + button at the top bar, then choose Tenor 
3. Type your keyword, select your images then tap Add to add them into Media Library 
4. Verify that the selected image(s) have been uploaded 

**Verify Tenor support in Aztec Post Editor**

1. Go to Site -> Posts
2. Open the Post and switch to Classic Editor if needed
3. Tap the Editor to activate edit mode
4. Tap the ellipsis button at the media toolbar, then select Tenor  
5. Search and select your images, then tap Add button at the bottom right, verify that the selected image(s) have been added into the Post while being uploaded to Media library 

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [] I have considered adding accessibility improvements for my changes.
- [] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #13861, opened 2020-04-09
Implement Tenor picker

ngoctr:
Thi PR implements the Tenor Picker as part of #13803. It is split from #13832 in an effort to break (pretty) large PR into small ones. Other PRs for Tenor support for Media Library, Aztec/Block Editors will be submitted after this one is reviewed and merged @ScoutHarris.

**Testing**
This PR doesn't include any UI-related functionality. It just needs code review.

PR submission checklist:


WordPress/Classes/ViewRelated/Media/Tenor/TenorMedia.swift
```diff
@@ -0,0 +1,123 @@
...
+    convenience init?(tenorGIF gif: TenorGIF) {
+        let largeGif = gif.media.first { $0.gif != nil }?.gif
+        let previewGif = gif.media.first { $0.tinyGIF != nil }?.tinyGIF
+        let thumbnailGif = gif.media.first { $0.nanoGIF != nil }?.nanoGIF
+
+        guard largeGif != nil, previewGif != nil, thumbnailGif != nil else {
+            return nil
+        }
+
+        let images = TenorImageCollection(largeURL: largeGif!.url,
```
ScoutHarris, line 41: I know you've checked that these gifs are not `nil`, however we try to avoid force unwrapped whenever possible.

WordPress/Classes/ViewRelated/Media/Tenor/TenorMedia.swift
```diff
@@ -0,0 +1,123 @@
...
+        self.updatedDate = date ?? Date()
+        self.images = images
+    }
+}
+
+// MARK: - Create Tenor media from API GIF Entity
+
+extension TenorMedia {
+    convenience init?(tenorGIF gif: TenorGIF) {
+        let largeGif = gif.media.first { $0.gif != nil }?.gif
```
ScoutHarris, line 33: I could be misunderstanding what's happening here, but it seems like there's (probably a slim) chance the 3 gifs could be different. For each gif, you're getting the first media where that gif exists. It seems to me we should get the first media (or at least, the same one), then check the 3 exist on that one.

WordPress/Classes/ViewRelated/Media/Tenor/TenorPageable.swift
```diff
@@ -0,0 +1,41 @@
+
+struct TenorPageable: Pageable {
+    let itemsPerPage: Int
+    let position: String?
+    let currentPageIndex: Int
+
+    static let defaultPageSize = 36 // 6 rows x 6 columns on iPad Landscape, up to 9 rows x 4 cols on iPhone
```
ScoutHarris, line 7: > 6 rows x 6 columns on iPad Landscape, up to 9 rows x 4 cols on iPhone

Is this always true? I'm wondering why this is `36`, but Giphy and Stock Photos are `40`.

WordPress/Classes/ViewRelated/Media/Tenor/TenorPicker.swift
```diff
@@ -0,0 +1,141 @@
...
+
+        observeDataSource()
+    }
+
+    private func observeDataSource() {
+        observerToken = dataSource.registerChangeObserverBlock { [weak self] _, _, _, _, _ in
+            self?.updateHintView()
+        }
+        dataSource.onStartLoading = { [weak self] in
+            NoResultsTenorConfiguration.configureAsLoading(self!.searchHint)
```
ScoutHarris, line 65: Please remove this force unwrap.

WordPress/Classes/ViewRelated/Media/Tenor/TenorMedia.swift
```diff
@@ -0,0 +1,123 @@
...
+        guard largeGif != nil, previewGif != nil, thumbnailGif != nil else {
+            return nil
+        }
+
+        let images = TenorImageCollection(largeURL: largeGif!.url,
+                                          previewURL: previewGif!.url,
+                                          staticThumbnailURL: thumbnailGif!.url,
+                                          largeSize: largeGif!.mediaSize)
+
+        self.init(id: gif.id, name: gif.title ?? "", caption: "", images: images, date: gif.created)
```
ScoutHarris, line 46: It appears there is never a Tenor `caption`. Should there be? If not, can this property be removed?

WordPress/Classes/ViewRelated/Media/Tenor/TenorService.swift
```diff
@@ -0,0 +1,42 @@
...
+
+class TenorService {
+    static let tenor: TenorClient = {
+        TenorClient.configure(apiKey: ApiCredentials.tenorApiKey())
+        return TenorClient.shared
+    }()
+
+    func search(params: TenorSearchParams, completion: @escaping (TenorResultsPage) -> Void) {
+        let tenorPageable = params.pageable as? TenorPageable
+        let currentPageIndex = tenorPageable != nil ? tenorPageable!.pageIndex : 0
```
ScoutHarris, line 22: Please remove this force unwrap.

WordPress/Classes/ViewRelated/Media/Tenor/TenorPageable.swift
ngoctr: Hi @ScoutHarris thanks for the review. I chose 36 because that will fit for the first visible page of the picker on both iPad (10.5″) (I didn't take the iPad 12" into consideration) and on any iPhone. Please take a look at screenshots posted here https://wp.me/pbMoDN-hD.

ScoutHarris: > I didn't take the iPad 12" into consideration

I'm curious as to why not? We do support this device. And if you do, (if I understand correctly) wouldn't that explain the other two pickers being 40, and why then Tenor should as well?

WordPress/Classes/ViewRelated/Media/Tenor/TenorMedia.swift
ngoctr: Thanks @ScoutHarris, removed it ;-)

WordPress/Classes/ViewRelated/Media/Tenor/TenorService.swift
ngoctr: Done! Thanks @ScoutHarris

WordPress/Classes/ViewRelated/Media/Tenor/TenorPicker.swift
ngoctr: Done. Thanks @ScoutHarris

WordPress/Classes/ViewRelated/Media/Tenor/TenorMedia.swift
ngoctr: Thanks for the guideline @ScoutHarris. Fixed it!

WordPress/Classes/ViewRelated/Media/Tenor/TenorMedia.swift
ngoctr: Yeah, that looks confusing, however, as the TenorAPI returns a `TenorGIF` which contains an array of `TenorGIFCollection`, and in each `TenorGIFCollection` there will be only one key has value. To avoid various `TenorGIFTinyMedia`, `TenorGIFMediumMedia`, etc. I chose to put all of `tiny`, `medium` as optional values in `TenorGIFCollection` so the code is much shorter.

WordPress/Classes/ViewRelated/Media/Tenor/TenorPageable.swift
ngoctr: Well in iPad 12.9" portrait, the first visible page take 8 rows x 6 cols = 48 items. I didn't take it into consideration as I owned an analytics data where there is only 2% out of 3M devices are iPad 12". That couldn't be the same here though, but with 40 or 36, on the Pro two requests will be issued any way (the Tenor also has a maximum 50 images per page). Any way, I can put it back to 40 as other pickers, what do you think?

ScoutHarris: Thanks for the explanation @ngoctr .

We don't support specific _devices_ per se, but iOS versions. Basically, if it can run iOS 11+, then we support it. 

Regarding which default to use (40 or 48) I'll leave that up to you, whichever works best for Tenor. We don't have to match the others if that's not what's best. I suspect it will be more obvious from my perspective when the UI is in place and testable.

ngoctr: Thank you very much @ScoutHarris. I think it would be (better) consistent, to keep it as 40 as other pickers. I agree that we don't have any UI in this PR so it is really not obvious to evaluate.

Issue #13581, opened 2020-03-02
Information Architecture: Reader Navigation - Filter


Gio2018:
Related issues

#13322
#13580
#13582

Design: https://wp.me/pauD4L-Dx

PR #13871, opened 2020-04-10
13581: Reader Filter Sheet

bjtitus:
#13581 This adds the Filter Sheet control for filtering the Reader stream.

### Testing

- Open to the Reader screen and select the "Filter" button

### Known Issues

* No Empty Sites or Tags views. I'll add these in a follow up PR.
* No Unit tests. I will also add these in a follow up PR (maybe the same with the empty views)

### UX Questions

@osullivanchris I have a few questions about some things I noticed that we might want to add/change:

- I think we may want to update the designs to show the site's icon to match what the current app is doing. I can do this in a follow up PR.
- Should we show a checkmark next to the currently selected site / tag in the filter table?
- On Landscape iPhone, the bottom sheet starts out "collapsed" (half way down the screen) this doesn't leave much room for scrolling (see the first iPhone landscape screenshot below). Should we transition to an expanded height in that orientation?

### Summary

A bottom sheet modal appears half way up the screen on iPhone and can be swiped up to enlarge:

<a href="https://user-images.githubusercontent.com/3250/79028384-08e2c900-7b4d-11ea-9aa6-f23233495701.gif"></a>

On iPad, a popover is used:

<a href="https://user-images.githubusercontent.com/3250/79028389-0c765000-7b4d-11ea-8bc4-eb73c2311773.gif"></a>

<details>
  <summary>Screenshots</summary>
Screen Shot 2020-04-10 at 4 58 36 PM
Screen Shot 2020-04-10 at 4 58 44 PM
Screen Shot 2020-04-10 at 4 52 16 PM
</details>

PR submission checklist:

- [x] ~I have considered adding unit tests where possible.~ These will be added in a follow up PR
- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ Unreleased

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
@@ -29,6 +29,11 @@ class FilterTableViewDataSource: NSObject, UITableViewDataSource {
 
 class SiteTableViewCell: UITableViewCell, GhostableView {
 
+    enum Constants {
+        static let textLabelCharacterWidth = 40 // Number of characters in text label
+        static let detailLabelCharacterWidth = 80 // Number of characters in detail label
+    }
+
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: .subtitle, reuseIdentifier: reuseIdentifier)
         detailTextLabel?.textColor = UIColor.systemGray
@@ -42,9 +47,9 @@ class SiteTableViewCell: UITableViewCell, GhostableView {
         contentView.subviews.forEach { view in
             view.isGhostableDisabled = true
         }
-        textLabel?.text = "TEST LABEL Text Label"
+        textLabel?.text = String(repeating: " ", count: Constants.textLabelCharacterWidth)
         textLabel?.isGhostableDisabled = false
-        detailTextLabel?.text = "TEST LABEL sdflsjlwe lsdfjsldjsl sidjflsidj"
+        detailTextLabel?.text = String(repeating: " ", count: Constants.detailLabelCharacterWidth)
         detailTextLabel?.isGhostableDisabled = false
     }
 }
--- a/WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
@@ -1,0 +1,50 @@
+struct TableDataItem {
+    let topic: ReaderAbstractTopic
+    let configure: (UITableViewCell) -> Void
+}
+
+class FilterTableViewDataSource: NSObject, UITableViewDataSource {
+
+    let data: [TableDataItem]
+    private let reuseIdentifier: String
+
+    init(data: [TableDataItem], reuseIdentifier: String) {
+        self.data = data
+        self.reuseIdentifier = reuseIdentifier
+    }
+
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return data.count
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let item = data[indexPath.row]
+        let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath)
+
+        item.configure(cell)
+
+        return cell
+    }
+}
+
+class SiteTableViewCell: UITableViewCell, GhostableView {
+
+    override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: .subtitle, reuseIdentifier: reuseIdentifier)
+        detailTextLabel?.textColor = UIColor.systemGray
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func ghostAnimationWillStart() {
+        contentView.subviews.forEach { view in
+            view.isGhostableDisabled = true
+        }
+        textLabel?.text = "TEST LABEL Text Label"
+        textLabel?.isGhostableDisabled = false
+        detailTextLabel?.text = "TEST LABEL sdflsjlwe lsdfjsldjsl sidjflsidj"
+        detailTextLabel?.isGhostableDisabled = false
+    }
+}
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -208,7 +208,8 @@ extension ReaderTabView {
 
     /// Filter button
     @objc private func didTapFilterButton() {
-        viewModel.presentFilter(from: filterButton) { [weak self] title in
+        /// Present from the image view to align to the left hand side
+        viewModel.presentFilter(from: filterButton.imageView ?? filterButton) { [weak self] title in
             if let title = title {
                 self?.resetFilterButton.isHidden = false
                 self?.setFilterButtonTitle(title)
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -187,11 +187,10 @@ extension ReaderTabView {
 
     /// Filter button
     @objc private func didTapFilterButton() {
-        viewModel.presentFilter(from: filterButton) { [weak self] topic in
-            if let topic = topic {
+        viewModel.presentFilter(from: filterButton) { [weak self] title in
+            if let title = title {
                 self?.resetFilterButton.isHidden = false
-                self?.setFilterButtonTitle(topic.title)
-                self?.viewModel.tabSelectionCallback?(topic)
+                self?.setFilterButtonTitle(title)
             }
         }
     }
@@ -207,7 +206,6 @@ extension ReaderTabView {
         viewModel.presentSettings()
     }
 }
-
 
 // MARK: - Appearance
 extension ReaderTabView {
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -187,10 +187,6 @@ extension ReaderTabView {
 
     /// Filter button
     @objc private func didTapFilterButton() {
-        //TODO: - READERNAV - Remove. This test code is for UI prototyping only
-        guard filterButton.titleLabel?.text == "Filter" else {
-            return
-        }
         viewModel.presentFilter(from: filterButton) { [weak self] topic in
             if let topic = topic {
                 self?.resetFilterButton.isHidden = false
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -97,8 +97,6 @@ extension ReaderTabView {
         buttonsStackView.alignment = .fill
         buttonsStackView.addArrangedSubview(filterButton)
         buttonsStackView.addArrangedSubview(resetFilterButton)
-        let spacer = UIView()
-        buttonsStackView.addArrangedSubview(spacer)
         buttonsStackView.addArrangedSubview(verticalDivider)
         buttonsStackView.addArrangedSubview(settingsButton)
     }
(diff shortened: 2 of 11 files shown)


WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -191,16 +191,20 @@ extension ReaderTabView {
         guard filterButton.titleLabel?.text == "Filter" else {
             return
         }
-        setFilterButtonTitle("Phoebe's Photos")
-        resetFilterButton.isHidden = false
-        viewModel.presentFilter()
+        viewModel.presentFilter() { [weak self] topic in
+            if let topic = topic {
```
Gio2018, line 195: would it be possible to leave the view agnostic of any model (in this case ReaderAbstractTopic)? More details in the next comment

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabViewController.swift
```diff
@@ -7,6 +7,15 @@ class ReaderTabViewController: UIViewController {
     init(viewModel: ReaderTabViewModel) {
         self.viewModel = viewModel
         super.init(nibName: nil, bundle: nil)
+        self.viewModel.filterTapped = { [weak self] completion in
```
Gio2018, line 10: can we possibly move the creation of specific instances of `FilterSheetViewController` and `BottomSheetViewController` out of `ReaderTabViewController`?
It would be preferable for two reasons:
- it would be more consistent with the 'MVVM-Like' architecture we are using
- it would leave `ReaderTabViewController` unaware of any model and child viewController. If we confine these in `ReaderTabViewModel` it maybe easier to evolve.

if we feel that `ReaderTabViewModel` grows too much, we can also break it up

Here is a possible implementation; does not have to be like this, it's just an idea

[bottomSheet.patch.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4463966/bottomSheet.patch.zip)

osullivanchris:
Hi @bjtitus thanks for the ping, this is looking great. First, re; your questions

> I think we may want to update the designs to show the site's icon to match what the current app is doing. I can do this in a follow up PR.

Hmm I can't remember why we didn't do this on Android now. Where do you mean on the current app that we're doing it? I think it was maybe just a case that it was a little heavy visually and needed bigger rows but honestly can't remember right now.

> Should we show a checkmark next to the currently selected site / tag in the filter table?

No lets leave it out. I originally had this in early designs. But it added some complexity and ambiguity in some cases. For example, if you tapped the selected item again it wouldn't be clear if that would re-select or unselect it, and it became a kinda question of is it a radio button or is it a check mark, and is there multiple selection, etc. That's why I went away from it, kept the sheet simple and have the clear action in the filter bar. Let me know if any more info needed.

> On Landscape iPhone, the bottom sheet starts out "collapsed" (half way down the screen) this doesn't leave much room for scrolling (see the first iPhone landscape screenshot below). Should we transition to an expanded height in that orientation?

Yeah agree this looks better! (FYI I'm not seeing this above, guess you updated the PR. But I saw the screenshots in the email notification I got re; this PR)


Thanks @Gio2018 for the issues you brought up too, agree on all those.

1 more issue for me - in the latest mocks I didn't have dividers on the rows. I know they are standard on iOS table rows, but I think they look a little heavy in the context of a sheet. How hard is it to hide the dividers?

Lastly I tried to download [the build mentioned above](https://github.com/wordpress-mobile/WordPress-iOS/pull/13871#issuecomment-612260109) to review on device and look for anything else, but for some reason I'm getting the current reader experience when I install it.

bjtitus:
@Gio2018 Thanks for taking a look!

> if I set a filter, then navigate to another tab, then go back to 'following', the filter title stays on, but the content is unfiltered

Should be fixed now!

> Tapping on the filter button after setting a filter, does not reset the filter/present the bottom sheet

Oops, I forgot to remove the TODO code from the prototype. Should be fixed now.

> when applying a 'Tag' as a 'Filter', this would not just filter among the followed sites, but present all posts with that specific tag

I'm not sure I follow. I'm able to see posts from sites that I do not follow when filtering by a tag. Are you seeing different behavior?

I didn't love the idea of assuming things about the responder to chain to track down the view controller so my changes have a bit more indirection, which I don't really love. It definitely does isolate the logic more in the view model, though. Let me know what you think.

bjtitus:
@osullivanchris 

> in the latest mocks I didn't have dividers on the rows. I know they are standard on iOS table rows, but I think they look a little heavy in the context of a sheet. How hard is it to hide the dividers?

Not hard at all. I just pushed a change to do this.

> I tried to download the build mentioned above to review on device and look for anything else, but for some reason I'm getting the current reader experience when I install it.

I had forgotten to enable to tag on this branch. Sorry about that! I'll get a new build made.

Gio2018:
> > when applying a 'Tag' as a 'Filter', this would not just filter among the followed sites, but present all posts with that specific tag
> 
> I'm not sure I follow. I'm able to see posts from sites that I do not follow when filtering by a tag. Are you seeing different behavior?
> 

To clarify, the posts that I am seeing when I select a tag, come from sites that I follow, but also from sites that I don't follow. 

Strictly speaking, a '_filter_' function applied to the '_Following_' section should _filter_ results from that section, thus it should only show posts from sites that I follow. Again this behavior is consistent with the previous experience, and I don't think it should be addressed in this PR, but maybe we want to consider this.

> I didn't love the idea of assuming things about the responder to chain to track down the view controller so my changes have a bit more indirection, which I don't really love. It definitely does isolate the logic more in the view model, though. Let me know what you think.

Following up our discussion on slack, I think this implementation is fine; we might want to address a refactor of `ReaderTabView` and `ReaderTabViewController` in a separate PR.

bjtitus:
@osullivanchris version 14.6 (27037) contains the latest changes.

osullivanchris:
> Strictly speaking, a 'filter' function applied to the 'Following' section should filter results from that section, thus it should only show posts from sites that I follow. Again this behavior is consistent with the previous experience, and I don't think it should be addressed in this PR, but maybe we want to consider this.

Hey @Gio2018 ! Fully understand this point and logically this makes sense. Going back to where we started with this: it would ideally be a feed of _everything_ you follow (sites or tags). However right now the volume of content from tags would be too large, so the content from sites would be drowned out.

That's why we started with showing content from Followed Sites only. I was using that label in the filter bar for a while to indicate that as the current selection. However, users didn't realise in user testing that it was actionable. so the word 'Filter', though not perfect logically, was a better CTA.

Anyway in the medium-long term, we will hopefully actually show some posts from tags you follow within the feed. @mbshakti is working on this with the Ravenclaw team. Then logically things will be more in the right order. Default is a list of all things you follow, filter to narrow. But still it won't be the sum of _all_ tagged content initially. So I don't know if we'll ever get to a perfect logical solution on it. But from a UX point of view, we'll hopefully be showing users the type of content they wanted, with the ability to fine-tune/drill down to specific stuff if they want it! Maybe there's another word for that than "filter"

Hope that makes sense!

osullivanchris:
@bjtitus thanks I was able to download the build now! A few minor things I spotted:

- When I tap filter, the loading state of the list shows the dividers, although once they load in, the dividers are gone (thanks!)
- I'm seeing too much space above the first list item in 'sites' (screenshot below)
- When I expand the sheet, only dragging on the very top of it can hide/collapse it again. I think it would be better if dragging anywhere on the sheet when at the top of the list disappears it. So if you have scrolled lower on the list, the first scroll would bring you to the top of the list, the second would collapse the sheet. If this is unclear I can demo
- The 'Filter' touch target seems to only be the text/icon. I think it should be the whole row up until the divider for the settings cog icon (again can show if unclear)
- One for @Gio2018 to continue a previous conversation. The divider issue discussed previously now looks perfect, but on Dark Mode still looks incorrect

Should I also review this on iPad or are we just concerned with iPhone for now? Thanks!

bjtitus:
@osullivanchris Please do look at iPad when you get a chance!

I’m taking a look at these issues now.

bjtitus:
> When I tap filter, the loading state of the list shows the dividers, although once they load in, the dividers are gone (thanks!)

Fixed 👍

> I'm seeing too much space above the first list item in 'sites' (screenshot below)

> When I expand the sheet, only dragging on the very top of it can hide/collapse it again. I think it would be better if dragging anywhere on the sheet when at the top of the list disappears it.

Yep, this is an issue with the unsupported scroll -> pan feature mentioned in the Bottom Sheet post (https://wp.me/paNNhX-5K). I'll add an item to implement this before shipping.

> The 'Filter' touch target seems to only be the text/icon. I think it should be the whole row up until the divider for the settings cog icon (again can show if unclear)

Can you describe what we should do with the reset button? For now, I've extended the height of the filter button to take up the entire row to make it a bit easier to hit, but I'm not sure what should happen when the reset button is shown if we make the entire row tappable to change the filter.

osullivanchris:
> Can you explain how you ended up seeing this? I'm having trouble reproducing that. It could be that you're following a site without a title somehow?

Ah good guess! When I tapped the empty area, I got this. Which if I'm following a site with no title, no url and no posts seems fair. I didn't realise it was possible for a site to have no title or url though, I don't know what the site is (For that reason!)

IMG_7808


> Yep, this is an issue with the unsupported scroll -> pan feature mentioned in the Bottom Sheet post (https://wp.me/paNNhX-5K). I'll add an item to implement this before shipping.

Ah ok, thanks for the clarification!

> Can you describe what we should do with the reset button? For now, I've extended the height of the filter button to take up the entire row to make it a bit easier to hit, but I'm not sure what should happen when the reset button is shown if we make the entire row tappable to change the filter.

Oh good point! Thats true. Here's what I had in my head for unselected and selected state touch targets. But I hadn't put 2+2 together and realised that it means the touch target is smaller in the selected state.

Screenshot 2020-04-15 at 12 18 10

We could do the layout this way for consistent touch targets, but the reason I kept the x right next to the name is I thought the cog icon was meaning the x didn't stand out so much over on the right

Screenshot 2020-04-15 at 12 19 27

What do you think?

> @osullivanchris Please do look at iPad when you get a chance!

tried to do this but I haven't added the iPad yet for testing, and nothing happens when I tap '+ Add new device' from my iPad. Any idea how to get it up and running?

Thanks again

WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func ghostAnimationWillStart() {
+        contentView.subviews.forEach { view in
+            view.isGhostableDisabled = true
+        }
+        textLabel?.text = "TEST LABEL Text Label"
```
frosty, line 45: Should we remove these test titles from this PR?

osullivanchris:
> Probably one for @osullivanchris – having spent a bit of time with it, I'm not sure if the Filter title is a little too prominent? Should it be either grey like the other tab bar items, or blue like the selected items / links in the Reader?

I've been having a nagging feeling on this too @frosty , but when it was the same grey as the tabs it looked too weak. 

I just revisited the mocks and realised I was using 'medium' but since learned in the build we use semibold - which is a bit chunkier than medium. I took inspiration from the Stats tab but not sure such a heavy treatment is appropriate here.

What if we changed the font style to regular per the mock on the right?

Screenshot 2020-04-15 at 16 00 16

bjtitus:
> The scroll bar seems to be acting as though it has the full height of the expanded bottom sheet, even when compact. This means when you get a certain way down the list it scrolls off the bottom of the screen

I think this should be resolved with that pan on scroll change.

> We could do the layout this way for consistent touch targets, but the reason I kept the x right next to the name is I thought the cog icon was meaning the x didn't stand out so much over on the right

I actually like the x button over on the right. I think it makes it more obvious how to clear the filter out. I've made that change and will get a new build for you.

WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
bjtitus: Yep, this was a bad attempt at sizing these labels but without having to manipulate the default frames. I've switched to repeating characters with a given width; I don't really love this, but it seems to work alright.

https://github.com/wordpress-mobile/WordPress-iOS/blob/de1b872f75f11397fa04d4285d12a393335bbbcd/WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift#L50

bjtitus:
@osullivanchris Build 27247 should be the one with the latest changes.

> I haven't added the iPad yet for testing, and nothing happens when I tap '+ Add new device' from my iPad. Any idea how to get it up and running?

Maybe check if the certificate shows up in Settings > General > Profiles & Device Management. You should be able to Install/Trust from there.

If that doesn’t work can you try going to this page on your iPad to see if it asks you to register?

https://install.appcenter.ms/register-device

osullivanchris:
Filter text looks much better, and the 'x' works fine for me on the right hand side, you're right!

Still having trouble accessing anything on iPad for some reason.

PR #13874, opened 2020-04-12
Fixes loading issues in the new Reader

Gio2018:
Improves topic syncing in ReaderStreamViewController

Fixes NA
Ref: #13580

To test:

- Go to Reader tab
- Switch between tabs and make sure that the content displays as expected
- log out and log in again (if you have more than one account, use another account)
- Go to reader tab
- make sure that the posts load correctly and that, when switching between tabs, the content displays as expected

PR submission checklist:

- [x] ~~I have considered adding accessibility improvements for my changes.~~ accessibility for the new reader will be handled in a future PR
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ not released


WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -30,7 +30,20 @@ import WordPressFlux
...
-    private var syncHelper: WPContentSyncHelper!
+    private var syncHelpers: [ReaderAbstractTopic: WPContentSyncHelper] = [:]
+
+    private var syncHelper: WPContentSyncHelper? {
+        guard let topic = readerTopic else {
+            return nil
+        }
+        let currentHelper = syncHelpers[topic] ?? WPContentSyncHelper()
+
+        if syncHelpers[topic] == nil {
```
bjtitus, line 41: Not sure if you necessarily need to check for `nil` here since the same instance will be written. Writing should be quicker than doing another lookup in the dictionary anyway.

PR #13880, opened 2020-04-13
Fix crash on Bottom Sheet

### LINKED ISSUE #13879: Crash when displaying the Nudges view multiple times
### How to reproduce

* Tap "Publish..." to show the Nudges view
* Tap "Publish" option to schedule
* Dismiss the calendar
* Dismiss the Nudges view
* Tap "Publish..." to show the Nudges view again
* Tap "Tags": 💥

leandroalonso:
Fixes #13879

### What was the issue

The method `configureScrollViewInsets` was crashing due to a `nil` property only on devices.

### What was fixed

We just check if `presentingViewController.view` exists before moving on.

### To test

Using a DEVICE (the crash doesn't happen on Simulator):

* Tap "Publish..." to show the Nudges view
* Tap "Publish" option to schedule
* Dismiss the calendar
* Dismiss the Nudges view
* Tap "Publish..." to show the Nudges view again
* Tap "Tags"
* Tags show be displayed

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
+++ b/WordPress/Classes/ViewRelated/System/Action Sheet/DrawerPresentationController.swift
@@ -422,11 +422,12 @@ private extension DrawerPresentationController {
         guard
             let scrollView = presentableViewController?.scrollableView,
             !scrollView.isScrolling,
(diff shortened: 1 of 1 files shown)


leandroalonso:
@emilylaguna Done!

PR #13906, opened 2020-04-15
Fixes some issues with GIF images in the reader.

### LINKED ISSUE #13895: Reader dropping frames while scrolling
Ref: p4a5px-2xT-p2

In the latest 14.6 internal, when viewing this post in reader, scrolling is very choppy

I've noticed this seems to be triggered once all images in the reader posts being shown are loaded.  We need to investigate what's causing this behavior.

diegoreymendez:
Fixes #13895 (well kind of, read on).

The issue in #13895 is that the GIFs were really large, so we kept loading frames at all times.  This PR changes the amount of frames in the buffer to 300, which means there will be no more jumps when scrolling the reader **AFTER THE GIF IS DONE LOADING**.

There will still be some jumps before the GIFs finish loading, and fixing this IMHO is not great in terms of efforts vs benefit.

Additionally, an issue in the reader was causing attachments to be loaded twice, so these GIFs were animating twice.  This was caused by a strong reference to the images held by a the `mediaArray` we're removing.

## To test:

1. Load in the reader the post shown in p4a5px-2xT-p2.
2. Make sure that after the GIFs are done loading for good, scrolling is smooth in that post.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Media/GIFPlaybackStrategy.swift
+++ b/WordPress/Classes/Utility/Media/GIFPlaybackStrategy.swift
@@ -54,18 +54,18 @@ extension GIFPlaybackStrategy {
 
 class SmallGIFPlaybackStrategy: GIFPlaybackStrategy {
     var maxSize = 8_000_000  // in MB
-    var frameBufferCount = 25
+    var frameBufferCount = 50
     var gifStrategy: GIFStrategy = .smallGIFs
 }
 
 class MediumGIFPlaybackStrategy: GIFPlaybackStrategy {
     var maxSize = 20_000_000  // in MB
-    var frameBufferCount = 50
+    var frameBufferCount = 150
     var gifStrategy: GIFStrategy = .mediumGIFs
 }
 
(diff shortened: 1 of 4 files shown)


diegoreymendez:
Fyi, @yaelirub

diegoreymendez:
@leandroalonso - The core issue was that loading the frames was consuming a lot of CPU.  The reason why the jumpy behavior never ended before was that the GIF had more frames than the buffer size (141 frames in the gif vs 50 in our buffer).

So the number of frames is one factor but I believe the file size is probably a factor too, as loading larger frames would consume more cpu.

Additionally we were loading all media twice and animating twice, which also consumed more memory and CPU.

I’ll want to profile this a bit more before merging, so let me know if you habe any other concern.

loremattei:
Hey! I'm moving this to 14.8 because 14.7 has been cut. If you want this to make it to 14.7, please feel free to ping me.

diegoreymendez:
I've tested in all of my devices and haven't found any issue with the new code. I'm merging.

PR #13925, opened 2020-04-17
Docs: Add Issue Triage doc

rachelmcr:
This PR adds documentation for what issue triage is, why we do it, and how it's done (labels and prioritization).

@designsimply Anything you'd add/change/remove here? Do you think it's clear?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -23,6 +23,7 @@
 - [Localization](localization.md)
 - [Feature Flags](feature-flags.md)
 - [Formattable Content](formattable-content.md)
+- [Issue Triage](issue-triage.md)
 
 ## Further Reading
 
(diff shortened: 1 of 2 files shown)


rachelmcr:
> Under Severity, I'd like to propose splitting the broken and non-functional high priority areas between High and Critical like this:

For now, I'd like to document the definitions of severity and impact as-is, since product leads have signed off on these definitions and it would be good to get them documented. I'm totally open to further discussion about how they're defined, and we can then update the documentation, but I think that should be a separate discussion so everyone involved can weigh in.

>It may help to add a clarifying statement or footnote to the Impact section such as

That’s great! Added.

@designsimply Could you take another look?

PR #13952, opened 2020-04-21
Fix crash when rendering comments

### LINKED ISSUE #13950: Crash when rendering comments
When navigating to the list of comments for an internal p2 post, the app will crash if there are comments.

### Expected behavior
It should be possible to navigate to the list of comments for a given post

### Actual behaviour
The app crashes when rendering the list of comments.

1. Navigate to the Notifications tab
2. Tap a notification (for example, a notification about a new post)
3. Tap the "comments" icon at the bottom right
4. The app navigates to the list of comments, and will crash, apparently, when it is about to render the list of comments

The crash can be reproduced when initiating the navigation from the Reader.

Tested on iPhone 11 Pro Max, iOS 13.4.1, WPiOS 14.7.0.0

leandroalonso:
Fixes #13950

This PR fixes the crash when loading the list of comments.

### To test

1. Navigate to the Notifications tab
2. Tap a notification (for example, a notification about a new post)
3. Tap the "comments" icon at the bottom right
4. The app navigates to the list of comments, and should not crash

### Additional info

From my tests, all `comment`'s entities had a `nil` `blog` property. What I did was to change the way the comments and cell is configured. Given a post or siteID/blogID I try to load the `Blog` associated with these params.

Then, this value is given to the cell, who uses it to configure the `MediaHost`. This fixes both the crash while maintains the logic to load images from a private Atomic site.

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
@@ -49,6 +49,8 @@ class ReaderCommentCell: UITableViewCell {
 
     @objc var comment: Comment?
 
+    @objc var blog: Blog?
+
     @objc var attributedString: NSAttributedString?
 
     @objc var showReply: Bool {
@@ -144,8 +146,9 @@ class ReaderCommentCell: UITableViewCell {
     // MARK: - Configuration
 
 
-    @objc func configureCell(comment: Comment, attributedString: NSAttributedString) {
+    @objc func configureCell(comment: Comment, blog: Blog?, attributedString: NSAttributedString) {
         self.comment = comment
+        self.blog = blog
         self.attributedString = attributedString
 
         configureAvatar()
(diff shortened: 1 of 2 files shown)


leandroalonso:
@frosty That's a good question. I'm not sure if somethings have changed.

What I do know is that comments from Reader have no parent `Blog` while comments from blogs that you have access do have. As you can see below in this print from the database:

Screen Shot 2020-04-21 at 11 51 44

The whole `.blog` thing has been an issue for some time, ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/12028

leandroalonso:
@loremattei since this is a crash, should I merge this to `develop` too?

loremattei:
@leandroalonso I'll do that as soon as I release the new beta tomorrow or on Thursday. 
If you need this into `develop` before, feel free to merge it.

aerych:
I'm a little late to the party but I had a thought.  Surfacing a chat with @leandroalonso here.  (Thanks for listening to my ramblings @leandroalonso!) 

It would be really nice if we could avoid making the Reader code aware of the Blog model or BlogService.  After a bit of testing it looks like the crash can be resolved with just a nil check on comment.blog in the ReaderCommentCell.  That would just leave the issue of authenticating for images on private atomic site comments, which could be handled by getting a MediaHost via the comments ReaderPost reference (if it has one).  Something like the following seems like it would do the trick: 

```
@objc func configureText() {
        guard let _ = comment, let attributedString = attributedString else {
            return
        }

        if let mediaHost = getMediaHostIfNeeded() {
            textView.mediaHost = mediaHost
        }

        // Use `content` vs `contentForDisplay`. Hierarchcial comments are already
        // correctly formatted during the sync process.
        textView.attributedText = attributedString
    }

    func getMediaHostIfNeeded() -> MediaHost? {
        guard let comment = comment else {
            return nil
        }
        var mediaHost: MediaHost?
        if let blog = comment.blog {
            mediaHost = MediaHost(with: blog, failure: { error in
                // We'll log the error, so we know it's there, but we won't halt execution.
                CrashLogging.logError(error)
            })
        } else if let post = comment.post as? ReaderPost, post.isPrivate() {
            mediaHost = MediaHost(with: post, failure: { error in
                // We'll log the error, so we know it's there, but we won't halt execution.
                CrashLogging.logError(error)
            })
        }
        return mediaHost
    }
```
Maybe we could revisit this patch in `develop` at some point?

PR #13957, opened 2020-04-21
Refactor the fix for comments crash

### LINKED ISSUE #13950: Crash when rendering comments

leandroalonso:
Fixes #13950

After talking to @aerych we can simplify the fix for the comments section crash while keeping Reader unaware of `BlogService`.

### To test

1. Navigate to the Notifications tab
2. Tap a notification (for example, a notification about a new post)
3. Tap the "comments" icon at the bottom right
4. The app navigates to the list of comments, and should not crash

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
@@ -49,8 +49,6 @@ class ReaderCommentCell: UITableViewCell {
 
     @objc var comment: Comment?
 
-    @objc var blog: Blog?
-
     @objc var attributedString: NSAttributedString?
 
     @objc var showReply: Bool {
@@ -146,9 +144,8 @@ class ReaderCommentCell: UITableViewCell {
     // MARK: - Configuration
 
 
-    @objc func configureCell(comment: Comment, blog: Blog?, attributedString: NSAttributedString) {
+    @objc func configureCell(comment: Comment, attributedString: NSAttributedString) {
         self.comment = comment
-        self.blog = blog
         self.attributedString = attributedString
 
         configureAvatar()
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
```diff
@@ -205,17 +202,17 @@ class ReaderCommentCell: UITableViewCell {
...
+        guard let comment = comment, let attributedString = attributedString else {
+            return
+        }
+
+        if let blog = comment.blog {
             textView.mediaHost = MediaHost(with: blog, failure: { error in
                 // We'll log the error, so we know it's there, but we won't halt execution.
                 CrashLogging.logError(error)
             })
         }
```
aerych, line 214: Maybe add :   
```suggestion
        }  else if let post = comment.post as? ReaderPost, post.isPrivate() {
            textView.mediaHost = MediaHost(with: post, failure: { error in
                // We'll log the error, so we know it's there, but we won't halt execution.
                CrashLogging.logError(error)
            })
        }
```

WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
```diff
@@ -252,6 +243,22 @@ class ReaderCommentCell: UITableViewCell {
         textView.updateLayoutForAttachments()
     }
 
+    /// Returns the media host for the current comment
+    private func mediaHost() -> MediaHost {
+        if let blog = comment?.blog {
+            textView.mediaHost = MediaHost(with: blog, failure: { error in
```
aerych, line 249:
```suggestion
            return MediaHost(with: blog, failure: { error in
```

WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
```diff
@@ -252,6 +243,22 @@ class ReaderCommentCell: UITableViewCell {
...
+                // We'll log the error, so we know it's there, but we won't halt execution.
+                CrashLogging.logError(error)
+            })
+        } else if let post = comment?.post as? ReaderPost, post.isPrivate() {
+            textView.mediaHost = MediaHost(with: post, failure: { error in
```
aerych, line 254:
```suggestion
            return MediaHost(with: post, failure: { error in
```

Issue #13939, opened 2020-04-18
ULS: Signup Epilogue


ScoutHarris:
Implement new Signup Epilogue.

Refs:
- [Unified Auth Design](https://wp.me/pbArwn-mA)
- [Login Epilogue](https://github.com/wordpress-mobile/WordPress-iOS/issues/13413)

Tasks:
- [x] Update User Info section. (https://github.com/wordpress-mobile/WordPress-iOS/pull/13961, https://github.com/wordpress-mobile/WordPress-iOS/pull/13982)
- [x] Update table view:
  - [x] Remove Password table section (https://github.com/wordpress-mobile/WordPress-iOS/pull/13962)
  - [x] Update table view and cells (https://github.com/wordpress-mobile/WordPress-iOS/pull/13973)
  - [x] Password field: fix left margin. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14010)
  - [x] Username field: fix disclosure icon right margin. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14010)
- [x] Update button view. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14013)
- [x] Update iPad layout. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14018)
- [x] Accessibility. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14019)

Issues:
  - [ ] Status bar text is white in light mode.
  - [ ] Password field: Secure text icon color has not been updated. (https://github.com/wordpress-mobile/WordPress-iOS/pull/14049)
  - [ ] Password field: when the view refreshes (i.e. changing split mode, toggling light/dark mode), the field is inset by a few points.

Pre-existing issues:
- [x] Table doesn't adjust properly for dynamic text.

---
Designs from Zeplin:

| ![Screen : Login : Sign Up : Step 3](https://user-images.githubusercontent.com/1816888/80138152-9e288900-8561-11ea-9a42-61f223abff9c.png) | ![Screen : Login : Sign Up : Step 3 Dark](https://user-images.githubusercontent.com/1816888/80138166-a4b70080-8561-11ea-98cf-dab5698fa108.png) |
|--------|-------|

PR #13961, opened 2020-04-21
Signup Epilogue: update User Info section

ScoutHarris:
Ref #13939 

This updates the Signup Epilogue's User Info section to the new design. Specifically:
- 'New Account' section header removed.
- Blue 'add' gridicon is now used for the add gravatar button.
- Top and bottom separator lines removed.
- Background color updated (light and dark modes).

Please note the table itself hasn't been updated, so the table colors are still off (i.e. the area around the User Info section). This change only applies to the User Info section itself. I'll request Design review once the table is updated and it looks more final.

---
To test:
- Signup.
  - If you don't want to actually sign up, see below for WPAuth hack tip.
- On the Signup Epilogue, verify the User Info section looks like that on #13939 or the designs in Zeplin.

| Before | After |
|--------|-------|
| before_light | after_light |
| before_dark | after_dark |

---
Want to see the signup epilogue without actually signing up? Well you came to the right place!

```
        fatalError()
    }
    
//        }
}
```
- Then _login_ with an existing account, and the _signup_ epilogue will display. 🎉 

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/EpilogueUserInfoCell.swift
+++ b/WordPress/Classes/ViewRelated/NUX/EpilogueUserInfoCell.swift
@@ -91,8 +91,7 @@ private extension EpilogueUserInfoCell {
 
     func configureColors() {
         gravatarAddIcon.tintColor = .primary
-        // TODO: update when table background updated.
-        gravatarAddIcon.backgroundColor = .listForeground
+        gravatarAddIcon.backgroundColor = .basicBackground
 
         fullNameLabel.textColor = .text
         fullNameLabel.font = fullNameFont()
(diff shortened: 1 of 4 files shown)


ScoutHarris:
@Gio2018 @bjtitus since @mindgraffiti 's AFK this week, I'd appreciate it if either of you could review this. Thanks!

Gio2018:
> I think this is probably out of scope for this PR, but it looks like my Gravatar isn't showing up if I already have one. If I go to pick one from the picker but never select anything and dismiss the modal, it does show up. I sometimes see a similar thing in the app, though, so @Gio2018 may have a better idea.

I agree this might be related to the same issue we are seeing on the me gravatar button. I'll let you know what I find out!

ScoutHarris:
> I think this is probably out of scope for this PR

Indeed it is. I saw it happen prior to these changes, and these changes are UI only. 

> I agree this might be related to the same issue we are seeing on the me gravatar button. I'll let you know what I find out!

@Gio2018 that'd be great thank you!

Gio2018:
> @Gio2018 that'd be great thank you!

There's a PR for that now 🙂

PR #13962, opened 2020-04-21
Signup Epilogue: remove Password table section

ScoutHarris:
Ref #13939 

This removes the table section just for the Password field, and adds the Password field to the same section as the other account fields.

To note, this is still a WIP, so table colors are still off. This change is just to move the Password field. 

---
To test:

- Sign up with Email. (If you don't want to actually sign up, see below for WPAuth hack tip.)
- On the epilogue, verify:
  - The password field is displayed below the Username field, in the same section.
  - The `You can always...` text is displayed below the Password field (in the section footer).
  
| Before | After |
|--------|-------|
| wp_before | wp_after |

---
- Sign up with a social account. (If you don't want to actually sign up, see below for WPAuth hack tip.)
- On the epilogue, verify the password field and footer text are not displayed.

| Before | After |
|--------|-------|
| social_before | social_after |

---

```
        fatalError()
    }
    
//        }
}
```

- Login with an existing WP account to verify the Password field is shown.
- Login with a Google account to verify the Password field is not shown.

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/SignupEpilogueTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/NUX/SignupEpilogueTableViewController.swift
@@ -32,16 +32,15 @@ class SignupEpilogueTableViewController: NUXTableViewController, EpilogueUserInf
     private var reloaded: Bool = false
 
     private struct Constants {
-        static let numberOfSections = 3
-        static let namesSectionRows = 2
-        static let sectionRows = 1
+        static let numberOfSections = 2
+        static let userInfoRows = 1
+        static let noPasswordRows = 2
+        static let allAccountRows = 3
         static let headerFooterHeight: CGFloat = 50
     }
 
     private struct TableSections {
         static let userInfo = 0
-        static let names = 1
-        static let password = 2
     }
 
     private struct CellIdentifiers {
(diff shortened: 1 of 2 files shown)


bjtitus:
I think this may be out of scope for this PR but I noticed that the field inside of the cell doesn't seem to match the cell's background color in dark mode. I wonder if the background on the `cellField` should just be `clear`.

<a href="https://user-images.githubusercontent.com/3250/80005198-7a414680-8480-11ea-8649-c4c1e4c7e944.png"></a>

WordPress/Classes/ViewRelated/NUX/SignupEpilogueTableViewController.swift
```diff
@@ -103,24 +103,24 @@ class SignupEpilogueTableViewController: NUXTableViewController, EpilogueUserInf
...
+        return cell
     }
 
     override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
 
+        // User Info Row
         if indexPath.section == TableSections.userInfo {
             guard let cell = tableView.dequeueReusableCell(withIdentifier: CellIdentifiers.epilogueUserInfoCell) as? EpilogueUserInfoCell else {
-                fatalError("Failed to get a user info cell")
+                return UITableViewCell()
```
Gio2018, line 123: 👍

ScoutHarris:
Hey @bjtitus .

> I think this may be out of scope for this PR but I noticed that the field inside of the cell doesn't seem to match the cell's background color in dark mode.

Yea, that's what I meant by this:

> To note, this is still a WIP, so table colors are still off. 

I should have said table _and cell_ colors. 

It will be corrected in a future PR. Thanks!

PR #13968, opened 2020-04-22
Fix editor button from being truncated

### LINKED ISSUE #13966: Editor: Schedule button title truncated when scheduling from prepublish nudge flow
When selecting a future post date using the prepublishing nudges flow, you can see the Publish button in the editor toolbar changes to say Schedule but the title gets truncated:

IMG_2451

(Ignore that the prepublish sheet is on the Visibility screen, the bug happened when changing the date)

I saw this affecting both the block editor and Aztec.

Tested on iPhone 11 Pro, WPiOS 14.7 (internal)

leandroalonso:
Fixes #13966

### To test

1. Create a new post
2. Tap Publish
3. Schedule the post
4. Check if the top-right button is not truncated and displays the whole word


leandroalonso:
@frosty let me know if you can review this one. 🙇🏻‍♂️

PR #13970, opened 2020-04-22
Change Media Picker label from "Preview %@" to "Edit %@"

leandroalonso:
This PR changes the bottom-left button's label in Media Picker. Instead of showing "Preview X" (where X is the number of images) we now say "Edit X".

This was discussed with @mbshakti and @malinajirka.

My intention was to point this PR to the release branch. But since this adds a new string (and the translation for the release branch has already started) I'm targeting `develop`. 

#### Gutenberg

1. Add a "Gallery Block"
2. Tap "Add Media"
3. Tap "Choose from Device"
4. Select one or more images and check that the label of the left button says "Edit X" (where X is the number of images you selected)

#### Aztec

1. Add a "+" at the action bar
2. Tap the image icon (right next to the WP logo)
3. Select one or more images and check that the label of the left button says "Edit X" (where X is the number of images you selected)


yaelirub:
Works as expected

PR #13973, opened 2020-04-23
Signup Epilogue: update table and cells for new design

ScoutHarris:
Ref #13939 
Related WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/252 (to add the secure text toggle button to the Password field)

This updates the Signup Epilogue table and Account Details cells to match the new design. 

No functionality has changed, with the exception of the secure text entry toggle on the Password field. Tapping it will hide/show the text entered in the Password field.

---
Known issues:

Password field:
- There is a leading margin that makes the field inset by 8 points.
- The secure text toggle icon color doesn't match the design.

Username field:
- The disclosure icon is inset 4 points.

These use a custom field type defined in WPAuth, and is used in several other auth views. I'll look at these separately since I'll have to figure out how to update this view without affecting the others.

---
To test:
- Signup. (See below for Auth hack to avoid actually signing up.)
- On the Epilogue, verify everything above the `Continue` button looks like the design on #13939. 

| Before | After |
|--------|-------|
| before_light | after_light |
| before_dark | after_dark |

---

- Check out the accompanying `WordPressAuthenticator` branch locally. (`issue/13939-secure_text_entry_toggle`)

```
        fatalError()
    }
    
//        }
}
```

---
PR submission checklist:


ScoutHarris:
Hey @mattmiklic . Thanks for the feedback! Regarding the issues you noticed:

- Top padding on the footer has been changed to 8 pts.
- Top padding on the password field has been changed to 16 pts.
- Username detail/disclosure is inset a few px - has been added to the Known Issues in the description. I'll look at it when I look at the password field.
- Add 18pt of padding above the gravatar - will be done separately as it affects both epilogues.

I've updated the screenshots in the description.

@Gio2018 I think we're ready for a review now.

WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
```diff
@@ -64,27 +72,52 @@ class SignupEpilogueCell: UITableViewCell {
...
+        let isPassword = (cellType == .password)
+        cellLabel.isHidden = isPassword
+
+        cellField.isSecureTextEntry = isPassword
+        cellField.showSecureTextEntryToggle = isPassword
+        cellField.textAlignment = isPassword ? .left : .right
+        cellField.textColor = isPassword ? .text : .textSubtle
+
+        cellFieldLeadingConstraintWithLabel.isActive = !isPassword
+        cellFieldLeadingConstraintWithoutLabel.isActive = isPassword
```
Gio2018, line 117: Just a question/suggestion: would it be easier to use a stack view to handle the case of a label hidden/not hidden instead of explicitly manipulate constraints?

ScoutHarris: You'd think so wouldn't you? Well, I tried. 😄 But because of the behind the scenes magic done with the custom field in WPAuth, it did _not_ go well.

FWIW, I firmly believe after we update `LoginTextField` (the custom field) for the new auth design, this epilogue will have to be adjusted anyway. If we're clever about it, maybe it'll end up being smarter.

Gio2018: Oh I see now! Thanks for clarifying!!

PR #13984, opened 2020-04-24
Information Architecture Improvements

Gio2018:
Ref #13580, #13319

#### Adds a few improvements and fixes related to the above issues of the [IA project](https://github.com/wordpress-mobile/WordPress-iOS/projects/15)

- Fix reader bar divider colors in dark mode
- Fix the fallback image for the 'Me' button
- Remove the site header in the 'Reader' when selecting a site from 'Filter'
- Code & improvements in `ReaderTabViewModel`, introduced `ReaderTabItemStore`

To test:

- Go to 'My Site' and verify that the fallback image corresponds to `userCircle` in `GridIcons`
  - **Note**: usually you'll need an account with no gravatar for this check, but at the moment there's a bug (whose fix is not yet merged into develop) that prevents to load the gravatar image, so you can use any account.

- Enable Dark Mode, go to the 'Reader' and check that the styles of all the dividers are consistent between themselves and with other dividers in the app (e.g. the tab bar in 'Stats')

- From the reader, tap 'Filter' and select a site, verify that no header is displayed. Clear the filter, and select a site in a different way (e.g. from 'Search' or tapping on the site icon/name in a card), and verify that the header shows up.

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ will be in a eparate PR.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ will be in a eparate PR.


osullivanchris:
> Fix reader bar divider colors in dark mode

fixed, thanks!

> Fix the fallback image for the 'Me' button

Also fixed!

> Remove the site header in the 'Reader' when selecting a site from 'Filter'

Fixed, of course!

Two minor things which I have spotted, which you can create separate issues for if you want

**First issue**
- Kill app
- Reopen app (on My Site)
- Tap on 'Reader' tab
- for some reason, the tabs take a moment to load in. This is fine. But the filter bar displays immediately. I guess we are making some check which tabs the user should have. It just looks a little weird while loading that I can see 'Filter' but a gap where the tabs should be. Maybe we should not display the filter bar until the tabs load?

**Second issue**
- Open app
- leave app, go to iPhone settings, appearance and brightness
- change to dark/light mode
- go back to WordPress app
- Go to reader
- the cog and filter icons on the filter bar are the wrong color
- this happens whenever you change color mode - it will look wrong on the mode you change to, both light and dark
- killing and re-opening the app resolves

bjtitus:
> the tabs take a moment to load in

Would it make more sense to use the Ghost views for the unloaded tabs or just have a loading view for the entire screen?

Gio2018:
> Two minor things which I have spotted, which you can create separate issues for if you want
Thanks @osullivanchris for pointing out those two!

> Would it make more sense to use the Ghost views for the unloaded tabs or just have a loading view for the entire screen?

@bjtitus love the idea of the ghost view! I believe that could work, as the content would not show up before the tab bar is populated anyway. @osullivanchris what do you think?

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabViewModel.swift
```diff
@@ -48,7 +80,7 @@ extension ReaderTabViewModel {
         } else {
             selectedTopic = topic
         }
-        topicSelectionCallback?(selectedTopic)
```
bjtitus, line 51: 👍

ScoutHarris:
Hey @Gio2018 .

> From the reader, tap 'Filter' and select a site

On iPad, the Filter view doesn't display properly:

Screen Shot 2020-04-24 at 3 23 28 PM

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
```diff
@@ -73,49 +86,30 @@ private extension UIBarButtonItem {
...
-                       relatedBy: .equal,
-                       toItem: view,
-                       attribute: .trailingMargin,
-                       multiplier: 1,
-                       constant: 0)
-        .isActive = true
+        NSLayoutConstraint.activate([
+            imageView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            imageView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
```
ScoutHarris, line 96: 👍

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -252,7 +251,7 @@ extension ReaderTabView {
         static let settingsButtonWidth: CGFloat = 56
 
         static let dividerWidth: CGFloat = .hairlineBorderWidth
-        static let dividerColor: UIColor = UIColor(light: .lightGray, dark: .darkGray)
+        static let dividerColor: UIColor = .neutral(.shade10)
```
ScoutHarris, line 254: Looks like `.neutral(.shade10)` is the standard divider color, so maybe this could just be `.divider`?

WordPress/Classes/ViewRelated/System/FilterTabBar.swift
```diff
@@ -356,7 +356,19 @@ class FilterTabBar: UIControl {
         }
     }
 
+    var currentlyCelectedItem: FilterTabBarItem? {
```
ScoutHarris, line 359: I expect this probably should be `currentlySelectedItem`?

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
```diff
@@ -53,10 +55,21 @@ private extension UIBarButtonItem {
     func makeGravatarTappableView(with email: String?, target: Any?, action: Selector?) -> UIView {
         let gravatarImageView = GravatarButtonView(tappableWidth: GravatarConfiguration.tappableWidth)
 
+        gravatarImageView.adjustView = { [weak self] in
+            // if there's a gravatar, add the border, if not, remove it and resize the userCircle image
+            if $0.image == GravatarConfiguration.fallBackImage {
```
bjtitus, line 60: Could we the parameter be named here?

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
```diff
@@ -0,0 +1,91 @@
...
+            switch self {
+            case .loading:
+                return true
+            case .error, .ready:
+                return false
+            }
+        }
+    }
+
+    var state: State = .ready([]) {
```
bjtitus, line 23: Shouldn't `state` start out loading?

WordPress/Classes/ViewRelated/System/FilterTabBar.swift
```diff
@@ -356,7 +356,19 @@ class FilterTabBar: UIControl {
...
+        guard !items.isEmpty, selectedIndex < items.count else {
```
bjtitus, line 360: I think you can check if the indices contain this index here:

```suggestion
        guard items.indices.contains(selectedIndex) else {
```

WordPress/Classes/ViewRelated/System/FilterTabBar.swift
```diff
@@ -356,7 +356,19 @@ class FilterTabBar: UIControl {
...
+            return nil
+        }
+        return items[selectedIndex]
+    }
+
     func setSelectedIndex(_ index: Int, animated: Bool = true) {
+        guard index >= 0, index < items.count else {
```
bjtitus, line 367: I think you can check if the indices contain this index here:

```suggestion
        guard items.indices.contains(index) else {
```

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
Gio2018: The loading is requested by the view, so at the init time, the store it's not loading, and it needs to be set to '.loading' only when it's actually loading, to prevent multiple calls to getItems(), so I'd leave it as it is.

WordPress/Classes/ViewRelated/System/FilterTabBar.swift
bjtitus: Sorry, I just realized there _is_ an extension on Collection for safe access:

https://github.com/wordpress-mobile/WordPress-iOS/blob/c1446d7a1f84812815ace1b8d8ccf9fad15d2e13/WordPress/Classes/Extensions/CollectionType%2BHelpers.swift#L15

Should be something like

```swift
return items[safe: selectedIndex]
```

Gio2018: This looks pretty cool!! Thanks for the hint

Gio2018:
@osullivanchris :
> Do you mean for the entire view, or just for the tabs? Showing some empty space for the taps, whether a ghost grey text blocks, or even just an empty white bar, would solve this issue for me!

the empty bar is currently available here (build number 27761). The other improvements will be part of another PR. Thanks!!

Issue #13985, opened 2020-04-24
Prepublishing Nudges: Stuck in publish loop after changing publish date to immediately


rachelmcr:
### Expected behavior

I expect that if I decide not to publish a draft post, I will be able to save my changes and exit the editor.

### Actual behavior

If I start to publish a post, change the publish date to Immediately, and then close the bottom sheet (i.e. decide not to publish now), I get stuck in a publish loop.

When I try to exit the editor, I have the options to update the post or discard my changes. If I've made changes to the draft that I want to save, I choose update — but that reopens the pre-publish nudge rather than saving the draft.

The only way to save my draft without publishing is to go to the ellipsis menu and select Save as Draft, but that's not obvious.

### Steps to reproduce the behavior

1. Go to the My site tab > Blog posts > Drafts.
2. Select a draft to open it in the editor.
3. Optional: make some changes to the post content.
4. Open the ellipsis menu and tap "Publish now."
5. Notice that the publish date is set to when I last saved the draft. Tap Publish > select Publish immediately.
6. Dismiss the bottom sheet.
7. Tap "X" to close the editor.
8. Notice the "You have unsaved changes" prompt appears. Select "Update Post" to save the changes to the draft.

Result: The pre-publishing nudge appears again.

Screencast: https://cloudup.com/cM5yyYyggH5

rachelmcr:
The same kind of loop can occur if you change the post visibility to Private in the pre-publish nudge, since a private post has to be published (you can't save a draft with visibility set to private).

leandroalonso:
@rachelmcr I submitted a PR to fix the schedule issue (both in the Nudges view and Post Settings). 🎉

However, changing the visibility issue is a little more tricky I guess. In the previous version, the same change (but under Post Settings) produces similar results, but instead of displaying the Prepublishing Nudges an action sheet was shown:

<img src="https://user-images.githubusercontent.com/7040243/80228768-d7c2c800-8625-11ea-9864-b1f4b0161d47.PNG" width="300px" />

Do you think we can create another issue and prioritize it for Groundskeeping?

rachelmcr:
Thanks for the quick fix @leandroalonso!

> Do you think we can create another issue and prioritize it for Groundskeeping?

Yes, I think in this case that makes sense because it may take some additional thought around how to improve the UX for private posts. I opened https://github.com/wordpress-mobile/WordPress-iOS/issues/13995 to track that.

[...]

PR #13993, opened 2020-04-24
Fix scheduling issues

### LINKED ISSUE #13983: Prepublishing Nudges: customized date is ignored for previously saved drafts
Steps to reproduce:

1. Go to My Site > Blog Posts > +.
1. Add a title and some content.
1. Tap More > Save as Draft. 
1. Tap More > Publish Now. 
1. Change the date to something in the future.
1. Tap the "Schedule Now" button.
1. Notice the post is unexpectedly published with the current date and time. 

**Result:** customized date is ignored for previously saved drafts. ([1m15s](https://cloudup.com/cBnL0zq6OLZ))

Scheduled for May 9 on Apr 23 | Post was published immediately
--- | ---
scheduled-for-may-9|published-immediately
<p style="clear:both;"><sup>Tested with WPiOS 14.7.0.1 TestFlight beta on iPhone 11 iOS 13.4.1.</sup></p>

**Note:** this happens when scheduling posts for the future or backdating posts to the past. It happens if you close and re-open a draft and then try to schedule it or if you use "Save as Draft" and then publish in the same session.

<p style="clear:both;"><sup>h/t @rachelmcr for reporting this in Slack. Thank you!</sup></p>

leandroalonso:
Fixes #13983 and #13985

#### Customized date is ignored for previously saved drafts (#13983)

1. Go to My Site > Blog Posts > +.
1. Add a title and some content.
1. Tap More > Save as Draft. 
1. Tap More > Publish. 
1. Change the date to something in the future.
1. Tap the "Schedule Now" button.
1. The post should be scheduled

#### Stuck in publish loop after changing publish date to immediately (#13985)

For this one, I fixed two issues: we do not display a date in the past. If it's in the past, we show "Immediately". Also, if you schedule a draft and then remove the scheduling, the draft will behave as a draft (instead of "Publish" as it was before, you'll see "Update").

1. Go to the My site tab > Blog posts > Drafts.
2. Select a draft to open it in the editor.
3. Optional: make some changes to the post content.
4. Open the ellipsis menu and tap "Publish"
5. Notice that the publish date is set to "Immediately". Tap Publish > select Publish immediately.
6. Dismiss the bottom sheet.
7. Tap "X" to close the editor.
8. Notice the "You have unsaved changes" prompt appears. Select "Update Draft" to save the changes to the draft.

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
@@ -27,7 +27,7 @@ struct PublishSettingsViewModel {
     let title: String?
 
     var detailString: String {
-        if let date = date {
+        if let date = date, post.hasFuturePublishDate() {
             return dateTimeFormatter.string(from: date)
         } else {
             return NSLocalizedString("Immediately", comment: "Undated post time label")
--- a/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
@@ -83,7 +83,11 @@ struct PublishSettingsViewModel {
             } else {
                 post.status = .publish
             }
-        } else {
+        } else if post.originalIsDraft() {
+            // If the original is a draft, keep the post as a draft
+            post.status = .draft
+            post.dateCreated = Date()
+        } else if post.hasFuturePublishDate() {
             post.publishImmediately()
         }
 
--- a/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
@@ -77,6 +77,7 @@ struct PublishSettingsViewModel {
 
     mutating func setDate(_ date: Date?) {
         if let date = date {
+            // If a date to schedule the post was given
             post.dateCreated = date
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
```diff
@@ -83,7 +83,11 @@ struct PublishSettingsViewModel {
             } else {
                 post.status = .publish
             }
-        } else {
+        } else if post.originalIsDraft() {
+            // If the original is a draft, keep the post as a draft
+            post.status = .draft
+            post.dateCreated = Date()
+        } else if post.hasFuturePublishDate() {
             post.publishImmediately()
```
yaelirub, line 91: Im a bit confused by this. Why would we publishImmediately if the post has future publish date?

yaelirub:
@leandroalonso , thanks for fixing this!
It works as expected following the testing steps. 
I just have a small clarification question

WordPress/Classes/ViewRelated/Post/Scheduling/PublishSettingsViewController.swift
leandroalonso: In this case:

1. No date was given (aka the user tapped "Publish immediately"
2. The post is scheduled, which means we're changing it to be published immediately
3. The post is not a draft (it is an already scheduled published)

I'm open to suggestions because it's confusing. A comment might help?

yaelirub: A comment would be great!!

PR #14012, opened 2020-04-28
Login Epilogue: update content margins

ScoutHarris:
Ref #13413 

This updates the Login Epilogue content margins to be 20pts leading and trailing. The margins have been added to the cells and removed from the table.

---
To test:
- Login.
- On the epilogue, verify:
  - All the content has 20 pt leading and trailing margins.
  - The separator lines have 20 pt leading and 0 pt trailing margins.
  - On iPhone, the Done button as 20 pt leading and trailing margins.
  - On iPad, the Done button has 0 leading and trailing margins (i.e. is the same width as the table).

| iPhone | iPad |
|--------|-------|
| iphone | ipad |

---
PR submission checklist:


ScoutHarris:
> My sims aren't showing borders (again), so I tested on my devices. The border line on some of the site icons is a known bug, right?

@mindgraffiti Uh... not known to me. That I recall. I suspect it's a pre-existing issue, but I'll certainly take a look.

PR #14013, opened 2020-04-29
Signup Epilogue: update Done button

ScoutHarris:
Ref #13939 

This updates the `Done` button on the Signup Epilogue to the new design. Except the button color. That's still pink to match the rest of the flow.

Technically:

Previously, `NUXButtonViewController` was used to display the button. This added most of the style that we no longer want. So I replaced it with a simple `FancyButton` (which is what the _Login_ Epilogue uses).

Visually:
- The button title is now `Done` (it was `Continue`).
- The area around the button now matches the table view color.
- The shadow between the button and the table has been removed.

| light | dark |
|--------|-------|

---
To test:
- Signup. (See below if you don't want to actually signup)
- Verify the `Done` button looks like that on #13939 or in Zeplin.

---

```
        fatalError()
    }
    
//        }
}
```

- _Login_ with an existing WP account, and the _signup_ epilogue will be displayed.

---
PR submission checklist:


WordPress/WordPressUITests/Screens/Signup/SignupEpilogueScreen.swift
```diff
@@ -6,8 +6,10 @@ private struct ElementStringIDs {
     static let displayNameField = "Display Name Field"
     static let usernameField = "Username Field"
     static let passwordField = "Password Field"
+    // This is the Done button on the Login Epilogue
     static let doneButton = "Done"
-    static let continueButton = "Continue Button"
+    // This is the Done button on the Signup Epilogue
```
mindgraffiti, line 11: Thanks for the documentation 💯

PR #14018, opened 2020-04-29
Signup Epilogue: adjust table width for iPad

ScoutHarris:
Ref #13939 

This updates the Signup Epilogue's table width based on the size class and device orientation.
- For the Regular x Regular size class:
  - In portrait, the table is 66% of the view width.
  - In landscape, the table is 50% of the view width.
- For all other size classes, the table is 100% of the view width.

In addition, the Done button left/right margins always line up with the table's content.

To test:

---
- Signup on an iPad. (See below to avoid actually signing up)
- Verify the table and button width are sized properly in portrait and landscape.

| ipad_port | ipad_land |
|--------|-------|

---
- Signup on an iPhone, or multitask on an iPad and change the view size.
- Verify the table and button width are sized properly.

iphone

---

- Check out the accompanying `WordPressAuthenticator` branch locally.

```
        fatalError()
    }
    
//        }
}
```

---
PR submission checklist:


ScoutHarris:
@mattmiklic , @mindgraffiti  - I've fixed these issues:
- The Done button margins now match the table content margins.
- 50/50 split view on larger iPads should display properly now.

Ready for (another) review!

PR #14027, opened 2020-04-30
Epilogues: fix status bar color

ScoutHarris:
Ref: #13939, #13413

This corrects the status bar color on the new Epilogues.

To test:

Note: The problem didn't always happen in simulators, so I suggest testing on a real device.

- Login or Signup. They use the same status bar setting, so either will suffice.
- On the epilogue, verify:
  - In light mode, the status bar text is dark.
  - In dark mode, the status bar text is light.

---

| Light | Dark |
|--------|-------|
| light_mode | dark_mode |

---
PR submission checklist:


ScoutHarris:
Uh... nevermind. I clearly didn't think this fix through. Let's just pretend this didn't happen.

Issue #13787, opened 2020-04-01
iA Tooltip for FAB


bjtitus:
Add a tooltip for the Create FAB so new users are introduced to the navigation change.

https://ministryofmagicp2.wordpress.com/2020/01/28/ia-project-my-site-ios/#comment-571

designsimply:
Note: I removed the 14.6 milestone from this issue because I noticed the associated PR now shows the milestone as 14.8 and I think that PRs are traditionally the best place to track the milestone in most cases. 👍

PR #14029, opened 2020-04-30
Adds a Tooltip for the Create Button

bjtitus:
#13787

Adds the tooltip prompt to the Create Button. The Notice is repurposed to show this particular tooltip. It hasn't been fully genericized to place a tooltip anywhere (the arrow is in a fixed position), but can be positioned using a `sourceView` which is used to anchor the tooltip.

<a href="https://user-images.githubusercontent.com/3250/80668256-9df72480-8a5e-11ea-95e6-8422e8cb75d5.png"></a>
<a href="https://user-images.githubusercontent.com/3250/80668262-a51e3280-8a5e-11ea-9027-ba8ddc039cc0.png"></a>

To test:
- Open the app to any Blog Details page (My Sites > Select a blog)
- Notice the tooltip that shows above the Create Floating Action Button

PR submission checklist:


bjtitus:
@Gio2018 Just pushed a fix that should solve those rotation issues.

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -61,13 +61,9 @@ class NoticePresenter: NSObject {
     private init(store: NoticeStore) {
         self.store = store
```
Gio2018, line 63: I know it's pre-existing, but I'm not sure this needs to be exposed to Objective-C anymore, so maybe we can remove the inheritance from NSObject?

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -296,14 +298,22 @@ class NoticePresenter: NSObject {
...
+                    case .present(let keyboardHeight):
+                        return -keyboardHeight + noticeContainer.bounds.height
+                    case .notPresent:
+                        return self.window.untouchableViewController.offsetOffscreen
+                    }
+                }()
+            case .fade:
+                ()
+            case .none:
+                ()
```
Gio2018, line 315: will these be implemented in other PRs?

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -316,7 +326,15 @@ class NoticePresenter: NSObject {
...
             noticeContainer.noticeView.alpha = WPAlphaFull
-            noticeContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant
+
+            switch self.currentNoticePresentation?.notice.style.animationStyle {
+            case .moveIn:
+                noticeContainer.bottomConstraint?.constant = self.onscreenNoticeContainerBottomConstraintConstant
+            case .fade:
+                ()
+            case .none:
+                ()
```
Gio2018, line 336: same question as above

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
```diff
@@ -347,6 +358,72 @@ fileprivate extension UIView {
 
     struct Constants {
         static let borderColor = UIColor.white.withAlphaComponent(0.25)
+        static let visualEffect = UIBlurEffect(style: .extraLight)
+    }
+}
+
+// MARK: - Arrow
```
Gio2018, line 365: 🏹

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
bjtitus: This was a great catch! I was thinking the notification observers required this for some reason.

Thanks!

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
bjtitus: I changed this a bit to handle the optional value separately and added some notes in a comment.

loremattei:
Hey! I'm moving this to 14.9 since 14.8 has been cut. Please, feel free to ping me if you want this to make it to 14.8.

bjtitus:
@osullivanchris Yep! I must have copied the wrong value:

<a href="https://user-images.githubusercontent.com/3250/81217227-78f03d80-8f99-11ea-88dc-d91e2eb6117c.png"></a>

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -346,7 +360,7 @@ - (void)viewWillAppear:(BOOL)animated
     }
 
     if (self.splitViewControllerIsHorizontallyCompact) {
-        [self animateDeselectionInteractively];
+//        [self.tableView animateDeselectionInteractively];
```
Gio2018, line 363: can we remove the commented out code?

WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
```diff
@@ -60,6 +83,10 @@ import Gridicons
     }
 
     @objc private func showCreateSheet() {
+
+        shouldShowNotice = false
+//        ActionDispatcher.dispatch(NoticeAction.clear(notice))
```
Gio2018, line 88: here too, can we remove the commented out code?

WordPress/Classes/ViewRelated/System/Notices/NoticeAnimator.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    let duration: TimeInterval
+    let springDampening: CGFloat
+    let springVelocity: CGFloat
+
+    /// Present the `Notice` inside of any view. If the notice includes a `sourceView` the constraints will be attached to that view (so the `sourceView` and `view` parameter **MUST** be in the same view hierarchy).
+    /// - Parameters:
+    ///   - notice: The `Notice` to present.
+    ///   - view: The `UIView` to add the `Notice` to.
+    /// - Returns: <#description#>
```
Gio2018, line 13: Looks like the placeholder remained in the comments

WordPress/Classes/ViewRelated/System/Notices/NoticeAnimator.swift
```diff
@@ -0,0 +1,104 @@
...
+            case .fade:
+                // Fade just changes the alpha value which both animations need
+                break
+            }
+
+            view?.layoutIfNeeded()
+        }
+    }
+
+    func onscreenState(for noticeContainer: NoticeContainerView, in view: UIView? = nil, bottomOffset: CGFloat = 0) -> AnimationBlock {
```
Gio2018, line 65: can we do one main method and two convenience methods where we pass in the alpha value and call the main one with the alpha value as a param?

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -31,7 +31,7 @@ import WordPressFlux
 ///
 /// - SeeAlso: `NoticeStore`
 /// - SeeAlso: `NoticeAction`
-class NoticePresenter: NSObject {
```
Gio2018, line 34: 👍

WordPress/Classes/ViewRelated/System/Notices/NoticePresenter.swift
```diff
@@ -219,6 +216,9 @@ class NoticePresenter: NSObject {
         addNoticeContainerToPresentingViewController(noticeContainerView)
         addBottomConstraintToNoticeContainer(noticeContainerView)
 
+        // At regular width, the notice shouldn't be any wider than 1/2 the app's width
+        noticeContainerView.noticeWidthConstraint =             noticeView.widthAnchor.constraint(equalTo: noticeContainerView.widthAnchor, multiplier: 0.5)
```
Gio2018, line 220: probably these extra spaces are not here on purpose?

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
```diff
@@ -42,6 +42,10 @@ class NoticeView: UIView {
         configureLabels()
         configureForNotice()
 
+//        if notice.sourceView != nil {
+//            configureArrow()
+//        }
```
Gio2018, line 47: Some more commented out code here

WordPress/Classes/ViewRelated/System/Notices/NoticeAnimator.swift
bjtitus: I ended up adding a `Transition` type because I don't really love the idea of the call sites having to pass in 0 or full (1) values each time. I kind of like sticking with the old "onscreen" and "offscreen" states.

Let me know what you think!

Gio2018: this is good! Another option could be 
`func offScreenState(_:, _:, _:) {
    state(_:, alpha: WPAlphaZero...)
}`

but it's just a matter of preference

ScoutHarris:
Hey @bjtitus @osullivanchris .

I'm not sure if this is the final version of the tooltip, but my 2 cents - it's a persistent little bugger. 😄 
- It reappears _every time_ blog details is viewed.
- It doesn't time out.
- It can't be dismissed by tapping on it. It can only be dismissed by actually tapping the create button.

bjtitus:
@osullivanchris What do you think about adding a timeout or just dismissing the tooltip after it has been viewed once or x times? There isn't a terribly easy way to detect tapping on the tooltip itself at the moment, but that could also be added if the other solutions seem inadequate.

osullivanchris:
Haha appreciate the candid feedback @ScoutHarris 

I think when we addressed this on Android it felt like it was easy for someone to dismiss. Although it included tapping on the tooltip as well as the FAB. So if I hate the FAB for some reason, here I might feel forced to do it.

I'm open to dismissing after X times. Whatever you think is the easiest way to define that. From a UX point of view, we just want to make someone aware of it. And I don't want to bug them for more than a couple of times.

bjtitus:
Sorry, I forgot to follow up with comments about this. I had not included a user defaults implementation in this PR because we were still a little unsure about exactly how it should be removed. I'll be submitting a follow up with some thoughts and an initial implementation.

Issue #14021, opened 2020-04-29
Change the order of the options 


yaelirub:
The order should match dotcom: show visibility as first row and scheduling as second

Issue #13986, opened 2020-04-24
[GlobalStep] Empty scroll space after changing the date in the publishing options window.


wptester9845:
### Description
When the user goes to create a new post, selects publish and then changes the date in the publishing options, it can be observed the user is now able to scroll far down below the ‘Schedule now’ button, causing the publishing options window to appear empty.

### Expected behavior
Empty scroll space is unnecessary and cannot be seen.

### Actual behavior
Possible to see a lot of empty scroll space.

### Steps to reproduce the behavior
1.	Install WordPress 14.7
2.	Login to a valid account.
3.	Select the new post button.
4.	Enter any content and select ‘Publish’
5.	Change the post date.
6.	Scroll down the publishing options window.

##### Tested on the following:
iPhone 6S Plus (13.4.1)
iPhone X (12.3.1)

#### Please see the attached video for reference.
[publishoptions_empty_space.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4528059/publishoptions_empty_space.zip)

#### Submitted by:
Oliver Howe

PR #14034, opened 2020-04-30
Change options order and fix label and scroll issue

### LINKED ISSUE #14020: Change "Publish" to "Publish date" in scheduling row
Change should be made in Post Settings as well

leandroalonso:
Fixes #14020, #14021, #13986 (phew!)

#### Order (#14021)

1. Create a post
2. Tap Publish
3. Check that the order in the Nudges view is: Visibility, Publish Date and Tags

#### Label (#14020)

1. Create a post
2. Tap Publish
3. Check that the option that schedules a post now reads "Publish Date"

#### Scroll (#13986)

1. Create a post
2. Tap Publish
3. Tap "Publish Date"
4. Dismiss the calendar
5. Scroll the Nudges view and check that there's no additional empty scroll area

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
+++ b/WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
@@ -648,7 +648,7 @@ FeaturedImageViewControllerDelegate>
             
             cell.detailTextLabel.text = [self.postDateFormatter stringFromDate:self.apost.dateCreated];
         } else {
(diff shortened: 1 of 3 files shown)


yaelirub:
@leandroalonso , thank you so much for the quick turnaround on this! 
Code looks good and everything works as expected. 

I did notice something that I think is outside the scope of this PR:
When you schedule a post via nudges view or via post settings and then modify back to "publish immediately" the post status changes to "published" (before actually publishing)

PR #14037, opened 2020-04-30
New Reader: Self Hosted Sites Login and Offline Experience

Gio2018:
Ref #13580

Adds support for offline and login with self hosted sites to the new Reader.

#### Self hosted sites

<p align="center">
Self hosted experience
</p>

#### Note: the The empty sites screen in the 'Filter Sheet' for self hosted login is not there yet, you would see an indefinite ghost for now. Here is the final design for it

<p align="center">
Filter empty sites
</p>

#### Offline
- If the users go offline after successfully signing in and before ever accessing the reader, and then access the reader, they will see the following screen

<p align="center">
Reader main offline
</p>

- ~~if they tap 'Retry' after going back online, the normal Reader experience will be presented~~
 - The 'Retry' button has been removed, once users go back online, Reachability should detect the change and the Reader should load automatically

- if users access the reader when online and then go offline, will have the cached content accessible when offline. If they had not accessed a particular tab, that tab will look like the following picture

<p align="center">
Reader menu offline
</p>

- ~~after going back online, a pulldown will refresh the content and present the normal screen.~~
 - In this case too, the content should load automatically when they go back online.
 - Saved posts are local content, so the existing saved posts (or the absence of them) will not be affected by network conditions: saved posts, or the empty screen for saved posts, will be the same regardless the network connection

To test:

##### Self Hosted Sites
- log in with a Self Hosted Site
- verify that you see the three tabs 'Following', 'Discover' and 'Saved'
- verify you can access tags from the 'Filter' sheet, and the overall experience looks like the animated gif above

##### Offline - please test on a real device

 - Make sure you're online, fresh install the app and login
 - Go offline
 - Go back to the app and in the reader, and check that the screen looks like the above picture with the 'Retry' button
 - Go back online, and verify that ~~the 'Retry' button loads the reader~~ the posts start loading
 - Go back offline, and verify that the loaded sections are still accessible (scrolling down some images might not be displayed)
 - Verify that sections that you had not accessed present the second offline screen above
 - Go back online and to the Reader, and verify that ~~pull to refresh obtains the content in the sections that were not loaded~~ the content starts loading
 - Optionally, go back offline and check that the content you expected is still there

PR submission checklist:

- [x] ~~I have considered adding accessibility improvements for my changes.~~ To be added in a separate PR
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ Not yet released


Gio2018:
@osullivanchris would you please have a look at this (build 27931)? Thanks so much!!

osullivanchris:
> @osullivanchris would you please have a look at this (build 27931)? Thanks so much!!

Hey I've downloaded it and had a play with it! Its looking good! 

One issue I noticed is that I'm getting a perpetual loading state for 'Sites' in the filter bottom sheet, I'm never getting an empty state.

Gio2018:
> One issue I noticed is that I'm getting a perpetual loading state for 'Sites' in the filter bottom sheet, I'm never getting an empty state.

hey @osullivanchris thanks for looking at this!! That empty screen is yet to be added, so that's expected for now.

WordPress/Classes/ViewRelated/Reader/ReaderHelpers.swift
```diff
@@ -276,11 +292,31 @@ extension ReaderHelpers {
             }
             return true
         }
+
+        guard !mutableItems.isEmpty else {
```
bjtitus, line 296: Should this guard just move up to the top? Sorting an empty list won't do anything anyway.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -487,12 +493,14 @@ import WordPressFlux
...
-            header.centerXAnchor.constraint(equalTo: tableView.centerXAnchor).isActive = true
-            header.widthAnchor.constraint(equalTo: tableView.widthAnchor).isActive = true
-            header.topAnchor.constraint(equalTo: tableView.topAnchor).isActive = true
+        header.centerXAnchor.constraint(equalTo: tableView.centerXAnchor).isActive = true
+        let headerWidthConstraint = header.widthAnchor.constraint(equalTo: tableView.widthAnchor)
+        if FeatureFlag.newReaderNavigation.enabled {
+            headerWidthConstraint.priority = UILayoutPriority(999)
         }
+        headerWidthConstraint.isActive = true
+        header.topAnchor.constraint(equalTo: tableView.topAnchor).isActive = true
```
bjtitus, line 503: Probably not a huge deal but it might be better to activate these in bulk since it is supposed to be faster for the autolayout engine:

```suggestion
        let centerConstraint = header.centerXAnchor.constraint(equalTo: tableView.centerXAnchor)
        let topConstraint = header.topAnchor.constraint(equalTo: tableView.topAnchor)
        let headerWidthConstraint = header.widthAnchor.constraint(equalTo: tableView.widthAnchor)
        if FeatureFlag.newReaderNavigation.enabled {
            headerWidthConstraint.priority = UILayoutPriority(999)
        }

        NSLayoutConstraint.activate([
            centerConstraint,
            headerWidthConstraint,
            topConstraint
        ])
```

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -162,6 +164,12 @@ extension ReaderTabView {
         containerView.translatesAutoresizingMaskIntoConstraints = false
         childController.view.translatesAutoresizingMaskIntoConstraints = false
 
+        controller.children.forEach {
+            $0.willMove(toParent: nil)
```
bjtitus, line 168: According to the docs, you shouldn't have to do this since `removeFromParent` is called below:

> The removeFromParent() method automatically calls the didMove(toParent:) method of the child view controller after it removes the child.

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621405-didmove

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -162,6 +164,12 @@ extension ReaderTabView {
...
+            $0.view.removeFromSuperview()
+            $0.removeFromParent()
```
bjtitus, line 170: I think you can use [WordPressUI's `remove()` method](https://github.com/wordpress-mobile/WordPressUI-iOS/blob/70a05f0ac69551c832e834d0d0c693936b3d27aa/WordPressUI/Extensions/UIViewController%2BChildViewController.swift#L19) here. Unless we want to shy away from introducing another dependency on that library.

```suggestion
            $0.remove()
```

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabViewModel.swift
```diff
@@ -170,13 +170,13 @@ extension ReaderTabViewModel {
 extension ReaderTabViewModel {
 
     func makeChildContentViewController(at index: Int) -> ReaderContentViewController? {
-        guard index < tabItems.count else {
-            return nil
+        guard tabItems.indices.contains(index) else {
```
bjtitus, line 173: I don't know if it's clearer to use the [`[safe:]` subscript method](https://github.com/wordpress-mobile/WordPress-iOS/blob/0aa1510fffbd18a714dd390052cda8fe2826545c/WordPress/Classes/Extensions/CollectionType%2BHelpers.swift#L15) here:

```suggestion
        guard let tabItem = tabItems[safe: index] else {
```

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
Gio2018: 👍

Gio2018:
@ScoutHarris , @bjtitus the PR is ready for another review, whenever you can!. Note that I have removed the 'Retry' button and use Reachability to reload content (more details in the updated comment). Thanks!!!

Gio2018:
> I do see one one issue with the Saved tab:
> 
> * Logged in.
> * Went to Reader.
> * Went offline.
> * Went to Saved tab.
> * The no results view is displayed instead of the no connection view.

Hey @ScoutHarris , thanks for re-reviewing!

This is on purpose, since the saved posts are only local and should not be affected by network conditions. 
If you feel that this might be confusing to the user, we might still show the no connection even for the Saved Posts. Maybe @osullivanchris can advise on this point?

ScoutHarris:
> This is on purpose, since the saved posts are only local and should not be affected by network conditions.

Oh, right. Yea, that makes sense now. I'll approve this then, and if @osullivanchris requests a change it can always be done later if you like.

Gio2018:
> Oh, right. Yea, that makes sense now. I'll approve this then, and if @osullivanchris requests a change it can always be done later if you like.

Sounds great! Thanks!!

osullivanchris:
Sounds fine to me!

PR #14039, opened 2020-05-01
Restores an earlier copy of Localizable.strings from develop

aerych:
In #13799 included manual changes to `Localizable.strings`. The changes were rolled back but the file remained in the PR (maybe due to a different hash?).  

In this PR, we're restorying a copy of `Localizable.strings` from `develop` immediately prior to #13799.  See p1588344410269100-slack-ios-osx-dev for more context.


aerych:
Thank you @jleandroperez !

Issue #13788, opened 2020-04-01
iA Navigation: Startup Modal with link to Announcement Blog Post


bjtitus:
Add a modal introducing the Create FAB, Me, and Reader navigation changes on startup.

https://ministryofmagicp2.wordpress.com/2020/01/28/ia-project-my-site-ios/#comment-571

designsimply:
I removed the milestone from this issue since 14.6 has already shipped and I think tracking this issue in the Information Architecture Refresh project board makes the milestone on the issue unnecessary—the PR is the place to add a milestone, once a PR is avail. 👍

PR #14042, opened 2020-05-01
Adds IA Navigation announcement modal

bjtitus:
#13788 

Adds a modal to announce new IA changes. I have a couple of questions:

* Do we need to track that this modal was shown? Do we need to track when the "Read More" button is tapped?
* Do we need more advanced logic to hide the modal after a certain amount of time has passed after the release so that we aren't showing users this prompt 2 years from now? Alternatively, we could remove it in a future release.

<a href="https://user-images.githubusercontent.com/3250/80828524-a789a500-8ba2-11ea-9fa6-681883e60980.png"></a>

To test:
- Open the app and notice the IA Navigation announcement modal
- Tap on the "Read more" button to navigate to the wordpress blog (blog URL to be added separately once post is created)

#### Resetting the modal state

To reset the User Defaults key so that the modal appears again pause using the debugger and run the following command:

```
e -l swift -- import WordPress; UserDefaults.standard.createButtonAlertWasDisplayed = false
```

After doing this, the modal should appear on next launch.

PR submission checklist:


loremattei:
Hey! I'm moving this to 14.9 since 14.8 has been cut. Please, feel free to ping me if you want this to make it to 14.8.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+FancyAlerts.swift
```diff
@@ -153,6 +155,30 @@ extension BlogDetailsViewController {
...
+
+        guard !UserDefaults.standard.createButtonAlertWasDisplayed else {
+            return
+        }
+
+        UserDefaults.standard.createButtonAlertWasDisplayed = true
+
+        let alert = FancyAlertViewController.makeCreateButtonAnnouncementAlertController { [weak self] (controller) in
+            controller.dismiss(animated: true)
+            if let url = URL(string: "https://wordpress.com/blog/") {
```
Gio2018, line 171: I wonder if it would be feasible, once we remove the placeholder and use the actual url, to retrieve it from the backend in some way instead of hardcoding it

bjtitus: Yeah, I don't know what the options are here. A remote config store like Firebase could help with this and turning on/off the announcement.

michelleweber:
Can you tell me a little more about the changes? Right now, it reads kinda like "Simplified navigation: We've made navigation more simple," which doesn't tell the user anything -- ideally, we can give them some little nugget of info here that hints at *how* it's easier, with all the detail in the post. If you tell me a little more about the changes (or just link me to a post somewhere that talks about it), I can pull something out.

Also, for the buttons, I might do "Thanks!" or "Got it!" for the top button, instead of "Okay."

osullivanchris:
Thanks for the quick response @michelleweber ! 

Good points. I don't have a good enough draft of the full post to share yet. I'll try get it in shape over the next day. I haven't been finding it easy to write.

The really obvious, specific changes, that a user might notice

- We've consolidated tabs/features so that there's now only 3 tabs to go to find things, rather than 5 with overlaps
- There's a big pink Floating Action Button as a shortcut to create content for your site
- We removed a layer of depth to the reader navigation and put it in tabs, so its quicker to get to content and switch between types of content

Now to just say that in very short bullet points 😄

osullivanchris:
@bjtitus also thinking that it does look a little strange getting chopped off. What do you think about doing the fade out like we had in a similar announcement? Here's how it could look, and a pdf asset. I've managed to create it with transparency rather than a color gradient, so it should work on dark/light mode and not matter what color is behind it.

Screenshot 2020-05-06 at 13 28 36

[ia-announcement-v2.pdf](https://github.com/wordpress-mobile/WordPress-iOS/files/4586766/ia-announcement-v2.pdf)

michelleweber:
@osullivanchris that's super helpful. And we don't have to say all of it, thanks to the blog post! We could do something like:

**Streamlined navigation**

Now there are fewer and better-organized tabs, posting shortcuts, and more, so you can find what you need fast.

(Got it!)
(Learn more)

osullivanchris:
@michelleweber forgot to reply before, but that all sounds great to me. I'm working on the blog post now so I'll send that for review when the draft is ready. Thanks!

bjtitus:
@ScoutHarris Could you take one more look at this?

I've made a change to only show the announcement modal on a version upgrade.

### Testing
- Delete the app from your device (to clear any user defaults)
- Install this new version
- Dismiss the Notifications modal, as most all users will have already done when upgrading.
- Notice that a modal does not appear on subsequent launches
- Change the App version from 14.8 to any other version (I used 14.9)
- Notice that the modal does appear after upgrading
- Make sure that the modal does not appear on subsequent launches after upgrading

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+FancyAlerts.swift
```diff
@@ -156,7 +156,7 @@ extension BlogDetailsViewController {
     }
 
     private func shouldShowCreateButtonAnnouncement() -> Bool {
-        return !UserDefaults.standard.createButtonAlertWasDisplayed
+        return AppRatingUtility.shared.didUpgradeVersion == true && !UserDefaults.standard.createButtonAlertWasDisplayed
```
ScoutHarris, line 159: The `== true` isn't necessary.

PR #14045, opened 2020-05-04
Gutenberg - Starter Page Templates Tooltip

geriux:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2169

`Gutenberg Mobile PR` -> https://github.com/wordpress-mobile/gutenberg-mobile/pull/2216
`Gutenberg PR` -> https://github.com/WordPress/gutenberg/pull/21974
`WordPress Android PR` -> https://github.com/wordpress-mobile/WordPress-Android/pull/11807

To test:

- Download and install the build.
- Open the app and go to Pages.
- Tap on the add a new page button.
- **Expect** to see the templates picker with the tooltip on top.
- Scroll through the template options.
- **Expect** the tooltip to stay visible.
- Tap on the title of the page.
- **Expect** to see the tooltip fade out.
- Close the app.
- Open the app and go to Pages.
- Tap on the add a new page button.
- **Expect** to see the templates picker **without** the tooltip.

PR submission checklist:


WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -157,6 +158,15 @@ class GutenbergSettings {
...
 
+    /// True if it should show the tooltip for the starter page templates picker
+    func getStarterPageTemplatesTooltipShown() -> Bool {
+        return database.bool(forKey: Key.starterPageTemplatesTooltipShown)
+    }
+
+    func setStarterPageTemplatesTooltipShown(_ tooltipShown: Bool) {
+        database.set(tooltipShown, forKey: Key.starterPageTemplatesTooltipShown)
+    }
+
```
chipsnyder, line 169: This suggestion is a minor style change, in Swift, we usually omit the `get` keyword for methods like these such as in:
https://github.com/wordpress-mobile/WordPress-iOS/blob/617a7a3092c8184bc14bf9ef572297a58ab021f0/WordPress/Classes/Utility/Editor/GutenbergSettings.swift#L96-L102

So maybe something like this would be more inline with the pattern you have if you wanted to keep it as function calls:
```suggestion
    /// True if it should show the tooltip for the starter page templates picker
    func starterPageTemplatesTooltipShown() -> Bool {
        return database.bool(forKey: Key.starterPageTemplatesTooltipShown)
    }

    func setStarterPageTemplatesTooltipShown(_ tooltipShown: Bool) {
        database.set(tooltipShown, forKey: Key.starterPageTemplatesTooltipShown)
    }

```

---

Another option that is more "Swift-like" would be to write it as  computed property:
```suggestion
    /// True if it should show the tooltip for the starter page templates picker
    var starterPageTemplatesTooltipShown: Bool {
        get {
            database.bool(forKey: Key.starterPageTemplatesTooltipShown)
        }

        set {
            database.set(newValue, forKey: Key.starterPageTemplatesTooltipShown)
        }
    }

```

You could then access it with `GutenbergSettings().starterPageTemplatesTooltipShown` and `GutenbergSettings().starterPageTemplatesTooltipShown = tooltipShown `

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -614,6 +614,14 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
         self.present(controller, animated: true)
     }
+
+    func gutenbergDidRequestGetStarterPageTemplatesTooltipShown() -> Bool {
+       return GutenbergSettings().getStarterPageTemplatesTooltipShown()
+    }
+
+    func gutenbergDidRequestSetStarterPageTemplatesTooltipShown(_ tooltipShown: Bool) {
+       GutenbergSettings().setStarterPageTemplatesTooltipShown(tooltipShown)
+    }
```
chipsnyder, line 624: This style of calling `GutenbergSettings` seems to be more in line with refactoring these methods to be static methods or maybe share the instance of `GutenbergSettings` somewhere. Right now it looks like it's only used in one spot as a lazy-loaded var
https://github.com/wordpress-mobile/WordPress-iOS/blob/617a7a3092c8184bc14bf9ef572297a58ab021f0/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L226-L228

This might be a good time to refactor that to share the instance in this file such as:
```
lazy var gutenbergSettings: GutenbergSettings = { 
     return GutenbergSettings()
 }() 
```

Then these methods become:
```suggestion

    func gutenbergDidRequestGetStarterPageTemplatesTooltipShown() -> Bool {
       // return gutenbergSettings.starterPageTemplatesTooltipShown // depending on if you use the other suggestion or not
       return gutenbergSettings.getStarterPageTemplatesTooltipShown()
    }

    func gutenbergDidRequestSetStarterPageTemplatesTooltipShown(_ tooltipShown: Bool) {
        // gutenbergSettings.starterPageTemplatesTooltipShown = tooltipShown // depending on if you use the other suggestion or not
       gutenbergSettings.setStarterPageTemplatesTooltipShown(tooltipShown)
    }
```

chipsnyder:
Looks good functionally 👍. 
Just want to get your thoughts on a couple of style-related things.

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
geriux: TIL! Thank you so much for these suggestions Chip! I'll update the code 🙌

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
geriux: Sounds good! I'll refactor that part and update the code.

PR #14060, opened 2020-05-06
Deselect start-over cell on settings screen

### LINKED ISSUE #13307: [GlobalStep] Start Over button remains highlighted after the user navigates back to the Settings screen
### Description
While in a Site’s Settings, pressing the Start Over button and closing the displayed page causes the Start Over to remain highlighted.

### Reproduction Rate
4/4 100%

### Expected behavior
The Start Over button should only be highlighted while the user is pressing it.

### Actual behavior
The Start Over button remains highlighted after being pressed.

1. Install WordPress 14.1
2. Log in to an account.
3. Navigate to the “My Sites” tab.
4. Navigate to Settings.
5. Press the “Start Over” button.
6. Close the “Empty Site” page.
##### Tested on the following
iPhone SE (13.1.3)

#### Please see the attached video for more information
[iOSStartOver.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4127842/iOSStartOver.zip)

### Submitted by:
Luis Pimenta

guarani:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/13307

Fixes a visual glitch where after returning from the Start Over screen (Site | Settings | Start Over),  the tapped cell remains highlighted.

### Testing steps

1. Open the Start Over modal (Site | Settings | Start Over)
2. Tap the close button to close the modal
3. Observe that the tapped cell is no longer highlighted.

### Demo

<kbd></kbd>

PR submission checklist:


bjtitus:
@Guarani Do you think it's worth only doing this for the case where the modal is presented?

By always deselecting, we lost the default `clearsSelectionOnViewWillAppear` behavior that animates when returning back to the table view.



I wish there was a better way to handle the lack of appearance methods. I tried adding a presentation controller delegate and manually calling its `didDismiss` method but it takes SO MUCH extra code to do that for such a simple thing.

guarani:
> By always deselecting, we lost the default clearsSelectionOnViewWillAppear behavior that animates when returning back to the table view.

That's an excellent point that I overlooked: the animation helps the user know which option they're coming back from.

> @guarani Do you think it's worth only doing this for the case where the modal is presented?

I'm considering this approach as it fixes almost all issues:
- ✅ the Start Over cell doesn't remain selected after closing the modal (as it did before)
- ✅ the other cells don't lose their default `clearsSelectionOnViewWillAppear` behavior 
But, 
- ⚠️ upon closing the Start Over modal, there's no visual cue as to which cell was pressed

loremattei:
Hey! I'm moving this to 15.0 because 14.9 has been cut. If you want this to make it to 14.9, please feel free to ping me.

guarani:
I think there is a solution to interactively animating the Start Over cell when the modal is dismissed, but it's quite complex and perhaps not worth the effort.

Here's how it works. Inspired by this post, I went ahead and used `UIAdaptivePresentationControllerDelegate` to listen for the modal sheet closing:

```
- (void)presentationControllerWillDismiss:(UIPresentationController *)presentationController {
    
    // The Advanced > Start Over cell opens a modal (as opposed to pushing a vc), so cell selection must be handled manually.
    // See: http://mikeabdullah.net/thl-diary-17.1.html
    NSIndexPath *selectedIndexPath = [self.tableView indexPathForSelectedRow];

    if (selectedIndexPath &&
        [[self tableSections][selectedIndexPath.section] isEqual: @(SiteSettingsSectionAdvanced)] &&
        selectedIndexPath.row == SiteSettingsAdvancedStartOver) {
        
        [self.transitionCoordinator animateAlongsideTransitionInView:self.tableView animation:^(id<UIViewControllerTransitionCoordinatorContext> _Nonnull context) {
            [self.tableView deselectRowAtIndexPath:selectedIndexPath animated:YES];
        } completion:^(id<UIViewControllerTransitionCoordinatorContext> _Nonnull context) {
            if (context.isCancelled) {
                [self.tableView selectRowAtIndexPath:selectedIndexPath animated:YES scrollPosition:UITableViewScrollPositionNone];
            }
        }];
    }
}
```

This is called when the modal sheet is dismissed, and the if-statement checks that the selected cell is indeed the Start Over cell. Then the deselection animation is done in sync with the dismiss transition animation.

 A few caveats:
1. This works great when the user drags the modal sheet to dismiss – but if the close button is press, the modal is closed programmatically and we would need to use another mechanism for listening to the event (e.g. a delegate protocol to communicate from the modal view controller to this settings screen)
2. Bottom sheets changed on iOS 13 so we'd need additional code for other versions.
3. This adds even more code to an already large view controller ~1400 lines (this could be mitigated but the gain here is small)

---

How about we just fix deselect the cell as soon as it's tapped and lose this animation?

guarani:
This should be ready for a re-review @bjtitus.

bjtitus (approved):
👍 

I think this works fine.

I wonder if something general like this would work:

```objectivec
- (void)presentViewController:(UIViewController *)viewControllerToPresent animated:(BOOL)flag completion:(void (^)(void))completion {
    [super presentViewController:viewControllerToPresent animated:flag completion:completion];
    
    // UIKit doesn't automatically manage cell selection when a modal presentation is triggered,
    // which is the case for Start Over when there's no paid plan, so we deselect the cell manually.
    if (viewControllerToPresent.modalPresentationStyle != UIModalPresentationFullScreen) {
        [self.tableView deselectRowAtIndexPath:self.tableView.indexPathForSelectedRow animated:NO];
    }
}

This may be too fragile to ship so I think what you have is probably a good approach to close this issue.
```

guarani:
Your code looks great, it is more generic. For now though I'll just leave it as-is I think, because since this is a huge VC, it might not be obvious to devs where the change in behavior comes from.

🙇 thanks for the review, @bjtitus!

PR #14061, opened 2020-05-06
Epilogues: code cleanup

ScoutHarris:
Ref #13413, #13939

This is just some code cleanup in the Signup and Login epilogues. Primarily:
- Remove usage of NUX view controllers.
- Cleanup private methods and properties.
- Remove some unnecessary code.

There should be no functional or visual changes.

---
To test:
- Login.
  - Verify the view appears as it should.
  - Verify the `Done` button dismisses the view.
  - Verify the `Connect another site` displays the site address view.
- Signup.
  - Verify the view appears as it should.
  - Verify the view does not rotate in iPhone landscape.
  - Select `Username`. Verify the `Help` nav bar button opens the `Support` view.
  - Go back to the epilogue. Verify the blue nav bar does not appear.

---

```
        fatalError()
    }
    
//        }
}
```

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
+++ b/WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
@@ -115,66 +115,7 @@ class SignupEpilogueCell: UITableViewCell {
         cellField.backgroundColor = .basicBackground
     }
 
-    // MARK: - Private behavior
-
-    private func configureForPassword() {
-        let isPassword = (cellType == .password)
-        cellLabel.isHidden = isPassword
-
-        cellField.isSecureTextEntry = isPassword
-        cellField.showSecureTextEntryToggle = isPassword
-        cellField.textAlignment = isPassword ? .left : .right
-        cellField.textColor = isPassword ? .text : .textSubtle
-
-        cellFieldLeadingConstraintWithLabel.isActive = !isPassword
-        cellFieldLeadingConstraintWithoutLabel.isActive = isPassword
-        cellFieldTopConstraint.constant = isPassword ? passwordTopMargin : 0
-    }
-
-    private func configureAccessibility(for cellType: EpilogueCellType) {
-        if cellType == .username {
-            accessibilityTraits.insert(.button) // selection transitions to SignupUsernameViewController
-            isAccessibilityElement = true       // this assures double-tap properly captures cell selection
-        }
-
-        switch cellType {
-        case .displayName:
-            cellField.accessibilityIdentifier = "Display Name Field"
-        case .username:
-            cellField.accessibilityIdentifier = "Username Field"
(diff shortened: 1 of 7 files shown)


WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
```diff
@@ -115,9 +115,44 @@ class SignupEpilogueCell: UITableViewCell {
...
     }
 
-    // MARK: - Private behavior
+}
+
+extension SignupEpilogueCell: UITextFieldDelegate {
+
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        if let cellType = cellType, cellType == .displayName || cellType == .password {
+            let updatedText = NSString(string: textField.text!).replacingCharacters(in: range, with: string)
```
mindgraffiti, line 124: Should textField.text be force unwrapped? I thought `shouldChangeCharactersIn range:` could be called after a user deletes all the text from the textField as well.

WordPress/Classes/ViewRelated/NUX/SignupEpilogueViewController.swift
```diff
@@ -2,7 +2,7 @@ import SVProgressHUD
 import WordPressAuthenticator
 
 
-class SignupEpilogueViewController: NUXViewController {
+class SignupEpilogueViewController: UIViewController {
```
mindgraffiti, line 5: 🎉

WordPress/Classes/ViewRelated/NUX/SignupUsernameTableViewController.swift
```diff
@@ -2,7 +2,7 @@ import SVProgressHUD
 import WordPressAuthenticator
 
 
-class SignupUsernameTableViewController: NUXTableViewController, SearchTableViewCellDelegate {
+class SignupUsernameTableViewController: UITableViewController, SearchTableViewCellDelegate {
```
mindgraffiti, line 5: 🎉

WordPress/Classes/ViewRelated/NUX/SignupEpilogueCell.swift
ScoutHarris: > Should textField.text be force unwrapped?

Ah yes, that was from the original implementation. Tis been removed.

> I thought `shouldChangeCharactersIn range:` could be called after a user deletes all the text from the textField as well.

I'm not quite sure what this means. This method is called every time a character is added/deleted, so that would include when all characters are deleted. If I missed something let me know.

Also, while testing this change, I found a pre-existing bug (https://github.com/wordpress-mobile/WordPress-iOS/issues/14067), which I'll fix separately.

Issue #13892, opened 2020-04-14
Unresolved Core Data save error (Derived Context) in SourcePostAttribution.post


loremattei:
**Sentry Url**: https://sentry.io/share/issue/1c7329b09b694798b6a0df9eea315e1f/
**User Count**: 541
**Count**: 6356
**First Release**: 13.4.1.1
**First Seen**: 2019-11-05T22:56:33Z
**Last Release**: 14.5.0.2
**Last Seen**: 2020-04-14T03:16:03Z
**24 Hours**: 10
**30 Days**: 380

**Stacktrace**:
```
Unresolved Core Data save error (Derived Context): Non-optional property with a nil value on SourcePostAttribution.post
    at null.__exceptionPreprocess(null:null)
    at null.objc_exception_throw(null:null)
    at null.-[NSException raise](null:null)
    at null.ContextManager.handleSaveError(ContextManager+ErrorHandling.swift:69)
    at null.ContextManager.handleSaveError(<compiler-generated>:null)
    at null.-[ContextManager internalSaveContext:](Classes/Utility/ContextManager.m:328)
    at null.__65-[ContextManager saveDerivedContext:andWait:withCompletionBlock:]_block_invoke.65(Classes/Utility/ContextManager.m:153)
    at null.developerSubmittedBlockToNSManagedObjectContextPerform(null:null)
    at null._dispatch_client_callout(null:null)
    at null._dispatch_lane_serial_drain$VARIANT$mp(null:null)
...
(4 additional frame(s) were not displayed)
```

yaelirub:
SourcePostAttribution is being used in the reader so I'd like to bump this one's priority. 
We've been dealing with context saving issues on the derived and main for some time (for example: https://github.com/wordpress-mobile/WordPress-iOS/issues/13429). I'd love a fresh set of eyes on this. @emilylaguna , would you mind digging in seeing if you can find something?

jleandroperez:
### Few notes on this one:
- **ReaderPost** has a N to 1 relationship with **SourcePostAttribution**
- Deletion rule is set to cascade (from ReaderPost > SourcePostAttribution)
- Inverse deletion rule is nullify

Deleting a ReaderPost should, effectively, get rid of the SourcePostAttribution.
My theory on this one would be a race condition in the **ReaderPost / SourcePostAttribution** sync'ing code.

[...]

Issue #13867, opened 2020-04-10
Unresolved Core Data save error (Derived Context): Merge policy failed - unable to complete merging on null.null


loremattei:
**Sentry Url**: https://sentry.io/share/issue/8c458ae1669e407594703b6b5341b2ad/
**User Count**: 514
**Count**: 4696
**First Release**: 13.4.1.1
**First Seen**: 2019-11-05T16:34:05Z
**Last Release**: 14.5.0.2
**Last Seen**: 2020-04-10T17:48:59Z
**24 Hours**: 1115
**30 Days**: 2816

yaelirub:
Triage: priority high by impact and has been for a while. There are many tickets floating around about save error on either derived or main context. Most of them you can find here:
https://github.com/wordpress-mobile/WordPress-iOS/search?q=save+context+error&type=Issues

@mindgraffiti , are you going to tackle this during Groundskeeping?

mindgraffiti:
@yaelirub thanks for the link! That's helpful. 

I made the mistake of thinking "the [landscape layout corrections](https://github.com/wordpress-mobile/WordPress-iOS/issues/13792) will only take a day" and thought I would get to start researching this issue today. Famous last words. 

I plan to take on this task if I complete the landscape layout issues in time.

yaelirub:
Gotcha. Could you please comment here either way once you know?

mindgraffiti:
@yaelirub can confirm. The landscape layouts are moving very slowly. Some of the views haven't been [edited for 3 years](https://github.com/wordpress-mobile/WordPress-iOS/tree/develop/WordPress/Classes/ViewRelated/Themes). I'm not able to look into this crash.

designsimply:
This crash came up in a 1-start app review and I used the username to see if I could find a crash report from them in Sentry and this issue came up as a result of that search.

> Super buggy by evoskuil – Apr 26, 2020 - Version 14.6
> Gets in to a mode where it fails to update a post and will then crash on every attempt until the app is deleted and restored. The new editor cannot view posts with either the old mobile editor or the desktop. Fails to recognize certain basic edits created on the desktop version. You can first create a paragraph on the mobile app and then edit it but don’t try creating paragraphs in the reverse order.

rezzap:
I think we have a report from a user who's having trouble with this same crash. The logs they sent include: 

[...]

PR #14068, opened 2020-05-06
Issues/13867 null nulls

aerych:
Refs #13867 
Refs #13892
This is a _very_ experimental patch that attempts to do two things:

1) Improve error messages logged when there is a Core Data merge error. Currently a merge error is appended with a message that reads "on null.null".  This isn't helpful as the message is intended to reference the object and property that caused the error, something that is not relevant for a merge conflict.  With this change, instead the source entity is logged so its at least easy to see which model is having the merge problem. 

2) We're theorizing a lot of the core data issues we're seeing is due to race conditions between saves happening across multiple derived contexts along with the main context. Somewhere a context is getting out of sync with the rest causing an error.  In this patch we're tweaking things to create one and only one derived context which will be used whenever a derived context is requested.  We're also setting all child contexts of the mainContext to auto merge changes from their parent.  This, in theory, should keep things from getting out of sync. 🤞 

To test:
Run tests. Make sure they all pass. 
Use the app. Like, seriously put it through the passes. Do things that are going to cause multiple background save operations in a short period of time. Monitor performance. Ensure the app performs as expected. Ensure there are no crashes.  Look particularly at media operations. 

Gonna ping lots of smart folks to help test this one because 😬 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/ContextManager+ErrorHandling.swift
+++ b/WordPress/Classes/Utility/ContextManager+ErrorHandling.swift
@@ -89,10 +89,23 @@ private extension ContextManager {
     }
 
     func reasonForIndividualError(_ error: NSError) -> String {
-        let entity = entityName(for: error) ?? "null"
-        let property = propertyName(for: error) ?? "null"
         let message = coreDataKnownErrorCodes[error.code] ?? "Unknown error (domain: \(error.domain) code: \(error.code), \(error.localizedDescription)"
-        return "\(message) on \(entity).\(property)"
+        var messageDetail = ""
+
+        if error.userInfo.keys.contains(NSValidationObjectErrorKey) {
+            let entity = entityName(for: error) ?? "null"
+            let property = propertyName(for: error) ?? "null"
+            messageDetail = "on \(entity).\(property)"
+
+        } else if error.code == NSManagedObjectMergeError,
+            let conflicts = error.userInfo["conflictList"] as? [NSMergeConflict],
+            let conflict = conflicts.first,
+            let name = conflict.sourceObject.entity.name {
+
+            messageDetail = "source entity: \(name)"
+        }
+
+        return "\(message) \(messageDetail)"
     }
 
     func entityName(for error: NSError) -> String? {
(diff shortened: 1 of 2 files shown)


aerych:
@astralbodies @jleandroperez would be grand to have your keen eyes on this one.

jleandroperez:
Code looks great sir!! I do have few thoughts about sharing the same derivedContext all over. We have several `performBlockAndWait` calls across the app, annnnd we might risk deadlocks.

How about if we:
- Add a new property called... `backgroundContext` or `sharedDerivedContext` (former is cooler)
- And wire such context, service by service, while testing all the things?

It's waaaay (way) more effort, huge downside, though. Thoughts?

aerych:
> Add a new property called... backgroundContext or sharedDerivedContext (former is cooler)
And wire such context, service by service, while testing all the things

I think it's a grand idea.  This PR is intentionally rough and I think after we vet the approach it makes sense to spend the extra time doing what you suggest. :)  This patch is just a few lines of code but its crazy impactful. We definitely do not want to introduce new issues. 😬

leandroalonso:
@aerych two things:

First thing: some time ago I amended one flow that caused a crash (`*** Terminating app due to uncaught exception 'Unresolved Core Data save error (Main Context)', reason: 'Non-optional property with a nil value on Category.blog'`) here: #13483

I checked out your branch and reverted the code I did to fix this specific error — to check if the code here prevented the error there. Unfortunately, the crash persists.

However, this is a good way of exploring if our changes produce any result since we can indeed easily reproduce the concurrent issue (you can see more details on the mentioned PR).

Second, the first time I run this PR on a device it froze (probably due to the issues @jleandroperez mentioned). 😬

yaelirub:
Hey all,
I was able to reproduce this almost consistently by doing:

1. Upload gallery on an existing draft
2. quit the app (remove from background) before upload complete
3. launch the app -> blog posts and see the failed post
4. Hit "Retry" and quit the app before upload complete
5. Launch the app -> Blog posts see failed post
6. tap the post and remove the failed gallery
7. Quit the app
8. Relaunch the app, wait 5 seconds after splash screen and -> blog posts
9. See the post trying to upload media automatically
10. wait and see "Uploading media" changing briefly to "Local changes" and then "Uploading post"
See the crash

If it didn't work the first time, repeating these steps the second time for the same post worked 99% of the time. Not sure if this helps at all but I think the post status changes on step 10 might be an indicator to something done wrong in auto uploading

aerych:
Closing this one in favor of #14087 and because it was mostly a proof of concept.

PR #14085, opened 2020-05-08
[RNMobile] Support Theme Colors and Gradients

chipsnyder:
**Fixes:** https://github.com/wordpress-mobile/gutenberg-mobile/issues/1744

**Related PRs**
`gutenberg` https://github.com/WordPress/gutenberg/pull/22197
`gutenberg-mobile` https://github.com/wordpress-mobile/gutenberg-mobile/pull/2241

----

## Description

Adds support for theme defined colors and gradients to be injected into the editor from the mobile apps. 

Colors can then be displayed by the same mechanisms that utilize the settings lookup for colors and gradients. This is currently supported on Cover and Button blocks

## How has this been tested?

***Note*** Atomic sites don't seem to support the `wp/v2/themes` API changes yet so this can be tested with a free site or a self-hosted site
To help with testing this theme can be added to a self-hosted site: [twentytwenty-copy.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4657974/twentytwenty-copy.zip)

### Colors
1.) Select a theme with custom colors (such as TwentyTwenty)
2.) Create a post with blocks that have a custom color set (such as Cover or Button)
3.) Load editor on mobile 

**Expect** to see the custom color on the block

### Gradients
1.) Select a theme with custom gradients or [add gradients to a theme](https://github.com/WordPress/gutenberg/blob/12a6ee49c16dd6f327a8a141b6858197536da800/docs/designers-developers/developers/themes/theme-support.md#block-gradient-presets)
2.) Create a post with blocks that have a custom gradient set (such as Cover or Button)
3.) Load editor on mobile 

**Expect** to see the custom gradient on the block

### No Wifi
1.) Set up blocks for custom color and gradient
2.) View the post on mobile to cache the theme
3.) Turn off wifi to the device
4.) Reload the editor

**Expect** to see the custom colors on the blocks

### No Wifi - Force kill the app
1.) Set up blocks for custom color and gradient
2.) View the post on mobile to cache the theme
3.) Turn off wifi to the device
4.) Stop the app from running in the background
5.) Reload the editor

**Expect** to see the custom colors on the blocks

## Screenshots 
<kbd></kbd>

## Types of changes

New feature: Theme support

## PR submission checklist:


WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -268,6 +272,16 @@ class GutenbergViewController: UIViewController, PostEditor {
...
+
+        let themeSupportStore = StoreContainer.shared.editorTheme
+        themeSupportQuery = themeSupportStore.query(EditorThemeQuery(blog: post.blog))
+        themeSupportReceipt = themeSupportStore.onStateChange { [weak self] (_, state) in
+            DispatchQueue.main.async {
+                if let strongSelf = self, let themeSupport = state.editorTheme(forBlog: strongSelf.post.blog)?.themeSupport {
+                    strongSelf.gutenberg.updateTheme(themeSupport)
+                }
+            }
+        }
```
chipsnyder, line 284: I've considered moving these calls outside of the GutenbergViewController. Ultimately I liked it here better because:

- It's closer to the location where it is being used.
- It's a fast enough call to where the average case will have this loaded before the user sees the editor. 
- If on a significantly slow network and a user navigates quickly to the editor, then there would still need to be some hook to the Gutenberg editor to call update theme.
- This only matters for the initial load of the editor for a blog, or if the theme changes; because the theme will load from a cache if no changes were detected.

Con: 
- There is a hiccup where the event can fire while the editor is still loading. I put a workaround for that issue [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/14085/files#diff-ae5d7f8fb9cae1bc2a433c8133c06674R578).

chipsnyder:
Hey @SergioEstevao 

@pinarol and I thought you might be a good person to ask for in this review when you have time. Not sure how your review plate is right now though so let me know if you want me to ping someone else.

SergioEstevao:
Testing in iOS:

### Button
 
- [ ] Expect to see the custom solid color on the block
- [ ] Expect to see the custom gradient color on the block

### Cover
 
- [ ] Expect to see the custom solid color on the block
- [ ] Expect to see the custom gradient color on the block

### No Wifi

- [ ] Expect to see the custom solid color on the block
- [ ] Expect to see the custom gradient color on the block

### No Wifi - Force kill the app

- [ ] Expect to see the custom solid color on the block
- [ ] Expect to see the custom gradient color on the block

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
SergioEstevao: Can we move this code to an extension file: `GutenbergViewController+ThemeSupport`?
And have all the methods associated with theme support moved there?

WordPress/Classes/Stores/EditorThemeStore.swift
```diff
@@ -0,0 +1,126 @@
...
+                DDLogError("Error loading active theme: \(error)")
+            }
+        }
+    }
+
+    func processResponse(_ response: Any, for blog: Blog) {
+        guard let responseData = try? JSONSerialization.data(withJSONObject: response, options: []) else { return }
+        let themeSupports = try? JSONDecoder().decode([EditorTheme].self, from: responseData)
+
+        if let themeKey = EditorThemeStoreState.key(forBlog: blog),
```
SergioEstevao, line 116: To make it easier to read this code can we use guard statements instead of having these nested conditions?

WordPress/Classes/Models/EditorTheme.swift
```diff
@@ -0,0 +1,46 @@
...
+        case colors = "editor-color-palette"
+        case gradients = "editor-gradient-presets"
+    }
+
+    let colors: [[String: String]]?
+    let gradients: [[String: String]]?
+
+    init(from decoder: Decoder) throws {
+        let map = try decoder.container(keyedBy: CodingKeys.self)
+        self.colors = try? map.decode([[String: String]].self, forKey: .colors)
```
SergioEstevao, line 43: Are we using `try?` here because this key can be optional?

chipsnyder: Yeah if the colors or gradients are missing then the JSON will have `editor-color-palette: false` or `editor-gradient-presets: false` so `try?` here is to protect against the type change. 

It's then left as an optional to make sure we don't send an empty array to Gutenberg and override the default colors that we might want.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
chipsnyder: Yeah that sounds like a good plan.

chipsnyder: I ended up moving these to an extension in the same file with a comment. Moving it to a new file would have resulted in changing the access level to a few objects like `gutenberg` and the `themeSupportQuery` and `themeSupportReceipt` which I feel are better left as `private`

Let me know if this works for you.

chipsnyder:
@SergioEstevao This should be good to go again. I'll handle the Podfile.lock conflict once I have the real commit from the GB side.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -747,6 +753,10 @@ extension GutenbergViewController: GutenbergBridgeDataSource {
             "mentions": post.blog.isAccessibleThroughWPCom()
         ]
     }
+
+    func gutenbergEditorTheme() -> GutenbergEditorTheme? {
```
SergioEstevao, line 757: Could this method move to the extension also?

chipsnyder: Yup I can move that. My thought here was that since it's a required method of `GutenbergBridgeDataSource` that it belonged with that extension, but I can just leave a comment on the method to help signify that.

chipsnyder: This has been moved now.

chipsnyder:
> Looking good, feel free to merge after the merge conflict is solved.

🙇 Thank you @SergioEstevao

chipsnyder:
I'll plan to merge this as soon as the Gutenber Pr's are approved and merged

SergioEstevao:
Can we delay this merge until next week, we are in the process of cutting a release of GB Mobile and this functionality will not be part of the release. If you wish you can merge it on the after-release branch

chipsnyder:
Yup that shouldn't be a problem to wait. I'm still working on getting reviews for the other layers any way.

PR #14087, opened 2020-05-08
ContextManager: Shared Derived Context API

jleandroperez:
### Details
In this PR we're implementing a new ContextManager API, meant to help us prevent multiple concurrent sync OP(s).

I'm also borrowing @aerych 's kung fu from #14068 (cherry picked a commit of your sir!).
#### Expected usage: Swift / ObjC
```
let context = ContextManager.sharedInstance().derivedContext(forKind: ReaderPostService.self)
let service = ReaderPostService(managedObjectContext: context)
```

```
NSManagedObjectContext *context = [[ContextManager sharedInstance] derivedContextForKind:[ReaderPostService class]]
ReaderPostService *service = [[ReaderPostService alloc] initWithManagedObjectContext:context]

```

Refs #13867
Refs #13892

#### Next step:
Audit the top crasher services and implement this API when possible.

### To test:
- [ ] Verify the Unit Tests pass

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/ContextManager.m
+++ b/WordPress/Classes/Utility/ContextManager.m
@@ -70,7 +70,7 @@ static ContextManager *_override;
     return [self newChildContextWithConcurrencyType:NSMainQueueConcurrencyType];
 }
 
-- (NSManagedObjectContext *const)sharedDerivedContextForKind:(Class)kind
+- (NSManagedObjectContext *const)derivedContextForKind:(Class)kind
 {
     NSAssert([NSThread isMainThread], @"This API is not threadsafe. Should only run on the main thread");
 
--- a/WordPress/Classes/Utility/ContextManager.m
+++ b/WordPress/Classes/Utility/ContextManager.m
@@ -20,6 +20,7 @@ static ContextManager *_override;
 @property (nonatomic, strong) NSManagedObjectModel *managedObjectModel;
 @property (nonatomic, strong) NSManagedObjectContext *mainContext;
 @property (nonatomic, strong) NSManagedObjectContext *writerContext;
+@property (nonatomic, strong) NSMapTable *derivedContextMap;
 @property (nonatomic, assign) BOOL migrationFailed;
 
 @end
@@ -33,6 +34,7 @@ static ContextManager *_override;
 {
     self = [super init];
     if (self) {
+        _derivedContextMap = [NSMapTable strongToStrongObjectsMapTable];
         [self startListeningToMainContextNotifications];
     }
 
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Utility/ContextManager.m
```diff
@@ -33,6 +34,7 @@ - (instancetype)init
 {
     self = [super init];
     if (self) {
+        _derivedContextMap = [NSMapTable strongToStrongObjectsMapTable];
```
aerych, line 37: Maybe use the setter vs assigning to the ivar directly, just as a best practice?

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -68,6 +70,28 @@ - (NSManagedObjectContext *const)newMainContextChildContext
     return [self newChildContextWithConcurrencyType:NSMainQueueConcurrencyType];
 }
 
+- (NSManagedObjectContext *const)sharedDerivedContextForKind:(Class)kind
+{
+    NSAssert([NSThread isMainThread], @"This API is not threadsafe. Should only run on the main thread");
```
aerych, line 75: Excellent thought

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -68,6 +70,28 @@ - (NSManagedObjectContext *const)newMainContextChildContext
...
+    }
+
+    derivedContext = [self newDerivedContext];
+    derivedContext.automaticallyMergesChangesFromParent = YES;
+    [self.derivedContextMap setObject:derivedContext forKey:key];
+
+    return derivedContext;
+}
+
+- (NSString *)derivedContextKeyForKind:(Class)kind
```
aerych, line 90: For a future thought, we might decide it's a best practice for these to only be used by a certain type of object. Would it make sense to define a protocol, or some other strategy, to restrict what can get a specific derived context?

WordPress/Classes/Utility/ContextManager.m
jleandroperez: Good call, thank you Eric!

WordPress/Classes/Utility/ContextManager.m
jleandroperez: Definitely! I thought about limiting this one to, perhaps, `LocalCoreDataService`, but felt a bit odd to couple the ContextManager with such.

(If) we end up adopting this API, we could, perhaps implementing a protocol, right here. Something like... `DerivedContextWorker` (OR something like that!!).

Still tinkering, since... MediaCoordinator is so complex, and there are so many moving parts, that a straightforward replace causes several deadlocks all over the place.

Thanks for stepping by sir!!

aerych: > I thought about limiting this one to, perhaps, LocalCoreDataService, but felt a bit odd to couple the ContextManager with such

Totally agree. No need for the ContextManager to have awareness of other random types. I like the idea that the context manager could define a protocol that other classes could implement.  Or something like that.

jkmassel:
Howdy friends – we're freezing `15.0` today, so I'm bumping this PR forward to `15.1`. If it needs to land in the `15.0` series, please merge it to `release/15.0` and DM me – I'll be happy to create a new beta release!

jleandroperez:
No need to @jkmassel !! Actually closing this one for now, since the immediate crash was already solved.

Thanks for reaching out!!

PR #14090, opened 2020-05-08
Hide Create FAB Tooltip after usage or # times

bjtitus:
* Persists tooltip state & display count via User Defaults
* Sets a maximum of 5 of times that tooltip will appear

## Testing

- Open the Blog Details screen
- Click on the Create button and ensure that the tooltip is no longer shown

Reset using these command in LLDB:
```
e -l swift -- import WordPress; UserDefaults.standard.createButtonTooltipWasDisplayed = false
e -l swift -- import WordPress; UserDefaults.standard.createButtonTooltipDisplayCount = 0
```

- Check that the tooltip is shown again
- Try switching to other screens (non-Blog Detail) and ensure that the tooltip is hidden after 5 times navigating away and back.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
@@ -4,9 +4,10 @@ import WordPressFlux
 @objc class CreateButtonCoordinator: NSObject {
 
     private enum Constants {
-        static let padding: CGFloat = -16
-        static let heightWidth: CGFloat = 56
-        static let popoverOffset: CGFloat = -10
+        static let padding: CGFloat = -16 // Bottom and trailing padding to position the button along the bottom right corner
+        static let heightWidth: CGFloat = 56 // Height and width of the button
+        static let popoverOffset: CGFloat = -10 // The vertical offset of the iPad popover
+        static let maximumTooltipView = 5 // Caps the number of times the user can see the announcement tooltip
     }
 
     var button: FloatingActionButton = {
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
```diff
@@ -141,7 +150,10 @@ import WordPressFlux
     }
 
     @objc func showCreateButton() {
-        noticeContainerView = noticeAnimator.present(notice: notice, in: viewController!.view, sourceView: button)
+        shownTooltipCount += 1
```
Gio2018, line 153: incrementing before presenting means that we will see the button `maximumTooltipView - 1` times, unless we change the condition in `shownTooltipCount` to `>`.

WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
```diff
@@ -4,9 +4,10 @@ import WordPressFlux
 @objc class CreateButtonCoordinator: NSObject {
 
     private enum Constants {
-        static let padding: CGFloat = -16
-        static let heightWidth: CGFloat = 56
-        static let popoverOffset: CGFloat = -10
+        static let padding: CGFloat = -16 // Bottom and trailing padding to position the button along the bottom right corner
+        static let heightWidth: CGFloat = 56 // Height and width of the button
+        static let popoverOffset: CGFloat = -10 // The vertical offset of the iPad popover
+        static let maximumTooltipView = 5 // Caps the number of times the user can see the announcement tooltip
```
Gio2018, line 10: `maximumTooltipViews`?

bjtitus:
@Gio2018 I was thinking of the view count as a fallback and that the number was low enough it wouldn’t be a concern, but I could see both sides here.

@osullivanchris can you see if the logic outlined above makes sense — build 28284 includes this work (from [here](https://install.appcenter.ms/orgs/automattic/apps/WPiOS-One-Offs/)) if you want to give it a try.

osullivanchris:
Looks good to me!

bjtitus:
@Gio2018 I've added persistence for that display count and also the ability to tap on the tooltip to dismiss! Should be much easier to get rid of now.

PR #14094, opened 2020-05-11
Remove media deletion to prevent Core Data null.null errors

### LINKED ISSUE #12142: Media thumbnails may disappear because of choice of storage
In #12060, we found that whenever you attach a media to a post, a thumbnail is generated. This thumbnail is stored in `Library/Cache`. This location may not be ideal. Based on the [docs](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW28) say:

> Put data cache files in the Library/Caches/ directory. Cache data can be used for any data that needs to persist longer than temporary data, but not as long as a support file. Generally speaking, the application does not require cache data to operate properly, but it can use cache data to improve performance. Examples of cache data include (but are not limited to) database cache files and transient, downloadable content. Note that the system may delete the Caches/ directory to free up disk space, so your app must be able to re-create or download these files as needed.

It looks like the `Library/Cache` has a tendency to survive longer than the temporary folder. Still, we have to consider it as unreliable.

#### Use a Persistent Folder

In the case of "thumbnails" generated for use in the Editor and Post List, I believe they shouldn't be in the `Library/Cache` but in `Documents` instead. They should:

1. Be created when a media is attached.
2. Deleted once the media is uploaded to the server.

It looks like this can be done by changing the `MediaThumbnailExporter.mediaDirectoryType` to `.uploads` so the storage is persistent.

https://github.com/wordpress-mobile/WordPress-iOS/blob/4b71c77527ce91d39b898690275467dc997503d0/WordPress/Classes/Services/MediaThumbnailService.swift#L45-L47

I have not fully explored the repercussions of this. There might be other places that are using that method.

#### On-the-fly Generation

Another possible solution is if we can solve _Problem A_ described in #12060, we can try having an on-the-fly generation of the thumbnail. For example, I discussed above that the URL should be changed to this:

```html
<img src="file://Library/Caches/Media/thumbnail-p4-2436x2436.jpeg">
```

Since that is technically a placeholder now, we could insert additional data in there, like the `mediaID`:

```html
<img src="file://Library/Caches/Media/thumbnail-p4-2436x2436.jpeg?mediaID=1234">
```

We can then use that `mediaID` to generate the thumbnail if it is no longer available in `Library/Cache`. With this, we can keep the thumbnails in `Library/Cache` and do not have to worry about when to delete them.

leandroalonso:
This PR aims to fix the `*** Terminating app due to uncaught exception 'Unresolved Core Data save error (Derived Context)', reason: 'Merge policy failed - unable to complete merging on null.null' crashes.

Please, keep in mind that this PR doesn't solve all of our Core Data errors, only this specific one.

### Context

This code was originally submitted to fix #12142. The removal of this part will probably cause a regression on the fix of this bug, but the impact it's nothing compared to the number of crashes we've seen.

#### Crash: Scenario 1

1. Create a draft post and add one or more images
2. Close the post and save as draft
3. Re-open the post
4. Background the app
5. Kill/relaunch: the app would re-open to that same post
6. Close the post: no crash should occur

#### Crash: Scenario 2

1. Upload gallery on an existing draft
1. quit the app (remove from background) before upload complete
1. launch the app -> blog posts and see the failed post
1. Hit "Retry" and quit the app before upload complete
1. Launch the app -> Blog posts see failed post
1. tap the post and remove the failed gallery
1. Quit the app
1. Relaunch the app, wait 5 seconds after splash screen and -> blog posts
1. See the post trying to upload media automatically
1. wait and see "Uploading media" changing briefly to "Local changes" and then "Uploading post"
1. No crash should occur

### DIFF
--- a/WordPress/Classes/Utility/Media/MediaFileManager.swift
+++ b/WordPress/Classes/Utility/Media/MediaFileManager.swift
@@ -249,14 +249,6 @@ class MediaFileManager: NSObject {
                         onCompletion()
                     }
                 }
-
-                // Update media objects local urls that were deleted to be updated to nil
-                let editedObjectsRequest = NSBatchUpdateRequest(entityName: Media.classNameWithoutNamespaces())
-                editedObjectsRequest.predicate = NSCompoundPredicate(notPredicateWithSubpredicate: predicate)
-                editedObjectsRequest.propertiesToUpdate = [localURLProperty: NSNull(), localThumbnailURLProperty: NSNull()]
-                editedObjectsRequest.resultType = .updatedObjectsCountResultType
-                try context.execute(editedObjectsRequest)
-                try context.save()
             } catch {
                 DDLogError("Error while attempting to clean local media: \(error.localizedDescription)")
                 if let onError = onError {


Gio2018:
Tried both scenarios and the crash did not seem to occur!

one thing: scenario 2, point n. 11

> See the crash

Should this be: 'no crash should occur'?

leandroalonso:
@Gio2018 oops, yes!

SergioEstevao:
@leandroalonso I tested these changes and besides not having the crash anymore, we still avoid the thumbnail issue because the check for the file presence is still being done.

For the record, I tried to keep the core data code, by changing the merge policy in the context to `NSRollbackMergePolicy` but it did avoid the crash, because the exception happens on the saving of the post and associated media, and not in this code.

leandroalonso:
Thanks for looking into this, @SergioEstevao! 

I closed in favor of #14097, pointing to a correct branch.

PR #14100, opened 2020-05-12
Automated Development Environment Setup Rake Tasks

charliescheer:
This PR adds new tasks to the WordPress iOS rakefile to make it easy for new internal or external developers to get setup with WPiOS quckly.

The new tasks, rake init:oss and rake init:developer, automate much of the setup process that is in the current read me file into a single command that can be run on a new download of WordPress to do the following:
Check for xcode is installed with the correct version
Install all dependencies
install additional development tools
install lint hooks
setup API credentials and generate a filled out wpcom_app_credentials file
And help internal developers create a gpg key for access to production credentials

I have also created a new version of the read me that reflects how to use these new scripts when getting started with WPiOS for the first time.  

To test:
1. Clone a copy of WordPress-ios into a new folder.
2. in terminal go to that directory and run `rake init:oss` or `rake init:developer`  
3. The scripts will run mostly without input 
4. Make sure to have a valid API client App ID and Client secret, you will be prompted to add this info
5.  A new gpg key will also be created in this process (there is an option to skip that)

Once complete make sure that:
The dependencies have been installed and are up to date
the swift link hooks are installed and setup
WordPress ios can compile and log in

PR submission checklist:


Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+namespace :init do
+task :oss => %w[
+  install:xcode:check
+  dependencies
+  install:tools:check_oss
+  install:lint:check
+  credentials:setup
+]
+
+desc "Install and configure WordPress iOS and it's dependencies - a8c Developers"
```
mokagio, line 276:
```suggestion
desc "Install and configure WordPress iOS and it's dependencies - Automattic Developers"
```

Before joining, I wasn't familiar with what A8C meant.

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+  namespace :xcode do
+    task :check => %w[xcode_app:check xcode_select:check]
+
+    #xcode_app namespace checks for the existance of xcode on developer's machine,
+    #checks to make sure that developer is using the correct version per the CI specs
+    #and confirms developer has xcode-select command line tools, if not installs them
+    namespace :xcode_app do
+      #check the existance of xcode, and compare version to CI specs
+      task :check do
+        puts "Checking for system for XCode"
```
mokagio, line 297: **Nitpick**, it's called [Xcode](https://developer.apple.com/xcode/) 😆
```suggestion
        puts "Checking for system for Xcode"
```

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+          puts "Please install XCode before setting up WordPressiOS"
+          puts "https://apps.apple.com/app/xcode/id497799835?mt=12"
+          abort("")
+        else
+          puts "Xcode installed"
+        end
+
+        puts "Checking CI recommendded installed XCode version"
+
+        if !xcode_version_is_correct?
```
mokagio, line 313: **Ruby nitpick**, I find `!value` as a conditional tricky to read, it's easy to miss the `!` while scanning the code and getting the wrong idea of what the code does.

Possible alternatives are

```ruby
if xcode_version_is_correct == false
  ... 
end
```

or 

```ruby
unless xcode_version_is_correct
  ... 
end
```

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+          #if xcode is the wrong version, prompt user to install the correct version and terminate rake
+          puts "Not recommended version of XCode installed"
+          puts "It is recommended to use Xcode version #{get_ci_xcode_version}"
+          STDOUT.puts "Please press enter to continue"
```
mokagio, line 317: Is this `STDOUT` required? How does this differ from the other `puts`?

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+          complete = STDIN.gets.strip
```
mokagio, line 318: **Nitpick**, it doesn't look like we need to read the value of `complete`, i.e., the key the user pressed. If so, we could just ignore it.

```suggestion
          STDIN.gets.strip
```

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+        system "xcodebuild -version", [:out, :err] => File::NULL
+      end
+
+      #compare xcode version to expected CI spec version
+      def xcode_version_is_correct?
+        if get_xcode_version == get_ci_xcode_version
+          puts "Correct version of XCode installed"
+          return true
+        end
+      end
```
mokagio, line 334: Is it ok for this to return `nil` if the two versions don't match? Or would it be clearer if it returned false?

Screen Shot 2020-05-27 at 3 22 40 pm

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+        version = %x[xcodebuild -version]
+
+        version.split(" ")[1]
+      end
+
+      def get_ci_xcode_version
+        ci_config = File.read(".circleci/config.yml")
+        specs = YAML.load(ci_config)
+
+        ci_version = specs["jobs"]["Build Tests"]["executor"]["xcode-version"]
```
mokagio, line 348: Nice idea going to check the Circle CI config for this value 👍 

I wonder if there's something that could be done to makes sure this code doesn't get out of date, say for example if the format in which Circle CI wants its YAML configs changes 🤔 

Just a question though, not something that should be addressed in this PR.

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+  namespace :tools do
+    task :check_oss => %w[homebrew:check addons:check_oss]
+    task :check_developer => %w[homebrew:check addons:check_developer]
+
+    #Check for Homebrew and install if missing
+    namespace :homebrew do
+      task :check do
+        puts "Checking system for Homebrew"
+        unless command?("brew")
+          Rake::Task["tools:homebrew:install"].invoke
```
mokagio, line 384: How would you feel about asking the user permission to run this?

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+      #Check for tool, install if not installed
+      task :check_oss do
+        tool_check(oss_tools)
+      end
+
+      task :check_developer do
+        tool_check(developer_tools)
+      end
+
+      #check machine for if the developer tool is present, if not install
```
mokagio, line 420:
```suggestion
      #check if the developer tool is present in the machine, if not install
```

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+
+  #End namespace Tools
+  end
+
+  namespace :lint do
+    task :check do
+      if !git_initialized?
+        puts "Initializing git repository"
+        sh "git init", verbose: false
+      end
```
mokagio, line 449: Is this check necessary? Is there a scenario in which a user might run this code on the WordPress repo _without_ the `.git` folder set up? 🤔

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+    puts '3. Set "Redirect URLs"= https://localhost and "Type" = Native and click "Create" then "Update".'
+    prompt_for_continue
+
+    puts "====================================================================================="
+    puts "4. Copy the Client ID and Client Secret from the OAuth Information."
+    prompt_for_continue
+  end
+
+  def prompt_for_continue
+    STDOUT.puts "Please press enter to continue"
```
mokagio, line 493: **Nitpick** This text will be read by users that are expected to do something _and only then_ press enter to continue.

How can we make this clearer?

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+
+  #user given app id and secret and create a new wpcom_app_credentials file
+  task :set_app_secrets do
+    create_secrets_file
+    set_app_secrets(get_client_id, get_client_secret)
+    remove_temp_file
+  end
+
+  def get_client_id
+    STDOUT.puts "Please enter your Cliet id"
```
mokagio, line 505:
```suggestion
    STDOUT.puts "Please enter your Client ID"
```

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+    sh "cp WordPress/Credentials/wpcom_app_credentials-example .configure-files/temp_wpcom_app_credentials", verbose: false
+  end
+
+  #create a new wpcom_app_credentials file combining the app secret and app id
+  def set_app_secrets(id, secret)
+    puts "Creating credentials file"
+    new_file = File.new(".configure-files/wpcom_app_credentials", "w")
+    File.open(".configure-files/temp_wpcom_app_credentials") do |file|
+      file.each_line do |line|
+        string = line.to_s()
```
mokagio, line 525: Is this conversion to String necessary?

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+    puts "====================================================================================="
+    puts "Press 'Y' for Yes.  Press 'N' for custom configuration"
+
+    display_prompt_response
+  end
+
+  #prompt for a Y or N response, continue asking if other character
+  #return true for Y and false for N
+  def display_prompt_response
+    response = STDIN.gets.strip.upcase
```
mokagio, line 632: Nice touch using the `.upcase` on the input 👍

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+    until response == "Y" || response == "N"
+        puts "Invalid entry, please enter Y or N"
+        response = STDIN.gets.strip.upcase
+    end
+
+    if response == "Y"
+      return true
+    elsif response == "N"
+      return false
+    end
```
mokagio, line 642: **Ruby nitpick**, given we already have the logic above to ensure the only valid values are `Y` or `N`, this could be simplified into a check for `Y`

```suggestion
    return response == "Y"
```

Rakefile
```diff
@@ -262,6 +263,399 @@ task :xcode => [:dependencies] do
...
+    puts ""
+    puts ""
+    puts "====================================================================================="
+    puts "You will need to enter the following info to create your key"
+    puts "Please enter your real name, email address, and a password for your key when prompted"
+    puts "====================================================================================="
+  end
+
+#end namespace GPG_key_setup
+end
```
mokagio, line 656: While I appreciate the information value these "end namespace" comments give when scanning the code, I do wonder whether the cost of maintaining them or them getting out of date (e.g., the `namespace` here is called `:gpg_key` not `:GPG_key_setup`) is too high.

README.md
```diff
@@ -7,28 +7,35 @@
...
-1. From a command line, `git clone git@github.com:wordpress-mobile/WordPress-iOS.git` in the folder of your preference.
-1. `cd WordPress-iOS` to enter the working directory.
-1. `rake dependencies` to install all dependencies required to run the project (this may take some time to complete).
-1. `rake xcode` to open the project in Xcode. 
-1. Compile and run the app on a device or an simulator.
+1. From a command line, run `git clone git@github.com:wordpress-mobile/WordPress-iOS.git` in the folder of your preference.
+1. Now, run `cd WordPress-iOS` to enter the working directory.
 
-In order to login to WordPress.com using the app:
+### Create WordPress.com API Credentials
```
mokagio, line 16: Using the same heading level here as above, `###`, made me think this was a different section, while it's a subsection of the setup instructions
```suggestion
#### Create WordPress.com API Credentials
```

README.md
```diff
@@ -7,28 +7,35 @@
...
+1. Once completed, run `rake xcode` to open the project in Xcode.
+
+If all went well you can now compile to your iOS device or simulator, and log into the WordPress app.
+
+Note: You can only log in with the WordPress.com account that you used to create the WordPress application.
 
-You can only log in with the WordPress.com account that you used to create the WordPress application.
+## Configuration Details
+
+The steps above will help you configure the WordPress app to run and compile.  But you may sometimes need to update or re run specific parts of the initial setup (like updating the dependencies)  To see how to do that, please check out the steps below
```
mokagio, line 34:
```suggestion
The steps above will help you configure the WordPress app to run and compile. But you may sometimes need to update or re-run specific parts of the initial setup (like updating the dependencies.) To see how to do that, please check out the steps below.
```

Rakefile
charliescheer: I hadn't put this in originally, but while testing, I setup a completely fresh install of WPiOS  (new OSX, new Xcode, new clone of the WPiOS repo) and the setup failed because I had not yet initialized the git repo on the folder.  So I thought I would build the check in just incase.

I agree that it is really unlikely that a user would run this setup without the .git folder having been created, but it the folder isn't there then the rake setup aborts.  

I do think that the check should be there, because if isn't and installing the swift lint hooks fails, there isn't a lot of indication as to why. But I could add some notes to clarify why the check is there.  What do you think?  Or is there a reason that should be left out that I missed?

Rakefile
charliescheer: Fixed

Rakefile
charliescheer: Eek found several of those... fixed

Rakefile
charliescheer: Ah that isn't required.

Rakefile
charliescheer: Thanks for pointing that out! originally I was using this return and didn't remove it later.

Rakefile
charliescheer: Thanks!

Rakefile
charliescheer: That is much simpler, thanks for the advice!  Fixed

Rakefile
charliescheer: Great point, I hadn't considered that at the time.

Rakefile
charliescheer: That's a good question.  I had not added any return to false because what it triggers happen if what comes back is not true.  So my thought on this was to only create the circumstance t return true, and left out other cases.  
But since it is a boolean check it would be clearer to return false if  the check fails.

Rakefile
charliescheer: Great idea!  Added

Rakefile
charliescheer: Added some more details to make what the script is waiting for to the developer running it

Rakefile
charliescheer: huh.... could have sworn that I was getting errors before when the to_s wasn't there... but removed it and it is working now...  Simplified that section.  Thanks!!

charliescheer:
Thanks for the great feedback @mokagio I have fixed the things that you suggested and added a check with the developer before installing homebrew.

Please let me know if you see anything else that needs to be looked at/fixed.

Rakefile
```diff
@@ -262,6 +263,397 @@ task :xcode => [:dependencies] do
...
+
+  #user given app id and secret and create a new wpcom_app_credentials file
+  task :set_app_secrets do
+    create_secrets_file
+    set_app_secrets(get_client_id, get_client_secret)
+    remove_temp_file
+  end
+
+  def get_client_id
+    STDOUT.puts "Please enter your Client id"
```
mokagio, line 513: Just a nitpick. You don't have to address this here, it an easily be a followup PR because it doesn't have to do with the actual workings of the code.

There's a bit of inconsistencies in how the "Client ID" and "Client Secret" are typed. Sometimes they are "Client ID", other times "client ID", or "Client id", or "client id". It would be great to have it consistent. To be honest, though, [our docs](https://developer.wordpress.com/docs/oauth2/) are not consistent with it either... 😳 

```suggestion
    STDOUT.puts "Please enter your Client ID"
```

Rakefile
```diff
@@ -636,6 +624,18 @@ namespace :gpg_key do
...
+#return true for Y and false for N
+def display_prompt_response
+  response = STDIN.gets.strip.upcase
+  until response == "Y" || response == "N"
+      puts "Invalid entry, please enter Y or N"
+      response = STDIN.gets.strip.upcase
+  end
+
+  return response == "Y"
+end
```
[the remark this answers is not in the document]
charliescheer: Thanks!

loremattei:
Thanks for this @charliescheer and @mokagio!
I've started the CI tests on this PR to make sure everything is fine as we use `rake` on CI.

PR #14102, opened 2020-05-12
Refactor Core Data initialization to use NSPersistentContainer

leandroalonso:
#### Updating the app

1. Run an old version of the app
2. Install this version
3. Check that the database behaves ok and all your data it's there

#### Fresh install

1. Run a fresh install
2. Check that the database is successfully created


Gio2018:
> I'm seeing some warnings after using NSPersistentContainer:

Could this be because we might be instantiating two managed object models? One in the persistent container and the other one is the property `managedObjectModel`, called from `migrateDataModelIfNecessary`?

leandroalonso:
@Gio2018 you're right! Refactored to keep a single `managedObjectModel`

astralbodies:
I love that this works without a ton of messing around! Here's what I would consider doing to take advantage of what NSPersistentContainer is:

1. Make the container a property inside of the context manager. There's a benefit to accessing everything important inside of container instead of the lazy vars we're using. The persistent store description could be a lazy property instead.
2. Update the iterative migrator to use a more modern approach (see attached Gist with the example from the Ray Wenderlich book).
3. Stop doing all of the manual merging when a notification comes in and try to rely heavily on `container.viewContext.automaticallyMergesChangesFromParent` as much as possible.
4. TestContextManager will need some refactoring to support all of the changes.

Example code with Core Data stacks & iterative migrator: https://gist.github.com/astralbodies/4e30ee85677b9e9321d8c6e8cfc2d9af

leandroalonso:
@astralbodies about this:

> Stop doing all of the manual merging when a notification comes in and try to rely heavily on container.viewContext.automaticallyMergesChangesFromParent as much as possible.

Do you mean using `mergeChanges` directly? Because I suspect only [`ReaderStreamViewController`](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift#L1191) does that.

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -0,0 +1,273 @@
+import Foundation
+import CoreData
+
+/// CoreDataIterativeMigrator: Migrates through a series of models to allow for users to skip app versions without risk.
+/// This was derived from ALIterativeMigrator originally used in the WordPress app.
```
leandroalonso, line 5: Maybe we want to remove this? lol

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -346,12 +342,14 @@ - (void)migrateDataModelsIfNecessary
         DDLogWarn(@"Migration required for persistent store.");
         NSError *error = nil;
         NSArray *sortedModelNames = [self sortedModelNames];
-        BOOL migrateResult = [ALIterativeMigrator iterativeMigrateURL:[self storeURL]
```
leandroalonso, line 349: I guess we can remove `ALIterativeMigrator` from the project.

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
Gio2018: I think we can 😀

WordPress/Classes/Utility/ContextManager.m
Gio2018: 👍

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -350,6 +350,9 @@ - (void)migrateDataModelsIfNecessary
 
         if (error != nil) {
             DDLogError(@"Unable to migrate store: %@", error);
+
+            SentryStartupEvent *startupEvent = [SentryStartupEvent new];
```
leandroalonso, line 354: We need to use the same `startupEvent` from `persistentContainer`. Otherwise we will end up with different events and this one will not be triggered (we need to call `sendWithTitle:` to fire this).

Take a look here: https://github.com/wordpress-mobile/WordPress-iOS/blob/fef99452a26bf739fa997e711d1f0817f57c9705/WordPress/Classes/Utility/ContextManager.m#L252

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -35,12 +35,10 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
     private var noticePresenter: NoticePresenter?
     private var bgTask: UIBackgroundTaskIdentifier? = nil
 
+    private let mainContext = ContextManager.shared.mainContext
```
aerych, line 38: I'm thinking this might conflict the way `ContextManager.overrideSharedInstance` is intended to work.  Assigning the context means the overridden context wouldn't be used later.  Maybe we could tweak this to be a computed property instead of a stored property?

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -0,0 +1,272 @@
...
+        do {
+            try migrator.migrateStore(from: url,
+                                      sourceType: storeType,
+                                      options: nil,
+                                      with: mappingModel,
+                                      toDestinationURL: tempDestinationURL,
+                                      destinationType: storeType,
+                                      destinationOptions: nil)
+        } catch {
+            return false
```
aerych, line 202: Maybe log the error here, and in the following do/try/catch block, before returning?

aerych: Never mind. I just saw they are logged where thrown :)

WordPress/Classes/System/WordPressAppDelegate.swift
astralbodies: I think this also has the side-effect of launching Core Data early in the lifecycle of the app before willfinish or didfinish is triggered.

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -203,95 +204,89 @@ - (void)mergeChanges:(NSManagedObjectContext *)context fromContextDidSaveNotific
...
+            DDLogError(@"Error opening the database. %@\nDeleting the file and trying again", error);
+
             SentryStartupEventAddError(startupEvent, error);
             error = nil;
-        }
 
-        // delete the sqlite file and try again
-        [[NSFileManager defaultManager] removeItemAtPath:storeURL.path error:&error];
+            // make a backup of the old database
+            [[NSFileManager defaultManager] copyItemAtPath:storeURL.path
```
astralbodies, line 233: SQLite actually creates up to three files if journaling is enabled (which it is by default in Core Data). If those journal files stick around when we create a new db, I could see that causing an issue bringing up the store after a migration would fail. In the iterative migrator we sort of account for this in the making of a backup and copying over the migrated file by doing the contents of the directory, not a single file. 

I don't know if it would fix anything, but it wouldn't hurt.

WordPress/Classes/System/WordPressAppDelegate.swift
leandroalonso: Our idea was to initialize the Core Data stack as soon as possible to avoid any concurrency issue.

> Assigning the context means the overridden context wouldn't be used later. Maybe we could tweak this to be a computed property instead of a stored property?

This `mainContext` is used only by the entities in `WordPressAppDelegate`. In fact, we've injected into them to justify the use here — otherwise, it would seem that this is here but not used at all.

We can keep the property here and do not use it to inject tho. And maybe add a comment explaining why is here?

> I think this also has the side-effect of launching Core Data early in the lifecycle of the app before willfinish or didfinish is triggered.

Maybe we can move that to be the first thing in willfinish?

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -203,95 +204,89 @@ - (void)mergeChanges:(NSManagedObjectContext *)context fromContextDidSaveNotific
...
-}
-
-- (NSPersistentStoreCoordinator *)persistentStoreCoordinator
+- (NSPersistentContainer *)persistentContainer
 {
-    if (_persistentStoreCoordinator) {
-        return _persistentStoreCoordinator;
+    if (_persistentContainer) {
+        return _persistentContainer;
     }
```
jleandroperez, line 214: I'm wondering if it wouldn't be best to rely on GCD to initialize all the things *once* (rather than having lazy non threadsafe properties).

WordPress/Classes/System/WordPressAppDelegate.swift
leandroalonso: Since this has been _controversy_ I reverted here: cb1a526890d548efd73412412967d6ff12154705

WordPress/Classes/Utility/ContextManager.m
leandroalonso: @astralbodies Do you mean copying the `-wal` and `-shm` files too?

leandroalonso: I changed to use `CoreDataIterativeMigrator` backup logic!

WordPress/Classes/Utility/ContextManager.m
leandroalonso: Since we're going to change this whole code to Swift very soon, I guess we can tackle that on the next task.

What do you think?

leandroalonso:
@astralbodies @aerych @jleandroperez sirs, ready for another 👀!

WordPress/Classes/Utility/ContextManager.m
astralbodies: > @astralbodies Do you mean copying the `-wal` and `-shm` files too?

Yup! We could also turn off journaling as an option.

leandroalonso:
Tks @astralbodies!

Issue #13683, opened 2020-03-18
Conflict Resolution Dialog


leandroalonso:
We've been receiving a lot of complaints of users confused about posts not being updated and displaying the "Local Changes" label (ref #9449).

To reproduce this issue:

1. Create a post using the app and save as a draft
2. Open the post, type some content and then close the app (without exiting the editor)
3. Go to the web, change the post content and publish it
4. Go back to the app, the post will have a "Local changes" label and will never update unless the user discards the changes

Our goal is to match the implemented behavior in Android, which displays a clear message to the user and prompts for an action:

| Post list        | Prompt           |
| ------------- |:-------------:|
| <img src="https://user-images.githubusercontent.com/7040243/76760446-711bd600-676c-11ea-8ea3-34e50ec721c4.png" width="300px" /> | <img src="https://user-images.githubusercontent.com/7040243/76760543-99a3d000-676c-11ea-804d-d514f9a8c7d4.png" width="300px" /> |

#### Showing label

* When the user has a post with local changes
* That has been published with another device (ie.: web)
* The app shows "Version conflict" label

#### Showing prompt to the user

* When the user has a post with local changes
* That has been published with another device (ie.: web)
* And the user tap to edit the post
* We prompt the user to choose between the local version or the remote version

#### Choosing "Discard web"

* We update the post using the local post

#### Choosing "Discard local"

* We discard the local post

## References

* Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/8989
* Revision Dialog on iOS (which is quite similar, but not the same thing): https://github.com/wordpress-mobile/WordPress-iOS/issues/12141

embryoconcepts:
Subtasks (will add PR links when available):

- UI 
	- Add label for `Version Conflict` on `PostCardCell`
	- Add an alert to `Resolve sync conflict` on `PostListViewController`
	- Add toast to `Undo` conflict resolution on `PostListViewController `
- Logic
	- Create a flag for a post having a conflict, based on last modified date
	- Show label when a post has a version conflict in `PostCardCell`
	- Show alert when tapping on a post with a version conflict
	- Present Undo toast, handle behavior for discarding local or web update

diegoreymendez:
I've chatted with @leandroalonso about closing this.  Please DM Leandro or myself if you have questions.

PR #14104, opened 2020-05-12
Create flag and update label for a conflicted post

embryoconcepts:
- Create a BOOL flag for a post that has a conflict in the `dateModified` property between the local and web versions
- Update the status on the PostCard and PostCompact cells to display ‘Version Conflict` in `error` coloring, when a post has a version conflict

NOTES: 
- Android uses `isConflicted`, but I felt `hasVersionConflict` was more precise, and more future-proof (ie, there may be other types of conflicts in the future, and this leaves room for those names
- I opted to only update the local status, ala “Local Changes”, and not the post status, since this is something that is restricted to local - please let me know if I should incorporate it as a proper Post Status, and I can do that 

Fixes (in part) #13683 

To test:
1. Create a draft while I am online in the app. That’ll create the draft in remote.
2. Go offline
3. Make changes to the post and go back. This will save the changes as local draft
4. Go to web or any other client and make changes to the post.
5. Publish the post.
6. on the app, you should be on the posts list. turn airplane mode off
7. pull to refresh.
8. you should see now the post has a “Conflicted” state.

PR submission checklist:


WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,13 @@ class PostTests: XCTestCase {
         XCTAssertFalse(revision.hasLocalChanges())
     }
 
+    func testThatVersionConflictWorks() {
+        #warning("create tests")
```
leandroalonso, line 380: Here we want two tests: one that given the conditions `hasVersionConflict` is `true` and another that is false.

What we want to do is to have a post that has an `original` and manipulate their dates, that way we'll ensure the logics of `hasVersionConflict`. Using .revision() [like here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressTest/ViewRelated/Post/Views/PostCardStatusViewModelTests.swift#L96) might work. :)

WordPress/Classes/ViewRelated/Post/PostCardStatusViewModel.swift
```diff
@@ -65,6 +65,8 @@ class PostCardStatusViewModel: NSObject {
             return NSLocalizedString("Uploading post...", comment: "Message displayed on a post's card when the post has failed to upload")
         } else if !post.hasLocalChanges() && post.hasAutosaveRevision {
             return StatusMessages.hasUnsavedChanges
+        } else if post.hasVersionConflict() {
+            return StatusMessages.hasVersionConflict
```
leandroalonso, line 69: We have some tests for the messages, you can see them here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressTest/ViewRelated/Post/Views/PostCardStatusViewModelTests.swift#L106

We just need to manipulate the post so it has an `.original` with a different date.

WordPress/Classes/ViewRelated/Post/PostCardCell.xib
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="15400" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="16096" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES">
```
leandroalonso, line 2: Sometimes just by opening a `.xib` Xcode changes its code (well, tks Xcode).

While I don't think this is a huge problem, check those files before committing so we can keep the PRs as small as possible!

WordPress/Classes/Models/AbstractPost.m
```diff
@@ -581,6 +581,12 @@ - (BOOL)hasRemoteChanges
             || self.remoteStatus == AbstractPostRemoteStatusFailed);
 }
 
+- (BOOL)hasVersionConflict
+{
+    AbstractPost *original = (AbstractPost *)self.original;
```
leandroalonso, line 586: I'm just wondering here: I think we can have a scenario when the user has a draft in which the date is more "recent" than the `original`.

That wouldn't configure a version conflict. Wdyt?

leandroalonso (changes requested):
Good work here @embryoconcepts.

Left a few suggestions for unit tests and a question about the conflict logic.

WordPress/Classes/ViewRelated/Post/PostCardCell.xib
embryoconcepts: ack, I already discarded those changes once, didn't see them slip in again!

WordPress/Classes/Models/AbstractPost.m
embryoconcepts: I don't think you can, because you can't currently save the post if there's a conflict. Let me try to make one happen.

WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,20 @@ class PostTests: XCTestCase {
         XCTAssertFalse(revision.hasLocalChanges())
     }
 
+    func testThatVersionConflictWorks() {
```
leandroalonso, line 379: I'm not a big fan of multiple tests inside a single function.

But I can see that there's a lot of others here like that. 😬

So it's up to you to make those changes or not.

WordPress/WordPressTest/ViewRelated/Post/Views/PostCardStatusViewModelTests.swift
```diff
@@ -111,6 +111,26 @@ class PostCardStatusViewModelTests: XCTestCase {
         expect(viewModel.status).to(equal(i18n("We'll publish the post when your device is back online.")))
         expect(viewModel.statusColor).to(equal(.warning))
     }
+
+    func testVersionConflictStatusMessage() {
```
leandroalonso, line 115: I think we can break those tests into 3 different tests. Also, you can add breaklines to make it easy to spot the [Arrange-Act-Assert](https://wiki.c2.com/?ArrangeActAssert) parts.

One thing that I did on the tests above was to add a comment explaining what the test does. I find it easier to read the comment instead of the function name. Anyway, the comment itself isn't a pattern adopted over the project, so up to you. :)

WordPress/Classes/Models/AbstractPost.m
leandroalonso: I tested here with a published post and I'm seeing "Version Conflict" while there's no version conflict.

Steps to reproduce:

1. Create and publish a post
2. Open this post, add some content and close the app
3. Open the app and check the post list, this post will be displayed with the "Version Conflict" label

That's because we created a new draft, with a recent date. Recent date != post.original date.

leandroalonso (changes requested):
Thank you so much for the changes @embryoconcepts!

Almost there!

WordPress/Classes/Models/AbstractPost.m
```diff
@@ -581,6 +581,17 @@ - (BOOL)hasRemoteChanges
...
+    if (original.dateModified) {
+        return ![self.dateModified isEqualToDate:original.dateModified] && self.hasLocalChanges;
+    } else {
+        // This is to accommodate tests - many of the mock Posts do not have an original with a dateModified property
```
leandroalonso, line 590: This is not only for tests. :)

When you create a new post in the app, it will not have the `original`.

WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,54 @@ class PostTests: XCTestCase {
...
 
+    func testThatVersionConflictWorksShouldNotConflict() {
+        let original = newTestPost()
+        XCTAssertFalse(original.hasVersionConflict(), "Post should return false for `hasVersionConflict`")
+    }
+
+    // Local changes alone should not trigger `hasVersionConflict`
+    func testThatVersionConflictWorksShouldNotConflictWithOnlyTagChange() {
+        let original = newTestPost()
+        let local = original
```
leandroalonso, line 387: I'm not sure I follow this.

`local` and `original` are the same entity (they don't have an `original`, for example). When you create a local change based on a post you'll have two entities: the new `Post` entity and the `original` one.

WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,54 @@ class PostTests: XCTestCase {
...
+
+        local.tags = "test"
+
+        XCTAssertFalse(local.hasVersionConflict(), "Post should return false for `hasVersionConflict`")
+    }
+
+    // dateModified changes alone should not trigger `hasVersionConflict`
+    func testThatVersionConflictWorksShouldNotConflictWithOnlyDateChanges() {
+        let original = newTestPost()
+        let local = original
```
leandroalonso, line 397: Same as above.

WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,54 @@ class PostTests: XCTestCase {
...
+
+        local.setPrimitiveValue((Date() - 5), forKey: "dateModified")
```
leandroalonso, line 399:
```suggestion
        local.dateModified = (Date() - 5)
```

WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,54 @@ class PostTests: XCTestCase {
...
+
+        XCTAssertTrue(local.hasVersionConflict(), "Local post with a previous date should return true for `hasVersionConflict`")
+    }
+
+    func testThatVersionConflictWorksWithLaterPost() {
+        let original = newTestPost()
+        original.setPrimitiveValue(Date(), forKey: "dateModified")
+        let local = original.createRevision() as! Post
+
+        local.setPrimitiveValue((Date() + 5), forKey: "dateModified")
```
leandroalonso, line 421: You can assign `dateModified` without the need for using `setPrimitiveValue` :)

WordPress/WordPressTest/PostTests.swift
```diff
@@ -376,6 +376,54 @@ class PostTests: XCTestCase {
...
+
+    func testThatVersionConflictWorksWithPreviousPost() {
+        let original = newTestPost()
+        original.setPrimitiveValue(Date(), forKey: "dateModified")
+        let local = original.createRevision() as! Post
+
+        local.setPrimitiveValue((Date() - 5), forKey: "dateModified")
+        local.tags = "test"
+
+        XCTAssertTrue(local.hasVersionConflict(), "Local post with a previous date should return true for `hasVersionConflict`")
```
leandroalonso, line 413: Jen, just a minor thing. Your `act` here is the call to `local.hasVersionConflict()` itself. :)

So this could be rearranged like:

```suggestion
        // Arrange
        let original = newTestPost()
        original.setPrimitiveValue(Date(), forKey: "dateModified")
        let local = original.createRevision() as! Post
        local.setPrimitiveValue((Date() - 5), forKey: "dateModified")
        local.tags = "test"
        
        // Act
        let hasVersionConflict = local.hasVersionConflict()

        XCTAssertTrue(hasVersionConflict, "Local post with a previous date should return true for `hasVersionConflict`")
```

But don't worry about that. Just keep in mind for the next tests, it's not a big deal, really.

leandroalonso (changes requested):
Just minor changes @embryoconcepts and we would be good to go! o/

leandroalonso (approved):

leandroalonso (changes requested):
@embryoconcepts All good to go. Just one detail:

Can you please remove `Localizable.strings`? We should not change this file (you can see a warning above about it).

leandroalonso (approved):

PR #14108, opened 2020-05-13
[FIX] Gutenberg - use allowMultipleSelection flag from JS in the GutenbergStockPhotosPicker

dratwas:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/953#issuecomment-627709554
In this PR I removed the hardcoded value of `allowMultipleSelection` in StockPhotosPicker and used flag passed from JS side instead.

To test:
- Run app
- Open Gutenberg editor
- Add the cover block and choose "Free photo library"
- You should be able to select only one media

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/Utils/GutenbergStockPhotos.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/Utils/GutenbergStockPhotos.swift
@@ -15,8 +15,7 @@ class GutenbergStockPhotos {
     func presentPicker(origin: UIViewController, post: AbstractPost, multipleSelection: Bool, callback: @escaping MediaPickerDidPickMediaCallback) {
         let picker = StockPhotosPicker()
         stockPhotos = picker
-        // Forcing multiple selection while multipleSelection == false in JS side.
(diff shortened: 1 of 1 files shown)


pinarol:
@loremattei Fyi, we'd like to include this one to the release as well.

pinarol:
> On a previously saved post with galleries, if I try to append new photos, the new photos will replace the existing ones. After this, adding new ones will be appended. Not sure if this is expected but it didn't feel right to me

@etoledom Very interesting case, is it possible to repro that by redo-ing your changes on that post? If yes, please open an issue by granting us some editing access to that site so that we can debug.

etoledom:
> @etoledom Very interesting case, is it possible to repro that by redo-ing your changes on that post? If yes, please open an issue by granting us some editing access to that site so that we can debug.

So, this looks like a false alarm.

Looking at the HTML code for the gallery, there's this:
` {"ids":[null,null,null],"columns":2} `

Probably it malformed somehow. This post in particular seems to have been a copy/paste from somewhere else.

On web it also will create a new gallery.

pinarol:
Cool thanks 😌 @etoledom

PR #14122, opened 2020-05-14
Added a conflict resolution alert to the PostList

embryoconcepts:
- Added an alert to the PostList view when posts with a version conflict are selected
- The user is given two options: Discard Web, or Discard Local
- Upon selection, the appropriate version is discarded, and the editor is loaded with the version that has been retained

Fixes (in part) #13683

To test:

Create

1. Create a post using the app and save as a draft
2. Open the post, type some content and then close the app (without exiting the editor)
3. Go to the web, change the post content and publish it
4. Go back to the app, the post will have a "Local changes" label and will never update unless the user discards the changes
5. You should see now the post has a “Conflicted” state.
6. Tapping on the post with a version conflict with give two options: Discard Local, or Discard Web
7. Discard Local should remove local changes and load the editor with the latest version from the web
8. Discard Web should remove the latest web revision, and load the editor with the local changes

PR submission checklist:


WordPress/Classes/ViewRelated/Post/PostListEditorPresenter.swift
```diff
@@ -8,8 +8,22 @@ struct PostListEditorPresenter {
 
     static func handle(post: Post, in postListViewController: PostListViewController) {
 
-        // Autosaves are ignored for posts with local changes.
-        if !post.hasLocalChanges(), post.hasAutosaveRevision, let saveDate = post.dateModified, let autosaveDate = post.autosaveModifiedDate {
+        if post.hasVersionConflict() {
+            let conflictResolutionAlert = presentConflictResolutionAlert(for: post) { keepLocal in
+                switch keepLocal {
```
leandroalonso, line 13: I'm very curious about that... using a `switch` for a Boolean has any advantage over an `if`?

embryoconcepts: it's just cleaner, in my eyes. I don't have strong feelings about it.

embryoconcepts:
@leandroalonso Should be good for a final review! I abstracted the resolution logic, fixed the issues you were seeing, and added tests.

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
```diff
@@ -0,0 +1,70 @@
+import Foundation
+
+struct PostListConflictResolver {
```
leandroalonso, line 3: I guess the conflict resolver itself doesn't need to know about Post List. What do you think?

We can give a view controller to it, but it doesn't necessarily need to be a `PostListViewController`. It could be any `UIViewController`.

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
```diff
@@ -0,0 +1,70 @@
...
+    /// Presents an alert with an option to discard the local or web version of a post, and opens the editor with the non-discarded version
+    static func handle(post: Post, in postListViewController: PostListViewController) {
+        let conflictResolutionAlert = presentConflictResolutionAlert(for: post) { discardWeb in
+            guard ReachabilityUtils.isInternetReachable() else {
```
leandroalonso, line 7: I don't think we need to handle offline support yet in this task.

Anyway, one of the principles that we follow is to always try to execute the request instead of relying only on the `ReachabilityUtils`. You can see that in action in `PostCordinator.dispatchNotice()`.

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
```diff
@@ -0,0 +1,70 @@
...
+            }
+            discardWeb ? keepLocalVersion(post: post) : keepWebVersion(post: post)
+            PostListHelper.openEditor(with: post, loadAutosaveRevision: false, in: postListViewController)
+        }
+        postListViewController.present(conflictResolutionAlert, animated: true)
+    }
+
+    /// Discard web revision, load post with local changes
+    private static func keepLocalVersion(post: Post) {
+        PostCoordinator.shared.autoSave(post)
```
leandroalonso, line 20: While this works, I guess that the correct method that should be used is `PostCoordinator.save()`.

Also, this is an async method and it might fail. If I'm not wrong, the Android version uploads the local version and only then show the editor. We can double-check with them. :)

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// Discard local changes, load latest web revision
+    private static func keepWebVersion(post: Post) {
+        post.updateRevision()
+    }
+
+    /// An alert that is presented when a post has a version conflict, and the user needs to select discarding either the local or web version
+    static func presentConflictResolutionAlert(for post: Post,
```
leandroalonso, line 29: The name of the method is `presentConflictResolutionAlert` but it just returns an `UIAlertController`, right? It doesn't actually present the alert. We can rename it.

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
```diff
@@ -0,0 +1,70 @@
...
+
+        if let localDate = post.dateModified {
+            localDateString = PostListHelper.dateAndTime(for: localDate)
+        }
+        if let webDate = post.original?.dateModified {
+            webDateString = PostListHelper.dateAndTime(for: webDate)
+        }
+
+        let str = """
+        This post has two versions that are in conflict. Select the version you would like to discard.
```
leandroalonso, line 44: I'm afraid that this `String` will not be extracted by our i18n script. You can see a few examples of formatting strings, like here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Extensions/UIViewController+Notice.swift#L18

WordPress/Classes/ViewRelated/Post/PostListHelper.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+
+struct PostListHelper {
```
leandroalonso, line 3: While `openEditor` is directly related to the Post List I'm not sure the same applies to the date formatted. What do you think?

Any idea of how can we refactor that?

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
```diff
@@ -528,7 +528,7 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
             return
         }
 
-        PostListEditorPresenter.handle(post: post, in: self)
+        post.hasVersionConflict() ? PostListConflictResolver.handle(post: post, in: self) : PostListEditorPresenter.handle(post: post, in: self)
```
leandroalonso, line 531: Do you think the Post Lists really need to be aware of the version conflict?

I guess that it can delegate that to some entity to present the post and this entity will do any other work that needs to be done.

WordPress/WordPressTest/ViewRelated/Post/Controllers/PostListConflictResolverTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    override func tearDown() {
+        context = nil
+        TestContextManager.overrideSharedInstance(nil)
+        super.tearDown()
+    }
+
+    func testAlertPresentsCorrectOptions() {
+        let post = createConflictedPost()
```
leandroalonso, line 26: While testing the alert is good and valid I think that the most important thing to be testes is the logic, like:

1. Given a post in version conflict and I discard local, this entity discards the local copy
2. Given a post in version conflict and I discard the web version, this entity calls the `PostCoordinator` to update the post with the local revision

The latter might be tricky, and you would need to inject the PostCoordinator to be able to mock it.

WordPress/WordPressTest/ViewRelated/Post/Controllers/PostListHelperTests.swift
```diff
@@ -0,0 +1,24 @@
...
+        super.setUp()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
+
+    func testDateAndTimeFormatter() {
+        let date = Date.init(timeIntervalSince1970: 10)
+        let expectedDateString = "Dec 31, 1969 @ 7:00 PM"
```
leandroalonso, line 18: This test is failing for me. Probably due to timezone issues?

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
embryoconcepts: Agreed. I could see this being used on the Post editor, as well.

WordPress/Classes/ViewRelated/Post/PostListConflictResolver.swift
embryoconcepts: Will remove, thanks for background!

WordPress/Classes/ViewRelated/Post/PostListHelper.swift
embryoconcepts: The date formatter should really be in the other repo, but I'm not sure how to go about that. It was originally just stuck in the PostListEditorPresenter, so it wasn't in the right location in the first place.

WordPress/Classes/ViewRelated/Post/PostListViewController.swift
embryoconcepts: I'm not sure what this comment means - it's already shifting the presentation work to the PostListEditorPresenter or the PostListConflictResolver?

WordPress/WordPressTest/ViewRelated/Post/Controllers/PostListHelperTests.swift
embryoconcepts: I think I've resolved the timezone issue, but I'm still having problems visualizing the shape of the Post, and don't think I'm accessing the right value for the date/time of local changes. I _think_ the web date/time will be `post.dateModified` or `post.latest().dateModified` (shouldn't these be the same?), but I don't know how to get a `dateModified` for the local changes.

embryoconcepts: I almost wonder if it would be beneficial to have a convenience property clearly labeled `localChanges`, that could be accessed? There are properties for `hasLocalChanges` and `isLocalDraft`, so but no clear way to grab those changes.

WordPress/WordPressTest/ViewRelated/Post/Controllers/PostListConflictResolverTests.swift
embryoconcepts: I believe the tests are correct now, please let me know if you have feedback.

PR #14124, opened 2020-05-15
Test PR to verify state of Danger integration (take 2)

mokagio:
I'm opening this PR without a label and no changes compared to `develop`.
I'm expecting Danger to leave a comment regarding the missing label and no comment from the consistency checks.

---

Fixes #

To test:

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14110, opened 2020-05-13
Upgrade Danger & TypeScript packages

mokagio:
There is a new major version of Danger, 4.x. Now that I'm back working on the Peril to Danger migration, it's time for an upgrade.

To test: if the GitHub action build works, then we should be good.

_Update:_ All the builds [did work](https://github.com/wordpress-mobile/WordPress-iOS/actions?query=branch%3Aupgrade-danger-tools). Some of the commits show a ❌, but that's due to Circle cancelling the builds, which has nothing to do with Danger.

Screen Shot 2020-05-13 at 10 05 50 pm

To test the integration yourself, remove the label from the PR and wait for Danger to post the warning comment (you may have have to reload the page). Then add the label and verify that the comment gets removed.

---

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14128, opened 2020-05-16
Disable broken Danger to avoid noise for the dev team

mokagio:
Builds have been starting to fail with a 404, but nothing has changed on our end. Something might have changed on the GitHub end.

While we sort this out, it's better to disable the integration so we don't disturb the devs' workflow.

See:

- https://github.com/wordpress-mobile/WordPress-iOS/pull/14124
- https://github.com/wordpress-mobile/WordPress-iOS/pull/14110

**To test**: If the GitHub action passes and no Danger step runs, then this works

---

PR submission checklist:

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -34,14 +34,21 @@ jobs:
         # frozen-lockfile will make the build fail if the lockfile is not there
         run: yarn install --frozen-lockfile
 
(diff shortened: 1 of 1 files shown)


mindgraffiti:
Catching up. Thank you @mokagio for quieting Danger while GitHub had bugs! 😀

PR #14133, opened 2020-05-18
Migrates ContextManager to Swift

Gio2018:
Implements a new `CoreDataManager` type that is the Swift version of `ContextManager`. If the feature flag is enabled, `ContextManager` is replaced by `CoreDataManager`.

The same way, the correct mocks are applied during the tests based on the flag.

To test:

- try as many use cases as possible and verify that the app behaves as expected (I know, it's pretty broad..)

PR submission checklist:

- [x] ~~I have considered adding accessibility improvements for my changes.~~ This PR does not contain UI changes.
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ Not user-facing changes.


WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -0,0 +1,391 @@
...
+        migrateDataModelIfNecessary()
+
+        let container = NSPersistentContainer(name: Constants.name, managedObjectModel: managedObjectModel)
+        container.persistentStoreDescriptions = [storeDescription]
+
+        container.loadPersistentStores { [weak self] (storeDescription, error) in
+            guard let `self` = self, let error = error else {
+                return
+            }
+            DDLogError("⛔️ [CoreDataManager] loadPersistentStore failed. Attempting to recover... \(error)")
```
leandroalonso, line 42: One thing that we need to add here and test is the usage of [`SentryStartupEvent`](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/Logging/SentryStartupEvent.swift), the same way as we do in the `ContextManager.m`.

WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    private func performSave(context: NSManagedObjectContext, completion: (() -> Void)? = nil) {
+
+        let objects = Array(context.insertedObjects)
+        do {
+            try context.obtainPermanentIDs(for: objects)
+
+            guard context.hasChanges else {
+                return
```
aerych, line 128: In cases where the guard fails, or there is an error and the catch block executes, should we still all the completion block?  It seems like if we don't we could reach a dead end execution path in some parts of the app, similar to API service calls that return vs calling their success or failure methods.  Thoughts?

WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -0,0 +1,401 @@
...
+
+
+// MARK: - Notifications
+extension CoreDataManager {
+
+    private func observe() {
+        NotificationCenter.default.addObserver(forName: NSNotification.Name.NSManagedObjectContextDidSave,
+                                               object: self.mainContext,
+                                               queue: nil) { [weak self] notification in
+            guard let self = self else {
```
aerych, line 246: Should `self` be enclosed in `s?

WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -0,0 +1,401 @@
...
+        let bundle = Bundle(for: type(of: self))
+        guard let url = bundle.url(forResource: Constants.name, withExtension: "momd") else {
+            fatalError("Missing Model Resource")
+        }
+        return url
+    }
+
+    private var storeURL: URL {
+        guard let url = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else {
+            fatalError("Okay: Missing Documents Folder?")
```
aerych, line 277: Maybe a differently worded error message? I can imagine a future someone wondering why an error starts with "Okay"

WordPress/Classes/Utility/ContextManager.h
```diff
@@ -3,7 +3,22 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface ContextManager : NSObject
+@protocol CoreDataStack
+@required
+@property (nonatomic, readonly, strong) NSManagedObjectContext *mainContext;
+//@optional
```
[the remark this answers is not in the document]
Gio2018: thank you!!

WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -0,0 +1,401 @@
...
+                fatalError("☠️ [CoreDataManager] Recovery Failed! \(error) [\(error.userInfo)]")
+            }
+        }
+        return container
+    }()
+
+    private func childContext(with concurrencyType: NSManagedObjectContextConcurrencyType) -> NSManagedObjectContext {
+        let childManagedObjectContext = NSManagedObjectContext(concurrencyType: concurrencyType)
+        childManagedObjectContext.parent = self.mainContext
+        childManagedObjectContext.mergePolicy = NSMergeByPropertyObjectTrumpMergePolicy
```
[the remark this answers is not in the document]
Gio2018: Maybe.. It's been implemented like this, for now, as it's the Swift-migrated version of `newChildContextWithConcurrencyType` in `ContextManmager.m`, which contains the same policy. Should we change that one too?

WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -0,0 +1,401 @@
...
+        }
+    }
+
+    private func performSave(context: NSManagedObjectContext, completion: (() -> Void)? = nil) {
+
+        let objects = Array(context.insertedObjects)
+        do {
+            try context.obtainPermanentIDs(for: objects)
+
+            guard context.hasChanges else {
```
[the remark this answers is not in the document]
Gio2018: Yup! Thanks!!

WordPress/Classes/Utility/CoreDataManager.swift
Gio2018: Thanks for spotting this!! In fact it was actually happening..

WordPress/Classes/Utility/CoreDataManager.swift
Gio2018: Since `Swift 4.2`, it is allowed to use the same keyword without back ticks
https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md

aerych: til!

WordPress/Classes/Utility/CoreDataManager.swift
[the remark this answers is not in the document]
aerych: I'm honestly not sure. @astralbodies ? Thoughts?

WordPress/Classes/Services/NotificationSyncMediator.swift
```diff
@@ -25,7 +25,7 @@ let NotificationSyncMediatorDidUpdateNotifications = "NotificationSyncMediatorDi
 class NotificationSyncMediator {
     /// Returns the Main Managed Context
     ///
-    fileprivate let contextManager: ContextManager
+    fileprivate let contextManager: CoreDataStack
```
astralbodies, line 28: Should we change all of the contextManager properties to be "coreDataStack" to be consistent with what it is?

WordPress/Classes/Utility/CoreDataManager.swift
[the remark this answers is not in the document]
astralbodies: Honestly if I were to write this from scratch I'd probably pick `NSMergeByPropertyObjectTrumpMergePolicy` because what's in memory is probably "newer" than what's on disk.

WordPress/Classes/Services/NotificationSyncMediator.swift
leandroalonso: We can do that as soon as we decide to keep only the Swift initialization. Wdyt?

I just want to prevent changing that, then (I hope not, but...) for some reason reverting this change, then having to rename all `coreDataStack` back to `contextManager`.

And since we're still using `ContextManager.sharedInstance` all over the code it would be good to change that too just to keep it consistent.

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
```diff
@@ -48,9 +45,33 @@ startup time. This will block the thread. Do not use unless you're sure.
...
+
+     - Parameters:
+     - error: The error object
+     - userInfo: A dictionary containing additional data about this error.
+     - level: The level of severity to report in Sentry (`.error` by default)
+    */
+    static func logErrorAndWait(_ error: Error, userInfo: [String: Any]? = nil, level: SentrySeverity = .error) {
+        let event = Event(level: .error)
+        event.message = error.localizedDescription
+        event.extra = userInfo ?? (error as NSError).userInfo
```
jleandroperez, line 64: Since this API is only consumed by the method above ` send(title: String)`, and the NSError [instantiated here](https://github.com/wordpress-mobile/WordPress-iOS/pull/14133/files#diff-9631a66a3adecbce84ba4545aceff4e3R48) carries the same String for `domain` / `localizedDescription`, would it be possible to simplify the signature a bit?

```
static func logErrorAndWait(_ description: String, userInfo: [String: Any]? = nil, level: SentrySeverity = .error) {
```

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
```diff
@@ -48,9 +45,33 @@ startup time. This will block the thread. Do not use unless you're sure.
...
+        }
+
+        let semaphore = DispatchSemaphore(value: 0)
+
+        Client.shared?.send(event: event) { _ in
             semaphore.signal()
-        })
+        }
 
         semaphore.wait()
```
jleandroperez, line 76: If `Client.shared` is nil, this line could cause a deadlock. WDYT if we bail out if the `Client.shared` is nil? (We can call `fatalError`, not sure who sets the Client, or when!)

```
guard let client = Client.shared else {
   fatalError() 
}
```

WordPress/WordPressTest/CoreDataManagerMock.swift
```diff
@@ -0,0 +1,122 @@
...
+        set {
+            _persistentStoreCoordinator = newValue
+        }
+
+        get {
+            return _persistentStoreCoordinator ?? _persistentContainer.persistentStoreCoordinator
+        }
+    }
+
+    var _mainContext: NSManagedObjectContext?
```
jleandroperez, line 33: Perhaps we wanna set to private all of the internal instances? (`var _XXXX`)

WordPress/WordPressTest/TestContextManager.m
```diff
@@ -30,90 +26,97 @@ - (instancetype)init
...
 {
     [self saveContext:context withCompletionBlock:^{
-        if (self.testExpectation) {
-            [self.testExpectation fulfill];
-            self.testExpectation = nil;
-        } else if (self.requiresTestExpectation) {
+        if (self->_stack.testExpectation) {
+            [self->_stack.testExpectation fulfill];
+            self->_stack.testExpectation = nil;
+        } else if (self->_stack.requiresTestExpectation) {
```
jleandroperez, line 73: Perhaps `self.` instead of the old school `->`?

jleandroperez: (Same note in the methods below!)

leandroalonso: I actually had to use `self->`, the compiler didn't liked the `self.` and I have no idea why

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
leandroalonso: My idea was to kept the same signature because after this PR being merged I'll move this method to `WordPressTracks`, and in that case, it would be good to keep it generalized. Wdyt?

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
leandroalonso: I just used the same logic and kept that way for consistency, ref: https://github.com/Automattic/Automattic-Tracks-iOS/blob/develop/Automattic-Tracks-iOS/Crash%20Logging/CrashLogging.swift#L130

(Since I'm going to move this method to there)

jleandroperez: This one is sliiiightly (slightly) different because of the Semaphore. A nil in there could result in a deadlock (no idea if that could ever happen, though!), but it wouldn't generate a crashlog. I think!

An in between approach would maybe be... using a force unwrap when calling `Client.shared`, so that worst case scenario we'd get a crashlog. WDYT?

leandroalonso: Oooh ok, I see your point now!

Makes sense, done in b41c4d5fa71090775442ba78f21dd65121861998

I just changed it slightly: instead of crashing, if the Client doesn't exist the method return and then the execution will follow (and probably it will crash on the original crash).

WordPress/WordPressTest/CoreDataManagerMock.swift
leandroalonso: Done in 9bfcef3d2ca221bdd070318683aff57179962657

WordPress/WordPressTest/TestContextManager.m
leandroalonso: Nevermind, done in 363c7fb1391760c0a5a27e38cfa1581860c03b12

WordPress/Classes/Utility/CoreDataManager.swift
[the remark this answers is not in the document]
aerych: I've gone back through the older code. Looks like we added the NSMergeByPropertyObjectTrumpMergePolicy in https://github.com/wordpress-mobile/WordPress-iOS/pull/1183 to deal with a bunch of merge errors we were seeing logged. 

Removing the policy might reveal a new set of merge issues that we're unaware of because the policy has been handling them.  This might be nice to do for developer builds (fix all the things!) but I think I've come around to keeping the policy for production builds since (as @Gio2018 pointed out) its been like that for a while now. :)   Maybe we can revisit in the future.

WordPress/Classes/Utility/Logging/SentryStartupEvent.swift
jleandroperez: Thanks for doing this @leandroalonso !!

PR #14135, opened 2020-05-18
Merge release 14.8.1 into 14.9

loremattei:
Merge 14.8.1 hotfix into 14.9

PR submission checklist:


loremattei:
This is what you get for trying to use new tools at late night 😅
I didn't notice that the files where not saved after the conflict was solved... I usually use a different tool that autosaves when I accept one revision...

It should be good now!

PR #14140, opened 2020-05-19
Gutenberg - Media Inserter - Check for internet connection

geriux:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2106

When uploading media from the block editor without **internet connection**, the placeholder will hang instead of showing the retry message. This affects any block that handles media upload, `Image`, `Video`, `Media & Text`, and `Cover`.

This PR adds an internet connection check in the `thumbnailReady` state to avoid showing the progress bar when it is not uploading anything and display the retry message instead.

For a bit more insight, when uploading media from the media library outside the editor it won't let the user upload anything without an internet connection. In this case, we do want to start the upload process but let the user know if it failed by having no connection.

### To test
- Open the app.
- Open the editor.
- Add one of these blocks: `Image`, `Video`, `Media & Text` or `Cover`.
- Turn on Airplane mode or disable any internet connection from the device.
- Tap on the block and upload a file from the device.
- **Expect** the block to display a retry message.
- Turn on the internet connection.
- Tap on the block and retry the upload.
- **Expect** the media to be uploaded successfully.

| Before | After |
|     :---:      |     :---:      |
|  |   |

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 15.0
 -----
- 
+* [**] Block editor: Fix media upload progress when there's no connection.
+
 14.9
 -----
 * Streamlined navigation: now there are fewer and better organized tabs, posting shortcuts and more, so you can find what you need fast.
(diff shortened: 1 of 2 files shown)


geriux:
Thanks for the review @chipsnyder !

Issue #13825, opened 2020-04-06
Testing: Tab bar navigation UI test fails inconsistently


rachelmcr:
We have started seeing the `testTabBarNavigation()` navigation test fail inconsistently in `develop` UI test runs, on the Reader step. We expect the test to open the Reader tab, showing the Followed Sites screen. Instead, it opens the Reader search screen:

<img src="https://user-images.githubusercontent.com/8658164/78581887-61078b80-782c-11ea-879c-48d600188f93.jpg" width="50%">

Because the keyboard is on screen, it blocks any further navigation via the bottom tab nav bar.

I disabled this test in https://github.com/wordpress-mobile/WordPress-iOS/pull/13824, pending further investigation. Issues to consider:

* We should investigate the expected app behavior for the Reader, i.e. is what we're seeing in these failing tests a bug or a scenario that the test isn't accounting for?
* We should ensure the test checks that the correct Reader screen is opened (not just any Reader screen).
* We should update the test to only navigate through the tabs in the bottom nav bar. (The test still opens the block editor, but that's no longer part of the bottom nav bar. The editor is exercised as part of the editor tests, so it shouldn't be required here.)

PR #14144, opened 2020-05-19
Testing: Fix and enable MainNavigationTests - TestTabBarNavigation

geriux:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/13825

This test case was [disabled](https://github.com/wordpress-mobile/WordPress-iOS/pull/13824) because it was failing inconsistently. This PR enables it again and also addresses suggestions to improve it from the issue:

> We should ensure the test checks that the correct Reader screen is opened (not just any Reader screen).

Now it will check the Reader screen has loaded.

> We should update the test to only navigate through the tabs in the bottom nav bar. (The test still opens the block editor, but that's no longer part of the bottom nav bar. The editor is exercised as part of the editor tests, so it shouldn't be required here.)

It removes opening the block editor and just navigates throughout the navigation bar.

To test:

MainNavigationTests > testTabBarNavigation should pass in the UI Tests.

PR submission checklist:

### DIFF
--- a/WordPress/WordPressUITests/Screens/ReaderScreen.swift
+++ b/WordPress/WordPressUITests/Screens/ReaderScreen.swift
@@ -10,4 +10,8 @@ class ReaderScreen: BaseScreen {
 
         super.init(element: readerTable)
     }
+    
+    static func isLoaded() -> Bool {
+        return XCUIApplication().tables["Reader"].exists
+    }
 }
--- a/WordPress/WordPressUITests/Screens/ReaderScreen.swift
+++ b/WordPress/WordPressUITests/Screens/ReaderScreen.swift
@@ -10,7 +10,7 @@ class ReaderScreen: BaseScreen {
 
         super.init(element: readerTable)
     }
-    
+
     static func isLoaded() -> Bool {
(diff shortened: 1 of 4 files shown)


geriux:
Hey @etoledom 👋 I added you as a reviewer for this PR, let me know if you would be able to review it, if not I'll assign someone else =) Thank you!

WordPress/WordPressUITests/Screens/ReaderScreen.swift
```diff
@@ -10,4 +10,8 @@ class ReaderScreen: BaseScreen {
 
         super.init(element: readerTable)
     }
+
+    static func isLoaded() -> Bool {
+        return XCUIApplication().tables["Reader"].exists
```
etoledom, line 15: Maybe we can have a `ElementStringIDs` struct/enum similar to the one declared on `MySiteScreen` to avoid repeating literal strings?

geriux: Thanks for reviewing @etoledom ! I'll update the code to have it just like `MySiteScreen`

PR #14149, opened 2020-05-19
Add UITest for checking removal of featured images in posts.

SergioEstevao:
Added extra steps for the Aztec and Gutenberg editors on Posts to check if the removal of featured images is working correctly.

To test:
  - Check if UITests work correctly.

PR submission checklist:


WordPress/WordPressUITests/Screens/Editor/EditorPostSettings.swift
```diff
@@ -6,6 +6,9 @@ class EditorPostSettings: BaseScreen {
     let categoriesSection: XCUIElement
     let tagsSection: XCUIElement
     let featuredImageButton: XCUIElement
+    var changeFeaturedImageButton: XCUIElement {
+        return settingsTable.cells["CurrentFeaturedImage"]
+    }
```
rachelmcr, line 11: Nitpick: Any reason why this is declared this as a `var` like this instead of `let` like the other elements on the screen? Not a huge deal but it would be nice to be consistent.

WordPress/WordPressUITests/Screens/Editor/FeaturedImageScreen.swift
```diff
@@ -0,0 +1,20 @@
...
+import XCTest
+
+class FeaturedImageScreen: BaseScreen {
+
+
+    let navigationBar: XCUIElement
+
+    init() {
+        let app = XCUIApplication()
+        navigationBar = app.navigationBars["Featured Image"]
```
rachelmcr, line 11: Since the navigation bar titles are localized strings, it would be nice to use something else with an accessibility identifier to verify we're on the correct screen. (We'd like to be able to run UI tests in any language.)

I'd suggest using the "Remove Featured Image" button instead, since it's specific to this screen and you know you'll need to use it on this screen.

WordPress/WordPressUITests/Screens/Editor/FeaturedImageScreen.swift
```diff
@@ -0,0 +1,20 @@
...
+        super.init(element: navigationBar )
+    }
+
+    func tapRemoveFeaturedImageButton() {
+        navigationBar.buttons["Remove Featured Image"].tap()
+        app.sheets["Remove this Featured Image?"].scrollViews.otherElements.buttons["Remove"].tap()
```
rachelmcr, line 17: Similar to the navigation bar, it would be good to avoid using localized strings here. For action sheets, we've been identifying them like this (tapping the first button on the action sheet):

```suggestion
        app.sheets.buttons.element(boundBy: 0).tap()
```

WordPress/WordPressUITests/Screens/Editor/EditorPostSettings.swift
SergioEstevao: It's a var because it's a dynamic variable, every time you access it, it returns a new value.
I need to do this because the cell is not available when you go in the screen, only after you set an featured image.

WordPress/WordPressUITests/Screens/Editor/EditorPostSettings.swift
```diff
@@ -41,6 +44,14 @@ class EditorPostSettings: BaseScreen {
         return TagsComponent()
     }
 
+    func removeFeatureImage() -> EditorPostSettings {
+        changeFeaturedImageButton.tap()
+        FeaturedImageScreen()
+            .tapRemoveFeaturedImageButton()
+
```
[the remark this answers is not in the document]
SergioEstevao: I went for option 1.

WordPress/WordPressUITests/Screens/Editor/FeaturedImageScreen.swift
SergioEstevao: Done.

PR #14151, opened 2020-05-20
Fix State Restoration issue + feature flag check

### LINKED ISSUE #14150: State Restoration does not work on Blog Details or any of its descendants in the navigation hierarchy
### Expected behavior

App should reopen to the Stats screen when restoring from saved state.

### Actual behavior

App does not reopen to the Stats screen when restoring from saved state.

- Navigate to the Stats Screen inside from the My Sites Screen
- Send the app to the background
- Kill the app in Xcode to trigger state restoration

bjtitus:
Fixes #14150 

This fixes an issue where state restoration would fail on Blog Details View Controller (and view controllers below it in the navigation stack) because the restoration properties were not set.

There are two additional preventative changes.

### Changes

- Calls the proper initializer in `BlogDetailsViewController` so that `restorationClass` and restorationIdentifier` are set.
- Defaults to hiding the create button instead of leaving it shown.
- Also adds a check for the CreateButton feature flag to prevent a possible crash if that feature flag is disabled.

#### State Restoration
- Navigate to the Blog Details or Stats Screens inside the My Sites Screen
- Send the app to the background
- Kill the app in Xcode to trigger state restoration
- Notice that the Stats Screen is restored (this _does not_ happen in current the 14.9 branch)

#### `floatingCreateButton` Feature Flag
- Disable the `floatingCreateButton` feature flag. The app should no longer crash when launching to the Blog Details screen. It would previously as it tries to add the tooltip to a button which is not on screen.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
@@ -257,7 +257,7 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 
 - (id)initWithMeScenePresenter:(id<ScenePresenter>)meScenePresenter
 {
-    self = [super init];
+    self = [self init];
     self.meScenePresenter = meScenePresenter;
     return self;
 }
@@ -377,7 +377,7 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
-    if ([self.tabBarController isKindOfClass:[WPTabBarController class]]) {
+    if ([self.tabBarController isKindOfClass:[WPTabBarController class]] && [Feature enabled:FeatureFlagFloatingCreateButton]) {
         [self.createButtonCoordinator showCreateButton];
     }
     [self createUserActivity];
--- a/WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
@@ -83,8 +83,10 @@ import WordPressFlux
         }
     }
 
+    /// Button must be manually shown _after_ adding using `showCreateButton`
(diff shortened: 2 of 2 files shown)


WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -257,7 +257,7 @@ - (void)dealloc
 
 - (id)initWithMeScenePresenter:(id<ScenePresenter>)meScenePresenter
 {
-    self = [super init];
+    self = [self init];
```
bjtitus, line 260: Fixes state restoration issue (https://github.com/wordpress-mobile/WordPress-iOS/issues/14150)

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -377,7 +377,7 @@ - (void)viewWillAppear:(BOOL)animated
 - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
-    if ([self.tabBarController isKindOfClass:[WPTabBarController class]]) {
+    if ([self.tabBarController isKindOfClass:[WPTabBarController class]] && [Feature enabled:FeatureFlagFloatingCreateButton]) {
```
bjtitus, line 380: Fixes a potential crash when `floatingCreateButton` flag is `false`.

WordPress/Classes/ViewRelated/System/Floating Create Button/CreateButtonCoordinator.swift
```diff
@@ -83,8 +83,10 @@ import WordPressFlux
         }
     }
 
+    /// Button must be manually shown _after_ adding using `showCreateButton`
     @objc func add(to view: UIView, trailingAnchor: NSLayoutXAxisAnchor, bottomAnchor: NSLayoutYAxisAnchor) {
         button.translatesAutoresizingMaskIntoConstraints = false
+        button.isHidden = true
```
bjtitus, line 89: Defaults the Floating Create Button to hidden. This requires that the `showCreateButton` method _must_ be called before the button is shown. The only visual change is, when restoring to the Stats screen and hitting the "Back" button, the FAB now appropriately animates in.

PR #14155, opened 2020-05-20
Issue/14141 detect change of featured image

### LINKED ISSUE #14141: Changing a featured image on a existing post is not detected as a change
### Expected behavior

If editing an existing post and we change the featured image, then the editor should detect this change and activate the Publish/Save button to allow to update it.

### Actual behavior

After a change is done to a featured image to an existing post, the editor doesn't detect the change and the Publish button is not active.

 - Create a new post and publish it.
 - Open the same post
 - Open post settings and add a featured image
 - Close Post settings
 - See that the Update button is not active.

SergioEstevao:
Fixes #14141

This PR adds checks for feature image changes to the method `hasLocalChanges` of `AbstractPost`.

It also updates the `GutenbergViewController` code to refresh the navigation bar every time it comes back from `PostSettings`.

To test:
 - Please test the steps bellow with Aztec and Gutenberg editor, and with **post** and **pages**.

 - Create a new post and publish it.
 - Open the same post
 - Open post settings and add a featured image
 - Close Post settings
 - Check that the Update on the top right is active.
 - Update the post
 - Exit the post and open it again
 - Open Post settings
 - Tap on the Featured Image
 - Remove it
 - Close Post Settings
 - Check that the update on the top right is active.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/AbstractPost.m
+++ b/WordPress/Classes/Models/AbstractPost.m
@@ -572,8 +572,8 @@
         return YES;
     }
 
-    if ( (self.featuredImage == NULL && self.original.featuredImage != NULL) ||
-        ![self.featuredImage.objectID isEqual: original.featuredImage.objectID] ) {
+    if ( ((self.featuredImage != nil) && ![self.featuredImage.objectID isEqual: original.featuredImage.objectID]) ||
+        (self.featuredImage == nil && self.original.featuredImage != nil) ) {
         return YES;
     }
 
--- a/WordPress/Classes/Models/AbstractPost.m
+++ b/WordPress/Classes/Models/AbstractPost.m
@@ -572,6 +572,11 @@
         return YES;
     }
 
+    if ( (self.featuredImage == NULL && self.original.featuredImage != NULL) ||
+        ![self.featuredImage.objectID isEqual: original.featuredImage.objectID] ) {
+        return YES;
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Models/AbstractPost.m
```diff
@@ -572,6 +572,11 @@ - (BOOL)hasLocalChanges
         return YES;
     }
 
+    if ( ((self.featuredImage != nil) && ![self.featuredImage.objectID isEqual: original.featuredImage.objectID]) ||
```
leandroalonso, line 575: What do you think about adding unit tests to ensure this logic? Under `PostTests`.

Something like:

```
    /// When removing the featured image hasLocalChanges returns true
    func testLocalChangesWhenfeaturedImageIsRemoved() {
        let post = newTestPost()
        post.featuredImage = Media.makeMedia(in: context)
        let revision = post.createRevision()
        revision.featuredImage = nil

        XCTAssertTrue(revision.hasLocalChanges())
    }
```

And besides that test we can:

1. Test when the post originally didn't had any featured image and the user adds one
2. The user sets the same featured image (based on objectID) and hasLocalChanges returns false

SergioEstevao: Good call, I will add the tests to it!

Issue #14167, opened 2020-05-22
Homepage Settings Master Issue


frosty:
Implement a settings screen to allow the user to configure their Homepage settings via the app. This will currently only support WordPress.com and Jetpack websites, due to limitations in the XMLRPC API.

![blueprint_ios](https://user-images.githubusercontent.com/4780/82658550-15b00d80-9c1f-11ea-8edf-e4825e8d3c89.png)

- [x] Add a new row to the Site Settings screen, showing the value of the current setting. This should only be visible for wpcom and Jetpack sites, and only to admin users.
- [x] Add a new Homepage Settings view controller, which includes:
- [x] Rows to toggle between posts and page homepage types
- [x] When page homepage type is selected, shows rows which display the current Home and Posts page selections.
- [x] When those rows are tapped, show a list of all pages and allow the user to select which one should be used for the Home / Posts page.
- [x] Also requires integration with the wpcom `/sites/:/homepage` endpoint to set these values.

jkmassel:
Howdy friends – we're freezing `15.0` today, so I'm bumping this PR forward to `15.1`. If it needs to land in the `15.0` series, please merge it to `release/15.0` and DM me – I'll be happy to create a new beta release!

frosty:
@jkmassel All of the PRs for this issue were merged in time for 15.0, but we neglected to update the issue itself. I switched the milestone back and I'm marking as closed.

ceyhun:
> This will currently only support WordPress.com and Jetpack websites, due to limitations in the XMLRPC API.

Hey there @bjtitus @frosty 👋 Do you think it will ever be possible to add Homepage settings to non-jetpack self-hosted sites somehow?

cc @chipsnyder

PR #14169, opened 2020-05-22
Homepage Settings #1

frosty:
Refs #14167. This PR adds some of the functionality for the Homepage Settings feature, detailed in the linked issue.

homepage-settings-3

* Added a new row to Site Settings, allowing the user to navigate to Homepage Settings. Also shows the current value of the 'homepage type' setting
* Added a new Homepage Settings screen, which allows the user to toggle their homepage type between posts and pages. I've done the toggle optimistically, so it updates the model straight away and reverts the change if the API request fails.
* If the homepage type is set to Pages, we show rows that display the current selection for the Home and Posts pages, if any have been selected.
* We fetch the list of all pages on the site when the Homepage Settings screen is displayed so that we can find the titles of any currently selected pages.

**What's not included**

* Navigation into the Home and Posts page rows in order to select a different page – that will be coming in a followup PR.
* PR coming up shortly for the associated WordPressKit changes.

#### Scenario 1

* Self-hosted non-Jetpack site – ensure that the Homepage Settings row on the Site Settings screen is not visible

#### Scenario 2

* With a Jetpack or WordPress.com site where you have never adjusted this setting, visit the Site Settings screen.
* Ensure that the Homepage Settings row shows 'Classic Blog'
* Tap into the Homepage Settings row and ensure the Classic Blog selection is reflected on the Homepage Settings screen
* Tap Static Homepage and check that the Homepage and Posts Page rows appear, and that they have no detail value
* Tap back out to Settings and ensure that the Homepage Settings row now reads Static Homepage

#### Scenario 3

* On the web, visit the Customizer for your test site, and find the Homepage Settings section there. Change it to Static Homepage, and choose both a Homepage and a Posts Page. Don't forget to save your changes!
* In the app, refresh your sites list and then navigate into your test site
* Go to Site Settings, and check that the Homepage Settings row says "Static Homepage"
* Tap into Homepage Settings, and ensure that both of your selected page titles are shown.
* Change the site back to Classic Blog
* Reload the customizer on the web for the site, and check it now shows Classic Blog there too

#### Scenario 4

* Build and run on your device, and put it into airplane mode
* Navigate into Site Settings for your site, and tap Homepage Settings
* Try changing the homepage type. The operation should fail (the setting should not change), and a notice should be displayed

PR submission checklist:

- [x] I have considered adding unit tests where possible – I started working on some tests, but I can submit those in a followup PR. I'd like to get these first steps merged for now.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary – will add to the final PR for this feature


WordPress/Classes/ViewRelated/Blog/Site Settings/HomepageSettingsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    var homepageTypeRows: [ImmuTableRow] {
+        guard let homepageType = blog.homepageType else {
+            return []
+        }
+
+        return [
+            CheckmarkRow(title: HomepageType.posts.title, checked: homepageType == .posts, action: { _ in
+                self.setHomepageType(.posts)
```
bjtitus, line 99: I think these `self` references need to be `weak` to prevent a retain cycle.

I'm seeing `HomepageSettingsViewController` sticking around after leaving and it looks like a few copies are also made when visiting repeatedly:

Screen Shot 2020-05-26 at 8 05 08 PM

frosty:
Thanks for catching this! Changes pushed, and I don't see any HomepageSettingsViewControllers hanging around any more.

PR #14185, opened 2020-05-26
Update Reader Card Design

emilylaguna:
**Fixes** #14005

**Subtasks:**
- [X] Relative Date Formatter: https://github.com/wordpress-mobile/WordPress-iOS/pull/14186
- [x] Update cross post design: https://github.com/wordpress-mobile/WordPress-iOS/pull/14207

This applies the new reader card design in both dark and light modes. 

### Light Mode
| Before | After |
|---|---|
|  before-discover-light | Simulator Screen Shot - iPhone 11 Pro - 2020-05-25 at 19 36 39 |

### Dark Mode
| Before | After |
|---|---|
| before-discover-dark|Simulator Screen Shot - iPhone 11 Pro - 2020-05-25 at 19 36 19|

### Other Card States

| Default | Text Only | Photo and title only | 
|---|---|---|
| Screen Shot 2020-05-25 at 7 38 41 PM|Screen Shot 2020-05-25 at 7 43 14 PM |Screen Shot 2020-05-25 at 7 44 58 PM|

| Photo Only | No site icon w/ comments disabled | Title only |
|---|---|---|
|Screen Shot 2020-05-25 at 7 46 18 PM|Screen Shot 2020-05-25 at 7 46 48 PM|Screen Shot 2020-05-25 at 7 41 56 PM|

### Other Devices
| iPhone 5s | iPad |
|---|---|
| Simulator Screen Shot - iPhone 5s - 2020-05-25 at 20 07 13 | Simulator Screen Shot - iPad Air (3rd generation) - 2020-05-25 at 20 11 02|

### Notes
- This does not include the updated relative time logic yet
- The action icons on the discover view are disabled even though when you tap on the post itself its enabled. This is because the discover post itself has comments disabled, but when you tap on it, it loads the attributed post.
- Some elements have not been updated such as the attribution label since the design wasn't updated for them yet.

## To test:
1. Launch the app
2. Tap on the Reader tab
3. Tap on any section
4. Scroll and make sure the card content and design matches
5. Close the app
6. Enter dark mode
7. Follow steps 1 - 4 to make sure the card looks good in dark mode

## PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14186, opened 2020-05-26
Implements the new relative date logic for reader card improvements

emilylaguna:
Subtask for #14185

The logic used (as defined by the `figma - card anatomy` section)

- If date is within the past hour display: `{Val}m`
	- `m` is an abbreviation for minutes 
	-  **Example:** 1m, 30m, 59m
- If date is  within the past 24 hours display: `{Val}h`
	- `h` is abbreviation for hours
	- **Example:** 1h, 5h, 20h
- If date is older than 24 hours but earlier than 7 days: `{Val}d`
	- `d` is an abbr for days
	- **Example:** 1d, 2d, 6d
- If date is older than 7 days but within the current year: `MMM DD`
	- **Example:** Feb 11, Jan 2
- If date is not from the current year `MMM DD, YYYY` or the users chosen `medium` style date format
	- **Example:** Feb 11 2019, Jan 2, 2010

### Screenshots
| Within an hour | Within 24 hours | With 7 days | Older than 7 days | Last Year |
|---|---|---|---|---|
| Screen Shot 2020-05-25 at 9 15 06 PM|  Screen Shot 2020-05-25 at 9 11 18 PM | Screen Shot 2020-05-25 at 9 11 59 PM | Screen Shot 2020-05-25 at 9 12 10 PM |  Screen Shot 2020-05-25 at 9 13 47 PM |

### To test:
1. Launch the app
2. Tap on the Reader tab
3. Tap on the Discover tab
4. For recent posts you should see `{Val}m` or `{Val}h` depending on how new/older the post is
5. Continue scrolling down to the older posts
6. For posts made after a day you should see `{Val}d` 
7. Continue scrolling down to the older posts
8. For posts made more than a week ago you should see the month and day: `May 18`
9. Continue scrolling down to the older posts
10. For posts made in Dec 2019 you should see the full date and year: `Dec 19, 2019` 

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderRelativeTimeFormatter.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+
+class ReaderRelativeTimeFormatter: NSObject {
```
leandroalonso, line 3: I don't think this is coupled to Reader, `RelativeTimeFormatter` would be ok. :)

emilylaguna: I thought a lot about the naming of the file and I should have mentioned it in the description. 

So, while this is a general "relative"-ish formatter it currently contains logic very specific to the reader, there's no configuration options, and defaults to displaying "just now" for any potential future dates. 

I don't want to confuse anyone in the project who sees `RelativeTimeFormatter` to think it's going to work like `NSRelativeDateTimeFormatter` or `TTTimeIntervalFormatter` 

If this logic is going to be applied in other locations in the app I can change this. 

So, at this point I'd think it's best to keep it relative to the reader. Thoughts?

leandroalonso: Agree

PR #14181, opened 2020-05-25
PR to test new approach to Danger integration

mokagio:
The new approach is to use a single remote `Dangerfile` with multiple imports. This should take care of both keeping the setup tidy, no new `danger ci` runs for new checks, and solving the multiple comments issue, because now we'll have a single `danger` instance running all the checks.

---

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14187, opened 2020-05-26
Re-enable Danger with new centralized setup

mokagio:
Hat tip to @loremattei for thinking of this option. Now instead of having new steps in the GitHub Actions workflow for each check we want to run with Danger, we simply execute a [dedicated remote Dangerfile](https://github.com/Automattic/danger-settings/blob/a2c8edff32be081595509d34bf25bc18a6c1bd1e/wordpress/ios.ts).

The checks run by this file are currently:

- [labels](https://github.com/Automattic/danger-settings/blob/a2c8edff32be081595509d34bf25bc18a6c1bd1e/org/label.ts)
- [iOS specific safety checks](https://github.com/Automattic/danger-settings/blob/a2c8edff32be081595509d34bf25bc18a6c1bd1e/org/pr/ios.ts)

I had to make a new repo, `danger-settings` because in order to get this setup to work I had to make a few changes to the code originally in [`peril-settings`](https://github.com/Automattic/peril-settings/).

I actually think the dedicated repo is not that bad of an idea. For once, the code there is meant to be used by Danger itself, not Peril –even though it's just a matter of semantics. It's also the only way to make sure we don't break Peril in the meantime, since I couldn't find a way to make remote imports work from a branch other than the base one.

### To test

Add/or remove a label to see whether Danger makes the appropriate comment.

The milestone check doesn't work properly, I believe due to the fact that the milestone even is bound to issues not PRs, and we have no handling for it yet. ~~I think this is worth fixing in a dedicated PR. I'm happy to disable the milestone check on the `danger-settings` side if you think this is confusing.~~ **Update:** Actually, I [disabled it](https://github.com/Automattic/danger-settings/commit/fc43c7bbb8bcaad66188f9bd9fbf3f8245d7cce7) because it was too confusing for me. I think it's best to leave it out and fix/enable it in a dedicated PR so this one is clearer to tests, and also the devs won't get confusing "missing milestone" warnings when they actually have the milestone there.

You can also push a [commit like this one](https://github.com/wordpress-mobile/WordPress-iOS/pull/14181/commits/69566fd4bfb49a7fdd5c8a4fd16832e379e624b1) to verify that the iOS safety checks do run.

Overall, for what concerns the rules above, the comments made by Danger and Peril should be consistent.

E.g.:
Screen Shot 2020-05-26 at 3 51 21 pm

---

PR submission checklist:

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -11,6 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # Till https://github.com/danger/danger-js/issues/1042 is fixed, we need
+      # this workaround
+      DANGER_GITHUB_API_BASE_URL: "https://api.github.com"
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -37,21 +37,7 @@ jobs:
         # frozen-lockfile will make the build fail if the lockfile is not there
         run: yarn install --frozen-lockfile
 
-      # Disabled because of issues that look like outside of our control on the
-      # GitHub end.
-      #
-      # See:
-      # - https://github.com/wordpress-mobile/WordPress-iOS/pull/14110
-      # - https://github.com/wordpress-mobile/wordpress-ios/pull/14124
-      #
-      # - name: Validate Labels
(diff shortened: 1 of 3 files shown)


mokagio:
_I'm using this PR as a test case for https://github.com/Automattic/peril-settings/pull/49, expect a few confusing changes of Milestone and label here. Sorry for the noise, but I need a PR that is open do to the testing._

PR #14191, opened 2020-05-26
Do not allow scheduling a private post

### LINKED ISSUE #13966: Editor: Schedule button title truncated when scheduling from prepublish nudge flow

leandroalonso:
Fixes #13966 

### To test

Please, test with the Prepublishing Nudges and the Post Settings.

1. Schedule a post
2. Change the visibility to private
3. Check that the post is not scheduled anymore and you can't schedule
4. Change the visibility to public
5. Check that you can schedule again

### DIFF
--- a/WordPress/Classes/Extensions/UITableViewCell+enableDisable.swift
+++ b/WordPress/Classes/Extensions/UITableViewCell+enableDisable.swift
@@ -1,6 +1,6 @@
 import WordPressShared
 
-extension WPTableViewCell {
+extension UITableViewCell {
     /// Enable cell interaction
     @objc func enable() {
         isUserInteractionEnabled = true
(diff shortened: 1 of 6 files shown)


yaelirub:
Tested according to described steps and works as expected. 
The only thing is that there is some weirdness when publishing a private post with scheduled date from web. It will appear in the published tab and when going to post settings it would show the scheduled date disabled  (though it was published).

I consider this an edge case and outside of the scope of this as the way it's handled on web is not super clear( the date is not updated to "immediately" although they notify the user)

PR #14194, opened 2020-05-26
Show Reader content in a WebView

### LINKED ISSUE #14183: [Reader WebView] Create a new screen that shows a webview and display the content
## Acceptance criteria

1. When a user selects a Reader post
2. Show a view containing an WebView with the content inside

## Additional information

No styles are expected in this task.

leandroalonso:
Fixes #14183

This adds the additional architecture for rendering post content inside a WebView. Some important things:

* It's all under a FeatureFlag that is *always* `false`
* My main goal is to keep the ViewController as simple as possible
* The coordinator should handle service and other stuff

### To test

First, you need to override the feature flag value. Open any site > tap on your profile (top-right icon) > tap "App Settings" > tap "Debug" (last option) > Toggle "Reader content displayed in a WebView" ON. [Like this](https://user-images.githubusercontent.com/7040243/82946518-e0842200-9f74-11ea-82b3-8108ba877568.png).

Then, open the Reader and tap some articles, you should see the content being displayed there (with no style at all).

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -0,0 +1,74 @@
...
+        service.fetchPost(postID.uintValue,
+                          forSite: siteID.uintValue,
+                          isFeed: isFeed,
+                          success: { [weak self] post in
+                            guard let post = post else {
+                                return
+                            }
+
+                            self?.view?.render(post)
+        }, failure: { [weak self] error in
```
yaelirub, line 69: Should we check / pass the error here?

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -0,0 +1,74 @@
...
+            self?.view?.showError()
+        })
+    }
+
```
yaelirub, line 73: extra line

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -0,0 +1,83 @@
+import UIKit
+
+protocol ReaderDetailView: class {
```
yaelirub, line 3: I could be a bit confused by the naming of this protocol thinking it is a view

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+    func render(_ post: ReaderPost)
+    func showError()
+}
+
+class ReaderDetailWebviewViewController: UIViewController, ReaderDetailView {
+    @IBOutlet weak var webView: WKWebView!
+
+    /// The coordinator, responsible for the logic
+    var coordinator: ReaderDetailCoordinator!
```
yaelirub, line 12: Can this be an optional instead of force unwrapping?

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+
+        coordinator.start()
+    }
+
+    func render(_ post: ReaderPost) {
+        webView.loadHTMLString(post.contentForDisplay(), baseURL: nil)
+    }
+
+    func showError() {
+        // Show error
```
yaelirub, line 25: TODO?

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+        return controller
+    }
+
+    /// A View Controller that displays a Post content.
+    ///
+    /// Use this method to present content for the user.
+    /// - Parameter url: an URL of the post.
+    /// - Returns: A `ReaderDetailWebviewViewController` instance
+    @objc class func controllerWithPostURL(_ url: URL) -> ReaderDetailWebviewViewController {
+        let controller = ReaderDetailWebviewViewController.loadFromStoryboard()
```
yaelirub, line 50: Are we going to use the url here?

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+    /// - Parameter post: a Reader Post
+    /// - Returns: A `ReaderDetailWebviewViewController` instance
+    @objc class func controllerWithPost(_ post: ReaderPost) -> ReaderDetailWebviewViewController {
+        if post.sourceAttributionStyle() == .post &&
+            post.sourceAttribution.postID != nil &&
+            post.sourceAttribution.blogID != nil {
+            return ReaderDetailWebviewViewController.controllerWithPostID(post.sourceAttribution.postID!, siteID: post.sourceAttribution.blogID!)
+        } else if post.isCross() {
+            return ReaderDetailWebviewViewController.controllerWithPostID(post.crossPostMeta.postID, siteID: post.crossPostMeta.siteID)
+        } else {
```
yaelirub, line 67: In what cases will this else happen?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -1290,7 +1290,14 @@ extension ReaderStreamViewController: NewsManagerDelegate {
     }
 
     private func presentReaderDetailViewControllerWithURL(_ url: URL) {
-        let viewController = ReaderDetailViewController.controllerWithPostURL(url)
+        var viewController: UIViewController!
```
yaelirub, line 1293: Can remove the "!" here

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: Based on my tests, most of the time. I'm still not sure what `sourceAttributionStyle` means, but I haven't created this code. I just adapted from here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift#L1551

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: Yes! But this will be covered in the task that handles loading links. :)

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: Yes, this is a `TODO`. But I guess the lint fails if we add `TODO` in the code :P

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: So, my reason behind that is that this VC is useless without the coordinator. This is not intended to be used without a coordinator at all.

If that happens for any reason, I'd prefer the app to crash then to display a white screen.

Wdyt?

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: It is intentional. This is generic: it could be a view, a ViewController, a custom class that represents a view, anything!

I'm also not a big fan of declaring protocols like:

```
protocol ReaderDetailViewController

class ReaderDetailViewControllerImpl: UIViewController, ReaderDetailViewController
```

Let me know if this is reasonable for you. Open to suggestions too!

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: Not sure yet! During the loader task, I'll work on that and I'll figure out. I've supressed this error for now.

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: Removed

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: Added!

leandroalonso:
@yaelirub ready for another look!

I replied to some of your comments, let me know what you think!

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
yaelirub: I was thinking that the "View" here is confusing. Your decision though

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
yaelirub: I never prefer the app crashing!!!! we can show an error when guarding for the coordinator 
Where is the coordinator getting initialized when you're calling `controllerWithPostURL` from `presentReaderDetailViewControllerWithURL' in `ReaderStreamViewController` ?

leandroalonso: Done in 5ac5cb374a957a16b0c271bc42d323e30bbab397

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: We have something similar here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Post/Revisions/ShowRevisionsListManger.swift#L1

Anyway, I can change. I'm just not sure how to call that without reproducing the bad naming convention mentioned above. :(

leandroalonso:
@yaelirub back to you!

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
yaelirub: If you think its good and clear then Im ok with it

PR #14200, opened 2020-05-27
Do not change the date of published posts when changing it to private

### LINKED ISSUE #13996: Editor: Post with private visibility and future publish date is published immediately
### Expected behavior

I expect that when I schedule a post, my selected publish date will be respected.

### Actual behavior

Private posts can't be scheduled, but I'm able to set a future date and set a post to private in the pre-publishing nudge. When I select "Schedule Now," the post is immediately published but has a future publish date (verified by viewing the published post on the blog).

1. Create a new post.
2. Add a title and/or post content.
3. Select Publish.
4. In the pre-publishing nudge, set the publish date to the future.
5. In the pre-publishing nudge, set the visibility to private.
6. Select "Schedule Now."
7. Notice that a "Post published" notice appears and the post is published, but the post has a future date.

Pre-publish nudge settings|Published with future date
-|-
IMG_2386|IMG_2388

leandroalonso:
Fixes #13996

This PR makes sure that published posts' original date isn't changed when changing its status to private.

cc @jd-alexander 

### To test

Change a published post to private and make sure that the original date is not changed.

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostVisibilitySelectorViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostVisibilitySelectorViewController.swift
@@ -35,8 +35,10 @@ import UIKit
             }
 
             if visibility == AbstractPost.privateLabel {
-                // Make sure the post is not scheduled anymore. The user can't schedule a private post
-                post.publishImmediately()
+                if post.isScheduled() {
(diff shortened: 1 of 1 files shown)


jd-alexander:
Nice @leandroalonso 😄

Issue #14202, opened 2020-05-27
Push Notifications Permissions iOS - Master Issue


Gio2018:
Update the existing flow to request permission for push notifications.

- Designs: 
   - First iteration Ref.pbArwn-os-p2
  - Second iteration Ref.pbArwn-zf-p2

- Master Thread Ref.pbArwn-pF-p2

There are three parts to this issue:

1. **Dialog on Landing**
     Instead of showing the custom dialog to request notifications permissions immediately, show it the first time users access the 'Notifications' tab

<img src="https://user-images.githubusercontent.com/34376330/83039181-749cca80-a003-11ea-9bd4-a274b11c0248.png" height="500" />

2. **In-line notifications tab**

- Today: we show an in-line message from your first visit to the notifications tab.

- Change: we will now only display the in-line message after the user has tapped on the notifications tab 5 times.

- Why: Its somewhat obtrusive to dismiss a dialog and then get the same request again, with similar buttons and copy. Especially after the change to request 1, where the user would: tap on the notifications tab, say no to a dialog, then get an in-line request immediately.

- Existing users who have not approved: won’t see this again unless they get a new phone/do a fresh install, and then tap the tab 5 times.

<img src="https://user-images.githubusercontent.com/34376330/83443234-0b57f580-a40f-11ea-9354-7ec80b4b3fe6.png" height="300" />

3. **Return of the dialog**
    The notification permission request dialog would return after users receive 10 notifications, if they have tapped 'not now' but not explicitly granted or denied notifications permissions to the app

![ReturnOfTheDialog](https://user-images.githubusercontent.com/34376330/83040335-c3972f80-a004-11ea-9f05-e4fa95a00a86.png)

- [x] move the logic to show the notification dialog from `BlogDetailsViewController` to `NotificationsViewController`
- [x] add logic to `NotificationsViewController` to display the in-line notifications tab only after users have visited the notification tab five (5) times.
- [x] add logic to `NotificationsViewController` to re-display the dialog after 10 notifications, if users have not explicitly granted or denied notifications permissions
- [x] handle the case of existing users that delete the app, for in-line notifications tab
- [x] update unit tests as needed
- [x] copy reviewed

PR #14203, opened 2020-05-27
Move push notifications permission alert to Notifications tab

Gio2018:
Ref #14202

To test:

- Build and run as fresh install
- Navigate to 'Notifications'
- Verify that the notifications permissions alert shows up
- verify that the buttons are actionable
- verify (on a real device) that tapping on 'allow notifications' presents the system screen to allow notifications

**Notes**
- The in-line prompt logic is yet to be updated. At this time, if users tap on 'not now', the in-line prompt will show up after navigating away and returning to the 'Notifications' tab
- To show the permission alert without deleting the app, pause the execution using the debugger and run the following command
```
e -l swift -- import WordPress; UserDefaults.standard.notificationPrimerAlertWasDisplayed = false
```

This will only work if the notifications permissions have not been set yet (that is, if you tap on 'not now' in the alert)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -1283,7 +1283,7 @@ extension NotificationsViewController: NoResultsViewControllerDelegate {
 //
 internal extension NotificationsViewController {
     func showInlinePrompt() {
-        guard inlinePromptView.alpha != WPAlphaFull else {
+        guard inlinePromptView.alpha != WPAlphaFull, UserDefaults.standard.notificationPrimerAlertWasDisplayed else {
             return
         }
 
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -176,6 +176,9 @@ class NotificationsViewController: UITableViewController, UIViewControllerRestor
             setupAppRatings()
             self.showInlinePrompt()
         }
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
```
frosty, line 179: Can I suggest we move this call into `showNotificationPrimerAlert`, after the `guard`, so we don't need to do this dispatch every time the view appears? If the alert has already been seen we can bail out of that method, otherwise we can dispatch after the timeout.

Perhaps an interim method could be introduced:

```
private func showNotificationPrimerAlertIfNeeded()
        guard !UserDefaults.standard.notificationPrimerAlertWasDisplayed else {
            return
        }

        DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
                self.showNotificationPrimerAlert()
        }
}
```

then just call `showNotificationPrimerAlertIfNeeded` from here.

Gio2018: Yup! This makes perfect sense, thank you!!

PR #14212, opened 2020-05-28
Add support for Iterable deep links

frosty:
This PR adds preliminary support for Iterable deep links. The server side of things isn't set up yet, so we can't actually verify that the deep links work. But we can verify that existing links still work okay.

I've deviated a little from the [Iterable example code](https://support.iterable.com/hc/en-us/articles/115003909366-Handling-iOS-Deep-Links-Without-the-SDK-), as I can't see how thier way of detecting an Iterable deep link would work – they use a regex, which to me looks like it'll basically just match any URL? (`/a/[a-zA-Z0-9]+`) so instead, I'm checking for our Iterable domain.

**To test:**

* Build and run in the simulator
* Background the app
* In the Terminal, run: `xcrun simctl openurl booted https://wordpress.com/notifications` and verify that the app launches to the Notifications screen. If one of our normal deep links loads, the others should all be fine.
* Check the code to see how we'd detect and handle an Iterable deep link.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/System/WordPressAppDelegate.swift
+++ b/WordPress/Classes/System/WordPressAppDelegate.swift
@@ -458,7 +458,9 @@ extension WordPressAppDelegate {
                 return
         }
 
-        UniversalLinkRouter.shared.handle(url: url)
+        trackDeepLink(for: url) { url in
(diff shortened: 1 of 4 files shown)


frosty:
Thanks for the review!

Issue #13864, opened 2020-04-10
Reader: too much space between sections


designsimply:
There are a few areas in the Reader where we can check on block cadence and make it flow a lot more nicely. 

Steps to reproduce:

1. Create a new post.
1. Add several different types of blocks with paragraphs in between.
1. Publish the post.
1. View the post in the Reader or Notifications.

**Result:** too much space between sections and some blocks simply aren't displaying correctly, here are some examples we could look into:

- [ ] Too much space after lists.
- [ ] Too much space between headings and paragraphs.
- [ ] Button doesn't display correctly.
- [ ] Cover block with linear gradient doesn't display correctly.
- [ ] Too much space between video and video caption.
- [ ] Not enough space between caption and the following content, such as a paragraph block.
- [ ] Not enough space between caption and separator.
- [ ] Too much space between separator and the next content, such as an image block.

<img src="https://user-images.githubusercontent.com/1119271/78957404-7980db80-7aa2-11ea-892b-0a601d4b6ec4.PNG" width="150" alt="IMG_3823"> <img src="https://user-images.githubusercontent.com/1119271/78957409-7dacf900-7aa2-11ea-8ede-ca9d7a4a1d69.PNG" width="150" alt="IMG_3824"> <img src="https://user-images.githubusercontent.com/1119271/78957410-7e458f80-7aa2-11ea-9892-2cd8f7790e91.PNG" width="150" alt="IMG_3825"> <img src="https://user-images.githubusercontent.com/1119271/78957412-7ede2600-7aa2-11ea-828d-094ae7fe7ceb.PNG" width="150" alt="IMG_3826">

<p style="clear:both;"><sup>Tested with WPiOS 14.6.0.0 TestFlight beta on iPhone 11 iOS 13.4.1.</sup></p>

<sub>(internal reference: p4a5px-2wF-p2 h/t @elibud)</sub>

elibud:
cc: @yaelirub @bummytime as we'll want to tackle these fixes as part of the Reader improvements project. Thanks @designsimply .

designsimply:
While testing the Custom HTML block today, I found that the following HTML is nice to use for testing spacing issues: https://raw.githubusercontent.com/cbracco/html5-test-page/master/index.html

More testing steps:

1. Go to https://raw.githubusercontent.com/cbracco/html5-test-page/master/index.html
1. Copy the content inside the `body` tags.
1. Create a new post in the app or on the web.
[...]

Issue #13862, opened 2020-04-10
Reader: quote block citation spacing is wrong


designsimply:
Steps to reproduce:

1. Create and publish a new post with a quote block and a citation.
1. View the post in the Reader or Notifications.

**Result:** the quote doesn’t look good and it’s missing some space between the quote and the post body.

Block Editor in app | Reader in app | Webview in app | Mobile Web
--- | --- | --- | ---
<img src="https://user-images.githubusercontent.com/1119271/78952833-0f613a00-7a94-11ea-9748-523fb8a70c3a.PNG" width="150" alt="IMG_3810"> | <img src="https://user-images.githubusercontent.com/1119271/78953050-d4133b00-7a94-11ea-9b8e-dd156464fabc.PNG" width="150" alt="IMG_3808"> | <img src="https://user-images.githubusercontent.com/1119271/78952902-4df6f480-7a94-11ea-9afa-88b1383e8915.PNG" width="150" alt="IMG_3812"> | <img src="https://user-images.githubusercontent.com/1119271/78953089-eb522880-7a94-11ea-8999-14fb36f03a3d.PNG" width="150" alt="IMG_3813">

<p style="clear:both;"><sup>Tested with WPiOS 14.6.0.0 TestFlight beta on iPhone 11 iOS 13.4.</sup></p>

<sub>(internal reference: p4a5px-2wF-p2 h/t @elibud)</sub>

Issue #9525, opened 2018-06-05
Blog post: bullet points are sometimes blue


ScoutHarris:
### Expected behavior
All bullet points should be black.

### Actual behavior
If the line starts with a URL the bullet point is blue (the URL color).

### Steps to reproduce the behavior
- Go to Notifications, select a Notification, then select a Post with a list.
- Notice the items that start with  URL have a blue bullet point, while the others are black. In a browser, all are black. (example post: p77Llu-apb-p2)

![list_bullets](https://user-images.githubusercontent.com/1816888/40995758-44d4ec9e-68bd-11e8-97bd-9fa75131a8d1.png)

rachelmcr:
This came up again in an internal report (p4a5px-2v3-p2).

designsimply:
This came up again in another internal report, and this should be prioritized in upcoming Reader improvements—we need to make sure the vast majority of user created content is properly displayed in the apps.

<p style="clear:both;"><sub>(internal reference: p4a5px-2wF-p2 h/t @elibud)</sub></p>

PR #14213, opened 2020-05-28
Apply post styles into webview

### LINKED ISSUE #14210: Adds CSS styles to Reader webview

leandroalonso:
Fixes #14210

(Also address #13862, #9525 and #13864)

This PR renders the post content into a webview applying the correct styles.

The styles were extracted from Calypso with a few modifications.

Noticed that header and footer are going to be tackled in the following task.

### To test

(To render the post into a webview remember to enable the feature flag in app settings)

My recommendation to test this PR is to keep one device (or simulator) with the current in-production Reader and the other with this branch. Then, open a post on both and compare — the post should appear **similar** on both. In this phase I'm looking for very outstanding bugs (instead of fixing everything).

Besides that, please:

1. Test with iPhone and iPad (note that in iPad we're using readable margins)
2. Test light and dark mode
3. Test portrait and landscape

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
```diff
@@ -0,0 +1,96 @@
...
+    /// Loads a HTML content into the webview and apply styles
+    ///
+    @discardableResult
+    override func loadHTMLString(_ string: String, baseURL: URL?) -> WKNavigation? {
+        let content = """
+        <!DOCTYPE html><html><head><meta charset='UTF-8' />
+        <title>Reader Post</title>
+        <meta name='viewport' content='initial-scale=1, maximum-scale=1.0, user-scalable=no'>
+        <link rel="stylesheet" type="text/css"
+              href="https://wordpress.com/calypso/fallback/async-load-blocks-reader-full-post.feb2901cfbd98579fa81.min.css">
```
leandroalonso, line 23: In the final version, we won't have this link hardcoded here, even because that might change.

We just need to discover a better way to refer to this.

yaelirub:
styling for lists and the 3 mentioned issues looks better. 
A few notes:
1. The dividers and quote lines are a bit weak 
Simulator Screen Shot - iPhone 11 Pro - 2020-05-29 at 13 41 43
Simulator Screen Shot - iPhone 11 Pro - 2020-05-29 at 13 42 48

2. When tapping on links it takes some time to load and for a second or two I wasn't sure if the tap was recognized.

3. In some post there seem to be a missing space at the end 
Simulator Screen Shot - iPhone 8 - 2020-05-29 at 13 47 57
Simulator Screen Shot - iPhone 11 Pro - 2020-05-29 at 13 47 20

*** its worth mentioning that this PR does not handle media, headers, footers or links an @s

leandroalonso:
> On iPad it didn't open a webview but used the previous rendering

@yaelirub Did you override the feature flag? (Profile/Me -> App Settings -> Debug -> Toggle on "Reader content displayed in a WebView")

yaelirub:
@leandroalonso  shoot I forgot about that 🤦

leandroalonso:
@yaelirub about your comments:

> The dividers and quote lines are a bit weak

I changed the divider to always be displayed with 1px height. It was configured to be displayed with 0.5 in retina displays, but I removed that.

The quotes are actually larger than they were before (9px, before it was 6px). Maybe it's the different color that is giving you this impression?

> When tapping on links it takes some time to load and for a second or two I wasn't sure if the tap was recognized.

This will be soon addressed in a separate task!

> In some post there seem to be a missing space at the end

I'll also tackle that on the task that adds the toolbar. Is that ok?

jkmassel:
Howdy friends – we're freezing `15.0` today, so I'm bumping this PR forward to `15.1`. If it needs to land in the `15.0` series, please merge it to `release/15.0` and DM me – I'll be happy to create a new beta release!

yaelirub:
@leandroalonso , since some issues are going to be handled in following PRs I'm going to approve this one.

PR #14226, opened 2020-06-01
Add tag should be gray and disabled when there isn't text inside

### LINKED ISSUE #14119: [GlobalStep] [iOS] "Add Tag" button should appear disabled/greyed out when the tag field is empty.
**Environment:**
iPhone 6(v12.4.4), iPhone 11 Pro (v13.3.1).

**Reproduction Rate:** 
4/4 100%
 
**Steps To Reproduce:**
 1. Install and open the application.
 2. log in with valid credentials.
 3. Go to Reader
 4. Tap the cog in the top corner
 5. Tap the 'Add a Topic' option
 4. Note that there is an empty field and "Add Tag" button.
 5. Now tap on "Add Tag" button without entering any tag
 6. Observe that "Add Tag" button fails to disable and user navigates back to Reader tab upon tapping "Add Tag" button.
  
**Actual Result:**
[GlobalStep] [iOS] "Add Tag" button should appear disabled/greyed out when the tag field is empty.
 
**Expected Result:**
 "Add Tag" button should remain disabled when the tag field is empty or a validation "Please enter a tag" should be displayed upon tapping Add tag button.

[#2229.zip](https://github.com/wordpress-mobile/gutenberg-mobile/files/4588273/2229.zip)

ciccioska:
Fixes #14119 
With this change, the ActionButton used to add a tag is disabled until at list a char is inserted. 
To test:

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
+++ b/WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
@@ -123,6 +123,8 @@ typedef NS_ENUM(NSInteger, SettingsTextSections) {
 {
     [super viewDidAppear:animated];
     [self.textField becomeFirstResponder];
+    // Fire initial status
+    [self validateTextInput:_textField];
 }
 
 - (void)viewWillDisappear:(BOOL)animated
@@ -172,22 +174,52 @@ typedef NS_ENUM(NSInteger, SettingsTextSections) {
 {
     // Hook up to Change Events
     [_textField addTarget:self action:@selector(validateTextInput:) forControlEvents:UIControlEventEditingChanged];
-    
-    // Fire initial status
-    [self validateTextInput:_textField];
 }
 
 - (BOOL)textPassesValidation
 {
-    BOOL isEmail = (self.mode == SettingsTextModesEmail);
-    return (self.validatesInput == false || isEmail == false || (isEmail && self.textField.text.isValidEmail));
+    switch(self.mode){
+        case SettingsTextModesEmail:
+            return !self.validatesInput || self.textField.text.isValidEmail;
+            break;
+        case SettingsTextModesLowerCaseText:
+        case SettingsTextModesText:
+            return !self.validatesInput || [self isValidText:self.textField.text];
+            break;
+        case SettingsTextModesURL:
+        case SettingsTextModesPassword:
+        case SettingsTextModesNewPassword:
+        default:
+            return (!self.validatesInput);
+            break;
+    }
+}
+//TODO: This method should be stay in the WorpressShared project
+- (BOOL)isValidText:(NSString*)text
+{
+    NSString *regex = @"\\w+";
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"SELF MATCHES %@" argumentArray:@[regex]];
+
+    return [predicate evaluateWithObject:text];
 }
 
 - (void)validateTextInput:(id)sender
 {
     self.doneButtonEnabled = [self textPassesValidation];
+    [self evaluateActionButton];
 }
 
+- (void)evaluateActionButton
+{
+    BOOL isValid = self.textPassesValidation;
+    if (isValid){
+        [_actionCell setUserInteractionEnabled: isValid];
+        _actionCell.textLabel.textColor = UIColor.murielText;
+    }else{
+        [_actionCell setUserInteractionEnabled: isValid];
+        _actionCell.textLabel.textColor = UIColor.murielTextSubtle;
+    }
+}
 
 #pragma mark - Properties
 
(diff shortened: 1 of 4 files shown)


ciccioska:
Hi, there is any problem with this feature, who can help me? Thanks

bjtitus:
Thanks for the contribution @ciccioska! This looks great!

One small thing I noticed is that the regex pattern for a word doesn't allow multi-word tags. Multi-word tags should be separated by hyphens:

`street-photography`: https://wordpress.com/tag/street-photography

I think it probably makes sense to just check that the text is not empty for now. If we want to do something fancier for the data type then we probably need to provide that to the `SettingsTextViewController` since the pattern for a tag may be different than the pattern for another settings text item.

WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
```diff
@@ -123,6 +123,8 @@ - (void)viewDidAppear:(BOOL)animated
 {
     [super viewDidAppear:animated];
     [self.textField becomeFirstResponder];
+    // Fire initial status
+    [self validateTextInput:_textField];
```
bjtitus, line 127: It's probably better to use the property here instead of the instance variable in case we needed to change the getter.

```suggestion
    [self validateTextInput:self.textField];
```

WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
```diff
@@ -172,22 +174,52 @@ - (void)startListeningTextfieldChanges
...
-    
-    // Fire initial status
-    [self validateTextInput:_textField];
 }
 
 - (BOOL)textPassesValidation
 {
-    BOOL isEmail = (self.mode == SettingsTextModesEmail);
-    return (self.validatesInput == false || isEmail == false || (isEmail && self.textField.text.isValidEmail));
+    switch(self.mode){
```
bjtitus, line 181: Just a minor spacing adjustment here to match the rest of the file.

```suggestion
    switch (self.mode) {
```

WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
```diff
@@ -172,22 +174,52 @@ - (void)startListeningTextfieldChanges
...
+        case SettingsTextModesURL:
+        case SettingsTextModesPassword:
+        case SettingsTextModesNewPassword:
+        default:
+            return (!self.validatesInput);
+            break;
+    }
+}
+//TODO: This method should be stay in the WorpressShared project
+- (BOOL)isValidText:(NSString*)text
```
bjtitus, line 198: We have a method in called [`slugForTopicName`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/ReaderTopicServiceRemote.m#L364) which contains some of the parsing done before sending this request. It could be a good idea to reuse that here for validation, but I think it's fine to just check the length for now.

WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
```diff
@@ -172,22 +174,52 @@ - (void)startListeningTextfieldChanges
...
 - (void)validateTextInput:(id)sender
 {
     self.doneButtonEnabled = [self textPassesValidation];
+    [self evaluateActionButton];
 }
 
+- (void)evaluateActionButton
+{
+    BOOL isValid = self.textPassesValidation;
+    if (isValid){
```
bjtitus, line 215: Small spacing tweak to be consistent with the file.

```suggestion
    if (isValid) {
```

WordPress/Classes/ViewRelated/Tools/SettingsTextViewController.m
```diff
@@ -172,22 +174,52 @@ - (void)startListeningTextfieldChanges
...
+        [_actionCell setUserInteractionEnabled: isValid];
+        _actionCell.textLabel.textColor = UIColor.murielText;
+    }else{
```
bjtitus, line 218: Small spacing tweak to be consistent with the file.

```suggestion
    } else {
```

bjtitus (changes requested):

ciccioska (commented):
Hi, I made all the changes suggested.

bjtitus (approved):
Looks good to me! I'll be gone over the weekend but we'll need to update with develop and merge from there.  

:shipit:

ciccioska:
Hi, this PR has not been merged, how can I help to close it? thanks

aerych:
Hi @ciccioska 👋  
Based on the earlier comment `but we'll need to update with develop and merge from there` it sounds like this branch was needing to have the `develop` branch merged into it (maybe to resolve a conflict?) before the changes could be merged to `develop`. 

The branch is currently showing conflicts with the `develop` branch, specifically `WordPress/WordPress.xcodeproj/project.pbxproj`.  This will need to be resolved before the PR can be merged.

If you would, `git pull` the latest changes in `develop` to your local branch and resolve any conflicts. Then we can take a look at merging. :)

aerych:
I think this might have been abandoned by the author.  Closing.

PR #14229, opened 2020-06-01
Push Notifications Permissions - In-line Prompt After 5 Visits

Gio2018:
Ref #14202 14202

To test:

- perform a fresh install and login or signup
- go to 'Notifications' tab
- When the alert shows, tab 'Not Now'
- Go to another tab and back to 'Notifications' four more times
- Verify that the in-line prompt shows up

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -170,6 +170,10 @@ class NotificationsViewController: UITableViewController, UIViewControllerRestor
 
         markWelcomeNotificationAsSeenIfNeeded()
 
+        if userDefaults.notificationsTabAccessCount < Constants.inlineTabAccessCount {
+            userDefaults.notificationsTabAccessCount += 1
+        }
+
         if shouldShowPrimeForPush {
             setupNotificationPrompt()
         } else if AppRatingUtility.shared.shouldPromptForAppReview(section: InlinePrompt.section) {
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1673,4 +1680,9 @@ extension NotificationsViewController: UIViewControllerTransitioningDelegate {
             self?.tabBarController?.present(alert, animated: true)
         }
     }
+
+    private enum Constants {
+        static let inlineTabAccessCount = 5
```
frosty, line 1685: I think we might want to bump this up by 1, as I think it should appear on the 5th appearance of the tab after dismissing the original alert. It currently appears on the 4th. (or change our logic to use `<=` and `>` when checking. That might make more sense.

Gio2018:
> Looking good! I just posted one code-level comment.
> 
> One other note: the tab access count (and I guess the primer alert) survives a log out event. I wonder if we should clear these on log out, and when somebody logs in we can re-show them again (providing that permission wasn't already granted).


Good point. 
I was wondering what are the use cases when users log out and back in? If they had a reason to do it frequently, we might want to persist the status, otherwise we should probably clear it. If the reason is switching between accounts, then it would likely make sense to clear it. @osullivanchris what's your opinion on this?

frosty:
I think clearing it should be fine. If the user already approved notifications, they wouldn't see it again anyway.

osullivanchris:
Sorry I just want to clarify something. Are we only talking about the case where the user has not taken action yet?

I ask because if the user has tapped **not now** I don't think they should see the in-line alert again if they log out and log back in. For someone who does it frequently that would be annoying.

However if we're just talking about users who have not chosen an option yet,  I don't mind whether we persist or reset the status of that alert. I think whatever makes most sense technically. If the user has tapped on the tab 3 times, and it resets to 0, or stays at 3; it makes little difference.

frosty:
> I ask because if the user has tapped not now I don't think they should see the in-line alert again if they log out and log back in. For someone who does it frequently that would be annoying.

Hmm, I guess this makes sense! So the setting should really be per device, not per user (which is what this PR does already).

Gio2018:
> Hmm, I guess this makes sense! So the setting should really be per device, not per user (which is what this PR does already).

Looks like we all agree on this, so I'm going to merge this one then. In case we have more thoughts about logout/login behavior, I'll address them in the next PRs

Issue #14009, opened 2020-04-28
UL&S: Master Issue


mindgraffiti:
The purpose of this ticket is to track all WPiOS PRs related to Unified Login & Signup. The early PRs are listed here. The later PRs are auto-linked (below this summary).

### Spikes
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13669

### To Production
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13703
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13704
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13720
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13730
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13733
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13786
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13827
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13828
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13896
- https://github.com/wordpress-mobile/WordPress-iOS/pull/13904
- https://github.com/wordpress-mobile/WordPress-iOS/pull/14003

mindgraffiti:
This became unwieldy very fast. Here's a search filter for each of the host apps.

- [Newspack-iOS PRs](https://github.com/Automattic/newspack-ios/pulls?q=is%3Apr+is%3Aclosed+label%3AAuthentication+)
- [WooCommerce-iOS PRs](https://github.com/woocommerce/woocommerce-ios/pulls?q=is%3Apr+is%3Aclosed+label%3ALogin+label%3A%22%5BType%5D+Enhancement%22)
- [WordPress-iOS PRs](https://github.com/wordpress-mobile/WordPress-iOS/pulls?q=is%3Apr+is%3Aclosed+project%3Awordpress-mobile%2F60)

U&LS project board: https://github.com/orgs/wordpress-mobile/projects/60

PR #14230, opened 2020-06-01
UL&S: Add tableView and continue button to SiteAddressViewController

mindgraffiti:
Ref. #14009 
Ref. https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/294

This PR points to the latest Auth feature branch and also creates a second instance of `WordPressAuthenticatorStyle` for use with unified auth changes.

### To test
1. Enable the .unifiedSiteAddress feature flag, `return BuildConfiguration.current == .localDeveloper`
2. `rake dependencies`
3. Build and run
4. Log In > "entering your site address" 
**Expected:** a Continue button appears. The tableview isn't visible at this point because it has no content.

<a href="https://user-images.githubusercontent.com/1062444/83460189-6f88b280-a42b-11ea-88e1-e6cee06985de.png"></a>

PR submission checklist:


WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
```diff
@@ -41,32 +41,61 @@ class WordPressAuthenticationManager: NSObject {
...
+                                                       disabledBorderColor: .neutral(.shade10),
+                                                       primaryTitleColor: .white,
+                                                       secondaryTitleColor: .text,
+                                                       disabledTitleColor: .neutral(.shade20),
+                                                       textButtonColor: .primary,
+                                                       textButtonHighlightColor: .primaryDark,
+                                                       instructionColor: .text,
+                                                       subheadlineColor: .textSubtle,
+                                                       placeholderColor: .textPlaceholder,
+                                                       viewControllerBackgroundColor: .white,
```
Gio2018, line 89: Since the only difference between the two styles is 
`viewControllerBackgroundColor: .listBackground` vs
`viewControllerBackgroundColor: .white`
would it be possible to use the ternary condition in the initializer instead of initializing `WordPressAuthenticatorStyle` twice?

mindgraffiti: I like it! It's a smaller footprint, especially when I don't know how many colors I will need to change.

PR #14233, opened 2020-06-02
Add toolbar in the new Reader detail screen

### LINKED ISSUE #14223: Add toolbar to the new reader detail screen
Add toolbar and action bar buttons: save for later, comments, star, and reblog.

leandroalonso:
Fixes #14223

Screen Shot 2020-06-03 at 18 26 06

This PR extracts and adds the toolbar in the new Reader detail screen.

* I extracted the toolbar to its own View so it could be easily reused in the future.
* I haven't extracted logic to another layer. I really don't like to have service and database in a View, but I'm afraid that refactoring those might increase the project scope. I'm leaving as it is for now.

### To test

#### iPad and iPhone, light and dark mode

1. Open a post in the reader
2. Checks the toolbar appears correctly in:

* [x] iPhone in light mode
* [x] iPhone in dark mode
* [x] iPad in light mode
* [x] iPad in dark mode

#### Actions

1. Open a post in the reader
2. Then:

* [x] Save the post for later
* [x] Like/dislike a post
* [x] Tap the comment icon to see comments
* [x] Tap reblog to reblog a post

PR submission checklist:


leandroalonso:
@yaelirub ready for review! :)

I created a small checklist so it can help you with your testing.

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
```diff
@@ -0,0 +1,323 @@
...
+
+        // Show likes if logged in, or if likes exist, but not if external
+        if (ReaderHelpers.isLoggedIn() || post.likeCount.intValue > 0) && !post.isExternal {
+            configureLikeActionButton()
+        }
+
+        // Show comments if logged in and comments are enabled, or if comments exist.
+        // But only if it is from wpcom (jetpack and external is not yet supported).
+        // Nesting this conditional cos it seems clearer that way
+        if (post.isWPCom || post.isJetpack) && !shouldHideComments {
```
yaelirub, line 138: The only place we set `shouldHideComments` is in https://github.com/wordpress-mobile/WordPress-iOS/blob/0bdc8da476a29b1574d17e6a027783ffa7e46d40/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m#L1182 but it checks the variable setting in `ReaderDetailViewController`

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
```diff
@@ -0,0 +1,323 @@
...
+    @IBAction func didTapReblog(_ sender: Any) {
+        guard let post = post, let viewController = viewController else {
+            return
+        }
+
+        ReaderReblogAction().execute(readerPost: post, origin: viewController, reblogSource: .detail)
+    }
+
+    @IBAction func didTapComment(_ sender: Any) {
+        guard let post = self.post, let viewController = viewController else {
```
yaelirub, line 82: can remove `self.` here

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
```diff
@@ -0,0 +1,323 @@
...
+    // MARK: - Configuration
+
+    private func configureActionButtons() {
+        resetActionButton(likeButton)
+        resetActionButton(commentButton)
+        resetActionButton(saveForLaterButton)
+        resetActionButton(reblogButton)
+
+        guard let post = post else {
+            assertionFailure()
```
yaelirub, line 126: why do we have assertion failure here but just `return` in other guards for post ?

yaelirub:
In iPad dark mode the post content color is not right, making it looks like content is missing :
Screen Shot 2020-06-05 at 5 02 51 PM

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
leandroalonso: This is for the future. :)

In a subsequent branch I added that: https://github.com/wordpress-mobile/WordPress-iOS/blob/issue/14252_new_reader_detail_missing_features/WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift#L58

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
leandroalonso: Done in 3249f7eded6c8ac62b92f76b4223d5f268b9c0d9

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
leandroalonso: Probably different people writing code?

To keep consistency I removed this in 9c1775a272640aa992fe4b7a4cabbdccc7a3dcba

leandroalonso:
@yaelirub The problem you're seeing in dark mode is because the URL of the CSS has expired (it changes on every Calypso deploy).

I can change that but it will probably expire in a few hours. I'm about to open a PR that will create a static URL that will not expire. :)

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderDetailToolbar.swift
yaelirub: And you'll set it to true where needed as well?

leandroalonso: Yes! Also here: https://github.com/wordpress-mobile/WordPress-iOS/blob/6f792020947308b8053672488eff1a91c1aaa141/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m#L1183 :)

PR #14235, opened 2020-06-02
ULS: Use unified Google auth flows

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/282
Auth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/295

To test:

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -184,10 +184,9 @@ target 'WordPress' do
 
     pod 'Gridicons', '~> 1.0.1'
 
-
-    pod 'WordPressAuthenticator', '~> 1.18.0-beta.2'
+    # pod 'WordPressAuthenticator', '~> 1.18.0-beta.4'
     # While in PR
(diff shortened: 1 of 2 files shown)


ScoutHarris:
> I'm unable to verify sign up with Google using a new Google account that hasn't been previously connected to a WP.com account.

Gah! I forgot to mention. If you use `Sign up` from the `Connected But...` view, that will attempt to create a normal WP account, not a Google connected account. While you would have signed up, it would technically be the same as "sign up with email", not "sign up with Google" (because they use different endpoints). 

I'm currently working on this view, specific to Google signup, that will be shown when there is no existing WP account (i.e. login is redirected to signup) . This will create a Google connected WP account.

<kbd>Screen : Google : Sign up confirmation</kbd>

ScoutHarris:
Thanks @mindgraffiti  & @Gio2018 !

PR #14238, opened 2020-06-03
Restore support for animated GIFs in Gutenberg.

SergioEstevao:
When we implemented our custom ImageLoader for RN, to support private sites, we accidentally removed support for playing animated GIFs while inside GB editor.
This PR changes the loading system of images to use the `RCTAnimated` class when a GIF is being downloaded.

One of the side effects of that change is that if the post featured image on the post/page is an animated GIF we show the animation.

Fixes #

To test:
 - Start the app
 - Create a new Post using GB
 - Add an image block and configure it to use an animated GIF
 - See if you can see the GIF animation on the editor
 - Publish the post
 - Go to the list of posts
 - See if the post card in the list of posts shows the animated GIF
 - Test this with WP.com sites, WP.org sites, Atomic sites, and WP.com private sites.

PR submission checklist:


WordPress/Classes/Utility/Media/ImageDownloader.swift
```diff
@@ -53,13 +53,21 @@ class ImageDownloader {
     @discardableResult
     func downloadImage(for request: URLRequest, completion: @escaping (UIImage?, Error?) -> Void) -> ImageDownloaderTask {
         let task = session.dataTask(with: request) { (data, _, error) in
-            guard let data = data, let image = UIImage(data: data) else {
-                let error = error ?? ImageDownloaderError.failed
```
etoledom, line 57: Aren't we loosing the chance to pass a better error?

WordPress/Classes/Utility/Media/ImageDownloader.swift
```diff
@@ -53,13 +53,21 @@ class ImageDownloader {
...
+            var image: UIImage?
+            if let url = request.url, url.pathExtension.lowercased() == "gif" {
+              image = RCTAnimatedImage(data: data, scale: 1)
+            } else {
+              image = UIImage.init(data: data)
+            }
+            if let finalImage = image {
+              completion(finalImage, nil)
+            } else {
+              completion(nil, ImageDownloaderError.failed)
```
etoledom, line 69: On my tests this works great!
But I was thinking if we could make it a bit more readable. What about something like:

```swift
if let gif = self.gifImage(from: request, data: data) {
    completion(gif, nil)
} else if let image = UIImage(data: data) {
    completion(image, nil)
} else {
    completion(nil, ImageDownloaderError.failed)
}
```

In this way we know right away what kind of image are we passing to the completion handler.

SergioEstevao: Thanks for the suggestion @etoledom . I updated the code according to your suggestions, do you mind to give it another look?

WordPress/Classes/Utility/Media/ImageDownloader.swift
SergioEstevao: Good catch!

PR #14239, opened 2020-06-03
Fix post_install hook to use the project directory when patching a ReactNative file

pinarol:
Fixes #

We saw that the patch introduced with [this PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/14062/files) was not getting applied. It manifested itself in the BottomSheet, mover buttons aren't disabled properly.

Reality:

Screen Shot 2020-06-03 at 10 14 59

Expected:
Screen Shot 2020-06-03 at 10 14 59

**Root Cause**

```
[12:56:13]: ▸ sh: patches/react-native+0.61.5.patch: No such file or directory
[12:56:13]: ▸ Patching RCTActionSheet to add possibility to disable action sheet buttons -
[12:56:13]: ▸ it could be removed once PR with that functionality will be merged into RN
[12:56:13]: ▸ sh: patches/react-native+0.61.5.patch: No such file or directory
```

It turns out fastlane builds are run from the WordPress/Scripts folder, so `patch` doesn't find the file since there's no dir as `Scripts/patches`. 

(fastlane  scripts that we use to build the releases live in the Scripts folder)

To fix this, we are updating the paths to use project_root.

To test:

- Clear Pods folder
- cd to `Script` folder and run `bundle exec pod install`
- Check the logs to see that no errors are produced during patching
- Open the patched file `/Users/pinarolguc/Development/WordPress-iOS/Pods/React-RCTActionSheet/RCTActionSheetManager.m` make sure this line exists:

`NSArray<NSNumber *> *disabledButtonIndices = [RCTConvert NSArray:options[@"disabledButtonIndices"]];`

- clear Pods folder
- run `rake dependencies` in the root folder(WordPress-iOS)
- Check the logs again

bonus: It'd be good if we can open the app and see the mover buttons getting disabled as well

PR submission checklist:


pinarol:
Thank you @loremattei Opened another [issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/14240) to tackle that.

PR #14249, opened 2020-06-05
Revert post status setting when Prepublishing bottom sheet dismissed

### LINKED ISSUE #13995: Editor: Setting post to private makes it stuck in a publish loop
Posts set with private visibility are tricky because private posts can't be saved as a draft or scheduled; they have to be published immediately.

### Expected behavior

I expect clear messaging around why a private post must be published immediately, with help to change the post's visibility if I want to save/schedule it instead of publishing it now.

On the web, this is handled by immediately publishing the post when private post visibility is selected. A dialog appears to confirm the publish action:

Screenshot 2020-04-24 17 38 39

If I choose "Cancel" in that dialog, the post visibility is not changed to private.

### Actual behavior

In the app, once I change the post visibility to private (in the post settings or in the pre-publish nudge introduced in 14.7), I get "stuck" in a publish loop. It isn't obvious how to save the post as a draft or why I'm being forced to publish:

When I try to exit the editor, I have the options to update the post or discard my changes. If I've made changes to the draft that I want to save, for example, I'll choose update — but that reopens the pre-publish nudge rather than saving the draft.

I have to manually change the post visibility in post settings or select "Save as Draft" in the ellipsis menu to resolve it so I can save it as a draft (which automatically reverts the post visibility change). This isn't obvious if you don't understand the restrictions on private posts.

These steps are to reproduce the issue in v14.7+ (with the pre-publishing nudge) although the issue also exists in previous versions:

1. Go to the My site tab > Blog posts > + to start a new post.
2. Add a title and/or post content.
3. Open the ellipsis menu and tap "Publish now."
4. Tap on "Visibility."
5. Select "Private."
6. Dismiss the bottom sheet.
7. Tap "X" to close the editor.
8. Notice the "You have unsaved changes" prompt appears. Select "Update Post" to save the changes to the draft.

Result: The pre-publishing nudge appears again.

< v14.7 (publish confirmation) | 14.7+ (pre-publish nudge)
-|-
IMG_2384|IMG_2385

##### Tested on [iPhone XS], iOS [13.4.1], WPiOS [14.5 & 14.7.0.20200421]

Originally raised in https://github.com/wordpress-mobile/WordPress-iOS/issues/13985#issuecomment-618891951.

yaelirub:
Fixes #13995 
Related PR: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/69

To test:
1. start a new post.
2. Add a title and/or post content.
3. Tap publish
4. Tap on "Visibility."
5. Select "Private."
6. Dismiss the bottom sheet.
7. Tap "X" to close the editor.
8. Notice the "You have unsaved changes" prompt appears with an option to save draft

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -38,7 +38,7 @@ def wordpress_ui
     ## for development:
     #pod 'WordPressUI', :path => '../WordPressUI-iOS'
     ## while PR is in review:
-    pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :branch => 'task/bottomsheet_child_handle_dismiss`'
+    pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :branch => 'task/bottomsheet_child_handle_dismiss'
     # pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :commit => '85b2a8cfb9d3c27194a14bdcb3c8391e13fbaa0f'
 end
 
--- a/Podfile
+++ b/Podfile
@@ -34,11 +34,11 @@ end
 
 def wordpress_ui
     ## for production:
-    pod 'WordPressUI', '~> 1.7.0'
+    #pod 'WordPressUI', '~> 1.7.0'
     ## for development:
     #pod 'WordPressUI', :path => '../WordPressUI-iOS'
     ## while PR is in review:
-    #pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :branch => 'feature/bottom-sheet-pan-gesture'
+    pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :branch => 'task/bottomsheet_child_handle_dismiss`'
     # pod 'WordPressUI', :git => 'https://github.com/wordpress-mobile/WordPressUI-iOS', :commit => '85b2a8cfb9d3c27194a14bdcb3c8391e13fbaa0f'
 end
 
--- a/Podfile
+++ b/Podfile
@@ -34,11 +34,11 @@ end
 
 def wordpress_ui
     ## for production:
-    #pod 'WordPressUI', '~> 1.7.0'
+    pod 'WordPressUI', '~> 1.7.1'
     ## for development:
     #pod 'WordPressUI', :path => '../WordPressUI-iOS'
     ## while PR is in review:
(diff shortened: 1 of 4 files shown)


leandroalonso:
@yaelirub I did a few changes: https://gist.github.com/leandroalonso/00ca15bc5851515e3c3cc218ee5b7657

* We only revert the post status if it was changed to private
* We check if the publish button wasn't pressed

There's an additional issue though:

1. Open the nudges
2. Change visiblity to password protected
3. Dismiss the nudge
4. Open it again, set it to private
5. Dismiss the nudge
6. Open it again: password lost

To fix that we would need to keep a copy of the post when the Nudges is created for the first time. But, to be honest, I'm not sure that's an issue.

elibud:
@leandroalonso if we only lose the password when the user changes the post status or doesn't publish I think it's a fair compromise to make.

leandroalonso:
@elibud Just committed those changes (hope you don't mind @yaelirub).

Can you test again?

elibud:
This is working as expected for me and the code looks good, but  given how rusty I am let's get another pair or eyes to approve before merging.

rachelmcr:
I tested this change, and functionally it's working well for me in numerous scenarios including:

* Change the visibility to `private` in either the post settings or pre-publishing nudge, and then either publish the post privately or dismiss the nudge and save as a (public) draft.
* Set visibility to `password protected` and have that saved when I dismiss the pre-publishing nudge. (I agree it's ok if the password is lost in the scenario described above.)
* Set publish date and have that saved when I dismiss the pre-publishing nudge.
* Set tags and have them saved when I dismiss the pre-publishing nudge, regardless of what the visibliity is set to.

My only concern is that because we reset the status to `public` without notifying the user explicitly, they will need to notice and set it to to `private` again when they are ready to publish. That should be ok because they'll see the pre-publishing nudge again, but if they don't notice that it was reset it could result in accidentally publishing a post as `public` when it wasn't intended.

(I'm also seeing some other unexpected behavior when I set posts to `private`, but it seems unrelated to this change. I'll file those issues separately.)

leandroalonso:
@rachelmcr another option is to not reset the status, and then show "Publish Post" instead of "Update Post" in the Action Sheet. Tapping it would open the Prepublishing Nudges.

The problem that I see with that is that the user might be confused and will not be able to leave the editor...

rachelmcr:
Yeah, I think that's basically the same problem as the original issue, where you're stuck in a loop where you have to publish to leave the editor. This feels like a definite improvement on that. 👍 

Unless a designer has a suggestion for a better way to communicate what's happening here to the user, I'm ok with giving this a try.

PR #14266, opened 2020-06-08
Quick Start: Dismiss Notices using clearWithTag

### LINKED ISSUE #11394: Quick Start: Dismiss Notices using clearWithTag
In #11281, we added the ability to dismiss Notices using tags:

```swift
ActionDispatcher.dispatch(NoticeAction.clearWithTag(noticeTag))
```

Using `NoticeAction.clearWithTag` is preferred to `NoticeAction.dismiss`. The `.clearWithTag` helps in avoiding dismissing Notices that the current page _did not create_. 

In #11281, we only changed `QuickStartTourGuide` to use `.dismiss`. This can be improved to use `.clearWithTag`.

mindgraffiti:
Fixes #11394 

This PR uses the recommended `.clearWithTag` behavior instead of `.dismiss` on the Quick Start tour guide.

### To test:
Walk through the "Customize Your Site" tour and "Grow Your Audience" tour. Ensure there are no regressions to the toasts and how they behave.

**Known issues:** the iA work has changed where our navigation is located, so now some steps are marked incorrectly in the Quick Start guide. See issues #14267 and #14268.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -7,6 +7,7 @@ open class QuickStartTourGuide: NSObject {
     private var currentTourState: TourState?
     private var suggestionWorkItem: DispatchWorkItem?
     private weak var recentlyTouredBlog: Blog?
+    private let noticeTag: Notice.Tag = "QuickStartTour"
     static let notificationElementKey = "QuickStartElementKey"
 
     @objc static func find() -> QuickStartTourGuide? {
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
```diff
@@ -389,7 +391,7 @@ private extension QuickStartTourGuide {
     }
 
     func dismissCurrentNotice() {
-        ActionDispatcher.dispatch(NoticeAction.dismiss)
+        ActionDispatcher.dispatch(NoticeAction.clearWithTag(noticeTag))
         ActionDispatcher.dispatch(NoticeAction.empty)
```
shiki, line 395: I'm curious whether we still need this. 🤔 But I think we should leave it for now cause it's not a good idea to cause a regression. It doesn't look like `QuickStartTourGuide` has unit tests so that makes things more difficult.

mindgraffiti: I don't know this feature deeply enough to make that call. It might be superfluous. I'll add a comment before merging.

mindgraffiti:
@shiki thanks for the review!

PR #14271, opened 2020-06-08
Handle links in the new Reader detail

### LINKED ISSUE #14228: Handle opening links and media in the new reader detail

leandroalonso:
Fixes #14228

This PR handles links in a Reader post.

### To test

There are 4 types of links handled:

* Images (gif, jpg, jpeg or png)
* Internal routes
* Link to a WP.com post
* All the other cases

To help testing it, I created this post: https://charlietest12345.blog/2020/06/08/handling-links/

#### Images

1. Open the post above in the Reader
2. Tap the first image, it should load in a new screen
3. Tap the GIF
4. The gif will be loaded and shown in full screen

#### Internal routes

1. Open the post above in the Reader
2. Tap /read/a8c
3. Check that the app is now showing the A8C session

#### WP.com link

1. Open the post above in the Reader
2. Tap the link "notes"
3. Check that it opens a new Reader detail screen with the contents of the post

#### Other links

1. Open the post above in the Reader
2. Tap the link to WP.com
3. Check that it opens in a new screen

Feel free to create other posts and try different links. :) 

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -124,6 +145,26 @@ class ReaderDetailCoordinator {
         })
     }
 
+    /// Requests a ReaderPost from the service and updates the View.
+    ///
+    /// Use this method to fetch a ReaderPost from a URL.
+    /// - Parameter url: a post URL
+    private func fetch(_ url: URL) {
+        service.fetchPost(at: postURL,
```
yaelirub, line 153: shouldn't this be ` service.fetchPost(at: url,` ?

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -124,6 +145,26 @@ class ReaderDetailCoordinator {
...
+                          success: { [weak self] post in
+                            guard let post = post else {
```
yaelirub, line 155: Why do we need this guard here? 
self?.post is optional, there's a guard in renderPostAndBumpStats and it could be `self?.view?.show(title: post?.postTitle)`

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -124,6 +145,26 @@ class ReaderDetailCoordinator {
...
+                                return
+                            }
+
+                            self?.post = post
+                            self?.renderPostAndBumpStats()
+                            self?.view?.show(title: post.postTitle)
+        }, failure: { [weak self] error in
+            DDLogError("Error fetching post for detail: \(String(describing: error?.localizedDescription))")
+            self?.view?.showError()
```
yaelirub, line 164: Shouldn't we pass the error to showError ?

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -240,3 +253,18 @@ extension ReaderDetailWebviewViewController: UIViewControllerTransitioningDelega
...
+
+// MARK: - Navigation Delegate
+
+extension ReaderDetailWebviewViewController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        if navigationAction.navigationType == .linkActivated {
+            if let url = navigationAction.request.url {
+                coordinator?.handle(url)
+            }
+            decisionHandler(.cancel)
```
yaelirub, line 265: Why .cancel ?

yaelirub:
everything works as expected in the test post
but when I tap on the images in other articles like : https://wordpress.com/blog/2020/06/04/premium-content-and-paid-newsletters/
nothing happens (it works when I disable the reader webview flag in App settings)

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: Fixed in 285e1372d90ff096426537c6470256d72147cc1a

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: Changed in 0c01b80b1702ad75995e008bc63241b40c0757a4

Thanks for pointing that out!

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: This is out of the scope for this task.

But even in the subsequent, we don't make distinction on what error happened. Any error will always display the same message, same for the current reader (you can see [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/ReaderDetailViewController.swift#L419)).

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
leandroalonso: `.cancel` tells the WebView to not open this link. Since we handle all the links, we don't let the WebView open then.

leandroalonso:
@yaelirub that happens because those images don't link to their media files, that's why it wasn't working.

However, to keep consistency with the current Reader, I implemented that in 3eec68afa2f2a822942a5aafa373475596cb564c

Can you please try again? Thank you!

PR #14283, opened 2020-06-10
Show loader indicator and errors on the new Reader detail

### LINKED ISSUE #14237: Add loader indicator and errors to the new Reader detail screen

leandroalonso:
Fixes #14237

This PR adds a loader indicator that is shown while the content hasn't been retrieved and rendered. It also shows a message in case of a failure or if the user if without connectivity.

It also:

* Improves WebView rendering time: we make sure to load the CSS and the text before everything, just after that step the other things are loaded. This results in the content appearing very quickly so the user can start reading right away.
* Removes `autoplay` from audio and videos. The current `WKWebView` plays that automatically, which is very annoying.
* Points the CSS to https://wordpress.com/calypso/reader-mobile.css (per https://github.com/Automattic/wp-calypso/pull/43073)

| Loading  | Error | Connectivity |
| ------------- | ------------- | ------------- |
|  |  |  |
 
### To test

Please, test the screens above in:

* iPhone: light and dark mode
* iPad: light and dark mode

To generate the error screen you can try to access a post which you don't have permission or either change `renderPostAndBumpStats`:

```swift
    private func renderPostAndBumpStats() {
        view?.showError()
    }
```

PR submission checklist:


jkmassel:
Howdy friends – we're freezing `15.1` today, so I'm bumping this PR forward to `15.2`. If it needs to land in the `15.1` series, please merge it to `release/15.1` and DM me – I'll be happy to create a new beta release!

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
```diff
@@ -19,25 +22,56 @@ class ReaderWebView: WKWebView {
         <!DOCTYPE html><html><head><meta charset='UTF-8' />
         <title>Reader Post</title>
         <meta name='viewport' content='initial-scale=1, maximum-scale=1.0, user-scalable=no'>
-        <link rel="stylesheet" type="text/css"
-              href="https://wordpress.com/calypso/fallback/async-load-blocks-reader-full-post.feb2901cfbd98579fa81.min.css">
+        <link rel="stylesheet" type="text/css" href="https://wordpress.com/calypso/reader-mobile.css">
```
emilylaguna, line 25: Thoughts on adding a some parameters to the URL?

Something like `platform=iOS`, `app-version=XX`, `os-version=XX`?

That way in the future if there's a bug with a specific version we can remotely fix it without having to do an update. 

A current example of this is the `<audio>` tag doesn't work on in the reader on iOS 11.4.

WordPress/Classes/ViewRelated/Reader/ReaderDetailWebviewViewController.swift
```diff
@@ -268,3 +312,44 @@ extension ReaderDetailWebviewViewController: WKNavigationDelegate {
         }
     }
 }
+
+// MARK: - Error View Handling (NoResultsViewController)
+
+private extension ReaderDetailWebviewViewController {
+    func configureAndDisplayLoadingView(title: String, accessoryView: UIView? = nil) {
```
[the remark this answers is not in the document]
leandroalonso: Done in 6608017bf279327336cca9cf337b3abda7d58396

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
```diff
@@ -4,6 +4,9 @@ import UIKit
 ///
 class ReaderWebView: WKWebView {
 
+    /// HTML elements that will load only after the text has appeared
+    private let elements = ["audio", "embed", "iframe", "img", "input", "script", "source", "track", "video"]
```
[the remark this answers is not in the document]
leandroalonso: HTML elements that has `src`: https://www.w3schools.com/tags/att_src.asp

Most of the posts will contain `img` or `audio` or `video`. But I decided to be extra cautious here and added the others.

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
leandroalonso: We can do that. But I have a few concerns:

* This will change the CSS endpoint for every version we release. This is not a problem, but if a user updated the app and then later try to open an article while offline the CSS will be broken (because this file has never be loaded)
* We don't have the logic in the backend to deal with that parameters.
* I'd avoid adding conditionals and keeping the CSS as simple as possible. But who knows about the future?

Let me know what do you think. :)

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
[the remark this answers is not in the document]
emilylaguna: Ah makes sense. Maybe add a comment above it with a link to that for future reference?

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
emilylaguna: > This will change the CSS endpoint for every version we release. This is not a problem, but if a user updated the app and then later try to open an article while offline the CSS will be broken (because this file has never be loaded)

True. I think this may expose a deeper issue though. If someone upgrades to the new WP version but doesn't open it until they're offline, what happens for them? Is it worth considering adding a local fallback CSS file just in case?

We could bundle a version of the CSS in the app and have a CSS rule above the web version that points to it. When the web version is loaded it will override the local CSS properties. 

`<link rel="stylesheet" type="text/css" href="file:///Applications/WordPress.app/Resources/reader-mobile.css">` 

> We don't have the logic in the backend to deal with that parameters.

Currently it doesn't but the backend can be updated in the future if needed. 

> I'd avoid adding conditionals and keeping the CSS as simple as possible. But who knows about the future?

I agree, but speaking from experience we are at the whim of what WebKit supports in each OS version. There could be a bug that occurs only on iOS 12 when rendering a specific block.

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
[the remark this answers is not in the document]
leandroalonso: Done in ffee56f468052a695aa2737e770a44f2c4ee75da

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
leandroalonso: > True. I think this may expose a deeper issue though. If someone upgrades to the new WP version but doesn't open it until they're offline, what happens for them? Is it worth considering adding a local fallback CSS file just in case?

Offline support is yet to come, but if a user saved a post for later we will preload the whole post in a hidden webview (same strategy as Android uses).

Yes, if someone just upgraded, has saved articles, and tries to open a saved post while offline the CSS won't be available.

> Is it worth considering adding a local fallback CSS file just in case?
Maybe... but this has its own challenges:

* How to deal with the style conflicts? This is not a "fallback" but something that will always be loaded
* We would keep that updated over time?

> There could be a bug that occurs only on iOS 12 when rendering a specific block.

I agree. But how would that be different from fixing in the browser?

emilylaguna: I agree that this has its own technical challenges and is outside the scope of this PR. We can continue discussing this separately if needed.

PR #14288, opened 2020-06-10
Track site changes inside the editor

### LINKED ISSUE #13041: Track event for the site switcher within the editor
In the iOS app, it's possible to switch sites from inside the editor. We would like to know how often that is used so we can use that information in deciding whether to add that feature to the Android app as well.

IMG_3179 IMG_3180

<sub>(internal reference: p4k3M4-2BP-p2#comment-2106)</sub>

elibud:
Fixes #13041

To test:

For both editors, for pages and posts, check that when switching the site the entry belongs to the new event is fired, and that is not fired when canceling the site switching.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -32,6 +32,7 @@ import Foundation
     case editorPostLocationChanged
     case editorPostSlugChanged
     case editorPostExcerptChanged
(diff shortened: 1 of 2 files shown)


elibud:
Thanks for the review @etoledom. We really just want to track this event to gather feature usage, so I kept it minimal. @maxme let me know if you'd like me to add more info to the event or this is good to go as it is.

maxme:
> @maxme let me know if you'd like me to add more info to the event or this is good to go as it is.

PR #14298, opened 2020-06-11
Send highest priority plan in Zendesk Tickets

### LINKED ISSUE #14224: Add Zendesk plan field
Copying this 👇 from [the Android issue](https://github.com/wordpress-mobile/WordPress-Android/issues/12064).

---

We are currently sending the site plans as tags and these are converted to a single Plan field on the Zendesk side. When the user has more than one site, each one with different plans, the last unique plan will be selected. This is not the behavior we want. Instead, we want the Plan field to show the "highest" plan the user has. Here are the changes we need to make to accomplish that:

1. [x] Remove the plan tags [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/support/ZendeskHelper.kt#L433-L435). This is so that when we send the plan field Zendesk doesn't try to override it. Even if that wasn't an issue, when multiple plan tags are sent, there are some cases where some plans are removed, so it's just simpler to not send them.
2. [x] Add a new ticket field id [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/support/ZendeskHelper.kt#L491) which I believe is `25175963`. Please verify this value with HEs if the plan you're sending is not showing up.
3. [x] Add the "highest" plan as a custom field with the id from Step 2 [here](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/support/ZendeskHelper.kt#L369-L378). You can use the `allSites` parameter in the function to figure that out. The [`shortPlanName`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java#L92) property is localized, so it's probably best to not rely on that. Instead, we can hopefully use the [`planId`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/model/SiteModel.java#L91) property to add a hard-coded matching. If that doesn't work, we have [`FETCH_PLANS`](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/action/SiteAction.java#L81) action to fetch the plan details for a site, but I suggest avoiding that because Zendesk config build is probably not a great time to do multiple fetch requests. (due to multiple sites) Instead, we might want to add a new property to `SiteModel` and set it [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/develop/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/site/SiteRestClient.java#L1024-L1035). (This is only if we can't use the `planId` property)

Note that there is a little bit of research involved in this issue. Specifically, we'll need to figure out what the "highest" plan is and that'll depend on which property we can utilize. 
- [x] Regardless, my suggestion is to first get the connection between app and Zendesk working and verify that you are able to set the plan to whatever you want through a hardcoded value. Once that's working, all we'll need to do is to find the highest plan which we can probably extract somewhere and unit test for future-proofing and possible maintenance. (in case we need to modify it)

Here are the Plan field values I can see in Zendesk:
```
Free, Blogger, Personal, Premium, Business/Professional, eCommerce, Tier 1, Tier 2, Tier 3, Tier 4, Other, Add-On
```
I think some of these will not be applicable to us, so I wouldn't worry about using all of them. I also don't know for sure if these are the values we need to send.

---

Hope these details will be useful to whoever is working on this. Please don't hesitate to ping me if you have any questions. Also please consider testing the integration very early to verify that you're setup correctly just so you don't waste time thinking something doesn't work when in fact the issue might be elsewhere. (speaking from experience :) ) Finally, please remember to use an email that is not your Zendesk agent email for your Zendesk identity. (which will be set the first time you visit Help through a dialog) If you use your Zendesk agent email, the integration will not work.

---
- [x] there seems to be an issue with the Business/Professional plan: Zendesk wants the value `business_professional`, while we are sending `Business`. To be investigated
  - there is a WP.com Automation that will assign the Business/Professional plan if there is a `Business` tag. to be determined what to do with the custom field.

Gio2018:
Fixes #14224

Related WordPressKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/250

Previously, if users had more than one site/plan, Zendesk could assign the wrong plan.
What we want, is to always set the 'highest' plan in the ticket, regardless of the site.

The priority should be:
1. eCommerce
2. Business/Professional
3. Premium
4. Personal
5. Blogger
6. Free

More context can be found here:
- p8wKgj-1eQ-p2
- p8wKgj-1gx-p2

#### Notes
1. Uses the /plans/mobile endpoint to retrieve the plan priority, in order to be 'future proof' if the list of available plans changes

2. I am now populating `sitePlanCache` even in the case of English language users. This is because, if a user adds/removes a plan, the `allBlogs` collection might not be updated (it gets updated only with login/logout or visiting `BlogListViewController`, that contains the `syncBlogs()` method), so for these users, the list might not be updated just yet, when they request a ticket. It's likely an edge case, but we are still doing that request for non English users anyway

3. In the event that there are unrecognized plans (new plans that have been added before we could update the app), these will be sent to Zendesk as tags in the ticket, in the form of
  "unrecognized_plan_[plan_name]"

To test:

Prerequisites:
- a Zendesk account
- Another WordPress.com account, different from the previous one, that has a variety of plans (or just add remove plans on the store admin)

1. Log in with the second account
2. Open one or more requests on Zendesk and verify that the assigned plan priority list matches the above one.
3. Optionally, hardcode a list of random strings in `plans` in the method `getHighestPriorityPlans()` of `ZenDeskUtils` and verify that the unrecognized plans are sent to Zendesk as ticket tags.

PR submission checklist:

- [x] I have considered adding unit tests where possible. - Added
- [x] ~~I have considered adding accessibility improvements for my changes.~~ No impact on the UI
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ No user facing changes


WordPress/Classes/Utility/ZendeskUtils.swift
```diff
@@ -188,24 +190,74 @@ extension NSNotification.Name {
...
+        ticketFields.append(CustomField(fieldId: TicketFieldIDs.plan, value: getHighestPriorityPlan()))
+        requestConfig.customFields = ticketFields
+
+        // Set tags
+        var tags = ZendeskUtils.getTags()
+        // send unrecognized plans as tags to Zendesk
+        if let plans = unrecognizedPlans {
+            tags.append(contentsOf: plans.map { "unrecognized_plan_" + $0 })
+        }
+        requestConfig.tags = tags
```
ScoutHarris, line 253: Can this be done in `getTags` so that all tag stuff is done in the same place?

WordPress/Classes/Utility/ZendeskUtils.swift
```diff
@@ -188,24 +190,74 @@ extension NSNotification.Name {
...
-        let planService = PlanService(managedObjectContext: context)
+        let planService = planService ?? PlanService(managedObjectContext: context)
         planService.getAllSitesNonLocalizedPlanDescriptionsForAccount(account, success: { plans in
             self.sitePlansCache = plans
         }, failure: { error in })
     }
 
+    /// Retrieves the highest priority plan, if it exists
+    /// - Returns: the highest priority plan found, or an empty string if none was foundq
+    private func getHighestPriorityPlan() -> String {
```
ScoutHarris, line 208: Can this be a static method to follow the pattern of the other methods? Also, there is a private extension this might could be moved into.

aerych:
Been chatting with @Gio2018 about this and we think we have a way to make this future proof so if the list of plans ever change it wouldn't require an update to the app as well.   We'll make an update to the API endpoint (D44839-code) and let the plans endpoint specify the priority for each plan.

WordPress/Classes/Utility/ZendeskUtils.swift
```diff
@@ -956,13 +1017,14 @@ private extension ZendeskUtils {
         static let logFieldCharacterLimit = 50000
         static let sourcePlatform = "mobile_-_ios"
         static let gutenbergIsDefault = "mobile_gutenberg_is_default"
+        static let planPriority = ["ecommerce", "business_professional", "premium", "personal", "blogger", "free"]
```
aerych, line 1020: Maybe this is no longer needed?

WordPress/Classes/Utility/ZendeskUtils.swift
```diff
@@ -54,6 +54,8 @@ extension NSNotification.Name {
     private static var zdClientId: String?
     private weak var presentInController: UIViewController?
 
+    private var unrecognizedPlans: Set<String>?
```
aerych, line 57: Maybe no longer needed?

Gio2018:
> Code wise I just had a couple of questions. I'm also wondering if it would be a good idea to give the new fields default values to be sure that the lightweight migration to the new model version works as we expect:

Thank you @aerych ! I think this would make perfect sense! Just pushed the requested changes!

PR #14299, opened 2020-06-11
Fix 'Take a Photo' option failing after adding an image to gallery block

ceyhun:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/1965

Could potentially fix https://sentry.io/organizations/a8c/issues/1535683817/?project=1438083

To test:

1. Add a gallery block
1. Add an image using "Choose from device"
1. Tap the "+" button at the bottom of the gallery block
1. Add an image using the "Take a Photo" option
1. Newly taken photo should be added to the gallery block
1. Now add an image block
1. Try adding an image to it using the "Take a Photo" option
1. Newly taken photo should be added to the image block

PR submission checklist:


guarani:
I'm not sure why this crash isn't closing the app in production, it must be caught somewhere 🤔.

ceyhun:
>This crash only happens when you first pick a photo from the library—which initializes navigationPicker—and then you take a photo, which causes navigationPicker to be accessed.

@guarani I couldn't reproduce the crash with debugger attached. Do you mean it happens selecting first from "Choose from device" or "Free Photo Library"? Both worked fine for me.

<p align="center">

</p>

> I think there's also a (minor) bug in WPMediaPicker because it probably shouldn't be calling its delegate method emptyViewController(forMediaPickerController:) when a photo is taken since an empty state isn't applicable when taking a photo.

For me this method is not called when a photo is taken via "Take a Photo":
https://github.com/wordpress-mobile/WordPress-iOS/blob/issue/1965-take-a-photo-failure/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift#L137

Could it be an iOS or Xcode version difference? I'm using iOS 13.5.1 and Xcode 11.5?

Can you please try again maybe with the build [in this comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/14299#issuecomment-642868503)? It also worked fine for me.

guarani:
> Do you mean it happens selecting first from "Choose from device" or "Free Photo Library"? Both worked fine for me.

Hey @ceyhun, it happened to me when selecting "Choose from device".

> For me this method is not called when a photo is taken via "Take a Photo":

It's strange this method is being called on my device but not on yours. 

> Could it be an iOS or Xcode version difference? I'm using iOS 13.5.1 and Xcode 11.5?
I was also on Xcode 11.5, but my device was on iOS 13.4.1.

> Can you please try again maybe with the build in this comment? It also worked fine for me.

I should have mentioned I tested build 29996 yesterday. Although I'll try the new 30055 now.

--- 

Just tested 30055 and it's crashing (closing) the app for me.
I can also see the crash 100% of the time on the the current App Store version 14.9, as well as the soon to be released 15.0. 

It happens when using just the gallery block (first photo from device gallery, second photo captured by camera), or when using two image blocks (again, first from device gallery, second from camera).

I'm going to update my device to the latest iOS and try again.

guarani:
I've updated to 13.5.1 and still see the crash, both with the current App Store version and build 30055. I tried to follow the exact sequence of steps you performed in the [gif above](https://github.com/wordpress-mobile/WordPress-iOS/pull/14299#issuecomment-643381134) but  the second image in the gallery never gets added for me (or crashes the app if I'm on build 30055 instead of the App Store version).

If we were to try to implement a fix at the moment, I think changing `unowned` to `weak` is a safe move:
https://github.com/wordpress-mobile/WordPress-iOS/blob/5721e1496799ab9b0130aea0766ad63da1b136e3/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift#L19
because this will allow `navigationPicker` to become `nil` and no longer crash the app when accessed from `emptyViewController(forMediaPickerController:)`. We could take this one step further and leave it as a plain optional (no weak or unowned). (`WPNavigationMediaPickerViewController` is used in a bunch of places in the app and in none of them is it `unowned` or `weak`.)

I know this must be awkward to debug since it's not crashing on your end, but I thought I'd mention this fix since 15.1 is getting cut on Monday.

guarani:
@SergioEstevao – do you know why this is a weak reference (it crashes the gallery block on some devices but not others)?

https://github.com/wordpress-mobile/WordPress-iOS/blob/5721e1496799ab9b0130aea0766ad63da1b136e3/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift#L19

SergioEstevao:
> @SergioEstevao – do you know why this is a weak reference (it crashes the gallery block on some devices but not others)?

I'm assuming it was to avoid a retain cycle, but is this reference being used after the picker is dismissed? I normally prefer to use a weak references to avoid unexpected crashes when accessing it references after releases.

guarani:
> I'm assuming it was to avoid a retain cycle, but is this reference being used after the picker is dismissed?

@SergioEstevao yes, this crashes on some devices because it gets accessed after being deallocated. 
The scenario is: when choosing a photo from the gallery, `navigationPicker` is only retained by the window it's presented on. Then when taking a photo, it's accessed again but is now deallocated.

ceyhun:
> If we were to try to implement a fix at the moment, I think changing `unowned` to `weak` is a safe move:

@guarani I agree with that. Just made this change. But this still makes me think what could be the reason that make this crash for you and not me 🤔

jkmassel:
Howdy friends – we're freezing `15.1` today, so I'm bumping this PR forward to `15.2`. If it needs to land in the `15.1` series, please merge it to `release/15.1` and DM me – I'll be happy to create a new beta release!

ceyhun:
@guarani Thanks for testing again 🙇 Good to hear that it's fixed now! The crash you had could be this one: https://sentry.io/organizations/a8c/issues/1535683817/?project=1438083
Since it's not affecting many users (including me :)), I'm going to target the next release for this and merge to develop.

PR #14301, opened 2020-06-11
Remove empty box flashes

### LINKED ISSUE #14080: [GlobalStep] [iOS] Under "Menus" black empty box flashes for few seconds before "Nothing Found" message is prompted while adding new item.
Environment:
iOS: iPad Air 2(v13.3.1), iPhone 11 Pro (v13.3.1)

Steps To Reproduce:
 
 1. Install and launch the application.
 2. Tap on "Log In".
 3. Log in to the app using valid credentials.
 4. Open "My site" tab.
 5. Tap on "Menus".
 6. Select or add new menu.
 7. Tap on New item and list will open.
 8. Tap on any option from the list (eg. Page,Link, Tag, Post,Product)
 9. Observe that   black empty text box is displayed for few second and then "Nothing found" message is displayed.
  
 Actual Result:
Under "Menus" black empty box flashes for few seconds before "Nothing Found" message is prompted while adding new item

 Expected Result:
Upon adding new item in Menu from "Menus" blank empty box should not display.

[#14080.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4594795/14080.zip)

negna1:
Fixes #14080 

To test:
1. Install and launch the application.
2. Tap on "Log In".
3. Log in to the app using valid credentials.
4. Open "My site" tab.
5. Tap on "Menus".
6. Select or add new menu.
7. Tap on New item and list will open.
8. Tap on any option from the list (eg. Page,Link, Tag, Post,Product)
9. Should't be white box in light theme case and dark box in dark theme case.

PR submission checklist:


frosty (changes requested):
Thanks for looking into this issue!

This change does remove the label, however it seems like there's a lot of code in this view class that's there specifically to make that label pulse as a loading animation. I think it would be best if we remove that animation code, and instead add a `UIActivityIndicatorView` centered within the view that we start and stop instead. Is that something you could look into doing? I don't think it would be a big change, but let me know if you need any pointers.

Also, please could you ensure that your branch is up to date with the latest code from the `develop` branch? I see a lot of extra commits here that don't belong in this PR. Thank you!

negna1:
> Thanks for looking into this issue!
> 
> This change does remove the label, however it seems like there's a lot of code in this view class that's there specifically to make that label pulse as a loading animation. I think it would be best if we remove that animation code, and instead add a `UIActivityIndicatorView` centered within the view that we start and stop instead. Is that something you could look into doing? I don't think it would be a big change, but let me know if you need any pointers.
> 
> Also, please could you ensure that your branch is up to date with the latest code from the `develop` branch? I see a lot of extra commits here that don't belong in this PR. Thank you!

If i get it wright , I must add indicator instead of this label animation. I will do that.

etoledom:
Closing this PR in favour of https://github.com/wordpress-mobile/WordPress-iOS/pull/14312

PR #14303, opened 2020-06-12
Remove news card feature

### LINKED ISSUE #14204: Remove NewsCard
As we have the new feature announcement and the new designs for Reader, please remove the News card feature

negna1:
Fixes #14204

To test: 
1. log in wordpress
2. Tap to reader tab bar
3. there must be any header which tells 

PR submission checklist:


frosty (changes requested):
Thanks for working on this issue!

Please can you ensure that your branch is up to date with the latest code from `develop`, and only contains commits relevant to the changes here? As well as a lot of merge commits, I also see this one: `remove empty box flash on menu item`, which I think is from one of your other PRs.

Also, please could you remove the code you've commented out completely? We generally don't like to leave commented out code in the project, and it can always be restored from git if we needed it in the future. I noticed you removed a number of files from the Xcode project but left the actual files present on disk. We can also remove the `News.strings` file, which contains the actual news items.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -483,56 +483,19 @@ import WordPressFlux
 
     // MARK: - Configuration / Topic Presentation
 
-    @objc private func configureStreamHeader() {
```
frosty, line 486: This method shouldn't be removed, as it is required to display the header view for many parts of the Reader.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -483,56 +483,19 @@ import WordPressFlux
...
     // topic's following status changed.
+    
+    //remove news card . as I see ther is no need to be thisfunction because there must not be any header , but it is possible to add some
+    //header , so i only delete configureStreamHeader()
     private func refreshTableHeaderIfNeeded() {
         guard let _ = readerTopic else {
             return
         }
-        configureStreamHeader()
+       // configureStreamHeader()
```
frosty, line 495: We still need to call this method here to provide a header to various views in the Reader.

WordPress/Classes/ViewRelated/System/WPTabBarController.m
```diff
@@ -120,16 +120,8 @@ - (instancetype)init
                                                  selector:@selector(updateIconIndicators:)
                                                      name:NSNotification.ZendeskPushNotificationClearedNotification
                                                    object:nil];
+//remove news card notifications
```
frosty, line 123: There's no need to include a comment here :)

frosty (changes requested):

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
negna1: Okay , I see. Nothing was using this method and that's why i deleted it. I will try to fix what you  say.

twstokes:

PR #14314, opened 2020-06-15
Remove ReaderPost content format from the Service

### LINKED ISSUE #14313: Remove the ReaderPost content format from the Service and move to the view
Since we're changing the Reader Detail view to use a WebView, the [content sanitization](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/develop/WordPressShared/Core/Utility/RichContentFormatter.swift) in WordPressKit is making some of the content to be displayed incorrectly (like Gutenberg Galleries).

The idea is to:

* Stop saving the content sanitized in the database
* Sanitize the content before displaying the old Reader Detail

leandroalonso:
Fixes #14313

`WordPressKit` related PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/251

This PR removes the `RichContentFormatter` from the Reader's service and move it to `ReaderDetailViewController`.

This change is necessary because the formatting is stripping some HTML tags that are making the post to render incorrectly in the new ReaderDetail that uses a WebView.

We lose a little bit of performance for doing that way, but soon the rendering with a WebView will be available and that will not be an issue.

### To test

1. Open posts in the Reader
2. Make sure they appear the same way as it was before


leandroalonso:
@aerych Thanks for the suggestion! Done in 524b2de7b6908883348541cd8ce8a456220e2f9e

So much cleaner. :)

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -1227,6 +1227,12 @@ - (ReaderPost *)createOrReplaceFromRemotePost:(RemoteReaderPost *)remotePost for
         post.sourceAttribution = nil;
     }
 
+    if ([Feature enabled:FeatureFlagReaderWebview]) {
+        post.content = remotePost.content;
+    } else {
+        post.content = [RichContentFormatter formatContentString:remotePost.content isPrivateSite:remotePost.isBlogPrivate];
+    }
+
```
aerych, line 1235: Looks good, but I think we'd want to either remove the earlier assignment on line 1156, or we'd want to replace it with this snippet.

leandroalonso: Fixed in 0d64166f0ee91b53c503227ee4f3f58ce812f064

PR #14330, opened 2020-06-16
Fixes Web Preview bug

### LINKED ISSUE #13437: Preview: Tapping preview twice in a row causes errors
### Expected behavior

I expect that if I'm able to tap on elements on the screen, even while waiting for another action to complete, I won't cause problems.

### Actual behavior

If I select "Preview" twice in a row, it causes two problems:

* A "Preview Unavailable" screen appears.
* The editor no longer works as expected.

1. Create a new post and add some content.
2. Select "Preview" from the ellipsis menu.
3. Immediately (while the status bar still shows "Generating preview") select "Preview" from the ellipsis menu again.
4. After the preview window appears, close it and notice that a screen is visible that says "Preview Unavailable."
5. Go back and notice that the editor no longer works as expected (post settings don't work and can cause a crash, new content can be lost, etc - some similar results as reported in https://github.com/wordpress-mobile/WordPress-iOS/issues/13436)

bjtitus:
Fixes #13437 

This fixes an issue where the web preview logic to upload a post could occur multiple times while the post is being uploaded to the server. It checks whether the post is currently being pushed before trying again and showing the web preview.

### Testing
- Create a new post
- Make a few changes to the content
- Throttle connection using Network Link Conditioner (or other tool). These were the settings I used:
Screen Shot 2020-06-16 at 8 34 09 AM

- Go to "Web Preview"
- While the Preview is being generated, tap "Web Preview" again and do this 2-3 times.
- Before the PR this should produce a Preview Unavailable message. Afterward, this should produce an alert about the post being uploaded.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+MoreOptions.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+MoreOptions.swift
@@ -70,6 +70,11 @@ extension PostEditor where Self: UIViewController {
             return
         }
 
+        guard post.remoteStatus != .pushing else {
+            displayPostIsUploadingAlert()
+            return
+        }
+
         savePostBeforePreview() { [weak self] previewURLString, error in
             guard let self = self else {
                 return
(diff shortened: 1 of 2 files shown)


bjtitus:
Hmmm. Somehow I missed a bullet 🤦 

This is the expected behavior, though. Thanks!

PR #14338, opened 2020-06-17
Fix word count to ignore HTML tags/comments when doing analytics.

### LINKED ISSUE #13618: Word Count is including HTML tags
It looks like the code to count words is not stripping HTML tags, which means it will overcount words on Gutenberg posts.

To reproduce:

1. Watch the logs for the `editor_post_published` event, it should have a `word_count` property.
2. Set your editor to classic and publish a post with just the word `Hello`: word count shows 1
3. Set your editor to block and  publish a post with just the word `Hello`: word count shows 6

SergioEstevao:
Fixes #13618 

Added a new method to the PostEditor protocol to allow the editor
send the information about content metrics.
This allows them to use a more optimized way to calculate metrics
and match with the way they display content information in the editor.

@koke I think you may enjoy reviewing this one :)

I also have a question for you I notice that on `word_diff_count` prop for analytics we are sending the original number of words in the post, is this correct or should we send the actual difference of numbers words between the original post and new post content?

**Update (22-06-2020)**: After some discussion offline we decided to remove the `word_diff_count` property altogether.
The rationale is that we are not using it for any stats, and we also don't implement it in Android and web. 

To test:
 - Start the app
 - Edit a post
 - Add a breakpoint on this method: https://github.com/wordpress-mobile/WordPress-iOS/compare/issue/13618_fix_word_count?expand=1#diff-ae677e43073df51ab2a9be66a47d74b0R189
 - Check that when you publish a post the metrics are correct.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -182,17 +182,8 @@ extension PostEditor where Self: UIViewController {
             return
         }
 
-
-        var originalWordCount = post.original?.content?.wordCount() ?? 0
-        var wordCount = post.content?.wordCount() ?? 0
-        if let (original, current) = self.getContentWordMetrics() {
-            originalWordCount = original
-            wordCount = current
-        }
+        let wordCount = self.wordCount ?? 0        
         var properties: [String: Any] = ["word_count": wordCount, WPAppAnalyticsKeyEditorSource: analyticsEditorSource]
-        if post.hasRemote() {
-            properties["word_diff_count"] = originalWordCount
-        }
 
         properties[WPAppAnalyticsKeyPostType] = postTypeValue
 
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -182,8 +182,13 @@ extension PostEditor where Self: UIViewController {
             return
         }
 
-        let originalWordCount = post.original?.content?.wordCount() ?? 0
-        let wordCount = post.content?.wordCount() ?? 0
+
+        var originalWordCount = post.original?.content?.wordCount() ?? 0
+        var wordCount = post.content?.wordCount() ?? 0
+        if let (original, current) = self.getContentWordMetrics() {
+            originalWordCount = original
+            wordCount = current
+        }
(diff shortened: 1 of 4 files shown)


maxme:
Can we check we get consistent results with the Android implementation? If results are not consistent, please open a wpandroid ticket to change the existing way of counting words:

https://github.com/wordpress-mobile/WordPress-Android/blob/78cfb8812c6fb0edb608bee18cfe59ea64cb47f1/WordPress/src/main/java/org/wordpress/android/util/analytics/AnalyticsUtils.java#L163-L167

maxme:
Related ticket in wpandroid: https://github.com/wordpress-mobile/WordPress-Android/issues/11342

WordPress/Classes/ViewRelated/Post/PostEditor.swift
```diff
@@ -111,6 +111,9 @@ protocol PostEditor: class, UIViewControllerTransitioningDelegate {
     var autosaver: Autosaver { get set }
     /// true if the post is the result of a reblog
     var postIsReblogged: Bool { get set }
+
+    /// Returns the word counts of the content in the editor.
+    var wordCount: UInt? { get }
```
etoledom, line 116: Would it make sense to have this non-optional and default to zero in Gutenberg-side implementation?

In this case I'm not sure what would be the difference between `nil` and 0, and seems like needed just because of the special case of how we get the count on Gutenberg.

etoledom, line 116: Would it make sense to make this non-optional?

I'm not sure how to interpret the difference between 0 and nil. It seems more like this is needed because of how we count words in Gutenberg. Maybe it makes more sense to default to 0 in the gutenberg side implementation instead?

SergioEstevao: The idea was that nil is used when we don't know the count yet. When that happens we are sending zero anyway in stats so I agree with your suggestion and just return zero on that cases.

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -182,7 +182,7 @@ extension PostEditor where Self: UIViewController {
             return
         }
 
-        let wordCount = self.wordCount ?? 0
+        let wordCount = self.wordCount
```
etoledom, line 185: Thanks for the update!

I think this line is now not needed 🤔

PR #14339, opened 2020-06-17
Save for later: basic offline support

### LINKED ISSUE #14337: Add basic offline support in save for later

leandroalonso:
Fixes #14337

Review only after #14328 is merged

This PR adds basic support to offline when saving a post for later in the Reader.

Why *basic* support? What we do is:

1. We open a hidden webview to preload the post, so when offline must of the resources (mainly images) will be there
2. Why not all of them? Because the WebView respects the cache from the server. If an image comes from a server that disables caching its resources, this will not work (but it will work with WP.com)

### To test

1. Open the Reader
2. Save a few posts for later
3. Close the app
4. Go offline
5. Open the app
6. Open the saved posts, most of the images should be there

You can use this snippet to make sure the webview content is erased:

```swift
let websiteDataTypes = NSSet(array: [WKWebsiteDataTypeDiskCache, WKWebsiteDataTypeMemoryCache])
let date = Date(timeIntervalSince1970: 0)
WKWebsiteDataStore.default().removeData(ofTypes: websiteDataTypes as! Set<String>, modifiedSince: date, completionHandler:{ })
```

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/DetailWebView/OfflineReaderWebView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/DetailWebView/OfflineReaderWebView.swift
@@ -4,9 +4,19 @@ import Foundation
 /// The mechanism is quite simple: open the post in a hidden webview so the images are cached
 ///
 class OfflineReaderWebView: ReaderWebView {
-    func saveForLater(_ string: String) {
+    func saveForLater(_ post: ReaderPost, viewController: UIViewController) {
         navigationDelegate = self
 
+        frame = CGRect(x: 0, y: 0, width: viewController.view.frame.width, height: viewController.view.frame.height)
+
+        isHidden = true
+
+        viewController.view.addSubview(self)
+
+        load(post.contentForDisplay())
+    }
+
+    private func load(_ string: String) {
         // Remove all srcset from the images, only the URL in the src tag will be cached
         let content = super.formattedContent(string, additionalJavaScript: """
             document.querySelectorAll('img').forEach((el) => {el.removeAttribute('srcset')})
--- a/WordPress/Classes/ViewRelated/Reader/DetailWebView/OfflineReaderWebView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/DetailWebView/OfflineReaderWebView.swift
@@ -1,0 +1,27 @@
+import Foundation
+
+/// A WKWebView used to save post content to be available when offline
+/// The mechanism is quite simple: open the post in a hidden webview so the images are cached
+///
+class OfflineReaderWebView: ReaderWebView {
+    func saveForLater(_ string: String) {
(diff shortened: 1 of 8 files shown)


WordPress/Classes/ViewRelated/Reader/DetailWebView/OfflineReaderWebView.swift
```diff
@@ -0,0 +1,41 @@
...
+
+        super.loadHTMLString(content, baseURL: Bundle.wordPressSharedBundle.bundleURL)
+    }
+}
+
+extension OfflineReaderWebView: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        // We wait 5 secs before removing the WebView to give it time to be rendered
+        // If its removed before it was rendered, the images won't be cached
+        DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
```
emilylaguna, line 37: I know the technical challenges with properly detecting whether all the images have successfully loaded in the WebView, but 5 seconds is a pretty long time.  Have you tested using a smaller time interval?

leandroalonso: My idea is that the webview might have different times to render from device to device, so even if I'm able to find a certain amount of seconds for the simulator or the iPhone 11, will that work for an iPhone 7?

5s it's a lot, but it gives me enough confidence that this would work for most of the cases.

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderWebView.swift
```diff
@@ -17,9 +17,21 @@ class ReaderWebView: WKWebView {
 
     /// Loads a HTML content into the webview and apply styles
     ///
-    @discardableResult
-    override func loadHTMLString(_ string: String, baseURL: URL?) -> WKNavigation? {
-        let content = """
+    func loadHTMLString(_ string: String) {
+        // If the user is offline, we remove the srcset from all images
+        // This is because only the image inside the src tag is previously saved
+        let additionalJavaScript = ReachabilityUtils.isInternetReachable() ? "" : "document.querySelectorAll('img-placeholder').forEach((el) => {el.removeAttribute('srcset')})"
```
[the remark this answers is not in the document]
leandroalonso: Done in 072f961b989586e35152846153f95e5142a4242d

WordPress/Classes/ViewRelated/Reader/DetailWebView/OfflineReaderWebView.swift
emilylaguna: That makes sense. Plus it the total time will vary depending on network speeds as well. I think in the future we should try to come up with a better solution.

PR #14340, opened 2020-06-17
UL&S: Add TextFieldTableViewCell to SiteAddressViewController

mindgraffiti:
Fixes # n/a
Ref PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/306

### To test:
1. `rake dependencies`
2. Build and run
3. Log in > "Or log in by entering your site address."
4. The SiteAddressViewController appears and you can interact with the textfield. (The Continue button doesn't do anything yet.)

### Design review:
The spacing between table cells in Zeplin isn't selectable. I can see there is extra space between the label and the textfield but I don't know how much. What's rendered above is a minimum height of 44 points for the 2 cells. How much top / bottom margin should be between the 2 items?

**Zeplin design**
Screen Shot 2020-06-17 at 11 29 01 AM

**iPhone 11 Pro screenshots**
<a href="https://user-images.githubusercontent.com/1062444/84951840-a52dcc80-b0b6-11ea-9768-dfa478c7a065.png"></a>

<a href="https://user-images.githubusercontent.com/1062444/84951886-b8d93300-b0b6-11ea-9d80-3abfb0b8329e.png"></a>

PR submission checklist:


mindgraffiti:
Hey Matt 👋 . Thanks by the way! I'm still getting used to which repos have fancy features (like the Needs Design Review label) and which ones don't.

>What I've used here is the standard text field component from the iOS design library, the separatorColor and inset is defined by the system. I'm thinking that if we use the same component in code we should get the right color automatically, but let me know if that's the case.

So the textfield component is standard, yes, but the border line isn't for this case. Where that line typically comes from is a table cell inside a tableView, with separators turned on for the whole table. 

On the code side, we're breaking up each piece to be its own table cell: instruction text, textfield, text link, error label, fine print, so that dynamic text is much easier to handle and multiline text resizes correctly. 

The table is there, but the separators are hidden because the lines would show up for everything. (Right now "everything" is just the instruction label and the textfield). To make the border line show up for just the textfield, I built a custom view, sized it, and added a muriel color to it. If something in the screenshots looks misplaced, I can definitely change it!

mattmiklic:
> I built a custom view, sized it, and added a muriel color to it. If something in the screenshots looks misplaced, I can definitely change it!

Thanks for this background! On an earlier screenshot I was thinking the border was too dark/thick but it looks like in these screenshots it's looking solid.

Is it possible to adjust the height of these text fields? The component in Apple's design system is 44pt tall, which seems like a good choice since it aligns with the minimum suggested tap target of 44pt. If it's difficult to change don't worry about it, but if it's easy that would be a good standard across the board for text inputs.

ScoutHarris:
> Is it possible to adjust the height of these text fields? The component in Apple's design system is 44pt tall, which seems like a good choice since it aligns with the minimum suggested tap target of 44pt.

Another FWIW from me. We often do >= 44 (to allow for larger text), so this shouldn't be an issue.

mindgraffiti:
>Is it possible to adjust the height of these text fields?

@mattmiklic oops, yes. I didn't specify the height. Used >= 44.0 per @ScoutHarris recommendation. I also noticed that I missed a keyboard setting. It is now the expected height, with autocorrect turned off:

<a href="https://user-images.githubusercontent.com/1062444/85311821-9e6ad500-b47b-11ea-9c16-e289ca1d524c.png"></a>

<a href="https://user-images.githubusercontent.com/1062444/85311854-aaef2d80-b47b-11ea-832b-1e7770c125f4.png"></a>

mindgraffiti:
Thanks for the reviews @ScoutHarris and @mattmiklic!

PR #14341, opened 2020-06-17
Core Data: Tweaks thrown errors to include more context about what went wrong

aerych:
A quick PR to add some extra information to errors that can be thrown during a core data migration to help better understand what went wrong.  Changes include:
- Including the to and from version involved when something goes 💥 
- Including information from an original error (if one exists) as part of our thrown error description.
- A slight refactor in a couple of places where a thrown error could be out right ignored.

Xcode was complaining in a few places about the newly added `error(with:description)` method so there is a bit of a mix-match with its usage. I prefer not having to include the class as part of the usage so I've only done that where necessary. Not sure why this is happening tho.

To test:
Run the branch and make sure it builds. 
Try migrating from a previous version and make sure there are no errors. 

PR submission checklist:

@Gio2018 and @leandroalonso, hoping y'all could both take a peek (when it comes to our core data stack the more eyes the better!)

### DIFF
--- a/WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
+++ b/WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
@@ -30,7 +30,7 @@ class CoreDataIterativeMigrator: NSObject {
 
         // Get the persistent store's metadata.  The metadata is used to
         // get information about the store's managed object model.
-        // If metadataForPersistentStore throws an error that error is propigated, not replaced by the throw
+        // If metadataForPersistentStore throws an error that error is propagated, not replaced by the throw
         // in the guard's else clause.  If metadataForPersistentStore returns nil then an error is thrown.
         guard let sourceMetadata = try metadataForPersistentStore(storeType: storeType, at: sourceStore) else {
             throw error(with: .failedRetrievingMetadata, description: "The source metadata was nil.")
@@ -252,7 +252,8 @@ private extension CoreDataIterativeMigrator {
 
     static func metadataForPersistentStore(storeType: String, at url: URL) throws -> [String: Any]? {
         do {
-            return try NSPersistentStoreCoordinator.metadataForPersistentStore(ofType: storeType, at: url, options: nil)
+            let metadata = try NSPersistentStoreCoordinator.metadataForPersistentStore(ofType: storeType, at: url, options: nil)
+            return metadata
         } catch {
             let originalDescription: String = (error as NSError).userInfo[NSLocalizedDescriptionKey] as? String ?? ""
             let description = "Failed to find source metadata for store: \(url). Original Description: \(originalDescription)"
--- a/WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
+++ b/WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
@@ -5,6 +5,10 @@ import CoreData
 ///
 class CoreDataIterativeMigrator: NSObject {
 
+    private static func error(with code: IterativeMigratorErrorCodes, description: String) -> NSError {
+        return NSError(domain: "IterativeMigrator", code: code.rawValue, userInfo: [NSLocalizedDescriptionKey: description])
+    }
+
     /// Migrates a store to a particular model using the list of models to do it iteratively, if required.
     ///
     /// - Parameters:
@@ -26,8 +30,10 @@ class CoreDataIterativeMigrator: NSObject {
 
         // Get the persistent store's metadata.  The metadata is used to
         // get information about the store's managed object model.
+        // If metadataForPersistentStore throws an error that error is propigated, not replaced by the throw
+        // in the guard's else clause.  If metadataForPersistentStore returns nil then an error is thrown.
         guard let sourceMetadata = try metadataForPersistentStore(storeType: storeType, at: sourceStore) else {
-            throw MigrationError.migrationFailed
+            throw error(with: .failedRetrievingMetadata, description: "The source metadata was nil.")
         }
 
         // Check whether the final model is already compatible with the store.
@@ -75,6 +81,15 @@ class CoreDataIterativeMigrator: NSObject {
             modelsToMigrate = modelsToMigrate.reversed()
         }
 
+        // Nested function for retrieving a model's version name.
+        // Used to give more context to errors.
+        func versionNameForModel(model: NSManagedObjectModel) -> String {
+            guard let index = objectModels.firstIndex(of: model) else {
+                return "Unknown"
+            }
+            return modelNames[index]
+        }
+
         // Migrate between each model. Count - 2 because of zero-based index and we want
         // to stop at the last pair (you can't migrate the last model to nothingness).
         let upperBound = modelsToMigrate.count - 2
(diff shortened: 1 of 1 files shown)


WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -203,35 +238,33 @@ private extension CoreDataIterativeMigrator {
...
-
-        guard let sourceMetadata = try? NSPersistentStoreCoordinator.metadataForPersistentStore(ofType: storeType, at: url, options: nil) else {
-            let description = "Failed to find source metadata for store: \(url)"
-            throw NSError(domain: "IterativeMigrator", code: 102, userInfo: [NSLocalizedDescriptionKey: description])
+        do {
+            return try NSPersistentStoreCoordinator.metadataForPersistentStore(ofType: storeType, at: url, options: nil)
+        } catch {
+            let originalDescription: String = (error as NSError).userInfo[NSLocalizedDescriptionKey] as? String ?? ""
+            let description = "Failed to find source metadata for store: \(url). Original Description: \(originalDescription)"
+            throw CoreDataIterativeMigrator.error(with: IterativeMigratorErrorCodes.noMetadataForStore, description: description)
```
Gio2018, line 259: how would this error propagate to the caller? I believe the one from `NSPersistentStoreCoordinator.metadataForPersistentStore` would prevail if an error is thrown.

Gio2018:
Did some migration from scratch from a random number of version between 14.5 and 15.0, as well as migrating back and forth, and observed no crash

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
aerych: I'm not sure I follow.  🤔

Gio2018: I believe that, since we are doing `return try...` , in case `NSPersistentStoreCoordinator.metadataForPersistentStore` throws an error, the `catch` block will be executed, but the error thrown in that block will not be thrown by our `static func metadataForPersistentStore(storeType: String, at url: URL)`.

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -26,8 +30,10 @@ class CoreDataIterativeMigrator: NSObject {
 
         // Get the persistent store's metadata.  The metadata is used to
         // get information about the store's managed object model.
+        // If metadataForPersistentStore throws an error that error is propigated, not replaced by the throw
```
leandroalonso, line 33:
```suggestion
        // If metadataForPersistentStore throws an error that error is propagated, not replaced by the throw
```

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -75,24 +81,53 @@ class CoreDataIterativeMigrator: NSObject {
             modelsToMigrate = modelsToMigrate.reversed()
         }
 
+        // Nested function for retrieving a model's version name.
+        // Used to give more context to errors.
+        func versionNameForModel(model: NSManagedObjectModel) -> String {
```
leandroalonso, line 86: Does this need to be public?

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
leandroalonso: @Gio2018 I would expect that the return wouldn't succeed and the execution will then follow to the `catch` block.

But I'm not sure, is something that we should double-check.

aerych: Right. If the `try` throws an error execution is interrupted before the `return` is evaluated and instead execution switches to the `catch` block.  Happy to split the logic of that line up so the return is part of a separate statement if it will clarify things!

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
aerych: Well, this is a nested function with local scope. It actually can't be declared `public`, which makes sense as there would be no way for code outside of its containing method to call it.

aerych: Arg! Typos! Thanks for spotting that!

aerych:
Updated to fix a typo and to clarify some code execution at a glance.  Ready for another look!

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
Gio2018: I just did some very quick tests on a playground: if the function throws, looks like the error of the throwing function will be thrown.. So regardless the return, this might be a common problem for any error thrown in a catch block of a throwing function (sounds like a riddle :) ). I'll do some more tests and share with you later

aerych: If I understand correctly, it sounds like if there is a method `bar` that throws, and its called by a method `baz` via a `do/try/catch`, and `baz` throws inside its catch block, that error thrown by `baz` would actually be the error thrown by `bar`?  Did I understand that right? 
I did a test in a playground with the following code block and it seems that `baz` throws its own error, not the error that was thrown by `bar`.  Did I misunderstand something? 

```
import UIKit
import Foundation

class Foo {

    static func bar() throws -> Bool {
        throw NSError(domain: "bar", code: 1, userInfo: nil)
    }

    static func baz() throws -> Bool {
        do {
            return try bar()
        } catch {
            throw NSError(domain: "baz", code: 2, userInfo: nil)
        }
    }

}

do {
    try Foo.baz()
} catch {
    let err = error as NSError
    print( err.domain )
}

```

Gio2018: Yes you are right. I thought I was picturing something, but I was pointing to the wrong direction. On the other hand, there still might be a problem (which predates this PR) in the caller, which is `iterativeMigrate`. The following code in a playground:

`import Foundation

enum FakeError: Error {
    case first
    case second
    case third

    var localizedDescription: String {
        switch self {
        case .first:
            return "FIRST"
        case .second:
            return "SECOND"
        case .third:
            return "THIRD"
        }
    }
}

// Equivalent of NSPersistentStoreCoordinator.metadataForPersistentStore
func firstThrowingFunction() throws -> String {
    print("throwing FIRST Error")
    throw FakeError.first
}

// Equivalent of metadataForPersistentStore
func secondThrowingFunction() throws -> String? {

    do {
        return try firstThrowingFunction()
    } catch {
        print("throwing SECOND Error")
        throw FakeError.second
    }
}

// Equivalent of iterativeMigrate
func thirdThrowingFunction() throws {

    guard let _ = try secondThrowingFunction() else {
        print("throwing THIRD Error")
        throw FakeError.third
    }
}


func test() {

    do {
        try thirdThrowingFunction()
    } catch {
        if let error = error as? FakeError {
            print("THE ERROR IS: " + error.localizedDescription) // prints THE ERROR IS: SECOND
        }
    }
}

test()`

`thirdThrowingFunction` (which is the "equivalent" of our `iterativeMigrate` would actually throw the second, not the third error. Does this make sense?

aerych: Yep. Makes total sense because there's no do/catch around the try. However, I think this might actually be what we want in `iterativeMigrate`.  The call to `metadataForPersistentStore` can throw, and we want that error to surface.  But if it doesn't throw, there is still a chance that it could return a nil value, at least the way its coded I'm not sure if that's the case in practice. It seemed to me that `iterativeMigrate`'s `throw` statement was intended to handle a cases where nil was returned from `metadataForPersistentStore` rather than try to handle the errors it throws.  Does that make sense?  Am I off base with my read of the code?

Gio2018: Yep I think it does make sense and it's pretty clear now, sorry that this took so long.

aerych: Not at all! Thanks for being so thoughtful and thorough :)

aerych:
Thanks @Gio2018 and @leandroalonso !

PR #14343, opened 2020-06-17
Core Data: Normalizes model config for stats entities.

aerych:
Refs #12028

While digging into #12028 I found a number of events where the cause was `TopCommentedPostStatsRecordValue.postURLString` being `nil` when Core Data was saved.  The data model did not define the `postURLString` field for this entity as optional, even tho the code across the app did. (The API code is written such that a URL should always be returned so an optional here is a little odd, but I digress.) 

Looking at other stats model entities I found inconsistent property definitions where numerical fields might have default values or not, and where string, date, and numerical fields might be optional or not.  

It's important to resolve the core data crash, and since the code seems to be pretty consistent in how it handles these different fields the best thing seems to be to update the data model to be consistent with the code in how it treats optionals and to ensure numerical fields have default values.

This PR makes a number of changes across the stats models. A detailed list is below but the TL;DR is: 
- A new model version is added (97).
- Scalar numeric fields now all default to 0 if they didn't previously. Any that were marked optional are no longer optional.
- String and Date fields that were not marked optional are now marked optional for consistency with String and Date fields that WERE marked optional.

One `postID` field was defined as a string and I've opted to not change this yet. 

To test:
- Check the data model changes and ensure nothing was actually supposed to be the way it previously was. 
- Run the app and give the stats feature a thorough testing. Confirm there are no crashes. 

PR submission checklist:

@ScoutHarris could I trouble you for a review of this one? 

-------

### Changes to Stats Models

VisitsSummaryStatsRecordValue
- Defaults to Ints 
- date made optional

TopViewedVideoStatsRecordValue
- postID default to 0

TopViewedPostStatsRecordValue
- viewCount, type, postID default to 0

TopViewedAuthorStatsRecordValue
- no changes

TopCommentsAuthorStatsRecordValue
- commentCount default to 0

TopCommentedPostStatsRecordValue
- postURLString optional
- commentCount default to 0
(??? postID is a string???) (Not changed as this would be more impactful)

TodayStatsRecordValue
- all ints default to 0

TagsCategoriesStatsRecordValue
- ViewCount no longer optional
- name optional
- type default to 0

StreakStatsRecordValue
- postCount no longer optional

StreakInsightStatsRecordValue
- currentStreakLength  no longer optional
- longestStreakLength  no longer optional

StatsRecord
- period defaults to 0

SearchResultsStatsRecordValue
- viewsCount defaults to 0

ReferrerStatsRecordValue
- viewCount defaults to 0
- label made optional

LastPostStatsRecordValue
- title made optional
- publish date made optional
- postID defaults to 0

FollowersStatsRecordValue
- type defaults to 0
- name is optional

FollowersCountStatsRecordValue
- type and count,  not optional, defaults to 0

FileDownloadsStatsRecordValue
- downloadcount defaults to 0

CountryStatsRecordValue
- viewsCount defaults to 0
- countryCode made optional

ClicksStatsRecordValue
- label made optional
- clicksCount defaults to 0

AnnualAndMostPopularTimeStatsRecordValue
- all fields default to 0

AllTimeStatsRecordValue
- bestViewsDate is optional
- int fields default to 0

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -3,6 +3,12 @@
 This file documents changes in the data model. Please explain any changes to the
 data model as well as any custom migrations.
 
+## WordPress 97
+
+@aerych 2020-06-17
+
+- All stats entities were reviewed for consistency of Optional settings for strings and dates and default values for scalar numerical fields.
+
 
 ## WordPress 94
 
--- a/WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
+++ b/WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>WordPress 96.xcdatamodel</string>
+	<string>WordPress 97.xcdatamodel</string>
 </dict>
 </plist>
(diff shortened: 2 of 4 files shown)


aerych:
Thanks @ScoutHarris !!

PR #6451, opened 2017-01-23
Categories: Add refresh control and sync for categories.

kurzee:
Fixes #2573 

This adds a simple `UIRefreshControl` to the `PostCategoriesViewController` and reloads the table once synced. This resolves the issue of users not being able to pull in any recent changes to categories if the app hasn't synced the blog. It also wasn't obvious how to refresh categories without force-closing the app and relaunching.

To test:
1. Add a self-hosted site.
2. Go straight to Post Options > Categories.
3. Check that the list is currently in sync with the web.
4. Make a change on the web.
5. Pull down the refresh, check that the list is again in sync with the web.

To test:
1. Repeat steps for a .com or Jetpack site.

Needs review: @frosty can you give this is a gander?

PR #14352, opened 2020-06-18
Categories: Makes fields scalar non optional. Adjusts order of execution

aerych:
Refs #12028 

This PR attempts to address some core data save errors where `Category.parentID` is `nil`, or where `Category.blog` is `nil`.  The exact circumstance that leads to the error remains unknown so the changes here are a best guess about what could contribute to the error. 

Changes:
- I've removed where Category.blog was explicitly set to `nil` when deleting a Category. The Category entity is already configured to nullify the `blog` relation when deleted so this should have been unnecessary and it's not clear why it [was added](https://github.com/wordpress-mobile/WordPress-iOS/pull/6451/files).
- The data model is updated so numeric fields are scalar and non optional. This should result in there always being a default value as opposed to nil when saving.
- I've modified code in PostCategoryService so `success` is called _after_ the context is saved. The same for where Categories are synced for self-hosted sites when creating a new category.  The reason for this is work is done asychronously across multiple `performBlock` calls. The way things were, a success block was called before the managed object context was saved.  Same for a call to sync categories (tho this was unlikely to finish before the context was saved).  Taken together this presented a scant window of time where inflight changes were assumed to be complete, even tho they weren't, and it may be the case this contributes to the core data crashes.  It's hard to say for sure.

Note that the model version was already bumped to 97 in a previous PR for this release cycle so I have not bumped it again. Expect to be logged out when launching the app since there is not a migration for the changes in this PR.

To test:
Perform the following smoke tests and confirm the app does not crash and categories function as expected.

### Scenario 1 - Try this with both a wpcom and self-hosted site.
- Launch the branch and browse to the Categories feature under post settings. 
- Add some new categories.
- Tap back to post settings. 
- Delete some categories via the web.
- Back in the app return to the list of categories and confirm that the list updates to remove the categories you deleted.

### Scenario 2 
- Browse to the Menus feature and add try to add a new menu item. 
- Select a category menu item
- Confirm the list of categories appears as expected.

PR submission checklist:

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -8,7 +8,7 @@ data model as well as any custom migrations.
 @aerych 2020-06-17
 
 - All stats entities were reviewed for consistency of Optional settings for strings and dates and default values for scalar numerical fields.
-
+- Categories entity updated to make numeric fields scalar and non-optional.
 
 ## WordPress 94
 
--- a/WordPress/Classes/Services/PostCategoryService.m
+++ b/WordPress/Classes/Services/PostCategoryService.m
@@ -155,10 +155,11 @@ NS_ASSUME_NONNULL_BEGIN
                            }
                            PostCategory *newCategory = [self newCategoryForBlog:blog];
                            newCategory.categoryID = receivedCategory.categoryID;
+                           BOOL needsSync = NO;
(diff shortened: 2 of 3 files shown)


aerych:
Thank you @jleandroperez !!

Issue #14354, opened 2020-06-19
Crash opening Gutenberg after logging out and in again


etoledom:
### Expected behavior
Editor opens normally

### Actual behavior
💥

### Steps to reproduce the behavior
- Open Gutenberg on a blog (tested on a Premium wpcom site)
- Logout from the app
- Log back in right away
- The app should show the same blog as before
- Go to Posts
- Open a post
- 💥

The crash seems to be related to an early network call from here:
https://github.com/wordpress-mobile/WordPress-iOS/blob/9f2a606159efaaea4d8caec5332f761a88140ee1/WordPress/Classes/Stores/EditorThemeStore.swift#L102

The app decided to make the request as a **Self Hosted** site and ends up crashing here:

https://github.com/wordpress-mobile/WordPress-iOS/blob/9f2a606159efaaea4d8caec5332f761a88140ee1/WordPress/Classes/Models/Blog.m#L202

While trying to get the Rest API object.

I don't think it's a problem specific to the Editor Theme feature but pinging @chipsnyder just in case. There might be something wrong with the Blog object passed.

cc @pinarol

pinarol:
Thanks for reporting and the analysis @etoledom! Can you also attach the crash log?

Can you elaborate what you mean by "early" network call? Is there something like a state in the app which is too early to make a network call?

etoledom:
Crash log:
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[NSRegularExpression enumerateMatchesInString:options:range:usingBlock:]: nil argument'

Thread backtrack:

```
   frame #0: 0x00000001a4a3cbe0 libobjc.A.dylib`objc_exception_throw
    frame #1: 0x00000001a502fc50 Foundation`-[NSRegularExpression(NSMatching) enumerateMatchesInString:options:range:usingBlock:] + 1856
    frame #2: 0x00000001a5052748 Foundation`-[NSRegularExpression(NSMatching) matchesInString:options:range:] + 184
    frame #3: 0x00000001a505c374 Foundation`-[NSRegularExpression(NSReplacement) stringByReplacingMatchesInString:options:range:withTemplate:] + 168
  * frame #4: 0x0000000100c6c0f0 WordPress`-[Blog urlWithPath:](self=0x0000000283ff7f60, _cmd="urlWithPath:", path=@"wp-json/") at Blog.m:202:12
[...]

PR #14359, opened 2020-06-19
Add guard around self-hosted site generation to help limit the risk o…

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/WordPress-iOS/issues/14354

The exception reported in the ticket was a result of the xmlrpc path not being populated yet. The code attempts to [catch](https://github.com/wordpress-mobile/WordPress-iOS/blob/ea2bf8d5d942f6c735a9ee66c374c31f773823b8/WordPress/Classes/Networking/WordPressOrgRestApi%2BWordPress.swift#L34) possible errors however there seems to be an interesting difference between how errors are thrown from Objective-C and how swift manages them.

> Although Swift error handling resembles exception handling in Objective-C, it is entirely separate functionality. If an Objective-C method throws an exception during runtime, Swift triggers a runtime error. There is no way to recover from Objective-C exceptions directly in Swift. Any exception handling behavior must be implemented in Objective-C code used by Swift.
Excerpt From: Apple Inc. “Using Swift with Cocoa and Objective-C (Swift 2.1).” iBooks.
https://stackoverflow.com/a/35141887/1350218

As a result, I added a guard around this point to return nil which would have followed the logic if this catch was working as expected.

## To test:

- Open Gutenberg on a blog (tested on a Premium wpcom site)
- Logout from the app
- Log back in right away
- The app should show the same blog as before
- Go to Posts
- Open a post
## PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 15.2
 ----
 * [**] Block editor: Display content metrics information (blocks, words, characters count).
+* [*] Fixed a crash that results in navigating to the block editor quickly after logging out and immediately back in.
 
 -----
  
--- a/WordPress/Classes/Networking/WordPressOrgRestApi+WordPress.swift
+++ b/WordPress/Classes/Networking/WordPressOrgRestApi+WordPress.swift
@@ -30,7 +30,6 @@ private func makeCookieNonceAuthenticator(blog: Blog) -> Authenticator? {
 
 private func apiBase(blog: Blog) -> URL? {
     precondition(blog.account == nil, ".com support has not been implemented yet")
-    guard blog.xmlrpc != nil else { return nil }
     return try? blog.url(withPath: "wp-json/").asURL()
 }
 
--- a/WordPress/Classes/Networking/WordPressOrgRestApi+WordPress.swift
+++ b/WordPress/Classes/Networking/WordPressOrgRestApi+WordPress.swift
@@ -30,6 +30,7 @@ private func makeCookieNonceAuthenticator(blog: Blog) -> Authenticator? {
 
 private func apiBase(blog: Blog) -> URL? {
     precondition(blog.account == nil, ".com support has not been implemented yet")
+    guard blog.xmlrpc != nil else { return nil }
     return try? blog.url(withPath: "wp-json/").asURL()
 }
 
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
@@ -341,14 +341,14 @@ class GutenbergViewController: UIViewController, PostEditor {
     private var previousFirstResponder: UIView?
 
     private func setupKeyboardObservers() {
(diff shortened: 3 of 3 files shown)


chipsnyder:
> My test site is a wpcom site, so the execution shouldn't have gone to the .org side to begin with.

Yeah I think the fallthrough happens here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Gutenberg/GutenbergNetworking.swift#L15-L21

We might be better off adding a second check that includes a new boolean along the lines of `blog.isAccessibleThroughXMLRPC()` then call the failure completion if neither API is available.

> I also noticed that every time we close and re-open the editor, the amount of requests we make to /wp/v2/themes?status=active increases. This regardless of the logout/login step.

Good catch on that. Sounds like we have a retain cycle somewhere. I'll take a look at that and make the adjustment.

> What it seems it's happening is those accumulated requests are storing a reference to a blog object that becomes faulty after login out, and re-triggering those requests things will go wrong.
>
> Could you please take another look at this? 

If it is holding on to bad blog instance that might reduce the need to add the `blog.isAccessibleThroughXMLRPC` suggested above. I wonder if that might still be useful though just in case. I'll try it without first then assess from there.

chipsnyder:
@etoledom 

> Sounds like we have a retain cycle somewhere.

This ended up being the case. Looks like this was introduced as part of [15.0](https://github.com/wordpress-mobile/WordPress-iOS/blob/3c00578ee73b352d4416e28ffa9c43c1e2178100/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L304-L317) I reverted the change I originally made so that way we could focus in on the root cause. 

FYI: @pinarol @hypest not sure if we're seeing many out of memory crashes in the app but this is causing a retain cycle every time we open the editor.

hypest:
> FYI: @pinarol @hypest not sure if we're seeing many out of memory crashes in the app but this is causing a retain cycle every time we open the editor.

How bad can this be, and is it a different issue than the ticket this PR is trying to address? We should extract it as its own hotfix if bad enough.

chipsnyder:
> is it a different issue than the ticket this PR is trying to address?

This retain-cycle is the root cause of the issue in the ticket. 

> How bad can this be

There is one other component to the retain cycle that I'm narrowing in on now. But it basically will translate to crashes from out of memory and slowness in the app for users who frequently use the editor. 

My gut says this will be fine until 15.2 but I wanted to drop the FYI  in case we see complaints that the app is slowing down or crashing when launching the editor.

chipsnyder:
> I have one more tweak to make then this should be ready

@etoledom This is actually good to go now. My "one last tweak" was me forgetting I had Zombies enabled so I was tracking down allocations that didn't actually exist.

**Before**
Screen Shot 2020-06-24 at 11 08 45 AM

**After** (With the GB editor open)
Screen Shot 2020-06-24 at 1 52 00 PM


> How bad can this be

I double-checked Sentry and I don't see any out memory exceptions/crashes so this doesn't seem like a big concern. So you can probably ignore the FYI before.

pinarol:
Nice catch @chipsnyder !

PR #14360, opened 2020-06-22
Fix WebViews Auth on WPCom after login in.

etoledom:
This PR fixes an auth issue on Previews due to changes in the backend.
We are not setting the WordPress User Agent to the auth cookies request to avoid being blocked.

Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2411

To test:
- Clean install the WordPress app.
- Login to a WPCom account.
- Create a new post on a blog that's not private Atomic.
- Write something and choose to preview (without publishing).
- Check that the preview works.

**Note:** This PR was initially intended to fix an auth issue on Gutenberg Web-View for the unsupported block fallback project. So this also fixes parts of https://github.com/wordpress-mobile/gutenberg-mobile/issues/2358

PR submission checklist:

cc @marecar3 @aerych

### DIFF
--- a/WordPress/Classes/Services/AuthenticationService.swift
+++ b/WordPress/Classes/Services/AuthenticationService.swift
@@ -94,7 +94,9 @@ class AuthenticationService {
                     // so that the user sees a reasonable error situation on screen.
                     // We could opt to create a special screen but for now I'd rather users report
                     // the issue when it happens.
-                    failure(error)
+                    DispatchQueue.main.async {
+                        failure(error)
+                    }
                 }
         }
     }
@@ -141,6 +143,7 @@ class AuthenticationService {
         headers.forEach { (key, value) in
             request.setValue(value, forHTTPHeaderField: key)
         }
+        request.setValue(WPUserAgent.wordPress(), forHTTPHeaderField: "User-Agent")
 
         let task = session.dataTask(with: request) { data, response, error in
             if let error = error {
--- a/WordPress/Classes/Services/AuthenticationService.swift
+++ b/WordPress/Classes/Services/AuthenticationService.swift
@@ -94,9 +94,7 @@ class AuthenticationService {
                     // so that the user sees a reasonable error situation on screen.
                     // We could opt to create a special screen but for now I'd rather users report
                     // the issue when it happens.
-                    DispatchQueue.main.async {
-                        failure(error)
-                    }
+                    failure(error)
                 }
         }
     }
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Utility/CookieJar.swift
```diff
@@ -156,19 +156,13 @@ extension WKHTTPCookieStore: CookieJarSharedImplementation {
...
         }
 
-        for cookie in cookies {
-            DispatchQueue.main.async {
-                self.setCookie(cookie, completionHandler: completionIncrement)
+        DispatchQueue.main.async {
+            self.setCookie(cookie) { [weak self] in
+                self?.setCookies(cookies.dropLast(), completion: completion)
             }
         }
```
etoledom, line 167: This should ensure that `completion()` is called if `cookies` is empty, while still adding cookies sequentially.

WordPress/Classes/Services/AuthenticationService.swift
```diff
@@ -94,7 +94,9 @@ class AuthenticationService {
                     // so that the user sees a reasonable error situation on screen.
                     // We could opt to create a special screen but for now I'd rather users report
                     // the issue when it happens.
-                    failure(error)
+                    DispatchQueue.main.async {
+                        failure(error)
+                    }
```
etoledom, line 99: `success()` was called on main thread while `failure()` was called on background thread. This is for consistency.

aerych: I'm wondering which success was called.  The one on line 81 (which would be main thread) or the one on line 87 (which doesn't look like it would be the main thread).  
I agree we should dispatch to the main thread but I wonder if we could move this into the `requestAuthCookies` method and wrap both `success` and `failure` calls there so its all done in one place? 
(I coincidentally ran across the threading issue while putting together #14363 and remembered you'd made this change.)

etoledom: Moving it to `requestAuthCookies` sounds good!

For `success` on line 87, execution has been already moved to main thread by `setCookies`. At least in the `WKHTTPCookiesStore` implementation of it. That's why it works.

But I agree this is shady, so let's improve :)

etoledom:
Switched base branch to `hotfix/15.0.1` to release as a Hotfix

cc @elibud @maxme @hypest

etoledom:
Cherry-picked https://github.com/wordpress-mobile/WordPress-iOS/pull/14316 to fix failing CI test

Issue #13953, opened 2020-04-21
The operation couldn’t be completed. (WordPressKit.WordPressComRestApiError error 7.)


loremattei:
**Sentry Url**: https://sentry.io/share/issue/9ff043231d4c4b978ff53ddb3c367c3b/

**User Count**: 27
**Count**: 122
**First Release**: 14.6.0.0
**First Seen**: 2020-04-08T16:22:24Z
**Last Release**: 14.6.0.3
**Last Seen**: 2020-04-21T14:10:23Z
**24 Hours**: 97
**30 Days**: 122

```
The operation couldn’t be completed. (WordPressKit.WordPressComRestApiError error 7.)

```

yaelirub:
Triage: Priority high by version.
Seems to be an issue in PreviewWebKitViewController or WebKitViewController
@bjtitus , would you  mind taking a look at this?

bjtitus:
It seems like the `unknown` `WordPressComRestApiError` comes from the Atomic authentication service (added in https://github.com/wordpress-mobile/WordPress-iOS/pull/13712) when it fails to fetch a cookie from the API. I tried on my own ephemeral atomic site and was unable to reproduce the issue.

@yaelirub Is there anyone familiar with the atomic stuff who might be able to take a look or provide some tips?

yaelirub:
Thanks so much for looking into this. I hope @aerych might be able to provide some help on this crash 🙏

bjtitus:
Here are a few updates on this from my investigation over the last few days:

- I was able to reproduce this error 100% of the time on a wordpress.com atomic site.
- To be clear, this is **NOT** a crash but is an error logged whenever the atomic proxy cookie fetch fails.
- I was unable to reproduce user facing bugs related to the cookie failure. The `frame-nonce` query parameter allows private page and post previews to be shown without the cookies. I'm not sure if there are cases where that parameter does not work.
- My best guess is that the required Jetpack xmlrpc method (`getReadAccessCookies`) isn't available on all instances yet.

### Details

Can reproduce by:
* Setting up an Atomic Site on WordPress.com
* Attempting to view any page or post preview on the blog.

Kibana/logstash shows the following logged from the cookie request:
```json
"url": "/wpcom/v2/sites/170037822/atomic-auth-proxy/read-access-cookies?_locale=en",
"@version": "1",
"message": "Auth failed: server error. requested method jetpack.getReadAccessCookies does not exist.",
```

[...]

PR #14363, opened 2020-06-22
Auth: Use atomic cookie auth only if atomic and private.

aerych:
Refs #13953 

This fixes an issue where the app would attempt to fetch auth cookies for atomic sites that were not private.  The endpoint in question assumes that it will only be used with a private atomic site and returns a 500 error in other cases. 

To test:
Use a private atomic site, set to coming soon or not, and confirm you can successfully get auth cookies, view the site's media, and preview a draft post
Repeat the test with a public atomic site and confirm auth cookies are not fetched and not needed to preview posts or view media.

@etoledom Could I trouble you with this one?

PR submission checklist:


aerych:
Thanks @etoledom !

PR #14366, opened 2020-06-22
Fix spotlight on the reader tab icon when rotating devices

### LINKED ISSUE #14286: Quick Start v2 > Follow another site > the spotlight does not rotate
### Expected behavior
See also comment in https://github.com/wordpress-mobile/WordPress-iOS/pull/14276#pullrequestreview-427947921. When the tab bar rotates, the spotlight needs redrawn.

### Actual behavior
The spotlight UI does not refresh and it moves near the "My Site" tab bar item.

1. Create a new site
2. Yes, Help Me when prompted
3. Close the "Customize Your Site" modal view
4. Open "Grow Your Audience"
5. Choose "Follow other sites"
6. Rotate the device or sim

Gio2018:
Fixes #14286

To test:

1. Create a new site
2. Yes, Help Me when prompted
3. Close the "Customize Your Site" modal view
4. Open "Grow Your Audience"
5. Choose "Follow other sites"
6. Rotate the device or sim and verify that the spotlight stays close to the reader icon tab

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ no new functionalities were added
- [x] ~~I have considered adding accessibility improvements for my changes.~~ no changes in UI elements
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ only fixes an issue, no user facing features changes or added

### DIFF
--- a/WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
+++ b/WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
@@ -16,7 +16,7 @@ extension WPTabBarController {
             let newSpotlight = QuickStartSpotlightView()
             self?.view.addSubview(newSpotlight)
 
-            guard let tabButton = self?.getTabButton(at: 1) else {
+            guard let tabButton = self?.getTabButton(at: Int(WPTabType.reader.rawValue)) else {
                 return
             }
 
@@ -49,7 +49,7 @@ extension WPTabBarController {
     }
 
     private func getTabButton(at index: Int) -> UIView? {
-        var tabs = self.tabBar.subviews.compactMap { return $0 is UIControl ? $0 : nil }
+        var tabs = tabBar.subviews.compactMap { return $0 is UIControl ? $0 : nil }
         tabs.sort { $0.frame.origin.x < $1.frame.origin.x }
         return tabs[safe: index]
     }
--- a/WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
+++ b/WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
@@ -1,5 +1,4 @@
 private var spotlightView: QuickStartSpotlightView?
-private let spotlightViewOffset: CGFloat = -5.0
 private var quickStartObserver: NSObject?
 
 extension WPTabBarController {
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
```diff
@@ -10,17 +9,26 @@ extension WPTabBarController {
...
+                [.readerTab].contains(element) else {
                     return
             }
 
             let newSpotlight = QuickStartSpotlightView()
-            tabBar.addSubview(newSpotlight)
+            self?.view.addSubview(newSpotlight)
 
-            let x = tabBar.bounds.size.width / 2.0
+            guard let tabButton = self?.getTabButton(at: 1) else {
```
frosty, line 19: Can we use the `WPTabReader` value here, instead of `1`?

WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
```diff
@@ -10,17 +9,26 @@ extension WPTabBarController {
 
             guard let userInfo = notification.userInfo,
                 let element = userInfo[QuickStartTourGuide.notificationElementKey] as? QuickStartTourElement,
-                [.readerTab].contains(element),
-                let tabBar = self?.tabBar else {
+                [.readerTab].contains(element) else {
                     return
             }
 
             let newSpotlight = QuickStartSpotlightView()
```
frosty, line 16: I wonder if we should move the code in this observer block into a separate method, just to make things a little cleaner.

WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
```diff
@@ -39,4 +47,16 @@ extension WPTabBarController {
         NotificationCenter.default.removeObserver(quickStartObserver as Any)
         quickStartObserver = nil
     }
+
+    private func getTabButton(at index: Int) -> UIView? {
+        var tabs = self.tabBar.subviews.compactMap { return $0 is UIControl ? $0 : nil }
```
frosty, line 52: Nitpick, can remove `self` here

WordPress/Classes/ViewRelated/System/WPTabBarController+QuickStart.swift
Gio2018: So the method `startWatchingQuickTours` already does basically just that. I can create another method to include only what's in the closure, but then it would be a method with a notification and a view as arguments, it would sort of look like the old way using a `selector` instead of the closure. Do you mind if we leave it as it is?

frosty: Sure, totally fine!

PR #14367, opened 2020-06-23
Fix navigation bar title when scrolling

leandroalonso:
This PR fixes the navigation bar title behavior when scrolling down/up in the Reader Detail.

It also makes sure that the Reader content doesn't have any `script` tag, preventing possible JavaScript injections.

#### Reader

1. Open a post in the Reader
2. Scroll down/up
3. Check that the navbar fades in or out depending on how much you scroll

#### In the app

Make sure the nav bar has the same font size / weight.

### DIFF
--- a/WordPress/Classes/System/WordPressAppDelegate.swift
+++ b/WordPress/Classes/System/WordPressAppDelegate.swift
@@ -883,7 +883,7 @@ extension WordPressAppDelegate {
         let maximumPointSize = WPStyleGuide.maxFontSize
 
         UINavigationBar.appearance().titleTextAttributes = [NSAttributedString.Key.foregroundColor: UIColor.white,
-                                                            NSAttributedString.Key.font: WPStyleGuide.fixedFont(for: UIFont.TextStyle.headline, weight: UIFont.Weight.bold)]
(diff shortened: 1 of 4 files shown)


WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -883,7 +883,7 @@ extension WordPressAppDelegate {
         let maximumPointSize = WPStyleGuide.maxFontSize
 
         UINavigationBar.appearance().titleTextAttributes = [NSAttributedString.Key.foregroundColor: UIColor.white,
-                                                            NSAttributedString.Key.font: WPStyleGuide.fixedFont(for: UIFont.TextStyle.headline, weight: UIFont.Weight.bold)]
+                                                            NSAttributedString.Key.font: WPStyleGuide.fixedFont(for: UIFont.TextStyle.headline, weight: UIFont.Weight.semibold)]
```
leandroalonso, line 886: For more info on this change check the commit message: https://github.com/wordpress-mobile/WordPress-iOS/pull/14367/commits/10776b6576e878d8c70eaf36615c141fa1127053

Issue #14190, opened 2020-05-26
Reader Card: Tapping More button does nothing


emilylaguna:
### Expected behavior
When tapping the more (...) button you are shown the list of additional options

### Actual behavior
When tapping the more (...) button on some posts nothing happens

### Steps to reproduce the behavior
1. Open the app
2. Open the Reader tab
3. Tap the following tab
4. Tap the cog to add a new site
5. Add 'en.blog.wordpress.com'
6. Refresh the following feed
7. Tap the 'more' button

##### Tested on [device], iOS [version], WPiOS [version]
iPhone 11 Pro, 13.3.3, 14.9

rezzap:
There was a report by a user asking where the 'visit' button which shows in the overflow menu accessed using the more (...) button was in 3073074-zen

On testing this, I was unable to open the overflow menu for quite some time while trying to click on it. It seems very specific to where you tap on the screen. I'm asking the user to report if they are having difficulty doing so also.

<img width="603" alt="Image 2020-06-19 at 9 25 10 am" src="https://user-images.githubusercontent.com/10325627/85113942-061ce800-b210-11ea-8373-a7a4ea0792de.png">

Tested on iPhone X, iOS 13.5.1, WPiOS 15.0

emilylaguna:
I looked into this briefly and found that:
- The button event is being triggered but the event is being fizzled down the chain 
- It gets to line [14 of ReaderMenuAction](https://href.li/?https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/ReaderMenuAction.swift#L14) and stops because the topic is nil

designsimply:
<sup>(internal reference: pbArwn-gP-p2 for related design discussion, scroll to "Context" section 2nd bullet)</sup>

designsimply:
I tested with WPiOS 15.0 App Store version and WPiOS 15.1.0.1 Testflight beta and, as @rezzap mentioned, it seems very specific to where you tap on the screen. 

I found that it's super easy to accidentally tap to the right of the three dots to open an individual post view (as if you tapped the image or title just below it). I wasn't able to trigger the problem where tapping on the three dots in the Reader cards is nil or unresponsive.

<img src="https://user-images.githubusercontent.com/1119271/85181429-2619cf00-b243-11ea-8a44-0ca6cd9478c6.gif" width="270" alt="RPReplay_Final1592601805">

[...]

PR #14370, opened 2020-06-23
Fixes reader more menu not appearing

emilylaguna:
**Fixes** #14190

## Bug description
When a user taps the more menu the event looks for a `ReaderSiteTopic` for the `ReaderPost`'s site, and if the site topic doesn't exist for some reason the the event would be ignored and nothing will happen. 

```
let service = ReaderTopicService(managedObjectContext: context)
--> if let topic = service.findSiteTopic(withSiteID: post.siteID) { <-- if this is nil, nothing happens
    showMenuForPost(post, context: context, topic: topic, readerTopic: readerTopic, fromView: anchor, vc: vc)
    return
}
```

### Why would the site topic be nil?
Prior to 14.7 the `ReaderMenuViewController` called the [`syncTopics`](https://github.com/wordpress-mobile/WordPress-iOS/blob/14.6/WordPress/Classes/ViewRelated/Reader/ReaderMenuViewController.swift#L253) method which sync'd both the reader more menu and the followed sites.

The `fetchFollowedSites` method creates a `ReaderSiteTopic` object for each of the users followed sites.

When the new reader tab navigation was introduced this file is no longer used and the followed sites network request is triggered asynchronously when the user taps the tab bar filter button, then the followed sites tab. This means that `ReaderSiteTopic` objects aren't created until then which causes the `service.findSiteTopic(withSiteID:)` method to return nil

#### Part 1: Sync the followed sites at the same time we pull down the reader more menu
When the reader tab is loaded we'll pull down both the reader more menu and the followed sites to mimic how it used to work before the new tab bar change in 14.7. 

#### Part 2: Remove the `ReaderSiteTopic` requirement from the show menu action
The  `ReaderPostMenu.showMenuForPost` method already supports passing an optional for the site topic. I updated the instances where this is called to ignore the requirement to have a site topic before displaying. 

This is a fallback just in case the topic doesn't exist for the user. It presents all options except the `Turn on/off site notifications` item. 

## To test:
1. Delete the app, this resets your database
2. Reinstall the app
3. Tap on the Reader tab
4. Tap on the more button (`⋮`, vertical ellipsis) on a post in the reader
5. Verify the more menu appears with the following options:
```
- Turn on/off site notifications
- Unfollow site
- Visit
- Share
```
6. Verify each of the options work as intended

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
@@ -86,14 +86,35 @@ extension ReaderTabItemsStore {
             return
         }
         state = .loading
-        service.fetchReaderMenu(success: { [weak self] in
-                                    self?.fetchTabBarItems()
-            },
-                                failure: { [weak self] error in
-                                    let actualError = error ?? NSError(domain: WordPressComRestApiErrorDomain, code: -1, userInfo: nil)
-                                    DDLogError(ReaderTopicsConstants.remoteFetchError + actualError.localizedDescription)
-                                    self?.fetchTabBarItems()
+
+        let dispatchGroup = DispatchGroup()
+
+        // Sync the reader menu
+        dispatchGroup.enter()
+        service.fetchReaderMenu(success: {
+            dispatchGroup.leave()
+        }, failure: { (error) in
+            let actualError = error ?? NSError(domain: WordPressComRestApiErrorDomain, code: -1, userInfo: nil)
+            DDLogError(ReaderTopicsConstants.remoteFetchError + actualError.localizedDescription)
+
+            dispatchGroup.leave()
         })
+
+        // Sync the followed sites
+        dispatchGroup.enter()
+        service.fetchFollowedSites(success: {
+            dispatchGroup.leave()
+        }, failure: { (error) in
+            let actualError = error ?? NSError(domain: WordPressComRestApiErrorDomain, code: -1, userInfo: nil)
+            DDLogError(ReaderTopicsConstants.remoteFetchError + actualError.localizedDescription)
+
+            dispatchGroup.leave()
+        })
+
+        // Wait for both the requests to finish
+        dispatchGroup.notify(queue: .main) { [weak self] in
+            self?.fetchTabBarItems()
+        }
     }
 
     private enum ReaderTopicsConstants {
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
```diff
@@ -86,14 +86,35 @@ extension ReaderTabItemsStore {
...
-                                    self?.fetchTabBarItems()
+
+        let dispatchGroup = DispatchGroup()
+
+        // Sync the reader menu
+        dispatchGroup.enter()
+        service.fetchReaderMenu(success: {
+            dispatchGroup.leave()
+        }, failure: { (error) in
+            let actualError = error ?? NSError(domain: WordPressComRestApiErrorDomain, code: -1, userInfo: nil)
```
Gio2018, line 97: This needed fixing before your changes: can we use ReaderTopicsConstants.remoteServiceError instead of creating the default error here and in `fetchFollowedSites` below?

emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14370/commits/7c7ff3010eeb891704b011d52c6f1d8901afe380

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
```diff
@@ -86,14 +86,35 @@ extension ReaderTabItemsStore {
...
+            dispatchGroup.leave()
+        })
+
+        // Sync the followed sites
+        dispatchGroup.enter()
+        service.fetchFollowedSites(success: {
+            dispatchGroup.leave()
+        }, failure: { (error) in
+            let actualError = error ?? ReaderTopicsConstants.remoteServiceError
+            DDLogError(ReaderTopicsConstants.remoteFetchError + actualError.localizedDescription)
```
yaelirub, line 109: Shouldn't this error logging be something like `DDLogError("Could not sync sites: \(String(describing: error))") ?

emilylaguna: The `ReaderTopicsConstants.remoteFetchError` maps to a string `Error syncing menu:`.

I updated the error logging to reflect the errors `ReaderMenuViewController` was using and switch to `String(describing:)`

PR #14386, opened 2020-06-25
ULS: Use unified navigation bar

ScoutHarris:
Ref #https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/182
Auth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/311

To test:

- Enable the `unifiedSiteAddress` feature flag.
- Run the app. Log out if necessary.
- Select Log In > Site Address.
- Verify the nav bar looks like this: (details are noted on the Auth PR)

<kbd>new_nav
</kbd>

- Go to any other flow.
- Verify the nav bar looks like the old style, i.e:

<kbd>old_nav</kbd>

PR submission checklist:


mindgraffiti:
@Gio2018 @ScoutHarris 

Ref. pbArwn-C9-p2

If it's a high effort task to change the back buttons from blue to .text color, I'm not sure it's worth putting in the effort. The Large Titles project uses blue navbar items, so we'd use the .text color only temporarily.

ScoutHarris:
Hey @Gio2018 .

My apologies, I didn't explain the "partially styled" bit very well. The blue 'Back' button on Site Address is correct. The unified nav bar style will have all the nav buttons blue. This is the design:

Screen Shot 2020-06-26 at 11 40 07 AM

As noted on the Auth PR I haven't done these:
- The 'Help' button is still white (on the white background, you may not see it, but it's there).
- The app icon is still displayed (it's white, so you may not see it, but it's there).

So, right now, the unified nav bar should look like you noticed.

Except on the My Site page. 'Switch Site' shouldn't be blue, which I'll ping you about directly as I don't know how you accomplished that. 😄

ScoutHarris:
@Gio2018 and I worked out why 'Switch Site' was blue.

He first accessed Site Address, which turned the button blue. He then logged in with Apple. Per his account settings, he didn't have to enter any information after the Apple creds (i.e. no other Auth view was displayed). Thus the button color was never reset back to white. Right now, since Site Address is disabled, it will never be set to blue, so this won't happen in the live app. However, it will be a valid issue when Site Address is enabled, so it is most definitely something I need to fix. But it can be done later.

Issue #14396, opened 2020-06-29
Reader Improvements: Select Interests Main Issue


emilylaguna:
Parent Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14395

<img  alt="Screen Shot 2020-06-29 at 12 33 55 PM" src="https://user-images.githubusercontent.com/793774/86048115-d090af80-ba04-11ea-8121-d75ef0fb31cc.png" width="50%">

### Goal
The user is presented with a screen that displays a list of interests for them to select that will allow them to fine tune the posts displayed to the on the Discover feed.

This screen should be shown only to users who don't follow any tags (new users or users who haven't used Reader much).

### Designs
[Figma](https://www.figma.com/file/NuCzgHcTmaW0YkGpX3YKD0/%F0%9F%93%90-Interests-Picker-Screen-(Specs)?node-id=0%3A1)

### Back Log
- [ ] Design review

### Done
- [x] Create a custom collection view layout for the interests selection: https://github.com/wordpress-mobile/WordPress-iOS/pull/14389
- [x] Create a new view controller for select interests view: https://github.com/wordpress-mobile/WordPress-iOS/pull/14402
- [x] Add new API call to pull down the list of interests from the server: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/257 and https://github.com/wordpress-mobile/WordPress-iOS/pull/14408
- [x] Handle selection view state: https://github.com/wordpress-mobile/WordPress-iOS/pull/14459
- [X] RTL Support: https://github.com/wordpress-mobile/WordPress-iOS/pull/14389
- [x] Update API to remove `-en` from the slug: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/259 and https://github.com/wordpress-mobile/WordPress-iOS/pull/14469
- [x] Write logic to determine when to display the new interests picker: https://github.com/wordpress-mobile/WordPress-iOS/pull/14476
- [x] Save the selected interests when tapping the next button: https://github.com/wordpress-mobile/WordPress-iOS/pull/14487
- [x] Write logic to present the interests view above the reader tabs below the tab view controller, see Android implementation
- [x] Add Error Handling
- [x] iPad Support
- [x] Accessibility
- [X] Add Tracking
- [x] Polish the UI

PR #14402, opened 2020-06-30
Reader Improvements: Add select interests view controller

emilylaguna:
Main Issue: #14396

- Initial view controller creation and design. 
- This also includes some a mock data source to use for testing, and some mock json files.

### Screenshots
| Light Mode | Dark Mode |
|---|---|
|Simulator Screen Shot - iPhone 11 Pro (13 3) - 2020-06-30 at 13 39 27|Simulator Screen Shot - iPhone 11 Pro (13 3) - 2020-06-30 at 13 39 28|

### To test:
**Note:** This isn't accessible in the app yet.

To test you can trigger the view by adding the following line to the end of the `viewDidAppear` method in the `ReaderStreamViewController`
```
        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500)) {
            self.navigationController?.present(ReaderSelectInterestsViewController(), animated: true)
        }
```

Then tap on the Reader tab to trigger the view.

### PR submission checklist:


WordPress/Classes/ViewRelated/Reader/Select Interests/Mock Data/InterestsDataSource.swift
```diff
@@ -0,0 +1,52 @@
...
+class InterestsDataSource {
+    private(set) var count: Int = 0
+    private(set) var interests: [ReaderInterest] = []
+
+    init(fileName: String) {
+        parseJSON(fileName: fileName)
+    }
+
+    private func parseJSON(fileName: String) {
+        guard
```
leandroalonso, line 24: Since this is a brand new code I highly suggest using [Decodable](https://developer.apple.com/documentation/swift/decodable) to parse JSON, we'll be able to remove a loot of lines here.

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsCollectionViewFlowLayout.swift
```diff
@@ -75,7 +75,7 @@ class ReaderInterestsCollectionViewFlowLayout: UICollectionViewFlowLayout {
 
         // Update content size
         contentSize.width = maxContentWidth
-        contentSize.height = currentRow * (cellHeight + itemSpacing) + contentInsets.top + contentInsets.bottom
+        contentSize.height = (currentRow+1) * (cellHeight + itemSpacing) + contentInsets.top + contentInsets.bottom
```
leandroalonso, line 78:
```suggestion
        contentSize.height = (currentRow + 1) * (cellHeight + itemSpacing) + contentInsets.top + contentInsets.bottom
```

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        return cell
+    }
+}
+
+// MARK: - UICollectionViewDelegate
+extension ReaderSelectInterestsViewController: UICollectionViewDelegate {
+    func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        let interest: ReaderInterest = fakeDataSource.interest(for: indexPath.row)
+        interest.isSelected = !interest.isSelected
```
leandroalonso, line 85: When I tap an interest the background and the text color changes in different speeds (first the background, then the text) — and for a brief moment I can see black text in the dark blue background. Do you think this is an easy fix?

WordPress/Classes/ViewRelated/Reader/Select Interests/Mock Data/InterestsDataSource.swift
```diff
@@ -0,0 +1,52 @@
...
+    let title: String
+    let slug: String
+
+    init(title: String, slug: String) {
+        self.title = title
+        self.slug = slug
+    }
+}
+
+class InterestsDataSource {
```
leandroalonso, line 15: I'm not sure if you pretend to use `InterestsDataSource` in the final implementation, but we can add some unit tests here to make sure that given a JSON this outputs the correct list of interests.

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        super.viewDidLoad()
+
+        configureCollectionView()
+        applyStyles()
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        let layout = collectionView.collectionViewLayout as! ReaderInterestsCollectionViewFlowLayout
```
yaelirub, line 33: Can we guard instead of forced unwrapping here ?

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// MARK: - UICollectionViewDataSource
+extension ReaderSelectInterestsViewController: UICollectionViewDataSource {
+    func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
+        return fakeDataSource.count
+    }
+
+    func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        let cell: ReaderInterestsCollectionViewCell = collectionView.dequeueReusableCell(withReuseIdentifier: Constants.reuseIdentifier, for: indexPath) as! ReaderInterestsCollectionViewCell
```
yaelirub, line 67: I prefer guard instead of forced unwrapping

yaelirub:
Nice work, @emilylaguna ! 
One general question: Why is there no localization anywhere?

WordPress/Classes/ViewRelated/Reader/Select Interests/Mock Data/InterestsDataSource.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+class ReaderInterest {
```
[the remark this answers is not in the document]
emilylaguna: I agree, but since this code is just for testing purposes and is going to be removed once the network side of things are done later this week I'm going to leave it as it is.

WordPress/Classes/ViewRelated/Reader/Select Interests/Mock Data/InterestsDataSource.swift
emilylaguna: This is going to be removed in the future.

emilylaguna:
> Nice work, @emilylaguna !
> One general question: Why is there no localization anywhere?

That's a good question, I was thinking I would do it in future task but there isn't a reason why I shouldn't do it in this one. I've added it in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14402/commits/88c19b1ddc16629bb21e763370c7d81851d0aae9

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14402/commits/f6de145dc492798e13896f7c9f26585a834d933f

emilylaguna:
> By [checking the original designs](https://www.figma.com/file/NuCzgHcTmaW0YkGpX3YKD0/%F0%9F%93%90-Interests-Picker-Screen-(Specs)?node-id=194%3A1177) I think the title is way too close to the description (and maybe the description from the buttons?)

When I measure using iOS 13 compared to the design specs everything aligns correctly, there's a 24px gap between the title and subtitle, and a 16px gap between the subtitle and the first row. 

I do see some slight reduction in the spacing in your screenshot on iOS 12 but it looks to be due to the difference in the font (New York on iOS 13 and Noto for iOS 12) compared to iOS 13. 

Thoughts?

Screen Shot 2020-07-01 at 11 42 57 AM


> In iOS 12/11 the background below the button is grey:

This is because the designs call for the use of semantic colors, and the button container view background is using the [tertiary system background](https://developer.apple.com/documentation/uikit/uicolor/3173154-tertiarysystembackground) color which is iOS 13 and above. On iOS 12 and below [we default it](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Extensions/Colors%20and%20Styles/UIColor%2BMurielColors.swift#L122) to: `.neutral(.shade10)`.

I'm not sure we should add a custom reader interests only tertiary background color to enforce it to always be white, thoughts?

leandroalonso:
> I do see some slight reduction in the spacing in your screenshot on iOS 12 but it looks to be due to the difference in the font (New York on iOS 13 and Noto for iOS 12) compared to iOS 13.

Hmm, I had this feeling while testing in iOS 13 (iPhone 11) too.

Anyway, as we talked, let's handle that in a future UI polishment task.

> This is because the designs call for the use of semantic colors, and the button container view background is using the tertiary system background color which is iOS 13 and above. On iOS 12 and below we default it to: .neutral(.shade10).

Having a dynamic color that is different between iOS 12- and iOS 13+ seems like an issue, don't? If something in light mode in iOS 13 is white, it should be white in iOS 12 or below too.

emilylaguna:
I'd like to add another design difference between the dark mode color in the design and the dark mode color used in the app.

The design calls for selected background color to be blue 40 which is what is being used in the app as well. But the color that's defined in the `ColorPalette.xcassets` file has its own definition of blue 40 for light and dark modes:

Screen Shot 2020-07-01 at 12 03 44 PM

which is why it appears slightly lighter on dark mode.

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift
emilylaguna: That's weird, the background color isn't animating and just changing which the text color is animating at the right rate. Not entirely sure why at this point, I just disabled the fade animation for now: https://github.com/wordpress-mobile/WordPress-iOS/pull/14402/commits/bda0ae794a160eeaac858a2a9dfaab2086095aa8

emilylaguna:
> Having a dynamic color that is different between iOS 12- and iOS 13+ seems like an issue, don't? If something in light mode in iOS 13 is white, it should be white in iOS 12 or below too.

I agree, for now I've changed the background color to white for iOS 12 and below

PR #14439, opened 2020-07-08
Reader Improvements: Displays Select Interests using FeatureFlag, adds real network calls

emilylaguna:
Main Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14396

WordPressKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/258

- Displays the select interests view if the FeatureFlag is enabled when opening the reader tab
- Updates the select interests view to use the real network calls

#### 01 - Test feature flag
1. Launch the app
2. Tap on the reader tab
3. Confirm you are not presented the select interests view
4. Tap on the My Site tab
5. Enable the `Reader Improvements Phase 2` feature flag
6. Follow the steps in test 02

#### 02 - Test presenting of the view
Make sure the feature flag is enabled:

1. Launch the app
2. Tap on the Reader tab
3. Wait a second
4. You should be presented with the select interests view

Note: You can dismiss on iOS 13 by dragging the view down

#### 03 - Test interests network
1. Follow the steps from test 02
2. Confirm the interests are displayed in your devices language

#### 04 - Test localization of network interests
1. Change your device language to another language
2. Follow the steps from test 02
3. Confirm the interests are localized to the device language
4. If the language is a right to left language, confirm the interests are aligned to the right

#### 05 - Test slow loading
1. Throttle your internet
1.a alternatively you can add a delay to the `ReaderSelectInterestsViewController refreshData` method as below:

```
    private func refreshData() {
        activityIndicatorView.startAnimating()

        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
            self.dataSource.reload()
        }
    }
```
2. Follow the steps from test 02
3. Confirm:
3.a  You see the activity indicator
3.b. The interests load in successfully
3.c The activity indicator is no longer visible

### PR submission checklist:


WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    var selectedInterests: [ReaderInterestViewModel] {
+        return interests.filter({ $0.isSelected })
+    }
+
+    private var interestsService: ReaderInterestsService
+
+    /// Creates a new instance of the data source
+    /// - Parameter topicService: An Optional `ReaderTopicService` to use. If this is `nil` one will be created on the main context
+    init(service: ReaderInterestsService? = nil) {
```
leandroalonso, line 57: Another way to write that would be:

```swift
init(service: ReaderInterestsService? = ReaderTopicService(managedObjectContext: ContextManager.sharedInstance().mainContext)) {
    self.interestsService = service
```

emilylaguna: I had it like that at first but I didn't like how long the line was so I broke it up

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
```diff
@@ -0,0 +1,91 @@
...
+        }) { (error: Error) in
+            DDLogError("Error: Could not retrieve reader interests: \(String(describing: error))")
+
+            self.interests = []
+        }
+    }
+
+    /// Toggles the selected state for the item at the specified row
+    /// - Parameter row: The row of the interest you want to toggle
+    public func toggleSelected(for row: Int) {
```
yaelirub, line 81: Shouldn't the public interface for this toggling be done through the viewModel? It seems kinda off to me that this would be done through the data source . wdyt?

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
```diff
@@ -0,0 +1,91 @@
...
+            return
+        }
+
+        self.interestsService = service
+    }
+
+    /// Fetches the interests from the topic service
+    public func reload() {
+        interestsService.fetchInterests(success: { interests in
+            self.interests = interests.map({ ReaderInterestViewModel(interest: $0)})
```
yaelirub, line 71: You don't need the "( )" when you're using $0 :) so this could be: 
```interests.map{ ReaderInterestViewModel(interest: $0)}```

File: WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+
+            return
+        }
+
+        self.interestsService = service
+    }
+
+    /// Fetches the interests from the topic service
+    public func reload() {
+        interestsService.fetchInterests(success: { interests in
```
yaelirub, line 70: Also, I know the closures are escaping but I'd prefer using [weak self] for both closures.

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// MARK: - ReaderInterestsDataSource
+class ReaderInterestsDataSource {
+    var delegate: ReaderInterestsDataDelegate?
+
+    private(set) var count: Int = 0
+    private(set) var interests: [ReaderInterestViewModel] = [] {
+        didSet {
+            count = interests.count
```
yaelirub, line 43: What is this count used for?

yaelirub:
@emilylaguna , wonderful work here! I left a few comments in code but a couple of other things:
 1. My Zombie checker is enabled and I get these errors:
```Class _NSZombie_WordPress.ReaderTagTopic_ReaderTagTopic_ is implemented in both ?? (0x7b0c006b3d90) and ?? (0x7b0c0066cf00). One of the two will be used. Which one is undefined.```

```objc[40954]: Class _NSZombie_WordPress.ReaderTeamTopic_ReaderTeamTopic_ is implemented in both ?? (0x7b0c0068b470) and ?? (0x7b0c0066d5c0). One of the two will be used. Which one is undefined.```

2. For some reason the word business was not translated :
Screen Shot 2020-07-08 at 5 30 22 PM

File: WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift [hunk 6]
```diff
@@ -23,16 +23,20 @@ class ReaderSelectInterestsViewController: UIViewController {
...
 
-    // MARK: - Mock Data Source
-    private let fakeDataSource = InterestsDataSource(fileName: "interests.json")
+    // MARK: - Data
+    private let dataSource: ReaderInterestsDataSource = ReaderInterestsDataSource()
 
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        dataSource.delegate = self
```
yaelirub, line 34: 🤖  DANGER 🤖  this is a retain cycle.
You should declare the delegate in the data source as weak but before you change it:
* Put a breakpoint in `BlogDetrailsViewController viewDidAppear` 
* Display select interests screen 
* Dismiss select interest screen
* Go to "My Site". Would hit breakpoint
* Tap on Debug Memory Graph icon and see the retain cycle that leaks the data source, view models and view controller
Screen Shot 2020-07-08 at 5 54 35 PM

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
leandroalonso: Or `[unowned self]` if you're confident that `self` will exist when the block is called. :)

emilylaguna:
> 1. For some reason the word business was not translated :

Looks like this an issue on the API side of things, you can see the translation in the JSON response is still in English: 

- https://public-api.wordpress.com/wpcom/v2/read/interests?_locale=he
Screen Shot 2020-07-09 at 8 31 00 AM

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
emilylaguna: Good plan!

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
emilylaguna: It keeps track of how many interests there are and is used in the `numberOfItemsInSection` method. 

The idea is that an external class doesn't access the interests array itself. 

To keep to that point I've made the interests array completely private to ensure the right methods are called.

emilylaguna:
> 1. `Class _NSZombie_WordPress.ReaderTagTopic_ReaderTagTopic_ is implemented in both ?? (0x7b0c006b3d90) and ?? (0x7b0c0066cf00). One of the two will be used. Which one is undefined.`
> 
> `objc[40954]: Class _NSZombie_WordPress.ReaderTeamTopic_ReaderTeamTopic_ is implemented in both ?? (0x7b0c0068b470) and ?? (0x7b0c0066d5c0). One of the two will be used. Which one is undefined.`

I just checked and I do see these as well, but I see them in develop so it doesn't look related to anything I changed specifically.

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderSelectInterestsViewController.swift
emilylaguna: D'oh.. 🤦‍♀️  Thanks for catching that!

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
emilylaguna: `[weak self]` it is!

WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsDataSource.swift
```diff
@@ -68,12 +68,12 @@ class ReaderInterestsDataSource {
 
     /// Fetches the interests from the topic service
     public func reload() {
-        interestsService.fetchInterests(success: { interests in
-            self.interests = interests.map({ ReaderInterestViewModel(interest: $0)})
-        }) { (error: Error) in
+        interestsService.fetchInterests(success: { [weak self] interests in
+            self?.interests = interests.map { ReaderInterestViewModel(interest: $0)}
```
yaelirub, line 72: super nitpicky but interests.map { ReaderInterestViewModel(interest: $0)} missing a space at the end between ) and }. I may add the space myself instead of bothering you with this

yaelirub:
@emilylaguna , if this gets merged tomorrow it can go to 15.2 just sayin ;)

PR #14443, opened 2020-07-08
Fix basic offline support in Reader

leandroalonso:
This PR fixes the basic offline suport in the new Reader detail.

### To test

1. Open the Reader
2. Save a few posts 
3. Go offline
4. Open the saved posts, they should show the cached images

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderCSS.swift
+++ b/WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderCSS.swift
@@ -29,7 +29,6 @@ struct ReaderCSS {
             return url(appendingTimestamp: now)
         }
 
-        // If the last time we fetched the CSS was 2 days ago and the user is online, refresh it
         return url(appendingTimestamp: lastUpdated)
     }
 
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderCSS.swift
```diff
@@ -21,8 +21,8 @@ struct ReaderCSS {
     ///
```
emilylaguna, line 21: The comments reference a 2 day update period but we've changed to 5. What about changing to reference the `expirationDays` variable?

leandroalonso:
@emilylaguna ready for another look.

I implemented unit tests for 3 different scenarios (each of 'em when online and offline). I guess the logic is now correct.

WordPress/Classes/ViewRelated/Reader/DetailWebView/ReaderCSS.swift
```diff
@@ -6,23 +6,25 @@ import Foundation
...
 
     private var expirationDaysInSeconds: Int {
         return expirationDays * 60 * 60 * 24
     }
 
-    private let updatedKey = "ReaderCSSLastUpdated"
+    static let updatedKey = "ReaderCSSLastUpdated"
 
     /// Returns the Reader CSS appending a timestamp
     /// We force it to update every 2 days
```
emilylaguna, line 22:
```suggestion
    /// We force it to update based on the `expirationDays` property
```

leandroalonso:
@emilylaguna done!

PR #14448, opened 2020-07-09
Remove spotlight from site title and site icon after navigating away of My Site

### LINKED ISSUE #14441: Quick Start: Blue Indicator shown on My Site without notice
### Expected behavior

Blue indicator should always be accompanied by a notice to describe the current tutorial action.

### Actual behavior

The blue indicator is sometimes shown without the notice.

* Enter the "Customize Your Site" screen from your new site
* Tap the "Upload a site icon"
* Tap on any cell in the table view
* Go back to the My Site screen

<a href="https://user-images.githubusercontent.com/3250/86966288-1f85c580-c126-11ea-816e-15fd5a096f0d.png"></a>
##### Tested on iPhone SE Simulator, iOS 13.5, WPiOS from: https://github.com/wordpress-mobile/WordPress-iOS/pull/14422

Gio2018:
Fixes #14441

To test:

* Enter the "Customize Your Site" screen from your new site
* Tap the "Upload a site icon" or "Set your site title"
* Tap on any cell in the table view
* Go back to the My Site and verify that the spotlight does not show anymore

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -177,7 +177,7 @@ open class QuickStartTourGuide: NSObject {
             return
         }
         if element != currentElement {
-            let blogDetailEvents: [QuickStartTourElement] = [.blogDetailNavigation, .checklist, .themes, .viewSite, .sharing]
+            let blogDetailEvents: [QuickStartTourElement] = [.blogDetailNavigation, .checklist, .themes, .viewSite, .sharing, .pages, .stats, .plans]
             let readerElements: [QuickStartTourElement] = [.readerTab, .readerSearch]
 
             if blogDetailEvents.contains(element) {
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -177,7 +177,7 @@ open class QuickStartTourGuide: NSObject {
             return
         }
         if element != currentElement {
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
```diff
@@ -177,7 +177,7 @@ open class QuickStartTourGuide: NSObject {
             return
         }
         if element != currentElement {
-            let blogDetailEvents: [QuickStartTourElement] = [.blogDetailNavigation, .checklist, .themes, .viewSite, .sharing]
```
frosty, line 180: What's the reason for adding these extra events here?

I'm not sure if this is an issue, but makes a slight change to the behaviour of the pages tour:

* Start the Pages tour
* Tap into the Pages screen
* Before this change, if you tap back the notice would stay visible and you could tap back into Pages and the notice would still be there.
* With this change, the notice disappears if you tap back into Pages.

Gio2018: Thanks for pointing this out @frosty ! I think I had misinterpreted the usage of these events, thinking that just tapping on the initial action should dismiss the notice (and the spotlight), while instead, in the case of pages, it's the + that dismisses the notice. This came by observing that when I tap on "Blog Posts", it would dismiss any notice and spotlight, including the notice on the FAB that pops up from "Grow your audience -> Publish a post".
So maybe I should revert these changes and then consider if the blog post needs changes, what do you think?

Gio2018: Went ahead and did the change, we can adjust it further in future PRs

PR #14449, opened 2020-07-09
[TESTING] Columns percentage

lukewalczak:
This PR is created only for testing columns percentage via peril build.

🚧 DO NOT MERGE 🚧

To test:

PR submission checklist:


iamthomasbishop:
Thanks for sharing a build, @lukewalczak ! I tested using the build above (31285) and tried to jot down some thorough notes, split into setup vs. editing. I think there's going to be at least a couple of rounds of design review but I'll try to gather as succinctly as possible.
 
## Setup

- The margins on the left/right of the sheet look a bit narrow (8px?). Can we bump them up a little bit, perhaps to the same width as the gutters between layout preset buttons? ([screenshot](https://cloudup.com/cFBx7aXRB-y))
- Can we change the title of the layout selector sheet to a more approachable label like `Select a layout`? (seen in the screenshot above)
- Would it be possible to smoothly transition from the inserter sheet to the layout selection sheet when Columns is chosen from the inserter — similar to the sub-sheet transitions on the color picking flow? If that's possible, the layout selection sheet (like the color sub-sheets) would have a `< Back` button on iOS and a back arrow (`←`) on Android.
- If we aren't able to do the last suggestion ☝️, we might want to have a "cancel" button in the header of the bottom sheet. This could be a red text label `Cancel` on iOS and a gray `X` on Android. It might also make sense for tapping on the scrim to perform the cancel action.
- When a layout has first been applied, I think we should probably select the parent Columns block first

## Editing

- Column blocks don't seem to be always getting the proper starting widths. For example, if I choose the `33/33/33` layout, my actual column widths are 50/50/50 (as seen in the block settings sheet).
- Related to the last point, when applying the `30/70` or `70/30` blocks, they're actually set to 33.3/66.6 and vice versa. The `50/50` layout option seems correctly applied.
- If I adjust the width of a column — let's say Column A — it appears that the slider values in the other columns' settings aren't getting updated. They do seem to be adjusting appropriately on the front-end. To reproduce: start with the `33/33/33` layout, edit one column to something like 15%, then preview. You'll notice that the column appears correctly on the front-end, but when you look at the other columns' widths on the block settings, they're still at their original (incorrect) value.
- I'm seeing some weird output on the front-end when I use more than 3 columns. It's happening on multiple themes, so I think it's related to our side (screenshots: [mobile editor](https://cloudup.com/cJhNgh8R21u) / [front-end](https://cloudup.com/cWwkGFpkC-T))
- I think we can round slider values to whole numbers, instead of decimal (e.g. 33% vs. 33.3333%)

lukewalczak:
## Setup

> - The margins on the left/right of the sheet look a bit narrow (8px?). Can we bump them up a little bit, perhaps to the same width as the gutters between layout preset buttons? ([screenshot](https://cloudup.com/cFBx7aXRB-y))
Done! ✅ 

> - Can we change the title of the layout selector sheet to a more approachable label like `Select a layout`? (seen in the screenshot above)
Done! ✅ 

> - If we aren't able to do the last suggestion ☝️, we might want to have a "cancel" button in the header of the bottom sheet. This could be a red text label `Cancel` on iOS and a gray `X` on Android. It might also make sense for tapping on the scrim to perform the cancel action.
Done! ✅ your first suggestion would have required more structure changes related to interfering to an inserter.

> - When a layout has first been applied, I think we should probably select the parent Columns block first
Done! ✅ 

## Editing

> - Column blocks don't seem to be always getting the proper starting widths. For example, if I choose the `33/33/33` layout, my actual column widths are 50/50/50 (as seen in the block settings sheet).
Done! ✅ 

> - Related to the last point, when applying the `30/70` or `70/30` blocks, they're actually set to 33.3/66.6 and vice versa. The `50/50` layout option seems correctly applied.
It's the logic which we share with the web. I think we have to keep that.

> - If I adjust the width of a column — let's say Column A — it appears that the slider values in the other columns' settings aren't getting updated. They do seem to be adjusting appropriately on the front-end. To reproduce: start with the `33/33/33` layout, edit one column to something like 15%, then preview. You'll notice that the column appears correctly on the front-end, but when you look at the other columns' widths on the block settings, they're still at their original (incorrect) value.

Please let  me know if it's fixed ✌️ 

> - I think we can round slider values to whole numbers, instead of decimal (e.g. 33% vs. 33.3333%)

I would like to do that, but on web you can set the percentage with one decimal. That change can introduce some issues with syncing values.

iamthomasbishop:
@lukewalczak Nice work on the fixes, it's looking a bunch better. Here are some updated notes based on the new build.

## Layout Picker Sheet

- I think the padding on the left of the horizontal scroll container still a little bit off ([screenshot](https://cloudup.com/cxUI5wE2-DF))
- Can we either remove ~8px margin between the horizontal scroll container and the footer message? ([screenshot](https://cloudup.com/cAex3RcbXgk)). Note: If doing this will cause conflicts w/ the base margin/paddings on bottom sheets, we can probably leave it as-is
- When I duplicate a Columns block (by tapping `•••` then `Duplicate` ), it’s showing the layout selection sheet
- The cancel button feels solid in the flow, nice work

## Editing

> It's the logic which we share with the web. I think we have to keep that.

Ok, fair enough, that’s really confusing on *web* then 🙂 I can’t help but feel we should change the label so it’s accurate. Perhaps something like `one-third / two-thirds`, `1/3 - 2/3`, or `33/66`? Not sure.

I’m also still seeing the issue where adjusting one column doesn’t affect the percentage of the other(s) in their block settings. My test case was: created a 2-columns 50-50 block, then edited the first column to 70% (saw no % value change in the second column’s settings).

## Other

- I'm seeing the bug again where removing nested children causes a crash
- I was seeing a weird bug where the layout picker was only showing when I add my first Columns block (not on subsequent tries) but I'm not able to reproduce it every time and I think re-installing the build fixed it

lukewalczak:
## Layout Picker Sheet

> - I think the padding on the left of the horizontal scroll container still a little bit off ([screenshot](https://cloudup.com/cxUI5wE2-DF))


> - Can we either remove ~8px margin between the horizontal scroll container and the footer message? ([screenshot](https://cloudup.com/cAex3RcbXgk)). Note: If doing this will cause conflicts w/ the base margin/paddings on bottom sheets, we can probably leave it as-is

Fixed both of them, please look at the screenshot and let me know if both are correct:

before | after
--- | ---
Screenshot 2020-07-14 at 12 33 06 | Screenshot 2020-07-14 at 12 33 14

 
> - When I duplicate a Columns block (by tapping `•••` then `Duplicate` ), it’s showing the layout selection sheet

Working on that, thanks for reporting that! 

> - The cancel button feels solid in the flow, nice work

Thanks!

## Editing

> Ok, fair enough, that’s really confusing on *web* then 🙂 I can’t help but feel we should change the label so it’s accurate. Perhaps something like `one-third / two-thirds`, `1/3 - 2/3`, or `33/66`? Not sure.

Labels and block icons we are sharing with a web as well, it can be confusing for the user to have different labels on mobile and web (_at least in that case_), however, I understand that values should correspond to labels.
I asked about that naming on core-editor slack and the answer is:

> “Thirds” — I assume just a shorthand to communicate this. I don’t have more context

I think we can accept that, wdyt @pinarol?

> I’m also still seeing the issue where adjusting one column doesn’t affect the percentage of the other(s) in their block settings. My test case was: created a 2-columns 50-50 block, then edited the first column to 70% (saw no % value change in the second column’s settings).

Should the value of the second column change? I think we've decided to not interfere in other columns values. If I understand your vision wrongly, please describe how it should work :v:

## Other

> - I'm seeing the bug again where removing nested children causes a crash
> - I was seeing a weird bug where the layout picker was only showing when I add my first Columns block (not on subsequent tries) but I'm not able to reproduce it every time and I think re-installing the build fixed it

If you are facing crashes please try to describe some steps to let me reproduce it and dive into issue

iamthomasbishop:
> Fixed both of them, please look at the screenshot and let me know if both are correct:

Looks much better, thank you!

> I asked about that naming on core-editor slack and the answer is: “Thirds”

"Thirds" would make more sense to me.

> Should the value of the second column change? I think we've decided to not interfere in other columns values. If I understand your vision wrongly, please describe how it should work ✌️

Yea, I think it feels like a bug in practice, when you change the value of one column and there is no change in the others. I understand there was a lot of discussion and complexity around how these work, but I thought I'd raise it because I think it'll feel like a bug to users.

> If you are facing crashes please try to describe some steps to let me reproduce it and dive into issue

I can't reproduce the crash (a new build might have fixed it?), but here's what I was doing to facilitate the crash:

1. Add a Group block
2. Within the Group, add a Columns block
3. Within one of the Column block, add some other block
4. Try to remove that nested "other" block
5. Observe crash

Note: I think the crash was also happening when deleting a Column after step 2, but I can't reproduce at the moment.

iamthomasbishop:
@lukewalczak Actually one tiny thing I noticed on second glance of the screenshots you attached in your last comment -- can we reduce the horizontal margins between the items slightly so that the fourth item in the list peeks out of the right edge? This should help to hint that the container is scrollable.

lukewalczak:
> "Thirds" would make more sense to me.

Just to clarify, I understand their answer as an explanation of how they perceive labels `30/70` and `70/30` according to their values, and we have to rather keep them (labels and values) without changes.

> Yea, I think it feels like a bug in practice, when you change the value of one column and there is no change in the others. I understand there was a lot of discussion and complexity around how these work, but I thought I'd raise it because I think it'll feel like a bug to users.

I see what you mean, but since on web each column can have 100% width, it's reasonable that changing a value in one column does not affect the other. In order to provide users with the same editing options, _currently_, we need to accept this solution.

Thanks for the feedback! I will do some styling tweaks, check the crash and produce new build asap.

iamthomasbishop:
> I see what you mean, but since on web each column can have 100% width, it's reasonable that changing a value in one column does not affect the other. In order to provide users with the same editing options, currently, we need to accept this solution.

@lukewalczak Yea, this is one of those confusing things on the web 😆 Let's roll with it for now. I have some thoughts on how we can improve the usability despite this constraint.

## Blind spots while editing

Considering our constraints and how it's hard to see the effects when you're editing, I think we might want to consider adding some visual cues near where you edit the widths. Perhaps we could show a little "preview" of the row that visualizes the composition of the row. Here are [some examples](https://cloudup.com/cYw3ev9tZP4).

Another thing that would probably help this, which can probably be part of a future iteration — is if we could surface all percentage width controls on the parent level Columns block in addition to each child. This will present some other challenges so we can avoid it for now, but I think this would be nice in the future.

## Column Widths and Viewport Size

As I test on my iPad Pro, I'm becoming more convinced that forcing equal-width columns on all viewport sizes is a bit limiting. While our content container is narrow (~580px), I think we might want to display columns *proportionally* — specifically on wider viewports like tablet in landscape orientation (although split-screen can complicate things a bit, I imagine we should be able to scope out what the application's viewport width is at?). 

Something in the neighborhood of web's breakpoint of ~782px would be a good starting point, although if we're not required to match that, I think we can go lower. The idea here would be to find a nice breakpoint that will support proportional column display for typical widths of tablet full-screen and *wide* split-screen, then defer to our standard equal-width display on smaller breakpoints.

Does that make sense? I think this will become a lot easier when we implement wide and full-width containers, but I think this will be beneficial regardless.

pinarol:
> While our content container is narrow (~580px), I think we might want to display columns proportionally — specifically on wider viewports like tablet in landscape orientation

@iamthomasbishop Would you mind elaborating on that a bit? Do you mean we should change our container width to a more dynamically calculated value rather than having it fixed to 580px? If yes then we should know the rules for this calculation and preferably handle it on a different PR(even if this means we might need to freeze columns percentage work). Having a breakpoint at ~782px is a different topic and can be done after we change the width calculations.

lukewalczak:
> Another thing that would probably help this, which can probably be part of a future iteration — is if we could surface all percentage width controls on the parent level Columns block in addition to each child. This will present some other challenges so we can avoid it for now, but I think this would be nice in the future.

I would like to have that feature as well, but currently, we are facing the issue with slider within the scrollable container, where it's easy to change the slider value unexpectedly when scrolling, so firstly we have to solve that issue.

> Considering our constraints and how it's hard to see the effects when you're editing, I think we might want to consider adding some visual cues near where you edit the widths. Perhaps we could show a little "preview" of the row that visualizes the composition of the row. Here are some examples.

Based on the examples I assume that "preview" will present the current column width in the context of other columns. It looks nice for 2-3 columns, however what about a case where we'll have ~6 columns - what should be the _width_ of that preview? Correct me please if I understand it wrongly ✌️ 

> While our content container is narrow (~580px), I think we might want to display columns proportionally — specifically on wider viewports like tablet in landscape orientation

To display columns proportionally like on the web, firstly we have to unblock our viewport to have the ability to display all the columns in one row. Btw (_probably I forgot the reason_) why on mobile we can have more than 6 columns where on the web the number is limited?

pinarol:
> Btw (probably I forgot the reason) why on mobile we can have more than 6 columns where on the web the number is limited?

There used to be such a limitation on web but now it is only in the settings, you can add many columns as you want using the inserter and you'll only see a warning in the settings indicating that output may start to suffer. So practically there's no real limit there as well.

lukewalczak:
> There used to be such a limitation on web but now it is only in the settings, you can add many columns as you want using the inserter and you'll only see a warning in the settings indicating that output may start to suffer. So practically there's no real limit there as well.

So it's another reason why we should have unblocked viewport to display columns' width proportionally.
@iamthomasbishop could please explain also how that _preview_ should work?

iamthomasbishop:
@pinarol @lukewalczak Per our discussions in Slack earlier, I wanted to comment on some clarifications and answer the latest questions y'all had.

## Columns Display WRT Breakpoints

As a starting point I wanted to clarify that my notes above regarding how we might want to display the column "proportionally" can be part of an upcoming iteration. I think we're in a good place for this specific iteration, but I wanted to note my concerns.

My suggestion above was about getting a little more granular for breakpoints ~768px+ by showing a layout more representative of the output at that screen/viewport size (which is what happens on the web-side already if you view on a tablet landscape mode). 

Let's use the example of a Columns block that has 3 columns — split `25` | `50` | `25`. With my suggestions, there would be no changes on smaller viewports -- we'd continue showing those columns as equal-width ( `33` | `33` | `33`) on "medium" size viewports (e.g. landscape phone) and would continue stacking columns on narrow (e.g. portrait phone). However, when the application viewport width is ≥ 768px, in this scenario we would lay out the columns as they would be on the web -- at their "natural" `25` | `50` | `25` proportional widths. Here is a sketch that hopefully helps to visualize this:

<a href="https://user-images.githubusercontent.com/1200432/88227020-93e65b80-cc32-11ea-8cc3-4451827e1118.jpeg"></a>

> I would like to have that feature as well, but currently, we are facing the issue with slider within the scrollable container, where it's easy to change the slider value unexpectedly when scrolling, so firstly we have to solve that issue.

@lukewalczak sounds good 👍

## Column Preview in Block Settings

> Based on the examples I assume that "preview" will present the current column width in the context of other columns. It looks nice for 2-3 columns, however what about a case where we'll have ~6 columns - what should be the width of that preview?

Good question! I'm not *super* worried about showing a large and very detailed preview — showing how the column fits into the general composition is probably enough to give a visual of how the column will look relative to the others (with the note about stacking directly below).

Let’s say there are 6 equal-width columns (each == 16.666%, or `x`), and you have the third one selected. The highlighted column width would be X and would have an offset of `x * 2` (bc it's the third column). It would look something like this (using a couple of options for layout...ignore the slider track position):

<a href="https://user-images.githubusercontent.com/1200432/88337202-eb9cc980-ccfb-11ea-95d3-8880dd093703.png"></a>

Let me know if you have any questions, happy to answer and discuss!

pinarol:
@iamthomasbishop so we are not talking about changing the content width but only the proportional-ness of the columns in different screen sizes, right? The content width on tablet and on the iphone landscape mode will be the same. 

Then my question is why don’t we equalize the behavior for the same content width? After all if the columns will be sharing the same space, we can bring the proportional columns into phone-landscape mode as well? I don’t understand the need for the different behavior for the same content size actually.
Web may be considering the screen width but there’s something quite different there. The empty space around the content gets smaller as the screen gets smaller on web, leaving the content larger space. So maybe they found this mechanism just enough, like no more fine tuning needed according to content width. But it starts to act poorly when I open the side bar, for example:
639D26C9-C382-46B7-9882-1A1A5A7F77F6

If we make the rules based on the content width then we won’t need to do extra changes in the future if we add more features like wide/full width. It’ll just adapt.


Another thing that needs clarification is, do we want to render all columns in 1 row? (assuming column count is bigger than 3)

iamthomasbishop:
> so we are not talking about changing the content width but only the proportional-ness of the columns in different screen sizes, right? The content width on tablet and on the iphone landscape mode will be the same.

@pinarol That's correct -- no changes to the current content container until we implement wide/full-width blocks.

> Then my question is why don’t we equalize the behavior for the same content width? After all if the columns will be sharing the same space, we can bring the proportional columns into phone-landscape mode as well? I don’t understand the need for the different behavior for the same content size actually.

My intention was to align with the web's behavior (lay out columns proportionally on larger _viewports_), but it probably  does make more sense to lay them out relative to its container width instead of viewport-width -- at least when the container is max ≤ 580px. With that, we now have to decide whether we want to lay columns out proportionally or equally (like we're currently doing) on landscape phone (and larger):
 
* Proportionally: this works fine when you have ≤ 3 columns, but gets very dense with more (or narrow) columns.
* Equal: this is what we have currently. It's more manageable and flexible but might not be as obvious to the user what the output will be (which is why I've proposed various workarounds for visualizing those details). Works well on "medium" breakpoints like landscape phone.

My proposal (for now, bc our container is 580px) would be to stick with our current equal-width/3-across mode. From there, once we add support for wider containers, we should explore displaying them proportionally in cases that the container is > 580. (There will probably be some caveats there, but I think this would set us up well.)

To summarize, does the following make sense moving forward:

* Stick with what we have now -- which is 3 columns per row, split equally inside a static 580px container
* When we add support for wider widths, we will allow columns to be displayed proportionally (provided the viewport is large enough to support > 580)

> do we want to render all columns in 1 row?

I think if/when we support the proportional method mentioned above, yes. A 6-column layout could theoretically have a wide container -- let's just say 1000px for example -- but when viewed on landscape phone that container will not actually be 1000px, so we would translate that to use the "equal width" layout.

iamthomasbishop:
@lukewalczak Tested out the latest build (#32051) and based on our conversations above, it's feeling pretty solid. As previously discussed, I think there are a number of things we can/should do in the near-term to make adjusting column widths easier (putting all of the column width sliders on the parent, etc) but it sounds like we're in agreement that those are things we can do in the next iteration.

With that said, I have a couple of very minor visual requests which I'll note below. I think functionally we're about ready to ship this iteration and can leave other enhancements to the next one.

## Column Preview

Screen Shot 2020-07-27 at 4 52 15 PM

I refined the design for the column preview indicator. I think this should feel much better.

- Single color for the border and column — can use the same color we're using for other icons (e.g. on the block toolbar)
- Border
    - Width: 1px
    - Placement: inset
    - Radius: 2px
- Dimensions
    - Width: 24
    - Height: 18

## Footer Messaging

I still am not entirely happy with what I originally proposed. Perhaps we should refine the messaging (`Note: columns may stack`...) to something a little clearer like `Column layout may vary between themes and screen sizes`.  Thoughts?

lukewalczak:
Thanks for the feedback! I addressed your suggestions and that's the result:

light | dark
--- | ---
Screenshot 2020-07-28 at 14 43 08 | Screenshot 2020-07-28 at 14 42 50

iamthomasbishop:
Looks great, @lukewalczak! Thank you.

PR #14454, opened 2020-07-10
Fixes reader UI too slow to load at first load

### LINKED ISSUE #14412: Reader: main stream placeholder and content load too slowly
Steps to reproduce:

1. Do a fresh install of the app or log out and back in to a different account.
1. Go to Reader.
1. Observe how long it takes for any placeholder or content to appear.

**Result:** it takes a long time for any content or placeholder to load and feels like something is broken. ([1m57s](https://cloudup.com/c8RexPkRuAS))

screen-shot-2020-07-02-at-jul-2-8 26 05-am IMG_4334
<p style="clear:both;"><sup>Tested with WPiOS 15.2.0.0 TestFlight beta on iPhone 11 iOS 13.5.1.</sup></p>

Gio2018:
Ref #14412 (improves the performance but does not totally fixes the issue)

Loads followed sites asynchronously to avoid blocking the Reader UI when it first loads.

To test:
1. Do a fresh install of the app or log out and back in to a different account.
2. Go to Reader.
3. Verify that the reader UI loads quickly and there is no noticeable blank screen before it loads
4. verify that issue #14370 does not occur

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
@@ -94,7 +94,6 @@ extension ReaderTabItemsStore {
             self?.fetchTabBarItemsAndFollowedSites()
             let actualError = error ?? ReaderTopicsConstants.remoteServiceError
             DDLogError("Error syncing menu: \(String(describing: actualError))")
-
         })
     }
 
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
@@ -87,34 +87,30 @@ extension ReaderTabItemsStore {
         }
         state = .loading
 
-        let dispatchGroup = DispatchGroup()
-
         // Sync the reader menu
-        dispatchGroup.enter()
-        service.fetchReaderMenu(success: {
-            dispatchGroup.leave()
-        }, failure: { (error) in
+        service.fetchReaderMenu(success: { [weak self] in
+            self?.fetchTabBarItemsAndFollowedSites()
+        }, failure: { [weak self] (error) in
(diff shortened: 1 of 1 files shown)


Gio2018:
> Thanks, this change does speed things up for me – although there is still a flash of blank screen while the reader menu is fetched. Hopefully we can do some more there to improve this :)
> 

I agree. I was going to see if we can add a ghost there anyway, as in low network speed we could still see some delay there

> I'm curious – we fetch followed sites here, but then it seems we also re-fetch them every time the filter menu is displayed, as it shows ghost views whenever it appears. What's the reason for that?

If I remember correctly, the idea in the filter was to re-fetch them in order to always have them up to date. There's a couple of options to improve this:

- fetch the followed sites only when sites get added/removed
- instead of re-fetching them every time, use some logic that limits that (time based, etc)

Gio2018:
Thanks for the very thorough testing @emilylaguna ! 
I totally agree that there should be a ghost instead of the blank screen; however, since the loading times in average network connections seem faster (less than one second, on average, as opposed to a few seconds before this change), I am wondering if we should still merge this change into 15.2 to provide some improvement, and then work on adding the ghost?
@frosty what's your opinion?

frosty:
Hmm, I agree that this is an improvement over the current state as it removals a call, but I agree it is still dependent on the network speed.

I think the old Reader menu contained a default set of items (https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/ReaderMenuViewModel.swift) - is it not possible to do something like that in this case?

Alternatively, are we able to ‘preload’ the menu as soon as a user logs in? (Or are we already doing that?)

emilylaguna:
@Gio2018 @frosty I agree with merging this in 15.2 since it is an improvement. 

That being said, I don't think we should close out the original issue (https://github.com/wordpress-mobile/WordPress-iOS/issues/14412) yet so we can still keep track of its status.

Thoughts?

Gio2018:
> That being said, I don't think we should close out the original issue (#14412) yet so we can still keep track of its status.
> 
> Thoughts?

Sounds good to me @emilylaguna !

PR #14459, opened 2020-07-10
Reader Improvements Phase 2: Select Interests: Add next button enable/disable selection toggle

emilylaguna:
Main Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14396

#### 01 - Testing Toggle State
1. If it's not already enabled, enable the Feature Flag for Reader Improvements Phase 2
2. Launch the app
3. Tap the Reader tab
4. The view will be presented
5. Verify the button is now disabled and says "Select a few to continue"
6. Tap an interest to select it
7. Verify the button is now enabled and says "done"
8. Deselect the interest
9. Verify the button is now disabled and says "Select a few to continue"

#### 02 - Testing Temporary next action
1. Follow the steps above
2. When the button is enabled
3. Tap it
4. Verify the view is dismissed

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsStyleGuide.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsStyleGuide.swift
@@ -34,6 +34,17 @@ class ReaderInterestsStyleGuide {
     }
 
     public class func applyNextButtonStyle(button: FancyButton) {
-        // TODO
+        let disabledBackgroundColor: UIColor
+
+        if #available(iOS 13.0, *) {
+            // System Gray 4 on Dark mode is the same color as tertiarySystemBackground
+            disabledBackgroundColor = UIColor(light: .systemGray4, dark: .systemGray3)
+        } else {
+            disabledBackgroundColor = .lightGray
+        }
+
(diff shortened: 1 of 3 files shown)


emilylaguna:
> Just a question though: "Select a _few_ to continue" is the final copy? I was expecting that I'd have to select 2 or 3, but just one is enough. We're sure about this logic?

Good Question. I'm pulling the label from the designs. 

I'm not 100% on the final logic for this right now. When we talked about this a while ago it was something like "Select 5" to continue. I'll confirm the logic with @mbshakti in the design review step.

leandroalonso:
@emilylaguna JFYI: `ReaderSelectInterestsViewController.xib:a7r-YY-Vl7: warning: Medium Style before iOS 13.0 [9]`

mbshakti:
To reduce the friction as much as possible, we decided to enable the button even if the user just picks one topic. The reason the button label says "select a few" instead of "select one" is because "select one" sounds too limiting / odd. And ideally we want to push the user to select a few, not just one; but at the same time, we don't want to prevent them from progressing if they only select one. 

@leandroalonso do you think that's a problem?

leandroalonso:
@mbshakti not at all! Just wanted to confirm, thanks for the info!

PR #14461, opened 2020-07-12
Fix editor More button display on iOS 14b2

### LINKED ISSUE #14460: [iOS 14] Editor More menu content not visible in popover on iPad
In the latest beta of iOS 14 (beta 2), there is an issue in both editors where the content of the More menu isn't visible when displayed in a popover on iPad. It may be visible on the first use of the menu, but is not visible on subsequent taps.

* Run the app on iPad on the latest iOS 14 beta
* Open an editor and tap the More button to view the More menu
* Try opening and closing the More menu several times

frosty:
Fixes #14460.

There's an issue on iOS 14b2 where the contents of the editor More menu isn't visible when displayed as a popover on iPad. I discovered that this is fixed if we display the popover from a `sourceRect` instead of a `barButton`. This seems like an OS-level issue as I can't see we're doing anything wrong, so hopefully we can remove this change in a future PR.

Note that this issue affects both Aztec and Gutenberg.

editor-popover-1

**To test:**

* You'll need to build using Xcode 11.5 to an iPad running iOS 14.2. To do this, you'll also need the latest Xcode 12 beta installed, and to symlink to the device support folder as detailed here: https://gist.github.com/steipete/d9b44d8e9f341e81414e86d7ff8fb62d#file-gistfile1-txt
* Build and run on an iPad running iOS 14b2
* Open Aztec and open the More menu multiple times. Ensure you can see the contents.
* Open the block editor and open the More menu multiple times. Ensure you can see the contents.

PR submission checklist:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. (I don't think we should or can mention iOS 14 changes in release notes).


SergioEstevao:
I've tested this and found the following issue:

- You open the initial popover by tapping ...
- You rotate the device 
- The popover is out of place

I this the problem is because we are using the `sourceRect` property it does not auto-adapt when rotating the device.

jkmassel:
Bumping this to 15.4 :)

frosty:
@SergioEstevao I looked at repositioning the popover after rotation but it felt a little messy (it seemed that I'd have to dismiss and re-present), so I opted just to dismiss the popover on rotation for the purposes of working around this bug.

Issue #14399, opened 2020-06-30
Reader Improvements: Display posts by interests — Main Issue


leandroalonso:
Parent Issue: #14395

<img width="597" alt="Screen Shot 2020-06-30 at 10 09 43" src="https://user-images.githubusercontent.com/7040243/86129969-db594c00-bab9-11ea-882b-e46acb6f5729.png">

### Goal

Based on the Interests the user has selected from #14396 present them as specified in the image above.

# Tasks

* [x] Extract model logics from `ReaderPostServiceRemote` — https://github.com/wordpress-mobile/WordPressKit-iOS/pull/255
* [x] Map the Cards API response to a plain object — https://github.com/wordpress-mobile/WordPressKit-iOS/pull/256
* [x] Extract ReaderPost logic to the model — https://github.com/wordpress-mobile/WordPress-iOS/pull/14421
* [x] Connect the call with the WPiOS — https://github.com/wordpress-mobile/WordPress-iOS/pull/14438
* [x] Persist and present the returned data — https://github.com/wordpress-mobile/WordPress-iOS/pull/14462
* [x] Add welcome banner — https://github.com/wordpress-mobile/WordPress-iOS/pull/14467
* [x] Ghost cells (when loading) — https://github.com/wordpress-mobile/WordPress-iOS/pull/14485
* [x] Retrieve the selected interests and give it to the API — https://github.com/wordpress-mobile/WordPress-iOS/pull/14486
* [x] Check performance — probably fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/14532
* [x] Integration with the real API — https://github.com/wordpress-mobile/WordPress-iOS/pull/14532
* [x] Do small refactorings after #14487 is merged
* [x] Check to make sure quick start is working correctly: https://github.com/wordpress-mobile/WordPress-iOS/pull/14642

PR #14462, opened 2020-07-13
Reader Cards pagination, pull to refresh and interests persistency

leandroalonso:
Part of #14399

This PR adds:

* Cards pagination
* The Core Data models for: cards that display posts and interests you might like
* Pull to refresh

#### Database migration

**Make sure that you have overridden the Reader Improvements Phase 2 Feature Flag (Me -> App Settings -> Debug -> "Reader Improvements Phase 2" should be ON)**

I just added new models and an optional new field in `ReaderPost`.

1. Install the app from `develop` (or any other branch)
2. Switch to this branch and install the app
3. Check that the app runs without any issues

#### Pagination

As discussed before, when the user does a pull to refresh we clean all the data and display it again.

1. Go to Discovery tab
2. Scroll down and check that a new page is loaded every time you hit the bottom
3. Check that the API result is always kept

#### Pull to refresh

As discussed before, when the user does a pull to refresh we clean all the data and display it again. In fact, the logic I created is: if the first page is being retrieved we clear all the data. This can probably be refined in the near future since every time the user goes to the "Discover" tab we will clean the data and reload again.

1. Go back to the top
2. Pull to refresh
3. Check the scroll indicator, it should be way smaller than before

#### Related interests

We're persisting in the list of returned interests.

1. Put a breakpoint in `ReaderCardsStreamViewController.swift` line `16`.
2. Run the app and go to discover
3. When the breakpoint is hit you can debug the persisted Interests using `po (content.content?.first as? ReaderCard)?.interests`
4. You can also open the .sqlite directly

PR submission checklist:


WordPress/Classes/Services/ReaderCardService.swift
```diff
@@ -3,44 +3,69 @@ import Foundation
...
-                            self.coreDataStack.save(syncContext) {
+                            self?.coreDataStack.save(syncContext) {
                                 success(cards.count, true)
                             }
         }, failure: { error in
             failure(error)
         })
     }
+
+    private func clean() {
```
emilylaguna, line 50: Can we make this method name a bit more descriptive?

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -25,6 +31,28 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         }
     }
 
+    // MARK: - Sync
+
+    override func fetch(for topic: ReaderAbstractTopic, success: @escaping ((Int, Bool) -> Void), failure: @escaping ((Error?) -> Void)) {
+        cardsService.fetch(page: 1, success: success, failure: failure)
```
emilylaguna, line 37: Since this is resetting the data wouldn't we want to reset `currentPage` here?

WordPress/Classes/Models/ReaderCard+CoreDataProperties.swift
```diff
@@ -9,5 +9,6 @@ extension ReaderCard {
 
     @NSManaged public var sortRank: Double
     @NSManaged public var post: ReaderPost?
+    @NSManaged public var interests: Set<ReaderInterest>?
```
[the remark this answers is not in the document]
leandroalonso: Core Data only supports `Set`, the data is not ordered by default. If this is changed to `Array` the app crashes.

WordPress/Classes/Services/ReaderCardService.swift
```diff
@@ -3,44 +3,69 @@ import Foundation
...
+    }()
 
     init(service: ReaderPostServiceRemote = ReaderPostServiceRemote(wordPressComRestApi: WordPressComMockrestApi()),
          coreDataStack: CoreDataStack = ContextManager.shared) {
         self.service = service
         self.coreDataStack = coreDataStack
     }
 
-    func fetch(success: @escaping (Int, Bool) -> Void, failure: @escaping (Error?) -> Void) {
+    func fetch(page: Int = 1, success: @escaping (Int, Bool) -> Void, failure: @escaping (Error?) -> Void) {
```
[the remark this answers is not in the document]
leandroalonso: This is one of the tasks I have planned for when we have the real API. :)

WordPress/Classes/Services/ReaderCardService.swift
leandroalonso: Done in 01d9cd0dd3b5b686e23b1ceb52ad8cb662143fed

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
leandroalonso: Yes! Done in 03ba076a452547ccd09dd4dcdd75c7304f10b278

leandroalonso:
@emilylaguna fixed the issue with the blank cells in 91508a17d0ca6054158dd2f306eb60863a9b01f0.

Regarding the warning you mentioned, this is triggered in this line: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/WPTableViewHandler.m#L149

The reason it happens it's because it tries to access the objects of the first cells, but *everything* was deleted — and this is intended.

The code proceeds without any issue (it has logic to deal with this failure).

I will take a note here and we can extensively test this once we have the final API, but so far the warning seems harmless (famous last words? I hope not).

WordPress/Classes/Models/ReaderCard+CoreDataProperties.swift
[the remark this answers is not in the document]
emilylaguna: Ah 🤦‍♀️ yes that makes sense.

WordPress/Classes/Services/ReaderCardService.swift
```diff
@@ -3,44 +3,65 @@ import Foundation
...
-                            syncContext.perform {
-                                cards.forEach { remoteCard in
-                                    guard remoteCard.type != .unknown else {
-                                        return
-                                    }
+                            guard let syncContext = self?.syncContext else {
+                                return
+                            }
+
+                            self?.syncContext.perform {
```
emilylaguna, line 24: Shouldn't this be:

```suggestion
                            syncContext.perform {
```

?

PR #14474, opened 2020-07-15
[Gutenberg] React-navigation in bottom-sheet

dratwas:
In this PR I added a `react-navigation` to use in the bottom-sheet. The react-navigation requires a few native modules to be linked so i added them as well.

Gutenberg PR: https://github.com/WordPress/gutenberg/pull/23782
Gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2549

To test:

- Run editor
- Check settings of each block
- Open Button Block settings
- Click on Text Color / Backgrounds color
- The navigation inside bottom sheet should be possible
- Play a bit with colors and check if the button reflecting these changes
- Try the hardware back button on Android - the back button should back to the previous screen or hide the bottom-sheet if it's the main settings screen

PR submission checklist:


iamthomasbishop:
@dratwas Apologies for the delayed review! I just took the latest iOS build for a spin on iPhone (still have to test on Android and a bit more on iPad). Overall, it’s feeling pretty solid in terms of snap points, heights, etc. but there are some rough edges that still feel a little clunky. I jotted down my feedback/thoughts, and I apologize if any of it has already been answered in other posts or conversations.

## Sheet transitions

I think there is probably some tightening up that we could do in terms of transitions, like for instance when sheets enter/exit or when they transition from a sheet to a sub-sheet and vice versa.

I'm unsure what our transition properties are for various instances — so before I go too deep into the specific values right now, is there a place where I can look at the values that we’re using for the various transitions/easings? That way we can work on consolidating and documenting these into a set of standards or variables?

_Note: I know @lukewalczak worked on a bunch of this during the colors work, so maybe some of this is already there — I can't recall for sure._

### Sheet ↔ sub-sheet

IIRC, our "ideal" for sub-sheet transitions was originally to work on a "push" transition mechanism. With that said, the current cross-fading approach seems to have served us well and is a nice middle-ground and I'd be okay with attacking the push transition mechanism separately if necessary.

I am curious whether the duration (timing) of the transition feels slightly fast. Or perhaps the duration of the transition is correct but the easing curve makes it feel slower? — hard to tell.

For reference: here’s a video comparison of one of our sheets and one in Pages for iOS at a similar height:

https://user-images.githubusercontent.com/1200432/87732872-aaecff80-c793-11ea-9b83-5f3d134d5de4.gif

It's pretty hard to tell the difference in this GIF, but we're talking about 20-40ms difference which makes a big difference on device.

### Sheet entrance timing
While comparing our Inserter sheet entrance timing against the iOS system Activity Sheet (which is usually about the same half-height that our standard sheet is).

I played around with timing a little bit and noticed that the entrance of sheets is a bit slow compared to a common bottom sheet's entrance. Again, not sure if it's because the duration is slow or the easing makes it feel that way.

## Scroll/drag refinements (mostly iOS)

There are a couple of potentially related things here, both of which are mostly specific to iOS and are very small behavioral details that make navigating the sheet surface feel a bit clunky. Neither are new complaints with the library, but I wanted to raise them again in case there's something we can do to solve for them in the short or medium-term.

### Over-scroll at top of containers
The first one is that there is no over-scroll "bounce" when you reach the top edge of a scroll container within a bottom sheet. For example, let's look at the Inserter sheet. If you scroll quickly down to the bottom of the sheet, you'll notice a nice smooth "bounce" animation that scrolls a bit past the end of the container and snaps back. However, this doesn't happen when you quickly scroll to the top of the container — it feels like it's "stuck in the middle" at times — like it should either be closing with the swipe down or the scroll container should bounce. Here's a [video of the behavior](https://cloudup.com/cuJzl_4esT9).

In some cases, it might actually make sense to simply dismiss the sheet if enough scroll force is applied, instead of waiting to reach the top of the sheet then require another swipe to dismiss. For example, a medium-height sheet like the heading levels sheet that's shown when you tap the heading level button on the block toolbar. Note in this example that the sheet is just barely tall enough to allow a few pixels of scrolling, but it feels like it should just close when you swipe down rather than scroll to the top and get "stuck". ([Video](https://cloudup.com/c1gFdoGkQml))

### Over-drag past natural height

Another thing that feels a little off to me is the lack of over-drag of the sheet's surface. Let me explain exactly what I'm referring to, and I will drop a couple of examples because it's hard to explain.

On iOS it's expected that if I drag a short sheet upward past its snap point, the surface should expand a bit past its snap point, with a little friction, then snap back when released.
Here are a couple of [visual examples](https://cloudup.com/ihkFIy3AVkU) — first in Google Tasks on iOS, and second in Apple Notes on iOS.

Are either (or both) of these things possible to solve with the library we're using? They seem like they could be tricky details on the surface, but they would go a long way in making these feel more native on iOS.

## Other

There was one other thing I noticed, and it's related to something I thought was already fixed so maybe it is resolved. I noticed that when dismissing the inserter bottom sheet, there is no smooth transition — the sheet closes instantly.

dratwas:
Hey @iamthomasbishop, thanks for your review :)

>I'm unsure what our transition properties are for various instances — so before I go too deep into the specific values right now, is there a place where I can look at the values that we’re using for the various transitions/easings?

There are 2 kinds of animations, let's call them:
- height animation - the animation of the height of bottom-sheet - in the build you tested it was set to 200 with easing  set to `ease`
- screen animation - the animation between screens. Right now it's the fade one - I believe the react-navigation uses default one from the system so - 350ms, however, we can set whatever config we need 
Here https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/TransitionConfigs/TransitionSpecs.tsx is also a file with duration for Android animations.

>IIRC, our "ideal" for sub-sheet transitions was originally to work on a "push" transition mechanism. With that said, the current cross-fading approach seems to have served us well and is a nice middle-ground and I'd be okay with attacking the push transition mechanism separately if necessary.

The "push" transition is a default one in react-navigation but i have changed it to the fade one because it looks really weird with the animation of height in the bottom-sheet.
rightleftnavi

>While comparing our Inserter sheet entrance timing against the iOS system Activity Sheet (which is usually about the same half-height that our standard sheet is).
I played around with timing a little bit and noticed that the entrance of sheets is a bit slow compared to a common bottom sheet's entrance. Again, not sure if it's because the duration is slow or the easing makes it feel that way.

I did no touch the entrance timing of the bottom sheet but it is set to 600ms for all of bottom-sheet. So, it could be a 3rd animation in the list above :)
```js
animationInTiming={ 600 }
animationOutTiming={ 300 }
```

>Are either (or both) of these things possible to solve with the library we're using? They seem like they could be tricky details on the surface, but they would go a long way in making these feel more native on iOS.

Unfortunately, nothing has changed in these areas since I didn't change the bottom-sheet library (i tried to do that but couldn't find better lib that we use already) So basically in this PR i added a library that handles navigation inside the bottom-sheet and it replaces our "primitive" implementation of subsheets.

>There was one other thing I noticed, and it's related to something I thought was already fixed so maybe it is resolved. I noticed that when dismissing the inserter bottom sheet, there is no smooth transition — the sheet closes instantly.

Yeah, I noticed this as well, it seems like we remove the menu inserter before the animation even starts. I think it should be fixed in separate PR - already created an issue for it https://github.com/WordPress/gutenberg/issues/24060

pinarol:
I opened https://github.com/wordpress-mobile/gutenberg-mobile/issues/2501 to track Slider spacing.

iamthomasbishop:
@pinarol @dratwas I think @antonis has already got this Slider issue on his radar, as we discussed the other day.

iamthomasbishop:
Ok, thanks for the info on animations/transitions, @dratwas! I still have the concerns I mentioned in my last comment WRT the micro-interactions (scroll/drag/etc refinements), but I'll keep this focused on the transitions that we're discussing.

I agree that the fade transition looks better as it exists, but I think with a couple of adjustments it's not far off from being workable. I've got a lot of notes jotted down, but **I am totally fine with tackling some of the more granular animation/transition properties in a separate PR** if necessary — unless of course, you'd rather bundle them in with this work. So I'll start with the push transition part and then get into more detail WRT granular transition/animation properties.

## Push Transition

> The "push" transition is a default one in react-navigation but i have changed it to the fade one because it looks really weird with the animation of height in the bottom-sheet.

I think we may be closer than it seems by looking at the examples. I'm not positive, but perhaps we just need to choreograph the 2 parts a little better. Can we try allowing the horizontal transition (the page push) to complete first, and *then* transition the height? I imagine this would feel a lot more natural — and if we can nail down the easing, it should feel very close to the native iOS transition.

## Sheet Entry Transition

> I did no touch the entrance timing of the bottom sheet but it is set to 600ms for all of bottom-sheet. So, it could be a 3rd animation in the list above :)

```
animationInTiming={ 600 }
animationOutTiming={ 300 }
```

Okay, so we can definitely speed the entry transition up a bit. I took a look at the values in `TransitionSpecs.tsx` and I think I've figured out a common easing curve that should work for both platforms, provided that we can set different durations for each. 

It turns out that the easing curve for `RevealFromBottomAndroidSpec`  ([line 53 in particular](https://github.com/react-navigation/react-navigation/blob/main/packages/stack/src/TransitionConfigs/TransitionSpecs.tsx)) is *very close* to lining up with the iOS system Activity Sheet, provided you set the duration to 500ms.

For Android, I think we might want to shorten the duration on Android, to better align w/ the Material guidelines — which recommends ~250-300ms depending on the transition area the sheet has to cover (they recommend 250ms for bottom sheets, but their example covers only a smaller transition area, so I'm rounding up to 300ms because I think it'll feel more natural.

## Sheet Exit Transition

For the sheet *exit*, we should speed up the transition a bit, and I think we can use a slightly different curve — an `easeIn` curve should work. We'll want to speed up the durations a bit, so let's start with 250ms for both platforms. 

Note: we might want to increase the duration for iOS if it feels too fast, but this is the standard duration for Android so I think it's a good starting point.

## Height/Screen Transitions

This will largely depend on what we decide to do WRT fade vs. page transitions, but I think we can probably try applying the same bezier curve to the height and screen transitions as a starting point, and then speed up or slow down if it doesn't feel right.

dratwas:
Thanks for your feedback @iamthomasbishop!

>I think we may be closer than it seems by looking at the examples. I'm not positive, but perhaps we just need to choreograph the 2 parts a little better. Can we try allowing the horizontal transition (the page push) to complete first, and then transition the height? I imagine this would feel a lot more natural — and if we can nail down the easing, it should feel very close to the native iOS transition.

I spent some time implementing it but managed to achieve something like this:
pushanimationheightnew

I need to refactor it a bit (from the code perspective) but before that, I would love to hear from you if it is something we want to polish :)

**Sheet Entry Transition/Sheet Exit Transition**

I did not mention that but the Bottom-sheet in/out animations are handled by `react-native-modal` and there is no possibility to set easing function in the version we currently use. I think that these animations should be done in separate PR since it needs an update of `react-native-modal` from `6.5.0` to `11.5.~` which could generate some additional issues. Ofc I can change the timing of animations because it is possible in the current version.

**Height/Screen Transitions**
Since the `height` animation is not performant and could be laggy on low android devices we have to use the `LayoutAnimation` for it. Unfortunately, there is no possibility to set custom easing. The list of available easings is here https://reactnative.dev/docs/layoutanimation#properties We can use one of them and set the duration, that's all :(

iamthomasbishop:
Thanks @dratwas for the update!

> I did not mention that but the Bottom-sheet in/out animations are handled by react-native-modal and there is no possibility to set easing function in the version we currently use. I think that these animations should be done in separate PR since it needs an update of react-native-modal from 6.5.0 to 11.5.~ which could generate some additional issues. Ofc I can change the timing of animations because it is possible in the current version.

Ok, that makes sense. Let's just try changing the duration of the transition to speed them up a bit. Can we try 300ms for entry and 250ms for exit?

> I spent some time implementing it and managed to achieve something like this:

Hmm, I think this looks _okay_, but still the cross-fade transition feels more natural. I have a couple of suggestions before we give up on this:

* Perform both transitions (height and x-position) at the same time and duration, so that they act 100% in tandem. Make the duration of both transitions equal to each other and faster, so that it is a bit more quick/fluid.
* Keep what you have now, and cut the duration of the height-change in half.

If neither of those win out, let's stick with the cross-fade transition as it feels solid on both platforms -- I just wanted to see if we could facilitate a more native push transition 😄

dratwas:
Hi @iamthomasbishop, i applied changes that you asked for :)

>Ok, that makes sense. Let's just try changing the duration of the transition to speed them up a bit. Can we try 300ms for entry and 250ms for exit?

openclosefast

>Perform both transitions (height and x-position) at the same time and duration, so that they act 100% in tandem. Make the duration of both transitions equal to each other and faster, so that it is a bit more quick/fluid.

This is not that easy, because the height animation can be run right after the content is rendered because we need to know the height of the content. In most cases, it can be done like that, but for some reason, the `custom color picker` screen runs the animation with a weird lag. I spent some time investigating why it happens but couldn't find fix for that. (I'm afraid that on Android it will work even slower :( )You can see what I mean on the GIF below.  

Note: I observed it during the previous animation that the render of `custom color picker` is really slow.
iosparalelfast

>Keep what you have now, and cut the duration of the height-change in half.

iossequencefast

Please leave your feedback and then I could create a test build with chosen version :)

iamthomasbishop:
@dratwas Enter/exit transitions look better! I'll test vs. other apps when new test builds are available.

Regarding the sub-page transitions, it definitely looks _better_ sped up, but it still doesn't feel quite as natural as the cross-fade transition. So let's stick with the cross-fade and if we feel the need, we can always come back to refine.

dratwas:
Hey @iamthomasbishop i created a new build with latest changes, please let me know what you think about it :)

iamthomasbishop:
@dratwas The faster transitions feel a lot better already, and very close. As discussed previously, ideally we could apply a custom easing curve, but if we aren't able to, let's just slow down the transition-in slightly -- it looks like there's about ~100-200ms diff, so let's try 400ms for the entry and 300ms for the exit.

Here's a side-by-side, slowed to 50% speed: 

bottom-sheets-comparison-wpios-notes-maps 2020-08-05 17_35_19

(Note: the exit is still a bit jarring, but I think you mentioned there's a resolution coming for that)

dratwas:
Hey @iamthomasbishop thanks for quick feedback :)
> but if we aren't able to, let's just slow down the transition-in slightly

It is not possible with the current version of `react-native-modal`, however, I think we really should update that lib since we use out-dated one.

I already changed entry duration to 400 and 300 for the exit. The build should be available soon.

>(Note: the exit is still a bit jarring, but I think you mentioned there's a resolution coming for that)

It happens only for the inserter menu, right? I believe that somewhere upper in the tree we use the `isVisible` to determine if the menu should be rendered and the BottomSheet is removed from the tree before the animation starts. Edit: We use the `dropdown` component which removes the BottomSheet before the animation even starts - https://github.com/WordPress/gutenberg/blob/master/packages/components/src/dropdown/index.native.js#L53

I'm wondering if we could use the local state in the bottom-sheet to determine if it is visible or not, and after the animation, we could call the `onClose` from a parent. It is a definitely thing that should be done in separate PR.

iamthomasbishop:
> It is not possible with the current version of react-native-modal, however, I think we really should update that lib since we use out-dated one.

Hmm, I thought we had sped up the duration of the sheet entrance (this is what I meant by "transition-in" 😄) in the last round of changes? Or do you mean we can't do a custom curve? I think the current curve is workable if we can simply slow it down slightly.

> I'm wondering if we could use the local state in the bottom-sheet to determine if it is visible or not, and after the animation, we could call the onClose from a parent. It is a definitely thing that should be done in separate PR.

That's fine by me.

dratwas:
>Hmm, I thought we had sped up the duration of the sheet entrance (this is what I meant by "transition-in" 😄) in the last round of changes? Or do you mean we can't do a custom curve? I think the current curve is workable if we can simply slow it down slightly.

Sorry for the confusion. I meant we can not change the easing of entrance/exit animation. I set entry duration to 400 and exit to 300 [here](https://install.appcenter.ms/orgs/automattic/apps/WPiOS-One-Offs)

dratwas:
Hey @iamthomasbishop could you please check the build one more time? I fixed the slider issue.

iamthomasbishop:
Looks great! Thank you @dratwas !

PR #14476, opened 2020-07-15
Adds logic to determine when to show the select interests view

emilylaguna:
**Main Issue:** https://github.com/wordpress-mobile/WordPress-iOS/issues/14396
**Sub Task:** Write logic to determine when to display the new interests picker

- This PR adds a service (`ReaderFollowedInterestsService`) to retrieve the followed interests (tags) from the locally and remotely
- It also adds the logic in a coordinator (`ReaderSelectInterestsCoordinator`) to determine if the app should present the select interests view to the user 

**Note:** During testing you will be required to delete all of your followed interests. You may want to use a test account if you don't want to remove all your existing followed tags.

### Testing Logic 
The select interests view should **not** be displayed if:
- The user has followed tags
- The reader has been opened and the logic to check if the view should be displayed has been ran

The last item above will prevent the following scenarios from happening:

**User A:** Has used WordPress iOS before, and has followed some tags. 
> - They update to a newer version of the app where the new reader has been released
> - They decide to delete all of their followed tags
> - They are then forced to select new interests using the Select Interests View

**User B:** Is a new WordPress iOS user
> - They install as version of the app where the new reader has been released
> - They go to the Reader, and see the 'Select Interests' view
> - They select some interests
> - Later they decide to delete all their followed interests
> - They are then forced to select new interests using the Select Interests View

---

**Please enable the "Reader Improvements Phase 2" Feature Flag**

#### 01 - Reset your app
1. Run the app
2. Tap on the Reader tab
3. Tap on the Cog in the filter bar
4. Tap on the 'Followed Tags' item
5. Delete all of your followed interests by clicking the `x`
6. Close the popover

**Expectation:** Verify you are **not** shown the Select Interests view. 
**Reason:** You will not be displayed the view because the check logic has been ran before. 

---

#### 02 - You are presented the select interests view if you have no interests
1. Reset your app using the steps in test 01 if needed
2. Tap on the Reader tab
3. If you have ran the app before you will not be presented with the select interests view
4. Reset has seen before:
4.a. In the Simulator hit the following key command: <kbd>Control</kbd>+<kbd>Command</kbd>+<kbd>Z</kbd>
4.b On device: Shake the device
5. You will see the following in the log:
> Resetting hasSeenBefore: Reader.SelectInterests.hasSeenBefore.{numbers}
6. Tap on the My Site tab
7. Tap back to the reader

**Expectation:** You are shown the Select Interests View
**Reason:** You are presented the view because the "has seen before" user defaults key is false

---

#### 03 - You are not presented the select interests view if you have interests
1. If you have no followed interests add them using the Cog in the filter tab bar
2. Tap on the Reader tab
3. Reset has seen before:
3.a. In the Simulator hit the following key command: <kbd>Control</kbd>+<kbd>Command</kbd>+<kbd>Z</kbd>
3.b On device: Shake the device
6. Tap on the My Site tab
7. Tap back to the reader

**Expectation:** You are not shown the Select Interests View
**Reason:** You not presented the view because the followed interests count is more than 0

---

### PR submission checklist:


WordPress/Classes/Services/ReaderTopicService+Interests.swift
```diff
@@ -1,6 +1,19 @@
 import Foundation
 import WordPressKit
 
+// MARK: - ReaderInterestsService
+
+/// Protocol representing a service that retrieves a list of interests stored remotely
```
leandroalonso, line 6: One may wonder that `stored remotely` means that this is the user's list of interests.

This just returns the list of the interests that the user can follow, right? If so, I guess we can update the comment to better reflect that.

WordPress/Classes/Services/ReaderTopicService+Interests.swift
```diff
@@ -1,6 +1,19 @@
...
+protocol ReaderInterestsService: AnyObject {
+    /// Fetches a large list of interests from the server
+    /// - Parameters:
+    ///   - success: Called upon successful completion and parsing, provides an array of `RemoteReaderInterest` objects
+    ///   - failure: Called upon network failure, or parsing errors, provides an Error object
+    func fetchInterests(success: @escaping ([RemoteReaderInterest]) -> Void,
+                        failure: @escaping (Error) -> Void)
```
leandroalonso, line 13: The name of this method and the parameters are pretty straightforward. Do you think we really need the comment?

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    /// Creates a new instance of the coordinator
+    /// - Parameter service: An Optional `ReaderFollowedInterestsService` to use. If this is `nil` one will be created on the main context
+    ///   - store: An optional backing store to keep track of if the user has seen the select interests view or not
+    ///   - userId: The logged in user account, this makes sure the tracking is a per-user basis
+    init(service: ReaderFollowedInterestsService? = nil,
+         store: KeyValueDatabase = UserDefaults.standard,
+         userId: NSNumber? = nil) {
+
+        let defaultContext = ContextManager.sharedInstance().mainContext
```
leandroalonso, line 20: I would avoid creating inline instances. If `ReaderSelectInterestsCoordinator` needs a context that should be a param in the constructor.

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
```diff
@@ -0,0 +1,93 @@
...
+            let acctServ = AccountService(managedObjectContext: defaultContext)
+            let account = acctServ.defaultWordPressComAccount()
+
+            return account?.userID
+        }()
+    }
+
+    // MARK: - Display Logic
+
+    /// Deteremines whether or not the select interests view should be displayed
```
leandroalonso, line 34:
```suggestion
    /// Determines whether or not the select interests view should be displayed
```

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    private func shouldDisplaySelectInterests(with interests: [ReaderTagTopic]?) -> Bool {
+        guard let interests = interests else {
+            return false
+        }
+
+        let hasSeen = hasSeenBefore()
+        DDLogDebug("Reader Improvements: Count: \(interests.count), has seen before? \(hasSeen)")
```
leandroalonso, line 49: Do we really need this log here?

If so, "Reader Improvements" is directly related to how we're naming this project, but it's not directly related to this part. It might create some confusion in the future.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +323,38 @@ import WordPressFlux
         }
     }
 
+
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
         syncIfAppropriate()
 
+        debugCheckIfNeedToDisplaySelectInterests()
```
leandroalonso, line 332: Why this method was named this way?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +323,38 @@ import WordPressFlux
...
+
+    }
+    private var interestsCoordinator = ReaderSelectInterestsCoordinator()
```
leandroalonso, line 335: Can we move that to the top of the class?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +323,38 @@ import WordPressFlux
...
+                if shouldDisplay {
+                    let controller = ReaderSelectInterestsViewController()
+                    self.navigationController?.present(controller, animated: true)
+                }
+                self.interestsCoordinator.markAsSeen()
+            }
+        }
+    }
+
+    override func becomeFirstResponder() -> Bool {
```
leandroalonso, line 350: Why do we need this? Just debug purposes?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +323,38 @@ import WordPressFlux
...
+    }
+    private var interestsCoordinator = ReaderSelectInterestsCoordinator()
+
+    private func debugCheckIfNeedToDisplaySelectInterests() {
         // TODO: Remove this
         if FeatureFlag.readerImprovementsPhase2.enabled {
-            DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500)) {
-                let controller = ReaderSelectInterestsViewController()
-                self.navigationController?.present(controller, animated: true)
+            interestsCoordinator.shouldDisplay { (shouldDisplay) in
```
leandroalonso, line 340:
```suggestion
            interestsCoordinator.shouldDisplay { shouldDisplay in
```

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +323,38 @@ import WordPressFlux
...
+                if shouldDisplay {
+                    let controller = ReaderSelectInterestsViewController()
+                    self.navigationController?.present(controller, animated: true)
```
leandroalonso, line 343: I'm not sure if this is the final solution, but we're sure this class should handle that?

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
```diff
@@ -0,0 +1,169 @@
+import XCTest
+@testable import WordPress
+
+// MARK: - MockFollowedInterestsService
+class MockFollowedInterestsService: ReaderFollowedInterestsService {
```
leandroalonso, line 5: When I open a test the main thing I'm interested is in... the tests! :P

Here I have to scroll all the way down until I find 'em. Mocks can be after the main test suite, wdyt?

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        } catch {
+            XCTAssertNil(error, "Error fetching interests")
+
+            return nil
+        }
+    }
+}
+
+class ReaderSelectInterestsCoordinatorTests: XCTestCase {
+    func testShouldDisplay() {
```
leandroalonso, line 93:
```suggestion
    func testShouldDisplayReturnsTrue() {
```

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        coordinator.shouldDisplay { (result) in
+            displayExpectation.fulfill()
+
+            XCTAssertTrue(result)
+        }
+
+        waitForExpectations(timeout: 4, handler: nil)
+    }
+
+    func testShouldNotDisplayIfUserHasFollowedInterests() {
```
leandroalonso, line 111:
```suggestion
    func testShouldDisplayReturnsFalseIfUserHasFollowedInterests() {
```

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        coordinator.shouldDisplay { (result) in
+            displayExpectation.fulfill()
+
+            XCTAssertFalse(result)
+        }
+
+        waitForExpectations(timeout: 4, handler: nil)
+    }
+
+    func testShouldNotDisplayIfUserHasSeenBefore() {
```
leandroalonso, line 131:
```suggestion
    func testShouldDisplayReturnsFalseIfUserHasSeenBefore() {
```

Has seen what?

leandroalonso:
@emilylaguna also, if the user selects "Discover" and it's not following any Interest I'd display this screen every time.

Otherwise, we need a different error message and the option for the user to select the interests.

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14476/commits/f8df5f7bde0afb37349061e7a2f9e603a17f9ec7

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14476/commits/f8df5f7bde0afb37349061e7a2f9e603a17f9ec7

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14476/commits/f8df5f7bde0afb37349061e7a2f9e603a17f9ec7

WordPress/WordPressTest/ReaderSelectInterestsCoordinatorTests.swift
emilylaguna: Good idea. Pushed: https://github.com/wordpress-mobile/WordPress-iOS/pull/14476/commits/7af34cf571c6c2657a05df09d29d297b9e760c30

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: The logic here is just for testing.

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
emilylaguna: This was just for testing to help determine why the view was not displaying.

emilylaguna: Removed in: bce9d04a935825c2a7ed134467b8b2f3cc52aeba

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
emilylaguna: Fixed in: f222e163204f908ae76cc2aa30495ee02bc98bcd

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
emilylaguna: Good catch, done in: 1b6a10258e1234026741cf11e6502bfb485ab949

WordPress/Classes/Services/ReaderTopicService+Interests.swift
emilylaguna: Done! 263b50fa21ab548902d78b95a44f46f3559121c5

WordPress/Classes/Services/ReaderTopicService+Interests.swift
emilylaguna: Done 263b50fa21ab548902d78b95a44f46f3559121c5

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: Because it's just debug code that will be removed once the presentation of the view is done in a later task. This PR is just creating the logic.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: Yes, it enables the motionEnded method

emilylaguna:
> 1. After the user selects the interests, we're showing the "Following" tab. Shouldn't we show the Discover session?

Correct, this will be changed in my next task. 

> 2. The logic to save the selected interests will be in another task?

Yes, next one.

> 3. ~I tested the flow as logged out user and the screen never appeared for me.~ (my mistake)

I did find an issue where the markAsSeen wasn't working correctly for logged out users.

emilylaguna:
@leandroalonso back at you!

WordPress/Classes/Services/ReaderTopicService+FollowedInterests.swift
```diff
@@ -0,0 +1,56 @@
...
+
+// MARK: - CoreData Fetching
+extension ReaderTopicService: ReaderFollowedInterestsService {
+    public func fetchFollowedInterestsLocally(completion: @escaping ([ReaderTagTopic]?) -> Void) {
+        completion(followedInterests())
+    }
+
+    public func fetchFollowedInterestsRemotely(completion: @escaping ([ReaderTagTopic]?) -> Void) {
+        fetchReaderMenu(success: { [weak self] in
+            guard let `self` = self else {
```
yaelirub, line 24: Whats the purpose of this guard? Can we just do: `self?.fetchFollowedInterestsLocally(completion: completion)` ?

emilylaguna: I originally had it like that, but if self fails nothing happens. This just ensures the callback is called no matter what.

yaelirub: In what cases was self nil? Usually that would point to a sever issue

WordPress/Classes/Services/ReaderTopicService+FollowedInterests.swift
```diff
@@ -0,0 +1,56 @@
...
+                completion(nil)
+                return
+            }
+
+            self.fetchFollowedInterestsLocally(completion: completion)
+        }) { (error) in
```
yaelirub, line 30: Tiny nit picking, you don't have to use parentheses when you have only one parameter in the block 
`{ error in
}`

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        return !hasSeenBefore() && interests.count <= 0
+    }
+
+    // MARK: - View Tracking
+
+    /// Generates the user defaults key for the logged in user
+    /// Returns nil if we can not get the default WP.com account
+    private var userDefaultsKey: String? {
+        get {
```
yaelirub, line 55: The 'get' is not needed here.

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
```diff
@@ -0,0 +1,87 @@
...
+            return String(format: Constants.userDefaultsKeyFormat, userId ?? 0)
```
yaelirub, line 56: I'd rather guard for the userId and return nil in the else instead of `userID ?? 0` that way the code is self explanatory and I d on't have to wonder what will this return if userId is nil and we use 0

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +325,37 @@ import WordPressFlux
...
+
+    }
+
+    private func debugCheckIfNeedToDisplaySelectInterests() {
         // TODO: Remove this
         if FeatureFlag.readerImprovementsPhase2.enabled {
-            DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500)) {
-                let controller = ReaderSelectInterestsViewController()
-                self.navigationController?.present(controller, animated: true)
+            interestsCoordinator.shouldDisplay { shouldDisplay in
```
yaelirub, line 341: consistency is important 😬  no parentheses here

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -323,16 +325,37 @@ import WordPressFlux
...
+            }
+        }
+    }
+
+    override func becomeFirstResponder() -> Bool {
+        return true
+    }
+
+    override func motionEnded(_ motion: UIEvent.EventSubtype, with event: UIEvent?) {
+        if FeatureFlag.readerImprovementsPhase2.enabled {
```
yaelirub, line 356: What is this function all about? also, can you combine the if statements?

WordPress/Classes/Services/ReaderTopicService+FollowedInterests.swift
leandroalonso: If we're pretty sure that `self` will exist we can use `[unowned self]`.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: It's just a debug method to help with testing. I am removing it before committing to develop. Without having a way to easily reset the has seen key you would need to completely delete the app and reinstall to test each flow.

WordPress/Classes/ViewRelated/Reader/ReaderInterestsCoordinator.swift
emilylaguna: If we return nil here the logged out user key will be nil which means the view tracking wouldn't be registered. 

I've updated how the default works by using `Constant.loggedOutUserKey` and removing the optional from the property: https://github.com/wordpress-mobile/WordPress-iOS/commit/2be6cdb2bdaac0a4d7efcb45a3aaa9a96c497d50

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: I've removed it: https://github.com/wordpress-mobile/WordPress-iOS/commit/4014d4b173d09ea2fde2a80c5f3454c929fb4988

WordPress/Classes/Services/ReaderTopicService+FollowedInterests.swift
emilylaguna: Changed here: 875e1a8567a3c87e244590a83e64327d3fcd601c

emilylaguna:
@emilylaguna also, if the user selects "Discover" and it's not following any Interest I'd display this screen every time.

Makes sense. I'll keep that in mind when I do the real display implementation.

emilylaguna:
Back to you guys @leandroalonso @yaelirub

Issue #14483, opened 2020-07-17
Cannot create a new page from the FloatingActionButton on iOS 12.4 and below


chipsnyder:
### Expected behavior
The editor should open a new page.

### Actual behavior
The modal dismisses without opening a page. 

### Steps to reproduce the behavior
From "My Site" screen
- On a Device or Simulator running iOS 12.4 or Below
- Select the Floating Action Button (FAB)
- Select "Site Page"

<img width=300 src=https://user-images.githubusercontent.com/3384451/87797444-af0f3080-c818-11ea-8544-0f29bf069ba1.gif />

##### Tested on 
This was tested on `develop` branch at `752d9c6f4836fcb55b0a2cc301f0d3319a25d8a3`
iPhone Simulator - iPhone 11, iOS 13.6 - No Problems 👍 
iPhone Simulator - iPhone 7, iOS 12.4 - Saw Error ❌ 
iPhone Simulator - iPhone 7, iOS 11.4 - Saw Error ❌

PR #14484, opened 2020-07-17
Fix issue with presenting the New Page editor when selecting from the FAB icon

chipsnyder:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/14483

### To test:
On a Simulator or Device running iOS 12.4 or below and again running 13.0 or above:
- Select the Floating Action Button (FAB)
- Select "Site Page"
- Expect to see the page editor open

### Additional Context
This could be related to a timing issue that my simulators for 12.4 and below hit more regularly than my 13.0 and above device and simulators.

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
@@ -434,15 +434,15 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
         __weak __typeof(self) weakSelf = self;
         _createButtonCoordinator = [[CreateButtonCoordinator alloc] init:self newPost:^{
             [weakSelf dismissViewControllerAnimated:true completion:nil];
-            [((WPTabBarController *)self.tabBarController) showPostTabWithCompletion:^(void) {
-                [self startAlertTimer];
+            [((WPTabBarController *)weakSelf.tabBarController) showPostTabWithCompletion:^(void) {
+                [weakSelf startAlertTimer];
             }];
         } newPage:^{
-            [weakSelf dismissViewControllerAnimated:true completion:nil];
-            
-            WPTabBarController *controller = (WPTabBarController *)self.tabBarController;
-            Blog *blog = [controller currentOrLastBlog];
-            [controller showPageEditorForBlog:blog];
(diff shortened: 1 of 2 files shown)


chipsnyder:
Hey @Gio2018 and @bjtitus 👋 

I found a bit of a bug while working on some changes for Gutenberg. The fix was pretty straight forward, so I think it only really needs one of you to take a look. GitHub suggested both of you so whoever gets time is first is fine but I wanted to add each of you because I'm not sure who would be the best to take a look.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -434,15 +434,15 @@ - (CreateButtonCoordinator *)createButtonCoordinator
         __weak __typeof(self) weakSelf = self;
         _createButtonCoordinator = [[CreateButtonCoordinator alloc] init:self newPost:^{
             [weakSelf dismissViewControllerAnimated:true completion:nil];
-            [((WPTabBarController *)self.tabBarController) showPostTabWithCompletion:^(void) {
-                [self startAlertTimer];
+            [((WPTabBarController *)weakSelf.tabBarController) showPostTabWithCompletion:^(void) {
+                [weakSelf startAlertTimer];
```
Gio2018, line 438: Thanks for fixing this!!

bjtitus:
@Gio2018 I noticed exactly the issues you mentioned but didn't have time to dig into the details yet.

> I'm wondering if we should just remove the two `[weakSelf dismissViewControllerAnimated:true completion:nil]`

I think this makes sense if it works for this situation. 👍

chipsnyder:
@Gio2018 and @bjtitus, thank you both for taking the time to review. I considered that briefly but held off because I wasn't entirely sure what our overall recommend pattern was. (i.e.) Should the one being presented perform the cleanup or the one doing the presenting? I'm glad you called it out. I agree with both of you; I would prefer only to have one piece of code own that behavior. 

I made the requested updates; this is ready for review again.

chipsnyder:
Hey, @bjtitus and @Gio2018 just wanted to check in and see if either of you were able to rereview after the suggested changes were made?

chipsnyder:
No problem 😀. Thanks, @Gio2018 !

Issue #14395, opened 2020-06-29
Reader Improvements: Phase 2 - Project Issue


emilylaguna:
This is the main project issue for Phase 2 of the Reader Improvements Project

### Tasks
- [x] Create a feature flag for the project: https://github.com/wordpress-mobile/WordPress-iOS/pull/14406
- [x] Add a new Select Interests screen: https://github.com/wordpress-mobile/WordPress-iOS/issues/14396
- [x] ~Add a new loading view that is displayed to the user while the posts are loading:~ N/A this is being done in https://github.com/wordpress-mobile/WordPress-iOS/issues/14396
- [x] Add a new Discover feed that displays a list of posts for the selected interests: https://github.com/wordpress-mobile/WordPress-iOS/issues/14399
- [x] Update Reader Card to display a list of recommended interests: #14500
- [x] Update Reader Card to display a list of tags for the posts in the Discover feed: https://github.com/wordpress-mobile/WordPress-iOS/pull/14645
- [x] Final Design Review: https://github.com/wordpress-mobile/WordPress-iOS/issues/14745

PR #14500, opened 2020-07-21
Reader Improvements: show topics the user might like

leandroalonso:
Part of #14395

This PR is part of the Reader Improvements and list topics the user might be interested in following under the "Discover" tab.

### To test

First, make sure you have the flag Reader Improvements Phase 2 **enabled**.

#### Showing the suggested topics

1. Go to Reader
2. Tap Discover
3. Check that you're presented with "Topics you might like"

#### Tapping a suggested topic

1. Follow the steps above
2. Tap a suggested topic
3. Check that posts about this topic are shown

### Known issue

I've been noticed that something is probably blocking the main thread for a few seconds. I did a timebox to discover but was unable to find the culprit, I added this in #14399 to be checked.

#### iPhone

| Portrait  | Landscape |
| Simulator Screen Shot - iPhone 11 - 2020-07-23 at 16 56 09 | Simulator Screen Shot - iPhone 11 - 2020-07-23 at 16 56 16 |

#### iPad

| Portrait  | Landscape |
| Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-07-23 at 17 05 09 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-07-23 at 17 07 53 |

PR submission checklist:


mbshakti:
Stoked for this. 

Already discussed on slack, but two comments:

1. Can we have the separator go edge-to-edge on iPhone
2. Show 4 rows of topic names instead of 5

Thank you!

leandroalonso:
@mbshakti 

> Can we have the separator go edge-to-edge on iPhone
Sure! I'll try to add that without compromising the readable margin in bigger screens.

> Show 4 rows of topic names instead of 5
The number of rows comes from the API, so it's pretty straightforward. If we want 4, the API just has to return 4 results. It shows 5 because I added 5 interests in the API response.

WordPress/Classes/Models/ReaderTagTopic.swift
```diff
@@ -4,6 +4,7 @@ import Foundation
     @NSManaged open var isRecommended: Bool
     @NSManaged open var slug: String
     @NSManaged open var tagID: NSNumber
+    @NSManaged open var card: NSOrderedSet?
```
yaelirub, line 7: Should't this be cards if its a set?

WordPress/Classes/Services/ReaderCardService.swift
```diff
@@ -27,25 +27,32 @@ class ReaderCardService {
             self.service.fetchCards(for: slugs,
                                success: { [weak self] cards in
 
-                                guard let syncContext = self?.syncContext else {
+                                guard let `self` = self else {
```
yaelirub, line 30: can just be self 😬

WordPress/Classes/Services/ReaderTopicService.m
```diff
@@ -964,14 +964,19 @@ - (void)mergeMenuTopics:(NSArray *)topics withSuccess:(void (^)(void))success
                         // If the user adds a locally saved tag/interest prevent it from being deleted
                         // while the user is logged out.
                         if ([Feature enabled:FeatureFlagReaderImprovementsPhase2]) {
-                            if (!ReaderHelpers.isLoggedIn && [topic isKindOfClass:ReaderTagTopic.class]) {
-                                ReaderTagTopic *tagTopic = (ReaderTagTopic *)topic;
+                            ReaderTagTopic *tagTopic = (ReaderTagTopic *)topic;
 
-                                if (tagTopic.wasAddedWhileLoggedOut) {
+                            if (!ReaderHelpers.isLoggedIn && [topic isKindOfClass:ReaderTagTopic.class]) {
```
yaelirub, line 969: Can we not just use tagTopic under this if (tagTopic.title...) here and avoid the kind of class check

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -3,20 +3,32 @@ import Foundation
...
     override func viewDidLoad() {
         super.viewDidLoad()
         ReaderWelcomeBanner.displayIfNeeded(in: tableView)
+        tableView.register(ReaderSuggestedTopicsCell.self, forCellReuseIdentifier: readerCardTopicsIdentifier)
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        // Only sync if the tableview is at the top, otherwise this will change offset
+        if isTableViewAtTheTop() {
+            super.viewWillAppear(animated)
```
yaelirub, line 21: Not sure this is "allowed" from the documentation:
`If you override this method, you must call super at some point in your implementation`
Maybe you always call `super.viewWillAppear(animated)` but do:  
 if isTableViewAtTheTop() {
  tableView.reloadData() 
}
?

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -3,20 +3,32 @@ import Foundation
...
+        }
     }
 
     // MARK: - TableView Related
 
     override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         if let posts = content.content as? [ReaderCard], let cardPost = posts[indexPath.row].post {
             return cell(for: cardPost, at: indexPath)
+        } else if let posts = content.content as? [ReaderCard], let interests = posts[indexPath.row].interests?.array as? [ReaderTagTopic] {
```
yaelirub, line 30: So something doesn't feel right in this function. 
1. First, take the data source out:
Something like : 
```
 lazy var readerCards: [ReaderCard]? = {
        return content.content as? [ReaderCard]
    }()
```
And then look at the specific card:
 `let readerCard = readerCards[indexPath.row]`
I also recommend implementing:
```
 func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
        return readerCards?.count ?? 0
       }
```
So you'll have more confidence that you're not out of bound when calling cell for row.

2. We want to make this scaleable for other types of cards, right ? Why don't we save the type of card from the remote and switch on that?

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
```diff
@@ -0,0 +1,159 @@
...
+protocol ReaderTopicsCellDelegate: class {
+    func didSelect(topic: ReaderTagTopic)
+}
+
+/// A cell that displays topics the user might like
+///
+class ReaderSuggestedTopicsCell: UITableViewCell {
+    private let containerView = UIView()
+
+    private let tableView = TopicsTableView()
```
yaelirub, line 12: Im confused by the tableView in the cell? Do we have a table inside the cell? I only see it used for reloading data but I don't understand why

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
```diff
@@ -0,0 +1,159 @@
...
+    func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
+        return 1
+    }
+}
+
+extension ReaderSuggestedTopicsCell: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        let header = UIView()
+        let headerTitle = UILabel()
+        headerTitle.text = "You might like"
```
yaelirub, line 133: Localization please

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
```diff
@@ -0,0 +1,159 @@
...
+        header.addSubview(headerTitle)
+        headerTitle.translatesAutoresizingMaskIntoConstraints = false
+        header.pinSubviewToAllEdges(headerTitle, insets: UIEdgeInsets(top: 10, left: 15, bottom: 10, right: 0))
+        headerTitle.font = WPStyleGuide.serifFontForTextStyle(.title2)
+        return header
+    }
+
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let topic = interests[indexPath.row]
```
yaelirub, line 142: Isn't it weird / confusing  to use interests and topic instead of just one of them?

yaelirub:
Great work here!

I also noticed the main thread blockage. Maybe by fixing other things it will lead to a fix or and if not, I'd love to jump on a call and be your rubber duck 🦆 

other issues:
1. I got to this weird state (not consistently) after relaunching the app:


2. I think I remember testing the add interests screen in Hebrew and seeing the interests localized but they are not in the new section:
Screen Shot 2020-07-25 at 5 53 01 PM

3. I left some comments in code

WordPress/Classes/Services/ReaderTopicService.m
leandroalonso: It looks like that I've added this line but it was already here.

The main issue is that we have different topics: `ReaderTagTopic`, `ReaderListTopic`, `ReaderDefaultTopic`, `ReaderTeamTopic`....

I guess this piece of logic is exclusive to `ReaderTagTopic`.

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
leandroalonso: Hmm, I wasn't sure about that.

I'll figure out another solution.

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
leandroalonso: Do you mean renaming `ReaderSuggestedTopicsCell` to `ReaderInterestsCardCell`?

WordPress/Classes/Models/ReaderTagTopic.swift
leandroalonso: Yes!

Done in 3a6cfe67a1ae21593e87fda24956511dd94309e1

WordPress/Classes/Services/ReaderCardService.swift
leandroalonso: Done in a3899f4642ab0476001331fb02ff075193b70cbf

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
leandroalonso: Feito em 3d61ef34f18b8f7f19ecc37f49982a380e15251f

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
leandroalonso: Yes, I completely agree and that is something that I thought while implementing that.

Since the title says `Topics you might like` and the entity is `ReaderTagTopic` I renamed all to `topics`: e1fbd0161744758df36a12850cb886208d4ef84c

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
leandroalonso: Yeah, I agree with you.

I don't like this implementation, I'll try to make it better.

leandroalonso:
> Maybe by fixing other things it will lead to a fix or and if not, I'd love to jump on a call and be your rubber duck 🦆

Deal

> I got to this weird state (not consistently) after relaunching the app:

Interesting. What are the steps to reproduce? Just restart the app?

> I think I remember testing the add interests screen in Hebrew and seeing the interests localized but they are not in the new section:

This is because the API data is mocked and the i18n comes from the API. Once we connect to the real API this will work!

WordPress/Classes/ViewRelated/Reader/ReaderSuggestedTopicsCell.swift
leandroalonso: I renamed to `ReaderTopicsCardCell` in e9ce1d7471fd94c7f76345e429a64291505731e5

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
leandroalonso: Changed in 9f28f40c2288a5a5f79803dda39dab3143b15ca4

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
leandroalonso: Ok! I simplified that a lot! Let me know what do you think. :)

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -3,21 +3,35 @@ import Foundation
...
-        if let posts = content.content as? [ReaderCard], let cardPost = posts[indexPath.row].post {
-            return cell(for: cardPost, at: indexPath)
-        } else {
+        guard let card = cards?[indexPath.row] else {
+            return UITableViewCell()
+        }
+
+        switch card.type {
+        case .post:
+            return cell(for: card.post!, at: indexPath)
```
leandroalonso, line 31: @yaelirub I know you hate force unwraps! 😝 But in this case, I already know that `post` isn't `nil`, so its safe to do that.

leandroalonso:
@yaelirub ready for another look!

WordPress/Classes/Services/ReaderTopicService.m
yaelirub: 👍

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
yaelirub: So much better!!! 👏

WordPress/Classes/Models/ReaderCard+CoreDataClass.swift
```diff
@@ -2,6 +2,28 @@ import Foundation
 import CoreData
 
 public class ReaderCard: NSManagedObject {
+    enum CardType {
+        case post
+        case topics
+        case unknown
+    }
+
+    var type: CardType {
```
yaelirub, line 11: Interesting. Wouldn't it be simpler to just do 
`var type: CardType = .unknown`
and then:
```
switch remoteCard.type {
        case .post:
            post = ReaderPost.createOrReplace(fromRemotePost: remoteCard.post, for: nil, context: managedObjectContext)
            self.type = .post
        case .interests:
            topics = NSOrderedSet(array: remoteCard.interests?.map {
                ReaderTagTopic.createIfNeeded(from: $0, context: context)
            } ?? [])
            self.type = .topics
        default:
            break
        }
```

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
yaelirub: lol fair enough

yaelirub:
>Deal

Quack! lets have a call tomorrow 
>Interesting. What are the steps to reproduce? Just restart the app?

yep. Just relaunch the app from Xcode. 

>This is because the API data is mocked and the i18n comes from the API. Once we connect to the real API this will work!

WordPress/Classes/Models/ReaderCard+CoreDataClass.swift
leandroalonso: Any reason why you prefer this implementation?

My main points with the current are:

1. I find it a little bit cleaner since the logic of the card type is detached from the parsing;
2. It's a little bit safer: `ReaderPost.createOrReplace` can return nil, in that case, `type` would be `.unknown`. With this implementation, it will be `.post` and it might lead to weird issues.

Anyway, no big feelings. Let me know what do you prefer.

leandroalonso:
> one more comment and the issues in my last comment + main thread blocking

About the main thread blocking, my plan is to leave that for a subsequent task (since it's not part of this). Plus, I'm working on integrating the Discover with the cards API, and apparently this fixed the issue (it might be related to reading the mocked file).

> yep. Just relaunch the app from Xcode.

I'm really not being able to reproduce that. I have a theory on what is happening (the data is being excluded from Core Data) but I'm not sure why this happens to you but not to me. There's anything on your log about deleting posts and topics?

WordPress/Classes/Models/ReaderCard+CoreDataClass.swift
yaelirub: As discussed in DM, In my mind, the local type should represent the remote type

leandroalonso: @yaelirub oops, I just realized that this needs to be a computed property.

`type` is not persisted to the database, so what happens is that it's always returning `.unknown` (because it's set only at the moment the object is created).

We can keep that way or I can remove the `type` property and refactor `cellForRowAt` to be like this:

```swift
    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
        guard let card = cards?[indexPath.row] else {
            return UITableViewCell()
        }

        if let post = card.post {
            return cell(for: card.post!, at: indexPath)
        } else if !card.topicsArray.isEmpty {
            return cell(for: card.topicsArray)
        } else {
            return UITableViewCell()
        }
    }
```

Thoughts?

yaelirub: Gotcha. Up to you

leandroalonso:
@yaelirub I'll keep the `ReaderCard.type` as it is now but we can rediscuss in the next tasks. 🙏

PR #14456, opened 2020-07-10
Create Modal Layout Picker Container for new Gutenberg pages and feature flag for development

chipsnyder:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2419

Related PRs:
- `gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2478
- `gutenberg`: https://github.com/WordPress/gutenberg/pull/23872

## Description
Adds feature flags and a container view for creating a new modal layout 

## To test:
<details>
<summary>To disable or enable the development version of Modal Layout Picker</summary>

- Open the app from the build https://github.com/wordpress-mobile/WordPress-iOS/pull/14456#issuecomment-656759183 or a local development build
    - By default, the modal layout picker will be enabled.
    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
        - Toggle "Gutenberg Modal Layout Picker" to enable or disable the picker
            <details>
            <summary>To toggle:</summary>
            
            </details>
</details>

- Steps to test the Gutenberg side of the modal layout picker toggle can be found in https://github.com/WordPress/gutenberg/pull/23872

### Classic Editor and Gutenberg Editor with Modal Layout Picker Disabled.
**Note:** Modal Layout Picker should be disabled and uneditable in any Release build
- Create a new page from My Site
    - Create a new page by clicking on the Floating Toolbar Button on the My Site Page then select "Site Page"
    - Expect to see the page editor without being prompted with the new modal layout picker
- Create a new page from Pages
    - Create a new page by clicking on the ➕ on the Pages screen
    - Expect to see the page editor without being prompted with the new modal layout picker

### Gutenberg Editor with Modal Layout Picker Enabled
- Create a new page from My Site
    - Create a new page by clicking on the Floating Toolbar Button on the My Site Page then select "Site Page"
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor
- Create a new page from Pages.
    - Create a new page by clicking on the ➕ icon on the Pages screen.
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor

### On the Modal Layout Picker
- Selecting the ✖️icon in the top right-hand corner should close the modal without creating a page.
- Swiping down around the "Grip" in the top middle of the modal should close the modal without creating a page.
    <details>
        <summary>Dismissing the modal</summary>
        
    </details>
[...]

PR #14501, opened 2020-07-21
Create Modal Layout Picker Container for new Gutenberg pages and feature flag for development

### LINKED ISSUE #2419: [Modal Layout Picker] Create Modal window for Modal Layout Picker
### Is your feature request related to a problem? Please describe.
This task is part of the effort to rework the Starter Page Template Picker for new pages. This part of the initiative will focus on creating the base modal for the new Modal Layout Picker. Other tasks will then work off of the modal created in this task. It will also handle the feature flags to set up the build environments to enable the development and building of the new modal.

The goal will be to create a UI that looks something like:

| iOS | Android |
| --- | --- | 
| iOS-Container | Android-Container |

### Goals - Modal Container
+ When creating a new page, a modal should display with the following items:
    + Title
    + Subtitle
    + "Create Blank Page" Button
    + A button for dismissing the modal. 
    + These items should be styled for Light and Dark interfaces. 
+ Selecting the "Create Blank Page" button should create a new blank page.
+ Selecting the "X" (iOS) or Up button (Android) should dismiss the modal.
    + Android should validate the back button works as expected for dismissing the modal.
+ Swiping down on the modal should dismiss the modal. 

### Goals - Feature Flag
+ Selecting the Create page should only show this new modal when running:
    + iOS build configurations: `localDevelopment` and `a8cBranchTest`.
    + Android build configurations: `vanilla` and `jalapeno`.
+ Opening the editor **should not** show the existing STP picker if the modal **has** been shown
    + We can pass the feature flag to RN as part of capabilities ([iOS](https://github.com/WordPress/gutenberg/blob/7532a485b400f86638145b71f94f6f717e5add25/packages/react-native-bridge/ios/GutenbergBridgeDataSource.swift#L43), [Android](https://github.com/WordPress/gutenberg/blob/7532a485b400f86638145b71f94f6f717e5add25/packages/react-native-bridge/android/src/main/java/org/wordpress/mobile/WPAndroidGlue/WPAndroidGlueCode.java#L470))
    + The key to pass as part of the capabilities should be `modalLayoutPicker`. `true` would mean the app supports modal layout picker
+ Opening the editor **should** show the existing STP picker if the modal **has not** been shown

### Additional context
+ Creating a new page can happen from:
    + The FAB button on the bottom right of the site's home screen
    + The ➕ on the top right on the "Pages" screen in iOS or the FAB button on the "Pages" screen in Android

<details>
<summary>Blueprint Flow - iOS</summary>

</details>

<details>
<summary>Blueprint Flow - Android</summary>

</details>

chipsnyder:
Fixes: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2419

Related PRs:
- `gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2505
- `gutenberg`: https://github.com/WordPress/gutenberg/pull/24170

## Description
Adds feature flags and a container view for creating a new modal layout for page creation in Gutenberg.

## To test:
<details>
<summary>To disable or enable the development version of Modal Layout Picker</summary>

- Open the app from the build https://github.com/wordpress-mobile/WordPress-iOS/pull/14456#issuecomment-656759183 or a local development build
    - By default, the modal layout picker will be enabled.
    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
            <details>
            <summary>To toggle:</summary>
            
            </details>
</details>

- Steps to test the Gutenberg side of the modal layout picker toggle can be found in https://github.com/WordPress/gutenberg/pull/23872

### Classic Editor and Gutenberg Editor with Modal Layout Picker Disabled.
**Note:** Modal Layout Picker should be disabled and uneditable in any Release build
- Create a new page from My Site
    - Create a new page by clicking on the Floating Toolbar Button on the My Site Page then select "Site Page"
    - Expect to see the page editor without being prompted with the new modal layout picker
- Create a new page from Pages
    - Create a new page by clicking on the ➕ on the Pages screen
    - Expect to see the page editor without being prompted with the new modal layout picker

### Gutenberg Editor with Modal Layout Picker Enabled
- Create a new page from My Site
    - Create a new page by clicking on the Floating Toolbar Button on the My Site Page then select "Site Page"
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor
- Create a new page from Pages.
    - Create a new page by clicking on the ➕ icon on the Pages screen.
    - Expect to see the modal layout picker container
    - Select "Create Blank Page"
    - Expect to see the page editor

### On the Modal Layout Picker
- Selecting the ✖️icon in the top right-hand corner should close the modal without creating a page.
- Swiping down around the "Grip" in the top middle of the modal should close the modal without creating a page.
    <details>
        <summary>Dismissing the modal</summary>
        
    </details>
- Scrolling up in the stubbed rows should collapse the header. Collapsing the header should:
    - Minimize the title
    - Hide the prompt
    - Leave space for adding the category bar in the future.
- Scrolling down in the stubbed rows should reset the header when reaching the top.
    <details>
    <summary>Continuing to scroll down after reaching the top should dismiss the modal</summary>
        
    </details>
- Toggle Dark mode and check the style seems consistent with other views

## Screenshots:
| Container | Collapsed Header | Dark Mode |
| --- | --- | --- |
|   |  |  |

## Additional Context
This PR was branched off https://github.com/wordpress-mobile/WordPress-iOS/pull/14456 using a slightly different approach. For design feedback history please see that PR. 

There are some poor layout issues in landscape mode, specifically with larger dynamic fonts. The plan is to revisit landscape at a later point after more of the screen is filled out. 

## PR submission checklist:


WordPress/Classes/Services/PageCoordinator.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+class PageCoordinator {
+    typealias TemplateSelectionCompletion = (String?) -> Void
```
chipsnyder, line 4: The string return value here is in preparation for adding the template slug or contents depending on what path works out the best then.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -0,0 +1,193 @@
+import UIKit
+import Gridicons
+
+class GutenbergLayoutPickerViewController: UIViewController {
+
+    @IBOutlet weak var headerView: UIView!
+    @IBOutlet weak var closeButton: UIButton!
+    @IBOutlet weak var titleView: UILabel!
```
chipsnyder, line 8: Specifically using a `titleView` here instead of the navigation bar's title because there was a weird random animation issue when scrolling quickly and setting the title on the screen the label would "fly" in from the upper left.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+
+    override func viewDidDisappear(_ animated: Bool) {
+        (navigationController as? GutenbergLightNavigationController)?.shadowIsHidden = false
+        super.viewDidDisappear(animated)
+    }
+
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        (navigationController as? GutenbergLightNavigationController)?.shadowIsHidden = false
+        super.prepare(for: segue, sender: sender)
+    }
```
chipsnyder, line 58: There aren't any segues on the screen yet but they will be added with a preview functionality later. I wanted to add this functionality now so it wouldn't be forgotten when that work is done.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -0,0 +1,193 @@
...
+        } else {
+            seperator = .lightGray
+        }
+
+        [createBlankPageBtn].forEach { (button) in
+            button?.titleLabel?.font = WPStyleGuide.fontForTextStyle(.body, fontWeight: .medium)
+            button?.layer.borderColor = seperator.cgColor
+            button?.layer.borderWidth = 1
+            button?.layer.cornerRadius = 8
+        }
```
etoledom, line 114: Asking out of curiosity:
Why did you decide to use `forEach` in this way?

chipsnyder: The designs have one other button that will need to be added that needs the same styling applied. I built that additional button out at one point so I could test the animations when pushing another ViewController. So I just left this syntax in place for when that button is added back. 

I also considered using a function like `stylePlainFooterButton` instead of doing a `forEach` but didn't feel like a function added any more clarity in this case.

etoledom: Thanks for the explanation, sounds good! :) 

Another alternative I like to use is creating a `styledButtons` OutletCollection (or any other name).
The styling code would be almost the same. It just avoids the array literal in code, and avoids the need to add new elements in code.
Then you just add the buttons you want to style to that collection via ctrl-drag and it gets styled 👍 

But up to you, this is perfectly fine!

chipsnyder: >Another alternative I like to use is creating a `styledButtons` OutletCollection ...

That is a cool idea! I re-style the buttons for trait changes (dark mode) and content size changes (text) so I think I'll leave it as is, otherwise, I probably would switch to that because it sounds like a neat style.

etoledom (commented):
Overall it's looking very good! Great job @chipsnyder 🎉 

Most of my comments are UI/UX stuff, so I'd ask @iamthomasbishop to validate them. Thanks! 🙏 

### 1. iPhone 5 screen size support.

Testing on an iPhone 5S (that we still support with iOS 11/12) on landscape mode, there's no space at all to start scrolling the content.

We might also want to add top margin to the `close` button and reduce `create blank page` bottom margin on this config.

iPhone5

This also affects iPhone 6/7 screen sizes. On them there's little space to start scrolling and some users might have difficulties doing so.

IMG_0020

### 2. Accessibility.

I believe it's never too early to think about this ;)
- The close button would benefit from an accessibility label.
- Not sure why but I didn't manage to scroll the content with Voice-Over active. Maybe with actual content will work better.

### 3. Small details for ❤️ (not blocker)

- It would be great if we can animate also the appearance and disappearance of the Nav Bar title label.

gif-label

chipsnyder:
Thanks for the review @etoledom!

> 1. iPhone 5 screen size support.

I agree with your concerns on landscape mode. @iamthomasbishop and I were discussing that in DM a bit. We decided to take care of that in another issue (which I'll create today and update here). We'd like to add some more content to the screen to help us determine the best approach. This also affects larger devices as you increase the dynamic font size.

>2. Accessibility.
That's a great callout I didn't think about accessibility thank you for testing this! I'll be sure to add a label for the close button. For the scrolling, I think I'll add a note to the issue that adds content to the scroll area to validate if that becomes more managable.

> 3. Small details for ❤️ (not blocker)

The suggestion in here should be relatively easy to add! I think this is a great idea I'll get that in place.

iamthomasbishop:
Agreed with everything @etoledom mentioned! The subtle transition on the navbar title would be wonderful. As a stretch goal, can we try a opacity cross-fade between the large title and navbar title? This would happen as the title gets close to/is going under the navigation bar -- the large title would fade out and the collapsed title would fade in.

One more thing I'd like is to try applying the regular or chrome system material to the navbar and footer bar (in iOS 12+, iirc), so when scrolling they would become translucent. Is that doable here, @chipsnyder ?

chipsnyder:
@iamthomasbishop 

> One more thing I'd like is to try applying the regular or chrome system material to the navbar and footer bar (in iOS 12+, iirc), so when scrolling they would become translucent. Is that doable here, @chipsnyder ?

It looks like [chrome system material](https://developer.apple.com/documentation/uikit/uiblureffect/style/systemchromematerial) is available on iOS 13+ so I can apply the standard blur on lower versions (this is what the footer had on previous builds). 

The way I have the header set up I'll need to make some changes to the layout to get that working but I can get that wrapped up in the morning probably.

As for the animation of the title view, I got that setup. Let me know what you think about it. The only piece I couldn't get working great is when you scroll quickly you'll see both headers as one scrolls into view and the other is animating out.

I just triggered a new build so that should be ready soon if you want to give it a try while I adjust the header.

iamthomasbishop:
> The way I have the header set up I'll need to make some changes to the layout to get that working but I can get that wrapped up in the morning probably.

Sounds good 👍 

> As for the animation of the title view, I got that setup. Let me know what you think about it. The only piece I couldn't get working great is when you scroll quickly you'll see both headers as one scrolls into view and the other is animating out.

This feels really solid. The momentary double-title thing doesn't bother me a whole lot, but if you are able to solve it then even better. 😀

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -56,6 +56,7 @@ class GutenbergLayoutPickerViewController: UIViewController {
         closeButton.setImage(UIImage.gridicon(.crossSmall), for: .normal)
         styleButtons()
         layoutHeader()
+        navigationController?.navigationItem.rightBarButtonItem?.accessibilityLabel = NSLocalizedString("Close", comment: "Close the modal window")
```
etoledom, line 74: This instance of `rightBarButtonItem` is `nil`, so this line is not actually setting the localized accessibility label.
What we want is to find the button on the `GutenbergLayoutPickerViewController`'s instance of `navigationItem`.

Each Controller manages their own navigation items.

This currently works because the label is also set via Storyboard, so it's a bit confusing, but won't be localized.

etoledom (commented):
Thank you for the update @chipsnyder - It's looking a lot better now! 🙏

Landscape mode on iPhone 7 is now much better! 🎉

I added a few of code comments, only the one related to accessibility label might be a blocker for now.

> The momentary double-title thing doesn't bother me a whole lot, but if you are able to solve it then even better. 😀

About the animation, it also looks good to me :) 

I did a fast test anchoring the title's alpha value to the `largeTitleView.frame.maxY` value and got an interesting result:

title_animation

In this way the alpha animation speed is controlled by the scroll itself. Seems simple enough to be implemented in this round but I won't block on this one 👍

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
chipsnyder: > This currently works because the label is also set via Storyboard, so it's a bit confusing, but won't be localized.

Actually I don't think the storyboard Accessibility was being determined. Without this line VoiceOver reads "Button Possibly Close" however with this line it reads as "Close Button". 

However, with the changes @iamthomasbishop requested, I'll need to move this out of the navigation controller to make sure the Visual Effect line up. So it'll go away after I push those changes.

etoledom: > Without this line VoiceOver reads "Button Possibly Close" however with this line it reads as "Close Button".

Hah, Interesting! I did the exact same test (and just did it again now to be sure) and in both cases (with or without the line) VoiceOver reads it as "Close".

I'm using an iPhone 7 device.

Also `navigationController?.navigationItem.rightBarButtonItem` is nil to me. So ¯\\_(ツ)_/¯ 

> So it'll go away after I push those changes.

Ok then! 👍

WordPress/Classes/Services/PageCoordinator.swift
```diff
@@ -0,0 +1,25 @@
...
+    static func showLayoutPickerIfNeeded(from controller: UIViewController, forBlog blog: Blog, completion: @escaping TemplateSelectionCompletion) {
+        if FeatureFlag.gutenbergModalLayoutPicker.enabled && blog.isGutenbergEnabled {
+            showLayoutPicker(from: controller, completion)
+        } else {
+            completion(nil)
+        }
+    }
+
+    private static func showLayoutPicker(from controller: UIViewController, _ completion: @escaping TemplateSelectionCompletion) {
+        let storyboard = UIStoryboard(name: "LayoutPickerStoryboard", bundle: Bundle.main)
```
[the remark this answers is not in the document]
chipsnyder: > both the custom Navigation Controller and the Layout picker can also be used separately if needed in the future.

This would still be possible in one storyboard. The Navigation Controller can still be created programmatically with the standard UINavigationController init methods. I can add a `storyboardID` to the Gutenberg Layout Picker View Controller Scene to allow the instantiation of that separately if needed. 

> Alternatively, we could also instantiate the NavigationController in code and set its root controller

I'm happy to switch this over if that's our standard practice. My goal here was to follow Apple's recommendation that the storyboard contains all of the design-time elements that are related to the flow of a set of screens. `GutenbergLayoutPickerViewController` assumes that the navigation controller is an instance of `GutenbergLightNavigationController` to hide the bar's shadow properly. Although with the changes I'll be making to add the visual effects @iamthomasbishop requested, I'll hide the navigation bar for this view instead. However, `GutenbergLightNavigationController` does still have other considerations in it to enforce the flow displays correctly, and using the view without a navigation controller won't work as we build out the [preview button](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2452) which will push a new view controller.

Since this felt more connected in the flow of the UI I felt the storyboard should include it to provide the full picture. If that time came when we did want to reuse the root view controller, that would still be doable with minimal code changes. Being forced to make those changes would also prompt the future developer to consider the best way to handle the design settings of the navigation controller to accommodate all of the uses.

Ultimately I think either approach will work, I just wanted to share my thoughts on it, and I'm happy to proceed in whichever way you feel is best and most consistent.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/LayoutPickerStoryboard.storyboard
```diff
@@ -224,15 +224,17 @@
...
                         <barButtonItem key="rightBarButtonItem" id="y9q-Mg-b6a">
-                            <button key="customView" opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" lineBreakMode="middleTruncation" id="Fgx-ig-LXQ">
+                            <button key="customView" opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" lineBreakMode="middleTruncation" id="Fgx-ig-LXQ" userLabel="Close">
                                 <rect key="frame" x="364" y="11" width="30" height="34"/>
-                                <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
+                                <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                                 <color key="backgroundColor" systemColor="systemFillColor" red="0.47058823529999999" green="0.47058823529999999" blue="0.50196078430000002" alpha="0.20000000000000001" colorSpace="custom" customColorSpace="sRGB"/>
+                                <accessibility key="accessibilityConfiguration" label="Close"/>
```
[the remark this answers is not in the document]
chipsnyder: Is this more related to how we handle localization for our app? ie should I be setting all text programmatically? 

If I localize the storyboard I get:

```
/* Class = "UIButton"; accessibilityLabel = "Close"; ObjectID = "Fgx-ig-LXQ"; */
"Fgx-ig-LXQ.accessibilityLabel" = "Close";
```
But if we don't localize the storyboards then I can move the other text values over.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
chipsnyder: Very interesting! I'm on an iPhone 11 running 13.5 so I wonder if there was a change there although the release notes don't have anything. Good to know though!

chipsnyder:
>I did a fast test anchoring the title's alpha value to the largeTitleView.frame.maxY value and got an interesting result:

I'm going to revisit this path to solve the double case. It looks like you're using a different anchor point than I was. The way I tried it originally was phasing in the other small title by the % that the big title was off the screen and it didn't look great, because you would see both.

Can you share the snippit you used for your quick test?

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
chipsnyder: Just closing the loop on this one because I thought it was interesting. I retested and I'm getting the same results you are. So I probably had a build issue that was using a cached storyboard.

chipsnyder:
@iamthomasbishop and @etoledom This is updated now with a new build (`32185`) generated for the header chrome style. However, It's hard to see the difference without setting the background color for cells. So we might want to focus on that in the next PR when I add some content to the rows. 

I also didn't make any changes to the header animation. I started building in a cancel for that, but it started causing crashes. So I figured that might be better to adjust toward the end if we still want to.

WordPress/Classes/Services/PageCoordinator.swift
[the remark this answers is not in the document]
etoledom: Thanks for the detail explanation!

Originally the main idea behind my suggestion was to hide the required init steps and configuration inside the `GutenbergLayoutPickerViewController`, in a way that `PageCoordinator.swift` doesn't need to know about storyboard details that seems internal to the controller (specially with it being tightly bounded with the navigation controller).

The Storyboard config as you explained it makes perfect sense to me! Let's keep it this way.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/LayoutPickerStoryboard.storyboard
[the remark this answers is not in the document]
etoledom: > Is this more related to how we handle localization for our app? ie should I be setting all text programmatically?

Yes I believe so. Our Localization scripts make use of `NSLocalizedString` to gather and send the strings to GlotPress, so using localized storyboard I believe it won't work.

etoledom:
By `Anchoring` I just meant making it (inverse) relative to `largeTitleView.frame.maxY`.

> Can you share the snippit you used for your quick test?

This is a quite rough example:

```swift
func scrollViewDidScroll(_ scrollView: UIScrollView) {
       // ....

        if largeTitleView.frame.maxY > 0 {
            titleView.isHidden = true
        } else {
            titleView.isHidden = false
            titleView.alpha = (-largeTitleView.frame.maxY / 10.0).clamp(min: 0, max: 1)
        }
}
```
Then you removed everything related to `titleIsHidden`.
Not sure if `clamp` is needed. This is probably already clamped internally.

There might be other considerations too with the code to handle small screen sizes.

etoledom:
> It's hard to see the difference without setting the background color for cells. So we might want to focus on that in the next PR when I add some content to the rows.

I agree with this too! 👍

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/LayoutPickerStoryboard.storyboard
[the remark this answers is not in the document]
chipsnyder: Thanks for the info! I'll move those strings over.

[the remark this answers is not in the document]
etoledom: Thank you!

Here's some info about localization best practices (from our team):
https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/docs/localization.md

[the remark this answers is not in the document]
chipsnyder: 🙇‍♂️

chipsnyder:
@etoledom Those static strings have been moved to be programmatic now. I left the original values for the labels in place just to help the system guess an appropriate intrinsic size after loading.

chipsnyder:
> By Anchoring I just meant making it (inverse) relative to largeTitleView.frame.maxY.

Ah ok, I see what you're doing here. When I did it, I originally had 
````
let alpha = (largeTitleView.frame.height-largeTitleView.frame.maxY / largeTitleView.frame.height)
```` 
So as the large title was getting close to disappearing, the small title would appear, and then you ended up with a case where both were showing which I didn't like. 

I put this in place, and it seems to be working well. I just needed to make some adjustments to when it "snaps" into place and when entering compact views. There is an additional point in the scrolling where you don't see either header in this approach compared to the other however the "snap" fixes that so it wouldn't really be noticeable. 

Just as a quick aside where did you come up with the `10`? Or is this just a magic number that felt right?

etoledom (approved):
Thank you for the update! 

With the changes on the header, is it still necessary to have a custom Navigation Controller?

> After doing some more testing on this approach I'm noticing the view flashes a bit. Since we're happy with the animation the way it is I think I would prefer to explore this approach at a later time. WDYT?

Absolutely, let's not block this PR on the animation 👍 

> Just as a quick aside where did you come up with the 10? Or is this just a magic number that felt right?

Yeah, it's a magic number.
I noticed that maxY changed around 1 to 10 in the range we were interested in, so dividing that by 10 looked like a good option. But doesn't have to be strictly 10.

(Looks like kind of a `t = d/v` equation 🤔, so 10 would be the velocity of the transition).

About the glitch (for a future PR):
We could try fix that glitch if we start the alpha changes a bit earlier. Let's say half height of the `largeTitleView` earlier. Something like:

`let alpha = -(maxY - (height / 2)) / 10`
Very similar to what you had previously :) 
In this way on the snap position, the calculation already gives `alpha = 1` and it won't jump to other values.

----

Let's move ahead, this looks good and feels great! Nice job @chipsnyder 🎉

chipsnyder:
Thanks @etoledom, I appreciate your great review :) 

> With the changes on the header, is it still necessary to have a custom Navigation Controller?

The style for the navigation bar when we push the Preview screen is a bit different than the existing Light Navigation Controller so we'll still want the customized one for that.

PR #14503, opened 2020-07-21
Update translators review builds

loremattei:
This PR updates the build for translators to use the .com GlotPress instance. Currently, the .com instance is sync'ed only for some languages, so the script will download only those. 

To test:
1. Run `bundle exec fastlane build_for_translation_review` and verify that the build succeeds and that the translations for the languages in the  `REVIEW_LANGS` array have been downloaded. 
2. Clean up any local change. 
3. Run `.Scripts/update-translations.rb` from the project root folder and verify that it succeeds and that all the translations have been downloaded.  

PR submission checklist:

### DIFF
--- a/Scripts/update-translations.rb
+++ b/Scripts/update-translations.rb
@@ -84,12 +84,14 @@ ALL_LANGS={
 langs = {}
 strings_filter = ""
 strings_file_ext = ""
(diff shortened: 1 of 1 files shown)


akirk:
Fantastic, thank you!

PR #14508, opened 2020-07-22
Second notification prompt for users that have not set notifications permissions

Gio2018:
Ref #14202

This PR adds a second Notifications Permissions prompt for those users who chose "Not Now" on the inline prompt in the 'Notifications' tab. The alert will look like in the picture below:

To test:
pre requisites:

- a WordPress.com account on a device
- another WordPress.com account (on another device or simulator) that can trigger notifications on the first account (e.g. follow a site from the first account, where you can like posts or comment on them), or any other way you could think of, to send notifications to the first account

Test 1
1. Fresh install the app on the device and login with the first account
2. Go to 'Notifications' and, when prompted for permissions, tap "Not Now"
3. Visit 'Notifications' 5 more times until the inline prompt shows
4. On the inline prompt, tap "Not Now"
5a. While on 'Notifications', trigger 10 notifications (e.g. like or comment on a post with the second account)and verify that a second alert with title "Find out faster with push notifications" shows up (see example above)
5b. While on a tab other than 'Notifications', trigger 10 or more notifications, then go to 'Notifications' and verify that a second alert with title "Find out faster with push notifications" shows up (see example above)

Test2:
- go to steps 1 to 4 in Test 1 (or any subset of it) with any version of the app
- Update the app to this branch and proceed with the remaining steps 

##### Test tips:
1. To avoid waiting for 10 notifications, you can temporarily set the treshold to a lower value. This can be done tweaking 'Constants.secondNotificationsAlertTreshold` in `NotificationsAlertViewController.swift`
2. To re-test without re-installing, you can reset Userdefaults values as follows:

- pause the debugger
- to reset the second alert notifications counter, in the lldb console, type:
`e -l swift -- import WordPress; UserDefaults.standard.secondNotificationsAlertCount = 0`
- to reset the inline prompt value (so that it shows up again) in the lldb console, type:
`e -l swift -- import WordPress; UserDefaults.standard.notificationPrimerInlineWasAcknowledged = false`
- Relaunch the app and the values should be reset

3. The test will work as long as the push notifications permissions are not yet set. If at any moment before steps 5 you set the permissions (by tapping on "Allow push notifications" and then allowing or denying them in the system), the second alert will not show (regardless resetting the user defaults). In this case, you'll need to do a fresh install again to re-test.

PR submission checklist:


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1678,21 +1743,25 @@ extension NotificationsViewController: UIViewControllerTransitioningDelegate {
...
+            return
+        }
+        showSecondNotificationAlert()
+        userDefaults.secondNotificationsAlertCount = Constants.secondNotificationsAlertDisabled
+    }
+
     private enum Constants {
         static let inlineTabAccessCount = 6
         static let displayAlertDelay = 0.2
+        static let secondNotificationsAlertTreshold = 10
```
mindgraffiti, line 1764:
```suggestion
        static let secondNotificationsAlertThreshold = 10
```

mindgraffiti (approved):
💬  I know you didn't create this file. Would you write up a proposal for breaking apart this class? Nearly 1800 LoC and the method names scroll to boot 😱 .

✅ Test 1, steps 5a & 5b pass
✅ Test 2, steps 5a & 5b pass

:shipit: 

*Initial ask*
<a href="https://user-images.githubusercontent.com/1062444/88225302-e114fe00-cc2f-11ea-8338-bb69e413e895.PNG"></a>

(forgot to screenshot the inline ask, but it did appear as expected)

*Second ask*
<a href="https://user-images.githubusercontent.com/1062444/88225322-ea05cf80-cc2f-11ea-8a0b-c5f442c674a1.PNG"></a>

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
Gio2018: Thanks for spotting this @mindgraffiti !!!

Gio2018:
Thanks for reviewing @mindgraffiti !!

> Would you write up a proposal for breaking apart this class? Nearly 1800 LoC and the method names scroll to boot 😱 

Maybe. The use of extensions to break things up makes sense, but it's true it's pretty long (plus there are three more files containing separate extensions). I have been thinking for a while what could be a reasonable approach (for this and other ViewControllers) to start separating at least the views (so that we could be more ready for SwiftUI).

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -31,7 +31,7 @@ enum FeatureFlag: Int, CaseIterable {
         case .unifiedAuth:
             return BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
         case .unifiedSiteAddress:
-            return false
+            return true
```
ScoutHarris, line 34: I don't think you meant to commit this... At least, I hope not. 😄

Gio2018: oh no! Sorry about that, I had reviewed your PR just before and for some reason git allowed me to checkout my branch anyway... thanks for finding this!!

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1094,6 +1100,22 @@ extension NotificationsViewController: WPTableViewHandlerDelegate {
...
     }
 
+    private var shouldCountNotificationsForSecondAlert: Bool {
+        userDefaults.notificationPrimerInlineWasAcknowledged &&
+            userDefaults.secondNotificationsAlertCount != Constants.secondNotificationsAlertDisabled
+    }
+
+    func tableViewWillChangeContent(_ tableView: UITableView) {
+        guard shouldCountNotificationsForSecondAlert,
+            let note = tableViewHandler.resultsController.fetchedObjects?.first as? Notification,
```
ScoutHarris, line 1110: Total nitpick - but can this be `notification` (in all the places it's `note`) ? Here it's not too bad, but it is a bit confusing as it's not immediately clear `note` = `notification`.

ScoutHarris (approved):
One naming nitpick, but otherwise 👍 .

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
Gio2018: Totally! Note does not mean notification (I don't even recall why I named it that way..)

Gio2018:
thanks @ScoutHarris and @mindgraffiti for reviewing!!

osullivanchris (approved):
Tested and this works as expected! Thanks!

osullivanchris:
Added the status: Needs Copy Review. FAO Editorial:

- Today when a user installs the app, they see a pop up which says
**Headline**: Stay in the loop
**Body**: We'll notify you when you get followers, comments and likes. Would you like to allow push notifications?
**Actions**: Allow push notifications / Not now

They also see an in-line prompt at the top of the notifications tab
**Body**: We'll notify you when you get followers, comments, and likes.
**Actions**: Allow notifications / Not now

So we have tweaked the logic of when we're displaying those two prompts. And in this PR, we are adding a third prompt. The third prompt is showing to a user who has said 'Not now' twice. We are showing them an additional prompt after receiving 10 notifications. The hypothesis being that they have now received some value and might be inclined to accept. You can see the placeholder copy above
**Body**: Find out faster with push notifications
**Actions**: Allow notifications / Not now

- actions are staying the same as we want to keep that in-limbo pool for future attempts/experiments
- I said 'Find out faster' as I thought it was a way to emphasise the benefit of push notifications over in-app notifications
- Also thought it was important to say 'Push notifications' to emphasise its going to the device home screen ,not about in app notificaitons

Hope that's enough info. Would love your input. Let us know if you need more clarity.

benhuberman:
Thanks for the super-clear recap, @osullivanchris  -- it really helps with getting the proper context. 

First, a very quick fix: in the first pop-up, there's a serial comma missing. It's currently:
> We'll notify you when you get followers, comments and likes.

But should be:
> We'll notify you when you get followers, **comments,** and likes. 

As for the overall messaging, if there's space for an extra word in that first pop-up as well as in the inline prompt, I'd add "new":
> We'll notify you when you get **new** followers, comments and likes.

For the new prompt, my concern is that "Find out faster" is too vague and doesn't immediately tell the user what it is that they can find out. The most straightforward path I can think of would be something like this:
**Prompt**
> Get your notifications faster

**Action** (note the added "push" here)
> Allow push notifications
>Now now

Would that work from your end? 

Alternative prompts could be something like:
> Why not get your notifications faster?

> Stay up-to-date — faster

osullivanchris:
@benhuberman thanks so much. I discussed it with @Gio2018 now and we both like what you've done and think it makes sense, so we'll go ahead with that! The initial 'Get your notifications faster' sounds straightforward but we'll have your other options in our pocket if we run into any issues!

Gio2018:
Thanks @benhuberman and @osullivanchris !

Here are the screenshots after the changes to the first prompt, second prompt and second action (with the added "push")

first popup | second popup
:-------------------------:|:-------------------------:
 |

benhuberman:
Great, let me know if I can help with anything else here @osullivanchris @Gio2018

bummytime:
@osullivanchris @Gio2018 for the second popup:

88400160-8a253b00-cd8d-11ea-8b2b-3537809879e7

Is there a reason the body is empty? It feels like something is missing to me.

osullivanchris:
@bummytime I can totally see what you mean. I'm trying to figure out why. In the mock below I think it looked ok. I can see two reasons why it doesn't look so well in the build even though its implemented correctly by Giorgio:
- I think the font weight in the real build is heavier than it looks in my mock. So there's more emphasis on it being a heading (without a body) than it seemed from the mock, where to me it was just looking like one piece of text
- The final copy is quite short and to the point. But visually, looks maybe like there's more to come.

push-notifications-request-3

@benhuberman if a subheading was available would you use it? Or do you think the message we have is effective? Do you also think it looks not right now we see it in situ? 

I think we should only add more copy if we actually think its a better message. If not, maybe there's something we can tweak in the padding or layout to make it look more 'complete'

frosty:
Thanks for the response Chris! I agree that I think it needs a little more explanatory text – I’m not sure it’s obvious what the intent of the alert is just from the title. Could we use the same text from the earlier notice?

benhuberman:
@osullivanchris It's tough to say without seeing it in action, but if we wanted to fill up some of the empty space there, we _could_ go with something like:

**Heading**:
> Get your notifications faster

**Subheading**:
> Learn about new comments, likes, and follows in seconds.

Is that... erring on the side of too much copy? I like the conciseness of my original suggestion, but I confess that I didn't really take visual/design considerations into account.

frosty:
I like this extra suggestion, @benhuberman. I think my issue with the original is just that it doesn't explicitly say what kinds of push notifications users will receive. The new one does that by mentioning comments, likes, and follows.

osullivanchris:
I think that's a good argument for the additional copy in terms of information and not just layout @frosty. So lets go with what @benhuberman has said above! 

Thanks again Ben!

Gio2018:
just pushed the changes to the body text of the second alert, here is how it looks:

<p align="center">
Screen Shot 2020-07-27 at 4 29 38 PM
</p>

osullivanchris:
Looks good to me! Thanks @Gio2018 !

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
```diff
@@ -101,6 +109,18 @@ import Foundation
...
         case .gutenbergUnsupportedBlockWebViewClosed:
             return "gutenberg_unsupported_block_webview_closed"
+        case .pushNotificationsPrimerSeen:
+            return "notifications_primer_seen"
+        case .pushNotificationsPrimerAllowTapped:
+            return "notifications_primer_allow_tapped"
+        case .pushNotificationsPrimerNoTapped:
+            return "notifications_primer_no_tapped"
+        case .secondNotificationsAlertSeen:
+            return "second_notifications_alert_seen"
```
frosty, line 119: Just wondering if we should swap the event name around here so it's similar to the `notifications_primer` events – for example, `notifications_second_alert_seen`. It might help out when looking at tracks events and trying to filter them?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -1678,21 +1743,25 @@ extension NotificationsViewController: UIViewControllerTransitioningDelegate {
...
+            return
+        }
+        showSecondNotificationAlert()
+        userDefaults.secondNotificationsAlertCount = Constants.secondNotificationsAlertDisabled
+    }
+
     private enum Constants {
         static let inlineTabAccessCount = 6
         static let displayAlertDelay = 0.2
+        static let secondNotificationsAlertThreshold = 10
```
frosty, line 1764: Could we add one small comment to clarify what this is used to count?

frosty (commented):
Looking good! Just two more small code comments!

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
Gio2018: sure!!

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
Gio2018: 👍

PR #14509, opened 2020-07-22
Show post content when navigating between notifications

leandroalonso:
This PR adds the feature to show the post content when using the arrows to navigate between notifications.

Previous, we just showed the post title with an image or an excerpt, now the full post is shown.

| Before  | After |
|  |  |

### To test

1. Go in the Notifications tab
2. Make sure you have at least one notification of a new post published in any site
3. Tap the closest notification (below or above) that one
4. Use the arrow to navigate to the post notification
5. Make sure the content is rendered

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
@@ -68,6 +68,10 @@ class NotificationDetailsViewController: UIViewController {
     ///
     fileprivate let estimatedRowHeightsCache = NSCache<AnyObject, AnyObject>()
 
+    /// A Reader Detail VC to display post content if needed
+    ///
+    private var readerDetailViewController: ReaderDetailWebviewViewController?
+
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -496,6 +501,31 @@ extension NotificationDetailsViewController {
...
 
+// MARK: - Reader Helpers
+//
+private extension NotificationDetailsViewController {
+    func attachReaderViewIfNeeded() {
+        guard shouldAttachReaderView,
+            let postID = note.metaPostID,
+            let siteID = note.metaSiteID else {
+                readerDetailViewController?.remove()
+            return
```
yaelirub, line 512: `return` needs indentation 😬

leandroalonso: Ops! Fixed in fc8ff0d8533620c0e6f6d2b678971761df5f29a3

PR #14516, opened 2020-07-24
Update Fastlane to version 2.152.0

mokagio:
This is a preventive update to avoid issues with Fastlane's `deliver`.

While working on Simplenote iOS 4.21, I experienced a failure updating the localized metadata via `deliver`, which updating Fastlane to the latest version solved.

I didn't spend extra time tracking down the culprit or the exact version fixing the issue because staying relatively up to date with Fastlane doesn't hurt anyways.

Another thing to note is I removed the `rubyzip` version specification.  It's part of Fastlane, although version 2, not 1 like we did. While that's a major version change, the breaking change is on a method we don't use, according to `git grep` in this and the release-toolkit repos.

See https://github.com/rubyzip/rubyzip/blob/v2.0.0/Changelog.md#200-2019-09-25

Also bumped our Fastlane release plugin to its latest version, 0.9.11, along the way, too.

## To test

---

PR submission checklist:


mokagio:
I milestoned this 15.4 despite it being so close to code freeze, because I think you might run in the same issues as I did with `deliver` so it would be great to get this in.

jkmassel:
Yeah, because I've merged #14518, I'll close this one without merging just so that there are no weird issues – this change will land in `develop` with the release branch.

Thanks @mokagio!

PR #14518, opened 2020-07-24
Update Fastlane to version 2.152.0

mokagio:
This mirrors #14516, and is here as a convenience in case you incur in the same `deliver` issue I did when finalizing this release.


jkmassel:
This was already at the second-most recent version, so I don't think I would've run into issues.

That said, this can't possibly hurt, so I'll land it prior to the release finalization just in case 😄 

Thanks @mokagio!!

PR #14524, opened 2020-07-24
Use system colors for tooltip and snackbar components

### LINKED ISSUE #14136: Proposal: Use system-defined grays for Tooltip and Snackbar components'
Over on the Mobile Gutenberg project, we've been working on a [RN-based tooltip](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2216) and I noticed that the new native Tooltip components for WPiOS (and WPAndroid) use [Color Studio](https://color-studio.blog) grays, which conflicts a bit with our preference of using system-defined grays. 

Considering both WPiOS and WPAndroid are both using system-defined grays, I think we might as well change the background-color on both the Tooltip and Snackbar components to use the following colors on iOS (from the [HIG](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/)), inverted for each mode (light mode using a dark mode gray and vice versa):

* Light mode: Use _dark-mode values_ of `systemGray5` or `systemGray6`
* Dark mode: Use _light-mode values_ of `systemGray5` or `systemGray6`

Note: we can't exactly use _dynamic_ system colors on RN, but we can target the default system grays separately with static hex or RGB values for each platform. So if this proposal is accepted, I'll likely propose doing so on RN. // cc @geriux 

CC'ing @mattmiklic @osullivanchris @bjtitus for feedback and discussion. 😄

Gio2018:
Fixes #14136

This PR changes the colors used in all Notices (Tooltips and snackbar components), as listed below:

 os | iOS 13 or later | iOS 12 
--- | --- | --- | 
**Background** | inverted systemGray5 | rgba(44.0, 44.0, 46.0, 1.0) |
**Primary title** | inverted label | white | 
**Secondary title/text** | inverted secondaryLabel | rgba(235.0, 235.0, 245.0, 0.6) |
**Action** | primary blue (shade40) |  primary blue (shade40) |

#### Note: inverted colors are system colors of the opposite appearance. For example, inverted systemGray5 means that in light appearance, we are using the value of `systemGray5` from dark appearance, etc

here are some screenshots for reference

element | light appearance | dark appearance
--- | --- | --- | 
**FAB** |  | 
**QuickStart step** |  | 
**QuickStart suggest** |  | 
**Reader actions** |  | 

To test:
- make a fresh install (so that the FAB tooltip will be shown)
- create a site and tap "yes help me" when prompted, so that the quick start will begin
- perform actions in the reader
- in all these steps, verify that the tooltips, snackbars and toast look as expected and do not have visual bugs in both dark and light appearance, as well as on iOS 12.

PR submission checklist:


WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
```diff
@@ -239,7 +239,11 @@ class NoticeView: UIView {
         cancelBackgroundView.addTrailingBorder()
 
         actionButton.titleLabel?.adjustsFontForContentSizeCategory = true
-        actionButton.setTitleColor(.white, for: .normal)
+        if #available(iOS 13.0, *) {
+            actionButton.setTitleColor(.label, for: .normal)
```
frosty, line 243: Could you move this to the NoticeStyle extension as another var? `actionButtonTitleColor` or something, and handle the `if #available` there?

frosty (approved):
Looks good to me! I'm not a designer, so I'll let them weigh in about the visual appearance :)

osullivanchris:
@Gio2018 @frosty 

Hey the colors look good...But I think you might have flipped light and dark mode? i.e. I expect a dark tooltip/snackbar on light mode, and a light one on dark mode. But I'm seeing light on light, and dark on dark. Unless I'm missing a design decision or other reason that we changed it!

Gio2018:
Hey @osullivanchris !

> Hey the colors look good...But I think you might have flipped light and dark mode? 

the `systemGray5` background, which is what is specified in #14136 adjusts automatically in light and dark mode (will be darker in dark mode, lighter in light mode). Should we replace this background with a different shade for the background and the tiltles? @iamthomasbishop, thoughts?

iamthomasbishop:
Ah good catch, @osullivanchris. That was my intention, and we discussed this early on but I must've missed the detail when reviewing the screenshots in isolation last week (my bad @Gio2018 !). The colors should be inverted:

* Light mode: dark-gray background (`systemGray5` from dark mode)
* Dark mode: light-gray background (`systemGray5` from light mode)
 
Note: text colors should also be inverted as a result.

Gio2018:
Oh I see now @osullivanchris and @iamthomasbishop ! 

Sorry about the confusion, I misread this in the original issue, it was stated that the colors were to be inverted.

The only issue is that, as far as I know, there is no way to automatically "invert" the system colors between dark and light appearance, so the solution would be to use **custom** colors, that look like the system colors in the opposite appearance. While this is perfectly feasible, we would not be really using system colors (which I thought it was one of the goals of #14136), because, for example, iOS in light appearance has no system-equivalent of `rgba[44.0, 44.0, 46.0, 1.0]`, which is `systemGray5` in dark mode.
Another alternative is to find, among the available system colors, those who match the inverted look that we are looking for. What do you think?

frosty:
@Gio2018 I think you should be able to make use of some of our existing UIColor convenience methods to add a 'flipped' version of e.g. systemGray5 to UIColor in an extension. For example:

```
extension UIColor {
    static var invertedSystem5: UIColor {
        let originalDark = UIColor.systemGray5.color(for: UITraitCollection(userInterfaceStyle: .dark))
        let originalLight = UIColor.systemGray5.color(for: UITraitCollection(userInterfaceStyle: .light))

        return UIColor(light: originalDark, dark: originalLight)
    }
}
```

This only takes care of iOS 13+ – you'd need to check for that and do something a little different for earlier OS versions. Do you think this could work?

Gio2018:
Hey @frosty !

>  I think you should be able to make use of some of our existing UIColor convenience methods to add a 'flipped' version of e.g. systemGray5 to UIColor in an extension
> Do you think this could work?

I believe it could. To be noted that even if we are using colors that are derived from system colors, these are, in fact, custom colors. 
Not an issue by itself, I just wanted to point this out. It's a subtle difference, but especially for semantically defined system colors (e.g. `label` which we are using for the primary text in notices) we might not be using it strictly as intended.

> Apple recommends against hard-coding the color values for their system colors, so the above would be preferable.

I agree, this implementation (for iOS 13+) has the advantage of not hardcoding RGB values. For iOS 12, I think the only option is instead to hardcode them?

WordPress/Classes/ViewRelated/System/Notices/NoticeView.swift
Gio2018: Sure!

Gio2018:
@osullivanchris , @iamthomasbishop , @frosty , colors are now updated to use their "inverted" version

osullivanchris (approved):
Hey looks great now @Gio2018 ! Thanks for the new build.

I can totally understand why this is not fully out-of-the-box as we're effectively using dark colors in light mode and light colors in dark mode. But everything looks good visually to me, and as I understand you were able to do it without much customisation. And we're using the same code across all tooltips and snackbars which is a win too.

iamthomasbishop:
Looks really solid overall, @Gio2018 ! One minor thing: are we using `blue 40` on both light/dark mode (I believe so based on the description above)? If so, can we use our standard link colors, which are `blue 50` in light mode and `blue 30` in dark mode? That'd help to provide a bit more contrast against the backgrounds. 

Other than that, everything looks good to go. Thank you!

Gio2018:
Hey @iamthomasbishop !
> are we using `blue 40` on both light/dark mode (I believe so based on the description above)? 

I'd say so, I kept the existing setting on this one
> If so, can we use our standard link colors, which are `blue 50` in light mode and `blue 30` in dark mode? 

Sure! I assume they should also be inverted like all the other colors in the notices? I am asking because if I don't invert them, they seem a little less readable. Here are example screenshots of both scenarios:


colors | light appearance | dark appearance
--- | --- | --- | 
light: shade50, dark: shade30 | light50Dark30_light | light50Dark30_dark
ight: shade30, dark: shade50 | light30Dark50_light | light30Dark50_dark

iamthomasbishop:
> I assume they should also be inverted like all the other colors in the notices? I am asking because if I don't invert them, they seem a little less readable.

@Gio2018 Ahh okay, if we're automatically inverting the colors, the second examples you shared look correct (`light: shade30, dark: shade50`).

frosty:
@Gio2018 I just noticed one other thing on this PR – in dark mode (light backgrounds) we lose the dividers between the content and buttons of the larger quick start notice:

Screenshot 2020-07-28 at 20 15 55

Is that something we could address separately?

Gio2018:
@frosty sure! I just merged this PR but I can open a new one for this specific issue. It looks like right now we are using `white` with `alpha = 0.25` for both light and dark appearance. @iamthomasbishop , @osullivanchris what color should we use for the dividers in dark mode/light notice backgrounds? Is the existing ok for dark mode or should we change it too?
Thanks!!

iamthomasbishop:
@Gio2018 can we use [Separator](https://developer.apple.com/documentation/uikit/uicolor/3173139-separator) (flipped between light/dark modes, obviously)? If not, can we use the equivalent rgba values or opaque rgb values (happy to provide those if needed)?

Gio2018:
> @Gio2018 can we use [Separator](https://developer.apple.com/documentation/uikit/uicolor/3173139-separator) (flipped between light/dark modes, obviously)? If not, can we use the equivalent rgba values or opaque rgb values (happy to provide those if needed)?

Sure we can! I'll do similarly to what I did for the other colors

PR #14525, opened 2020-07-24
Update topic page header UI 

yaelirub:
This PR updates the topic header according to designs. 
It also removes the "old" header from a filtered following stream since the filter bar already shows the topic and the user is already in the Following tab (screenshots to explain more)

 ### Filtered Following screen - removed header
Before             |  After
:-------------------------:|:-------------------------:
  |  

### Topic page header before
Follow              |  Following
:-------------------------:|:-------------------------:
  |  

### Topic page header After - Dark mode
Follow              |  Following
:-------------------------:|:-------------------------:
  |  

### Topic page header After - Light mode 
Follow              |  Following
:-------------------------:|:-------------------------:
  |  

To test:

Filtered Following:
1. Open the reader
2. Tap on filter 
3. Tap on Tags
4. Tap on a tag
5. See no header 
6. Remove filter and make sure everything works 

Topic page
1. Tap on an article tag 
2. See topic page according to new designs 
3. Make sure things look good in dark and light mode

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14536, opened 2020-07-29
Update site header UI 

yaelirub:
After updating the topic page header in #14525 , we are now giving the site header a fresh new look as well.

To test:
1. Go to reader 
2. Tap on a site name in a post card 
3. See the new site header

Before             |  After
:-------------------------:|:-------------------------:
  |  

### With a longer site description 
Screen Shot 2020-07-28 at 6 36 31 PM

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 15.5
 -----
+* [*] Reader: revamped UI for your site header.
  
 15.4
 -----
--- a/WordPress/Classes/ViewRelated/Reader/ReaderSiteStreamHeader.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderSiteStreamHeader.swift
@@ -51,24 +51,37 @@ fileprivate func > <T: Comparable>(lhs: T?, rhs: T?) -> Bool {
         WPStyleGuide.applyReaderSiteStreamCountStyle(followCountLabel)
     }
 
+    open override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
+                   preferredContentSizeDidChange()
+        }
+
+        if #available(iOS 13.0, *) {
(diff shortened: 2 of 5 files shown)


yaelirub:
hey @mbshakti , I updated the site header like we talked about. 
Two things to notice: 
1. We had the followers count in the previous UI so I put it under the image but if you want to get rid of it, we can. 
2. In your designs, you limit the description to 3 lines and a "...more" text that I assume is meant to expend to the full description. I didn't have the time to work on that so the header will show the full description when initialized. I don't see it as a blocker for this PR approval but if you see it as a deal breaker, I'll add a task to tackle this later on.

mbshakti:
Thanks @yaelirub! Agree that the "more" link isn't a blocker. 

Some nits:
- Can we make sure we are using the same background color for the blog header as the postcards? In light mode seems fine, but dark mode, it should match the postcard. I think it's using Elevated Primary color. 
- As discussed on slack, I updated the design to include the followers # below the follow button. [Figma link here](https://www.figma.com/file/pUAK0Ft8TxC1DyIbstP3MY/Untitled?node-id=0%3A1).

To @leandroalonso regarding your comment on the font sizes, do you mean it seems big just on iPad or generally?

leandroalonso:
@mbshakti in general, for example:



The description text is bigger than the blog information on the post card, for example. Anyway, if you're ok with it I'm ok too. :)

yaelirub:
Screen Shot 2020-07-29 at 11 26 30 AM

@mbshakti , @leandroalonso , updated according to code review and design updates

mbshakti:
Thanks! Looking good! One visual design nit, but not blocking. I noticed that in the latest screenshot you shared there seems to be uneven spacing at the top and bottom of the blog header card, [for example here](https://drive.google.com/file/d/1FVt-AOKpTattUPWuh3h0tIGGcY_-9bWZ/view?usp=sharing). I didn't notice earlier because the header faded into the background. 

Regarding the blog description font size, thanks for clarifying @leandroalonso -- I know what you mean! Since this screen is dedicated to the blog, and I think the font size being slighter bigger on the header is nice because it makes it feel more prominent, and doesn't feel like just another card on the screen.

yaelirub:
>to be uneven spacing at the top and bottom

Updated

PR #14532, opened 2020-07-28
Integration with Cards API

leandroalonso:
Part of #14399

WPKit related PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/266

This PR connects the new Discover session to the real cards API.

### To test

First, make sure you have the flag Reader Improvements Phase 2 **enabled**.

#### 1. Showing post cards

1. Go to discover
2. You should see post cards based on the interests you follow

#### 2. Tapping a post

1. Go to discover
2. Tap any post
3. Make sure the content appears correctly

#### 3. Loading more

1. Go to Discover
2. Scroll down
3. Checks that more posts are loaded

#### 4. No main-thread blockers

1. Go to discover
2. Scroll down
3. Checks that more posts are loaded and the UI is never blocked as it was before

#### 5. Localization

1. Change your device/simulator language
2. Go to Discover
3. Check that the posts match your language

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14539, opened 2020-07-29
Select topics refactoring

leandroalonso:
Part of #14399

Review only after #14532 is merged.

This PR doesn't add any new features. Instead, it refactors logics related to the topics selecting screen.

What was done:

1. Code moved from `ReaderStreamViewController` to `ReaderCardsViewController`
2. Decoupling between the Interests Coordinator and Reader screen
3. Removed the logic to track if the user has seen the selects interests screen or not (this isn't relevant anymore, we always show this screen if the user is not following any topic)
4. Removed a few view animations to simplify code understanding

#### 01 - Select Interests Picker
1. Login using a WP.com account
2. Launch the app
3. Delete all of your followed tags if you have any
4. Tap on the Reader tab
5. Tap on the 'Discover' tab

**Expectation:** You are presented with the Select Interests View

#### 02 - Save Interests: Loading View
1. Continuing with the steps from 01
2. Select some interests
3. Tap the 'Done' button

**Expectation:** The select interests view fades out, and you're presented with posts based on the interests you selected.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -135,35 +135,28 @@ private extension ReaderCardsStreamViewController {
     func displaySelectInterestsIfNeeded() {
         selectInterestsViewController?.userIsFollowingTopics { [unowned self] isFollowing in
             if isFollowing {
-                self.makeSelectInterestsViewControllerIfNeeded()
-                self.showSelectInterestsViewIfNeeded()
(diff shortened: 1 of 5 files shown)


yaelirub:
Works as expected when it comes to selecting topics and the logic for showing  the picker screen👍 
The only thing is that the Discover content doesn't get updated when Im removing and adding tags so I have to pull to refresh to see  the new content:

leandroalonso:
>  The only thing is that the Discover content doesn't get updated when Im removing and adding tags so I have to pull to refresh to see the new content:

I'll create an issue for it and fix it. :)

PR #14546, opened 2020-07-30
[TESTING] Multiple slider in Columns

lukewalczak:
This PR is created only for testing multiple sliders in columns via peril build.

🚧 DO NOT MERGE 🚧

To test:

PR submission checklist:


iamthomasbishop:
@lukewalczak Nice work! Adding the column width controls to the parent feels really intuitive, so what you've implemented here looks great and my concerns/feedback are mostly around Columns percentages in general. My big thing is still the unclear UX in terms of how columns' percentages relate to each other. I'll talk about that first and propose a workaround, then go into more general refinements I would like to see. All in all, I think there are maybe only 2 or 3 changes that we might want to consider.

## Percentages

This has been discussed at length, but I still feel strongly that how we're handling width adjustments (using the web's logic) is going to cause a *lot* of confusion, and while the column previews are definitely helpful to *visualize* the layout, I expect the percentage strings specifically will be quite confusing to users so we should solve this ASAP because it's *critical* to giving our users the best UX with regard to columns.

Tl;dr — until we're able to implement a more intuitive model, I think the *simplest thing* we could do is to hide the percentages text input on this parent Columns sheet, because I believe it'd simplify the interaction a bit. Allow me to explain...

Right now when I adjust a slider, my eyes tend to focus on the exact percentage of the column (the text input), which would be fine if the columns dynamically adapted to my adjustments. However, if we remove the text inputs, my eyes are drawn to the previews, which *are* updated dynamically as you change the sliders and are more accurate to the output — even if the position of each slider handle isn't 1:1 with the percentage.

The side effect of this approach would be that it's harder to make granular adjustments (e.g. adjusting the slider 1-3% would be hard). If you want to make a *granular* adjustment to a single column, you have to navigate to an individual column's settings sheet, which would require a few steps, but I think I'm okay with this as a trade-off because the benefits of a simpler UI likely outweigh the need of a granular adjustment. If this is too much friction, I have some ideas for how me might address it — let's decide whether this is the right approach before going down that path 🙂

## General columns refinements

- I think I may have asked this previously, but is there a way to limit the slider adjustments to whole numbers while still allowing decimals in the text input?
- I think we can change the "cancel" button on the columns layout picker sheet to our primary blue — I originally thought red would make sense in context, but it feels a bit too strong bc it's not a destructive action

iamthomasbishop:
@lukewalczak Testing the latest build, and the updates from the last round of feedback look pretty solid. Removing the percentages from the parent sliders feels at least a little bit _better_, despite our aforementioned constraints. One tiny additional thing that might make the width adjustments slightly better is to remove the word "Percentage" from the child (Column) block settings -- because as we've discussed, the percentages aren't always very clear, and so simply labeling the slider cells as "Width" or "Column width" might have a similar affect. I understand we're getting this label from the web side, but if this is doable I think it will help.

PR #14549, opened 2020-07-30
Fix flaky unit tests

leandroalonso:
This PR fixes two tests that are failing in some machines but not on the CI:

* `ReaderSelectInterestsCoordinatorTests.testShouldDisplayReturnsTrue`: the order of tests being executed in this suite was causing false negatives because database changes were being persisted;
* `ReaderPostCardCellTests.testHeaderLabelMatchesExpectation`: this one was failing because of the timezone. Now we use a hardcoded timezone in the tests.

### To test

1. Run the whole unit tests suite in your machine
2. Make sure it's all green

PR submission checklist:


mindgraffiti (approved):
✅ works on dirty build and clean installs

Thank you!

:shipit:

PR #14556, opened 2020-07-31
ULS: Updating Auth to use new 2FA view for Apple

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/336
Auth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/351

This updates Auth to show the new 2FA view for Apple accounts that:
- _Do not_ have a WP password set.
- _Do_ have WP 2FA enabled.

Note the new `unifiedApple` feature flag is disabled by default.

To test:

---
- Select `Continue with Apple` from either login or signup.
- Select an Apple account that:
  - Does not have a WP password set.
  - Does have WP 2FA set.
- Verify the old 2FA view is displayed.

old

---
- Enable the `unifiedApple` feature flag.
- Go through the same steps as above.
- Verify the new 2FA view is displayed.

<kbd>new_</kbd>

---
PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -186,9 +186,9 @@ target 'WordPress' do
 
     pod 'Gridicons', '~> 1.0.1'
 
-    # pod 'WordPressAuthenticator', '~> 1.22.0-beta'
+    pod 'WordPressAuthenticator', '~> 1.22.0-beta'
     # While in PR
-    pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => 'feature/336-apple_2fa_view'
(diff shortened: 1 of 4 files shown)


mindgraffiti (approved):
Tested on an iPhone 11 Pro w/an Apple account w/no password + 2fa enabled. Whew!

✅ Old UI does not break
✅ New UI works as expected
🔢  The only thing I could not confirm was if the "Continue" button spins after pressing it. It navigated really quickly from 2fa to the epilogue screen.

:shipit:

ScoutHarris:
> The only thing I could not confirm was if the "Continue" button spins after pressing it.

I'll take a look. If I did it incorrectly, I've done it incorrectly in the next view. So I'll verify them both in one go. Thanks @mindgraffiti !

PR #14557, opened 2020-07-31
add ghost view on reader tab bar at the first load

### LINKED ISSUE #14412: Reader: main stream placeholder and content load too slowly

Gio2018:
Fixes #14412 (combined with #14454)

To test:
- delete and reinstall or logout/login
- go to the reader
- verify that a ghost animation is displayed in the tab bar area until the menu titles appear (along with the rest of the UI)

tip: to make the loading more noticeable, use a slow network (like in the example below)
<p align=center>

</p>

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -11,6 +11,7 @@ class ReaderTabView: UIView {
     private let verticalDivider: UIView
     private let horizontalDivider: UIView
     private let containerView: UIView
+    private var loadingView: UIView?
 
     private let viewModel: ReaderTabViewModel
 
@@ -39,6 +40,7 @@ class ReaderTabView: UIView {
             self?.tabBar.items = tabItems
             self?.tabBar.setSelectedIndex(index)
             self?.configureTabBarElements()
+            self?.hideGhost()
             self?.addContentToContainerView()
         }
         setupViewElements()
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
Thanks for fixing that @Gio2018! :shipit:

osullivanchris:
Hey @leandroalonso @Gio2018 is the idea that this will play well with #14485 I assume?

When both are merged, I would expect when I land on the tab to see both the tabs and the page area loading, and then the tabs and content appear as they are available.

Just checking if that is correct?

For example, the issue I am trying to avoid. It would be bad if the tabs loaded first, then were shown. And only after that the page loading state started. I think everything should show as loading as soon as the user loads the view. Tricky to explain, I hope that makes sense.

leandroalonso:
> For example, the issue I am trying to avoid. It would be bad if the tabs loaded first, then were shown.

The PR you referenced has this code currently in another branch, so it's not yet available. However, when we have everything in one place we can add "ghost cards" here too.

However, even with the lack of it right now I'd like to move forward with that solution since the current one seems broken (the blank screen). Wdyt @osullivanchris?

frosty (approved):
Happy to move ahead here, @leandroalonso – but I agree this will be really nice when we get ghosted cards in the main view as well! The ghosting menu gives an indication that _something_ is happening, but it's still a little empty 🙂

osullivanchris (approved):
agree with all the above!

Issue #14561, opened 2020-08-03
Image Upload Problems


azollafox:
### Expected behavior
When a feature image is added it should upload successfully to the post when published

### Actual behavior
>when we try to post featured image .. it shows the the upload done successfully however when we publish the post it doesn’t appear so when again we go back to edit we could not see the featured image so we have to upload it again ..

### Steps to reproduce the behavior
Not reproduced yet

#####Tested on iPad 7, iOS 13.5.1, WPiOS 15.3
User reported on 15.2 on multiple devices

#3189303-zen
#3175647-zen

designsimply:
Noting the site types and editors in use (at least according to the logs sent at the time the tickets were created) for reference:

* 3189303 - Atomic, Gutenberg
* 3175647 - Jetpack Personal, Aztec (note: they said they are having similar upload problems in Media as well)

And here are some of the snippets from the app logs mentioning media errors in case they're helpful:

```
2020-07-31 22:24:43:209 🔵 Tracked: editor_session_start <blog_type: jetpack, content_type: gutenberg, editor: gutenberg, has_unsupported_blocks: 0, post_type: post, session_id: 9381F1FE-72CF-4135-83F5-EBB265D694E6, startup_time_ms: 928, unsupported_blocks: (
)>
2020-07-31 22:24:48:471 <PostSettingsViewController: 0x10349e800> viewDidLoad
2020-07-31 22:24:48:472 🔵 Tracked: post_settings_shown <>
2020-07-31 22:24:53:749 🔵 Tracked: editor_post_featured_image_changed <action: removed, via: settings>
2020-07-31 22:24:54:613 <WordPress.WordPressAppDelegate: 0x281003400> applicationWillResignActive(_:)
2020-07-31 22:24:55:083 Stats Period: finished persisting Period Stats to disk.
2020-07-31 22:24:55:708 <WordPress.WordPressAppDelegate: 0x281003400> applicationDidEnterBackground(_:)
2020-07-31 22:24:55:717 🔵 Tracked: application_closed <last_visible_screen: Post Editor, time_in_app: 17>
2020-07-31 22:24:55:721 PingHub disconnecting
2020-07-31 22:24:56:218 PingHub disconnected: WSError(type: Starscream.ErrorType.protocolError, message: "", code: 1000)
2020-07-31 22:24:56:223 🔵 Tracked: media_service_upload_started <blog_id: 178491266>
[...]

PR #14563, opened 2020-08-03
Removes a silent failure.

diegoreymendez:
This is an attempt to fix what looks like a silent failure here: https://github.com/wordpress-mobile/WordPress-iOS/issues/14561

Unfortunately I wasn't able to reproduce the issue, but there's a chance this change could at least raise the error so that both our users and the App detect it.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/MediaCoordinator.swift
+++ b/WordPress/Classes/Services/MediaCoordinator.swift
@@ -305,9 +305,23 @@ class MediaCoordinator: NSObject {
                             success: {
                                 self.end(media)
         }, failure: { error in
-            guard let nserror = error as NSError? else {
-                return
-            }
+            // Ideally the upload service should always return an error.  This may be easier to enforce
+            // if we update the service to Swift, but in the meanwhile I'm instantiating an unknown upload
+            // error whenever the service doesn't provide one.
+            //
+            let nserror = error as NSError?
+                ?? NSError(
+                    domain: MediaServiceErrorDomain,
+                    code: MediaServiceError.unknownUploadError.rawValue,
+                    userInfo: [
+                        "filename": media.filename ?? "",
+                        "filesize": media.filesize ?? "",
+                        "height": media.height ?? "",
+                        "width": media.width ?? "",
+                        "localURL": media.localURL ?? "",
+                        "remoteURL": media.remoteURL ?? "",
+                ])
+
             self.coordinator(for: media).attach(error: nserror, toMediaID: media.uploadID)
             self.fail(nserror, media: media)
         })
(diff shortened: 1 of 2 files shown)


diegoreymendez:
Heads up @loremattei and @jkmassel  since this is aimed at 15.4, which is currently frozen.

diegoreymendez:
@leandroalonso - Since @SergioEstevao is on extended AFK I'll need someone else to review this change.  Can I ask you for a review?

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -305,9 +305,23 @@ class MediaCoordinator: NSObject {
                             success: {
                                 self.end(media)
         }, failure: { error in
-            guard let nserror = error as NSError? else {
-                return
-            }
+            // Ideally the upload service should always return an error.  This may be easier to enforce
+            // if we update the service to Swift, but in the meanwhile I'm instantiating an unknown upload
```
leandroalonso, line 309: There's no way to enforce an error in the Obj-C code?

WordPress/Classes/Services/MediaCoordinator.swift
```diff
@@ -305,9 +305,23 @@ class MediaCoordinator: NSObject {
...
+            // error whenever the service doesn't provide one.
+            //
+            let nserror = error as NSError?
+                ?? NSError(
```
leandroalonso, line 313: The idea is to attach this error so we can see that it's happening? Because in case of the `unknownUploadError` we don't have information about what happened, right?

Anyway, I agree with you. Is better than just ignoring it...

leandroalonso:
@diegoreymendez do we have the information if the user is experiencing the main issue while trying to upload an image from the phone or while selecting an existing media from the WP library?

WordPress/Classes/Services/MediaCoordinator.swift
diegoreymendez: Nopes, the idea is to avoid silencing errors.

Ideally this scenario should never be possible and the APIs we use should make sure an error is returned whenever there is one, but that's beyond the scope of this PR.

WordPress/Classes/Services/MediaCoordinator.swift
diegoreymendez: But I didn't say that.  I just said it's easier to enforce in Swift because the nullability of types isn't optional.  You have to clearly define if something is optional or not, whereas in Obj-C by default you can just avoid specifying anything.

diegoreymendez:
@leandroalonso - The tickets I saw all referenced setting the featured image from the post settings.  The log I saw matched that.

Btw, keep in mind I wasn't able to repro the issue, but the change in this PR is good regardless of if it fixes the referenced issue.  We should really not be silencing errors anywhere in our code, and a no-op `return` statement is really not great to have when things fail.

leandroalonso:
@diegoreymendez my question is: users are uploading images stored on their device or using the WP media library (aka an image already uploaded in the blog).

Because these two options go through different patches.

Anyway, the code here looks good. :shipit:

diegoreymendez:
Thanks @leandroalonso - let's keep chatting about your latest point via DM.

PR #14566, opened 2020-08-03
Ensure cookies for private sites when using the Reader

### LINKED ISSUE #14564: Reader fails to load Media
When you install the app and access the Reader, images won't display unless you open any other webview in the app.

### Steps to reproduce

1. Do a fresh install of the app
2. Log in into your A8C account
3. Go to the Reader
4. Go to "Automattic"
5. Open a post with images
6. Check that images won't load

This happens not only for A8C internal blogs but also for any private Atomic site.

leandroalonso:
Fixes #14564

### To test

1. Do a fresh install of the app
2. Log in into your A8C account
3. Go to the Reader
4. Go to "Automattic"
5. Open a post with images
6. Check that images load correctly

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
@@ -171,14 +171,14 @@ class ReaderDetailCoordinator {
     /// Use this method to make sure these cookies are downloaded.
     /// - Parameter webView: the webView where the post will be rendered
     /// - Parameter completion: a completion block
-    func authenticateIfNeeded(in webView: WKWebView, completion: @escaping () -> Void) {
+    func storeAuthenticationCookies(in webView: WKWebView, completion: @escaping () -> Void) {
         guard let postURLString = post?.permaLink,
             let postURL = URL(string: postURLString) else {
             completion()
             return
         }
 
-        authenticatedRequest(for: postURL, on: webView) { _ in
+        storeAuthenticationCookies(for: postURL, cookieJar: webView.configuration.websiteDataStore.httpCookieStore) {
             completion()
         }
     }
--- a/WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
@@ -19,6 +19,15 @@ class ReaderDetailCoordinator {
     /// Called if the view controller's post fails to load
     var postLoadFailureBlock: (() -> Void)? = nil
 
+    /// A Authenticator for private posts that needs special webview cookies
+    lazy var authenticator: RequestAuthenticator? = {
+        guard let account = AccountService(managedObjectContext: coreDataStack.mainContext).defaultWordPressComAccount() else {
+            return nil
+        }
+
+        return RequestAuthenticator(account: account)
+    }()
+
     /// Core Data stack manager
     private let coreDataStack: CoreDataStack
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -156,6 +165,23 @@ class ReaderDetailCoordinator {
...
+    /// Some posts have content from private sites that need special cookies
+    ///
+    /// Use this method to make sure these cookies are downloaded.
+    /// - Parameter webView: the webView where the post will be rendered
+    /// - Parameter completion: a completion block
+    func authenticateIfNeeded(in webView: WKWebView, completion: @escaping () -> Void) {
+        guard let postURLString = post?.permaLink,
+            let postURL = URL(string: postURLString) else {
+            completion()
+            return
```
diegoreymendez, line 177: Same question as above... is this an acceptable / normal / regular scenario?  Otherwise consider loggin an error here.

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -156,6 +165,23 @@ class ReaderDetailCoordinator {
...
+        }
+
+        authenticatedRequest(for: postURL, on: webView) { _ in
```
diegoreymendez, line 180: This was really hard to understand at first: we ask for an authenticated request and ignore the result! 🤔

It took me a while to understand.

I stopped to think about this a bit more and realized the method we're calling here does more than just authenticating a request - and we're actually calling it for that side effect, which isn't really documented and isn't self-evident from the method name.

Consider how the code would look if we, instead, had a line like this:

```swift
authenticator.storeAuthenticationCookies(for: postURL, in: webView.configuration.websiteDataStore.httpCookieStore)
```

And then have in Authenticator a new method, looking kind of like this:

```swift
func storeAuthenticationCookies(for url: URL, cookieJar: CookieJar, completion: () -> ()) {
    authenticator.request(url: url, cookieJar: cookieStore) { _ in
        completion()
    }
}
```

I think this would greatly improve the readability of your code.  Just my 2 cents.

diegoreymendez (commented):
Added some comments for your consideration.

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -19,6 +19,15 @@ class ReaderDetailCoordinator {
     /// Called if the view controller's post fails to load
     var postLoadFailureBlock: (() -> Void)? = nil
 
+    /// A Authenticator for private posts that needs special webview cookies
+    lazy var authenticator: RequestAuthenticator? = {
+        guard let account = AccountService(managedObjectContext: coreDataStack.mainContext).defaultWordPressComAccount() else {
+            return nil
```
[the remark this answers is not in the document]
leandroalonso: Done in 3631d2695041e8aa5881df2ae666d65057be3b57

I used `DDLog` instead of `CrashLogging.logError()` as we talked.

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
```diff
@@ -19,6 +19,15 @@ class ReaderDetailCoordinator {
     /// Called if the view controller's post fails to load
     var postLoadFailureBlock: (() -> Void)? = nil
 
+    /// A Authenticator for private posts that needs special webview cookies
```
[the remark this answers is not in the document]
leandroalonso: Done in d09dd1b8beb3d21ab4800104820e0fcef9343ac5

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: I'd say yes. If the permalink doesn't exist the post will be rendered either way.

WordPress/Classes/ViewRelated/Reader/ReaderDetailCoordinator.swift
leandroalonso: Loved this feedback, it was indeed confusing.

I changed in 6cdea72021b997c9687cc6939b776912c36a0952

Let me know what do you think: is it more clear?

diegoreymendez (approved):
Nice work!

PR #14569, opened 2020-08-04
Enable @mentions feature in all flavors for version 15.4

maxme:
The gutenbergMentions feature flag is always true, but we'll keep the flag in case we want to revert it quickly.

Note: the feature was enabled and tested in pre-releases. This doesn't require a new beta release.


maxme:
@mchowning I update the release notes - also fixed the targeted branch (was `develop` it's now `release/15.4`).

PR #14573, opened 2020-08-04
ULS: update Auth to show email and gravatar on Password view

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/352
Auth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/359

To test:

---
- Enable the `unifiedGoogle` feature flag.
- Select `Continue with Google` from either login or signup.
- Select a Google account that:
  - Is disconnected from WP.com. WP.com > Me > Security > Social Logins > Google > disconnect
  - Has a WP password set.
  - Does not have a gravatar.
- On the Password view, verify the placeholder icon and the email is displayed:

<kbd>no_gravatar</kbd>

---
- Select a Google account that:
  - Is disconnected from WP.com. WP.com > Me > Security > Social Logins > Google > disconnect
  - Has a WP password set.
  - Does have a gravatar.
- On the Password view, verify the gravatar and the email are displayed:

<kbd>gravatar</kbd>

---
PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14578, opened 2020-08-05
Relax WordPressKit and WordPressShared version control

mindgraffiti:
Diego I'm so sorry! I forgot to explain the WordPressKit / WordPressShared dance that happens between WordPressAuthenticator and WPiOS. When a dependency is updated in Auth that is also a dependency used in WPiOS, they have to be updated at the same time, otherwise it will break pods on `develop` for everyone else. In this case the conflict was that the WPiOS podfile had strict and outdated versions defined for WordPressKit and WordPressShared, while Auth did not.

### To test
To see the errors on `develop`:
1. Check out develop
2. `rm -rf pods`
3. `rake dependencies`
4. Build and run
5. Errors that `AnalyticsEvent` class cannot be found, build fails

To test the fix:
1. Check out this branch
2. `rm -rf pods`
3. `rake dependencies`
4. Build and run (should successfully build now)

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -10,7 +10,7 @@ workspace 'WordPress.xcworkspace'
 ## ===================================
 ##
 def wordpress_shared
-    pod 'WordPressShared', '1.9.2'
+    pod 'WordPressShared', ' ~> 1.10.-beta'
     #pod 'WordPressShared', :git => 'https://github.com/wordpress-mobile/WordPress-iOS-Shared.git', :tag => ''
     #pod 'WordPressShared', :git => 'https://github.com/wordpress-mobile/WordPress-iOS-Shared.git', :branch => ''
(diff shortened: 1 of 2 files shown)


ScoutHarris (changes requested):
Hey Thuy. I've already made these changes in my PR https://github.com/wordpress-mobile/WordPress-iOS/pull/14573/. We can just use mine if you like, or use yours (after one change).

Podfile
```diff
@@ -10,7 +10,7 @@ workspace 'WordPress.xcworkspace'
 ## ===================================
 ##
 def wordpress_shared
-    pod 'WordPressShared', '1.9.2'
+    pod 'WordPressShared', ' ~> 1.10.-beta'
```
ScoutHarris, line 13: This should be '~> 1.10-beta'.

mindgraffiti:
Closing since @ScoutHarris  already incorporated these changes in an earlier PR. Thanks Stephenie!

Issue #14594, opened 2020-08-06
Add What's New entry in App Settings


Gio2018:
After dismissing the "What's New" modal (more details in #14581), users can review it at any time, by accessing the related entry in "App Settings".

Below, the flow to access it, extracted from the design specs:

<img width="1280" alt="Screen Shot 2020-08-06 at 3 09 16 PM" src="https://user-images.githubusercontent.com/34376330/89577796-d1341700-d7f6-11ea-8385-87d5c97e6128.png">

Gio2018:
Addressed by #14597

PR #14597, opened 2020-08-07
Add What's New entry in App Settings

Gio2018:
Ref #14594

Adds the "What's New" Entry in app settings and presents a modal that will become the Feature Announcement screen (see the flow in #14594).

**Note: this PR only sets the basis. Things that will be addressed in future PRs:**

- Content of the "What's New" screen: at the moment, only the "continue" button is there (and actionable)
- Size of the modal on iPad: at the moment it uses the standard size and does not reflect the designs
- Accessibility
- Unit tests

To test:

- go to 'Me' -> 'App Settings'
- scroll to the bottom and verify that there is a "What's New in WordPress" row
- tap on it and verify that a modal with a 'Continue' button is presented (see below)
- tap 'Continue' and verify that the modal is dismissed

<p align=center>

</p>

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ Not in this PR
- [x] ~~I have considered adding accessibility improvements for my changes.~~ Not in this PR

### DIFF
--- a/WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift
@@ -302,6 +302,14 @@ class AppSettingsViewController: UITableViewController {
             ActionDispatcher.dispatch(NoticeAction.post(notice))
         }
     }
+
+    func presentWhatIsNew() -> ImmuTableAction {
+        return { [weak self] row in
+            let controller = WhatIsNewViewController()
+            self?.present(controller, animated: true)
+            self?.tableView.deselectSelectedRowWithAnimation(false)
+        }
+    }
 }
 
 // MARK: - SearchableActivity Conformance
@@ -461,6 +469,8 @@ private extension AppSettingsViewController {
             action: pushAbout()
         )
 
+        let whatIsNewRow = NavigationItemRow(title: NSLocalizedString("What's New in WordPress", comment: "Opens the What's New / Feature Announcement modal"), action: presentWhatIsNew())
+
         var rows: [ImmuTableRow] = [settingsRow, aboutRow]
         if #available(iOS 10.3, *),
             UIApplication.shared.supportsAlternateIcons {
@@ -469,6 +479,10 @@ private extension AppSettingsViewController {
 
         if FeatureFlag.debugMenu.enabled {
             rows.append(debugRow)
+        }
+
+        if FeatureFlag.whatIsNew.enabled {
+            rows.append(whatIsNewRow)
         }
 
         if #available(iOS 13.0, *) {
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/What's New/WhatIsNewViewController.swift
```diff
@@ -0,0 +1,66 @@
+
```
mindgraffiti, line 1: Please consider documenting this class, it would help me to understand what's going on :)

WordPress/Classes/ViewRelated/What's New/WhatIsNewViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+import UIKit
+
+
+class WhatIsNewViewController: UIViewController {
```
mindgraffiti, line 5: For example, you have a great summary in the PR that could go here:

```suggestion
/// WhatIsNewViewController: adds the "What's New" Entry in app settings and presents a modal that will become the Feature Announcement screen
///
class WhatIsNewViewController: UIViewController {
```

WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift
```diff
@@ -302,6 +302,14 @@ class AppSettingsViewController: UITableViewController {
             ActionDispatcher.dispatch(NoticeAction.post(notice))
         }
     }
+
+    func presentWhatIsNew() -> ImmuTableAction {
+        return { [weak self] row in
+            let controller = WhatIsNewViewController()
```
diegoreymendez, line 308: If you add a `guard let self = self else { return }` above this line, we can avoid instantiating the VC when self is `nil`.

Definitely not necessary for the code to behave though, so feel free to ignore if you prefer this way.

jkmassel:
Today is the `15.5` code freeze, so I've bumped this to `15.6`. If it needs to be shipped as part of `15.5`, please merge it to the `release/15.5` branch, then DM me – I'll be happy to cut a new beta release.

Gio2018:
> Today is the `15.5` code freeze, so I've bumped this to `15.6`. If it needs to be shipped as part of `15.5`, please merge it to the `release/15.5` branch, then DM me – I'll be happy to cut a new beta release.

Thanks so much @jkmassel ! I forgot to do it myself before going AFK.

WordPress/Classes/ViewRelated/What's New/WhatIsNewViewController.swift
Gio2018: Absolutely right! I'll add some description to both `WhatIsNewViewController` and `WhatIsNewView`

WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift
Gio2018: Totally agree! Good catch!!

PR #14598, opened 2020-08-07
Add hit area to suggestions UI to allow dismissing on smaller devices

guarani:
Addresses https://github.com/WordPress/gutenberg/issues/24425

## Description

The PR adds a hit area with a minimum height of 50 points to the suggestion UI list, which allows the UI to be dismissed no matter what the screen size. Currently, on smaller phones such as the iPhone 7, the suggestions UI takes up all available screen space and there is no room to tap outside it to dismiss.

## Screenshots

**Note: the gap in the second screenshot shows the hit area this PR adds to facilitate dismissing the list.**

| Before | After |
| -- | -- |
|  |  |

## Testing instructions

_You'll need to test with a smaller device such as an iPhone 7 and a site with enough users that the suggestion UI fills the screen (roughly 7 in my tests)_
1. Open a post
2. Type @ to trigger the mentions UI
3. Notice there is now a gap between the navigation bar and the top username in the list of suggestions
4. Tap anywhere in this gap and notice the suggestions UI dismisses

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMentionsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMentionsViewController.swift
@@ -4,6 +4,7 @@ import UIKit
 public class GutenbergMentionsViewController: UIViewController {
 
     static let mentionTriggerText = String("@")
+    static let minimumHeaderHeight = CGFloat(50)
 
     public lazy var backgroundView: UIView = {
         let view = UIView(frame: .zero)
@@ -153,6 +154,10 @@ extension GutenbergMentionsViewController: SuggestionsTableViewDelegate {
 
     public func suggestionsTableViewDidTapHeader(_ suggestionsTableView: SuggestionsTableView) {
         onCompletion?(.failure(buildErrorForCancelation()))
+    }
+
+    public func suggestionsTableViewHeaderMinimumHeight(_ suggestionsTableView: SuggestionsTableView) -> CGFloat {
+        return Self.minimumHeaderHeight
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Suggestions/SuggestionsTableView.h
+++ b/WordPress/Classes/ViewRelated/Suggestions/SuggestionsTableView.h
@@ -67,5 +67,10 @@
 /// @param suggestionsTableView the suggestion view.
 - (void)suggestionsTableViewDidTapHeader:(nonnull SuggestionsTableView *)suggestionsTableView;
 
+/// This method returns the header view minimum height.
+/// Can be used as a hit area for the user to dismiss the suggestions list.
+/// @param suggestionsTableView the suggestion view.
+- (CGFloat)suggestionsTableViewHeaderMinimumHeight:(nonnull SuggestionsTableView *)suggestionsTableView;
+
 
 @end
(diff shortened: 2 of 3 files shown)


guarani:
Thanks for the review @etoledom! I'd overlooked testing with landscape mode. Looks like that and swipe to dismiss could be worked on in the future.

I'd also love to hear @iamthomasbishop's thoughts on adding swipe-to-dismiss to the suggestions UI component. It would be great for the in Gutenberg editor suggestions UI. We could also evaluate it for Reader Comments as well, although in full screen mode it might be best disabled.

iamthomasbishop:
> I have a feeling that we should allow swipe down to dismiss

@guarani  @etoledom This is something I proposed during the early stages of the project, because I had hoped we could take an approach that was similar to a bottom sheet (which allows swipe-to-dismiss, obviously). However, bc of timeline constraints, we decided to stay closer to what existed at the time. So yes, swipe-to-dismiss would be fantastic in the next iterations, but it wasn't in the cards for this first iteration. 😄

In the meantime (for the next iteration), I think we could consider one of the following options as a sort of "dismiss" action, and I'm curious what y'all think:

<a href="https://user-images.githubusercontent.com/1200432/89683370-06f00300-d8be-11ea-980f-6156008bd7f2.png"></a>

1. Adds a "done" button on the same row as the text input
2. Adds a keyboard-hide button (like the one we have on the block toolbar)
3. Adds a "done" button in the same row as the keyboard's auto-suggestions

(Note: If we were to do either of the first two, I think we can safely hide the keyboard auto-suggestions, because it doesn't look like they're of much help in this context.)

Another word of caution, because this component also is shared with notifications/commenting, let's make sure everything we change works on that UI. From the looks of it, the max-height fix might have caused an issue when the commenting sheet is open on notifications. Screenshot:

<a href="https://user-images.githubusercontent.com/1200432/89684396-f17bd880-d8bf-11ea-964b-cb27154c077a.jpeg"></a>

guarani:
Thanks for sharing the designs @iamthomasbishop, they look very good — my favorite is the keyboard close button. 
Any thoughts on the above @mchowning?

Regarding the cut-off cell seen in the screenshot you shared above, I checked and this existed before this PR was made.

mchowning:
All those options look good to me as well. Nice job @iamthomasbishop ! :bow: Personally, I favor Options 1 or 2, but I don't feel strongly.

Are we considering this for Android as well, or just iOS?

iamthomasbishop:
After looking again, I'm not sure if the keyboard-hide one (option 2) makes sense in this context, because the action doesn't just hide the keyboard, it dismisses the entire UI. With that, I'm now torn between option 1 and 3. I think option 3 looks most intuitive to my eyes, especially when I take into account how this would look on Notifications. The inline "done" button on option 1 above could conflict with the "send" button that exists on the right side of that input, and I think positioning the "done" button on another row adds some separation and emphasis.

> Are we considering this for Android as well, or just iOS?

@mchowning I think we discussed last week that we might not need the same thing on Android because the system back button would essentially do the same thing, but it only works that way in Notifications -- not the editor. Currently if I tap the system back button, the keyboard dismisses, but the UI stays visible -- whereas if you're at-mentioning someone in Notifications and tap the system back button, the UI _does_ dismiss. Maybe we should do the same in the editor on Android?

PR #14601, opened 2020-08-07
Present Select Interests fullscreen

leandroalonso:
This PR has two different (but inter-connected) changes:

* It displays the Select Interests fullscreen
* It doesn't wait for a request to finish to show the Reader and preload the topics when the user login

### Showing the Reader right away

1. Open the app
2. Tap Reader
3. Check that it appears right away, without waiting for a request with the server to finish

#### Logged out

1. Do a fresh install of the app
2. Login with a self-hosted site
3. Tap the Reader, check that the Interests screen pops up
4. Select at least one topic
5. Tap Done and check that the Discover session is shown

#### Logged in

1. You can either do a fresh install or login into an account
2. Tap Reader
3. Make sure to remove any topic you're already following
4. Tap "My Site" and then Reader again
3. Check that the Interests screen pops up
4. Select at least one topic
5. Tap Done and check that the Discover session is shown

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -35,7 +35,7 @@ class ReaderTabView: UIView {
             self?.toggleButtonsView()
         }
 
-        viewModel.onDidChangeTabBar { [weak self] tabItems, index in
+        viewModel.onTabBarItemsDidChange { [weak self] tabItems, index in
             self?.tabBar.items = tabItems
             self?.tabBar.setSelectedIndex(index)
             self?.configureTabBarElements()
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -47,6 +47,16 @@ class ReaderTabView: UIView {
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    func selectDiscover() {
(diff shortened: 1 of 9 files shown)


WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabViewModel.swift
```diff
@@ -7,7 +7,7 @@ import WordPressFlux
     /// tab bar items
     private let tabItemsStore: ItemsStore
     private var subscription: Receipt?
-    private var setTabBarItems: (([ReaderTabItem], Int) -> Void)?
+    private var didChangeTabBarItems: [(([ReaderTabItem], Int) -> Void)] = []
```
emilylaguna, line 10: See comment above

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -35,18 +35,30 @@ class ReaderTabView: UIView {
             self?.toggleButtonsView()
         }
 
-        viewModel.refreshTabBar { [weak self] tabItems, index in
+        viewModel.onDidChangeTabBar { [weak self] tabItems, index in
```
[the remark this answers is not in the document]
leandroalonso: Done in cc545c4a151424fbb690776d948fe1d09a3351fb

Much better :)

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabViewModel.swift
leandroalonso: cc545c4a151424fbb690776d948fe1d09a3351fb

PR #14627, opened 2020-08-12
Reader: fix unfollowed sites not being removed from list of followed sites

### LINKED ISSUE #14387: New Reader: Unfollowed sites are not removed.
### Expected behavior

I would expect that a site unfollowed from the browser would be removed from the app followed site list. 

### Actual behavior

The site remains in the list and when I try to unfollow it I get an error message "Could not unfollow site. You are not subscribed to the URL" and the site remains on the list. 

1. Log into WordPress.com in a browser and go to your reader followed site list. 
2. Select a site to unfollow. 
3. Go to the reader in the app and click the gear icon to switch to "Followed Sites". 
4. Pull down to refresh the list, the site will still display. 
5. Click on the icon to unfollow the site. 
6. For a second the site is removed, then the error message pops-up and the site returns to the list. 

Logging out of WordPress.com and back in through the app cleared the site from the list. 

##### Tested on iPhone X, iOS 13.5.1, WPiOS 15.0.1.

Reported by a user in 3074502-zen
Youtube video showing the issue listed in the ticket.

momo-ozawa:
Fixes #14387

### Changes:
- If a site is unfollowed from the browser, remove it from the app's followed site list
  - Keep track of sites fetched from remote
  - Compare them against sites fetched from CoreData
  - If a site fetched from CoreData isn't in the the list of sites fetched from remote, mark that site as being unfollowed

### To test:

1. Log into WordPress.com in a browser and go to your reader followed site list.
2. Select a site to unfollow.
3. Go to the reader in the app and click the gear icon to switch to "Followed Sites".
4. Pull down to refresh the list.
5. Check that the site unfollowed from the browser is removed from the app followed site list ✅   

unfollow-and-remove-from-followed-sites

### PR submission checklist:

### DIFF
--- a/WordPress/WordPressTest/ReaderTopicServiceTest.swift
+++ b/WordPress/WordPressTest/ReaderTopicServiceTest.swift
@@ -95,6 +95,25 @@ final class ReaderTopicSwiftTest: XCTestCase {
         }
     }
 
+    func remoteSiteInfoForTests() -> [RemoteReaderSiteInfo] {
+        let foo = RemoteReaderSiteInfo()
+        foo.feedID = 1
+        foo.isFollowing = true
+        foo.postsEndpoint = "/sites/foo"
+
+        let bar = RemoteReaderSiteInfo()
+        bar.feedID = 2
+        bar.isFollowing = true
+        bar.postsEndpoint = "/sites/bar"
+
+        let baz = RemoteReaderSiteInfo()
+        baz.feedID = 3
+        baz.isFollowing = true
+        baz.postsEndpoint = "/sites/baz"
+
+        return [foo, bar, baz]
+    }
+
     func remoteTopicsForTests() -> [RemoteReaderTopic] {
         let foo = RemoteReaderTopic()
         foo.topicID = 1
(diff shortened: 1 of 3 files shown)


WordPress/WordPressTest/ReaderTopicServiceTest.swift
```diff
@@ -138,6 +157,51 @@ final class ReaderTopicSwiftTest: XCTestCase {
...
+    func testUnfollowedSiteIsUnfollowedDuringSync() {
+        guard let context = context else {
+            XCTFail("Context is nil")
+            return
+        }
+        let remoteSites = remoteSiteInfoForTests()
+
+        // Setup
+        var expect = expectation(description: "sites saved expectation")
+        let service = ReaderTopicService(managedObjectContext: context)
```
leandroalonso, line 173: I guess you can remove the whole `guard` condition and force unwrap `context` here. Wdyt?

WordPress/WordPressTest/ReaderTopicServiceTest.swift
```diff
@@ -138,6 +157,51 @@ final class ReaderTopicSwiftTest: XCTestCase {
 
     // MARK: Tests
 
+    /**
+    Ensure that followed sites a user unfollows from are set to unfollowed in core data when merging
+    results from the REST API.
+    */
+    func testUnfollowedSiteIsUnfollowedDuringSync() {
+        guard let context = context else {
```
leandroalonso, line 165: Wdyt about separating this test in 3 parts using comments? An example: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressTest/PostCoordinatorTests.swift#L369

You can use Arrange/Act/Assert or Given/Then/When, whichever you like the most.

I find them useful in big tests because I can clearly see the 3 different sections. Here, for example, it took me a while to understand that you were first populating the sites and then testing. :)

WordPress/WordPressTest/ReaderTopicServiceTest.swift
momo-ozawa: That makes sense! Makes the test case more concise 👍

WordPress/WordPressTest/ReaderTopicServiceTest.swift
momo-ozawa: Thanks for the suggestion! I totally agree, this is kind of a big test so I'll divide it up into 3 sections 😄

momo-ozawa: Fixed! 6d4135c

WordPress/WordPressTest/ReaderTopicServiceTest.swift
momo-ozawa: Fixed! 6d4135c

Issue #12881, opened 2019-11-06
Menus: the whole section needs improvements to work well with VoiceOver


designsimply:
Steps to reproduce:

1. Go to Settings > General > Accessibility > VoiceOver and toggle the setting on.
1. In the WordPress app, go to Menus.
1. Swipe right to navigate through each item.
1. Try to edit a menu.

**Result:** headings and logically grouped items need to be grouped and several elements are not accessible, for example the down arrows are not accessible at all (cannot be tapped) while using VoiceOver—the whole section needs improvements to work well with VoiceOver. ([8s](https://cloudup.com/cbcHo37Af_a))

<img src="https://user-images.githubusercontent.com/1119271/68268181-69a42d80-0012-11ea-83d5-7303eba5828d.png" width="270" alt="menus">
<sup>Tested with WPiOS 13.5.0.3 (App Store version) on iPhone 6S iOS 13.2.</sup>

<p style="clear:both;"><sub>/hat tip @etoledom for the report, internal reference: p77Llu-bQ6-p2</sub></p>

chipsnyder:
Re-testing this as part of some grounds keeping work. With the changes from @etoledom this has made some really great progress. The one thing I noticed while trying to use it is the selection reverts back a step when moving quickly through the VoiceOver flow. 

1. Go to Settings > General > Accessibility > VoiceOver and toggle the setting on.
1. In the WordPress app, go to Menus.
1. Highlight a menu 
1. Double-tap to expand the areas option
1. While VoiceOver is still speaking the first couple of words ~"Menu area primary" swipe right to navigate to the next item
- **Notice** that the app reverts you back to the expanded menu option forcing you to swipe right again to select the option you were just on


<kbd>![Voice-Over 2020-10-06 11_00_14](https://user-images.githubusercontent.com/3384451/95219578-89644880-07c3-11eb-8475-755b98e998ef.gif)</kbd>

mindgraffiti:
Confirmed fixed: https://d.pr/v/kGqprl. Tested on iPhone 11 Pro, iOS 13.7, WPiOS 16.1 (DEV).

PR #14639, opened 2020-08-13
Improve VoiceOver integration in Menus section - Mark I

etoledom:
Fixes parts of #12881

This PR improves VoiceOver integration of the Header part of the Menus section.

Menus

VoiceOver reads the Areas picker cell like this:
> Menu area: Header. 3 menu areas in this theme. Button. [hint] Expands to select a different menu area.

I'm not sure about the `3 menu areas in this theme` part. It might be useful information but makes the label a bit too long.
What do you think @iamthomasbishop ?

Same with the Menu picker cell. It reads:
> Menu in area Header: Primary. 3 menus available. Button. [hint] Expands to select a different menu to edit.

### To test:

[Using Voice Over](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/docs/using-voiceover.md):
- Go to the Menu section of a site.
- Check that the header section of the controller reacts as expected to VoiceOver.
- Expand the Areas and Menus section.
- Also test switch actions (swipes to the left and right).
- Test landscape where Areas and Menus are side by side.
  - Check that the areas and menus are navigated as a group via switch.
- Check out the [guidelines](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/docs/voiceover-guidelines.md#questions-to-ask).

cc @designsimply 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
+++ b/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
@@ -77,12 +77,14 @@ static CGFloat ViewExpansionAnimationDelay = 0.15;
 {
     MenusSelectionItem *locationItem = [self.locationsView selectionItemForObject:location];
     [self.locationsView setSelectedItem:locationItem];
+    [self prepareForVoiceOver];
 }
 
 - (void)setSelectedMenu:(Menu *)menu
 {
     MenusSelectionItem *menuItem = [self.menusView selectionItemForObject:menu];
     [self.menusView setSelectedItem:menuItem location:self.locationsView.selectedItem];
+    [self prepareForVoiceOver];
 }
 
 - (void)refreshMenuViewsUsingMenu:(Menu *)menu
@@ -138,6 +140,38 @@ static CGFloat ViewExpansionAnimationDelay = 0.15;
 {
     [self.delegate headerViewControllerSelectedForCreatingNewMenu:self];
     [self contractSelectionsIfNeeded];
+}
+
+#pragma mark - accessibility
+
+- (void)prepareForVoiceOver
+{
+    [self configureLocationAccessibility];
+    [self configureMenuAccessibility];
+}
+
+- (void)configureLocationAccessibility
+{
+    // Menu area: Header. 3 menu areas available. Button. [hint] Expands to select a different menu area.
+    NSString *format = NSLocalizedString(@"Menu area: %@, %d menu areas available", @"Screen reader string too choose a menu area to edit. %@ is the name of the menu area (Primary, Footer, etc...). %d is a number indicating the ammount of menu areas available.");
+    MenusSelectionItem *selectedItem = self.locationsView.selectedItem;
+    self.locationsView.accessibilityLabel = [NSString stringWithFormat:format,
+                                                        selectedItem.displayName,
+                                                        self.blog.menuLocations.count];//self.locationsView.subTitleLabel.text];
+    self.locationsView.accessibilityHint = NSLocalizedString(@"Expands to select a different menu area", @"Screen reader hint (non-imperative) about what does the site menu area selector button do.");
+}
+
+- (void)configureMenuAccessibility
+{
+    // Menu in area Header: Primary. 3 menus available. Button. [hint] Expands to select a different menu to edit.
+    NSString *format = NSLocalizedString(@"Menu in area %@: %@, %d menus available", @"Screen reader string too choose a menu to edit. First %@ is the name of the menu area (Primary, Footer, etc...). Second %@ is name of the menu currently selected. %d is a number indicating the ammount of menus available in the selected menu area.");
+    MenusSelectionItem *selectedItem = self.menusView.selectedItem;
+    MenusSelectionItem *selectedLocationItem = self.locationsView.selectedItem;
+    self.menusView.accessibilityLabel = [NSString stringWithFormat:format,
+                                                    selectedLocationItem.displayName,
+                                                    selectedItem.displayName,
+                                                    self.blog.menus.count];//self.detailView.subTitleLabel.text];
+    self.menusView.accessibilityHint = NSLocalizedString(@"Expands to select a different menu", @"Screen reader hint (non-imperative) about what does the site menu selector button do.");
 }
 
 @end
--- a/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
+++ b/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
@@ -82,7 +82,7 @@ static CGFloat ViewExpansionAnimationDelay = 0.15;
 - (void)setSelectedMenu:(Menu *)menu
 {
     MenusSelectionItem *menuItem = [self.menusView selectionItemForObject:menu];
-    [self.menusView setSelectedItem:menuItem];
+    [self.menusView setSelectedItem:menuItem location:self.locationsView.selectedItem];
 }
 
 - (void)refreshMenuViewsUsingMenu:(Menu *)menu
--- a/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
+++ b/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
@@ -83,7 +83,7 @@ static CGFloat ViewExpansionAnimationDelay = 0.15;
 - (void)setSelectedMenu:(Menu *)menu
 {
     MenusSelectionItem *menuItem = [self.menusView selectionItemForObject:menu];
-    [self.menusView setSelectedItem:menuItem location:self.locationsView.selectedItem];
+    [self.menusView setSelectedItem:menuItem];
     [self prepareForVoiceOver];
 }
 
--- a/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
+++ b/WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
@@ -109,6 +109,7 @@ static CGFloat ViewExpansionAnimationDelay = 0.15;
     self.textLabel.backgroundColor = [UIColor clearColor];
     self.textLabel.textColor = [UIColor murielNeutral];
     self.textLabel.text = NSLocalizedString(@"USES", @"Menus label for describing which menu the location uses in the header.");
+    self.textLabel.isAccessibilityElement = NO;
 }
 
 #pragma mark - MenusSelectionViewDelegate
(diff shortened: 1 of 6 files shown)


iamthomasbishop:
> I'm not sure about the 3 menu areas in this theme part. It might be useful information but makes the label a bit too long.
What do you think @iamthomasbishop ?

@etoledom Maybe we could shorten this to "3 menu areas"? 

> Same with the Menu picker cell. It reads:
> Menu in area Header: Primary. 3 menus available. Button. [hint] Expands to select a different menu to edit.

If we want to shorten this, what about something like "Menu in Header area: 3 menus available. Button. [hint] Expands to select a different menu to edit."

WordPress/Classes/ViewRelated/Menus/MenusSelectionItemView.m
```diff
@@ -194,6 +196,22 @@ - (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(UIEvent *)event
...
+{
+    self.isAccessibilityElement = YES;
+    self.accessibilityTraits = UIAccessibilityTraitButton;
+    self.accessibilityLabel = self.item.displayName;
+
+    NSString *selectedLocalizedString = NSLocalizedString(@"Selected", @"Screen reader text to represent the selected state of a button");
+    self.accessibilityValue = self.item.selected ? selectedLocalizedString : nil;
+    if (self.item.selected) {
+        self.accessibilityHint = NSLocalizedString(@"Selects this item", @"Screen reader hint (non-imperative) about what does the menu item selection button do");
+    }
```
guarani: In tests, VoiceOver read the "Selects this item" hint on _both_ selected and non-selected items. I thought this hint should only be added to non-selected items, to describe the result of the element's action. And since already selected items can't be further selected, I didn't expect to hear it there either.

Fwiw, I think this `prepareForVoiceOver` is being called only on item initialization, so `self.item.selected` never changes (it's `true` in all my tests, even when selecting different items).

WordPress/Classes/ViewRelated/Menus/MenusSelectionView.m
```diff
@@ -266,4 +281,46 @@ - (void)selectionItemObjectWasUpdatedNotification:(NSNotification *)notification
...
+                                          self.detailView.subTitleLabel.text];
+    self.detailView.accessibilityHint = NSLocalizedString(@"Expands to select a different menu area", @"Screen reader hint (non-imperative) about what does the site menu area selector button do.");
+}
+
+- (void)configureMenuAccessibility
+{
+    // Menu in area Header: Primary. 3 menus available. Button. [hint] Expands to select a different menu to edit.
+    NSString *format = NSLocalizedString(@"Menu in area %@: %@, %@", @"Screen reader string too choose a menu to edit. First %@ is the name of the menu area (Primary, Footer, etc...). Second %@ is name of the menu currently selected. Third is an already localized string saying the number of menus available in the menu area selected.");
+    self.detailView.accessibilityLabel = [NSString stringWithFormat:format,
+                                          self.selectedItemLocation.displayName,
```
guarani: I find the presence of `selectedItemLocation` in addition to the existing `selectedItem` a bit confusing tbh. I understand that this is used to reference the selected item _from the section above_ (a different `MenusSelectionView` instance, i.e. `MenuHeaderViewController.locationsView`), but it doesn't feel like it belongs here. Especially if it's not actually needed/used in the section above.

I'd suggest using a delegate to ask the `MenuHeaderViewController` which menu location is currently active, and use that — what do you think?

guarani (commented):
I tested this and overall it works great, but I found one potential bug. Here's the steps to reproduce:
1. In the top section (menu location), select a different selected menu location (changed from "Header" to "Social Menu" in my test)
2. With VoiceOver on, tap the header view (i.e. `MenusSelectionDetailView`) once to hear its accessibility message

I heard:

> Menu in area Header: No Menu. 2 menus available. Expanded. Button. [hint] Expands to select a different menu

but I expected:

> Menu in area **Social Menu**: No Menu. 2 menus available. Expanded. Button. [hint] Expands to select a different menu

because "Social Menu", not "Header", is the currently selected menu location.

guarani:
Regarding the accessibility messages in general, e.g.:

> Menu area: Header. 3 menu areas in this theme. Button. [hint] Expands to select a different menu area.

Would it make sense to change this like so:

> Header menu area. 3 menu areas in this theme. Button. [hint] Expands to select a different menu area.

This would feel more natural to me. Keeping in mind that "Header" could be "Social Menu" or any other name, I still think it would work. It's just an idea but thought I'd mention it.

WordPress/Classes/ViewRelated/Menus/MenusSelectionItemView.m
etoledom: Nice catch!

WordPress/Classes/ViewRelated/Menus/MenusSelectionView.m
etoledom: True that two `MenuSelectionitem`s to `MenusSelectionView` might be confusing, and it's used just on the `menus picker` use case.
Delegates are usually used to inform about stuff rather than requesting data, right? It would be better as a Data Source.

Since all the information is in the controller, I'll give it a try to set these labels in the controller itself, rather than passing the data to the views to set their own labels. Wdyt?

guarani: > Delegates are usually used to inform about stuff rather than requesting data, right? It would be better as a Data Source.

I was just referring to the delegate pattern (I didn't make that clear), i.e. the use of a protocol to fetch stuff  — which covers "delegates" or "data sources".

> Since all the information is in the controller, I'll give it a try to set these labels in the controller itself, rather than passing the data to the views to set their own labels. Wdyt?

I think that would also work 👍

etoledom:
@Guarani - Thanks for the review!

> I tested this and overall it works great, but I found one potential bug. Here's the steps to reproduce:
> 
> 1. In the top section (menu location), select a different selected menu location (changed from "Header" to "Social Menu" in my test)
> 2. With VoiceOver on, tap the header view (i.e. `MenusSelectionDetailView`) once to hear its accessibility message
> 
> I heard:
> 
> > Menu in area Header: No Menu. 2 menus available. Expanded. Button. [hint] Expands to select a different menu
> 
> but I expected:
> 
> > Menu in area **Social Menu**: No Menu. 2 menus available. Expanded. Button. [hint] Expands to select a different menu
> 
> because "Social Menu", not "Header", is the currently selected menu location.

Interesting, I couldn't reproduce the problem.
Each time I switch the Menu area, then the Menu picker updates and VoiceOver reads as expected.

This is what I do:
- Select and expand Menu Area picker
- Select a different menu area
- Select the Menu picker
  - Expect that VoiceOver reads the menu area name selected previously.


### About the labels:

#### Menu area picker
> Maybe we could shorten this to "3 menu areas"?

Cool let's do that!

#### Menu picker
> "Menu in Header area: 3 menus available. Button. [hint] Expands to select a different menu to edit."

I think it's important to mention the currently selected menu. I add the name of the area where this menu will be placed so the mental connection with the previous picker can be created. This UI can be a bit complex to understand even for non-voiceover users, so I'm trying to be extra careful with presenting the UI on a way that can be understood.

So I think it is important to mention the currently selected menu and in what menu area it is placed (even though the area is mentioned in the previous picker).

> Header menu area. 3 menu areas in this theme. Button. [hint] Expands to select a different menu area.

That reads better indeed, but I think it's important to give priority to what control it is, and second what information does it have. For example: This is the `Menu area` picker, and the selection is `Primary`.

If we start with the menu selected and then it changes, a VoiceOver user might think it is a different control all together.

etoledom:
@Guarani - Ready for another 👀 🙏

jkmassel:
We're cutting the `15.6` release today, so this PR is being bumped to `15.7`. If it needs to land in the `15.6` release, please merge it to the `release/15.6` branch, and DM me so that I can release a new beta build.

Thanks!

WordPress/Classes/ViewRelated/Menus/MenusSelectionItemView.m
guarani: Working great now, I hear:
- on selected items: "Footer menu. Selected. Button"
- on not selected items: "Primary. Button. Selects this item"
👍

WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
```diff
@@ -139,4 +142,36 @@ - (void)selectionViewSelectedOptionForCreatingNewItem:(MenusSelectionView *)sele
...
+
+- (void)configureLocationAccessibility
+{
+    // Menu area: Header. 3 menu areas available. Button. [hint] Expands to select a different menu area.
+    NSString *format = NSLocalizedString(@"Menu area: %@, %d menu areas available", @"Screen reader string too choose a menu area to edit. %@ is the name of the menu area (Primary, Footer, etc...). %d is a number indicating the ammount of menu areas available.");
+    MenusSelectionItem *selectedItem = self.locationsView.selectedItem;
+    self.locationsView.accessibilityLabel = [NSString stringWithFormat:format,
+                                                        selectedItem.displayName,
+                                                        self.blog.menuLocations.count];//self.locationsView.subTitleLabel.text];
+    self.locationsView.accessibilityHint = NSLocalizedString(@"Expands to select a different menu area", @"Screen reader hint (non-imperative) about what does the site menu area selector button do.");
```
guarani, line 161: From what I see here the accessibility hint is set irrespective of whether the item is selected or not, and I'm seeing some slightly inconsistent behavior. 
Here's the sequence of steps I do (with VoiceOver on):
1. Single tap the menu location button (`MenusSelectionDetailView`) to hear: the accessibility label plus the "Expands to select a different menu area" accessibility hint ✅
2. Double tap the same button to expand the options, now I hear: the accessibility label plus the accessibility value "Expanded" ✅
3. Single tap the same button (just to hear VoiceOver read it again), and I hear a different message: the accessibility label plus the "Expands to select a different menu area" accessibility hint ⚠️

That final message doesn't sound right, it should read the same message as in Step 2. Here's a video that includes sound (VoiceOver audio included): https://cloudup.com/cVFQgi9dJGV (31s)

guarani (approved):
I think this is ready, if you can't reproduce the new issue I mentioned in https://github.com/wordpress-mobile/WordPress-iOS/pull/14639/files#r475925103.

(It might be an iOS 14 beta thing.)

So if you can't reproduce this issue, feel free to merge as I think it's working great!

WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
etoledom: Thank you for noticing this!

For the hint to be the same is fine as long as the status shows that the control is already expanded. In this case, when I test the point 3, VoiceOver does say the value `Expanded` so I'd say it's Ok. I see in the video that this is also the case for you 👍 

After performing the action, is normal for VoiceOver to skip reading the Hint.

etoledom:
Thank you @Guarani and @iamthomasbishop for the reviews! 🙏

WordPress/Classes/ViewRelated/Menus/MenuHeaderViewController.m
guarani: Ok thanks for clarifying 👍

PR #14640, opened 2020-08-13
[TESTING] Rounded image

lukewalczak:
This PR is created only for testing a rounded image via peril build.

🚧 DO NOT MERGE 🚧

To test:

PR submission checklist:


iamthomasbishop:
@lukewalczak Just took the latest build for a spin, and it's feeling really great already! Functionally it works as expected, especially paired with our new-ish image editing capabilities in the editor. Here are my notes, mostly just very minor spacing things:

* I think we need a tiny bit more spacing between the section title ("Image settings") and the style variations cell. I took a screenshot of the current implementation and measured against our Figma component and it looks like we need an additional 7 or 8px between. ([example](https://cloudup.com/cxVsm7xNTHg)). I think we can probably just add 8px top/bottom padding to the scroll container, which I think is what we ended up doing on color picker scroll containers. ([example](https://cloudup.com/cY6kVzy2rBJ))
* There is a ~8-10px diff in the spacing between the bottom of the text label and the first table cell ([example](https://cloudup.com/cQCW1JQDwmK)). Note: If this spacing is set on all sheets, we can probably remove the padding on the bottom of the style variations' scroll container.
* The first item's leading edge is pushed rightward slightly, about 2-3px ([example](https://cloudup.com/cGz3XNulCl6)). Can we either apply the bordering in a way that not only keeps the container the same size, but that the image fills the entire container even when unselected -- meaning the selection border essentially "overlays" the image when it's selected, but the image background itself always stays 100% of the container. Alternatively, we could account for this by applying a negative `margin-left` that == the effective border size (sum of the blue + white borders), but that would mean when the first item is selected, the border would be "hanging" over the left edge, if that makes sense.
* I think the text label styling on production is slightly diff from the Figma component (14px font-size vs 12 in Figma), but 14 is probably better for readability, so I will update the Figma component to match. 👍 

That's all I've got! Really great (and fast!) work here 👏 Can't wait to start using this and expanding style variations to other blocks!

PR #14641, opened 2020-08-13
[Reader P2 follow conversations] add button in comments view

### LINKED ISSUE #14623: Reader P2 follow conversations: add button in comments view
### Parent Issue

### Tasks

- [x] Add a `Follow conversation` button in the comments view. 
- [x] Refine design according to detailed design specs.

momo-ozawa:
Fixes #14623

To test:

### Follow Conversation Button

1. Navigate to the “Reader” tab
2. Tap on the comments button on a post
3. Check that a follow conversation button is displayed ✅ 
8. Check that the follow button matches [design specs](https://wpmobilep2.wordpress.com/2020/08/06/follow-conversations-design/) ✅ 
2. Check that by setting the `isSubscribedToPost` flag in code, the button is updated accordingly ✅ 

Follow conversation | Following conversation
--- | ---
Screen Shot 2020-08-15 at 10 06 53 | Screen Shot 2020-08-15 at 10 00 35

PR submission checklist:

- [ ] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
+++ b/WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
@@ -281,6 +281,46 @@ extension WPStyleGuide {
         button.setTitleColor(disabledColor, for: .disabled)
     }
 
+    @objc public class func applyReaderFollowConversationButtonStyle(_ button: UIButton) {
+        // General
+        button.naturalContentHorizontalAlignment = .leading
+        button.backgroundColor = .clear
+        button.titleLabel?.font = WPStyleGuide.subtitleFont()
+
+        // Color(s)
+        let normalColor = UIColor.neutral(.shade50)
+        let highlightedColor = UIColor.neutral(.shade40)
+        let selectedColor = UIColor.success
+
+        button.setTitleColor(normalColor, for: .normal)
+        button.setTitleColor(selectedColor, for: .selected)
+        button.setTitleColor(highlightedColor, for: .highlighted)
+
+        // Image(s)
+        let side = WPStyleGuide.fontSizeForTextStyle(.headline)
+        let size = CGSize(width: side, height: side)
+        let followIcon = UIImage.gridicon(.readerFollowConversation, size: size)
+        let followingIcon = UIImage.gridicon(.readerFollowingConversation, size: size)
+
+        button.setImage(followIcon.imageWithTintColor(normalColor), for: .normal)
+        button.setImage(followingIcon.imageWithTintColor(selectedColor), for: .selected)
+        button.setImage(followingIcon.imageWithTintColor(highlightedColor), for: .highlighted)
+        button.imageEdgeInsets = FollowConversationButton.Style.imageEdgeInsets
+        button.contentEdgeInsets = FollowConversationButton.Style.contentEdgeInsets
+
+        // Strings
+        let normalText = FollowConversationButton.Text.followConversationStringForDisplay
+        let selectedText = FollowConversationButton.Text.followingConversationStringForDisplay
+
+        button.setTitle(normalText, for: .normal)
+        button.setTitle(selectedText, for: .selected)
+        button.setTitle(selectedText, for: .highlighted)
+
+        // Default accessibility label and hint.
+        button.accessibilityLabel = normalText
+        button.accessibilityHint = NSLocalizedString("Follows the blog.", comment: "VoiceOver accessibility hint, informing the user the button can be used to follow a blog.")
+    }
+
     @objc public class func applyReaderFollowButtonStyle(_ button: UIButton) {
         let side = WPStyleGuide.fontSizeForTextStyle(.headline)
         let size = CGSize(width: side, height: side)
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
```diff
@@ -281,6 +281,46 @@ extension WPStyleGuide {
...
+        let normalText = FollowConversationButton.Text.followConversationStringForDisplay
+        let selectedText = FollowConversationButton.Text.followingConversationStringForDisplay
+
+        button.setTitle(normalText, for: .normal)
+        button.setTitle(selectedText, for: .selected)
+        button.setTitle(selectedText, for: .highlighted)
+
+        // Default accessibility label and hint.
+        button.accessibilityLabel = normalText
+        button.accessibilityHint = NSLocalizedString("Follows the blog.", comment: "VoiceOver accessibility hint, informing the user the button can be used to follow a blog.")
```
leandroalonso, line 321: I guess the hint here is about following comments? :)

leandroalonso:
@osullivanchris I'm a little bit concerned by the height of the button. As for now is `15pt`, and the minimum touch area recommended is `44pt` — I'm pretty sure someone with fat fingers will have a hard time tapping on it, However, we don't have that much space in the header... so I'm not sure what to do. Thoughts?

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
momo-ozawa: Thanks for catching this!

momo-ozawa: Fixed! d75adc0

momo-ozawa: One sec, realized my new string reads "Follows the post", when it should really be "Follows the comments on a post" 🙈

momo-ozawa: Fixed, for real this time 924b4fa

osullivanchris:
> @osullivanchris I'm a little bit concerned by the height of the button. As for now is 15pt, and the minimum touch area recommended is 44pt — I'm pretty sure someone with fat fingers will have a hard time tapping on it, However, we don't have that much space in the header... so I'm not sure what to do. Thoughts?

👋  Hey @leandroalonso that is a fair point. I've had another look. First of all, looking at the builds and the design

- In the design, I increased the space in that row to 72px, which is a little more than I'm seeing in the build - around 56px. Similarly, the font size in the design is 13, 11~ in the build.
- I didn't give specs so no major issue with the build being different. But the design does have _a bit more_ space. 
- I still think even if we match the design exactly, the question is valid. 

follow conversation button options


So I explored a couple of alternatives.
- Alt 1: add a follow button to the top right of the nav bar
- Alt2: add an action to the reply row. This action would go away if the user activates the 'reply' field.
- In both Alt1 and Alt2, there is not enough space to write 'Follow conversation'. And I don't think the 'Follow' action is clear. What are you going to follow...
- So I tried Alt3: adding an ellipsis action to the 'comments' row. Downsides here...its a bit hidden. And it also make the row less obviously tappable 

Not sure if you have any other ideas. I feel like there's flaws in all those alternatives so maybe we should just tweak the build closer to the original design, and get the largest tap target we can manage out of it. 

TBH the follow conversations row is a bit limiting in UI. And the user journey is quite confusing clicking on that row to get from notifications -> comments When you get to the bigger design of comments and post details maybe you can look at redesigning that space and doing something more flexible in the UI. A quick example

header explorations

momo-ozawa:
> In the design, I increased the space in that row to 72px, which is a little more than I'm seeing in the build - around 56px. Similarly, the font size in the design is 13, 11~ in the build.

@osullivanchris @leandroalonso Thanks for the feedback!

- I changed the font size and defined a larger button height for the follow conversation button.
- The screenshot in the middle is what it looks like now. I think it matches the design fairly well. The total height comes out to ~77px, as shown in the screenshot below the table.
- I also tried increasing the button size to 44px, but that looked way too big.

Before | 13pt + 32px button height | 13pt + 44px button height
--- | --- | ---
Screen Shot 2020-08-18 at 20 10 27 | Screen Shot 2020-08-18 at 20 29 37 | Screen Shot 2020-08-18 at 20 26 55

Screen Shot 2020-08-18 at 20 29 01

Let me know what you think 👍

osullivanchris:
@momo-ozawa thanks for the tweaks! I think the middle screenshot looks a lot better. I agree, adding more space on the right is too far visually.

I think we can probably live with this right now as the layout looks right, even if the touch target is a little small...and consider a bigger change to this header (something like the examples I did) with the reader details work that's going on. WDYT @leandroalonso ?

cc @mbshakti

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
```diff
@@ -281,6 +281,46 @@ extension WPStyleGuide {
...
     }
 
+    @objc public class func applyReaderFollowConversationButtonStyle(_ button: UIButton) {
+        // General
+        button.naturalContentHorizontalAlignment = .leading
+        button.backgroundColor = .clear
+        button.titleLabel?.font = fontForTextStyle(.footnote)
+
+        // Color(s)
+        let normalColor = UIColor.primary(.shade40)
```
leandroalonso, line 291: Based on Chris' designs I think the right color here is just `.primary`. In light mode, the contrast is a little bit low when using `shade40`.

leandroalonso:
@osullivanchris sounds like a plan to me.

@momo-ozawa I just left a comment about the blue color. Just a minor change (sorry I haven't noticed that before).

WordPress/Classes/ViewRelated/Reader/WPStyleGuide+Reader.swift
momo-ozawa: Fixed! 7a7d9ce

mbshakti:
Sounds good to me too. Some thoughts for the future: I know we currently don't have any predefined button styles in the app for primary, secondary, tertiary buttons etc, maybe in the future when we define different button styles, we can use visually treat this like a secondary button as well (similar to your exploration).

Issue #14570, opened 2020-08-04
Update Reader Card to display an expandable list of tags for the posts on the Discover feed


emilylaguna:
Parent Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14395

<img width="436" alt="Screen Shot 2020-08-04 at 10 33 14 AM" src="https://user-images.githubusercontent.com/793774/89306484-cd509b00-d624-11ea-91e3-84ea88cd3ee8.png">

### Goal
- Display a single row of chips for the tags for the post
- If the number of tags will overflow out of the cell then display an overflow chip `X+`
- When tapped the chips will expand to multiple rows to display the remaining tags
- Track `reader_chips_more_toggled when the user taps the more/hide button
### Designs
- [Figma](https://www.figma.com/file/BL53gDs6URD0yRUmGRCbhr/%F0%9F%93%90-Discover-Postcard-UI-(Specs)?node-id=492%3A805)

leandroalonso:
@emilylaguna Could you please track `reader_chips_more_tapped` when the user taps "X+" to expand the topics list?

emilylaguna:
Is there an event to track if the user collapses?

leandroalonso:
Perhaps we can rename the event to `reader_chips_more_toggled` and add a property `action: expand|collapse`? cc @yaelirub @mbshakti

yaelirub:
I think that we can do with just `reader_chips_more_toggled` (no need for expand, collapse) . That's at least what I'd be interested in looking at.

mbshakti:
+1 to @yaelirub's comment.

Would we also be able to look at if the user tapped on any of these topics (to go to the topic page) or are we already tracking that?

Issue #14348, opened 2020-06-18
Reader iPad Card UI: Abnormal spacing between metadata and post thumbnail


mbshakti:
### Expected behavior
Normal spacing between metadata and post thumbnail

### Actual behavior
Large spacing between metadata and post thumbnail

### Steps to reproduce the behavior
Open app -> Go to reader
Here's where it gets tricky. I don't always see this? It seems up sometimes, and sometimes seems to be normal.

### Tested on [device], iOS [version], WPiOS [version]
iPad 11, iOS version 13.4.1, WP version 15.0

### Screenshot
Take a look at the second story here (gilly macmillan)
![IMG_0035](https://user-images.githubusercontent.com/6350997/85029873-c5a76680-b14a-11ea-8dfc-bf5291fe13b9.PNG)

Both the first and second story here seem to have this issue
![IMG_0036 2](https://user-images.githubusercontent.com/6350997/85030073-03a48a80-b14b-11ea-8c58-dba17dc351d6.PNG)

kean:
The design has changes multiple over the previous iterations. It no longer reproduces.

dangermattic:
Thanks for reporting! 👍

PR #14645, opened 2020-08-13
Reader Improvements Phase 2: Adds a tags chips group to the top of the reader card cell

emilylaguna:
Main Issue #14570

#### Portrait
| Collapsed | Expanded |
| :---: | :---: |
||  |

#### Landscape
| Collapsed | Expanded |
| :---: | :---: |
||  |

#### 01 - Tag chips appear on Discover tab
1. Tap on the Reader tab
2. Tap on the Discover tab

- [x] Expectation: Posts that have tags should now display the chips at the top of the cell

#### 02 - Expand Works
1. Follow the steps from 01
2. Locate a post that has an expansion cell indicated by the 'X+' where X is the number of chips left to be displayed
3. Tap on the expansion cell

- [x] Expectation: The rest of the chips should be displayed
- [x] Expectation 2: The number displayed in the expansion cell is the number of chips that appeared, example if the number displays 2+, then 2 more chips should be displayed
- [x] Expectation 3: A cell with the localized title "Hide" should be displayed as the last chip

#### 03 - Hide Button
1. Follow the steps in 02
2. Tap on the Hide chip

- [x] Expectation: The chips collapse, and display the expansion tooltip

#### 04 - Landscape
1. Following the steps from 03
2. Rotate the device so it appears in landscape mode

- [x] Expectation: If the device is wide enough to display all the chips, no expansion cell is displayed
- [x] Expectation 2: If there are still some chips that are hidden the expansion cell's number should be updated to reflect the number of remaining cells
- [x] Expectation 3: Rotating the device back to portrait updates the count as well

#### 05 - Chips only appear on Discover
1. With the Reader tab active
2. Tap on any other tab besides Discover

- [x] Expectation: The chips should not appear

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
@@ -29,7 +29,13 @@ final class ReaderCellConfiguration {
         cell.setSiteName(post.blogName)
     }
 
-    func configurePostCardCell(_ cell: UITableViewCell, withPost post: ReaderPost, topic: ReaderAbstractTopic? = nil, delegate: ReaderPostCellDelegate?, loggedInActionVisibility: ReaderActionsVisibility) {
+    func configurePostCardCell(_ cell: UITableViewCell,
+                               withPost post: ReaderPost,
+                               topic: ReaderAbstractTopic? = nil,
+                               delegate: ReaderPostCellDelegate?,
+                               loggedInActionVisibility: ReaderActionsVisibility,
+                               cardDelegate: ReaderPostCardCellDelegate? = nil,
+                               displayTopics: Bool = true) {
         // To help avoid potential crash: https://github.com/wordpress-mobile/WordPress-iOS/issues/6757
         guard !post.isDeleted else {
             return
@@ -38,8 +44,10 @@ final class ReaderCellConfiguration {
         let postCell = cell as! ReaderPostCardCell
 
         postCell.delegate = delegate
+        postCell.cardDelegate = cardDelegate
 
         postCell.loggedInActionVisibility = loggedInActionVisibility
+        postCell.displayTopics = displayTopics
         postCell.configureCell(post)
         postCell.layoutIfNeeded()
     }
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
@@ -41,6 +41,7 @@ final class ReaderCellConfiguration {
 
         postCell.loggedInActionVisibility = loggedInActionVisibility
         postCell.configureCell(post)
+        postCell.layoutIfNeeded()
     }
 
     func configureGapMarker(_ cell: ReaderGapMarkerCell, filling: Bool) {
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
@@ -34,7 +34,7 @@ final class ReaderCellConfiguration {
                                topic: ReaderAbstractTopic? = nil,
                                delegate: ReaderPostCellDelegate?,
                                loggedInActionVisibility: ReaderActionsVisibility,
-                               cardDelegate: ReaderPostCardCellDelegate? = nil,
+                               topicChipsDelegate: ReaderTopicsChipsDelegate? = nil,
                                displayTopics: Bool = true) {
         // To help avoid potential crash: https://github.com/wordpress-mobile/WordPress-iOS/issues/6757
         guard !post.isDeleted else {
@@ -44,7 +44,7 @@ final class ReaderCellConfiguration {
         let postCell = cell as! ReaderPostCardCell
 
         postCell.delegate = delegate
-        postCell.cardDelegate = cardDelegate
+        postCell.topicChipsDelegate = topicChipsDelegate
 
         postCell.loggedInActionVisibility = loggedInActionVisibility
         postCell.displayTopics = displayTopics
--- a/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
@@ -11,7 +11,7 @@ private struct Constants {
     static let summaryMaxNumberOfLines: NSInteger = 2
 }
 
-protocol ReaderPostCardCellDelegate: class {
+protocol ReaderTopicsChipsDelegate: class {
     func didSelect(topic: String)
     func heightDidChange()
 }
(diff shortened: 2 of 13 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -1926,3 +1943,17 @@ private extension ReaderStreamViewController {
         static let contentErrorImage = "cloud"
     }
 }
+
+extension ReaderStreamViewController: ReaderPostCardCellDelegate {
+    func heightDidChange() {
+        WPAnalytics.track(.readerChipsMoreToggled)
```
leandroalonso, line 1949: I think this needs to be in the place where the tap actually happens.

Otherwise, if any other place of the code starts using the chips view, it will have to implement this event.

Also, is quite strange to see this event in height change callback.

WordPress/Classes/ViewRelated/Reader/Tags View/ReaderTopicCollectionViewCoordinator.swift
```diff
@@ -0,0 +1,196 @@
...
+    case collapsed
+    case expanded
+}
+
+protocol ReaderTopicCollectionViewCoordinatorDelegate: AnyObject {
+    func coordinator(_ coordinator: ReaderTopicCollectionViewCoordinator, didSelectTopic topic: String)
+    func coordinator(_ coordinator: ReaderTopicCollectionViewCoordinator, didChangeState: ReaderTopicCollectionViewState)
+}
+
+class ReaderTopicCollectionViewCoordinator: NSObject {
```
leandroalonso, line 13: I guess this coordinator is making the solution to break the single-responsibility principle.

To display chips any entity needs to:

* Have a `UICollectionView`
* Have this coordinator

Maybe I might be being too simplistic here and there's underlying complexities I'm not aware of, but why to not make something like `TagsCollectionView` which inherits from `UICollectionView`?

That way all a view/VC/class needs to do is add the `TagsCollectionView`, feed it with the tags and voilá!

This would make things very separated and all the other entity needs to know is about the public API of `TagsCollectionView`.

emilylaguna:
> * In iPad, when going from portrait to landscape, there's additional space being shown before/after the blog header. I'm not sure if this is related to your changes though (I have the feeling I've seen that before):

This is a documented issue here: https://github.com/wordpress-mobile/WordPress-iOS/issues/14348

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: Done in: 5c0b1a884edba451763d1302719df988678ba4b7

WordPress/Classes/ViewRelated/Reader/Tags View/ReaderTopicCollectionViewCoordinator.swift
emilylaguna: Done in: ba76f8197c7c7d0df8a7f1035d127777337d1277

WordPress/Classes/ViewRelated/Reader/ReaderCellConfiguration.swift
```diff
@@ -29,7 +29,13 @@ final class ReaderCellConfiguration {
         cell.setSiteName(post.blogName)
     }
 
-    func configurePostCardCell(_ cell: UITableViewCell, withPost post: ReaderPost, topic: ReaderAbstractTopic? = nil, delegate: ReaderPostCellDelegate?, loggedInActionVisibility: ReaderActionsVisibility) {
+    func configurePostCardCell(_ cell: UITableViewCell,
+                               withPost post: ReaderPost,
+                               topic: ReaderAbstractTopic? = nil,
+                               delegate: ReaderPostCellDelegate?,
+                               loggedInActionVisibility: ReaderActionsVisibility,
+                               cardDelegate: ReaderPostCardCellDelegate? = nil,
```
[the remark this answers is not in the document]
emilylaguna: > > * In iPad, when going from portrait to landscape, there's additional space being shown before/after the blog header. I'm not sure if this is related to your changes though (I have the feeling I've seen that before):
> 
> This is a documented issue here: #14348

This PR may also fix that issue as well.

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -77,6 +84,10 @@ private struct Constants {
         return  width <= 320
     }
 
+    var topicsCoordinator: ReaderTopicCollectionViewCoordinator?
+    weak var cardDelegate: ReaderPostCardCellDelegate?
```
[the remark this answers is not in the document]
emilylaguna: Done in: 8b0c98c3f4489d557690838d138e660f92e6117e

emilylaguna:
> * When rotating from landscape to portrait with expanded tags, there's an additional space being shown. You can see here: https://d.pr/v/sMeLkc

Done in: 8504f973299d5a3f986c1d188eae4df426f67ded

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -139,6 +148,8 @@ private struct Constants {
         setupSummaryLabel()
         setupAttributionView()
         adjustInsetsForTextDirection()
+
+        topicsCollectionView.topicDelegate = self
```
leandroalonso, line 152: Maybe move that to `configureTopicsCollectionView`?

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -232,11 +245,13 @@ private struct Constants {
         summaryLabel.backgroundColor = .listForeground
         commentActionButton.titleLabel?.backgroundColor = .listForeground
         likeActionButton.titleLabel?.backgroundColor = .listForeground
+        topicsCollectionView.backgroundColor = .listForeground
```
leandroalonso, line 248: Maybe move that to `TopicsCollectionView`?

WordPress/Classes/ViewRelated/Reader/Tags View/TopicsCollectionView.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+class TopicsCollectionView: DynamicHeightCollectionView {
```
leandroalonso, line 3: Documentation? :)

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
emilylaguna: 👍

WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
emilylaguna: > Maybe move that to `TopicsCollectionView`?

I think that the this should be configured by the implementing view since it could change depending on where else it's being used. Thoughts?

leandroalonso: Makes sense! Ok!

WordPress/Classes/ViewRelated/Reader/Tags View/TopicsCollectionView.swift
emilylaguna: done: 793c851c99e7bf849562feff50b6ac388b416a54

Issue #14220, opened 2020-05-30
Editor: cannot preview drafts on a WordPress.com site with a custom domain


designsimply:
We received the following report:

> I am currently unable to use the View and Preview functions on my Wordpress App on iPad. I receive a 404 page not found error each time. The issue arises when I attempt to preview or view my drafts. Any of the drafts on my site will show the effect.

* Type of site: WordPress.com Premium with a custom domain
* WPiOS 14.8.2
* iPad Pro 9.7-inch (WiFi/Cellular) (iPad6,4)
* iOS 13.4.1

Note: the app logs provided by the user showing an editor session with a preview doesn't seem to indicate a problem (is it possible we aren't logging failed previews in this flow and so app logs won't be helpful in this type of case?):

```
2020-05-25 09:58:46:344 🔵 Tracked: editor_session_start <blog_type: wpcom, content_type: gutenberg, editor: gutenberg, has_unsupported_blocks: 0, post_type: post, session_id: 218FF820-7CD8-4A57-AB4E-1D4949111147, startup_time_ms: 1723, unsupported_blocks: (
)>
2020-05-25 09:58:58:388 🔵 Tracked: web_preview_opened <blog_type: wpcom, post_type: post>
2020-05-25 09:59:00:339 TracksService sendQueuedEvents completed. Sent 2 events.
2020-05-25 09:59:03:718 🔵 Tracked: editor_session_end <blog_type: wpcom, content_type: gutenberg, editor: gutenberg, has_unsupported_blocks: 0, outcome: cancel, post_type: post, session_id: 218FF820-7CD8-4A57-AB4E-1D4949111147>
2020-05-25 09:59:03:728 🔵 Tracked: editor_discarded_changes <blog_id: 151743515, editor_source: gutenberg, has_gutenberg_blocks: 1, post_id: 647>
2020-05-25 09:59:03:780 🔵 Tracked: editor_closed <blog_id: 151743515, editor_source: gutenberg, has_gutenberg_blocks: 1, post_id: 647>
```

<p style="clear:both;"><sup>(internal reference: p4a5px-2zK-p2, 2987505-zen)</sup></p>

designsimply:
I tested this today on three different types of sites (WordPress.com, Jetpack, Self-hosted) and found that previews did show a brief 404 error when first loading on a WordPress.com site but then eventually loaded. Based on the user report, it seems some sites aren't able to eventually resolve the preview link but I am not sure why.

[...]

PR #14650, opened 2020-08-14
Log error if WP.com preview auth cookie is not found

guarani:
Addresses #14220

When the user previews a post or page, the app fetches an authentication cookie from the server. There are user reports (see linked issue above) where the preview screen for WP.com sites mapped to custom domains show a "Page not found" error. While I could only ever reproduce this once, in that instance I saw that the `wordpress_logged_in` cookie had an empty string as its value. 

### The potential fix

This PR aims to provide logs that will allow us to better assess whether a missing cookie is indeed the problem on WP.com Simple sites. If the cookie request returns error will appear in Sentry as:

Response to request for auth cookie for WP.com site failed to return cookie.

### What's not included

- **Self-hosted sites**: although self-hosted sites may benefit from similar error logging, but in my brief tests, I found the requests were different and so were the cookies. Since there's no reports of this error on self-hosted sites, I think we can omit them.
- **Atomic sites**: atomic sites use an auth cookie named `wordpress_logged_in_` (note the trailing underscore) and their previews are handled different (looks like there's a distinction between private and non-private Atomic), so I've excluded them from the scope of this PR.

### How to test

Testing requires recreating the scenario where the `wordpress_logged_in` cookie is present in the server response, but has an empty value — and then checking Sentry to ensure the error is being logged correctly.

I used Charles proxy to remove the cookie from the server response. Setting up Charles proxy required some patience to get working correctly, but it did work in the end. I used the iOS simulator instead of a physical device because it was easier to setup the proxy there, but either should work. Here's the steps to test:

1. Setup Charles proxy to intercept traffic from an iOS simulator
2. Build and run this branch on the device (do a clean install to ensure no cookies are left over from a previous session)
3. Log in to a WP.com site that uses a custom domain
4. If running this from Xcode, you need to turn on error logging in the app (to allow errors to be sent to Sentry): Select a site > Tap profile photo > App Settings > Debug > Always Send Crash Logs
5. Create a post in the block editor, and add content
6. Import this [breakpoint](https://github.com/wordpress-mobile/WordPress-iOS/files/5077322/Breakpoints.xml.zip) into Charles (or manually add a breakpoint to Charles for the **response** of the `https://wordpress.com:443/wp-login.php` URL)
7. Ensure breakpoints are enabled in Charles (Proxy > Breakpoint Settings)
8. Tap the ellipsis menu (three dots) and select Preview
9. When Charles opens the Breakpoints window *, click the Edit Response button, and find both instances of the `wordpress_logged_in` cookie (technically, you only need to remove the one that's marked "secure" or for HTTPS) and remove their values
10. Click Execute in the Breakpoints window
11. Expect the app to show a "Page not found" error
11. Verify the error can be found in Sentry (search for the error message "Response to request for auth cookie for WP.com site failed to return cookie.")
12. Finally, remove the breakpoint (or quit Charles), open the preview screen in the app again and ensure it loads the post or page correctly

_* If the Breakpoints window doesn't open, quitting the simulator, restarting Charles, and running the simulator again fixed this for me_

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/AuthenticationService.swift
+++ b/WordPress/Classes/Services/AuthenticationService.swift
@@ -5,6 +5,17 @@ class AuthenticationService {
 
     static let wpComLoginEndpoint = "https://wordpress.com/wp-login.php"
 
+    enum RequestAuthCookieError: Error, LocalizedError {
+        case wpcomCookieNotReturned
+
+        public var errorDescription: String? {
+            switch self {
+            case .wpcomCookieNotReturned:
+                return "Response to request for auth cookie for WP.com site failed to return cookie."
+            }
+        }
+    }
+
     // MARK: - Self Hosted
 
     func loadAuthCookiesForSelfHosted(
@@ -84,7 +95,16 @@ class AuthenticationService {
 
                 self.getAuthCookiesForWPCom(username: username, authToken: authToken, success: { cookies in
                     cookieJar.setCookies(cookies) {
-                        success()
+
+                        // Verify that the WP.com auth cookie was indeed returned by the server.
+                        cookieJar.hasWordPressComAuthCookie(username: username, atomicSite: false) { hasCookie in
+                            guard hasCookie else {
+                                failure(RequestAuthCookieError.wpcomCookieNotReturned)
+                                return
+                            }
+                            success()
+                        }
+
                     }
                 }) { error in
                     // Make sure this error scenario isn't silently ignored.
--- a/WordPress/Classes/Services/AuthenticationService.swift
+++ b/WordPress/Classes/Services/AuthenticationService.swift
@@ -96,7 +96,6 @@ class AuthenticationService {
                 self.getAuthCookiesForWPCom(username: username, authToken: authToken, success: { cookies in
                     cookieJar.setCookies(cookies) {
 
-                        // Verify that the WP.com auth cookie was indeed returned by the server.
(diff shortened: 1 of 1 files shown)


guarani:
@diegoreymendez hey Diego, sorry to bug you — I noticed you have modified this area of the code base and would appreciate your review here 🙏

WordPress/Classes/Services/AuthenticationService.swift
```diff
@@ -84,7 +95,16 @@ class AuthenticationService {
 
                 self.getAuthCookiesForWPCom(username: username, authToken: authToken, success: { cookies in
                     cookieJar.setCookies(cookies) {
-                        success()
+
+                        // Verify that the WP.com auth cookie was indeed returned by the server.
```
diegoreymendez, line 99: I'd consider removing this comment as it's not adding any information to the code that's right below it.

It's not a big deal, so feel free to ignore it.  But I always try to avoid comments that explain what the code does since the value they provide is low (since any developer should be able to tell what the code does), vs the added effort of maintaining it over time as the code changes.  My 2 cents!

diegoreymendez (approved):
This looks like a great addition to the service.  It's always right to treat outside data as unsafe to begin with.

I just added one comment for your consideration.

WordPress/Classes/Services/AuthenticationService.swift
guarani: That's understandable, I'll go ahead and remove it.

guarani:
Thanks for the review, @diegoreymendez!

Issue #14609, opened 2020-08-10
[User Profiles] Show basic profile when tapping avatars


frosty:
We would like to add the ability to view user profiles within the app. Stage one will be to show profile information when tapping avatars within the app – particularly in Notifications and Reader comments.

Here's how things should look:

<img width="662" alt="image" src="https://user-images.githubusercontent.com/4780/89774934-b2879600-dafe-11ea-8d0b-10089aeb78c4.png">

A few notes:

- The designs above show a 'recent posts' section, but this can be omitted for v1.
- This should be able to use our `BottomSheetViewController` component to display the sheet. See `presentFilter` in `ReaderTabViewModel` for an existing usage example. As we'll be leaving out the recent posts area initially, I think the sheet can just be presented at a single size, and doesn't need to be expandable by the user.
- We'll have some of the information to populate the profile already, such as the avatar image itself and the display name. The rest can be fetched from [Gravatar](https://en.gravatar.com/). In particular, `https://gravatar.com/hash_of_email_address.json` will fetch a user's profile info.
- While we're waiting for the other information to load, we should show the fields with a 'ghost' animation like we use elsewhere in the app.

Task list:

- [x] Add feature flag.
- [x] Add the 'user profile' sheet
   - [ ] Retrieve appropriate profile information from the backend as needed.
   - [x] Add user profile to Notifications.
   - [x] Add user profile to Reader comments.
   - [x] Ensure that a 'ghost' animation is displayed when/if necessary.
- [ ] Ensure views are correct for both light and dark appearance.
- [ ] Add accessibility support.
- [ ] Add tracks.
- [ ] Enable feature flag for internal.
- [ ] Enable feature flag for all.
- [ ] Remove feature flag.

designsimply:
This is so great! I've always wanted this. 😍 

Would it be possible to add a gear icon or edit link to the profile sheet that appears for the logged-in user when viewing their own account? Would it be possible to add the ability to edit which sites are shown in the profile from the app directly before this is launched so users can change it if they find an unexpected list of sites they forgot about and/or would like to change or if they would like to add sites? It's not apparent from within the app how or where or if you can do that.

Here is what the My Profile editing screen currently looks like in the app:

<img src="https://user-images.githubusercontent.com/1119271/89960346-062bd980-dbfc-11ea-953b-84d4e3b3afbf.PNG" width="270" alt="IMG_4483">

Calypso has a "Profile links" section which the app does not have:

[...]

PR #14654, opened 2020-08-15
Display basic user profile on avatar tap

ghost:
Ref: [14609](https://github.com/wordpress-mobile/WordPress-iOS/issues/14609)
[WordPressKit PR 268](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/268)

Initially from Reader Comments and Notification Details pages, when a user taps on an avatar image we want to present a bottom sheet with a basic user profile view.

Design:
https://user-images.githubusercontent.com/4780/89774934-b2879600-dafe-11ea-8d0b-10089aeb78c4.png

This commit also depends on a `WordPressKit-iOS` pod update to parse some extra properties for gravatar profile.

### To test:
1- Proceed to the Reader tab, and select a comment of an item. Then tap on the gravatar of comment's owner.
2- Proceed to the Notifications tab, then select Comments in header. Go to the details of the comment, then tap on the gravatar of comment's owner.

Missing items:
* Sites are not displayed yet. Will submit a PR later.
* Size of the page will be updated according to visible items.
* Tracking not added. 

### PR submission checklist:
* [x]  I have considered adding unit tests where possible.
* [x]  I have considered adding accessibility improvements for my changes.
* [x]  I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.


Podfile
```diff
@@ -37,9 +37,9 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.14-beta'
+    # pod 'WordPressKit', '~> 4.14-beta'
     #pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :tag => ''
-    #pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => ''
+    pod 'WordPressKit', :git => 'https://github.com/ughuro/WordPressKit-iOS.git', :branch => 'develop'
```
Gio2018: Thanks for adding this! It will need to be changed to the actual release before merging

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -108,6 +111,8 @@ extension FeatureFlag: OverrideableFlag {
             return "Gutenberg Modal Layout Picker"
         case .whatIsNew:
             return "What's New / Feature Announcement"
+        case .avatarsBasicProfile:
+            return "Show Basic Profile from Avatars"
```
Gio2018: It would be better to add a bit more description to specify which avatars will actually show the profile

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        // TODO: Ghost animation
+        // TODO: Accessibility
+        // TODO: Dark mode
+        // TODO: Tracking
+        // TODO: Styling
```
Gio2018: unless these TODOs represent missing code in this specific location, I would not leave them here. Also, to easily identify them and avoid forgetting them in the code, I usually label them more specifically, so you can search them easily (e.g. TODO - BASICPROFILES: - ...)

Gio2018 (changes requested):
Hey @ughuro ! This looks like a good start!

I left some initial comments in the code, but more importantly I noticed some issues when testing your branch:
1. If I go to Reader/Discover then find a post with comments, tap on comments, tap on any avatar, then no user info is displayed at all (see screenshot):

<p align=center>

</p>

2. I am not sure if this will be fixed with the page size in future PRs, but the UITableView below looks off center and oddly position vertically (see the same screenshot above)

3. On iPad regular width, the sheet seems oddly anchored (and same problems with the UITableView); in compact view, it covers the whole view. See screenshots below

iPad regular width | iPad compact width
--- | ---
 |

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
ghost: Done, updated the description.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: Done.

ughuro:
I've fixed the anchor position on iPad, but will update the content height after implementing the sites table view.
Some users don't have any email data, I will use the hash in gravatar image URLs for them.

WordPress/Classes/Services/GravatarService.swift
```diff
@@ -72,4 +77,20 @@ open class GravatarService {
     func gravatarServiceRemote() -> GravatarServiceRemote {
         return GravatarServiceRemote()
     }
+
+    private static func parse(remoteProfile: RemoteGravatarProfile, completion: ((_ profile: GravatarProfile?) -> Void)) {
```
Gio2018, line 81: thanks for extrapolating this method!

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -108,6 +111,8 @@ extension FeatureFlag: OverrideableFlag {
             return "Gutenberg Modal Layout Picker"
         case .whatIsNew:
             return "What's New / Feature Announcement"
+        case .avatarsBasicProfile:
+            return "Show Basic Profile from Avatars on Reader Comments and Notification Details"
```
Gio2018, line 120: 👍

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    @IBOutlet private weak var gravatarImageView: CircularImageView!
+    @IBOutlet private weak var nameLabel: UILabel!
+    @IBOutlet private weak var usernameLabel: UILabel!
+    @IBOutlet private weak var locationLabel: UILabel!
+    @IBOutlet private weak var aboutMeLabel: UILabel!
+    @IBOutlet private weak var contentView: UIView!
+    @IBOutlet private weak var sitesTableView: UITableView!
+
+    var viewModel: BasicUserProfileViewModel!
```
Gio2018: I know that implicit unwrapping can be convenient, but I would prefer not to use it, especially in a situation where the property is not injected at initialization time (so it's not really guaranteed that it's always not nil). What do you think?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    let allowsUserTransition = false
+
+    @IBOutlet private weak var gravatarImageView: CircularImageView!
+    @IBOutlet private weak var nameLabel: UILabel!
+    @IBOutlet private weak var usernameLabel: UILabel!
+    @IBOutlet private weak var locationLabel: UILabel!
+    @IBOutlet private weak var aboutMeLabel: UILabel!
+    @IBOutlet private weak var contentView: UIView!
+    @IBOutlet private weak var sitesTableView: UITableView!
```
Gio2018: I believe this will contain sites in the future, just a warning that at the moment it's not connected to anything in the storyboard.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        let controller = BasicUserProfileViewController.loadFromStoryboard()
+        controller.viewModel = viewModel
+
+        let bottomSheet = BottomSheetViewController(childViewController: controller)
+        bottomSheet.show(from: context, sourceView: view, arrowDirections: .up)
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
```
Gio2018: nit: can we put the view life cycle methods in the same order they are called?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+        super.viewWillAppear(animated)
+        contentView.startGhostAnimation()
```
Gio2018: The description says that the ghost animation is not part of this pull request. Is that a leftover?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        nameLabel.font = WPStyleGuide.fontForTextStyle(.title3, fontWeight: .semibold)
+        usernameLabel.font = WPStyleGuide.fontForTextStyle(.callout)
+        locationLabel.font = WPStyleGuide.fontForTextStyle(.caption1)
+        aboutMeLabel.font = WPStyleGuide.fontForTextStyle(.caption1)
+
+        nameLabel.textColor = .text
+        usernameLabel.textColor = .textSubtle
+        locationLabel.textColor = .textSubtle
+        aboutMeLabel.textColor = .text
```
Gio2018: would it be possible to group all these style-related instructions in a helper method?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        downloadGravatarImageView()
+
+        viewModel.fetchUserDetails { [weak self] profile in
+            guard let strongSelf = self else {
```
Gio2018: since Swift 4.2, we can just do `guard let self = self...`, it might be more readable, what do you think?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+                strongSelf.aboutMeLabel.isHidden = true
+            }
+
+            strongSelf.contentView.stopGhostAnimation()
+
+            // TODO - BASICPROFILES: Update sites
+        }
+    }
+
+    private func downloadGravatarImageView() {
```
Gio2018, line 73: I'm wondering if there's a way to make the `UIViewController` agnostic of underlying properties (email, URL etc) and just let the `ViewModel` provide it with the right image? Thoughts?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+    @IBOutlet private weak var nameLabel: UILabel!
+    @IBOutlet private weak var usernameLabel: UILabel!
+    @IBOutlet private weak var locationLabel: UILabel!
+    @IBOutlet private weak var aboutMeLabel: UILabel!
+    @IBOutlet private weak var contentView: UIView!
+    @IBOutlet private weak var sitesTableView: UITableView!
+
+    var viewModel: BasicUserProfileViewModel!
+
+    @objc class func present(context: UIViewController, view: UIView, email: String?, avatarURL: URL?) {
```
Gio2018: would it be possible to move this in an extension? This way, if for any reason we need to move away the presentation from the class itself (e.g. in a coordinator) it would be easier.

Gio2018 (changes requested):
Hey @ughuro !! This is looking much better! Thank you.

I left some comments in the code, and noticed a couple of things:

1. On iPad, looks like the profile name appears truncated (see screenshot). Should we give at least two lines to it? @osullivanchris what are your thoughts?
<p align=center>
Screen Shot 2020-08-20 at 5 31 51 PM
</p>

2. Still on iPad compact, if I pan down the sheet and then pull it back up, a ghost appears and never goes away (see animation below)

<p align=center>

</p>

The view on iPad compact still covers the whole thing, but I guess this will be part of the page setup PR.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: For most of the time I strongly agree with you, but with this case the controller is meaningless without a view model and I wanted to emphasize that. Also we can clear lots of unwrapping from the code. But still if this is against the convention, I can update?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: Yes that was a dummy reference, will remove it :(

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: Done

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: I found a way to display them. So updating the description now.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: Done

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: I'm used to explicitly stating it as strong but that's fine for me too. 👍

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: My first intention was to take advantage of the existing UIImageView extension methods. Additionally I wanted to keep view model away from UIKit stuff.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: Sure

ghost:
Hi @Gio2018 ,
1- I will look for a way to update the content size on iPad if that's an only iPad issue. @osullivanchris We have the number-of-lines problem for about me text in the bottom too. To which line number do you prefer to limit that information?
2- Will fix that. I didn't realize the viewWillAppear function will be called multiple times. :/

Gio2018:
> Hi @Gio2018 ,
> 1- I will look for a way to update the content size on iPad if that's an only iPad issue. @osullivanchris We have the number-of-lines problem for about me text in the bottom too. To which line number do you prefer to limit that information?
> 2- Will fix that. I didn't realize the viewWillAppear function will be called multiple times. :/

Sounds good @ughuro and thank you for the updates! I'll give it another pass once the changes are pushed.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
Gio2018: I think the most important thing is to avoid crashes, and I totally agree that the `viewModel` is necessary. The ideal scenario would be to have it injected at initialization time, but I realize that this might require a bit of changes because we are using storyboards, and I don't want to over complicate your trial project. If you'd agree, we could proceed as follows:

- For now, let's just do the explicit unwrapping instead of the implicit: thinking about the impacts on the app, is certainly better if a profile page is shown blank or with missing parts than the app crashing after trying to view a profile

- In the future, we could refactor this `viewController` in order to have the `viewModel` injected at initialization time

What do you think?

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
Gio2018: It's really up to you, I just pointed it out because they introduced this in Swift 4.2 for this exact purpose.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
Gio2018: That's a very good point and I agree that the `viewModel` should not be concerned of `UIKit` stuff. Let's leave as it is for now, and in the future we can reason about the methods that download the images.

WordPress/Classes/ViewRelated/Basic Profile/BasicUserProfileViewController.swift
ghost: Sure, sounds great. Will update that with next PR. 👍

Gio2018:
Hey @ughuro ! Once the conflict on the WPKit PR is addressed and we can use the version in the Podfile, we should be able to merge this one!!

Gio2018 (approved):
:shipit:!

PR #14655, opened 2020-08-15
[Reader P2 follow conversations] hookup subscription endpoints

### LINKED ISSUE #14624: Reader P2 follow conversations: hook up subscription endpoints
### Parent Issue

### Tasks
- [x] Call the [get the current subscription state](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post/subscribers/mine/) endpoint and update the follow/following conversation button to reflect the current state.
- [x] Tapping on the button toggles between `Follow conversation` and `Following conversation`.
- [x] When following a conversation, call the [subscribe to a post comments stream](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/posts/%24post/subscribers/new/) endpoint. 
- [x] Similarly, when unfollowing a conversation, call the [unsubscribe from comments stream](https://developer.wordpress.com/docs/api/1.1/post/sites/%24site/posts/%24post/subscribers/mine/delete/) endpoint.

momo-ozawa:
Fixes #14624
Note: Please merge after https://github.com/wordpress-mobile/WordPress-iOS/pull/14641

To test:

### Get subscription status
1. In a browser, open a P2 post. Note the subscription status of the post.
2. In the app, navigate to the corresponding P2 post's comments view.
3. Check that the subscription status in the app matches that of the browser.

### Subscribe to post
0. In the app, navigate to a P2 post you are not subscribed to yet.
1. Toggle the subscription status for a P2 post.
2. Check that the follow button is toggled, and shows `Following conversation` ✅ 
3. Check that a feedback notification is displayed for a few seconds at the bottom of the screen: `Successfully subscribed to post` ✅ 
2. In the browser, navigate to the corresponding P2 post.
3. Check that the subscription status shows `Following conversation`. ✅ 

### Unsubscribe from post
0. In the app, navigate to a P2 post you are already subscribed to.
1. Toggle the subscription status for a P2 post.
2. Check that the follow button is toggled, and shows `Follow conversation` ✅ 
3. Check that a feedback notification is displayed for a few seconds at the bottom of the screen:  `Successfully unsubscribed from post` ✅ 
2. In the browser, navigate to the corresponding P2 post.
3. Check that the subscription status shows `Follow conversation`. ✅ 

PR submission checklist:


momo-ozawa:
@leandroalonso
Should I add tracking methods for the subscribe/unsubscribe actions?

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -1171,6 +1189,57 @@ - (void)presentWebViewControllerWithURL:(NSURL *)URL
...
+
+    // Optimistically toggle subscription status
+    self.postHeaderView.isSubscribedToPost = newIsSubscribed;
+
+    // Define success block
+    void (^successBlock)(void) = ^void() {
+        NSString *subscriptionString = newIsSubscribed
+            ? NSLocalizedString(@"subscribed to", "Past tense verb + preposition. A subscription action.")
+            : NSLocalizedString(@"unsubscribed from", "Past tense verb + preposition. A subscription action.");
+        NSString *successTitle = [NSString stringWithFormat: NSLocalizedString(@"Successfully %@ the post", @"The app successfully updated the subscription status for the post"), subscriptionString];
```
leandroalonso, line 1211: What do you think about creating the whole string at once?

I fear that this might be a little bit confusing to the translators and that we might have issues with how this sentence might be written in other languages.

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -1171,6 +1189,57 @@ - (void)presentWebViewControllerWithURL:(NSURL *)URL
 
 #pragma mark - PostHeaderView helpers
 
+- (void)handleFollowingConversationButtonTapped
+{
+    __typeof(self) __weak weakSelf = self;
```
leandroalonso, line 1194: I guess this is another piece of logic that we might want to extract to another layer?

WordPress/Classes/Services/ReaderPostService.h
```diff
@@ -95,6 +95,30 @@ extern NSString * const ReaderPostServiceToggleSiteFollowingState;
  */
 - (void)refreshPostsForFollowedTopic;
 
+/**
+ Get the subscription status of the specified post.
+
+ @param post The reader post to fetch the subscription status.
+ @param success block called on a successful fetch.
+ @param failure block called if there is any error. `error` can be any underlying network error.
```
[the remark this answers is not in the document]
momo-ozawa: Fixed! 27c55fe

(I'm assuming you were pointing out how the comment specifies "network errors" 🙏 )

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -192,6 +192,17 @@ - (void)fetchPostAtURL:(NSURL *)postURL
     }];
 }
 
+-(void)fetchSubscriptionStatusForPost:(ReaderPost *)post
+                              success:(void (^)(BOOL isSubscribed))success
+                              failure:(void (^)(NSError *))failure
+{
+    ReaderPostServiceRemote *remoteService = [[ReaderPostServiceRemote alloc] initWithWordPressComRestApi:[self apiForRequest]];
```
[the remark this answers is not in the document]
momo-ozawa: I ended up extracting the subscription related methods from `ReaderPostService `and defining them in a new `FollowCommentService.swift` 👍 

As for injecting the remote service -- I was looking at other Service classes to use as an example, and it seemed like RemoteServices were being instantiated inside the init method, instead of being injected. What do you think?

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -576,6 +581,19 @@ - (BOOL)shouldDisplaySuggestionsTableView
 
 #pragma mark - View Refresh Helpers
 
+- (void)updateSubscriptionStatus
+{
+    NSManagedObjectContext *context = [[ContextManager sharedInstance] mainContext];
```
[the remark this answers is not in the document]
momo-ozawa: Your comment made me realize I'm actually not using Core Data for the new subscription use cases. 

I did, however, swap to using the [new FollowCommentsService ](https://github.com/wordpress-mobile/WordPress-iOS/pull/14655#discussion_r472122291)when fetching the subscription status.

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
momo-ozawa: That sounds like a much simpler idea, thanks!

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
momo-ozawa: By extracting to another layer, do you mean just the part that deals with the service? If so, I've swapped over to using the new FollowCommentsService 👍 

If you meant the whole method, I was thinking I _could_  move it out to the ReaderPostHeaderView... what do you think?

WordPress/Classes/Services/ReaderPostService.m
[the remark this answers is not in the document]
leandroalonso: > As for injecting the remote service -- I was looking at other Service classes to use as an example, and it seemed like RemoteServices were being instantiated inside the init method, instead of being injected. What do you think?

This is tricky indeed. I had the same problem recently, I solved this way: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Services/ReaderCardService.swift#L20

It's still not good but at least I'm able to inject it through the constructor and mock it if necessary in unit tests. Anyway, the way you did is ok too. :)

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
leandroalonso: Now that I'm reading again I can see that most of this logic is related to the presentation.

My main idea was to extract any logic to another layer to avoid the ViewController from calling database/network directly. I think we can keep that way. :)

WordPress/Classes/Services/FollowCommentsService.swift
```diff
@@ -0,0 +1,72 @@
...
+    /// Toggles the subscription status of the specified post.
+    ///
+    /// - Parameters:
+    ///   - isSubscribed: The current subscription status for the reader post.
+    ///   - success: Success block called on a successful fetch.
+    ///   - failure: Failure block called if there is any error.
+    @objc func toggleSubscribed(_ isSubscribed: Bool,
+                                success: @escaping () -> Void,
+                                failure: @escaping (Error?) -> Void) {
+        if isSubscribed {
```
leandroalonso, line 47: Wdyt about adding a simple unit test for this method?

Just to make sure the correct remote methods are called.

momo-ozawa: Sounds good! Will do! 👍

momo-ozawa: Added tests! 14328a5

leandroalonso:
@momo-ozawa 

>  Should I add tracking methods for the subscribe/unsubscribe actions?

Yes, please! We can do that in a separate task, wdyt?

(sorry, I totally missed this comment!)

WordPress/WordPressTest/FollowCommentsServiceTests.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    override func tearDown() {
+        ContextManager.overrideSharedInstance(nil)
+        super.tearDown()
+    }
+
+    // MARK: - Mocks / Tester
+
+    class ReaderPostServiceRemoteMock: ReaderPostServiceRemote {
```
leandroalonso, line 24: I often prefer to add the mocks at the end of the file. Wdyt?

Mainly because when I open a test I'm interested in the tested behaviors, here I have to scroll down and find the tests.

WordPress/WordPressTest/FollowCommentsServiceTests.swift
```diff
@@ -0,0 +1,97 @@
...
+
+        override func unsubscribeFromPost(with postID: Int,
+                                          for siteID: Int,
+                                          success: @escaping () -> Void,
+                                          failure: @escaping (Error) -> Void) {
+            unsubscribeFromPostCalled = true
+        }
+    }
+
+    class FollowCommentsServiceTester: FollowCommentsService {
```
leandroalonso, line 44: Since `FollowCommentsService` is the object which we're testing here I'd try to avoid inherit from it.

What you can do instead is to inject the service into the class. There are two distinct, yet similar, ways you can accomplish that:

1. https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Services/PostCoordinator.swift#L46
2. https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Services/ReaderCardService.swift#L20

Both ways will allow you to inject your `ReaderPostServiceRemoteMock` without the need of creating a "Tester". :)

leandroalonso:
@osullivanchris regarding tracking the follow/unfollow action I thought about a single `reader_toggle_follow_conversation` with a property `action: following|unfollowing`. Wdyt?

We should apply the same event in Android too. cc @develric

WordPress/WordPressTest/FollowCommentsServiceTests.swift
momo-ozawa: @leandroalonso 
Thanks for the suggestions! I'll give this a shot 👍

momo-ozawa: It seems like Objective-C doesn't support Swift default parameters. So no matter what, it seems like you'll have to pass an argument for the service parameter... which is unfortunate, because that means we won't be able to leverage the convenience of default paramters.

While the default parameter approach works well for the tests, I think it looks a little unwieldy when called from Objective-C. What do you think?

### Approach 1
```
FollowCommentsService *service = [[FollowCommentsService alloc] initWithPost:(ReaderPost * _Nonnull) service:(ReaderPostServiceRemote * _Nullable)];

```

Screen Shot 2020-08-20 at 21 28 35


### Approach 2
```
FollowCommentsService *service = [[FollowCommentsService alloc] initWithPost:(ReaderPost * _Nonnull) service:(ReaderPostServiceRemote * _Nonnull)];
```

Screen Shot 2020-08-20 at 21 42 30

leandroalonso: Hmmm, I see... 🤔

What if we create a `class func` that acts as the constructor and use that in the Obj-C side? That way we can still inject it in the tests and use the "custom constructor" in Obj-C. Thoughts?

momo-ozawa: That's a great idea! 👍

momo-ozawa: Fixed! 5b90251
The dependency injection approach is much cleaner 😄

osullivanchris:
> @osullivanchris regarding tracking the follow/unfollow action I thought about a single reader_toggle_follow_conversation with a property action: following|unfollowing. Wdyt?

No strong opinions here. Of course its good if we can track something.

This suggestion would tell us how often people are following/unfollowing conversations, is that right?

The only other thing I wonder is whether we would want to get breakdowns. E.g
- do some users follow/unfollow conversations more than others
- do some sites or posts get a lot more follows/unfollows than others
maybe that is covered by putting in this one event

thanks for the ping!

leandroalonso:
> do some users follow/unfollow conversations more than others

We can know that!

> do some sites or posts get a lot more follows/unfollows than others
maybe that is covered by putting in this one event

I guess we might able to track that by adding `siteID` as a property. But I wonder if that has privacy issues?

PR #14656, opened 2020-08-16
Feature/14625 show or hide follow conversation button

### LINKED ISSUE #14625: Reader P2 follow conversations: show/hide button logic
### Parent Issue

### Tasks
- [x] Show the `Follow conversation` / `Following conversation` button only if it’s a P2 post. New P2 product blogs are marked as `"is_wpforteams_site": true` in the site options response.

momo-ozawa:
Fixes #14625
Note: Please merge after https://github.com/wordpress-mobile/WordPress-iOS/pull/14655

To test:

### Show Follow Conversation Button
1. Navigate to the comments view of a P2 blog post marked as `"is_wpforteams_site": true`
2. Check that the follow conversation button is visible ✅ 

### Hide Follow Conversation Button
1. Navigate to the comments view of a P2 blog post marked as `"is_wpforteams_site": false`
2. Check that the follow conversation button is hidden ✅ 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
@@ -1231,7 +1231,7 @@ static NSString *RestorablePostObjectIDURLKey = @"RestorablePostObjectIDURLKey";
     // Call the remote service to toggle the subscription status
     NSManagedObjectContext *context = [[ContextManager sharedInstance] mainContext];
     ReaderPostService *service = [[ReaderPostService alloc] initWithManagedObjectContext:context];
-    [service toggleSubscribed:newIsSubscribed
+    [service toggleSubscribed:oldIsSubscribed
                       forPost:self.post
                       success:successBlock
                       failure:failureBlock];
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
@@ -295,7 +295,6 @@ static NSString *RestorablePostObjectIDURLKey = @"RestorablePostObjectIDURLKey";
     headerView.translatesAutoresizingMaskIntoConstraints = NO;
     headerView.showsDisclosureIndicator = self.allowsPushingPostDetails;
     [headerView setSubtitle:NSLocalizedString(@"Comments on", @"Sentence fragment. The full phrase is 'Comments on' followed by the title of a post on a separate line.")];
-    // headerView.isSubscribedToPost = NO;  // FIXME: Get subscription state
     [headerWrapper addSubview:headerView];
 
     // Border
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
@@ -149,8 +149,6 @@ static NSString *RestorablePostObjectIDURLKey = @"RestorablePostObjectIDURLKey";
 {
     [super viewWillAppear:animated];
 
-    [self updateSubscriptionStatus];
-
     [self.keyboardManager startListeningToKeyboardNotifications];
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(handleApplicationDidBecomeActive:)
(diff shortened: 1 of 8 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -563,6 +566,18 @@ - (BOOL)canComment
     return self.post.commentsOpen && self.isLoggedIn;
 }
 
+- (BOOL)canFollowPost
+{
+    // FIXME: Older a8c internal P2s do not contain the isWPForTeams flag.
+    // In case we can't find a flag that marks an old P2 site as being a P2,
+    // we can assume that blogs in the Reader's Automattic tab are P2s.
+    // Note that blogs in the Reader's Automattic tab are only available to Automatticians.
+    NSManagedObjectContext *context = [[ContextManager sharedInstance] mainContext];
```
leandroalonso, line 575: Another logic that we can extract from the VC. :)

momo-ozawa: Thinking of extracting to the [FollowCommentsService](https://github.com/wordpress-mobile/WordPress-iOS/pull/14655#discussion_r472122291)!

momo-ozawa: Fixed! 65e25c0

WordPress/Classes/Services/FollowCommentsService.swift
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    /// Determines whether or not the comments on the post can be followed.
+    @objc func canFollowConversation() -> Bool {
+        // FIXME: Older a8c internal P2s do not contain the isWPForTeams flag.
+        // In case we can't find a flag that marks an old P2 site as being a P2,
+        // we can assume that blogs in the Reader's Automattic tab are P2s.
+        // Note that blogs in the Reader's Automattic tab are only available to Automatticians.
+        guard let blogID = self.post.siteID else { return false }
```
leandroalonso, line 66: Overall we always use some break lines when using `guard`:

```suggestion
        guard let blogID = self.post.siteID else {
            return false
        }

```

Just to Keep consistency. :)

WordPress/Classes/Services/FollowCommentsService.swift
```diff
@@ -0,0 +1,84 @@
...
+        let service = BlogService(managedObjectContext: ContextManager.shared.mainContext)
+        let blog = service.blog(byBlogId: blogID)
```
leandroalonso, line 68: I wonder if this logic of finding a `Blog` can be moved to `ReaderPost` (an extension, maybe?).

I remember to see that in a few places, so it would be def good to have it there. Also, a `ReaderPost` should know how to return it's own blog. Wdyt?

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -610,6 +613,20 @@ - (void)refreshPostHeaderView
...
     }
+    
+    FollowCommentsService *service = [[FollowCommentsService alloc] initWithPost:self.post];
+    self.postHeaderView.showsFollowConversationButton = [service canFollowConversation];
+}
+
+- (void)refreshSubscriptionStatus
+{
+    __weak __typeof(self) weakSelf = self;
+    FollowCommentsService *service = [[FollowCommentsService alloc] initWithPost:self.post];
```
leandroalonso, line 624: Two things about this:

* We're instantiating `FollowCommentsService` inline in two different places. Wouldn't be better to make it a class property?
* Correct me if I'm wrong, but we will always call the service even though the button is unavailable (hidden), right? Shall we prevent that?

WordPress/Classes/Services/FollowCommentsService.swift
momo-ozawa: Thanks! Totally missed [this](https://github.com/wordpress-mobile/swift-style-guide#braces) -- I'll keep this in mind!

WordPress/Classes/Services/FollowCommentsService.swift
momo-ozawa: > Also, a ReaderPost should know how to return it's own blog.

Oh! That's a really good idea... thanks!

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
momo-ozawa: > We're instantiating FollowCommentsService inline in two different places. Wouldn't be better to make it a class property?

Good point. Thanks for the suggestion 👍 

> we will always call the service even though the button is unavailable (hidden), right? Shall we prevent that?

Very, very good point. I totally missed this. Thank you for catching this 🙏

WordPress/Classes/Services/FollowCommentsService.swift
momo-ozawa: Fixed! 01db296

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
momo-ozawa: Fixed! 
045a214
9a0d9f7

PR #14661, opened 2020-08-17
Reader Improvements: Fix Discover indefinite ghost mode

### LINKED ISSUE #14660: Reader Cards: Discover gets stuck in ghost mode indefinitely
1. Load the new discover view
2. Scroll down to the next page
3. Tap on a topic
4. Tap back

### Expected behavior
The view should show the posts that were previously visible

### Actual behavior
The view gets stuck in the ghost mode until switching tabs

iPhone Simulator 13.3

emilylaguna:
Fixes #14660

### To test:
1. Run the app with the Reader Improvements Phase 2 flag enabled
2. Tap on the Reader tab
3. Tap on the Discover tab
3. Scroll down a few posts
4. Tap on a topic at the top of a post
5. Wait for it to load
6. Tap back

**Expectation:** The app displays the posts that were there before

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -103,6 +103,10 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         })
     }
 
+    override var topicPostsCount: Int {
+        return cards?.count ?? 0
+    }
+
     override func syncIfAppropriate() {
         // Only sync if the tableview is at the top, otherwise this will change tableview's offset
         if isTableViewAtTheTop() {
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -72,8 +72,11 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
             return
         }
 
-        readerTabBarViewController.displaySelectInterestsIfNeeded()
-        showGhost()
+        readerTabBarViewController.displaySelectInterestsIfNeeded { [unowned self] (isDisplaying) in
+            if isDisplaying {
+                self.showGhost()
+            }
+        }
     }
 
     // MARK: - Sync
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -100,6 +103,10 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         })
     }
 
+    override var topicPostsCount: Int {
```
yaelirub, line 106: After messing around with the ReaderStream and headers, tags, site topics, tag topics I realize even more the importance of simple naming so could this just be numberOfCards or cardsCount?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -915,13 +915,23 @@ import WordPressFlux
...
+        if canSync() && (interval >= refreshInterval || topicPostsCount == 0) {
             syncHelper?.syncContentWithUserInteraction(false)
         } else {
             handleConnectionError()
         }
     }
 
+    /// Returns the number of posts for the current topic
+    /// This allows the count to be overriden by subclasses
+    var topicPostsCount: Int {
```
yaelirub, line 928: No need to change necessarily but could have also been a one liner :
`return readerTopic != nil ? readerTopic!.posts.count : 0`

yaelirub:
@emilylaguna , good work here. 
Unrelated to this PR but needs to be discussed: is this really how the discover page going to look like before you scroll or is this just testing data:
Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 18 38 05

We should limit the number of interests suggestions to a fixed number of rows so that the user will see the next card on the screen before scrolling.

emilylaguna:
> Unrelated to this PR but needs to be discussed: is this really how the discover page going to look like before you scroll or is this just testing data:

This is coming directly from the server itself. The count changes sometimes, I think they're still working on the related interests code.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
emilylaguna: Makes sense, although I'm not a fan of the force unwrapping. I would change it to just:

`return readerTopic?.posts.count ?? 0`

yaelirub: lol great!

PR #14668, opened 2020-08-18
Use `main` branch for mobile secrets

jkmassel:
To test:

- In the project's root directory, run `rm -rf configure_files && git reset` 
- In the project's `Scripts` directory, run `bundle exec fastlane run configure_apply`, then build the app in Xcode.
- If it succeeds, this PR is good to go!


mokagio (changes requested):
Following up on @oguzkocer's comment, I think the command to test this is 

```
bundle exec fastlane run configure_apply
```

If I do that, I get this output, which is what I'd expect when updating the local repo according to what's in the `.configure` file and `.mobile-secrets`:

```
[12:12:31]: -----------------------------
[12:12:31]: --- Step: configure_apply ---
[12:12:31]: -----------------------------
[12:12:31]: .configure-files/wpcom_app_credentials has changes that need to be merged. Would you like to see a diff?
(y/n)
n
[12:12:36]: Would you like to make a backup of .configure-files/wpcom_app_credentials? (y/n)
n
[12:12:40]: Would you like to overwrite .configure-files/wpcom_app_credentials? (y/n)
y
[12:12:44]: .configure-files/wpcom_alpha_app_credentials has changes that need to be merged. Would you like to see a
diff? (y/n)
n
[12:12:46]: Would you like to make a backup of .configure-files/wpcom_alpha_app_credentials? (y/n)
n
[12:12:48]: Would you like to overwrite .configure-files/wpcom_alpha_app_credentials? (y/n)
y
[12:12:49]: .configure-files/wpcom_internal_app_credentials has changes that need to be merged. Would you like to see a
diff? (y/n)
n
[12:12:49]: Would you like to make a backup of .configure-files/wpcom_internal_app_credentials? (y/n)
n
[12:12:50]: Would you like to overwrite .configure-files/wpcom_internal_app_credentials? (y/n)
y
[12:12:51]: .configure-files/project.env has changes that need to be merged. Would you like to see a diff? (y/n)
n
[12:12:54]: Would you like to make a backup of .configure-files/project.env? (y/n)
n
[12:12:56]: Would you like to overwrite .configure-files/project.env? (y/n)
y
[12:12:56]: Applied configuration
[12:12:56]: Result: true
```

When I tried to run the unit tests locally, the build failed during the "Generate API credentials" build phase:

```
Checking for WordPress.com Oauth App Secret in /Users/gio/Developer/a8c/wpios/WordPress/../.configure-files/wpcom_app_credentials
Config found
warning: Missing WPCOM_APP_ID
Command PhaseScriptExecution failed with a nonzero exit code
``

That's the same failure we're seeing [in CI](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-iOS/13694/workflows/35684d31-bcf4-4ecb-88d6-c8096619cbe3/jobs/33165).

I think something is going wrong when decrypting the files, if I `cat .configure-files/wpcom_app_credentials` on `develop`, I get a text file which includes a line defining the `WPCOM_APP_ID`. But on this branch, after applying the new configurations, `cat` doesn't show anything because the file is a binary.

loremattei:
So, there's something weird here. I think the tool works as expected with the new branch, but something was wrong when the new encoded file have been created. 
Any chances that you didn't decrypt your mobile secrets repository before updating the files @jkmassel? It looks like the decrypted configuration files contain the git-crypted version of themselves 🤔. The only reason for this I can think of is that you run `configure_update` before unlocking the mobile secrets repository.

So, when running `configure_apply`, I get the same result as @mokagio. I believe the tool is doing its work well, as it's able to pick the unlock key from mobile-secrets and to decrypt the files. 

When running `configure_update`, things look good to me. The tool asks to switch branch and it creates the new encrypted files from the `HEAD` commit on `main`, which is what it's supposed to do. Running `configure_apply` after this brings back a good configuration.  `.configure` doesn't change after the update, so it looks good in this PR. 

> "The current branch is master. Would you like to switch branches?" which I don't think should say master but I am not sure

@oguzkocer, @jkmassel is more familiar than me with the tool so he can confirm this, but I think this is the expected behaviour, as long as you have the `master` branch checked out in your `mobile-secrets` repository. `configure_update` is meant to update the project configuration to use what you currently have in your `mobile-secrets` repository, so if you are on `master`, it tells you that and asks you if you want to pick the new configuration from that branch or you want to switch to another one.

jkmassel:
> Any chances that you didn't decrypt your mobile secrets repository before updating the files @jkmassel? It looks like the decrypted configuration files contain the git-crypted version of themselves 🤔. The only reason for this I can think of is that you run configure_update before unlocking the mobile secrets repository.

Oof...that's correct 😓. Hazards of using a new computer, I totally forgot to install `git-crypt` after I copied everything over.

Thanks for finding that @loremattei and for the feedback everyone 😄

jkmassel:
I've updated the testing steps to better represent the actual change going on here – sorry about the confusion!

mokagio (approved):
Works now on my end. Thanks @jkmassel.

---

Note. To test this, I only used the terminal. My pods sandbox wasn't in sync, and `pod install --repo-update` failed on both `main` and here due to some failed dependency resolution.

Once I updated the pods, everything worked well.

---

Other interesting note. You don't need to `cd` into `Scripts/` to run `configure_apply`, it works from the root, too.

I believe the reason is that, since we run it through Bundler and all the logic lives in the Fastlane plugin, the `Fastfile` is never read, everything runs through the Fastlane gem itself.

loremattei:
> Other interesting note: You don't need to cd into Scripts/ to run configure_apply, it works from the root, too.

Right! The key is the `run` keyword which tells `fastlane` to look for an `action` and not a `user lane`.

loremattei (approved):

PR #14669, opened 2020-08-18
Reader Improvements Phase 2: Merge reader cards branch into develop

emilylaguna:
Merges the changes from the new reader cards branch into develop. 

#### 00 - Feature Flag Enabled
1. Enable the Reader Improvements Phase 2 feature flag
2. Tap on the Reader tab
3. Tap on Discover 
4. You should now see the new cards layout
5. Test the normal flow of the reader such as paging, tapping on a post, tapping on a topic, etc

#### 01 - Feature Flag Disabled
1. Disable the Reader Improvements Phase 2 feature flag
2. Tap on the Reader tab
3. Tap on Discover 
4. You should now see the original Discover view with the new header, and new tag chips

Note: If you had the flag enabled before you will need to tap on the Following tab and back to the Discover to reset it.

5. Test the normal flows to make sure everything works as normal

### PR submission checklist:


guarani:
@emilylaguna I noticed the new Core Data model version v98 doesn't have an entry in [MIGRATIONS.md](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/MIGRATIONS.md) and was wondering if that's intentional or not.
I happened to come across this today since I'm adding another version (which will be v99).

leandroalonso:
@guarani My mistake! On it :)

Issue #14581, opened 2020-08-05
What's New - Announcement Modal UI


Gio2018:
Implement the UI of the "What's New" modal, according to the specs. Below are two screenshots extracted from the spec:

iPhone | iPad
--- | ---
<img height="400" alt="whatsnewiphone" src="https://user-images.githubusercontent.com/34376330/89450831-d32ba680-d720-11ea-82aa-edcf0422258f.png"> | <img height="400" alt="whatsnewipad" src="https://user-images.githubusercontent.com/34376330/89450879-e8a0d080-d720-11ea-868c-f153c7946759.png">

PR #14672, opened 2020-08-18
Add view elements to WhatIsNewView

Gio2018:
Ref #14581

Adds base view elements to the "What's New / Feature Announcements" page.

**Notes:**
**1. the table view that will contain the actual announcements, as well as the "Find out more" button is not populated yet. At the moment, you will just see a grey rectangle in place of it; the grey background will be removed in future PRs.
2. The modal size on iPad has not been customized at the moment and it's using the default (which on iOS 12 would be full screen).**

To test:
1. Go to Me -> App Settings -> What's new in WordPress
2. Verify that the title "What's New in WordPress" appears and the version number is consistent to the actual version. For reference, see design in #14581 and screenshots below. 
3. The Continue button should dismiss the modal.

iPhone | iPad
--- | ---
whatsnewiphone | whatsnewipad

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ Not yet added.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ Not yet added.


osullivanchris (approved):
Looks great so far!

WordPress/Classes/ViewRelated/What's New/WPTabBarController+WhatIsNew.swift
```diff
@@ -0,0 +1,25 @@
...
+
+    func makeWhatIsNewViewController() -> WhatIsNewViewController {
+        return WhatIsNewViewController(whatIsNewViewFactory: makeWhatIsNewView)
+    }
+
+    private func makeWhatIsNewView() -> WhatIsNewView {
+
+        var version = ""
+        if let versionNumber = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
+            version = NSLocalizedString("Version ", comment: "Description for the version label in the What's new page.") + versionNumber
```
ScoutHarris, line 12: What do you think about making a `private enum` for all the `NSLocalizedString`s in this method? It'd make it more readable. For example, I almost missed the tacked on `+ versionNumber` here as it was kind of obfuscated by the string definition.

WordPress/Classes/ViewRelated/What's New/WPTabBarController+WhatIsNew.swift
```diff
@@ -0,0 +1,25 @@
...
+extension WPTabBarController {
+
+    func makeWhatIsNewViewController() -> WhatIsNewViewController {
+        return WhatIsNewViewController(whatIsNewViewFactory: makeWhatIsNewView)
+    }
+
+    private func makeWhatIsNewView() -> WhatIsNewView {
+
+        var version = ""
+        if let versionNumber = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String {
```
ScoutHarris, line 11: We have a `NSBundle+VersionNumberHelper` that does this. Maybe you can use `Bundle.main.shortVersionString()`?

WordPress/Classes/ViewRelated/What's New/WPTabBarController+WhatIsNew.swift
Gio2018: Totally more readable! Thanks for pointing this out!

WordPress/Classes/ViewRelated/What's New/WPTabBarController+WhatIsNew.swift
Gio2018: yes absolutely, I had missed this helper, thank you!

PR #14677, opened 2020-08-18
Update Core Data model minimum tools version

guarani:
This PR updates the minimum "Tools Version" used by our current Core Data model **WordPress 97.xcdatamodel** from Xcode 7.3 to 9.0.

### Rationale

This was the minimum change necessary to enable the (newish) Core Data attribute type, URI, which allows URLs to be modeled directly as `URL` instead of as `String`. Without this change, using URI produced an error:

Error caused by old model version

### Changes

Updating the version changes the underlying XML data model automatically as seen in the diff. These are the changes I see:
- `minimumToolsVersion="Xcode 7.3"` becomes `minimumToolsVersion="Xcode 9.0"`
- Changes related to indexing:
  - Removal of `indexed="YES"` from model attributes
  - addition of `<fetchIndex>`, [which is related to a new indexing API](https://stackoverflow.com/a/47852563/1305067)

### Testing

This change doesn't involve a new Core Data model version, but I think it's important to test a simple app update scenario. I tested using the following steps:
1. Install the app on a device using the current `develop` branch
2. Log in to a site and perform some common actions such as updating a post
3. Switch over to this branch and re-run the app on the device
4. Open the app, open the post and ensure there's no crash

PR submission checklist:


guarani:
@aerych I wonder if you could review this? It would be much appreciated!

aerych:
Heya @guarani 👋. It looks like https://github.com/wordpress-mobile/WordPress-iOS/pull/14669 was merged a few hours ago and it bumped the model version to 98.  Since these changes target 97 I'm thinking that (out of an abundance of caution) it might be best to redo the patch on top of model version 98 to avoid any tricky merges.  Thoughts? 

I've locally made this change and tested as described and everything looks solid to me.  No issues or warnings from core data just by updating the tools version on the existing model as far as I can see. We should be good to go after addressing the new model version.

guarani:
@aerych thanks for letting me know! I've applied the change to model version 98 and reverted the changes to the previous version. 
Ready again for review 👍

guarani:
Thank you @aerych!

PR #14682, opened 2020-08-19
Reset the isExpanded flag on reuse

### LINKED ISSUE #14681: Reader Improvements: Expandable tags view stays expanded on refresh
If you tap the + button on a cell, then reload the table view, the cell stays expanded even though the tags have changed

1. Enable the Reader Improvements Phase 2 feature flag
2. Tap on the Reader tab
3. Tap on Discover
4. Reload until you see a post that has a lot of tags
5. Expand them
6. Reload the view
7. The cell will still show the "hide" button

emilylaguna:
Fixes #14681

To test:

1. Enable the Reader Improvements Phase 2 feature flag
2. Tap on the Reader tab
3. Tap on Discover
4. Reload until you see a post that has a lot of tags
5. Expand them
6. Pull top refresh the view
7. The cell should be collapsed and no longer show the 'hide' view

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
@@ -168,13 +168,8 @@ protocol ReaderTopicsChipsDelegate: class {
         imageLoader.prepareForReuse()
         displayTopics = false
 
-        guard let layout = topicsCollectionView.collectionViewLayout as? ReaderInterestsCollectionViewFlowLayout else {
-            return
-        }
-
-        layout.isExpanded = false
+        topicsCollectionView.collapse()
     }
-
 
     // MARK: - Configuration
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift
```diff
@@ -167,6 +167,12 @@ protocol ReaderTopicsChipsDelegate: class {
 
         imageLoader.prepareForReuse()
         displayTopics = false
+
+        guard let layout = topicsCollectionView.collectionViewLayout as? ReaderInterestsCollectionViewFlowLayout else {
```
leandroalonso, line 171: What about moving this logic to the `TopicsCollectionView`?

Then we can just call `topicsCollectionView.collapse()` or something like that — and avoid accessing the layout directly.

emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14682/commits/f1dd20f2968ceeeb8bebf2de092eec1807919982

PR #14686, opened 2020-08-19
Fix upload for HEIC files (live photos)

### LINKED ISSUE #14680: Unknown File Type Error for HEIC images on WordPress.com Business plan sites
### Expected behavior
I would expect that I could upload images (live or still) from my iPhone to my media library without needing to convert them. 

### Actual behavior
On a Business plan sites on WordPress.com (atomic), live photos from the iPhone media library get an error 'unknown file type' and cannot be uploaded or the posts published. 

1. Go to My Sites and select a WordPress.com business plan site with plugins installed to edit in the app.
2. Go to Blog Posts to create a new post or edit an existing one and use the image block or gallery block.
3. Select Choose from my device. 
4. Choose a 'live photo' (HEIC) from your albums and click 'Add New'.
5. The image will try to upload until you get an error 'failed to insert media. Please tap for options.
6. Options come up with Media Options, File type uknown' and retry or dismiss. 

The post can't be published 
24fa887f-4286-434c-bbb0-283653806734

I tried using Safari browser on the same device, same site and using the same image and was able to do so from the browser. 

##### Tested on iPhone X, iOS 13.6, WPiOS 15.3

Internal link: p1HpG7-9XE-p2

leandroalonso:
Fixes #14680

(I'll cherry-pick the fix to `release/15.5` after this is merged)

### To test

**Make sure to test using a real device**

1. Go to My Sites and select an Atomic site
2. Go to Blog Posts to create a new post or edit an existing one and use the image block or gallery block.
3. Select Choose from my device. 
4. Choose a 'live photo' (HEIC) from your albums and add it
5. The image should upload

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/MediaService.m
+++ b/WordPress/Classes/Services/MediaService.m
@@ -57,7 +57,10 @@ NSErrorDomain const MediaServiceErrorDomain = @"MediaServiceErrorDomain";
         }
         
         if (blog.allowedFileTypes != nil) {
-            allowedFileTypes = blog.allowedFileTypes;
+            // HEIC isn't supported when uploading an image, so we filter it out (http://git.io/JJAae)
+            NSMutableSet *mutableAllowedFileTypes = [blog.allowedFileTypes mutableCopy];
+            [mutableAllowedFileTypes removeObject:@"heic"];
+            allowedFileTypes = mutableAllowedFileTypes;
         }
 
         if (post != nil) {


WordPress/Classes/Services/MediaService.m
```diff
@@ -57,7 +57,10 @@ - (Media *)createMediaWith:(id<ExportableAsset>)exportable
         }
         
         if (blog.allowedFileTypes != nil) {
-            allowedFileTypes = blog.allowedFileTypes;
+            // HEIC isn't supported when uploading an image, so we filter it out (http://git.io/JJAae)
```
shiki, line 60: As discussed, it's better to fix this client-side (iOS) for now. I confirmed that the "heic" is returned by:

- WPCom
- Jetpack
- Self-hosted sites

But in the future, the API should not return `heic` if it can't accept the file type.

WordPress/Classes/Services/MediaService.m
```diff
@@ -57,7 +57,10 @@ - (Media *)createMediaWith:(id<ExportableAsset>)exportable
...
+            NSMutableSet *mutableAllowedFileTypes = [blog.allowedFileTypes mutableCopy];
```
shiki, line 61: @leandroalonso Since the constraint is for all types of sites, I wonder if we should consider placing this condition in [`MediaAssetExporter`](https://github.com/wordpress-mobile/WordPress-iOS/blob/issue/14680_fix_heic_upload/WordPress/Classes/Utility/Media/MediaAssetExporter.swift) instead?

jkmassel:
I merged this to be able to cut a new beta tonight, hope that's ok!

leandroalonso:
@jkmassel Thank you!

PR #14692, opened 2020-08-20
Update MIGRATIONS.md

leandroalonso:
Add documentation about the Core Data model number 98.


guarani:
Looks good to me @leandroalonso :D

PR #14732, opened 2020-08-22
What's New: header and version scroll with content

### LINKED ISSUE #14718: Make header and version scrollable on What's New
Following up discussion in  #14698, the header and version labels should scroll with the table view

Gio2018:
Fixes #14718

To test:

- go to Me/App Settings/What's New in WordPress
- scroll the announcements and verify that the header and version scroll with them.
- Try different device sizes and orientations and make sure that view elements look right.

here are a couple of example screenshots

iPhone | iPad compact
--- | ---
whatsnewiphone | whatsnewipad

**Note: I added a few more fake announcements to emphasize the scrolling. These will all go away once the real data come in.**

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ Will be added in a separate PR.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ Will be added in a separate PR.
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ Not released

### DIFF
--- a/WordPress/Classes/ViewRelated/What's New/Dependency container/WPTabBarController+WhatIsNew.swift
+++ b/WordPress/Classes/ViewRelated/What's New/Dependency container/WPTabBarController+WhatIsNew.swift
@@ -41,6 +41,18 @@ extension WPTabBarController {
                                         Announcement(heading: "Try write headings that don't go beyond 2 lines",
                                                      subHeading: "If combined with three lines of subheading this is the longest an item should be.",
                                                      image: nil,
+                                                     imageUrl: nil),
+                                        Announcement(heading: "Heading with a single line",
+                                                     subHeading: "Try to write subheadings that run to a max of three lines. See how the icon is centered.",
+                                                     image: nil,
+                                                     imageUrl: nil),
+                                        Announcement(heading: "Heading with a single line",
+                                                     subHeading: "Subheading with only one line.",
+                                                     image: nil,
+                                                     imageUrl: nil),
+                                        Announcement(heading: "Try write headings that don't go beyond 2 lines",
+                                                     subHeading: "If combined with three lines of subheading this is the longest an item should be.",
+                                                     image: nil,
                                                      imageUrl: nil)]
         // TODO - WHATSNEW: to be removed when the real data come in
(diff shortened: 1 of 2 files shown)


osullivanchris (changes requested):
Thanks for the ping @Gio2018 

Its definitely better not having the header fixed, I think its a change for the better. However, I think I would expect the header and list just to scroll off the top of the screen. Here we are retaining the padding and scrolling behind it. Perhaps the space could be part of the list so it scrolls away too?

I'm also not loving the scrollbar aesthically. Is it recommended or required for accessibility or other reasons? If we do need to keep it, could it be aligned to the right side of the view rather than right to the edge of the text? I'm guessing we could achieve this by having right-padding or margin  for the text within the scrollable area, rather than outside the content area. Image below explaining.

There were some other issues discussed in #14698  but as this PR only references scrolling with content, I'm assuming we're dealing with other issues in separate PRs 👍 

scroll behaviour

Gio2018:
Thank you for reviewing @osullivanchris !!

> I think I would expect the header and list just to scroll off the top of the screen

I'm in total agreement, the "fixed" padding looks kinda odd and it's now fixed

> I'm also not loving the scrollbar aesthically

I agree on this too, removed!!

> There were some other issues discussed in #14698 but as this PR only references scrolling with content, I'm assuming we're dealing with other issues in separate PRs 👍

That is correct, so we can keep PRs as small as possible.

The changes have been pushed, once CircleCI is finished a new installable build will be available. Meantime, here's an animation that shows the updates

<p align=center>

</p>

osullivanchris (approved):
Awesome, glad you agree on the changes and thanks for the quick fix!

> That is correct, so we can keep PRs as small as possible.

makes sense! 

> The changes have been pushed, once CircleCI is finished a new installable build will be available. Meantime, here's an animation that shows the updates

great! Happy to approve based on the gif. But I'll also check out the build once its available.

Gio2018:
Thank you for the quick re-review, @osullivanchris !!

Gio2018:
Thank you for reviewing @bjtitus !!

PR #14735, opened 2020-08-24
Reader Improvements Phase 2: Fixes a potential crash 

emilylaguna:
Fixes an out of bounds issue I encountered this morning. I have never seen this before, so I don't know how reproducible it is for everyone. 

### To test:
1. Open the app, and enable the Reader Improvements Phase 2 feature flag
2. Tap on the Reader
3. Tap on the Discover tab
4. Scroll down past the related interests

### PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #14741, opened 2020-08-24
Reader Improvements: Adds accessibility support to the reader topics collection view 

emilylaguna:
**Review after**: https://github.com/wordpress-mobile/WordPress-iOS/pull/14735

This adds VoiceOver support to the new topics collection view. 

**Note: Accessibility Inspector doesn't work well. Please run on a real device**

### To test:
Install the One Off Build: 33581

1. Run the app on a real device
2. Enable the Reader Improvements Phase 2 feature flag
3. Enable VoiceOver 
4. Navigate to the Discover tab in Reader
5. Tap on the stream
6. Swipe right to highlight the next item in the list
7. Swipe until you get to the topics view in the reader card
8. Swipe to highlight the more item
9. Double tap to expand
10. Swipe left to highlight a topic
11. Double tap to select it

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsCollectionViewCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Select Interests/ReaderInterestsCollectionViewCell.swift
@@ -2,4 +2,16 @@ import UIKit
 
 class ReaderInterestsCollectionViewCell: UICollectionViewCell {
     @IBOutlet weak var label: UILabel!
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+
+        guard let label = self.label else {
+            return
+        }
+
+        label.isAccessibilityElement = true
+        accessibilityElements = [label]
+    }
+
 }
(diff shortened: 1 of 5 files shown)


leandroalonso (commented):
@emilylaguna two details:

* When selecting an interest, there's no difference between the selected state and the normal state (VO always reads just the name of the topic)
* Maybe VO should announce the chip as "Topic: foobar"? It's just an idea. I'm just thinking if only the tag names would mean something to the user. Perhaps yes?

emilylaguna:
> Maybe VO should announce the chip as "Topic: foobar"? It's just an idea. I'm just thinking if only the tag names would mean something to the user. Perhaps yes?

I originally did that actually, but honestly when thumbing through it just got a bit annoying. If anything I would like the first one to read: "Post topics: animals", then the rest can just be the topic title.

leandroalonso:
@emilylaguna Let's just leave the chips as it is right now. It's inside the post card so I guess this givens the user the context.

emilylaguna:
> When selecting an interest, there's no difference between the selected state and the normal state (VO always reads just the name of the topic)

I'm fixing this 👍

leandroalonso (approved):

PR #14743, opened 2020-08-25
Use persistent storage for @-mentions suggestions

guarani:
This changes the storage of @-mentions from an in-memory cache to a persistent Core Data store. This will allow us to provide an offline experience for @-mentions in the Gutenberg block editor.

Update: Other changes include updating `SuggestionService` to use a callback-based rather than a notification-based mechanism. 

Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/2577

@-mentions are used in three places in the app: Reader comments, Notification comments/replies, and the Gutenberg block editor.

Known issues:
- Reader: @-mentions not working on Atomic sites: https://github.com/wordpress-mobile/WordPress-iOS/issues/14844 (**existing bug**, not related to this PR)
- ~~@-mentions data is shared between users on the same device: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2600~~ no longer the case on iOS after https://github.com/wordpress-mobile/WordPress-iOS/commit/2c343d7e4c10bde8b51947d6af146fae7812c7f4
- Supporting @-mentions on Jetpack connected self-hosted sites: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2601
- Even if multiple sites return (via `rest/v1.1/users/suggest`) largely the same suggestions list, the app stores each list separately in Core Data. In one test where the user visited 5 sites each with ~1500 results, Core Data contained ~6500 `AtMentionSuggestion` records. However, these records are deleted when the user logs out.
- If the suggestion list changes on the server (e.g. a user is added, deleted or edited), the app will not see the change if it has already fetched a non-empty suggestion list for that site, or until the user logs back in again

## How to test

#### Test checklist template

```
- [ ] Reader comments
- [ ] Comment on post (via notification)
- [ ] Reply to comment (via notification)
- [ ] Gutenberg editor mentions
```

**What you'll need**: A WordPress.com site (may be your own site) that you can comment on and that has one or more other WordPress.com users (in order to test @-mentions). Make sure some of the users have a profile photo. The user that you login with will also need:
- a notification about a post
- a notification about a comment

### Reader comments

<details>
<summary>Test steps</summary>
<ol>
<li>In the Reader tab, locate the site</li>
<li>Navigate to a post, tap the comment button</li>
<li>Tap the &quot;Reply to post...&quot; text box to bring up the comment UI</li>
<li>Type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions to add it to your comment and hit send</li>
<li>Verify that the mentioned user receives a notification</li>
</ol>
</details>

### Comment on post (via notification)

<details>
<summary>Test steps</summary>
<ol>
<li>In the Notifications tab, locate a notification about a post</li>
<li>Open it, then tap the comment button, then the &quot;Reply to post...&quot; text box to bring up the comment UI</li>
<li>Type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions to add it to your comment and hit send</li>
<li>Verify that the mentioned user receives a notification</li>
</ol>
</details>

### Reply to comment (via notification)

<details>
<summary>Test steps</summary>
<ol>
<li>In the Notifications tab, locate a notification about a comment</li>
<li>Open it, then tap the &quot;Write a reply...&quot; text box to bring up the comment UI</li>
<li>Type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions to add it to your reply and hit send</li>
<li>Verify that mentioned user receives a notification</li>
</ol>
</details>

### Gutenberg editor mentions

<details>
<summary>Test steps</summary>
<ol>
<li>Open the block editor</li>
<li>In the &quot;Start writing...&quot; section, type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions and publish the post</li>
<li>Verify that mentioned user receives a notification</li>
</ol>
</details>

In addition to the above tests, I also did extra testing (which is not necessary for PR review):
- filtering names by typing after the @ character and ensuring the list is filtered
- tested on sites that contain hundreds (or thousands) of users
- fresh installs for each of the above tests to make sure nothing is carried over from one test to the next
- Switch the device offline after doing the above tests, and ensure the offline (cached) results are shown
- Tested edge case where an empty list of suggestions is retrieved (e.g. when a user has the role Contributor on a site, that user is not able to retrieve a list of suggestions)

---

PR submission checklist:


guarani:
Testing build 33945:
The @-mentions suggestion list isn't shown sometimes when the @ character is typed — after navigating to previous screen and trying again it normally works.

guarani:
> Testing build 33945:
> The @-mentions suggestion list isn't shown sometimes when the @ character is typed — after navigating to previous screen and trying again it normally works.

Fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/14743/commits/72ec0d734432d0872e0b410751d119de0306c7ce

guarani:
Tested on build 34422
- [x] Reader comments
- [x] Comment on post (via notification)
- [x] Reply to comment (via notification)
- [x] Gutenberg editor mentions

guarani:
@leandroalonso we discussed offline whether Core Data is the right solution here and if an alternative like archived objects could work.

Sites can return hundreds (possibly thousands) of users, and because we're looking to implement offline support for @-mentions in Gutenberg, Core Data to me feels right — but I could be wrong. 

We've had some issues with Core Data, but I'm not sure if we've started moving away from it, or if other solutions are better in their own right.

While I could implement an alternative to Core Data here, such as using archived objects, and compare — I don't think it's scalable enough (Core Data will allow us to search on disk without fetching all objects to memory), and IMO Core Data seems to do the job here. I'm not adding relationships to existing entities, so hopefully this doesn't add more complexity to our setup than needed.

mchowning (approved):
This looks good and works well! Nice job @guarani ! 👍 

I'm not familiar with the database setup on iOS, so hopefully @emilylaguna  or @leandroalonso can take a look at this and give their more knowledgeable opinion about the implementation. 😄

leandroalonso:
On it!

Autocompleter+CoreDataClass.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+import CoreData
+
+@objc(Autocompleter)
+public class Autocompleter: NSManagedObject {
```
leandroalonso: Do you think we really need this entity? `Autocompleter` doesn't seem like an entity, it looks to me that we can add the `siteID` as a relationship in `UserAutocomplete` and things are going to work the same way. Wdyt?

UserAutocomplete+CoreDataClass.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+import CoreData
+
+@objc(UserAutocomplete)
+public class UserAutocomplete: NSManagedObject {
```
leandroalonso: I guess that even though this entity exists for the autocomplete, in the end, it's just usernames.

Then I was looking into our database schema and we already have a very similar entity which is `BlogAuthor`. All the username suggestions are users added to a specific blog, right?

Maybe that's something you have already tried but `BlogAuthor` wouldn't suit us here?

WordPress/Classes/Services/SuggestionService.m
```diff
@@ -91,23 +82,84 @@ - (void)updateSuggestionsForSiteID:(NSNumber *)siteID
     }];
 }
 
+- (Autocompleter *)retrieveAutocompleterForSiteID:(NSNumber *)siteID
```
leandroalonso: We don't need to do it now but keep in mind for upcoming changes: given the amount of code we changed/added here we could have used this opportunity to migrate this code to Swift. :D

WordPress/Classes/Services/SuggestionService.m
```diff
@@ -91,23 +82,84 @@ - (void)updateSuggestionsForSiteID:(NSNumber *)siteID
...
+{
+    NSManagedObjectContext *context = [[ContextManager sharedInstance] mainContext];
+
+    NSFetchRequest *request = Autocompleter.fetchRequest;
+    [request setPredicate:[NSPredicate predicateWithFormat:@"siteID == %@", siteID]];
+    NSError *error;
+    NSArray *sites = [context executeFetchRequest:request error:&error];
+    if (error) {
+        DDLogError(@"Could not fetch Autocompleter for site %@, error", siteID, error);
```
leandroalonso:
```suggestion
        DDLogError(@"Could not fetch usernames for site %@, error", siteID, error);
```

WordPress/Classes/Services/SuggestionService.m
```diff
@@ -91,23 +82,84 @@ - (void)updateSuggestionsForSiteID:(NSNumber *)siteID
...
+        return nil;
+    } else if (sites.count > 1) {
+        DDLogError(@"Retrieved more than one Autocompleter for site %@", siteID);
```
leandroalonso: In this case, can't we just return the firstObject?

WordPress/Classes/Services/SuggestionService.m
```diff
@@ -91,23 +82,84 @@ - (void)updateSuggestionsForSiteID:(NSNumber *)siteID
...
-    
-    WordPressAppDelegate *appDelegate = [WordPressAppDelegate shared];
-    
-    NSArray *suggestions = [self.suggestionsCache objectForKey:siteID];
-    
-    // if the device is offline and suggestion list is not yet retrieved
-    if (!appDelegate.connectionAvailable && !suggestions) {
+
+    // if the device is offline
+    if (![WordPressAppDelegate shared].connectionAvailable) {
```
leandroalonso: If the device is offline can't we just rely on the cached data? I guess it worked like that before. Or am I missing something?

leandroalonso (commented):
@guarani tested all the cases, it works as described!

I left a few suggestions about the implementation. Let me know what do you think. :)

WordPress/Classes/Services/SuggestionService.m
leandroalonso: Or perhaps, "username suggestions" or "user suggestions"

Autocompleter+CoreDataClass.swift
guarani: Originally I was going to just use `UserAutocomplete` with the `siteID` as a property, but that wouldn't allow the app to differentiate between a scenario where a site's mentions were fetched from the network and there were none (empty array), and a scenario where we haven't yet fetched a site's mentions. 

This would be analogous to an optional property, e.g. an a property of type array that can be either `nil` OR an empty array. I think the only way to model this in Core Data is with an extra entity, so I created `Autocompleter` (and moved the `siteID` there). Now, if there's no `Autocompleter` for a particular `siteID`, or there is but its `UserAutocomplete` is `nil`, the app knows that it hasn't yet fetched mentions for that site.

Soon, once we implement x-posting in Gutenberg, my idea here was to add another entity (e.g. `XPostAutocomplete`) which will also have a relationship to `Autocomplete`. (Or possibly refactor `UserAutocomplete` to support x-posting, if the data we need to store is generic enough.)

Does this sound like a good approach in your opinion?

UserAutocomplete+CoreDataClass.swift
guarani: Good question, I hadn't seen `BlogAuthor`. Since `Suggestion` already existed, I didn't think to look for another class that could handle this — thanks for bringing it up.

One thing I didn't mention in this PR is that the Gutenberg team is currently implementing x-posting, which will be very similar to @-mentions.

Now the question comes up of whether instead of `UserAutocomplete` I could reuse `BlogAuthor` — and while I'm at it, use `Blog` in the future for x-posting. There's a clear advantage to reusing the entities we already have instead of creating new ones, especially if they make sense semantically. On the other hand, some things to keep in mind are:
- to audit existing use of `BlogAuthor` and `Blog` to make sure that existing screens will not be affected by the fact that these same entities will now be used for additional functionality (mentions and x-posting)
- both entities have non-optional attributes that mentions and x-posts won't be able to use (e.g. `userID` on `BlogAuthor`) and many non-optional properties on `Blog` that the x-post endpoint doesn't return — converting to non-optional could require significant effort

I think a good idea would be to check if the backend services that provide data for mentions and x-posts internally use the same entities as the services that populate `BlogAuthor` and `Blog` objects. If they do, that would be a good indication that we should do the same, but if not, that we should keep the separate entities. I'll try to look into this a bit and see what I can find, but please let me know if you have an thoughts on the above.

WordPress/Classes/Services/SuggestionService.m
guarani: Thanks! I'll improve this.

WordPress/Classes/Services/SuggestionService.m
guarani: This is a tough call, I'd normally say, wouldn't it be better to fail right here as soon as we know the app is in an unexpected state? In any case, I would leave the logging there.

WordPress/Classes/Services/SuggestionService.m
guarani: I didn't change the logic here because this PR was only meant to change change the storage mechanism from in-memory to disk, in a future PR my plan is to implement offline mode (which might include other thing like an offline UI, etc).

WordPress/Classes/Services/SuggestionService.m
guarani: Good point. Do we normally rewrite to Swift things that get significant changes?

guarani:
> @guarani tested all the cases, it works as described!
> I left a few suggestions about the implementation. Let me know what do you think. :)

Thanks a lot for the review @leandroalonso! 

Sorry I didn't mention earlier about how x-posting will be implemented soon and will reuse some of this. Please let me know what you think based on my replies above. 

Just a quick question, we have multiple names for the same thing: mentions, suggestions, autocompleters. Perhaps I should stick to "suggestions" and refactor the names here, wdyt?

Autocompleter+CoreDataClass.swift
leandroalonso: Hmm, making additional calls would be a problem here?

In the first solution, we had a session cache-based, maybe we can implement a similar logic here and keep the database unaware of it? If the user closes the app and reopens again, it will make a new request, but I don't think that's a problem.

My main reasoning here is to try to remove this kind of knowledge from the database layer (the database should only model the objects).

I'm not sure I fully understood the x-posting. This will autocomplete based on blog URLs?

UserAutocomplete+CoreDataClass.swift
leandroalonso: I agree with all your points. It would be awesome if we could reuse those entities (since, correct me if I'm wrong, the mentions are blog authors) but we need to be cautious about changing what already exists.

I checked the payload for the suggestions and they have: `ID`, `user_login`, `display_name` and `image_URL` (it's missing `email`,`linkedUserID` and `primaryBlogID`).

I know that `BasePost` is used by `Post` and `ReaderPost`, for example. Maybe instead of using `BlogAuthor` we can just do a similar thing? Then this won't affect the parts of the code that already uses it.

Regarding `x-posts` I don't think `Blog` will suit us in this case.

WordPress/Classes/Services/SuggestionService.m
leandroalonso: Hmm, I was comparing with the previous logic, it was:

```
     if (!appDelegate.connectionAvailable && !suggestions) { 
```

The current logic doesn't check for existing suggestions, so it wouldn't show it while offline, right?

leandroalonso:
> Just a quick question, we have multiple names for the same thing: mentions, suggestions, autocompleters. Perhaps I should stick to "suggestions" and refactor the names here, wdyt?

Good point! I'd stick with suggestions like you mentioned. :)

WordPress/Classes/Services/SuggestionService.m
guarani: True! I should re-enable the previous logic for suggestions in Reader to keep things like they were before. I might disable offline support for @-mentions in Gutenberg though, since that will be handled separately in a later PR (with extra UI to indicate to the user that the list of @-mentions is not fresh).

UserAutocomplete+CoreDataClass.swift
guarani: > I checked the payload for the suggestions and they have: ID, user_login, display_name and image_URL (it's missing email,linkedUserID and primaryBlogID).

`BlogAuthor` is populated by the `sites/$site_id/users` endpoint, which doesn't actually include the user's display name.
<details>
<summary>See JSON:</summary>
<pre>
    {
      "ID": 190912407,
      "login": "paulvonschrottky",
      "email": "paulvonschrottky@gmail.com",
      "name": "paulvonschrottky",
      "first_name": "",
      "last_name": "",
      "nice_name": "paulvonschrottky",
      "URL": "",
      "avatar_URL": "https://0.gravatar.com/avatar/f2976ad632f6923d61ebe9acbb8d852f?s=96&d=identicon&r=G",
      "profile_URL": "https://en.gravatar.com/paulvonschrottky",
      "ip_address": "",
      "site_ID": 171152168,
      "site_visible": true,
      "roles": [
        "editor"
      ],
      "is_super_admin": false
    },
</pre>
</details>

> I know that BasePost is used by Post and ReaderPost, for example. Maybe instead of using BlogAuthor we can just do a similar thing? Then this won't affect the parts of the code that already uses it.

To do this properly, it would probably be best to remove `displayName` from `BlogAuthor` (since it's not currently used and doesn't belong there) and put it on the new `BlogAuthor` subclass. 

> Regarding x-posts I don't think Blog will suit us in this case.

Yeah, that's my feeling as well.

WordPress/Classes/Services/SuggestionService.m
guarani: I decided to port `SuggestionService` and `Suggestion` to Obj-C: https://github.com/wordpress-mobile/WordPress-iOS/pull/14888

guarani: @leandroalonso If you could take a look at that PR (https://github.com/wordpress-mobile/WordPress-iOS/pull/14888), that would be great! Once it's merged I'll update this PR, that should make this PR a bit smaller too.

guarani: https://github.com/wordpress-mobile/WordPress-iOS/pull/14888 was merged so resolving this 👍

WordPress/Classes/Services/SuggestionService.m
guarani: Marking this as resolved because the latest commits got rid of this code.

WordPress/Classes/Services/SuggestionService.m
guarani: Marking this as resolved because the latest commits get rid of this code.

WordPress/Classes/Services/SuggestionService.m
guarani: Marking this as resolved because the latest commits check this condition properly, so this is fixed.

UserAutocomplete+CoreDataClass.swift
guarani: (fyi: I've made significant changes since the previous comment)

Instead of using two entities as I did before, I now just use one new entity (`AtMentionSuggestion`). By adding a 1-to-many nullable relationship from `Blog` to `AtMentionSuggestion`, I haven't lost any of the functionality of the previous approach while making things much simpler.
My gut feeling is that this is the most intuitive approach (rather than involving `BlogAuthor`) and only adds one new entity.

Autocompleter+CoreDataClass.swift
guarani: My latest commits remove this entity, you were right about this not being needed for the reasons I shared above. An optional 1-to-many relationship from `Blog` to `AtMentionSuggestion` allows me to differentiate between the app not having fetched suggestions yet, and having fetched but getting an empty list.

The only drawback of my new approach is: imagine a user (e.g. an A11n) who has 5 sites each with 1000 @-mentions suggestions for each site would result in 5000 rows in the `AtMentionSuggestion` entity table. What's important to keep in mind though is that `AtMentionSuggestion` is cleared when the user closes their session (it's wired to cascade delete when the `Blog` are deleted, which happens on logout.)

guarani:
I'm going to do some more testing before re-requesting reviews.

guarani:
Testing build 36197
- [ ] Reader comments
- [ ] Comment on post (via notification)
- [ ] Reply to comment (via notification)
- [ ] Gutenberg editor mentions

guarani:
Testing build 36205
- [x] Reader comments
- [x] Comment on post (via notification)
- [x] Reply to comment (via notification)
- [x] Gutenberg editor mentions

guarani:
👋 @leandroalonso, would you mind giving this another look? 

I've made some changes based on feedback and the refactor in https://github.com/wordpress-mobile/WordPress-iOS/pull/14888, which has reduced the number of new Core Data entities from two to just one, `AtMentionSuggestion`.

WordPress/Classes/Models/AtMentionSuggestion+CoreDataClass.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+import CoreData
+
+@objc(AtMentionSuggestion)
+public class AtMentionSuggestion: NSManagedObject {
```
leandroalonso: It took me a little bit to understand that `At` refers to @, I was thinking in ATomic :P

What do you think about rename this entity to represent what it is? Like `UserSuggestion`? Then in the future we will also have `SiteSuggestion`. I guess it makes a little bit easier to understand. Wdyt?

leandroalonso (commented):
@guarani left one small suggestion, let me know what do you think. :)

WordPress/Classes/Models/AtMentionSuggestion+CoreDataClass.swift
guarani: Good idea. I spent too much time with @-mentions that I'd overlooked how AtMention might be confused with **At**omic.

guarani: Renamed in af9fc2cd7f08defb3157de4eef8216df55d827c2

leandroalonso (approved):
Thanks for the patience and changes, @guarani. :)

We're good to go!

guarani:
Not a problem at all. Thank you very much for the review, @leandroalonso! Much appreciated

PR #14631, opened 2020-08-12
[Modal Layout Picker] Create Layout Picker category bar UI using static data

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2437

## To test:

<details>
<summary><strong>To disable or enable the development version of Modal Layout Picker</strong></summary>

- Open the app from the build that allows FeatureFlags such as a PR build or a local development build
    - By default, the modal layout picker will be disabled.
    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
        - Toggle "Gutenberg Modal Layout Picker" to enable or disable the picker
            
</details>

When enabled, the Layout Picker should show when creating a new page from My Site and from the Page list.

Start by enabling the MLP (if needed) and by navigating to the Modal Layout Picker

### Select a Category
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Deselect the category
    - **Expect** to see the other categories return

### Select Multiple-Categories 
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Select a different category
    - **Expect** to see the new section added to the list
1. Deselect a category
    - **Expect** the list to update again to be only the selected categories
1. Deselect the rest of the categories
    - **Expect** the all of the categories to return

### Selecting a Layout then filter on the same category
1. Select a layout in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the selected layout to still be selected.
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout to still be selected.

### Selecting a Layout then filter on the same category deselect layout
1. Select a layout in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the selected layout to still be selected.
1. Deselect the layout
    - **Expect** the bottom bar to revert to "Create a Blank Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for no layout to be selected

### Selecting a Layout then filter on a different category
1. Select a layout in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout to still be selected.

[...]

PR #14750, opened 2020-08-25
[Modal Layout Picker] "Create Page" button functionality

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2446

## To test:
<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
            
</details>

🗒️ The layouts that are provided right now do contain a lot of unsupported blocks. This will be filtered out by a later PR.

### From My Site
1. On the My Site Page
1. Click the FAB icon 
1. Select Site Page
1. Choose a Layout
1. Select Create Page

**Expect** to see the Editor loaded with the contents of your selected block

### From Pages
1. On the My Site Page
1. Open the Page list
1. Select the ➕ icon
1. Choose a Layout
1. Select Create Page

**Expect** to see the Editor loaded with the contents of your selected block

### From Deep Link 
1. Make sure the test build is the only build installed
   - This is to make sure we're following the directions in https://github.com/wordpress-mobile/WordPress-iOS/pull/13707
     > All other copies of the app must be uninstalled since I believe it is random which copy of the app is chosen to open the URL
1. From another app (I used notes but sending a text in messages should also work) set up the link `https://apps.wordpress.com/get#/page`
1. Open the link by clicking on it
    - **Expect** to see the Layout picker
1. Choose a Layout
1. Select Create Page

**Expect** to see the Editor loaded with the contents of your selected block

### From Deep Link with specific domain
1. Make sure the test build is the only build installed
   - This is to make sure we're following the directions in https://github.com/wordpress-mobile/WordPress-iOS/pull/13707
     > All other copies of the app must be uninstalled since I believe it is random which copy of the app is chosen to open the URL
1. From another app (I used notes but sending a text in messages should also work) set up the link `https://apps.wordpress.com/get#/page/yourAwesomeDomain.com`
    - Make sure you replace `yourAwesomeDomain.com` with a domain linked to your app 😉 
1. Open the link by clicking on it
    - **Expect** to see the Layout picker
1. Choose a Layout
1. Select Create Page

**Expect** to see the Editor loaded with the contents of your selected block

### From Deep Link with content
1. Make sure the test build is the only build installed
   - This is to make sure we're following the directions in https://github.com/wordpress-mobile/WordPress-iOS/pull/13707
     > All other copies of the app must be uninstalled since I believe it is random which copy of the app is chosen to open the URL
1. From another app (I used notes but sending a text in messages should also work) set up the link `wpdebug://newpage?content=test`
1. Open the link by clicking on it

**Expect** to see the Editor loaded with the content from your deep link. Note this seems to be loading the classic editor even if Gutenberg is configured for your domain.

## Screenshots:

| My Sites | Pages |
| --- | --- |
| CreatePage | CreatePage_pages |

## Additional Context: 

This is a continuation of the work done in https://github.com/wordpress-mobile/WordPress-iOS/pull/14726 and will be followed up by other tickets as part of [Modal Layout Picker Project](https://github.com/wordpress-mobile/gutenberg-mobile/issues?q=is%3Aissue+is%3Aopen+%5BModal+Layout+Picker%5D)

There are some design considerations that are still being figured out. You can see them documented here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2563 

## PR submission checklist:


GhBsx-png (commented):
Thank you for your assistance

WordPress/Classes/ViewRelated/System/WPTabBarController+ShowTab.swift
```diff
@@ -15,17 +15,21 @@ extension WPTabBarController {
             }
             return
         }
-
         guard let blog = inBlog ?? self.currentOrLastBlog() else { return }
+        guard content == nil else {
+            showPageEditor(blog: blog, title: title, content: content)
```
chipsnyder: From what I can tell the content property of this call is only ever populated from a deeplink that is no longer hooked up. (path = `newPage`) But I kept the support in place.

guarani: If `content` is **not** `nil` (e.g. via the deeplink), this calls the same function again and appears to lead to an infinite loop.

There was [a recent PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/13707) that implemented a deeplink to open a new Page in the editor. I tested these below links in the current WPiOS 15.5 version and they work, and then went to check if they still work using the Alpha build you provided on this PR.

```
https://apps.wordpress.com/get#/page
https:/apps.wordpress.com/get#/page/<domain>
```

The trouble is that the alpha build uses a different URL scheme so these links actually took me to the App Store to download WPiOS 15.5. I thought that these links might redirect to `wpalpha://newpage?content=test` (as seen on that PR), but these opened the Classic editor — so I hit a bit of a dead end.

I hope this has been helpful, sorry I couldn't come up with something conclusive here. Would you like to test more yourself?

guarani (commented):
I gave this a test and it worked great overall. I didn't test different devices, iOS versions, orientations, etc. [like last time](https://github.com/wordpress-mobile/WordPress-iOS/pull/14631#pullrequestreview-473790959), because this PR is more about adding logic than UI.

I left a comment about the code [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/14750#discussion_r481242555). 

Regarding the design blueprint though, is the snack bar message supposed to be shown after creating a page from a template?

Screen Shot 2020-09-01 at 11 33 36

WordPress/Classes/ViewRelated/System/WPTabBarController+ShowTab.swift
guarani: I updated the above to fix a typo (bolded) 😅

chipsnyder: I did miss that that PR it looks like we should be able to deploy locally and test using `wpdebug://newpage?content=test` I'll give this a try and see what happens. Thanks so much for sharing this 🙇 

That method signature is actually the same that I use for the path but swift figures it out. Looking at this though it is very confusing and unclear on how it will behave 😅 . I'll rename that function. I must have been too focused when working on this one.

Let me take another pass at this and I'll drop a comment when it's ready for revie again. Thanks so much @guarani :)

chipsnyder:
> Regarding the design blueprint though, is the snack bar message supposed to be shown after creating a page from a template?

@guarani Good question on the toast message. We actually decided to handle that in a separate Issue just to keep this one focused https://github.com/wordpress-mobile/gutenberg-mobile/issues/2453

chipsnyder:
Hey @guarani 
This is ready to go again now I added some additional test cases for deeplink flows

guarani:
> From another app (I used notes but sending a text in messages should also work) set up the link 
> https://apps.wordpress.com/get#/page

@chipsnyder, I tested again and ran into the same issue mentioned [above](https://github.com/wordpress-mobile/WordPress-iOS/pull/14750#discussion_r481242555). I think I get why it's not working, since those URLs would only point to the production app, but I'm curious how you got it to work.

chipsnyder:
Hey @guarani I was able to get it to work by installing the app through XCode then I pasted the links in the notes app. From there clicking on the notes app worked fine. I only had one instance of WordPress installed though and no instances of the Gutenberg demo app either. Hopefully that helps.

guarani (approved):
Thanks for the tip about installing from Xcode, the deeplinks worked there.
I tested the three scenarios and all are working as expected 👍

PR #14772, opened 2020-08-26
Fixes an issue with the save password prompt.

diegoreymendez:
WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/417

Fixes an issue that I found while testing the login flow.

The original issue was that the App was prompting the user to update their password in the keychain, in a scenario where the user was dismissing the VC by pressing "< Back".

## Notes:

I'm not sure if you need to have existing iCloud Keychain credentials for the original issue to come up.  The steps below have been enough for me to reproduce the issue.

## Steps:

These steps can be used both for testing that the issue is fixed (in this branch) and for seeing the original issue (in `develop`).

1. Press "Login"
2. Press "Continue with WP.com"
3. Enter a valid email (don't use any form of autocompletion as it will skip the password view)
4. At the magic link screen, tap “Enter your password instead”
5. Type any wrong password but DO NOT click "Next"
6. Tap “< Back”
7. Wait a bit

In `develop`: iOS will ask if you want to update the credentials with the password you inserted.  This isn't correct since you have't submitted anything and cancelled the operation.

In this branch: iOS will NOT ask you if you want to update your saved password.

## PR submission checklist:


mindgraffiti (approved):
✅ Tested on my device

Issue #14767, opened 2020-08-26
Reader Improvements: Reader Detail Updates


emilylaguna:
Project Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14768

### Goal
Update the Reader Detail design

### Screenshots
<img width="1030" alt="Screen Shot 2020-08-26 at 3 16 21 PM" src="https://user-images.githubusercontent.com/793774/91346772-2a5bee80-e796-11ea-81bc-d0dd7a8fab06.png">

### Designs
- [Explorations](https://www.figma.com/file/a5O8CmzVbGbWBpqzrPUyiC/Reader-Detail-View?node-id=1%3A1285)
- [Specs](https://www.figma.com/file/yG8ieha0yESnr8wCtplr29/Post-Detail-View?node-id=0%3A1) 

### Tasks
- [x] Fix issues introduced by merging develop: https://github.com/wordpress-mobile/WordPress-iOS/pull/15051

#### Header
- [x] Update reader detail navigation buttons to be 44px: https://github.com/wordpress-mobile/WordPress-iOS/pull/14867
- [x] Add swipe to back: https://github.com/wordpress-mobile/WordPress-iOS/pull/14860
- [x] Fix slow loading of featured image: https://github.com/wordpress-mobile/WordPress-iOS/pull/14884
- [x] Fix scrolling back to the top when the user rotates the device: https://github.com/wordpress-mobile/WordPress-iOS/pull/14889
- [x] Create fake navigation controller buttons to appear over the featured images: https://github.com/wordpress-mobile/WordPress-iOS/pull/14821
- [x] Move the featured image to be pinned to the top edges of the view: https://github.com/wordpress-mobile/WordPress-iOS/pull/14821
- [x] Add a gradient view over the featured image: https://github.com/wordpress-mobile/WordPress-iOS/pull/14821
- [x] Add support for the scrolling transition for the nav bar: https://github.com/wordpress-mobile/WordPress-iOS/pull/14821
- [x] iPad support for the featured image: https://github.com/wordpress-mobile/WordPress-iOS/pull/14821

#### Content
- [x] Add the post tags section with the expandable tags view: https://github.com/wordpress-mobile/WordPress-iOS/pull/14822
- [x] Reposition the detail sections to be inline with the designs
- [x] Add a comment entry point at the bottom of the detail view: https://github.com/wordpress-mobile/WordPress-iOS/pull/15040
[...]

PR #14773, opened 2020-08-27
Hide the tab bar when presenting Reader Detail

leandroalonso:
Part of #14767 

This PR:

* Shows the Reader Detail hiding the Tab Bar (thanks @emilylaguna)
* Removes the show/hide effect in the navigation bar and action bar in Reader Detail

Note that `AMScrollingNavBar` is still present in `Podfile`. 

## To test

1. Run the app
2. Go to Reader
3. Tap any post
4. Check that the tab bar is not visible
5. Scroll the article
6. Check that the navigation bar and the action bar stays fixed

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -82,15 +82,9 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
     }
 
     override var hidesBottomBarWhenPushed: Bool {
-        set {
-
-        }
-
-        get {
-            true
-        }
+        set { }
+        get { true }
     }
-
     override func viewDidLoad() {
         super.viewDidLoad()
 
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -82,6 +81,16 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+    override var hidesBottomBarWhenPushed: Bool {
+        set {
+
+        }
+
+        get {
+            true
+        }
+    }
+
```
emilylaguna: Maybe condense this down a few lines? 

```suggestion
    override var hidesBottomBarWhenPushed: Bool {
        set { }
        get { true }
    }
```

emilylaguna (changes requested):
Looks good, 1 small comment.

emilylaguna (approved):

Issue #14240, opened 2020-06-03
Improve the script that patches the Pods inside post_install hook


pinarol:
We have a post_install hook in the Podfile and have a piece of Script there that patches a couple of files from React Native.

```
        puts 'Patching RCTShadowView to fix nested group block - it could be removed after upgrade to 0.62'
        %x(patch "#{project_root}/Pods/React-Core/React/Views/RCTShadowView.m" < "#{project_root}/patches/react-native+0.61.5.patch")
        puts 'Patching RCTActionSheet to add possibility to disable action sheet buttons -
        it could be removed once PR with that functionality will be merged into RN'
        %x(patch "#{project_root}/Pods/React-RCTActionSheet/RCTActionSheetManager.m" < "#{project_root}/patches/react-native+0.61.5.patch")
```

Currently when it fails, it fails silently and it is hard to notice [these kinds of issues](https://github.com/wordpress-mobile/WordPress-iOS/pull/14239). So, instead of using `%x(...)` can we experiment with replacing that line with `system('patch ... < ...') || exit(1)`. `system(...)` returns the exit value of the shell so it can help us detecting when the patch fails. But, we should make sure it is not producing false positives and causing the termination of the process unnecessarily. 

**To Test:**

Clear Pods folder
cd to Script folder and run `bundle exec pod install`
Check the logs to see that no errors are produced during patching
Open the patched file /Users/pinarolguc/Development/WordPress-iOS/Pods/React-RCTActionSheet/RCTActionSheetManager.m make sure this line exists:
NSArray<NSNumber *> *disabledButtonIndices = [RCTConvert NSArray:options[@"disabledButtonIndices"]];

 run `bundle exec pod install` again and make sure it is not producing errors

clear Pods folder
run `rake dependencies` in the root folder(WordPress-iOS)
Check the logs again

run `rake dependencies` again and make sure it is not producing errors

bonus: open the gutenberg editor and see the mover buttons getting disabled

chipsnyder:
This is being updated again as part of https://github.com/wordpress-mobile/WordPress-iOS/pull/14776

PR #14776, opened 2020-08-27
Upgrade React Native to 0.63

ceyhun:
Upgrading react-native to version 0.63.2

To test: Follow instructions on gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2580

PR submission checklist:


Podfile
```diff
@@ -222,14 +223,12 @@ target 'WordPress' do
...
         puts 'Patching RCTActionSheet to add possibility to disable action sheet buttons -
         it could be removed once PR with that functionality will be merged into RN'
-        %x(patch "#{project_root}/Pods/React-RCTActionSheet/RCTActionSheetManager.m" < "#{project_root}/patches/RN-RCTActionSheetManager.patch")
-        puts 'Patching RCTUIImageViewAnimated to fix a problem where images will not load when built using the iOS 14 SDK (Xcode 12) -
-        it can be removed once we upgrade Gutenberg to use RN 0.63 or later'
-        %x(patch "#{project_root}/Pods/React-RCTImage/RCTUIImageViewAnimated.m" < "#{project_root}/patches/RN-RCTUIImageViewAnimated.patch")
+        %x(patch "#{project_root}/Pods/React-CoreModules/RCTActionSheetManager.mm" < "#{project_root}/patches/RN-RCTActionSheetManager.patch")
+        puts 'Patching FBReactNativeSpec to add possibility to disable action sheet buttons -
+        it could be removed once PR with that functionality will be merged into RN'
+        %x(patch "#{project_root}/Pods/FBReactNativeSpec/FBReactNativeSpec/FBReactNativeSpec.h" < "#{project_root}/patches/FBReactNativeSpec.patch")
```
chipsnyder, line 231: This might fall too far outside of the scope of what we're trying to take on here (update to RN 0.63) but I wonder if it would be possible to apply the suggested changes from https://github.com/wordpress-mobile/WordPress-iOS/issues/14240 since we're modifying these patches.

https://github.com/wordpress-mobile/WordPress-iOS/issues/14240
> So, instead of using %x(...) can we experiment with replacing that line with system('patch ... < ...') || exit(1). system(...) returns the exit value of the shell so it can help us detecting when the patch fails. But, we should make sure it is not producing false positives and causing the termination of the process unnecessarily.

ceyhun: I think with the 0.64 upgrade none of these patches here will be necessary anymore. This final [remaining patch](https://github.com/facebook/react-native/commit/a7c1c5aff24671bba609caeb82092a8de3d3b232) seems [to be merged into 0.64](https://github.com/react-native-community/releases/blob/master/CHANGELOG.md#v0640).

PR #14777, opened 2020-08-27
Fix empty notifications on release builds failures

loremattei:
We noticed that sometimes empty notifications are posted on Slack when the Release Build task fails on CI. It turns out that we currently setup the notification message quite late in the flow in order to be able to extract the version number, so if failures happen during checkout or early install, the message is empty. 

This PR adds the setup of a generic failure message early in the flow that will be replaced by the current one if the process makes it to when we can extract the version number. 

To test:
 - Checkout a new branch from this one. 
 - Add a failure in the Release Build flow before the Build step, commit and push.
 - Tag with a beta tag and push it. 
 - Verify that the failure message is published on Slack.

** Clean up**
 - Remove the test beta tag.
 - Remove the test branch

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -153,6 +153,10 @@ jobs:
     environment:
       HOMEBREW_NO_AUTO_UPDATE: 1
     steps:
+      - run:
+          name: Setup Notifications
+          command: |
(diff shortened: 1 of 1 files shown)


jkmassel (approved):
This worked, and the notification appeared in Slack! (I just re-ran the job that was already queued up)

:shipit:

loremattei:
The _Release Build_ job failed because we tested on the HEAD commit of this PR, so it's expected.

PR #14788, opened 2020-08-28
Update Reader Detail's header

leandroalonso:
Part of #14767

This PR updates the Reader Detail's Header. Note that the navigation bar and featured image is out of the scope.

Also, as part of this PR, I fixed the transparent space at the bottom part of the screen.

| Light  | Dark |
| Simulator Screen Shot - iPhone 11 - 2020-08-28 at 13 20 25 | Simulator Screen Shot - iPhone 11 - 2020-08-28 at 13 20 34 |

## To test

Most of the changes are visual, make sure to test in iPad and iPhone, light and dark mode.

### Follow / Unfollow

1. Open any post in the Reader
2. Check that the button at the right side of the blog details reflects the correct state (following or not)
3. Tap on it
4. Check that you unfollowed or followed the blog

PR submission checklist:


leandroalonso:
An iPad screenshot:

mbshakti:
The changes are looking so fresh!!! So excited for this. A few comments:

- I wasn't able to see topic chips on the post detail view in this build. Is that intentional?

- For the author attribution, can we show the author name instead of author username? I think the post detail view today shows the name as well, not the user name. 

- What is the font style of the author attribution + time stamp? the attribution + time stamp seem a bit larger than the sitename/url but I think they were both supposed to use `subheadline` style. 

- What font style is the post title using? In the specs, I used `title 2` but in the build it seems smaller than title 2. 

- Maybe not relevant to this PR, but there seems to be some spacing missing above the post title (if there's no featured image). Attached screenshot below:
Screen Shot 2020-08-28 at 2 06 00 PM

leandroalonso:
@mbshakti 

> I wasn't able to see topic chips on the post detail view in this build. Is that intentional?

We'll work on that in another task!

> For the author attribution, can we show the author name instead of author username? I think the post detail view today shows the name as well, not the user name.

Done! My mistake. My intention was to display the author's name (not username).

> What is the font style of the author attribution + time stamp? the attribution + time stamp seem a bit larger than the sitename/url but I think they were both supposed to use subheadline style.

The blog title + address was using constant sizes. I changed them to be .subhead (Blog title) and .footnote (URL). Here's the result:



(.subhead = SF 15pt)

> What font style is the post title using? In the specs, I used title 2 but in the build it seems smaller than title 2.

Hmm, It's using `.title2` + semibold. Here's how it looks with `.title1`:



> Maybe not relevant to this PR, but there seems to be some spacing missing above the post title (if there's no featured image). Attached screenshot below:

This will be fixed with the featured image + nav bar at the top!

mbshakti:
Thanks for addressing the feedback, looks great!

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
```diff
@@ -320,6 +320,16 @@ class ReaderDetailCoordinator {
         viewController?.present(controller, animated: true)
     }
 
+    private func followSite() {
+        guard let post = post else {
+            return
+        }
+
+        ReaderFollowAction().execute(with: post, context: coreDataStack.mainContext) { [weak self] in
```
emilylaguna: Can we pre-emptively toggle the follow button state? Right now it feels super laggy when I hit the follow button

WordPress/WordPressTest/ReaderDetailCoordinatorTests.swift
```diff
@@ -355,6 +355,10 @@ private class ReaderDetailViewMock: UIViewController, ReaderDetailView {
         didCallScrollToWith = to
     }
 
+    func updateHeader() {
+
+    }
+
```
emilylaguna: Empty implementation? Maybe add a comment so it doesn't look incomplete?
```suggestion
    // Empty implementation to adhere to the protocol
    func updateHeader() { }
```

emilylaguna (changes requested):
Looks good, and works well. Left a couple comments.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
leandroalonso: Done in 938c701c63e798c91727cbd53d16a5489328cc34

WordPress/WordPressTest/ReaderDetailCoordinatorTests.swift
leandroalonso: Done in 7d7047a7a2053583a31bf71411ac1b1a7e90e37a

emilylaguna (approved):
:shipit:

PR #14795, opened 2020-08-28
[Modal Layout Picker] Request scaled Thumbnail images

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2445

## To test:
<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
            
</details>

🗒️ More persistent caching of the thumbnail will be handled by a later ticket this currently uses an in-memory cache that will be purged by going to the background.
🗒️ If multiple image downloads complete at the same time the app can start to feel choppy. I plan to revisit this when we handle the long term cache of thumbnails

### WPCom Site
1. Select a WPcom site and ensure a valid network connection
1. Navigate to the Modal layout picker by either:
    - Selecting the FAB icon on the site page and selecting "Site page"
    - Selecting the ➕ on the Page List
1. Using a network logger:
    - **Expect** to see the `preview` parameter return with data like: 
        ```
        "preview": "https:\/\/headstartdata.files.wordpress.com\/2020\/01\/about-3.png?w=828&zoom=2"
        ```
    - **Expect** to see a network request with: 
        ```
        https://public-api.wordpress.com/wpcom/v2/sites/{{SITEID}}/block-layouts?locale=en&preview_width=828.0&scale=2.0
        ```

### Self-Hosted Site
1. Select a Self-Hosted site and ensure a valid network connection
1. Navigate to the Modal layout picker by either:
    - Selecting the FAB icon on the site page and selecting "Site page"
    - Selecting the ➕ on the Page List
1. Using a network logger:
    - **Expect** to see the `preview` parameter return with data like: 
        ```
        "preview": "https:\/\/headstartdata.files.wordpress.com\/2020\/01\/about-3.png?w=828&zoom=2"
        ```
    - **Expect** to see a network request with: 
        ```
        https://public-api.wordpress.com/wpcom/v2/common-block-layouts?locale=en&preview_width=828.0&scale=2.0
        ```

## Screenshots:
Screen Shot 2020-08-27 at 5 17 58 PM

## Additional Context: 

This is a continuation of the work done in https://github.com/wordpress-mobile/WordPress-iOS/pull/14726 and will be followed up by other tickets as part of [Modal Layout Picker Project](https://github.com/wordpress-mobile/gutenberg-mobile/issues?q=is%3Aissue+is%3Aopen+%5BModal+Layout+Picker%5D)

## PR submission checklist:


WordPress/Classes/Services/PageLayoutService.swift
```diff
@@ -47,4 +52,20 @@ class PageLayoutService {
...
+        ]
+    }
+
+    static func previewWidth() -> String {
+        return "\(UIScreen.main.nativeBounds.width)"
+    }
+
+    static func scale() -> String {
+        return "\(UIScreen.main.nativeScale)"
+    }
```
etoledom: Just curious here: Any particular reason to not make this vars instead of functions?

etoledom (commented):
Thank you for this PR @chipsnyder ! It's working great 🎉 

Just a couple of UI details:

The ghost effect is not noticeable on Dark Mode:
ghost


The vertical scroll feels a bit choppy with all the images in place (It's more noticeable in real life, not so much on the gif. Tested on iPhone 7):
scroll

**Beside that, I have one suggestion**:

`w=828&zoom=2` This image dimension looks like an overkill for the cell size. The ideal would be to get the cell width * screen density and make the image request with those parameters. The image will load faster and we will use (maybe around) 10% of the current data usage. Thinking that they are many images and their size is about 100~300kb each, this could be considerable and save many MBs of data.

chipsnyder:
Thanks @etoledom for the review! 

> The ghost effect is not noticeable on Dark Mode:

I should have called this out more clearly. We actually have this problem on all of the ghost views for this screen my goal for this one was to just get them in place then come through as part of the design ticket and handle all of the ghost views with a shared style.

> The vertical scroll feels a bit choppy with all the images in place (It's more noticeable in real life, not so much on the gif. Tested on iPhone 7):

I wasn't getting the choppy ness when scrolling on my 11 but that's probably a device-specific thing. I'll see if the smaller image sizes help with this. 

> This image dimension looks like an overkill for the cell size.

That's a great point and going back to the [API ticket](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2416) I realized I had misread the expected inputs. We were using the overall screen size as an example but it really wants the desired size in pixels. I'll get this fixed thanks for calling it out 👍

WordPress/Classes/Services/PageLayoutService.swift
chipsnyder: If I'm honest, no real reason that I remember. Looking at them with a fresh set of eyes I probably would have made them vars if I were to do it today. Also, I probably should have marked these as private 😅

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/LayoutPickerCollectionViewCell.swift
```diff
@@ -4,13 +4,8 @@ import Gridicons
...
 
-    static var cellReuseIdentifier: String {
-         return "LayoutPickerCollectionViewCell"
-     }
-
-    static var nib: UINib {
-        return UINib(nibName: "LayoutPickerCollectionViewCell", bundle: Bundle.main)
-    }
+    static let cellReuseIdentifier = "LayoutPickerCollectionViewCell"
+    static let nib = UINib(nibName: "LayoutPickerCollectionViewCell", bundle: Bundle.main)
```
chipsnyder, line 8: This is directly related to these changes but I noticed it was inconsistent with other files in this group so I went ahead and updated it here

chipsnyder:
@etoledom Trying this on my other devices I did notice that choppiness when scrolling vertically this time. I also noticed that was fetching the images for the row I was about to scroll to. So I think that falls under this category: 

> 🗒️ If multiple image downloads complete at the same time the app can start to feel choppy. I plan to revisit this when we handle the long term cache of thumbnails

I feel like this will be improved when we handle the caching of data as part of https://github.com/wordpress-mobile/gutenberg-mobile/issues/2456 so I'd like to tackle that then as long as that plan works for you. That said though it was smoother with the smaller images sizes.

I just pushed the changes requesting the smaller image sizes and made an adjustment from using a function to using a constant. So this is ready for review again

etoledom (commented):
Thank you @chipsnyder the update!

Testing the new Image size, I noticed that the URL returned by the API call is something like:
`https://headstartdata.files.wordpress.com/2020/01/balasana.png?w=320&zoom=2`

This serves an image width 640px of width. But the with of the cell is 160pts (with density of 2x in my case), so the served image width should be 320px.

Still it's a good improvement from the previous case, but if we could get the correct image width would be great.
I don't think this is a blocker on this stage so let me know if you prefer to move forward.

> I'd like to tackle that then as long as that plan works for you.

Sounds good to me! Let's come back to this later on 👍

chipsnyder:
> This serves an image width 640px of width. But the with of the cell is 160pts (with density of 2x in my case), so the served image width should be 320px.

@etoledom You're right I should have checked that before posting this up again. I was going off the actual size instead of the point size. Sorry for the wasted time on having you re-review it again. 😅 This has been updated now and should be returning images with 320pxs and sending up width=160 in the request.

etoledom (approved):
Thank you for the update @chipsnyder !

Working great now 🎉 
Also tested on the `Max` family and the zoom value corresponds to the screen density ✅ 

Choppiness is fixed now too! Probably my old phone had problems loading many heavy images at once (even if they were already cached).

Great job!

chipsnyder:
Thanks @etoledom!

> Choppiness is fixed now too! Probably my old phone had problems loading many heavy images at once (even if they were already cached).

Yeah, I think the larger images were getting flushed from the cache on less powerful devices, so I'm glad to hear it's smoothed out for you now!

Issue #14774, opened 2020-08-27
Big Titles, White Headers Project Issue


frosty:
This is the main project issue for "Big Titles, White Headers". The goal of this project is to switch the navigation bars to use system materials, serif fonts for titles, and large titles for key sections:

![big-titles-white-headers-ios](https://user-images.githubusercontent.com/4780/91452761-dffe6e80-e876-11ea-93fb-461972308f80.png)

- [x] Add feature flag (currently disabled) for the feature (#14797)
- [x] Don't apply current navigation bar background styling when feature flag is active (#14797)
- [x] Use system serif font for navigation bar titles (#14803)
- [ ] Use large title style for top level of My Site, Reader, and Notifications
- [ ] Integrate search bars with navigation bars (this may need to be split out into its own issue)
- [ ] When top tab bars or search bars are present, divider should appear below them
- [ ] Adopt new style site switcher (again, this will probably be a separate issue)
- [ ] Fix status bar color in theme browser and close this [issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15714) 
- [x] Ensure Media Picker has correct appearance
- [x] Ensure editor has correct appearance
- [ ] Ensure Theme picker webview nav bar and titles look as expected 
- [ ] Ensure All web views look as expected

PR #14797, opened 2020-08-28
Big Titles, White Headers 1

frosty:
Refs #14774.

This PR is the first part of "Big Titles, White Headers". It sets up a feature flag, and updates the colours of navigation bars throughout the app. This does not yet include large titles, integrated search bars, or serif fonts.

I've set the feature flag to a8cBranchTesting so that we can do an installable build here for @osullivanchris to look at. I'll set it back to localDeveloper before merging.

iOS 13

iOS 13 Dark Mode

iOS 12

Most of the changes here are actually to consolidate the style used for `preferredStatusBarStyle` throughout the app, so that the status bar appears the correct colour. I think we can probably even remove many of the remaining calls, as most of this should be controlled through the navigation controllers and split view controllers we use. But I think that can be cleanup for a future PR.

I also removed the `customizeAppearanceForTextElements` method from the app delegate – this was called on launch and every time the system's content size changed. I didn't see the use in calling this method multiple times, and I think that the things it configured were either unnecessary or also done elsewhere.

**To test**

- Check the code
- Build and run in:
  - iOS 12
  - iOS 13 light mode
  - iOS 13 dark mode
- Take a look through as many screens as you can in the app and check the navigation bar and its contents look correct. Look out for things like:
  - Bars created in libraries outside of the main app, like Authenticator, Media Picker, Gutenberg, and Zendesk (@ScoutHarris I tagged you here specifically for auth because I don't want to disrupt the work you've been doing there). Please check that unified and non-unified auth flows look correct.
  - Places we use search bars
  - Modal windows
  - System views like the document picker used if you add media from the 'other apps' source
- Double check things still look okay with the feature flag disabled

PR submission checklist:


frosty:
Just kicked off an installable build for you, @osullivanchris!

mattmiklic:
FOLLOWING. 🥳

frosty:
Build issues should be fixed now.

osullivanchris (approved):
Wow it looks brilliant for a first pass! Particularly the reader and editor look much better and more content focused. There's a few areas that we might think about, but they're not necessarily needing to be addressed here:

- **Search bars**. They feel a bit out of place in between the other blocks with white/black background colour. Wonder if we should move to integrated search bars, or customise the background colour, or something else
- **Background material** on headers - At the moment it looks like its solid white? I can't 100% tell! 
- **Dividers**...as discussed, there's some dividers there that I eliminated in mocks. Doesn't look bad though. Maybe the time to decide this is when we see how the large titles fit with everything else
- **Sticky/fixed elements**: same as discussions we're having on Android. In Android we've ended up not keeping things fixed, so that there's more content focus. Defaults may be different on iOS. Again I think we should wait until we see the large titles, as it could affect this decision
- **Dark mode / light mode**: a lot of this was already this way on Dark Mode, but it feels like there's more colours going on now compared to the new light mode. E.G. on reader in light mode there's only really a white background colour for content, and a grey background colour for the view. On dark mode, the tabs background colour and the content background colour are slightly different. Wondering if we should reduce the number of background colours we're using, as it looks really fresh in Light Mode just seeing only a couple.

Looking forward to seeing more!

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
```diff
@@ -79,9 +79,9 @@ class WordPressAuthenticationManager: NSObject {
                                                               textButtonColor: .primary,
                                                               textButtonHighlightColor: .primaryDark,
                                                               viewControllerBackgroundColor: .basicBackground,
-                                                              navBarBackgroundColor: .basicBackground,
-                                                              navButtonTextColor: .brand,
-                                                              navTitleTextColor: .text)
+                                                              navBarBackgroundColor: .basicBackground,  // TODO: Replace with .appBarBackground once new nav bar styles are merged
```
Gio2018: can we add a feature-specific keyword to these TODOs so they are easily searchable when it's time to remove them? Like TODO - BIGTITLES

Gio2018 (changes requested):
Hey @frosty ! This looks awesome!

I had a very small code comment, and then found a couple of things:

- This is minor: the 'Me' button white border now blends in with the bar in light mode, I'm wondering if we should do something different with it, since in dark mode it's still visible
- on iPad Air with iOS 12, the titles in "My Site" are still white and not visible

<p align=center>

</p>

ScoutHarris:
Hey @frosty .

Couple things I noticed on an iPad.

On iOS13:
- `Switch Site` color is white.
- Placeholder gravatar color is white.

| Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-08-31 at 14 17 18 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-08-31 at 14 14 33 |
|--------|-------|


On iOS12: all the text is white.

| Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-08-31 at 14 46 11 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-08-31 at 14 43 43 |
|--------|-------|

ScoutHarris (changes requested):
See comment about iPad.

frosty:
Thanks all!

@osullivanchris:

- @Gio2018 had a question about adding a non-white border to the Me icon – I think we probably should, as an all-or-mostly white gravatar would blend in with the nav bar. Any thoughts?
- **Search bars.** Definitely the plan to try integrating them – I haven't touched them yet. Coming up in a future PR!
- **Background material on headers.** Currently solid white. I thought this was what we had agreed, but we can try some other options after this initial PR.
- **Dividers, sticky / fixed elements** – I haven't touched either of these yet. I agree, we should wait to see large titles and then work out what to do from there.
- **Dark mode** I agree, I think dark mode needs another pass. It does feel less cohesive in terms of the colours we're using across the UI.

osullivanchris:
> @Gio2018 had a question about adding a non-white border to the Me icon – I think we probably should, as an all-or-mostly white gravatar would blend in with the nav bar. Any thoughts?

Oh yeah agree. On [Android we have added a grey border](https://github.com/wordpress-mobile/WordPress-Android/pull/12799#issuecomment-683481033) for both light and dark mode.

Are we using system grey colors yet on iOS? We could use something like [SystemGray5 from here](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/). Subtle and translates across light and dark mode.

> Background material on headers. Currently solid white. I thought this was what we had agreed, but we can try some other options after this initial PR.

Ok cool! Yeah I couldn't remember for sure. 

Rodger on the other details, happy to address in separate PRs!

Thanks @Gio2018 and @ScoutHarris for spotting the iPad tweaks!

mattmiklic:
> Are we using system grey colors yet on iOS? We could use something like SystemGray5 from here. Subtle and translates across light and dark mode.

Apple recommends using one of the [systemFill](https://developer.apple.com/documentation/uikit/uicolor/3255070-systemfill) colors for situations where the color will appear on a semitransparent background. Assuming that we eventually move to using the system material for these backgrounds, one of the systemFill colors (which are semitransparent) would likely be a better choice tha one of the solid systemGrays.

frosty:
@Gio2018 @ScoutHarris Thanks for your reviews! I've pushed some changes that should address your comments.

@osullivanchris @mattmiklic I just tried `systemFill` but for some reason I just don't see an outline. For now, I've eyeballed a colour that looks close to the one used in Android – I went for `systemGray2` on iOS 13+ and our gray20 on iOS 11 and 12. Here's how the system gray looks:

Screenshot 2020-09-01 at 23 55 00

And for fun, here's how it looks with a 1px border instead:

Screenshot 2020-09-01 at 23 56 17

Let me know what you think!

ScoutHarris (approved):
Looks good!

:shipit:

osullivanchris:
I'd go with the 1px @frosty ! 

I think it was tricky to make the avatar look right on blue, and the border was the best we could do. However, on white I think the avatar looks totally fine without a border. We're only really adding it it in for the case of a mostly white/black image in there. So I would keep it to the minimum 1px.

I haven't worked with the System Fills before either, so lets see if @mattmiklic has any info on why that colour wouldn't be showing up.

Gio2018 (approved):
Awesome @frosty :shipit:!

Gio2018:
> On Android we have added a grey border for both light and dark mode.

hey @osullivanchris , the grey border works well, although my personal impression is that the circle is yet another color in the nav bar, so was wondering if we should just get rid of the border and use that space for the gravatar itself, kinda like in the App Store? No big deal anyway, just a thought.

mattmiklic:
I'm not familiar with `systemFill` in code, only in the design docs. :) I think what @Gio2018 suggested above may be the best route though; the border was more necessary to prevent avatars from clashing with the blue BG; less of an issue on a neutral background.

osullivanchris:
Totally agree with that @mattmiklic @Gio2018 I think the design looks better without a border. 

But the issue that I was worried about was when there is a mostly white avatar (or dark grey in dark mode) the shape of the user icon is lost and it looks a bit funny / less obvious. So it was a case of bringing it in for all avatars just to solve this case. 

I'd love if we could show it _only_ when there was a mostly white avatar but that seems an unnecessarily complicated thing to do when we could just have a subtle border always! 😄 

avatars

mattmiklic:
I don't think it's that big of a deal either way, so I'm happy to defer to y'all. I looked at the App Store again; with a full color photo I didn't notice any border but when I used a mostly-white background avatar, I did see an extraordinarily faint and subtle border (maybe 0.5pt?) so I don't think it hurts to do this if y'all want to.

osullivanchris:
Extraordinarily subtle and faint works for me TBH!!

osullivanchris:
Had a quick look at how this works on the App Store. 

They are adding the subtle border actually on top of the image, rather than around the outside. It looks like a black @ 10% opacity.

So a white icon like Simplenote has a defined border. But the WordPress one you don't really see grey around it. They're doing the same thing for the App Store screenshots. I don't think they're doing as good a job in dark mode.

I think the approach is good. But we've definitely spent more time on this than the value we're getting! Looks like Frosty has already made a change to a subtle border, and if its on the outside rather than on the top i'm sure that's good enough!

borders app store

frosty:
@osullivanchris So for now, I've gone with a 0.5pt border in the system `separator` color, which is intended for thin dividing lines:

Screenshot 2020-09-03 at 09 46 18

I might just merge as-is for now to get this code wrapped up, and we can always revisit before we ship the feature.

osullivanchris:
> @osullivanchris So for now, I've gone with a 0.5pt border in the system separator color, which is intended for thin dividing lines:

this looks good!

> I might just merge as-is for now to get this code wrapped up, and we can always revisit before we ship the feature.

agreed! I'd have no problem shipping the above anyway

PR #14799, opened 2020-08-28
Add in-app support for remote feature flags

jkmassel:
Adds app-level support for remote feature flags.

Should be tested alongside https://github.com/wordpress-mobile/WordPressKit-iOS/pull/273

To test:
1. Apply this PR
2. In the `WordPress` scheme, go to Run > Arguments, and enable `-extra_debug 1`
3. Run the app, and filter logs by the 🚩 character.
4. You should see that the app does the following:
   - Generates a new Device ID (note it for later by copying it to the clipboard or something)
   - Updates the local cache of remote feature flags
5. Close the app
6. Run it again – this time it doesn't generate a new device ID – it uses the same one as before
7. If you'd like to test that the cache works, you can add this code at the top of the AppDelegate's `didFinishLaunching` method:

```swift
debugPrint("🚩 \(remoteFeatureFlagStore.valueForFlag(.unifiedAuth))")
```

It will print `true` to the console.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
+++ b/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
@@ -7,11 +7,9 @@ class RemoteFeatureFlagStore {
         DDLogInfo("🚩 Remote Feature Flag Device ID: \(deviceID)")
     }
 
-    /**
-     Fetches remote feature flags from the server.
-
-     - Parameter callback: An optional callback that can be used to update UI following the fetch. It is not called on the UI thread.
-    */
+    /// Fetches remote feature flags from the server.
+    /// - Parameters:
+    ///     - callback: An optional callback that can be used to update UI following the fetch. It is not called on the UI thread.
     public func update(then callback: FetchCallback? = nil) {
         // We currently use an anonymous connection to WordPress.com, because we don't need user state.
         // In the future, we could provide login credentials to the endpoint, which would allow customizing flags server-side on a per-user basis.
--- a/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
+++ b/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
@@ -27,7 +27,7 @@ class RemoteFeatureFlagStore {
     }
 
     /// Checks if the local cache has a value for a given `FeatureFlag`
-    public func hasRemoteValueForFlag(_ flag: FeatureFlag) -> Bool {
+    public func hasValue(for flag: FeatureFlag) -> Bool {
         guard let remoteKey = flag.remoteKey else {
             return false
         }
@@ -40,7 +40,7 @@ class RemoteFeatureFlagStore {
     /// If the flag exists in the local cache, the current value will be returned.  If the flag does not exist in the local cache, the compile-time default will be returned.
     /// - Parameters:
     ///     - flag: The `FeatureFlag` object associated with a remote feature flag
-    public func valueForFlag(_ flag: FeatureFlag) -> Bool {
+    public func value(for flag: FeatureFlag) -> Bool {
         guard
             let remoteKey = flag.remoteKey, // Not all flags need remote keys, since they may not use remote feature flagging
             let value = cache[remoteKey]    // The value may not be in the cache if this is the first run
--- a/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
+++ b/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
@@ -28,7 +28,11 @@ class RemoteFeatureFlagStore {
 
     /// Checks if the local cache has a value for a given `FeatureFlag`
     public func hasRemoteValueForFlag(_ flag: FeatureFlag) -> Bool {
-        return flag.remoteKey != nil && cache[flag.remoteKey!] != nil
+        guard let remoteKey = flag.remoteKey else {
+            return false
+        }
+
+        return cache[remoteKey] != nil
     }
 
     /// Looks up the value for a remote feature flag.
--- a/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
+++ b/WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
@@ -1,0 +1,93 @@
+import Foundation
+import WordPressKit
+
+class RemoteFeatureFlagStore {
+
+    init() {
+        DDLogInfo("🚩 Remote Feature Flag Device ID: \(deviceID)")
+    }
+
+    /**
+     Fetches remote feature flags from the server.
+
+     - Parameter callback: An optional callback that can be used to update UI following the fetch. It is not called on the UI thread.
+    */
+    public func update(then callback: FetchCallback? = nil) {
+        // We currently use an anonymous connection to WordPress.com, because we don't need user state.
+        // In the future, we could provide login credentials to the endpoint, which would allow customizing flags server-side on a per-user basis.
+        let remote = FeatureFlagRemote(wordPressComRestApi: WordPressComRestApi.defaultApi())
+        remote.getRemoteFeatureFlags(forDeviceId: deviceID) { [weak self] result in
+            switch result {
+                case .success(let flags):
+                    self?.cache = flags.dictionaryValue
+                    DDLogInfo("🚩 Successfully updated local feature flags: \(flags)")
+                    callback?()
+                case .failure(let error):
+                    DDLogError("🚩 Unable to update Feature Flag Store: \(error.localizedDescription)")
+            }
+        }
+    }
+
+    /// Checks if the local cache has a value for a given `FeatureFlag`
+    public func hasRemoteValueForFlag(_ flag: FeatureFlag) -> Bool {
+        return flag.remoteKey != nil && cache[flag.remoteKey!] != nil
+    }
+
+    /// Looks up the value for a remote feature flag.
+    ///
+    /// If the flag exists in the local cache, the current value will be returned.  If the flag does not exist in the local cache, the compile-time default will be returned.
+    /// - Parameters:
+    ///     - flag: The `FeatureFlag` object associated with a remote feature flag
+    public func valueForFlag(_ flag: FeatureFlag) -> Bool {
+        guard
+            let remoteKey = flag.remoteKey, // Not all flags need remote keys, since they may not use remote feature flagging
+            let value = cache[remoteKey]    // The value may not be in the cache if this is the first run
+            else {
+                DDLogInfo("🚩 Unable to resolve remote feature flag: \(flag.description). Returning compile-time default.")
+                return flag.enabled
+        }
+
+        return value
+    }
+
+    /// Thread Safety Coordinator
+    private let queue = DispatchQueue(label: "remote-feature-flag-store-queue")
+}
+
+extension RemoteFeatureFlagStore {
+    private struct Constants {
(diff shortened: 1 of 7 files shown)


mindgraffiti:
cc: @diegoreymendez in case you're curious

Edit: @jkmassel I'm assigning all 3 of us. We have a "1 review then merge it" rule. So once you get an approval, feel free to merge 😁

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
```diff
@@ -0,0 +1,93 @@
...
+                    DDLogInfo("🚩 Successfully updated local feature flags: \(flags)")
+                    callback?()
+                case .failure(let error):
+                    DDLogError("🚩 Unable to update Feature Flag Store: \(error.localizedDescription)")
+            }
+        }
+    }
+
+    /// Checks if the local cache has a value for a given `FeatureFlag`
+    public func hasRemoteValueForFlag(_ flag: FeatureFlag) -> Bool {
```
diegoreymendez: I'd consider going for a more Swifty:

```swift
public func hasRemoteValue(for flag: FeatureFlag) {
```

Which would result in calls like this one:

```swift
remoteFeatureFlagStore.hasRemoteValue(for: .newLoginFlow)
```

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
```diff
@@ -0,0 +1,93 @@
...
+        return flag.remoteKey != nil && cache[flag.remoteKey!] != nil
+    }
+
+    /// Looks up the value for a remote feature flag.
+    ///
+    /// If the flag exists in the local cache, the current value will be returned.  If the flag does not exist in the local cache, the compile-time default will be returned.
+    /// - Parameters:
+    ///     - flag: The `FeatureFlag` object associated with a remote feature flag
+    public func valueForFlag(_ flag: FeatureFlag) -> Bool {
```
diegoreymendez: Same comment as above:

```swift
public func value(for flag: FeatureFlag) -> Bool {
```

Which would result in calls like this one:

```swift
remoteFeatureFlagStore.value(for: .newLoginFlow)
```

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
```diff
@@ -0,0 +1,93 @@
...
+        guard
+            let remoteKey = flag.remoteKey, // Not all flags need remote keys, since they may not use remote feature flagging
```
diegoreymendez: IMHO this comment here should be part of the documentation in `FeatureFlag.remoteKey` instead of being present here, since that's intrinsic to `remoteKey` and not really to this particular guard.

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
```diff
@@ -0,0 +1,93 @@
...
+     Fetches remote feature flags from the server.
+
+     - Parameter callback: An optional callback that can be used to update UI following the fetch. It is not called on the UI thread.
+    */
+    public func update(then callback: FetchCallback? = nil) {
+        // We currently use an anonymous connection to WordPress.com, because we don't need user state.
+        // In the future, we could provide login credentials to the endpoint, which would allow customizing flags server-side on a per-user basis.
+        let remote = FeatureFlagRemote(wordPressComRestApi: WordPressComRestApi.defaultApi())
+        remote.getRemoteFeatureFlags(forDeviceId: deviceID) { [weak self] result in
+            switch result {
```
diegoreymendez, line 24: Love that we're using the `Result` type more!

diegoreymendez (changes requested):
The comments I left within the code are minor and pretty much subjective.  They're for your consideration.

The only thing I'd really add to this PR, which I feel is important, is unit tests.  The new class you created looks great, and it's written in a way that unit testing it is just a step away.  Even the concurrency working fine could be tested to make sure all is good.

But overall this is a very nice PR.

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    init() {
+        DDLogInfo("🚩 Remote Feature Flag Device ID: \(deviceID)")
+    }
+
+    /**
+     Fetches remote feature flags from the server.
+
+     - Parameter callback: An optional callback that can be used to update UI following the fetch. It is not called on the UI thread.
+    */
```
mindgraffiti: nitpick. Let's go more swift-y with [comments](https://github.com/raywenderlich/swift-style-guide/#comments) instead of C-based

```suggestion
    /// Fetches remote feature flags from the server.
    ///
    /// - Parameter callback: An optional callback that can be used to update UI following the fetch. It is not called on the UI thread.
```

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
```diff
@@ -0,0 +1,93 @@
...
+                    callback?()
+                case .failure(let error):
+                    DDLogError("🚩 Unable to update Feature Flag Store: \(error.localizedDescription)")
+            }
+        }
+    }
+
+    /// Checks if the local cache has a value for a given `FeatureFlag`
+    public func hasRemoteValueForFlag(_ flag: FeatureFlag) -> Bool {
+        return flag.remoteKey != nil && cache[flag.remoteKey!] != nil
```
mindgraffiti: Let's deal with optionals early instead of force unwrapping

```suggestion
        guard let cachedKey = flag.remoteKey else {
            return false
        }

        return flag.remoteKey != nil && cache[cachedKey] != nil
```

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
jkmassel: Addressed in e481fa72713c06912e97e7274baf60300992f72d. I didn't use the commit suggestion just because I wanted to make the line break and indentation consistent with the other documented method in this class.

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
jkmassel: Addressed in c2015f3f1a05446e730105371d6f9b9d339ea7ba. I didn't use the commit suggestion because if we're using a `guard` statement we don't need the `flag.remoteKey != nil` check again in the `return` statement, so might as well get rid of it 🤷‍♂️

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
jkmassel: Addressed in 720d35f707545fff468ef82df7c0990ce22d8efa. I also changed the method name slightly – I realized I wasn't happy with one being `hasRemoteValue` and the other being `getValue`, so now it's `hasValue` and `getValue` for consistency.

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
jkmassel: Same as above!

jkmassel:
@diegoreymendez – testing is included in 155f49229289527ec160f19492a03a6a72c5dc53. I had to rework `FeatureFlag` protocol conformance in order to make it possible to substitute a `MockFeaturedFlag` in `RemoteFeatureFlagStore` methods.

diegoreymendez:
> I had to rework FeatureFlag protocol conformance in order to make it possible to substitute a MockFeaturedFlag in RemoteFeatureFlagStore methods.

Apologies for the extra work.  Thanks a ton for taking the time to do this.

WordPress/Classes/Stores/RemoteFeatureFlagStore.swift
jkmassel: I ended up putting it in both places in dd84a9bd8702897382058007f94d4ace1c393c8b. 

I agree that it's important to have it in the protocol definition for sure, but this _should_ be the only place in the codebase that this value is examined, and it may not be immediately clear _why_ it needs to be unwrapped when reading the code, which is why I left the comment in there.

Hope that's ok?

WordPress/WordPressTest/FeatureFlagTests.swift
```diff
@@ -1,11 +1,29 @@
 import XCTest
 @testable import WordPress
 
-fileprivate enum MockFeatureFlag: OverrideableFlag {
+enum MockFeatureFlag: OverrideableFlag {
```
diegoreymendez, line 4: I believe that if you make this conform to `CaseIterable` you should be able to get all cases by using `MockFeatureFlag.allCases` (and not require the `remoteCases` array below).

Ref: https://www.simpleswiftguide.com/how-to-iterate-over-enum-cases-in-swift/

WordPress/WordPressTest/RemoteFeatureFlagTests.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    override func setUpWithError() throws {
+        UserDefaults.standard.removeObject(forKey: RemoteFeatureFlagStore.Constants.CachedFlagsKey)
+        UserDefaults.standard.removeObject(forKey: RemoteFeatureFlagStore.Constants.DeviceIdKey)
+    }
+
+    func testThatDeviceIdIsTheSameForEveryInstanceOfTheStore() {
+        let mock = MockFeatureFlagRemote()
+        var deviceId = ""
+        mock.deviceIdCallback = {
```
diegoreymendez, line 18: Nitpicky: I'd consider adding an expectation so that we know the `deviceIDCallback` is called twice.  The reason is that if the callback is never executed this test will never fail.

Something like:

```swift
func testThatDeviceIdIsTheSameForEveryInstanceOfTheStore() {
    let callbackExecuted = expectation(description: "Callback executed twice")
    callbackExecuted. expectedFulfillmentCount = 2

    let mock = MockFeatureFlagRemote()
    var deviceId = ""
    mock.deviceIdCallback = {
        deviceId == "" ? deviceId = $0 : XCTAssertEqual(deviceId, $0)
    }

    RemoteFeatureFlagStore(remote: mock).update()
    RemoteFeatureFlagStore(remote: mock).update()

    waitForExpectations(timeout: 0.1)
}
```

diegoreymendez (approved):
@jkmassel - Some optional comments, but I'm approving since this works as described.

Really nice work!

WordPress/WordPressTest/FeatureFlagTests.swift
jkmassel: Yeah! I don't actually want all the cases – just the remote ones (so not `enabledFeature`, `disabledFeature`, and `nonOverrideableFeature` – I thought about using `CaseIterable` and filtering out the ones I _didn't_ want but thought maybe that'd be too clever? 🤷‍♂️

This way is a bit more explicit, if verbose 😅

WordPress/WordPressTest/RemoteFeatureFlagTests.swift
jkmassel: This seems like a good extra check, so added in 7ba7d0946ec64e0e86576eb0cfc92854203cc07d :)

WordPress/WordPressTest/FeatureFlagTests.swift
diegoreymendez: It's fine I didn't realize it was some.

diegoreymendez (approved):
I just retested this with the new WPKit version and it's working as described.

PR #14801, opened 2020-08-29
Re-initialize statsServiceRemote in StatsPeriodStore when site changes

### LINKED ISSUE #14759: Stats not changing when you switch from one site to the other
We have had 3 reports lately regarding an issue with the iOS WP app. However, each of them happened in a different "scenario". This user 3252442-zen has simple WP.com site, this one 3254212-zen has multiple self-hosted JP sites, and this one 3238193-zen is a beta tester.

So, users open the app and it shows the stats for site # 1. All good there. They then switch to another site, let's call site # 2 and open the stats screen. 

### Expected behavior

We would expect the stats screen to show the stats pertaining to site # 2 once you switch. 

### Actual behavior

The stats keep showing the information from site # 1 

I was not able to reproduce here, so it's not 100% consistent, but @rezzap was able to reproduce on the beta version. 

- Open the app 
- Go to the stats page
- Click on Switch site and pick a different site 
- see the stats showing for the last site and not te current one. 

##### Tested on:

Tested by @rezzap on 15.5 iPhone X, 2 WordPress.com sites (one atomic, one Premium) iOS 13.6.1

User on 3238193-zen tested on  iPhone X iOS, 14.0 WordPress for iOS 15.5 (they also provided a screencast of the issue on the ticket).

The other 2 users I don't have the information on their devices or iOS version but both are on WordPress for iOS 15.4

Gio2018:
Fixes #14759

WordPressKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/274

To test:

pre-requisite: have an account with more than one site and some activity that populates the stats (in at least one of them)

- Select a site and go to stats
- Observe charts and values in any of the tabs (particularly the periods)
- Go back to My site, Switch site then select a different one
- Go to Stats and make sure that they are consistent with the site you chose, and are not from the previous site

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ Update to unit tests not required.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ Update to accessibility not required
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. Updated.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -37,9 +37,9 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.15.0-beta.2'
+    #pod 'WordPressKit', '~> 4.15.0-beta.2'
     #pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :tag => ''
-    #pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => ''
+    pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => 'issue/14759-stats-not-updating'
     #pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :commit => ''
     #pod 'WordPressKit', :path => '../WordPressKit-iOS'
 end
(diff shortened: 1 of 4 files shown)


frosty (approved):
Works for me! Thanks for fixing! Don't forget to ship the pod and update the Podfile reference before merging :)

ScoutHarris (approved):
Thanks for fixing this @Gio2018 ! Much appreciated!

PR #14803, opened 2020-08-31
White Bars II: Serif fonts

frosty:
Refs #14774.

This PR switches the navigation bars to use serif fonts. We use the system font on iOS 13+, and Noto on older versions. This needed some extra methods [added to WordPress-Shared-iOS here](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/275).

white-headers-ii

**To test**

- Build and run
- Test on iOS 12 and 13.
- Do a smoke test of the app and verify that you see serif fonts for the navigation bar title throughout. In particular, also check the blog selector in the editor, and the header in reader detail views.

PR submission checklist:


osullivanchris (approved):
Need to get my hands on the build, but judging from the screenshot this looks great. New York is definitely looking nicer than Noto to my eyes, but that's expected and I'm not sure we can do much to improve Noto.

WordPress/Classes/Services/PageLayoutService.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    private static func fetchSharedLayouts(completion: @escaping CompletionHandler) {
+        let api = WordPressComRestApi.anonymousApi(userAgent: WPUserAgent.wordPress())
+        let urlPath = "/wpcom/v2/common-block-layouts"
+        fetchLayouts(api, urlPath, completion)
+    }
+
+    private static func fetchLayouts(_ api: WordPressComRestApi, _ urlPath: String, _ completion: @escaping CompletionHandler) {
+        api.GET(urlPath, parameters: nil, success: { (responseObject, _) in
```
Gio2018: I think that, since this is using a `Decodable` object, it should use the new `GETData` that I added to `WordPressKit`. I am going to open that PR by today

WordPress/Classes/Services/GutenbergLayouts.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        return groupedLayouts
+    }
+
+    public func layouts(forCategory slug: String) -> [GutenbergLayout] {
+        return groupedLayouts[slug] ?? []
+    }
+}
+
+public struct GutenbergLayout: Codable {
```
Gio2018: maybe just `Decodable` would suffice? Unless we need to encode the same for any reason, of course

WordPress/Classes/Services/PageLayoutService.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    private static func fetchWordPressComLayouts(forBlog blog: Blog, completion: @escaping CompletionHandler) {
+        guard let blogId = blog.dotComID as? Int, let api = blog.wordPressComRestApi() else {
+            let error = NSError(domain: "PageLayoutService", code: 0, userInfo: [NSDebugDescriptionErrorKey: "Api or dotCom Site ID not found"])
+            completion(.failure(error))
+            return
+        }
+
+        let urlPath = "/wpcom/v2/sites/\(blogId)/block-layouts"
```
Gio2018: Do you think it would make sense to move the remote network calls to `WordPressKit`?

WordPress/Classes/Services/PageLayoutService.swift
```diff
@@ -0,0 +1,50 @@
...
+                return
+            }
+            completion(.success(result))
+        }, failure: { (error, _) in
+            completion(.failure(error))
+        })
+    }
+
+    private static func parseLayouts(fromResponse response: Any) -> GutenbergPageLayouts? {
+        guard let data = try? JSONSerialization.data(withJSONObject: response) else {
```
Gio2018: see above, so we don't have do decode-encode-decode

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -118,8 +118,23 @@ class GutenbergLayoutPickerViewController: UIViewController {
             })
         }
     }
+    private var isLoading: Bool = true {
```
Gio2018: Could be done in a separate PR, but I wonder if this could be handled using some sort of `enum State` with associated values, instead of a boolean (example: `ReaderTabItemsStore`, which also uses `WordPressFlux`). This way, we don't need to explicitly set the bool in every situation, and would have more control on the actual state. What do you think?

Gio2018 (changes requested):
Hey @frosty ! I had a few code comments, but other than that saw the same issue of  #14797 on iPad iOS 12

frosty:
Sorry @Gio2018, you shouldn't have had to review all that Gutenberg code here. I think I accidentally pulled develop into this branch instead of `feature/white-bars-i`, putting it ahead of the base branch. I'll clean it up now.

frosty:
@Gio2018 I cleaned up the commits in this PR. Once the changes in #14797 are approved, is this one good to go too?

Gio2018 (approved):
> @Gio2018 I cleaned up the commits in this PR. Once the changes in #14797 are approved, is this one good to go too?

Hey @frosty  yes! :shipit: !

PR #14810, opened 2020-09-01
Match Android's Reader events

leandroalonso:
This PR adds some missing events introduced [in Android here](https://github.com/wordpress-mobile/WordPress-Android/pull/12818).

### Following

1. Tap Reader
2. Tap Following
3. Check that `reader_following_shown` is fired

### Saved

1. Tap Reader
2. Tap Saved
3. Check that `reader_saved_list_shown` is fired

### Liked

1. Tap Reader
2. Tap Liked
3. Check that `reader_liked_shown` is fired

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -54,6 +54,7 @@ import Foundation
     case selectInterestsPicked
     case readerDiscoverShown
     case readerFollowingShown
+    case readerLikedShown
     case readerBlogPreviewed
     case readerDiscoverPaginated
     case readerPostCardTapped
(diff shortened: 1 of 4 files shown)


yaelirub:
Hey @leandroalonso   👋 
A couple of questions:

1. This is probably not related to this PR but why am I seeing this:
Screen Shot 2020-09-02 at 2 35 38 PM

2. For Saved I see `reader_saved_list_viewed` not `reader_saved_list_shown` as described is that intentional?

leandroalonso:
@yaelirub 

1. This is just a log that I added when a post is not deleted because it's associated with a card (this avoid a card being there without any post). But maybe it would create more problems and we don't really need it. I can remove, wdyt?

2. Fixed in af8e506137f881f273d7350ba76fd996a64db5d2

yaelirub:
>But maybe it would create more problems and we don't really need it

Yeah I think we can remove this one.

yaelirub (approved):
👍

PR #14820, opened 2020-09-02
UL&S Tracking: tracks dismissal of the Help screen. 

diegoreymendez:
Adds missing tracking of the dismissal of the help dialogue during the sign in flows.

### Test 1: Tracking works

In your console in Xcode, filter by "Tracked: unified".

1. Pic one of the enabled flows (Site Address, Google, SIWA) to Sign In.
2. Once you see the Help button in the top right corner of any screen tap it.
3. When the help screen is up, tap "Close".

Make sure you see this event in Xcode's console:

🔵 Tracked: unified_login_interaction <click: dismiss, flow: XXXX, source: default, step: help>

### Test 2: no tracking in outside of Sign In

1. Once logged in go to "Me"
2. Tap "Help & Support"
3. Tap close.

Make sure you don't see the unified track event reported above.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
+++ b/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
@@ -177,6 +177,10 @@ extension WordPressAuthenticationManager: WordPressAuthenticatorDelegate {
     var showSupportNotificationIndicator: Bool {
         return ZendeskUtils.showSupportNotificationIndicator
     }
+    
+    private var tracker: AuthenticatorAnalyticsTracker {
+        AuthenticatorAnalyticsTracker.shared
+    }
 
     /// We allow to connect with WordPress.com account only if there is no default account connected already.
     var allowWPComLogin: Bool {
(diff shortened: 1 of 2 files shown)


diegoreymendez:
Hey folks, sorry I spotted something more that needs changing.  I'm moving this back to draft and submitting in a few.  Apologies!

diegoreymendez:
Re-checked and this is ready for review.

diegoreymendez:
@jkmassel - This also goes into 15.6.

mindgraffiti (changes requested):
I don't think it's correct to fire both the old Tracks and new one. When I tap the "Help" button, it reports both:
```
🔵 Tracked: unified_login_step <flow: login_site_address, source: default, step: help>
2020-09-03 08:51:32:656 WordPress[9945:2620759] 🔵 Tracked: support_opened <>
```

diegoreymendez:
> I don't think it's correct to fire both the old Tracks and new one. When I tap the "Help" button, it reports both:

This PR doesn’t change that event, so I’d prefer if we could move forward with this PR and handle the reported issue in a separate PR.

But it's not just the scope: part of the reason why I’d like to handle that request separately is because I'm not sure it will be an easy change.  The support screen is shown from several places within the App.

mindgraffiti (approved):
Unblocking this PR because I don't think the Help screens are the most important parts of the UL&S tracks

diegoreymendez:
Thanks @mindgraffiti

diegoreymendez:
@jkmassel - First one is merged.

PR #14821, opened 2020-09-02
Reader Improvements: Detail Update: Move the featured image to the top of the view

### LINKED ISSUE #14768: Reader Improvements: Phase 3 - Project Issue
This is the main project issue for Phase 3 of the Reader Improvements Project:

### Deferred
- [x] Discover: Sorting Posts: https://github.com/wordpress-mobile/WordPress-iOS/issues/14765

### Done
- [x] Sites/Blogs you might like: https://github.com/wordpress-mobile/WordPress-iOS/issues/14766
- [x] Remove Feature Flag: https://github.com/wordpress-mobile/WordPress-iOS/pull/14842
- [x] Change copy from "tags" to "topics" on the manage screen: https://github.com/wordpress-mobile/WordPress-iOS/issues/14837
- [x] Change copy from "tags" to "topics" on Add Tag screen: https://github.com/wordpress-mobile/WordPress-iOS/issues/14838
- [x] Request bigger size for site icon in Reader Site Header: [PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15045) - [Issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/14856)
- [x] Reader Detail Updates: https://github.com/wordpress-mobile/WordPress-iOS/issues/14767
- [x] Reader Detail Design Review: https://github.com/wordpress-mobile/WordPress-iOS/issues/15052
- [x] Move settings icon from following tab to reader app nav bar: https://github.com/wordpress-mobile/WordPress-iOS/issues/14835

emilylaguna:
Project: #14767

### Dark Mode
| Full Height | Scrolled |
|:---:|:---:|
|||

### Full Height: Portrait

| iPhone 5s - 11.4 | iPad Air - 12.4 | iPhone 11 Pro - 13.3 |
|:---:|:---:|:---:|
||||

### Full Height: Landscape

| iPhone 5s - 11.4 | iPad Air - 12.4 | iPhone 11 Pro - 13.3 |
|:---:|:---:|:---:|
||||

### Scrolled
| iPhone 5s - 11.4 | iPad Air - 12.4 | iPhone 11 Pro - 13.3 |
|:---:|:---:|:---:|
||||

## To test:
1. Launch the app
2. Tap on the Reader Tab
3. Tap on any card item

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
@@ -158,7 +158,10 @@ class ReaderDetailCoordinator {
     /// Open the postURL in a separated view controller
     ///
     func openInBrowser() {
-        guard let postURL = postURL else {
+        guard
+            let permaLink = post?.permaLink,
+            let postURL = URL(string: permaLink)
+        else {
             return
         }
 
@@ -430,6 +433,7 @@ class ReaderDetailCoordinator {
     }
 }
 
+// MARK: - ReaderDetailHeaderViewDelegate
 extension ReaderDetailCoordinator: ReaderDetailHeaderViewDelegate {
     func didTapBlogName() {
         previewSite()
@@ -447,12 +451,15 @@ extension ReaderDetailCoordinator: ReaderDetailHeaderViewDelegate {
         previewSite()
     }
 
-    func didTapFeaturedImage(_ sender: CachedAnimatedImageView) {
-        showFeaturedImage(sender)
-    }
-
     func didTapFollowButton() {
         followSite()
+    }
+}
+
+// MARK: - ReaderDetailFeaturedImageViewDelegate
+extension ReaderDetailCoordinator: ReaderDetailFeaturedImageViewDelegate {
+    func didTapFeaturedImage(_ sender: CachedAnimatedImageView) {
+        showFeaturedImage(sender)
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -134,7 +134,7 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
         guard !featuredImage.isLoaded else {
             return
         }
-        
+
         // Load the image
         featuredImage.load { [unowned self] in
             self.hideLoading()
@@ -293,7 +293,7 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
             UIBarButtonItem.fixedSpace(24),
             configuredBrowserButton()
         ]
-        
+
         navigationItem.leftBarButtonItem = configuredBackButton()
         navigationItem.rightBarButtonItems = rightItems.compactMap({ $0 })
     }
@@ -366,7 +366,7 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
             featuredImage.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: 0),
             featuredImage.topAnchor.constraint(equalTo: view.topAnchor, constant: 0)
         ])
-        
+
         headerContainerView.translatesAutoresizingMaskIntoConstraints = false
     }
 
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -147,6 +147,14 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
         ReaderTracker.shared.stop(.readerPost)
     }
 
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+
+        coordinator.animate(alongsideTransition: { _ in
+            self.featuredImage.deviceDidRotate()
+        })
+    }
+
     func render(_ post: ReaderPost) {
         configureDiscoverAttribution(post)
 
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -143,7 +143,6 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
 
         // Reapply the appearance, this reset the navbar after presenting a view
         featuredImage.applyTransparentNavigationBarAppearance(to: navigationController?.navigationBar)
-
     }
 
     override func viewDidDisappear(_ animated: Bool) {
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -113,8 +113,21 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+        configureFeaturedImage()
+
+        featuredImage.configure(scrollView: scrollView,
+                                navigationBar: navigationController?.navigationBar)
 
         featuredImage.applyTransparentNavigationBarAppearance(to: navigationController?.navigationBar)
+
+        guard !featuredImage.isLoaded else {
+            return
+        }
+
+        // Load the image
+        featuredImage.load { [unowned self] in
+            self.hideLoading()
+        }
     }
 
     override func viewWillDisappear(_ animated: Bool) {
@@ -131,14 +144,6 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
         // Reapply the appearance, this reset the navbar after presenting a view
         featuredImage.applyTransparentNavigationBarAppearance(to: navigationController?.navigationBar)
 
-        guard !featuredImage.isLoaded else {
-            return
-        }
-
-        // Load the image
-        featuredImage.load { [unowned self] in
-            self.hideLoading()
-        }
     }
 
     override func viewDidDisappear(_ animated: Bool) {
@@ -158,15 +163,21 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
     func render(_ post: ReaderPost) {
         configureDiscoverAttribution(post)
 
-        featuredImage.configure(scrollView: scrollView,
-                                navigationBar: navigationController?.navigationBar)
-
         featuredImage.configure(for: post)
         toolbar.configure(for: post, in: self)
         header.configure(for: post)
 
         coordinator?.storeAuthenticationCookies(in: webView) { [weak self] in
             self?.webView.loadHTMLString(post.contentForDisplay())
+        }
+
+        guard !featuredImage.isLoaded else {
+            return
+        }
+
+        // Load the image
+        featuredImage.load { [unowned self] in
+            self.hideLoading()
         }
     }
 
@@ -189,8 +200,8 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
             self.loadingView.alpha = 0.0
         }) { (_) in
             self.loadingView.isHidden = true
-            self.loadingView.alpha = 1.0
             self.loadingView.stopGhostAnimation()
+            self.loadingView.alpha = 1.0
         }
     }
 
@@ -357,6 +368,10 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
     }
 
     private func configureFeaturedImage() {
+        guard featuredImage.superview == nil else {
+            return
+        }
+
         featuredImage.delegate = coordinator
 
         view.insertSubview(featuredImage, belowSubview: loadingView)
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -294,78 +294,6 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
         }
     }
 
-    private func configureNavigationBar() {
-        let rightItems = [
-            UIBarButtonItem.fixedSpace(24),
-            configuredMoreButton(),
-            UIBarButtonItem.fixedSpace(24),
-            configuredShareButton(),
-            UIBarButtonItem.fixedSpace(24),
-            configuredBrowserButton()
-        ]
-
-        navigationItem.leftBarButtonItem = configuredBackButton()
-        navigationItem.rightBarButtonItems = rightItems.compactMap({ $0 })
-    }
-
-    private func configuredBackButton() -> UIBarButtonItem {
-        let image = UIImage.gridicon(.chevronLeft).withRenderingMode(UIImage.RenderingMode.alwaysTemplate)
-        let button = CustomHighlightButton(frame: CGRect(x: 0, y: 0, width: image.size.width, height: image.size.height))
-        button.setImage(image, for: UIControl.State())
-        button.addTarget(self, action: #selector(didTapBackButton(_:)), for: .touchUpInside)
-
-        let barButtonItem = UIBarButtonItem(customView: button)
-        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
-
-        return barButtonItem
-    }
-
-    @objc func didTapBackButton(_ sender: UIButton) {
-        navigationController?.popViewController(animated: true)
-    }
-
-    private func configuredBrowserButton() -> UIBarButtonItem? {
-        let image = UIImage.gridicon(.globe).withRenderingMode(UIImage.RenderingMode.alwaysTemplate)
-        let button = CustomHighlightButton(frame: CGRect(x: 0, y: 0, width: image.size.width, height: image.size.height))
-        button.setImage(image, for: UIControl.State())
-        button.addTarget(self, action: #selector(didTapBrowserButton(_:)), for: .touchUpInside)
-
-        let barButtonItem = UIBarButtonItem(customView: button)
-        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
-
-        return barButtonItem
-    }
-
-    private func configuredMoreButton() -> UIBarButtonItem? {
-        guard let icon = UIImage(named: "icon-menu-vertical-ellipsis") else {
-            return nil
-        }
-
-        let image = icon.withRenderingMode(.alwaysTemplate)
-        let button = CustomHighlightButton(frame: CGRect(x: 0, y: 0, width: image.size.width, height: image.size.height))
-        button.setImage(image, for: UIControl.State())
-        button.addTarget(self, action: #selector(didTapMenuButton(_:)), for: .touchUpInside)
-
-        let barButtonItem = UIBarButtonItem(customView: button)
-        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
-
-        return barButtonItem
-    }
-
-    /// REtrurn the share button at the right of the nav bar
-    ///
-    private func configuredShareButton() -> UIBarButtonItem? {
-        let image = UIImage.gridicon(.shareiOS).withRenderingMode(UIImage.RenderingMode.alwaysTemplate)
-        let button = CustomHighlightButton(frame: CGRect(x: 0, y: 0, width: image.size.width, height: image.size.height))
-        button.setImage(image, for: UIControl.State())
-        button.addTarget(self, action: #selector(didTapShareButton(_:)), for: .touchUpInside)
-
-        let barButtonItem = UIBarButtonItem(customView: button)
-        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
-
-        return barButtonItem
-    }
-
     private func configureFeaturedImage() {
         guard featuredImage.superview == nil else {
             return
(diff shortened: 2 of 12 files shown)


WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -103,10 +111,38 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+
+        featuredImage.applyTransparentNavigationBarAppearance(to: navigationController?.navigationBar)
+
+        guard !featuredImage.isLoaded else {
+            return
+        }
+
+        // Load the image
+        featuredImage.load { [unowned self] in
+            self.hideLoading()
```
leandroalonso, line 129: I have a big concern about that here: waiting for the featured image to load to display the content might be ok in good connections, but almost impracticable in 3G for example.

I tested a few articles here (in iPad and iPhone using 3G) and for some articles it took so long that I give up (loading a GIF in that case). You can see my [explorations here](https://d.pr/v/AAQimY).

I know changing that probably opens a lot of other issues. But I really think the most important thing is the content, and once available we should show it right away. We can add a grey placeholder for the featured image once it's being loaded.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
```diff
@@ -447,15 +451,18 @@ extension ReaderDetailCoordinator: ReaderDetailHeaderViewDelegate {
...
-
     func didTapFollowButton() {
         followSite()
     }
 }
 
+// MARK: - ReaderDetailFeaturedImageViewDelegate
+extension ReaderDetailCoordinator: ReaderDetailFeaturedImageViewDelegate {
+    func didTapFeaturedImage(_ sender: CachedAnimatedImageView) {
+        showFeaturedImage(sender)
```
leandroalonso, line 462: When I tap an image and then go back, I can see the nav bar blue and then this disappears. Is this something that you're aware of?

We can address that later, anyway.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -103,10 +111,38 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+        featuredImage.restoreNavigationBarAppearance()
+    }
+
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
         ReaderTracker.shared.start(.readerPost)
+
+        // Reapply the appearance, this reset the navbar after presenting a view
+        featuredImage.applyTransparentNavigationBarAppearance(to: navigationController?.navigationBar)
```
leandroalonso, line 145: A few concerns regarding the transparent navigation bar:

* While I think it's beautiful, it's kind clashes with the rest of the app (AFAIK we don't have that anywhere else)
* The above is not a big deal, but I find a little bit weird that the navigation is translucid but the toolbar isn't.
* I can see that when we scroll the featured image gets behind it and stays there. I'd expect to see the scrolling content behind it
* In light mode when you scroll the status bar is white (while the icons are black). Shouldn't the status bar change to black too?

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -115,14 +151,33 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
         ReaderTracker.shared.stop(.readerPost)
     }
 
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+
+        coordinator.animate(alongsideTransition: { _ in
+            self.featuredImage.deviceDidRotate()
```
leandroalonso, line 158: When I rotate the device the scroll goes back to the top. In the previous implementation, this wouldn't change the offset.

I know there's probably a good reason for that but I think this is really bad. If I'm reading an article and I rotate my device by mistake I'd have to scroll it all over again.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -233,17 +294,94 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+        button.addTarget(self, action: #selector(didTapBrowserButton(_:)), for: .touchUpInside)
+
+        let barButtonItem = UIBarButtonItem(customView: button)
+        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
+
+        return barButtonItem
+    }
+
+    private func configuredMoreButton() -> UIBarButtonItem? {
+        guard let icon = UIImage(named: "icon-menu-vertical-ellipsis") else {
```
leandroalonso: It's almost impossible to tap this button, the width should be at least 44pt:

Screen Shot 2020-09-03 at 11 25 30

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -233,17 +294,94 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+            configuredShareButton(),
+            UIBarButtonItem.fixedSpace(24),
+            configuredBrowserButton()
+        ]
+
+        navigationItem.leftBarButtonItem = configuredBackButton()
+        navigationItem.rightBarButtonItems = rightItems.compactMap({ $0 })
+    }
+
+    private func configuredBackButton() -> UIBarButtonItem {
```
leandroalonso:
```suggestion
    private func configureBackButton() -> UIBarButtonItem {
```

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -233,17 +294,94 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
+
+        return barButtonItem
+    }
+
+    @objc func didTapBackButton(_ sender: UIButton) {
+        navigationController?.popViewController(animated: true)
+    }
+
+    private func configuredBrowserButton() -> UIBarButtonItem? {
```
leandroalonso:
```suggestion
    private func configureBrowserButton() -> UIBarButtonItem? {
```

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -233,17 +294,94 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+        button.setImage(image, for: UIControl.State())
+        button.addTarget(self, action: #selector(didTapBrowserButton(_:)), for: .touchUpInside)
+
+        let barButtonItem = UIBarButtonItem(customView: button)
+        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
+
+        return barButtonItem
+    }
+
+    private func configuredMoreButton() -> UIBarButtonItem? {
```
leandroalonso:
```suggestion
    private func configureMoreButton() -> UIBarButtonItem? {
```

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -233,17 +294,94 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
+    }
+
+    private func configuredBackButton() -> UIBarButtonItem {
+        let image = UIImage.gridicon(.chevronLeft).withRenderingMode(UIImage.RenderingMode.alwaysTemplate)
+        let button = CustomHighlightButton(frame: CGRect(x: 0, y: 0, width: image.size.width, height: image.size.height))
+        button.setImage(image, for: UIControl.State())
+        button.addTarget(self, action: #selector(didTapBackButton(_:)), for: .touchUpInside)
+
+        let barButtonItem = UIBarButtonItem(customView: button)
+        barButtonItem.accessibilityLabel = NSLocalizedString("Share", comment: "Spoken accessibility label")
```
leandroalonso: Maybe we should extract this constant to an enum or something like that? Up to you.

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
```diff
@@ -0,0 +1,386 @@
...
+        struct multipliers {
+            static let maxPortaitHeight: CGFloat = 0.70
+            static let maxPadPortaitHeight: CGFloat = 0.50
+            static let maxLandscapeHeight: CGFloat = 0.30
+        }
+    }
+
+    struct Styles {
+        static let startTintColor: UIColor = .white
+        static let endTintColor: UIColor = UIColor(light: .black, dark: .white)
```
leandroalonso: I guess you can replace that by using `.text`

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
```diff
@@ -0,0 +1,386 @@
...
+        // Allow for large GIFs to animate on the detail page
+        return ImageLoader(imageView: imageView, gifStrategy: .largeGIFs)
+    }()
+
+    /// The reader post that the toolbar interacts with
+    private var post: ReaderPost?
+    private weak var scrollView: UIScrollView?
+    private weak var navigationBar: UINavigationBar?
+
+    /// An observer of the number of likes of the post
```
leandroalonso: Wrong comment

leandroalonso (changes requested):

mbshakti:
>While I think it's beautiful, it's kind clashes with the rest of the app (AFAIK we don't have that anywhere else)

@leandroalonso regarding this comment, I noticed (also confirmed with @iamthomasbishop) that the new layout picker on mobile is going to use translucent materials, and have also heard from chris that @mattmiklic is interested in exploring it more.

TBH I really like the translucent bar here! However, it does sometimes feel distracting from the text if the image is too loud/colorful. Would we be able to use the translucent material just for what's underneath instead of using the image as the base?

emilylaguna:
@emilylaguna I tried that before and I didn't think it looked good. The animation between the image disappearing and the text behind the effect view looks jarring.

For now, I'll remove the effect view and revert to the original designs. We can explore more options later on.

mbshakti:
Stoked about this! The double image display situation throws me off a bit. I looked in how the current experience works, and it doesn't happen there. Do you know how we can prevent this with the new design? 

IMG_0474

emilylaguna:
> Stoked about this! The double image display situation throws me off a bit. I looked in how the current experience works, and it doesn't happen there. Do you know how we can prevent this with the new design?

Unfortunately not. The image itself is within the post content. I looked into hiding it within the content itself but there are TOO many issues that can happen with that.

mbshakti:
> Unfortunately not. The image itself is within the post content. I looked into hiding it within the content itself but there are TOO many issues that can happen with that.

Just to clarify, those kind of posts don't have a featured image, correct? That's why we are seeing this issue? Or are they just using the same image for the featured post + the beginning of the post?

emilylaguna:
> > Unfortunately not. The image itself is within the post content. I looked into hiding it within the content itself but there are TOO many issues that can happen with that.
> 
> Just to clarify, those kind of posts don't have a featured image, correct? That's why we are seeing this issue? Or are they just using the same image for the featured post + the beginning of the post?

Correct, it is if they include an image in their post and don't set a featured image.

Edit: 
We talked about this a bit more, and agreed to not show the top featured image if it's included in the content. We're also thinking about other solutions to this.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: I agree, I have a plan to address this but I'll do it in a separate PR.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Fixed in: 8479f0d1ec8913acb4007319d850a95b9f115828

It's 33px because I wanted to keep the visual spacing from the share button to 24px

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/commit/8479f0d1ec8913acb4007319d850a95b9f115828

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/commit/8479f0d1ec8913acb4007319d850a95b9f115828

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/commit/8479f0d1ec8913acb4007319d850a95b9f115828

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/commit/8479f0d1ec8913acb4007319d850a95b9f115828

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
emilylaguna: Done in: 51261fdb999244caa1d9e1148af2bca5ae5137a5

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
emilylaguna: Done in: a244c4c5ba05725b1b0ec935853c924e045c7673

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
emilylaguna: yeah ☹️ I spent a few hours working on this but could not come up with a solution. For some reason the appearance is reset when the image view is opened, and doesn't reset itself in viewWillAppear, so I have to reset the appearance again in viewDidAppear.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: The statusbar fix is done here: dc06d075787b37684b84e79984fb247bb3a68999

emilylaguna:
@leandroalonso Back to you!

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
leandroalonso: One small issue:



Shouldn't the status bar follow the same color as the icons?

emilylaguna: It should! It was working perfectly last night, let me take another look.

emilylaguna: 🤔 I'm not able to reproduce this on my end: https://d.pr/i/VG6nPi

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: I made a separate task so we can address this.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Ah I reproduced it on device. My fix is a bit involved so I'm going to make a new PR for it.

iamthomasbishop:
> TBH I really like the translucent bar here! However, it does sometimes feel distracting from the text if the image is too loud/colorful. Would we be able to use the translucent material just for what's underneath instead of using the image as the base?

@mbshakti @emilylaguna I'm not exactly sure if it would solve what you're referring to above, but in general if you're worried about contents from a lower plane shining through too much, it might be worth trying a thicker system material (assuming you're currently using the regular/chrome material).

emilylaguna:
> @mbshakti @emilylaguna I'm not exactly sure if it would solve what you're referring to above, but in general if you're worried about contents from a lower plane shining through too much, it might be worth trying a thicker system material (assuming you're currently using the regular/chrome material).

The main issue isn't so much the content shining through, it's the initial transition where the effect view fades in and the title goes under it.. 

Give me a few mins and I'll post a video demonstrating this.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
leandroalonso: Will be fixed in another PR.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
leandroalonso: Will be fixed in another PR.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
leandroalonso: Will be fixed in another PR.

leandroalonso (approved):
@emilylaguna A W E S O M E  job @emilylaguna!

We still have some issues to fix but as we talked we'll work on them on separate PRs.

emilylaguna:
@mbshakti @iamthomasbishop Alright took a bit longer than I thought but here's a demo of what it looks like:
https://d.pr/v/HomEzN

Honestly, after I fixed some bugs with the visuals I think this looks pretty good. Would love your thoughts!

iamthomasbishop:
@emilylaguna It's obviously @mbshakti 's call, but personally I think this looks fantastic overall! If you don't mind, I have a couple of bits of feedback (take with a grain of salt 😄):

* I think the navigation bar in light mode would benefit from a separator/border at the bottom for a tiny bit of separation, but I'm not sure if/how that aligns with other places that we're implementing new nav/header bars
* To my eyes, the icons in the navigation bar seem a bit stronger than we're using elsewhere in the apps (particularly in light mode)? If memory serves, I think we're using `systemGray` as the default icon color, although to be fair we might need something a bit more dynamic considering they're over translucent materials

As an aside, cc'ing @osullivanchris to make sure we're aligning on logic/details as other works-in-progress. 👋

emilylaguna:
> * I think the navigation bar in light mode would benefit from a separator/border at the bottom for a tiny bit of separation, but I'm not sure if/how that aligns with other places that we're implementing new nav/header bars

I agree, I did a quick test with fading in a shadow image that uses the dynamic `separator` color, you can see how it looks here: https://d.pr/v/VL7jNg

> To my eyes, the icons in the navigation bar seem a bit stronger than we're using elsewhere in the apps (particularly in light mode)? If memory serves, I think we're using systemGray as the default icon color, although to be fair we might need something a bit more dynamic considering they're over translucent materials

I'm using `label` for the color which is a iOS 13 dynamic color. Using vibrant colors for the icons could be an idea but it I'm not sure how hard it will be to perform the transition effect when the image disappears.

iamthomasbishop:
> I agree, I did a quick test with fading in a shadow image that uses the dynamic separator color, you can see how it looks here: https://d.pr/v/VL7jNg

That looks great! And now I find myself wanting the materials on the footer toolbar as well 😬 😆   [homer-disappear.gif]

Not 100% sure if we'd need a separator on dark mode, but I will let others chime in on how they think this feels.

> I'm using label for the color which is a iOS 13 dynamic color. Using vibrant colors for the icons could be an idea but it I'm not sure how hard it will be to perform the transition effect when the image disappears.

Ahh, okay. I think `secondaryLabel` would be a solid dynamic "equivalent" to `systemGray` in this context ([example](https://cloudup.com/cZRolLnn_BV)), maybe that's worth a shot.

osullivanchris:
> As an aside, cc'ing @osullivanchris to make sure we're aligning on logic/details as other works-in-progress. 👋

Thanks Thomas! We aren't as far yet on iOS as we are on Android on the new headers. But on iOS I think we get a separator as a default behaviour when we move to using a system white header rather than our blue one. And it looks good where I've seen it. So its likely we'll have that.

While I'm here...Personally I feel like the image blurred in the header looks cool in dark mode, but I don't like it as much in light mode. I'd probably vote for just blurring the content behind the nav bar rather than keeping the image in there regardless of where you scroll. But there is something cool about it, maybe there's a way to make it work.

> We talked about this a bit more, and agreed to not show the top featured image if it's included in the content. We're also thinking about other solutions to this.

Hey @mbshakti on this one - I can't remember if you plan to suggest to people to add a featured image in pre-publishing nudges. If you do, it might reduce the number of times this is happening. Maybe people add images to the top of the body intending it to be a sort of featured image, without knowing that exists. @iamthomasbishop you could even go one step further in the editor and when someone adds an image at the very top of the post, upsell them to a featured image 😬 😱 ?

mbshakti:
> can't remember if you plan to suggest to people to add a featured image in pre-publishing nudges

@osullivanchris for v2, featured image is part of the plan


> Maybe people add images to the top of the body intending it to be a sort of featured image, without knowing that exists.

Highly possible with new users

>you could even go one step further in the editor and when someone adds an image at the very top of the post, upsell them to a featured image

I do wish there was a way to do this in the editor. On Reader, it can get awkward if a user uploads an image onto the post, and then uploads the same image as the featured image. It could appear twice... which doesn't look great.

PR #14828, opened 2020-09-03
Debug menu: Add option to clear CSS cache

frosty:
This PR adds an option to the debug menu to clear the cache timestamp for CSS in the Reader.

Screenshot 2020-09-03 at 10 11 50

**To test:**

- Add breakpoints to the `address` property in `ReaderCSS.swift` where each possible URL is returned.
- Visit a couple of pages in the Reader and check that the same URL is returned each time.
- Head to **Me > App Settings > Debug** and tap **Clear Reader CSS Cache**.
- Go back to the Reader and load another article. Check that the URL is now returned with the `now` timestamp.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Me/App Settings/DebugMenuViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/App Settings/DebugMenuViewController.swift
@@ -72,7 +72,7 @@ class DebugMenuViewController: UITableViewController {
                 ReaderCSS().clearCache()
                 let alert = UIAlertController(title: NSLocalizedString("Cache cleared!", comment: "Debug message informing the user that the cache for CSS in the Reader has been cleared"),
                                               message: nil, preferredStyle: .alert)
-                alert.addCancelActionWithTitle(NSLocalizedString("Dismiss", comment: ""))
+                alert.addCancelActionWithTitle(NSLocalizedString("Dismiss", comment: "Dismisses an alert"))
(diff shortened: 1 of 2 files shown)


leandroalonso:
@frosty I'd rather try another approach.

When we're fixing issues in the CSS it's very common that you do some change and then check the result by opening a post (repeat 1000x). This current approach will require the user to go to Post Settings and clear the cache every time they need a fresh CSS.

I think we can add a flag that once enabled will always load a new CSS, something like (in `ReaderCSS.swift`):

```swift
    var address: String {
        guard !FeatureFlag.freshReaderCSS.enabled else {
            saveCurrentDate()
            return url(appendingTimestamp: now)
        }
```

If the flag is enabled this will always load a new CSS every time you open an article, which makes testing very easy. :)

frosty:
@leandroalonso Great! Thanks for the suggestion! 🙂

loremattei:
Hey there! I'm moving this to 15.8 because 15.7 has been cut. If you want this to make it to 15.7, please, feel free to ping me.

leandroalonso:
@frosty I created a PR to add the changes I suggested in my prev comment, hope you don't mind: https://github.com/wordpress-mobile/WordPress-iOS/pull/14859

frosty:
Of course not, thanks for doing that! Sorry I didn't get chance to tweak this one. Will close in preference to yours.

PR #14847, opened 2020-09-04
Add local cache for What's New - feature announcements

### LINKED ISSUE #14585: Add local cache for What's New
Add a local cache for feature announcements to be used until the user updates to a new version.

Gio2018:
Fixes #14585

WordPressKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/279

To test:

- put two breakpoints at lines 18 and 21 in `AnnouncementsCache.swift`
- go to Me/App Settings/What's New in WordPress
- Make sure that the execution pauses at line 21
- Resume the execution and make sure that the screen looks like the examples below
- do it again (except the breakpoints, that are already there)
- Make sure this time the execution pauses at line 18
- While you're paused, take a loot at the contents of `UserDefaults.announcements` and make sure they look consistent with the screenshots above
- Resume the execution and make sure that the screen has not changed

To reset the `UserDefaults`, pause the debugger and type

`e -l swift -- import WordPress; UserDefaults.standard.announcements = nil`

iPhone | iPad
--- | ---
 | 

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ Future PR.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ Future PR.


WordPress/Classes/ViewRelated/What's New/Data store/AnnouncementsStore.swift
```diff
@@ -45,16 +48,26 @@ class RemoteAnnouncementsStore: AnnouncementsStore {
...
 
+    init(cache: AnnouncementsCache) {
+        self.cache = cache
+    }
+
     func getAnnouncements() {
-        let service = AnnouncementServiceRemote(wordPressComRestApi: api)
         state = .loading
+        if let announcements = cache.announcements {
+            self.state = .ready(announcements)
```
frosty: Nitpick, but don't need `self` here.

WordPress/Classes/ViewRelated/What's New/Data store/AnnouncementsStore.swift
```diff
@@ -45,16 +48,26 @@ class RemoteAnnouncementsStore: AnnouncementsStore {
...
+
+        let service = AnnouncementServiceRemote(wordPressComRestApi: api)
         service.getAnnouncements(appId: Identifiers.appId,
                                  appVersion: Identifiers.appVersion,
                                  locale: Locale.current.identifier) { result in
 
             switch result {
             case .success(let announcements):
                 self.state = .ready(announcements)
+                self.cache.announcements = announcements
```
frosty: Should these two statements be the other way round? Setting the state emits the change event, but so the cached announcements should probably be in place before that happens.

WordPress/Classes/ViewRelated/What's New/Data store/Cache/AnnouncementsCache.swift
```diff
@@ -0,0 +1,60 @@
...
+            }
+            return announcements
+        }
+        set {
+            UserDefaults.standard.announcements = newValue
+        }
+    }
+
+    private func versionIsValid(for announcements: [Announcement]) -> Bool {
+        guard let minimumVersion = announcements.first?.minimumAppVersion, // there should not be more than one announcement
```
frosty, line 26: It might be good to have a couple of tests for this code.

frosty (commented):
Works for me – just a couple of comments!

WordPress/Classes/ViewRelated/What's New/Data store/AnnouncementsStore.swift
Gio2018: So what sets the state is only the `getAnnouncements` method.

The flow should be:

- if there are cached values, use them and return
- if not:
  - fetch them from the backend and, as soon as you have them, use them (do not wait to save them on disk)
  - "at the same time", save them on disk, but don't let this operation block your process

So I believe that the change here would be to dispatch the save-on-disk asynchronously on the global queue

WordPress/Classes/ViewRelated/What's New/Data store/Cache/AnnouncementsCache.swift
Gio2018: Sure, as the description says, Unit Tests will be added in a separate PR.

mindgraffiti (approved):
"It's clobbering time!" 😁 

:shipit:

WordPress/Classes/ViewRelated/What's New/Data store/Cache/AnnouncementsCache.swift
frosty: Sorry, missed that! :)

PR #14848, opened 2020-09-04
Update to Xcode 12

jkmassel:
This PR migrates our CI system to Xcode 12. It makes a few changes:

- It uses `fastlane` to build for testing, rather than using `xcodebuild` directly. This better aligns with how we do release builds, and it fixed a weird issue with a Swift compiler mismatch. IMHO moving forward we should probably deprecate the direct `xcodebuild` part of the orb and use `fastlane` for all the things?
- It uses the iPad Air 4th generation (X-style) for UI tests. If we want a simulator with a home button, we could use the 8th generation iPad instead.
- Instead of passing the exact path to the `.xctestrun` file that you want, now you can just pass part of the name of the file to match (ie – `UITest` or `UnitTest`), and it'll do the right thing. This should make future upgrades easier.
- It removes the "destination" string – fastlane was warning that we shouldn't use it, so I've removed that argument
- There are a couple specific fixes that needed to happen – we were linking the test bundles against a bunch of libraries unnecessarily, so I've removed that, and I fixed a compiler error in a test

To Test:
- Ensure CI passes

### DIFF
--- a/Scripts/fastlane/Fastfile
+++ b/Scripts/fastlane/Fastfile
@@ -614,7 +614,7 @@ import "./ScreenshotFastfile"
   # It requires a prebuilt xctestrun file and simulator destination where the tests will be run.
   # -----------------------------------------------------------------------------------
   # Usage:
-  # bundle exec fastlane test_without_building [name:<Partial name of the .xctestrun file>] [destination:<Simulator>] [try_count:<Number of times to try tests>]
+  # bundle exec fastlane test_without_building [name:<Partial name of the .xctestrun file>] [try_count:<Number of times to try tests>]
   #
   # Example:
   # bundle exec fastlane test_without_building name:UITests try_count:3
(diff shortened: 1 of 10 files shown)


Scripts/fastlane/Fastfile
```diff
@@ -604,19 +614,28 @@ import "./ScreenshotFastfile"
   # It requires a prebuilt xctestrun file and simulator destination where the tests will be run.
   # -----------------------------------------------------------------------------------
   # Usage:
-  # bundle exec fastlane test_without_building [xctestrun:<Path to xctestrun file>] [destination:<Simulator>] [try_count:<Number of times to try tests>]
+  # bundle exec fastlane test_without_building [name:<Partial name of the .xctestrun file>] [destination:<Simulator>] [try_count:<Number of times to try tests>]
```
AliSoftware: Wasn't the `destination:` option removed?

```suggestion
  # bundle exec fastlane test_without_building [name:<Partial name of the .xctestrun file>] [try_count:<Number of times to try tests>]
```

AliSoftware (approved):
🎉

I'm all for using fastlane instead of xcodebuild orb action directly too

loremattei (approved):
Code changes look good to me!
I tested it locally and everything worked as expected. 
CI is green. 

Other than the small issue in the comment that Olivier pointed out, this looks good to me!

Scripts/fastlane/Fastfile
jkmassel: Oops yep! Thanks!

Issue #13757, opened 2020-03-27
[GlobalStep] iOS - Full Screen images re-size and position incorrectly when rotating device.


wptester9845:
### Description
Opening any image on any iOS device will display it correctly. As soon as the user rotates the device in any direction, the image will begin to re-size and positioning incorrectly. The image will not display properly in Full Screen mode again until leaving and returning to this mode.

### Reproduction Rate
4/4 100%

### Expected behaviour
Full Screen images should re-size and stay within screen edges.

### Actual behavior
The images on Full Screen will re-size randomly and have portions out of screen.

### Steps to reproduce the behavior
1. Install and Launch WordPress 14.5
2. Log in with a valid email address.
3. Tap 'Continue'.
4. Scroll down to 'Publish' and tap 'Media'.
5. Tap on an image and tap again to full screen it.
6. Rotate the device and observe the image's behaviour.

##### Tested on the following
iPhone 8 (13.2.3)
iPhone SE (13.1.3)
iPhone 5S (12.0.0)
iPad Air 3 (13.2.3)

#### Please see the attached video for more information
[iOS fullscreen image rotating.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/4392788/iOS.fullscreen.image.rotating.zip)

### Submitted by:
Ricardo Silva

designsimply:
Tested and confirmed. Also found that the problem reported here where the fullscreen image doesn't look right after rotating the device also happens when accessing the preview in the block editor. ([43s](https://cloudup.com/cuvz9DQ7ZIN))

Confirmed it's happening as far back as 13.9.0.2 (downloaded from TestFlight). ([28s](https://cloudup.com/cGb4jWR41_c))

For reference: fullscreen previews were implemented in the block editor in https://github.com/wordpress-mobile/gutenberg-mobile/pull/1597.

Adding to Groundskeeping since the problem isn't new in this release.

dangermattic:
Thanks for reporting! 👍

PR #14852, opened 2020-09-07
Images: Fixes an issue with image layout after rotation.

### LINKED ISSUE #10099: WPImageViewController has broken layout on rotation
Edit: This isn't just limited to gifs, and happens for any images viewed full screen.

When previewing gifs in `WPImageViewController`, the image doesn't remain fixed to the centre of the view if you rotate the device:

This affects the new Giphy integration, as well as simply previewing gifs that are already present in your media library.

**To reproduce**

Either:

* Upload a gif to the site via your computer
* Open the Media Library for a site
* Find the gif you uploaded, and tap to view its details
* Tap the gif preview to view it full screen
* Rotate the device

Or

* Open the Media Library for a site
* Tap +
* Choose Giphy
* Tap one or more gifs to preview
* Tap Preview in the bottom toolbar
* Rotate the device

Tested on iPhone X (device) and iPhone 8 (simulator), iOS 11.4 and iOS 12, WPiOS 10.8

aerych:
Fixes #10099 
Refs #13757 (fixes?)

The call to `centerImage` was happening before final layout was determined, and was thus using stale geometry. Animating the change along side the transition seems to fix the issue.

To test:
With an iPhone (or iPhone simulator) preview an image either in the Reader or via the Media feature (they both use the WPImageViewController). 
Switch between portrait and landscape orientation.  Confirm the image is correctly positioned.

PR submission checklist:

@frosty could I trouble you for a review?


frosty (approved):
Thank you for fixing this! I can confirm that images are rotating correctly for me.

Should we add this as a bug fix note to the release notes?

aerych:
> Should we add this as a bug fix note to the release notes?

Sure thing.  How does that look?

frosty (approved):
Looks good to me! :shipit:

aerych:
Thanks @frosty !!

PR #14854, opened 2020-09-07
Share: Adjusts the Share Extension cancel behavior to not completely dismiss

aerych:
This PR modifies the way the Share Extension is dismissed when cancelled, so the user is left on the share selection screen rather than taken all the way out of sharing.  

For example: when sharing a photo from the Photos app, a user viewing a grid of photos might tap one one to view it, then tap the share button.  Then on the selection screen choose an app to share to.  If they choose WordPress, then clicked the cancel button, they'd be taken out of the share selection screen, rather than back to it.  This patch cancels sharing (rather than completing it) so the user is taken back to the previous screen. 

To test:
- [ ] Confirm the previous behavior by trying to share a photo from the photos app to WordPress and canceling the cancel button in the share extension. 
- [ ] Switch to this branch and confirm that you are taken back to the share selection screen. 
- [ ] Confirm that posting from the share extension still takes you all the way out of the sharing feature.
- [ ] Confirm that the dismissal behavior in the iPad is as expected in both scenarios.

@frosty @diegoreymendez would either of you have time to take a peek? 

PR submission checklist:


WordPress/WordPressShareExtension/ShareModularViewController.swift
```diff
@@ -214,17 +214,7 @@ class ShareModularViewController: ShareExtensionAbstractViewController {
 
 extension ShareModularViewController {
     fileprivate func dismiss() {
-        // In regular width size classes (iPad), action extensions are displayed
-        // in a small modal, which looks strange when this VC is dismissed
-        // before the main / presenting controller with its white background.
-        // This workaround simply dismisses the modular VC along with the main extension VC.
-        // See https://github.com/wordpress-mobile/WordPress-iOS/issues/8646 for more info.
-        guard UIDevice.isPad() == false && originatingExtension != .saveToDraft else {
```
aerych, line 222: As far as I can tell, the share UI has changed since this was originally done and this work around is no longer needed.

WordPress/WordPressShareExtension/MainShareViewController.swift
```diff
@@ -82,8 +82,13 @@ private extension MainShareViewController {
 
     func loadAndPresentNavigationVC() {
         editorController.context = self.extensionContext
-        editorController.dismissalCompletionBlock = {
-            self.extensionContext?.completeRequest(returningItems: nil, completionHandler: nil)
+        editorController.dismissalCompletionBlock = { (exitSharing) in
```
frosty: Should we `[weak self]` here?

WordPress/WordPressShareExtension/MainShareViewController.swift
```diff
@@ -82,8 +82,13 @@ private extension MainShareViewController {
...
+            if exitSharing {
+                self.extensionContext?.completeRequest(returningItems: nil, completionHandler: nil)
+            } else {
+                let error = NSError(domain: "com.wordpress.sharing", code: 0, userInfo: nil)
```
frosty: Should we use a constant for the error code, with a descriptive name like ShareCancelled?

WordPress/WordPressShareExtension/ShareModularViewController.swift
frosty: Right, I'm seeing the extension always displayed in a large modal, not a small one.

frosty (commented):
- [x] Confirm the previous behavior by trying to share a photo from the photos app to WordPress and canceling the cancel button in the share extension.
- [x] Switch to this branch and confirm that you are taken back to the share selection screen.
- [x] Confirm that posting from the share extension still takes you all the way out of the sharing feature.
- [x] Confirm that the dismissal behavior in the iPad is as expected in both scenarios.

Changes look great! Thanks for fixing this. I just left two small code comments – see what you think.

WordPress/WordPressShareExtension/MainShareViewController.swift
aerych: Hmm... good question.  After a second look I've opted to just make an error enum with a single canceled case.

aerych:
Thanks @frosty!  Quick peek at the changes?

frosty (approved):
:shipit:

aerych:
Thanks @frosty !!

PR #14859, opened 2020-09-08
Add Reader's CSS debug options

leandroalonso:
This PR adds two debug options for the Reader, to make it easier to update blocks and CSS:

1. It allows to always retrieve a fresh/uncached CSS from the server
2. It allows the user to change the CSS URL (using a local one, for example)

### Retrieving an uncached CSS

1. Go to Settings > Debug
2. Enable "Always fetches a fresh Reader CSS"
3. Add a breakpoint [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderWebView.swift#L40)
4. Load a post in Reader, check the timestamp
5. Open another post, check that the timestamp at the end of the URL is different

### Overriding the URL

1. Go to Settings > Debug
2. Tap "Reader CSS URL" (at the end of the view)
3. Change the URL to something else
4. Add a breakpoint [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderWebView.swift#L40)
5. Load a post in Reader and check that the returned URL is the one you just added

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
+++ b/WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
@@ -100,7 +100,7 @@ extension FeatureFlag {
         case .debugMenu:
             return "Debug menu"
         case .readerCSS:
-            return "Always fetches a fresh Reader CSS"
+            return "Ignore Reader CSS Cache"
         case .unifiedAuth:
             return "Unified Auth"
         case .unifiedSiteAddress:
--- a/WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
+++ b/WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
@@ -4,6 +4,7 @@
 enum FeatureFlag: Int, CaseIterable, OverrideableFlag {
     case jetpackDisconnect
     case debugMenu
+    case readerCSS
     case unifiedAuth
     case unifiedSiteAddress
     case unifiedGoogle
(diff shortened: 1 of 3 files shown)


WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -96,6 +99,8 @@ extension FeatureFlag {
             return "Jetpack disconnect"
         case .debugMenu:
             return "Debug menu"
+        case .readerCSS:
+            return "Always fetches a fresh Reader CSS"
```
emilylaguna: I am a bit confused by what this meant. Perhaps this is a better description?

```suggestion
            return "Ignore Reader Detail CSS Cache"
```

emilylaguna (changes requested):
This is working for me, made a slight suggestion to change the feature flag description.

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
leandroalonso: Done in 8fa1b5f1a95504ff15d32c530bbe6174e0319606

I just removed the word "Detail" to keep it more concise for everyone (iOS devs or not).

emilylaguna (approved):
Works for me!

PR #14866, opened 2020-09-09
UL&S iCloud Keychain: attempt to fix backgrounding issue.

diegoreymendez:
WPAuth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/446

As reported by @ScoutHarris via a private conversation, it seems that some conditions are making the iCloud Keychain flow to sometimes start under these conditions:

1. WPiOS is running in the background.
2. WPiOS is logged out.
3. The device is idling.

## Theory:

My theory is that there's some background process that triggers calls to `viewDidAppear` in the root VC.

### Examples of this being an issue:
- [PushKit background notifications triggering calls to `viewDidAppear` in root VC](https://stackoverflow.com/questions/53881314/pushkit-background-notification-triggers-viewdidappear-on-initial-view-controlle)
- [Location changes triggering calls to `viewDidAppear` in root VC](https://stackoverflow.com/questions/30584356/viewdidappear-is-called-when-app-is-started-due-to-significant-location-change)

## Attempts to reproduce the issue:

I tried to reproduce this issue by sending notifications to the device:

1. I first logged in to WPiOS to enable notifications.
2. Then I logged back out.

Then I tried simulating notifications using `xcrun simctl push <SIMULATOR_DEVICE_ID> <YOUR_APP_BUNDLE_ID> <APNS_FILE_NAME>` but those notifications weren't really triggering calls to:

```swift
func application(_:, didReceiveRemoteNotification:, fetchCompletionHandler:)
```

I also tried a tool provided by @jkmassel, but that tool would only work while logged in.  I wonder if there's any type of notification or background update that we use that can trigger activity in the app while logged out and in the background.

## Proposed solution:

The submitted code is a proposed solution IF the issue is caused by background processes triggering a call to `viewDidAppear`.

## Testing:

Due to what I described above, there are no known testing steps for this issue.  This is a tentative fix.

## PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -191,10 +191,10 @@ target 'WordPress' do
 
     pod 'Gridicons', '~> 1.0.1'
 
-    pod 'WordPressAuthenticator', '~> 1.24.0'
+    # pod 'WordPressAuthenticator', '~> 1.24.0'
     # While in PR
     # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
-    # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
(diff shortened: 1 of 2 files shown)


ScoutHarris (approved):

diegoreymendez:
I'm closing this PR as the WPAuth changes will be integrated in an upcoming PR.

PR #14874, opened 2020-09-10
Wordwrap NUX button

### LINKED ISSUE #155: NUX buttons are truncated on smaller devices
I just launched the app on the iPhone SE simulator, and noticed that the "continue with..." buttons in the NUX are truncated on smaller devices.

Per pbArwn-5X-p2, scale the button text to fit the width instead of truncating.

I guess we'll need to either wrap the text or shrink it when this happens.

mindgraffiti:
Fixes: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/155
Auth PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/452

There is an issue where iOS 12.0+ users on smaller devices do not get proper font scaling. This is an iOS 12 issue that doesn't happen on iOS 13. Instead of attempting to force iOS 12 to use the correct font scaling, we're going to word-wrap the long text in the button so it can be successfully read by the user.

Note that the buttons in the prologue are different because the UL&S project had updated designs to work from.

| Before | After |
| --- | --- |
|  |  <a href="https://user-images.githubusercontent.com/1062444/92657284-2a8ddb00-f2ba-11ea-8478-0fd1c9822abc.png"></a> |

### To test
Use a small device or simulator on iOS 12, such as the iPhone SE or iPhone 5S.

1. Log out if logged in
2. Observe the prologue buttons

PR submission checklist:


ScoutHarris (approved):
Looks good! Thanks for fixing this!!

Simulator Screen Shot - iPhone SE - 2020-09-10 at 10 13 14

PR #14880, opened 2020-09-11
[Gutenberg] UBE: Pressenting Jetpack Security settings on Missing Block

etoledom:
Gutenberg-mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/2610

This PR implements changes from https://github.com/wordpress-mobile/gutenberg-mobile/pull/2610 to present the Jetpack Security settings modally over Gutenberg, and refresh the editor capabilities to enable the Unsupported Block Editor.

**NOTE:** The action button text for UBE Enabled has been updated to `Edit using web editor`.
jetpack_ube

### To test:
- All tests must be run with metro running on https://github.com/wordpress-mobile/gutenberg-mobile/pull/2610

#### Jetpack connected sites (Self-Hosted and Atomic)
- Have a Jetpack connected site with a post with an unsupported block on it.
- Be sure to have SSO disabled for this site.
- Open the post with the mobile app.
- Press the (?) button on the missing block.
  - Check that the bottom sheet looks like the one from the image `Jetpack - UBE Disabled`.
- Press the action button `Open Jetpack security settings`.
 - Check that the Jetpack Security Settings appears modally.
- Activate the option `Allow WordPress.com login` and press Done.
- Press (?) button again.
  - Check that now the bottom sheet has changes, and looks like `Jetpack - UBE Enabled` from the image.
- Press again the action button  `Edit using web editor`.
  - Check that UBE works as expected. (If its a new site, you might need to allow WordPress.com login on the web view. This is expected behaviour.

#### Self-Hosted sites.
- Add a self-hosted site to the mobile app (not connected to wpcom).
- Open a post with an unsupported block in it.
- Press the (?) button on the missing block.
  - Check that it looks like the one from the image `Self-Hosted - Disabled`.

#### WPCom Simple sites.
- On a WPCom Simple site:
- Open a post with an unsupported block in it.
- Press the (?) button on the missing block.
  - Check that it looks like the one from the image `WPCom Simple - Enabled`.
- Press the action button `Edit using web editor`.
  - Check that UBE loads as expected.

PR submission checklist:


jkmassel:
👋 We're freezing `15.8` today, so I'm bumping this PR forward to `15.9`. If you need it to be part of the `15.8` release cycle, please merge to the `release/15.8` branch and DM me – I'll be happy to cut a new beta release!

marecar3:
Hey 👋 @etoledom 
I have updated gb-mobile reference to the latest and changed this PR to ready for review if that is ok?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -464,6 +464,25 @@ class GutenbergViewController: UIViewController, PostEditor {
...
+        let JetpackSSOEnabled = (blog.jetpack?.isConnected ?? false) && (blog.settings?.jetpackSSOEnabled ?? false)
+        if JetpackSSOEnabled == false {
+            let controller = JetpackSecuritySettingsViewController(blog: blog)
+            controller.navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(jetpackSettingsControllerDismissed))
+            let navController = UINavigationController(rootViewController: controller)
+            present(navController, animated: true)
+        }
+    }
+
+    @objc private func jetpackSettingsControllerDismissed() {
```
guarani: Similar to `handleMissingBlockAlertButtonPressed`, would this be better named `jetpackSettingsControllerDoneButtonPressed`?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -912,6 +940,7 @@ extension GutenbergViewController: GutenbergBridgeDataSource {
         return [
             .mentions: post.blog.isAccessibleThroughWPCom() && FeatureFlag.gutenbergMentions.enabled,
             .unsupportedBlockEditor: isUnsupportedBlockEditorEnabled,
+            .unsupportedBlockEditorSwitch: post.blog.jetpack?.isConnected ?? false,
```
guarani: I'm wondering if something like `unsupportedBlockEditorRequiresUserInteraction` would be more descriptive?

guarani: Also, now that we're handling Jetpack sites here, should we remove this comment:
https://github.com/wordpress-mobile/WordPress-iOS/blob/0416d76792078e0b2a8e902e2361709844252cf2/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L922

guarani (approved):
I tested all three scenarios and this works great! 🎉 I've commented on a couple of things above, but they are relatively minor.

It's worth noting that from the Jetpack settings modal screen, I can pretty much navigate anywhere (e.g. sign up for a Jetpack account) — I'm not sure if we want to block those off somehow (otherwise users may get lost).

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
etoledom: These names are really tricky 😬 
So, this capability is about the site being able to enable Unsupported Blocks Editor if it's disabled.
This would be the general description to no mention SSO or Jetpack. That's why it's `true` for any Jetpack site, even if SSO is already active and it doesn't require user interaction.

Maybe a better name could be `unsupportedBlockEditorSwitchable` or `...Activable`? 🤔 
I really don't like this "capability" but we need this information.

etoledom: From https://github.com/WordPress/gutenberg/pull/25238#discussion_r494792659
Maybe `unsupportedBlockEditorCanBeEnabled` could also work for the capability name 🤔 
Wdyt?

etoledom: cc @marecar3 - since we might need to change the capability name in both platforms

guarani: > That's why it's true for any Jetpack site, even if SSO is already active and it doesn't require user interaction.

True, so `unsupportedBlockEditorRequiresUserInteraction` wouldn't work either.

> I really don't like this "capability" but we need this information.

It seems like it isn't a "capability", but I don't know where else we could put it either.

etoledom: Hah yeah, that sounds right. It's more a capability of the site that's using using Gutenberg rather than of gutenberg itself.
This feels more like a `property` of the Unsupported Block Editor capability.

PR #14884, opened 2020-09-11
Reader Detail: Add async featured image loading

emilylaguna:
Project Issue: #14767

### To test:
1. Launch the app
2. Tap on the Reader
3. Locate a post with a featured image
4. Enable the Network Link Conditioner to reduce the internet speed
5. Tap on the post
6. You should see the ghost loading view
7. You should see the content load in with a gray image where the featured image will be
8. After a few seconds the featured image will load in

### Notes
- I originally tries to use the ghostable view for the loading, but since those use layers they have implicit animations when the user scrolls and the view height changes the animation is very slow and does not look great. 

### PR submission checklist:


WordPress/Classes/Utility/Image Dimension Parser/ImageDimensionFetcher.swift
```diff
@@ -0,0 +1,76 @@
...
+
+class ImageDimensionsFetcher: NSObject, URLSessionDataDelegate {
+    // Helpful typealiases for the closures
+    public typealias CompletionHandler = (ImageDimensionFormat, CGSize?) -> Void
+    public typealias ErrorHandler = (Error?) -> Void
+
+    let completionHandler: CompletionHandler
+    let errorHandler: ErrorHandler?
+
+    // Internal use properties
```
leandroalonso, line 11: The comment is redundant (given the `private` keyword)

WordPress/Classes/Utility/Image Dimension Parser/ImageDimensionFetcher.swift
```diff
@@ -0,0 +1,76 @@
...
+    private let request: URLRequest
+    private var task: URLSessionDataTask? = nil
+    private let parser: ImageDimensionParser = ImageDimensionParser()
```
leandroalonso: Can we add to the constructor as a default parameter?

```
init(request: URLRequest, success: @escaping CompletionHandler, error: ErrorHandler? = nil, parser: ImageDimensionParser = ImageDimensionParser()) {
```

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
```diff
@@ -259,14 +274,39 @@ class ReaderDetailFeaturedImageView: UIView, NibLoadable {
...
-        imageLoader.loadImage(with: imageURL, from: post, placeholder: nil, success: { [weak self] in
-            self?.didFinishLoading()
+        let completionHandler: (CGSize) -> Void = { [weak self] size in
+            guard let self = self else {
+                return
+            }
+
+            self.imageSize = size
+            self.didFinishLoading()
+            self.isLoading = false
```
leandroalonso, line 297: There are two remaining issues here:

* Saved posts (which are meant to work while offline or in a bad connection)
* Users in a really bad connection

I guess both cases can be solved by adding a timeout. It probably won't affect the current functionality. Wdyt?

leandroalonso (changes requested):
@emilylaguna tested and worked like described.

Left a few small suggestion for really bad networks.

WordPress/Classes/Utility/Image Dimension Parser/ImageDimensionFetcher.swift
emilylaguna: Correct, I like to have a visual separator comment that groups together various properties. I do the same with IBOutlet's and public use properties. I find this helps with scanning the "header" of the class a bit easier.

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
emilylaguna: I agree, how long of a timeout do you think we should use?

WordPress/Classes/Utility/Image Dimension Parser/ImageDimensionFetcher.swift
emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14884/commits/81f1341e58fc0e665f17d387ca7ce40e63db6388

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
leandroalonso: 3 or 4 seconds? I thought about 5 but it seems a lot, and in most cases (even in bad connections) the image size will load in less than a second, right?

emilylaguna: Done in: https://github.com/wordpress-mobile/WordPress-iOS/pull/14884/commits/0aa3f5080b86c8eedca091b3c6414656ae856ce7

emilylaguna:
@leandroalonso back to you!

leandroalonso (approved):

Issue #14844, opened 2020-09-04
@-mentions not working on Atomic sites


guarani:
### Expected behavior

@-mentions (or suggestions) should work for all types of WordPress.com sites, including those with plugins installed (Atomic). 

Note: This does work on WordPress for Android.

| Android ✅ | iOS ❌ |
| - | - |
| <img alt="Screenshot showing @-mentions working on Atomic sites on WordPress for Android" width="300" src="https://user-images.githubusercontent.com/1898325/92261184-063f9200-eea7-11ea-90ad-f2f9a1320e6f.jpg" /> | <img alt="Screenshot showing @-mentions working on Atomic sites on WordPress for Android" width="300" src="https://user-images.githubusercontent.com/1898325/92261241-1bb4bc00-eea7-11ea-9556-5e0b2b9d4a56.PNG" /> |

### Actual behavior

In Reader, the @ character does not trigger a list of users to be shown when the site in question is Atomic, but it does work if the site is a regular WordPress.com site. 
It doesn't work in Notifications or Gutenberg, either.

### Steps to reproduce the behavior

1. On the WordPress for iOS app, go to the Reader and locate a post from an Atomic site. If none available, create one like so:
  a. Find or create a WordPress.com site with a Business plan, install any plugin
  b. Publish a post from that site
  c. Add another user to the site
  d. Log in to the WordPress for iOS app using that recently added user
2. Create a comment on that post, type the @ character and notice that no UI is shown for adding mentions

guarani:
This is due to this code:
https://github.com/wordpress-mobile/WordPress-iOS/blob/ed9df296c60e0875e4121e3e8595aa9787638578/WordPress/Classes/Services/SuggestionService.m#L114-L118

The check `[blog supports:BlogFeatureMentions]` incorrectly interprets Atomic sites as not being hosted on WordPress.com:

https://github.com/wordpress-mobile/WordPress-iOS/blob/8bfca4663f4beeab131ec1d67e2d4cf6f946f5a0/WordPress/Classes/Models/Blog.m#L478-L479

Note: The `isHostedAtWPcom` property is false for Atomic sites in my tests. (It should really be named `isWPComSimpleSite` or similar.)

hypest:
Fixed by #15317

PR #14888, opened 2020-09-12
Port Suggestions from Objective-C to Swift

guarani:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/2577

Ports/migrates the code from `SuggestionService` and `Suggestion` from Obj-C to Swift in preparation for changes in https://github.com/wordpress-mobile/WordPress-iOS/pull/14743.

@-mentions are used in two places in the app: Reader comments and Notification comments/replies.

Known issues:
- Reader: @-mentions not working on Atomic sites: https://github.com/wordpress-mobile/WordPress-iOS/issues/14844 (**existing bug**, not related to this PR)
- Supporting @-mentions on Jetpack connected self-hosted sites: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2601

## How to test

#### Checklist 

```
- [ ] Reader comments
- [ ] Comment on post (via notification)
- [ ] Reply to comment (via notification)
```

**What you'll need**: A WordPress.com site (may be your own site) that you can comment on and that has one or more other WordPress.com users (in order to test @-mentions). Make sure some of the users have a profile photo. The user that you login with will also need:
- a notification about a post
- a notification about a comment

### Reader comments

<details>
<summary>Test steps</summary>
<ol>
<li>In the Reader tab, locate the site</li>
<li>Navigate to a post, tap the comment button</li>
<li>Tap the &quot;Reply to post...&quot; text box to bring up the comment UI</li>
<li>Type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions to add it to your comment and hit send</li>
<li>Verify that the mentioned user receives a notification</li>
</ol>
</details>

### Comment on post (via notification)

<details>
<summary>Test steps</summary>
<ol>
<li>In the Notifications tab, locate a notification about a post</li>
<li>Open it, then tap the comment button, then the &quot;Reply to post...&quot; text box to bring up the comment UI</li>
<li>Type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions to add it to your comment and hit send</li>
<li>Verify that the mentioned user receives a notification</li>
</ol>
</details>

### Reply to comment (via notification)

<details>
<summary>Test steps</summary>
<ol>
<li>In the Notifications tab, locate a notification about a comment</li>
<li>Open it, then tap the &quot;Write a reply...&quot; text box to bring up the comment UI</li>
<li>Type <code>@</code> and verify that the list of suggestions is shown</li>
<li>Make sure that the avatar images load for users who have a profile photo</li>
<li>Select one of the suggestions to add it to your reply and hit send</li>
<li>Verify that mentioned user receives a notification</li>
</ol>
</details>

---

PR submission checklist:


guarani:
Tested build 34789

- [x] Reader comments
- [x] Comment on post (via notification)
- [x] Reply to comment (via notification)

guarani:
Could I bother one of you for a review please? @leandroalonso @emilylaguna

yaelirub:
@guarani , I apologize for the long wait time for code review! 
If you still need this one reviewed, I wonder if @frosty , @Gio2018 can help 🙏

guarani:
@yaelirub no worries, I'd already been given a headsup that the team would be AFK this week, I just hadn't updated the reviews — thanks for the reminder!

guarani:
@frosty @Gio2018, would either of you be up for a review please? 🙌

Gio2018:
> @guarani , I apologize for the long wait time for code review!
> If you still need this one reviewed, I wonder if @frosty , @Gio2018 can help 🙏

For sure!! Will review later today.

jkmassel:
👋 We're freezing `15.8` today, so I'm bumping this PR forward to `15.9`. If you need it to be part of the `15.8` release cycle, please merge to the `release/15.8` branch and DM me – I'll be happy to cut a new beta release!

Gio2018 (changes requested):
Hey @Guarani !! Still have to do a full code review, but I noticed that while the test cases worked fine on iPhone X with iOS 13.7, mentions do not seem to work on iPad Air 2, iPadOS 13.7 (both real devices, not simulators)

guarani:
Thanks for the report @Gio2018! I'll test this on the iPad simulator (I don't have an iPad handy).

WordPress/Classes/Models/Suggestion.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+@objcMembers public class Suggestion: NSObject {
+    let userLogin: String?
```
frosty, line 4: Should any of these be non-optional? Is a suggestion valid without a login or display name? If so, we should make these non-optional and have a failable initializer.

WordPress/Classes/Models/Suggestion.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+@objcMembers public class Suggestion: NSObject {
```
frosty, line 3: While we're here, could we just add a a comment explaining what this is used for?

WordPress/Classes/Models/Suggestion.swift
```diff
@@ -0,0 +1,45 @@
...
+        let type = avatarSourceType(with: &hash)
+
+        if let hash = hash, let type = type, let success = success {
+            WPAvatarSource.shared()?.fetchImage(forAvatarHash: hash as String, of: type, with: size, success: success)
+        } else {
+            success?(nil)
+        }
+    }
+
+    func avatarSourceType(with hash: inout NSString?) -> WPAvatarSourceType? {
```
frosty, line 49: I'd love it if we could update `WPAvatarSource.parseURL(forAvatarHash:)` so that it doesn't require an inout parameter, but I guess that might be outside of the scope of these changes.

WordPress/Classes/Services/SuggestionService.swift
```diff
@@ -0,0 +1,118 @@
...
+
+extension NSNotification.Name {
+    static let suggestionListUpdated = NSNotification.Name("SuggestionListUpdated")
+}
+
+@objc extension NSNotification {
+    public static let suggestionListUpdated = NSNotification.Name.suggestionListUpdated
+}
+
+@objc public class SuggestionService: NSObject {
```
frosty: It would be good to add a comment here explaining what the class is used for. We have a number of different suggestion-related classes in the app (like domain suggestions, Reader suggestions), so would be helpful to differentiate.

WordPress/Classes/Services/SuggestionService.swift
```diff
@@ -0,0 +1,118 @@
...
+
+    // NSCache works with classes such as NSArray, not structs such as Suggestion or [Suggestion].
```
frosty: Suggestion currently isn't a struct, though?

WordPress/Classes/Services/SuggestionService.swift
```diff
@@ -0,0 +1,118 @@
...
+        let defaultAccount = accountService.defaultWordPressComAccount()
+        let params = ["site_id": siteID]
+
+        defaultAccount?.wordPressComRestApi.GET(suggestPath, parameters: params, success: { [weak self] responseObject, httpResponse in
+            guard let `self` = self else { return }
+            guard let payload = responseObject as? [String: Any] else { return }
+            guard let restSuggestions = payload["suggestions"] as? [[String: Any]] else { return }
+
+            let suggestions = restSuggestions.map { Suggestion(dictionary: $0) }
+            self.suggestionsCache.setObject(Wrapper(suggestions), forKey: siteID)
```
frosty: The original code sets a cost for the suggestions based on their count. Is that no longer required? (I think probably unnecessary)

WordPress/Classes/Services/SuggestionService.swift
```diff
@@ -0,0 +1,118 @@
...
+    */
+    @objc func shouldShowSuggestions(for siteID: NSNumber?) -> Bool {
+        guard let siteID = siteID, let appDelegate = WordPressAppDelegate.shared else {
+            return false
+        }
+
+        let suggestions = suggestionsCache.object(forKey: siteID)?.value
+
+        // if the device is offline and suggestion list is not yet retrieved
+        if !appDelegate.connectionAvailable && suggestions == nil {
```
frosty: Rather than accessing the app delegate here, I think it would be better to use `ReachabilityUtils.isInternetReachable()` (does the same thing, but abstracted a layer)

frosty (commented):
I ran through your testing scenarios and they all looked good to me! I've left some code level comments.

As there are only a handful of places in Objective-C that we access `Suggestion` and `SuggestionService`, I think it might be worthwhile us porting those small few pieces into Swift (we could add them as Swift extensions to their classes). This would allow us to drop the `@objcMembers` and `NSObject` subclassing. If we do that, we could also potentially turn Suggestion into a struct, which would be nice as it's really a simple value type. If we wanted, we could also make it Decodable. What do you think?

WordPress/Classes/Services/SuggestionService.swift
guarani: Thanks for catching! (It was at one point during my refactor, then I switched to a class.)
I'll see if we can switch it back over to a struct using the approach you suggested here: https://github.com/wordpress-mobile/WordPress-iOS/pull/14888#pullrequestreview-493611751

WordPress/Classes/Services/SuggestionService.swift
guarani: I'd overlooked that, thanks for catching!

Based on the docs for `setObject(_:forKey:cost:)`, it does seem unnecessary as you suggest (emphasis mine):

> Pass in 0 for the cost value if you otherwise have nothing useful to pass, **or simply use the setObject:forKey: method, which does not require a cost value to be passed in**.

So I'd agree this change is ok to include here in this refactor.

WordPress/Classes/Services/SuggestionService.swift
guarani: Thanks! I hadn't noticed that utility.

WordPress/Classes/Models/Suggestion.swift
guarani: That's an idea. I'll take a look and see if it's easy, if not I'll probably leave it out.

WordPress/Classes/Models/Suggestion.swift
guarani: Good question, I don't see optionality mentioned on [the WP.com REST API docs for this endpoint](https://developer.wordpress.com/docs/api/1.1/get/users/suggest/).

Based on the UI, I think that in order to properly display a suggestion, we'd need **either** `userLogin` or `displayName`, or both to be present (although if they were both `nil` the UI wouldn't look good). So my feeling is that any of these (including the image URL, can be optional).

WordPress/Classes/Models/Suggestion.swift
guarani: I think it's best to address it separately, either by porting `WPAvatarSource` to Swift or making its API more Swift-friendly. I created a ticket for me to follow up on that: https://github.com/wordpress-mobile/WordPress-iOS/issues/14996

WordPress/Classes/Models/Suggestion.swift
guarani: I've added https://github.com/wordpress-mobile/WordPress-iOS/pull/14888/commits/b0bd264e47ea70f7a019dc8e8ba90fe84c24f0bb to omit user suggestions if they are missing both a username and display name.

WordPress/Classes/Services/SuggestionService.swift
guarani: > I'll see if we can switch it back over to a struct using the approach you suggested here: #14888 (review)

It wasn't possible to make a straightforward switch over to structs, but I did get rid of some of the `@objc` attributes.

guarani:
@frosty this is ready for another look. I've managed to remove some of the @objc attributes, but couldn't move `Suggestion` over to a struct without refactoring `SuggestionTableView`.

WordPress/Classes/Services/SuggestionService.swift
```diff
@@ -8,37 +8,28 @@ extension NSNotification.Name {
...
-        init(_ value: T) {
-            self.value = value
-        }
-    }
-
-    private let suggestionsCache = NSCache<NSNumber, Wrapper<[Suggestion]>>()
+    private let suggestionsCache = NSCache<NSNumber, NSArray>()
     private var siteIDsCurrentlyBeingRequested = [NSNumber]()
 
     private static let shared = SuggestionService()
```
frosty: Is there a reason we don't just make this internal, instead of returning it in a separate function?

frosty (approved):
Looks good to me, thanks for making these changes! I just left one more small question, but otherwise :shipit:!

WordPress/Classes/Services/SuggestionService.swift
guarani: Good point! I think I kept this too close to the original implementation of Obj-C.

guarani:
> Looks good to me, thanks for making these changes! I just left one more small question, but otherwise :shipit:!

Thanks very much for the review @frosty!

guarani:
> Hey @guarani !! Still have to do a full code review, but I noticed that while the test cases worked fine on iPhone X with iOS 13.7, mentions do not seem to work on iPad Air 2, iPadOS 13.7 (both real devices, not simulators)

@Gio2018 I wasn't able to reproduce this, although I don't have an iPad to test. I tried the iPad simulator with iOS 12, 13, and 14 and in all I could @-mention.

Could you tell me what part exactly failed (e.g. reader comments, comment on post, or reply to comment, or all)?
A screen recording could also help. Thanks a lot!

Gio2018 (approved):
Hey @Guarani ! What was happening is that, when I tapped `@` on iPad, mentions did not showed up. However, I deleted and re-installed the app, and now they are showing just fine.. not sure why this is happening, but does not seem related to this PR anyway.
:shipit: !

guarani:
Hmm that's strange, hopefully it was unrelated @Gio2018

Thanks for the review!

PR #14889, opened 2020-09-14
Fix Reader Detail device rotation

leandroalonso:
Fixes rotation in the new Reader Detail.

### Without scrolling

1. Open a post in the new Reader Detail
2. Rotate the device
1. Check that the offset is kinda near the offset you were*

* Calculating the offset precisely seems very complex. I did a lot of explorations in iOS in the Safari app, Messages, and a few others and it seems that when rotating the content is not always kept in the same area.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
@@ -127,10 +127,7 @@ class ReaderDetailFeaturedImageView: UIView, NibLoadable {
     }
 
     public func deviceDidRotate() {
-        let contentOffset = scrollView?.contentOffset.y ?? 0
-        let isFeaturedImageVisible = contentOffset < 0
-
-        updateInitialHeight(resetContentOffset: isFeaturedImageVisible)
+        updateInitialHeight(resetContentOffset: false)
     }
 
     func applyTransparentNavigationBarAppearance(to navigationBar: UINavigationBar?) {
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
```diff
@@ -288,7 +288,11 @@ class ReaderDetailFeaturedImageView: UIView, NibLoadable {
 
         if let scrollView = self.scrollView {
             scrollView.contentInset = UIEdgeInsets(top: height, left: 0, bottom: 0, right: 0)
-            scrollView.setContentOffset(CGPoint(x: 0, y: -height), animated: false)
+
+            // Only reset offset if the user has scrolled just a little bit
```
emilylaguna: I'm not sure what "just a little bit" means in this context. Can we change it something a bit more descriptive?

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailFeaturedImageView.swift
```diff
@@ -288,7 +288,11 @@ class ReaderDetailFeaturedImageView: UIView, NibLoadable {
...
+            if scrollView.contentOffset.y < 0 {
```
emilylaguna: Thoughts on changing this check to a flag that's passed to the `updateInitialHeight` and move the `scrollView.contentOffset.y < 0 ` to the `deviceDidRotate` method?

emilylaguna (changes requested):
Couple of comments for you!

emilylaguna (approved):
Works for me! 👍

emilylaguna (approved):
🚢

PR #14890, opened 2020-09-14
Remove unused Reader Detail protocol conformance

frosty:
This PR fixes an issue in the release branch where the status bar was showing dark content when in a Reader Detail view:

Screenshot 2020-09-14 at 13 53 47

This was due to a leftover declaraction of `ReaderDetailViewController` conforming to `DefinesVariableStatusBarStyle`. This causes the navigation controller to query the view controller for the preferred status bar style, but this view controller no longer declares one. As such, the default style would be used.

I've also removed an unused use of `PrefersFullscreenDisplay`, which is no longer needed now that the Reader doesn't use a split view.

**To test:**

- Note: while I can see the original issue with the current test builds of the app, I was only able to recreate it building to an iOS 12.2 simulator. It wasn't happening with iOS 13.7
- Build and run, and navigate to a Reader detail view. Ensure that the status bar content remains white.
- Ideally check on both device and simulator.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -519,13 +519,3 @@ extension ReaderDetailViewController: UIViewControllerRestoration {
         return super.awakeAfter(using: aDecoder)
     }
 }
-
-// MARK: - PrefersFullscreenDisplay (iPad)
-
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
@frosty I've changed the `newNavBarAppearance` to false to test in both device and simulator, working as expected. :)

frosty:
Sorry, I forgot to add that part of the test instructions! Thanks for testing!

PR #14891, opened 2020-09-14
[TESTING] Full width columns

lukewalczak:
Fixes #

To test:

PR submission checklist:


iamthomasbishop:
@lukewalczak This is looking pretty close. Just a couple of notes:

* The margins/padding on the trailing appender and inline toolbar of the child (Column) blocks look a little bit off (I expected them to be a bit indented in the same manner as on the parent (Columns) block. (Example: [Columns](https://cloudup.com/cMMvhPQisbu) (preferred) vs. [Column](https://cloudup.com/cBfg5be0PS2))
* I'm hoping that we can force children within to utilize all of the available space, but it looks like we're probably applying the standard 580px container within them ([example](https://cloudup.com/cp5yCEFx-DC))
* Looks like there might be a 1px margin on the left/right sides of the full width parent ([screenshot](https://cloudup.com/cIV83Ul1Xdu)). really hard to see on the screenshot, but it's more obvious when it's flush against the black bezel of a device 🙈 

I still don't love the way that percentages are structured/used (which we're inheriting from the web side to some extent), but that's not really something we can fix at the moment. The good news is it sounds like there may be some work coming up on the core Columns block 🙌 . Regardless, I'm not seeing any issues with width percentages that are visible as a result of this work, so that's good news!

**Note: others should stress test this a bit to kick the tires, maybe there are things I just didn't notice. // would love @geriux's feedback on this to see if there's anything relevant that I missed or misspoke about based on his previous work.**

geriux:
> **Note: others should stress test this a bit to kick the tires, maybe there are things I just didn't notice. // would love @geriux's feedback on this to see if there's anything relevant that I missed or misspoke about based on his previous work.**

Of course! I'll test it out =)

iamthomasbishop:
@lukewalczak I'm testing again on iPhone. I think some of this is specific to the various widths, but some are harder to tell whether or not they are related. Here are my notes:

## Margins

* When I had an empty and unselected Columns block (parent), the child block borders looked a bit off. I think the left side is "proper" because on other full-width blocks we are aligning the inside block content with the edges of the container (which cuts off the left/right border typically), but as you can see on the screenshot that the child border on the right side is visible while unselected and not the Columns block is selected (as would be expected). It almost looks like the whole thing is maybe shifted to the left 1-2px or the width is just shy of 100% of the container. 🤔

| **Columns block unselected** | **Columns block selected** |
| --------------- | --------------- |
| <a href="https://user-images.githubusercontent.com/1200432/93405353-61cc3100-f852-11ea-9cf1-254901bdfe06.jpeg"></a><br>**Note**: Right side border is visible | <a href="https://user-images.githubusercontent.com/1200432/93405378-75779780-f852-11ea-829b-72a85b65d8d0.jpeg"></a> |

## Nested Columns > Column > Columns

* I tried some variations of nesting, such as Columns > Column > Columns, etc. This looked mostly fine in "normal" and wide widths, but looks a bit weird when set to full-width because the borders kind of collapse in on the block ([video](https://cloudup.com/cdh5DOAkoK3), in which I select a nested Columns block and switch from normal => wide => full and back). I think the borders in this situation (selected full-width Columns block nested in a Column) shouldn't change visibly when switching between normal/wide/full — this is how the web behaves. I almost wondered if we should hide those alignment controls on nesting, because there really is no effect in this scenario — but the web has them, so we can probably keep them, let's just apply the same bordering that is visible on normal/wide

## Trailing Inserter spacing

Probably not related to this work, but not sure — I noticed that the spacing on the trailing inserter differs between the parent and child, which isn't a huge issue but is especially obvious in portrait mode on phone. (This happens in the same way regardless of nesting level).

| **Parent selected** | **Child selected** |
| --------------- | --------------- |
| Spacing looks correct 👍  | Spacing looks off, also off-center |
| <a href="https://user-images.githubusercontent.com/1200432/93412713-503f5500-f863-11ea-82d0-fd3ed633792b.jpeg"></a> | <a href="https://user-images.githubusercontent.com/1200432/93412733-5af9ea00-f863-11ea-8e39-ed86e043b17b.jpeg"></a> |

When an empty column is selected on landscape phone, the inserter also looks mis-aligned and doesn't extend to the inner container (shown in red):

<a href="https://user-images.githubusercontent.com/1200432/93413608-3a329400-f865-11ea-9bc6-ca638f886a55.png"></a>


I'll take another look tomorrow, but I think it's pretty close 🤞 Nice work thus far!

talotabi040 (approved):

talotabi040 (approved):

iamthomasbishop:
@lukewalczak Had a chance to test the latest build on iPhone after futzing around with AppCenter for a bit because it wouldn’t let me install the build. I’m still going to test on iPad, but while I’m attempting to get that build installed, here are my notes from testing on iPhone:

- I noticed that when a nested Column block is selected — **specifically inside a full-width Columns block** — it doesn’t retain its full width ([video example](https://cloudup.com/crN8jufJhUi)), which results in a jump when selection is shifted. This only seems to happen on phone/portrait—landscape and wide-width seems fine in terms of positioning/shifting.
- The border of children (Column, additionally-nested blocks) are cut off at the safe areas when the Columns block is full-width ([examples](https://cloudup.com/c_Rk6dfgcIr) with parent/child selected). 
- If you change the width setting to `wide` while in landscape, the block seems to collapse to a single narrow column ([video](https://cloudup.com/cvARj30fDgP)). I’m having a hard time reproducing this when using more than 2 columns, so maybe it’s something specific to a two-column layout(?)
- When Column blocks are laid out horizontally, it doesn’t look like the last Column is reaching quite 100% — maybe a 1-2px gap ([example](https://cloudup.com/cbxhjgDoHfS)). It’s a tiny thing, but would be nice to resolve 😀

geriux:
Hey @lukewalczak 👋 

I was testing this a bit on my iPad and saw a small issue with the appender when the column size is changing, sometimes it overflows out of its container, here's a [video](https://cloudup.com/cHL1YLhMshE).

iamthomasbishop:
@lukewalczak This feels really solid overall, and I think the remaining things are very minor.

- (It looks like @geriux noticed this one already, but here are my notes anyway 😆). When child blocks that have appenders are selected (`Columns > **Column**`, `Columns > Column > **Columns**`, etc.), the appender is cut off on the edge of the screen or safe area ([screenshot](https://cloudup.com/cjAYxk-LGtq)). This isn't a deal-breaker but would be nice to give it some margin, like the other side of the button has)
- In the above screenshot, I also noticed that it looks like the block contents are cut off by maybe 1-2px (this is most noticeable on richtext blocks, bc characters are affected)
- There are also some _very_ (VERY) minor diffs in margins when you bounce selection between and within Columns <> Column blocks, most notably on the inline toolbar, where the icons shift up/down ever so slightly. ([video](https://cloudup.com/cs8Tc918kBL))

That's all I'm seeing at the moment, and I've tested on iPhone and iPad Pro. Nice work, as always!

twstokes:

PR #14922, opened 2020-09-16
Fix screenshot generation test

rachelmcr:
This PR fixes the failing screenshot generation test.

## Changes

I added some newer screen objects to the `WordPressScreenshotGeneration` target so they can also be used in the screenshot test.

I also removed the `dismissNotificationMessageIfNeeded()` method, which was failing because the `tap()` method requires the element to be `hittable` and for some reason that doesn’t always work. I replaced it with the `dismissNotificationAlertIfNeeded()` method that we’re using elsewhere in the UI tests, which is less flaky.

## Testing

1. Reset your chosen simulator by erasing all content and settings. (We reset the simulators before generating screenshots, so this will make sure you test in the same state.)
2. Run `rake mocks` to start the mock server.
3. Select the `WordPressScreenshotGeneration` scheme.
4. Run the `testGenerateScreenshots()` test and confirm it passes.

## PR submission checklist


startuptester (approved):
LGTM, thanks for the unblock @rachelmcr!

startuptester:
Tests ran successfully - no warnings or errors

PR #14928, opened 2020-09-17
Menus: Disable Save button if the menu selected hasn't changed

### LINKED ISSUE #9832: Menus: Don't display Save dialogue if content is unchanged
### Expected behavior
If I change content within the Menus section for my site but then change it back to the original value, I don't expect to see the Save button (since there's nothing to save).

### Actual behavior
The save button still appears.

1. Visit the Menus section in the app.
2. Change the menu in use for a section of your site to a different menu. Save button will appear in the top right-hand corner.
3. Change it back to the original menu. Save button is still present even though nothing is really set to save.

etoledom:
Fixes #9832  

This PR fixes the issue where the Save button on the Menus section was still tappable even if the selected menu was the same than the initial (saved) one. Now if the user selects the same menu as it is saved (not changing state), the Save button will be disabled.
 
menu_reset_save

To test:

- Go to the Menus section of a site.
- On the `Menu for location` section (second drop down), select a different menu.
  - Check that the `Save` button is now active.
  - Check that the `< Back` button now says `Discard`.
- Select again the menu initially selected.
  - Check that the `Save` button is now disabled.
  - Check that the `< Back` button is back.
- Smoke-test other interactions in the menu section.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Menus/MenusViewController.m
+++ b/WordPress/Classes/ViewRelated/Menus/MenusViewController.m
@@ -38,6 +38,7 @@ static CGFloat const ScrollViewOffsetAdjustmentPadding = 10.0;
 
 @property (nonatomic, strong) MenuLocation *selectedMenuLocation;
 @property (nonatomic, strong) Menu *updatedMenuForSaving;
+@property (nonatomic, strong) Menu *initialMenuSelection;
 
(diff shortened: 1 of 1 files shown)


leandroalonso (approved):
Tested and works as described. :)

etoledom:
Thank you! 🙏

PR #14929, opened 2020-09-17
Set clear backgrounds for stackviews on iOS 14

frosty:
This PR fixes some appearance issues in iOS 14. In iOS 14, stack views now render their backgrounds, which was never the case before. This led to some visual issues like this:

This PR 'fixes' that by setting all stack views to have a clear background with UIAppearance. This seems to have the desired effect in my testing, but it's possible we'll need to tweak some places individually.

**To test:**

- This PR does not include changes required to build correctly on Xcode 12. It's probably easiest to cherry-pick or just copy this change into a an existing Xcode 12 branch you may have.
- Build and run, put the app into dark mode, and navigate to the Reader.
- Check that cards appear as expected.
- Smoke test the rest of the app.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/System/WordPressAppDelegate.swift
+++ b/WordPress/Classes/System/WordPressAppDelegate.swift
@@ -830,6 +830,11 @@ extension WordPressAppDelegate {
         window?.backgroundColor = .black
         window?.tintColor = WPStyleGuide.wordPressBlue()
 
+        // iOS 14 started rendering backgrounds for stack views, when previous versions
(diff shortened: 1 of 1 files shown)


emilylaguna (approved):
Awesome, works well!  I can confirm this is working on:

- [x] iPhone 5s - 11.4
- [x] iPad Air - 12.4
- [x] iPhone 11 Pro Max - 13.3
- [x] iPhone 11 Pro - 14.0

PR #14934, opened 2020-09-17
Update syntax to support multiple trailing closures

Gio2018:
This PR fixes some syntax warnings related to calls that have more than one closure, and one of them is optional and defaulted to nil. In this case the compiler thinks that the second trailing closure should be named. There were a few ways to fix this, but I think just including the parameters explicitly is for now the more legible.

Fixes #NA

To test:

- build with Xcode 12 and make sure the following warning does not appear in the console

`Backward matching of the unlabeled trailing closure is deprecated;`

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
+++ b/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
@@ -63,7 +63,7 @@ extension UIImageView {
         with request: URLRequest,
         placeholderImage: UIImage?) {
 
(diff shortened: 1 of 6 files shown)


jkmassel (approved):
Thanks for this @Gio2018!

I took the liberty of rebasing it on the fixed CI changed in #14848, so this can be merged any time.

Gio2018:
Thanks for reviewing @jkmassel !

PR #14936, opened 2020-09-17
Fix "Topics You Might Like" interaction

leandroalonso:
Fixes the tap on the topics under the "Topics You Might Like" card.

## To test

1. Tap Reader
1. Tap Discover
1. In the "Topics You Might Like" card, tap any topic
1. The topic screen should be displayed

PR submission checklist:


emilylaguna (approved):
🚢

PR #14948, opened 2020-09-21
Update PostCategoriesViewController and WPCategoryTree to Swift 

yaelirub:
Updating `PostCategoriesViewController' to Swift as prep for implementing Categories in prepublishing bottom sheet.
 
## How to test

So:

## Existing categories
1. Open a post
2. Go to Post Settings 
3. See the list of expected categories 

## Select / Diselect Category
1. Open or create a post 
2. Go to Post Settings
3. Tap on Categories
4. Tap on unselected category
5. See the checkmark accessory 
6. Tap Back 
7. See that the Category was added to the list 
* To test unselecting category, repeat 1-3 but tap a selected category and see that the category was removed from the list.  

## Add Category
1. Open or create a post 
2. Go to Post Settings
3. Tap on Categories
4. Tap +
5. Type a category name
6. Tap Save
7. See that the Category was added to the list and is selected 

## Parent Category
1. Open or create a post 
2. Go to Post Settings
3. Tap on Categories
4. Tap +
5. Type a category name
6. Tap Parent Category
7. Change parent 
8. See selected parent in row
6. Tap Save
7. See that the Category was added to the list under the parent and is selected 

# Remove category
1. While you are in Post Settings -> Categories
1. Go to your sites wp-admin on the web
2. Go to Posts
3. Got to Categories 
4. Delete a category 
5. In the app, pull to refresh the list of categories 
6. See the list updating as expected 

Please test anything else you can think of 🙏


WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -3187,8 +3189,6 @@
 		5D577D351891360900B964C3 /* PostGeolocationView.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = PostGeolocationView.m; sourceTree = "<group>"; };
 		5D5A6E911B613CA400DAF819 /* ReaderPostCardCell.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ReaderPostCardCell.swift; sourceTree = "<group>"; };
 		5D5A6E921B613CA400DAF819 /* ReaderPostCardCell.xib */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = file.xib; path = ReaderPostCardCell.xib; sourceTree = "<group>"; };
-		5D5D0025187DA9D30027CEF6 /* PostCategoriesViewController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; lineEnding = 0; path = PostCategoriesViewController.h; sourceTree = "<group>"; xcLanguageSpecificationIdentifier = xcode.lang.objcpp; };
```
frosty, line 3190: While the references have been removed from the project, it looks like the actual files haven't been removed – they're still present in `WordPress/Classes/ViewRelated/Post/Categories/`. Can we remove those too?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+}
+
+@objc enum CategoriesSelectionMode: Int {
+    case post
+    case parent
+    case blogDefault
+}
+
+@objc class PostCategoriesViewController: UITableViewController {
+    @objc var delegate: PostCategoriesViewControllerDelegate?
```
frosty: Should this be `weak`?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+    private var selectionMode: CategoriesSelectionMode
+    private var blog: Blog
```
frosty, line 17: Could we space these out / group them, and possibly add some comments if necessary? (grouping a little like this maybe:)

```swift
@objc var delegate: PostCategoriesViewControllerDelegate?

private var blog: Blog
private var originalSelection: [PostCategory]?
private var selectionMode: CategoriesSelectionMode

private var categories = [PostCategory]()
private var categoryIndentationDict = [String: NSNumber]()
private var selectedCategories: [PostCategory]?

private var saveButtonItem: UIBarButtonItem?

private var hasSyncedCategories = false
```

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+
+    private func configureTableView() {
+        tableView.accessibilityIdentifier = "CategoriesList"
+        tableView.cellLayoutMarginsFollowReadableWidth = true
+        tableView.register(WPTableViewCell.self, forCellReuseIdentifier: Constants.categoryCellIdentifier)
+        WPStyleGuide.configureAutomaticHeightRows(for: tableView)
+    }
+
+    private func configureView() {
+        WPStyleGuide.configureColors(view: view, tableView: tableView)
```
frosty, line 62: Could we add some linebreaks to this method to separate things out a little?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        let rightBarButtonItem = UIBarButtonItem(image: UIImage(named: "icon-post-add"), style: .plain, target: self, action: #selector(showAddNewCategory))
+        switch selectionMode {
+        case .post:
+            navigationItem.rightBarButtonItem = rightBarButtonItem
+            title = NSLocalizedString("Post Categories", comment: "Title for selecting categories for a post")
+        case .parent:
+            navigationItem.rightBarButtonItem = rightBarButtonItem
+            title = NSLocalizedString("Parent Category", comment: "Title for selecting parent category of a category")
+        case .blogDefault:
+            title = NSLocalizedString("Default Category", comment: "Title for selecting a default category for a posty")
```
frosty: Typo: `posty`

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+
+    @objc private func refreshCategoriesWithInteraction() {
+        syncCategories()
+    }
+
+    @objc private func showAddNewCategory() {
+        let addCategoriesViewController = WPAddPostCategoryViewController(blog: blog)
+        addCategoriesViewController!.delegate = self
```
frosty: Could we use `if let` here or a `guard` to avoid the `!`?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        let addCategoriesNavigationControlloer = UINavigationController(rootViewController: addCategoriesViewController!)
```
frosty: Typo: `Controlloer`

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        present(addCategoriesNavigationControlloer, animated: true, completion: nil)
+    }
+
+    private func syncCategories() {
+        guard let context = blog.managedObjectContext else {
+            return
+        }
+        let service  = PostCategoryService(managedObjectContext: context)
+        service.syncCategories(for: blog, success: { [weak self] in
+            self?.reloadCategories()
```
frosty, line 103: We could use a `guard let self = self` here to reduce the optionals, or extract these calls into one method?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        let tree = WPCategoryTree(parent: nil)
+        if let sortedCategories = blog.sortedCategories() {
+            tree.getChildrenFromObjects(sortedCategories)
+        }
+
+        categories = tree.getAllObjects()
+        categories.forEach { category in
+            categoryDict[category.categoryID] = category
+            let indentationLevel = indentationLevelForCategory(parentID: category.parentID, categoryCollection: categoryDict)
+            categoryIndentationDict[category.categoryID .stringValue] = NSNumber(integerLiteral: indentationLevel)
```
frosty: Remove space here before `.stringValue`

WordPress/Classes/ViewRelated/Post/Categories/WPCategoryTree.swift
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        children.forEach {
+            allObjects.append(contentsOf: $0.getAllObjects())
+        }
+        return allObjects
+    }
+
+    private func isParentChild(category: PostCategory, parent: PostCategory?) -> Bool {
+        return parent == nil ? category.parentID == 0 : parent!.categoryID == category.parentID
```
frosty: I think this might be a little easier to understand written as:

```
let parentID = parent.categoryID ?? 0

return category.parentID = parentID
```

or

```
guard let parent = parent else {
    return category.parentID == 0
}

return category.parentID = parent.categoryID
```

WordPress/Classes/ViewRelated/Post/Categories/WPAddPostCategoryViewController.m
```diff
@@ -1,7 +1,7 @@
 #import "WPAddPostCategoryViewController.h"
 #import "Blog.h"
 #import "PostCategory.h"
-#import "PostCategoriesViewController.h"
+//#import "PostCategoriesViewController.h"
```
frosty: Can we just remove this line if we don't need it?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        // Notify the delegate of any changes for selectedCategories.
+        if didUpdateSelectedCategories {
+            delegate?.postCategoriesViewController?(self, didUpdateSelectedCategories: NSSet(array: selectedCategories!))
+        }
+
+        tableView.reloadData()
+    }
+
+    private func indentationLevelForCategory(parentID: NSNumber, categoryCollection: [NSNumber: PostCategory]) -> Int {
+        if parentID.intValue == 0 {
```
frosty: We could remove one of the returns here with something like this:

```
guard parentID != 0,
      let category = categoryCollection[parentID] else {
    return 0
}

return indentationLevelForCategory(parentID: category.parentID.intValue, categoryCollection: categoryCollection) + 1
```

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        var result = categories.count
+        if selectionMode == .parent {
+            result = result + 1
+        }
+        return result
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        guard let cell = tableView.dequeueReusableCell(withIdentifier: Constants.categoryCellIdentifier, for: indexPath) as? WPTableViewCell else {
+            return UITableViewCell()
```
frosty, line 199: It might be better to do `fatalError` here, rather than failing silently.

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,281 @@
...
+        var row = indexPath.row
+
+        // When showing this VC in mode CategoriesSelectionModeParent, we want the first item to be
+        // "No Category" and come up in red, to allow the user to select no category at all.
+        if selectionMode == .parent {
+            if row == 0 {
+                WPStyleGuide.configureTableViewDestructiveActionCell(cell)
+                cell.textLabel?.textAlignment = .natural
+                cell.textLabel?.text = NSLocalizedString("No Category", comment: "Text shown (to select no-category) in the parent-category-selection screen when creating a new category.")
+                if selectedCategories == nil {
```
frosty: Could be a little more concise with

```
cell.accessoryType = selectedCategories.isEmpty ? .checkmark : .none
```

and same for the similar accessory code below.

frosty (changes requested):
Nice to see code being converted to Swift!

I've left a number of code-level comments, as I think some extra cleanup could be done in various places. I also think there are two changes we could make which would allow a number of improvements throughout the class:

- If we make `selectedCategories` nonoptional, we can avoid a lot of unwrapping
- We can tidy things up a little by switching to `Int` for storing references to categories

I pushed a Gist here which makes these changes – let me know what you think: https://gist.github.com/frosty/dc12c641445f6ed741a034cb534395a0

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: Oh my, Yes! Thank you!

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: <3

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: <3

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: I actually prefer to leave self as weak here instead of adding the guard for both blocks. I also prefer those to be inline. Unless you have a strong preference

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: I know its more of the "Swift way" but I personally rather put the user experience first and not crash. How strongly do you feel about it?

WordPress/WordPress.xcodeproj/project.pbxproj
yaelirub: Done in 2db934aac44d3daa6d338533f88e217b945d8d2c

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: Done in 3d2b01b391c64e225d5c21458b0a3043249477b7

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: Done in 3d2b01b391c64e225d5c21458b0a3043249477b7

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: Done in 3d2b01b391c64e225d5c21458b0a3043249477b7

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: done in 3d2b01b391c64e225d5c21458b0a3043249477b7

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: done in 3d2b01b

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: done in 3d2b01b

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: done in 3d2b01b

WordPress/Classes/ViewRelated/Post/Categories/WPAddPostCategoryViewController.m
yaelirub: Done in 2db934aac44d3daa6d338533f88e217b945d8d2c

WordPress/Classes/ViewRelated/Post/Categories/WPCategoryTree.swift
yaelirub: Done in de7c8e6af0dcdf682e2541dea294bc2694e20968

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: Done in de7c8e6af0dcdf682e2541dea294bc2694e20968

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
yaelirub: Done in de7c8e6af0dcdf682e2541dea294bc2694e20968

yaelirub:
@frosty , thank you so so much for your review and suggestions! I applied them and fixed some issues.  Ready  for another look

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
frosty: Not too strongly – either way is an error and the categories view can't be used anyway 🤷‍♂️

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
frosty: 👍

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+            return UITableViewCell()
+        }
+
+        var row = indexPath.row
+
+        // When showing this VC in mode CategoriesSelectionModeParent, we want the first item to be
+        // "No Category" and come up in red, to allow the user to select no category at all.
+        if selectionMode == .parent {
+            if row == 0 {
+                WPStyleGuide.configureTableViewDestructiveActionCell(cell)
```
frosty: How about extracting this into a `configureNoCategoryRow` method?

WordPress/Classes/ViewRelated/Post/Categories/PostCategoriesViewController.swift
```diff
@@ -0,0 +1,282 @@
...
+        let indentationLevel = categoryIndentationDict[category.categoryID.intValue]
+        cell.indentationLevel = indentationLevel ?? 0
+        cell.indentationWidth = Constants.categoryCellIndentation
+        cell.textLabel?.text = category.categoryName.stringByDecodingXMLCharacters()
+        WPStyleGuide.configureTableViewCell(cell)
+        if selectedCategories.contains(category) {
+            cell.accessoryType = .checkmark
+        } else {
+            cell.accessoryType = .none
+        }
```
frosty: Likewise, we could extract this to a `configureRow(for category:)` method?

frosty (commented):
Thanks for making all the changes! I just posted two small comments about extracting two methods – see what you think :)

Otherwise, I think we're good to go!

yaelirub:
Hey @frosty , Thank you so so much for these suggestions. I updated accordingly here c7e19ff951586efa6b638adbe78e70b79af5c35f. Ready for another look 🙏

frosty (approved):
Looks great! :shipit:!

PR #14953, opened 2020-09-21
[iOS 14] Fix appearance of bottom toolbar in new photo picker

### LINKED ISSUE #14939: [iOS 14] Fix system photo picker appearance
On iOS 14, the system photo picker (presented if you choose "Select Photos..." on the photo permissions dialog) picks up some styles from the app, which don't look good. It also makes the text of the bottom button hard to read:

This affects the shipping iOS 13 build (15.6) that's currently available in the App Store.

frosty:
Fixes #14939. This PR fixes an issue in iOS 14 where the bottom toolbar of the new system photo picker is coloured blue and hard to read.

I couldn't see an easy way to target this toolbar specifically, as the view controller itself uses private classes. I opted to remove the global configuration we had for `UIToolbar`. I checked through every instance of `toolbar` we had in the app, and as far as I could tell there was only one place we didn't override this value anyway. I also tweaked the color of bar button items, as I found once instance where this was styled incorrectly.

**To test**

- Disable the `newNavBarAppearance` feature flag (you can also test with it enabled if you like)
- On iOS 14 / Xcode 12, build and run and go to Media > + and choose the 'select photos' option
- Ensure that the bottom toolbar has default styling and the text is readable
- Also check Me > App Settings > Debug > Encrypted Logs (near the bottom) and ensure the bottom toolbar is visible
- You can also check Media > tap and hold on a **video** thumbnail in your library. Check that the bottom toolbar of the video player is visible.
- Ideally also test in dark mode, and on iOS 12 as that uses different colors due to no dark mode.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
@@ -76,7 +76,7 @@ extension WPStyleGuide {
             barButtonItemAppearance.tintColor = .barButtonItemTitle
             barButtonItemAppearance.setTitleTextAttributes([NSAttributedString.Key.foregroundColor: UIColor.barButtonItemTitle], for: .normal)
 
-            let buttonBarAppearance = UIBarButtonItem.appearance()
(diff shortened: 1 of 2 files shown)


Gio2018 (approved):
Tested on Simulators: iPhone 11 Max Pro with iOS 14 (light + Dark) and iPhone SE iOS 12.4 and worked as expected.
:shipit:!

PR #14978, opened 2020-09-23
Show What's New on app update

### LINKED ISSUE #14583: Show What's New modal on app update
Add logic to show the "What's New" modal when users update the app (further details to be added)

Gio2018:
Fixes #14583

This will show the "What's New / Feature Announcements" screen on app update. The detailed behavior (for now) is:

- When users update the app, they will see the announcements on My Site
- New announcements are fetched from the backend: if the network connection is slow and users go to another screen, a few things can happen:
- If they go to another main tab (Reader or Notification) the behavior is the same: announcements will be displayed as soon as they are retrieved from the server
- If they navigate to a child level, announcements won't be displayed. When they go back to one of the main tabs (at the main level) announcements will pop up
- If they display a modal, announcements won't be shown. Announcements will be shown as soon as users navigate to another main tab (at the main level)
- Announcements will be shown once per version update, but can always be accessed via App Settings / What's New in WordPress
- Also, there is a little bit of duplication between announcements and the "streamlined navigation" alert. I disabled that in this version (it's feature flagged, so it won't affect production or internal testing)

@osullivanchris please let me know if you would like any changes to the above, thanks!!

To test:

**Normal network conditions**
1. Start with installing a version that is lower than 15.8
2. Update to this version
3. Make sure that the announcements screen appears (see examples below)

**Slow network conditions**
1. Start with installing a version that is lower than 15.8
2. Use a slow network, or simulate it with the network link conditioner
3. Update to this version
4.
  - On iPhone 
      - before the announcements appear, navigate to a child level
      - wait a few seconds, then go back to the main level and make sure the announcements appear
      - alternatively, display a modal (e.g. Me), wait a few seconds and then dismiss it. Then go to another tab (e.g. Reader) and verify that the announcements are shown)
  - On iPad (My Site is always on screen in the split view)
      - before the announcements appear, go to the Reader then on a post detail
      - wait a few seconds then go back to the main level and make sure announcements appear

**Fresh install**
- Install this version from scratch and make sure announcements do not appear.

**Feature Announcements examples**

iPhone | iPad
--- | ---
 | 

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ will be added in two next PRs.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ will be added in the next PR.


osullivanchris:
> @osullivanchris please let me know if you would like any changes to the above, thanks!!

Hey it mostly sounds great! You've really thought it through, and thanks for explaining so clearly. 


The only thing I can think to add is

> If they navigate to a child level, announcements won't be displayed. When they go back to one of the main tabs (at the main level) announcements will pop up

should we 'give up' at some point. Like if I'm doing a complex task and come up to the top level, will this feel weird? Should we not show it after like a certain amount of time, and wait for the next app launch and try again? I might be overthinking it

I tried to test this out to see if its a real concern. But I'm using a real device rather than emulator, and I didn't manage to trigger this behaviour.

Gio2018:
Thank you for reviewing @osullivanchris !

> should we 'give up' at some point. Like if I'm doing a complex task and come up to the top level, will this feel weird? Should we not show it after like a certain amount of time, and wait for the next app launch and try again?

We can do this! As discussed, let's reconnect tomorrow and see if we can install the developer tools on one or more of your devices, so you can test this out in slow network conditions, and then decide what route we want to follow.

mindgraffiti (commented):
✅ Normal network conditions
✅ Fresh install
❓ iPad slow network conditions
Do you have recommended Network Link Conditioner settings to test this? I used Edge Network on iPad, but when I navigated to the Reader, the Reader took forever to load posts, so I couldn't drill down into a post early enough, which caused "What's New" to appear. I also tried leaving the app drilled down into an individual post on 15.7 and building this branch on top of that, but it doesn't "save" my location, so that didn't work for testing either.
❓ iPhone slow network conditions - edge network
I managed to drill into a post detail view on iPhone. I waited a few seconds and tapped "My Site" to navigate to a main view. No What's New was presented.

Gio2018:
hey @mindgraffiti , thanks for reviewing!!

>❓ iPad slow network conditions [...] ❓ iPhone slow network conditions

These are kinda tricky to test. On iPhone, you can navigate to another level (e.g. stats or media) and then go back. One way to do it using the reader, could be to save for later one or more posts when you are on 15.7, then update, tap reader and then saved and then a post detail. This way it should be quicker since those posts are local. I haven't tested this scenario yet, but it could work, what do you think?

mindgraffiti (approved):
✅  iPhone slow network conditions (navigating into Stats worked great!)
🗨️  I still can't get iPad to slow down enough to present what's new after drilling down into a Reader post. I tried this on my device with an LTE cellular connection and it was *still* too fast at downloading. I'm not going to hold up this PR over testing difficulties, as your code looks correct.

osullivanchris (approved):

PR #15010, opened 2020-09-29
Add support for dark mode screenshots

rachelmcr:
Adds dark mode screenshots and updates the Fastlane `SnapshotHelper`.

### Changes

The UI tests now support setting the simulator to dark mode, and the Fastlane `screenshots` lane runs through the screenshots twice — once in dark mode and once in light mode. The screenshot test also uses a new helper method to include light/dark mode in the filename.

(These changes match the approach used in the WooCommerce iOS screenshots to support light/dark mode, h/t @jkmassel for that implementation.)

The `SnapshotHelper` update resolves an issue with the iPad landscape screenshots (part of the [2.159.0 Fastlane release](https://github.com/fastlane/fastlane/releases/tag/2.159.0)). Previously those screenshots would appear incorrectly rotated even when the simulator was in landscape orientation.

### To test

1. Open the fastlane directory: `cd Scripts/fastlane`
2. Generate the screenshots: `bundle exec fastlane screenshots` (or `bundle exec fastlane screenshots language:"en-US" ` to run it for a single language — the full run can take a long time and there aren't language-specific changes.)
3. Confirm there’s now a full set of screenshots in both light & dark mode, and iPad screenshots have the correct landscape orientation.

These generated screenshots will also come in handy for reviewing an upcoming PR (changing the app store screenshot composition).

### To review

This PR only requires a single reviewer.

### PR submission checklist

### DIFF
--- a/Scripts/fastlane/ScreenshotFastfile
+++ b/Scripts/fastlane/ScreenshotFastfile
@@ -54,12 +54,15 @@ platform :ios do
 
     puts languages
 
-    capture_ios_screenshots(
-      test_without_building: true,
-      derived_data_path: derived_data_path,
-      languages: languages,
-      clear_previous_screenshots: should_clear_previous_screenshots,
-    )
+    [true, false].each { | dark_mode_enabled |
+      capture_ios_screenshots(
+        test_without_building: true,
+        derived_data_path: derived_data_path,
+        languages: languages,
+        clear_previous_screenshots: should_clear_previous_screenshots,
+        dark_mode: dark_mode_enabled
+      )
+    }
   end
 
   #####################################################################################
--- a/Scripts/fastlane/ScreenshotFastfile
+++ b/Scripts/fastlane/ScreenshotFastfile
@@ -38,8 +38,6 @@ platform :ios do
       derived_data_path: derived_data_path,
     )
 
-    # By default, clear previous screenshots
-    should_clear_previous_screenshots = true
     languages = "da de-DE en-AU en-CA en-GB es-ES fr-FR id it ja ko no nl-NL pt-BR pt-PT ru sv th tr zh-Hans zh-Hant en-US".split(" ")
 
     # Allow creating screenshots for just one languages
@@ -47,9 +45,6 @@ platform :ios do
       languages.keep_if { |language|
         language.casecmp(options[:language]) == 0
       }
-
-      # Don't clear, because we might just be fixing one locale
-      should_clear_previous_screenshots = false
     end
 
     puts languages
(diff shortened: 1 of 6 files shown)


rachelmcr:
Let's hold off on reviewing/merging this one, so we can get the initial changes to the screenshot test done first (to support the new screenshot designs). I'll update this PR once those changes are done and we can take a look at it then.

jkmassel:
👋 We're freezing `15.9` today, so I'm bumping this PR forward to `16.0`. If you need it to be part of the `15.9` release cycle, please merge to the `release/15.9` branch and DM me – I'll be happy to cut a new beta release!

startuptester:
@jkmassel maybe I misunderstand, but I am pretty sure this does need to be in 15.9.

jkmassel:
No problem at all – just means that it needs to be merged into `release/15.9` instead of `develop` when it's ready! 😃

rachelmcr:
@startuptester this is actually ready for review now! I left it targeting `develop` because even though we plan to generate new screenshots for the 15.9 release, we don't expect any significant UI differences between the `release/15.9` and `develop` branches.

rachelmcr:
While generating a full set of screenshots I discovered that they're not all being generated — nothing is generated for `th` or `ru` because the test fails consistently in those languages. This is an issue with a full set of screenshots I generated before this branch, as well, so it isn't specific to this PR. (I didn't notice earlier because those languages don't appear at all in `screenshots.html`; you have to check each language in the `screenshots/` directory to notice some are missing.)

Since this isn't specific to this PR, I opened https://github.com/wordpress-mobile/WordPress-iOS/issues/15074 to follow up on those failures.

startuptester (commented):
Russian and Thai are empty as you noted but 
Spanish is the only language with both light _and_ dark screenshots, and the light ones are not rotated. (see attached image) 
Screen Shot 2020-10-09 at 4 33 32 PM

rachelmcr:
You ran the script in Spanish while testing https://github.com/wordpress-mobile/WordPress-iOS/pull/15055, right? Did you clear out those Spanish screenshots before testing this PR? Since this PR includes a change that doesn't automatically clear previous screenshots, those are likely copies from that previous run (which had an older version of Fastlane without the orientation fix).

startuptester:
> You ran the script in Spanish while testing #15055, right? Did you clear out those Spanish screenshots before testing this PR? 

Ah, genius! But I also ran this before with no specified language and while those screenshots are still stored, they aren't in the .html..? Just to make sure, I'll clear everything out and run again now.

startuptester (approved):
All good! 👍

PR #15011, opened 2020-09-29
Change CropViewController to 2.5.3 to remove a warning

leandroalonso:
## To test

1. Run `pod install`
2. Check that no warnings are shown

## Additional info

The `2.5.4` just fixes some SPM issues. Since we're not using it, we don't need to use this version.

PR submission checklist:

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -34,7 +34,7 @@ PODS:
   - CocoaLumberjack (3.5.2):
     - CocoaLumberjack/Core (= 3.5.2)
   - CocoaLumberjack/Core (3.5.2)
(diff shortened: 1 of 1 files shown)


emilylaguna (changes requested):
Confirmed the large warning when running pod install is gone although I do see they just released v2.5.5 that fixes this warning. Thoughts about switching to that instead? https://github.com/TimOliver/TOCropViewController/releases/tag/2.5.5

emilylaguna (approved):
Turns out the 2.5.5 doesn't fix this issue 😢 . Approving!

PR #15013, opened 2020-09-29
Reader: fix tables

### LINKED ISSUE #10333: Tables don't appear in Reader
When a table is included in a post, the content is hidden in the Reader:

The table content appears in the preview, but doesn't appear at all when opening the whole post. I couldn't replicate this in the Reader on a desktop, so I'm guessing it's an iOS issue, but please do point me in the right direction if that's not the case!

1. Create a new post that includes a table (either manually via HTML or via Gutenberg in wp-admin).
2. View post in the Reader on iOS app.
3. Table is missing!

Tested on an iPhone X running iOS 12.0, WPiOS 10.9

leandroalonso:
Fixes #10333

### To test

1. Go to App Settings -> Debug -> Turn "Ignore Reader CSS Cache" on
2. Open a post in Reader containing a table block (you can use [this one](https://charlietest12345.blog/2020/06/23/table-block/))
3. Make sure the table appears in both dark and light mode

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderCSS.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderCSS.swift
@@ -62,7 +62,7 @@ struct ReaderCSS {
     }
 
     private func url(appendingTimestamp appending: Int) -> String {
-        guard let customURL = customAddress else {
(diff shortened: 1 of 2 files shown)


emilylaguna (approved):
I can confirm the tables appear in both dark and light modes

|Light|Dark|
|:---:|:---:|
|||

Issue #15016, opened 2020-09-29
What's New - Hide Feature Announcements when needed


Gio2018:
Hide feature announcements at fresh install and when the current version does not have any announcement

PR #15020, opened 2020-09-30
What's New - Show announcements only when needed; + unit tests

### LINKED ISSUE #14984: Add unit tests for What's New
This adds unit tests to the What's New feature

Gio2018:
Fixes #14984, #15016

Feature announcements should appear when users update the app, but should not appear if:

- users do a fresh install
- the corresponding release does not have published announcements

this PR also adds unit tests

To test:

**Test case 1**
- Fresh install and make sure that no feature announcements screen (see below) appears at startup
- go to Me/App settings and make sure that the table view does not contain a row named "What's New in WordPress" (at the moment, if present, it's the last row at the bottom of the table"
- Manually change the version of the `WordPress` target to 15.9, then relaunch
- Make sure that an announcements screen like the one below appears at startup
- go to Me/App settings, make sure that the row "What's New in WordPress"
- tap on that row, and make sure you see the same screen
- Manually change the version of the `WordPress` target to 15.10 (or anything greater than 15.9), then relaunch
- Make sure that no feature announcements screen (see below) appears at startup
- go to Me/App settings and make sure that the table view does not contain a row named "What's New in WordPress"

Example screen shot
<p align=center>

</p>

**Test case 2**
- Run the `WordPressTest` test suite and make sure that all the tests succeed

PR submission checklist:

- [x] ~~I have considered adding accessibility improvements for my changes.~~ right on next PR.
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ Not yet released


frosty (approved):
Changes look good! I followed your testing steps, and ran unit tests, and everything worked as expected.

I wasn't sure if we would always want to show the last available 'what's new' at the bottom of the App Settings screen, even if it doesn't directly target the current version of the app 🤔 But I think this approach is fine for now, and we may revisit it if we list a fully version history in that location sometime in the future.

Gio2018:
Thank you for reviewing @frosty !!!

> I wasn't sure if we would always want to show the last available 'what's new' at the bottom of the App Settings screen, even if it doesn't directly target the current version of the app

That's a valid point

> we may revisit it if we list a fully version history in that location sometime in the future.

Sounds good to me. The approach I followed was to reflect what has been done on Android, but we can surely change this in the future

PR #15027, opened 2020-10-01
Reader: Fixes anchor links not responding to taps

### LINKED ISSUE #15017: Reader doesn't open anchor links correctly, or doesn't scroll to anchor
### Expected behavior
Tapping on a link that points to an anchor in a different webpage should open that link and scroll to the anchor position.

### Actual behavior
Couple of different behaviors observed:
a. Post is opened but content doesn't scroll to the anchor. That's what folks encounter in this ZD ticket: 3345628
b. Post is not opened. That's what I observed on my iPhone 6s, iOS v13.5.1, WPiOS 15.8 (and 15.6 Internal)

1. In the Reader, use the search box and put the URL of a post you know it includes an anchor link
2. Open the post
3. Tap on the anchor link
4. Either nothing happens, or the post is opened but content not scrolled to the anchor position

Example:
[Click here](https://makyo0117.wordpress.com/2019/11/18/shbdt-chuong-1/) and scroll to the bottom of the article.
Click Mục lục  to be taken to an anchor link and a list of chapter/article numbers

iPhone 6s, iOS v13.5.1, WPiOS v15.8 (link not opening case)
iPad Air (3rd gen), iOS 13.6.1 (17G80) (scroll not happening case)

emilylaguna:
Fixes #15017

#### Loading of posts with anchors
1. Open a post with a link to a post with an anchor in the reader, example URL: https://emilylaguna.wordpress.com/2020/09/30/anchor-1/
2. Tap on the link
3. If it's a WP.com post you should be directed to the post in the app, and sent to the anchor once the post loads

#### Loading of a post with internal anchors
1. Open a post that has anchors to other areas in the same post, example: https://emilylaguna.wordpress.com/2020/09/30/anchor-2/
2. Scroll to the anchor
3. Tap on it
4. You should be brought to the location in the post for that anchor

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
@@ -220,7 +220,8 @@ class ReaderDetailCoordinator {
         service.fetchPost(at: url,
                           success: { [weak self] post in
                             self?.post = post
-                            self?.renderPostAndBumpStats(with: url)
+                            self?.renderPostAndBumpStats()
+                            self?.scrollToHashIfNeeded(with: url)
                             self?.view?.show(title: post?.postTitle)
         }, failure: { [weak self] error in
             DDLogError("Error fetching post for detail: \(String(describing: error?.localizedDescription))")
@@ -229,13 +230,12 @@ class ReaderDetailCoordinator {
         })
     }
 
-    private func renderPostAndBumpStats(with url: URL? = nil) {
+    private func renderPostAndBumpStats() {
         guard let post = post else {
             return
         }
 
         view?.render(post)
-        scrollToHashIfNeeded(with: url)
 
         bumpStats()
         bumpPageViewsForPost()
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
@@ -61,6 +61,15 @@ class ReaderDetailCoordinator {
     /// If the site is an external feed (not hosted at WPcom and not using Jetpack)
     private(set) var isFeed: Bool?
 
+    /// The perma link URL for the loaded post
+    private var permaLinkURL: URL? {
+        guard let postURLString = post?.permaLink else {
+            return nil
+        }
+
+        return URL(string: postURLString)
+    }
+
     /// Initialize the Reader Detail Coordinator
     ///
     /// - Parameter service: a Reader Post Service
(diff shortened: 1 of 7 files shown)


WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
```diff
@@ -221,16 +229,32 @@ class ReaderDetailCoordinator {
         })
     }
 
-    private func renderPostAndBumpStats() {
+    private func renderPostAndBumpStats(with url: URL? = nil) {
```
leandroalonso: I was a little bit confused by the `url` as a param here.

I had to check the code to see that `bumpStats` does not rely on this URL, and was wondering why that was set in one of the places that is called but not on the other.

What do you thinnk about instead of having this logic here you just call `self?.scrollToHashIfNeeded(with: url)` in `service.fetchPost`? I guess it will avoid this confusion.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -72,6 +72,13 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
         }
     }
 
+    /// Tracks whether the webview has called -didFinish:navigation
+    private var isLoadingWebView = true
+
+    /// If the scroll(to:) has been called before the webview content has been loaded
+    /// the hash will be stored in this variable, and scrolled to once the content is loaded
+    private var pendingHashString: String? = nil
```
leandroalonso: I know this is such a simple logic (and the scroll) but I wonder if that shouldn't be in the coordinator (which is responsible for all the logic) — while the view simple do what the coordinator commands.

Something like: webView load -> tell coordinator -> coordinator tell VC to scroll if needed.

leandroalonso (changes requested):
It works as described.

Just two small suggestions though.

hypest:
👋 @emilylaguna , I removed myself from the reviewers list as I'm not currently planning to dive in, especially since I see Leandro has made a pass already. That said, let me know if you'd still want me to make some pass, although I won't be too productive on the iOS side of things :)

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
emilylaguna: Done in ee35f5cc93eecd4588b5b6ba05e35a474526e9e0

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
emilylaguna: This was changed in ee35f5cc93eecd4588b5b6ba05e35a474526e9e0

emilylaguna:
@leandroalonso back to you!

leandroalonso (approved):

PR #15028, opened 2020-10-01
Updates to the screenshot generation test

startuptester:
Fixes:
removed screenshots we no longer need and added new ones. 

To test:
make sure the test runs and the correct screenshots are taken

PR submission checklist:

- [n/a ] I have considered adding unit tests where possible.


WordPress/WordPressScreenshotGeneration/WordPressScreenshotGeneration.swift
```diff
@@ -24,6 +24,7 @@ class WordPressScreenshotGeneration: XCTestCase {
             XCUIDevice.shared.orientation = UIDeviceOrientation.portrait
         }
 
+
```
rachelmcr: Nitpick: It looks like this extra line break snuck in by accident.

WordPress/WordPressUITests/Screens/MySiteScreen.swift
```diff
@@ -83,6 +86,18 @@ class MySiteScreen: BaseScreen {
...
+        readerButton.tap()
+        return ReaderScreen()
+
+    }
+
+    func gotoEditorScreen() -> EditorScreen {
+        postsButton.tap()
+        return EditorScreen()
+
+    }
```
rachelmcr: I don't think this is being used anywhere, is it? Let's take it out unless we have a need for it.

WordPress/WordPressUITests/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -141,4 +141,14 @@ class BlockEditorScreen: BaseScreen {
...
         return XCUIApplication().navigationBars["Gutenberg Editor Navigation Bar"].buttons["Close"].exists
     }
+    @discardableResult
+    func openBlockPicker() -> BlockEditorScreen {
+            addBlockButton.tap()
+        return BlockEditorScreen()
+    }
+    @discardableResult
+    func closeBlockPicker() -> BlockEditorScreen {
+        editorCloseButton.swipeDown()
```
startuptester: This isn't ideal, but editorCloseButton was the only element on the BlockEditorScreen that wasn't located by string and these tests need to be language agnostic. Changing the other elements would have required more refactoring than we have time for in this project.

WordPress/WordPressUITests/Screens/MySiteScreen.swift
```diff
@@ -83,6 +86,18 @@ class MySiteScreen: BaseScreen {
         return ActionSheetComponent()
     }
 
+    func gotoReaderScreen() -> ReaderScreen {
+        readerButton.tap()
+        return ReaderScreen()
+
+    }
```
rachelmcr: We have a `gotoReaderScreen()` method that's part of `TabNavComponent`. Since we're using that screen object method in the test, we shouldn't need to add one specifically for this screen.

WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
```diff
@@ -46,6 +46,7 @@ extension XCTestCase {
 
         let app = XCUIApplication()
         app.launchArguments = ["-wpcom-api-base-url", WireMock.URL().absoluteString, "-no-animations"]
+        app.launchArguments = ["-no-animations"]
```
rachelmcr: Oops, looks like this extra `app.launchArguments` line was left in here. (We only want the one in the line above it that includes setting the base URL for WireMock.)

WordPress/WordPressScreenshotGeneration/WordPressScreenshotGeneration.swift
```diff
@@ -36,22 +37,22 @@ class WordPressScreenshotGeneration: XCTestCase {
...
 
-        let firstPostEditorScreenshot = postList.selectPost(withSlug: "summer-band-jam")
-        snapshot("1-PostEditor")
-        firstPostEditorScreenshot.close()
-
-        // Get a screenshot of the drafts feature
-        let secondPostEditorScreenshot = postList.selectPost(withSlug: "ideas")
-        snapshot("5-DraftEditor")
-        secondPostEditorScreenshot.close()
+        let blockPickerScreenshot = postList.selectPost(withSlug: "summer-band-jam")
```
rachelmcr: For clarity, since the block picker is opened in a separate step maybe instead of `blockPickerScreenshot` this constant could be something like `postEditor`? Then when it's used again in line 55 it's clear that it's the post editor being closed.

WordPress/WordPressUITests/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -141,4 +141,14 @@ class BlockEditorScreen: BaseScreen {
     static func isLoaded() -> Bool {
         return XCUIApplication().navigationBars["Gutenberg Editor Navigation Bar"].buttons["Close"].exists
     }
+    @discardableResult
+    func openBlockPicker() -> BlockEditorScreen {
+            addBlockButton.tap()
```
rachelmcr: Nitpick: This line has some extra indentation.

WordPress/WordPressUITests/Screens/Editor/BlockEditorScreen.swift
rachelmcr: This action was pretty flaky when I ran the test — the swipe didn't always close the picker. What if for now we try tapping on a specific coordinate outside the picker to close it?

rachelmcr (changes requested):
I ran `testGenerateScreenshots()` and the main issue at the moment is the extra launch arguments in `XCTest+Extensions.swift` (see my comment on that file) — that's preventing the tests from running against the mock server.

Right now running the tests with mocks will fail at the Reader step, but that should be resolved once we have new/updated mocks (planned for a followup PR) so it's expected for now. I'm also getting a failure due to the editor taking a surprisingly long time to open, but that also happens for me on `develop` so it isn't anything introduced in this PR. That editor issue might be something local to me but if not we can follow up on it in another PR.

My other comments are mainly cleanup and nitpicks, nothing blocking. Let me know if you have any questions about my comments here!

WordPress/WordPressUITests/Screens/Editor/BlockEditorScreen.swift
startuptester: I used the same element and set the offsets to 0 so hopefully that works more consistently? The previous method worked for me so please let know if this is just a flaky for you.

WordPress/WordPressUITests/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -141,4 +142,14 @@ class BlockEditorScreen: BaseScreen {
...
         return XCUIApplication().navigationBars["Gutenberg Editor Navigation Bar"].buttons["Close"].exists
     }
+    @discardableResult
+    func openBlockPicker() -> BlockEditorScreen {
+        addBlockButton.tap()
+        return BlockEditorScreen()
+    }
+    @discardableResult
+    func closeBlockPicker() -> BlockEditorScreen {
+        offBlockPickerCoordinate.tap()
```
rachelmcr: The test is consistently hanging for me while instantiating the `BlockEditorScreen`. Instead of defining an `offBlockPickerCoordinate` constant, which is causing it to hang, how about calculating the coordinate directly within this method like this:

```suggestion
        editorCloseButton.coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0)).tap()
```

rachelmcr:
> I'm also getting a failure due to the editor taking a surprisingly long time to open, but that also happens for me on develop so it isn't anything introduced in this PR. That editor issue might be something local to me but if not we can follow up on it in another PR.

This shouldn't require any followup. I forgot that I was proxying through Charles when I tested this before, and for some reason that massively delays the editor opening. Once I closed Charles I had no issue with this while running the test.

rachelmcr (approved):
These changes look great and checks are passing! :shipit:

PR #15030, opened 2020-10-01
Use Reader V2 posts endpoint for topics

### LINKED ISSUE #14930: Reader Improvements: Use the new v2 topic posts endpoint
### Description
The current endpoint the app uses to get the posts for a topic is: `https://public-api.wordpress.com/rest/v1.2/read/tags/{TAG}/posts`

We should switch this over to the `wpcom/v2` endpoint:  `https://public-api.wordpress.com/wpcom/v2/read/tags/posts?tags={TAG}`

### Tasks
- [x] Add the new API to WPKit
- [x] Update the ReaderStreamViewController to use the new API
- [x] Update the paging system to use the `next_page_handle` Note: the current implementation uses dates and offset counts

leandroalonso:
Fixes #14930 

### To test

1. Open Discover
2. Tap in any topic under "Topics you might like"
3. Check that the content appears correctly

Pull to refresh, scroll, etc. :)


WordPress/Classes/Services/ReaderPostService+PostsV2.swift
```diff
@@ -0,0 +1,80 @@
+import Foundation
+
+extension ReaderPostService {
+    func fetchCuratedPosts(for topic: ReaderTagTopic, isFirstPage: Bool = true, success: @escaping (Int, Bool) -> Void, failure: @escaping (Error?) -> Void) {
```
leandroalonso: Not sure about this name, tbh.

Open to suggestions. :)

WordPress/Classes/Services/ReaderPostService.h
```diff
@@ -9,6 +9,9 @@ extern NSString * const ReaderPostServiceToggleSiteFollowingState;
 
 @interface ReaderPostService : LocalCoreDataService
 
+@property (nonatomic) NSString *nextPageHandle;
```
emilylaguna: We should copy the string. 

```suggestion
@property (copy, nonatomic) NSString *nextPageHandle;
```

WordPress/Classes/Services/ReaderPostService+PostsV2.swift
emilylaguna: What about this? I think curated gives the wrong impression

```suggestion
    func fetchPostsV2(for topic: ReaderTagTopic, isFirstPage: Bool = true, success: @escaping (Int, Bool) -> Void, failure: @escaping (Error?) -> Void) {
```

emilylaguna (changes requested):
Works really well, a couple comments below.

WordPress/Classes/Services/ReaderPostService+PostsV2.swift
leandroalonso: Done in e4ba1f857a07c722c352f282dacbfa05a949e905

WordPress/Classes/Services/ReaderPostService.h
leandroalonso: Done in 57305aaf43f93709228d56bd60c7a767f7f6f0cf

leandroalonso:
@emilylaguna ready for another review! :)

emilylaguna (approved):
This works well for me! 👍

PR #15041, opened 2020-10-02
Revert CropViewController manual update

leandroalonso:
This PR removes the manual bump (from 2.5.3 to 2.5.4) in CropViewController that produces the huge warnings when running `bundle exec pod install`.


emilylaguna (approved):
Let's try this again!

PR #15048, opened 2020-10-05
Moves settings button to the navigation bar from the tab bar

### LINKED ISSUE #14835: Move settings icon from following tab to reader app nav bar
The settings icon can currently only be accessed from the filter bar in the following tab. Let's move the settings icon to the app nav bar next to the search icon so it's a global action that can be accessed from any tab (following, discover, likes, saved). This increases the discoverability of the "manage" tab and allows users to manage their followed topics & sites from anywhere on the Reader.

Screen Shot 2020-09-03 at 4 26 55 PM

emilylaguna:
Fixes #14835

### Screenshots
| iPhone 11.4 | iPad 12.4 | iPhone 14.0 |
|:---:|:---:|:---:|
||||

### To test:
1. Launch the app
2. Tap on the Reader tab
3. Tap on the Settings icon
4. Go through the flows of the manage screen, unfollowing, following, etc.

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -7,8 +7,6 @@ class ReaderTabView: UIView {
     private let tabBar: FilterTabBar
     private let filterButton: PostMetaButton
     private let resetFilterButton: UIButton
-    private let settingsButton: UIButton
-    private let verticalDivider: UIView
     private let horizontalDivider: UIView
     private let containerView: UIView
     private var loadingView: UIView?
@@ -22,8 +20,6 @@ class ReaderTabView: UIView {
         tabBar = FilterTabBar()
         filterButton = PostMetaButton(type: .custom)
         resetFilterButton = UIButton(type: .custom)
-        settingsButton = UIButton(type: .custom)
-        verticalDivider = UIView()
         horizontalDivider = UIView()
         containerView = UIView()
 
@@ -76,9 +72,7 @@ extension ReaderTabView {
         setupButtonsView()
         setupFilterButton()
         setupResetFilterButton()
-        setupVerticalDivider(verticalDivider)
         setupHorizontalDivider(horizontalDivider)
-        setupSettingsButton()
         activateConstraints()
     }
 
@@ -105,8 +99,6 @@ extension ReaderTabView {
         }
         buttonsStackView.isHidden = tabItem.shouldHideButtonsView
         horizontalDivider.isHidden = tabItem.shouldHideButtonsView
-        settingsButton.isHidden = tabItem.shouldHideSettingsButton
-        verticalDivider.isHidden = tabItem.shouldHideSettingsButton
     }
 
     private func setupButtonsView() {
@@ -116,8 +108,6 @@ extension ReaderTabView {
         buttonsStackView.alignment = .fill
         buttonsStackView.addArrangedSubview(filterButton)
         buttonsStackView.addArrangedSubview(resetFilterButton)
-        buttonsStackView.addArrangedSubview(verticalDivider)
-        buttonsStackView.addArrangedSubview(settingsButton)
         buttonsStackView.isHidden = true
     }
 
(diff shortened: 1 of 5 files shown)


Gio2018:
Thanks for working on this @emilylaguna !! Will review asap. Also adding @osullivanchris to this PR as we worked on this screen together, and he might have some useful hints.

jkmassel:
👋 We're freezing `15.9` today, so I'm bumping this PR forward to `16.0`. If you need it to be part of the `15.9` release cycle, please merge to the `release/15.9` branch and DM me – I'll be happy to cut a new beta release!

Gio2018 (commented):
Hey @emilylaguna , the code looks great and it works as described.

I had one question: on smaller iPhones, the settings and search buttons, although separated, still feel somewhat close to each other (and this might just be because I have bigger hands), but I was wondering if it would make sense to put one button on the left and one on the right instead of both on the right? Just a thought.

emilylaguna:
Hey @Gio2018 thanks for the review. Since I'm not the designer of the task I'll ask @mbshakti for her thoughts on this before proceeding.

mbshakti:
Good point on the tap target space @Gio2018, the only thing is that we are using a similar icon/top bar pattern on the new reader detail view, so this same issue might come up there.... 

I think it makes sense to have one button on the right and left @osullivanchris what do you think given all your work on the white headers projects? 

In term of which icon is positioned on which side, I noticed that the settings icon is on the right side on the notifications tab, maybe we should keep the settings icon position consistent for the Reader tab as well by having it placed on the right side. And move the search icon to the left side.

osullivanchris:
👋  Hey I think this is a good idea to fix. When I did usability testing, the settings icon and managing what you follow was a weak point. I actually think the cog icon is unclear as well as the position of it.

I am not sure if it should be omnipresent across tabs as suggested in #14835 , as it is about managing what you _follow_. But I can kinda see with the direction the discover tab is going that maybe it makes sense.

For the position, 2 icons top right is usually manageable. Its a bit more common on Android. I think putting the cog on the left would work fine currently. But it wont work when we bring in the left aligned large titles.

Ended up diving into this for a bit and thinking of some alternatives as well. I think search is part of the problem as well. I don't think the search feature is well framed or explained. I don't know why I should search or for what based on the position. It looks like it will search within the page or something.

**Option 1: single icon with action sheet**

Here, there is an icon, or it could be a text action. It brings up an action sheet. The action sheet allows us to show a longer label, making it clearer where you are going and what you are to do. The actions would bring the user to the existing location.

IMG_20201007_122356

**Option 2: single icon but shared space**

Similar to the first option, we have one access point ('Manage Feed' could be improved!). But rather than splitting the user off with an action sheet, we could provide the searching and managing functionality in one space. 

I tried two options on the right. In one, the search and managing stuff is all available immediately. On the sketch further to the right, its a menu, which kinda lets us explain/frame the options a bit better so that you know what you're getting into.

IMG_20201007_122333

**Option 3: Manage/Follow from Discover tab**

Another theory...what if 'Following' was just about content. And 'Discover' was the place you went for everything else. Its like following is your bookshelf in the house, and discover is going to the book store. You can get inspired by stuff on the tables as you walk in, or you can search for particular things on the shelves.

In the sketch below I've added an entry point to search at the top of discover, and a way to manage your tags/sites underneath the suggested ones. But there's a lot of ways you could do this, if you put this functionality into discover. Its a big departure though so it would need some work.

IMG_20201007_122423

Option 4: Keep two actions top right
If we don't want to shift so much. I don't think splitting the actions is gonna work due to large titles. See example in the sketch below. I also think we should change the cog to something else like a text label or a 'more' icon. 

IMG_20201007_122410


This last option would be my MVP/quick fix solution. But hope you can see there is a bigger problem here with searching and managing, that maybe we could go deeper on too. At the moment its totally unclear to me why I would search and what the cog icon is going to do.

mbshakti:
Had a chat on slack with @osullivanchris -- For now, let's keep it as is (as per @emilylaguna's screenshot here). We can't move one icon to the right & and the other left because it won't work with the white headers. We can think through different entry points with more intentionality when we revisit the manage topics/site experience. Thanks for the discussion, all! Feedback noted.

Gio2018 (approved):
:shipit: !!

osullivanchris (approved):
Nice one @mbshakti  thanks for prioritising. Agree its fine for now. But glad to have captured some ideas, as I think we can make further improvements to how the user finds sites/posts and manages what they follow.

emilylaguna:
Awesome, thanks everyone! Merged!

PR #14274, opened 2020-06-08
Use correct analytics event for save draft action

guarani:
`PostEditorStateContext`'s `action` property was not holding the correct value for a save draft operation. I've updated it to be `.save` instead of `.publish`.

Fixes #13729

### Testing steps
Taken from main issue:
1. Create a new Page or Post
2. Make a few edits Set title and content
3. Select "X" in the top left-hand corner
Expect: To see a dialog with "You have unsaved changes." and "Save Draft"
4. Select Save
5. The editor should fire the analytics event `editor_draft_saved`

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #13729, opened 2020-03-25
Analytics event `editor_draft_saved` fires as `editor_post_published`


chipsnyder:
### Expected behavior
After creating a new page or post and backing out via the X and save as draft flow, the editor should fire the analytics event `editor_draft_saved`

### Actual behavior

Editor fires the analytics event `editor_post_published`

### Steps to reproduce the behavior

1. Create a new Page or Post
2. Make a few edits Set title and content
3. Select "X" in the top left-hand corner
     - **Expect:** To see a dialog with "You have unsaved changes." and "Save Draft"
4. Select Save

**Expected** The editor should fire the analytics event `editor_draft_saved`
**Actual** The editor fires the analytics event `editor_post_published`

### Additional info 
This was discovered as part of https://github.com/wordpress-mobile/WordPress-iOS/issues/13527. We might want to combine this work with another ticket to adjust the way we fire the draft quicksave as well https://github.com/wordpress-mobile/WordPress-iOS/issues/13527#issuecomment-603743549

elibud:
@hypest @pinarol  can you please have someone on your team to look into this as it may be affecting your analysis?

guarani:
Thanks for reporting, @chipsnyder! I can reproduce it consistently.

The method that performs the save in this scenario is `publishPost(action:dismissWhenDone:analyticsStat:)` and is called here: 
https://github.com/wordpress-mobile/WordPress-iOS/blob/d798535dc663d24d3231caef236791f322c0c82e/WordPress/Classes/ViewRelated/Post/PostEditor%2BPublish.swift#L309

Notice that `action` is passed in explicitly _and_ is also a property of `postEditorStateContext` – this seems like two sources of truth for the action the user is performing, which is odd.

Going back to this scenario, `action == .saveAsDraft` while `self.postEditorStateContext.action == .publish`. I'd say the former is correct but the latter is wrong – both should be `.saveAsDraft`. The reason it's `publish` in this scenario seems to be because `publish` is the default and is never updated. See: 
https://github.com/wordpress-mobile/WordPress-iOS/blob/e3cd86b3cf984ff576bbef61a17df38fa7f06b20/WordPress/Classes/ViewRelated/Post/PostEditorState.swift#L152

One solution is to directly update `self.postEditorStateContext.action` to `saveAsDraft` when the "Save Draft" button is pressed. I then noticed that the mapping from action to event is not correct, I'm probably looking for `.save` instead:
[...]

PR #15050, opened 2020-10-05
Use correct analytics event for save draft action

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/WordPress-iOS/issues/13729

When selecting the ✖️ then selecting save as draft the action that was being sent along to `publishPost` was being sent as `save` however the analytics stat was being sent as `publish` to get the desired outcome this should be `save`. Updated the `postEditorStateContext` action to match the change from the user's selection. 

This also affects when a user opens an existing post and updates. The flow for `update` and `save` appear to be the same in `postEditorStateContext` so I added that metric to the list as well.

🗒️ This is another take on the research and work that @guarani did as part of https://github.com/wordpress-mobile/WordPress-iOS/pull/14274

### To test:

An easy way to test for these events is to run the app locally through XCode and filter the console to look for the string `🔵 Tracked:`

1. Create a new Page or Post
1. Make a few edits Set title and content
1. Select "X" in the top left-hand corner
     - **Expect:** To see a dialog with "You have unsaved changes." and "Save Draft"
1. Select Save

**Expect** The editor should fire the analytics event `editor_draft_saved`

---
#### Updates via ✖️ 
1. Open an existing draft
1. Make an edit
1. Select the ✖️ to back out
1. Select "Update Draft"
**Expect** The editor should fire the analytics event `editor_post_update`

#### Create a draft and update via ✖️ 
1. Create a new page or post
1. Make an edit
1. Select `...` in the top right corner
1. Select "Save as Draft"
    - The editor should fire the analytics event `editor_quick_draft_saved`
1. Make another edit
1. Select the ✖️ to back out
1. Select "Update Draft"
**Expect** The editor should fire the analytics event `editor_post_update`

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
@@ -308,7 +308,9 @@ extension PostEditor where Self: UIViewController {
 
             // The post is a local or remote draft
             alertController.addDefaultActionWithTitle(title) { _ in
-                self.publishPost(action: self.editorAction(), dismissWhenDone: true, analyticsStat: self.postEditorStateContext.publishActionAnalyticsStat)
+                let action = self.editorAction()
+                self.postEditorStateContext.action = action
+                self.publishPost(action: action, dismissWhenDone: true, analyticsStat: self.postEditorStateContext.publishActionAnalyticsStat)
             }
         }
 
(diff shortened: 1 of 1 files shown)


guarani (approved):
Nice work, thanks a lot for picking this up!
I tested the scenarios here and they worked fine, the code looks great.

Issue #14547, opened 2020-07-30
[iPad] Initial states for split view panes don't display correctly


diegoreymendez:
When logging into the app fresh on an iPad, there are multiple places where the initial state of the split views require optimization:

- My Sites: When no sites are present, no results view is shown in a small left pane.
- My Sites: When no sites are present, right pane background doesn't fill space (this can be fixed by resolving the above issue to fill the space with the no results view)
- Notifications: For a Jetpack site, the Jetpack prompt displays in a small left pane.
- Notifications: When no notifications are present, no results view is shown in a small left pane and the right pane shows gray.

### Original issue

After creating a new account without adding or creating a site, this is how the "My Sites" tab looks like.

<img width="636" alt="Screen Shot 2020-07-30 at 12 00 24" src="https://user-images.githubusercontent.com/1836005/88939653-27d0ac80-d25d-11ea-910d-a838f2a09073.png">

diegoreymendez:
Fyi @frosty, @rachelmcr

designsimply:
Tested and confirmed that the My Sites screen on iPad for an account with no sites displays the "Add new site" prompt in the sidebar on the left which looks a bit odd on iPad screens.

Portrait|Landscape
---|---
![IMG_0503](https://user-images.githubusercontent.com/1119271/89081924-c89c9600-d349-11ea-9f8c-f3b6efdb06f9.PNG)|![IMG_0504](https://user-images.githubusercontent.com/1119271/89081923-c6d2d280-d349-11ea-9048-080e254341ee.PNG)

<sup>Tested with WPiOS 15.4.0.1 TestFlight beta on iPad Mini 4 (WiFi) (iPad5,1) iPadOS 13.6.</sup>

diegoreymendez:
Thanks @designsimply and apologies I didn't loop you in!

designsimply:
While beta testing the Unified Login & Signup call for testing with WPiOS 15.7.0.1, I noticed that the Notifications > Install Jetpack prompt also displays small and to the left in contrast to a similar prompt in Stats where it's shown on the right.

"Install Jetpack" prompt in Notifications|"Install Jetpack" prompt in Stats
---|---
[...]

PR #15053, opened 2020-10-05
My Sites > no sites: show NRV on rotation

### LINKED ISSUE #15019: No Sites view: view disappears when rotating
- On an iPhone that switches between single and split view (ex: iPhone 11 Pro Max), login with an account that has no sites.
- Go to the My Site tab.
- Rotate the device.
- Portrait mode is correct, but in landscape the view disappears.

no_site_rotate

For reference, on an iPhone _without_ split view (ex: iPhone 8) the view rotates correctly:

rotate_single_view

ScoutHarris:
Fixes #15019 

This fixes an issue where the no sites view would disappear on rotation.

Details:
The problem was caused by the empty details view being rendered on top of the primary view (which displays the No Results View). Initially, I tried setting the details VC to `nil`. However, the `collapseSecondary` method crashed because there was no detail VC. So instead, when the view is split in the `separateSecondaryFrom` method, the details is just not shown.

To test:
- On an iPhone that switches between single and split view (ex: iPhone 11 Pro Max), login with an account that has no sites.
- Go to the My Site tab.
- Rotate the device.
- Verify the view can be seen in landscape. (The image is hidden in landscape, as it's meant to be.)

To note, this _only_ fixes the NRV "disappearing". You'll notice if you start the app in landscape, the view is incorrectly split, and the NRV appears on the left. This is related to https://github.com/wordpress-mobile/WordPress-iOS/issues/14547#issuecomment-702217917 and will be resolved separately.

nrv_rotation

PR submission checklist:


Gio2018 (approved):
Tested on iPhone 11 Pro Max and worked as expected.

:shipit:!!

PR #15057, opened 2020-10-06
Update LayoutChanged notification

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/WordPress-iOS/issues/12881#issuecomment-704332412

After expanding the menu item the app fires the accessibility notification `UIAccessibilityLayoutChangedNotification` with the parameter `[self.stackView.arrangedSubviews firstObject]`. From my testing, this seems to do two things.
1. Lets the app know that some of the accessible items have changed.
1. Schedules an event to update the focused item back to the first item of the stack view. In these cases, it's the header with the expansion button.

The first part seems to happen right away because I'm able to swipe to the new nested items. The second step seems delayed though because I'm able to swipe to a new item but then my focus is reverted.

Removing the second parameter allows part 1 (update the list of accessible elements) to still occur but prevents the focus change.

## To Test

1. In the WordPress app, go to Menus.
1. Highlight a menu 
1. Double-tap to expand the areas option
1. While VoiceOver is still speaking the first couple of words ~"Menu area primary" swipe right to navigate to the next item
- **Expect** your focus to not be reverted from the selected item.

| Before | After |
| --- | --- |
| <kbd>Voice-Over 2020-10-06 11_00_14</kbd> | <kbd>VoiceOver-after 2020-10-06 14_22_44</kbd> | 

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Menus/MenusSelectionView.m
+++ b/WordPress/Classes/ViewRelated/Menus/MenusSelectionView.m
@@ -128,8 +128,7 @@
 
         self.detailView.showsDesignActive = selectionItemsExpanded;
         [self prepareForVoiceOver];
-        UIAccessibilityPostNotification(UIAccessibilityLayoutChangedNotification,
-                                        [self.stackView.arrangedSubviews firstObject]);
+        UIAccessibilityPostNotification(UIAccessibilityLayoutChangedNotification, nil);
     }
 }
 


etoledom (approved):
This was done this way so VoiceOver would go back to the corresponding cell after selecting a different element in the list.

I do agree that the behaviour when opening the list can be more annoying than the issue this was solving, so let's go with it.

Thank you @chipsnyder !

chipsnyder:
> This was done this way so VoiceOver would go back to the corresponding cell after selecting a different element in the list.

Thank you for adding more details to the history of this, and for taking the time to review.

PR #15058, opened 2020-10-06
Notifications: Don't dim the detail view when the filtered NRV is displayed.

ScoutHarris:
Ref https://github.com/wordpress-mobile/WordPress-iOS/issues/14547

This removes the dimmed view when there are no Notifications to show for a selected filter.

Details:
I tried simply not showing the dimmed view. Without it, there is no detail view, so `WPSplitViewController` automatically creates one. Which was fine on iPad. However, on an iPhone that has single and split view, `collapseSecondary` adds it to the top of the view stack, covering the primary view. Not so fine. `WPSplitViewController` has behavior tied to the existence of the dimmed view. We want that behavior, but we don't want the view dimmed. So the solution ended up being to show the dimmed view with an alpha of 0. 

To test:
- Run the app on an iPad or an iPhone that has single and split view (ex: iPhone 11 Pro Max).
- Go to the Notifications tab.
- Select a filter that has no Notifications.
- Verify the detail view is no longer dimmed.

| Before | After |
|--------|-------|
| before | after |
| iphone_before | iphone_after |

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -1275,6 +1275,10 @@ private extension NotificationsViewController {
         if splitViewController.wpPrimaryColumnWidth != columnWidth {
             splitViewController.wpPrimaryColumnWidth = columnWidth
         }
+
+        if columnWidth == .default {
+            splitViewController.dimDetailViewController(shouldDimDetailViewController, withAlpha: WPAlphaZero)
+        }
     }
 
     var noConnectionTitleText: String {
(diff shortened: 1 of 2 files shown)


frosty (commented):
Thanks for looking into this! This fixes the dimming issue. However, it seems like just removing the call to `dimDetailViewController` has the same effect? (I might have missed something, though!)

ScoutHarris:
Hey @frosty 

>  it seems like just removing the call to dimDetailViewController has the same effect? 

Sorry I wasn't clear. That's what I meant by:

> I tried simply not showing the dimmed view. Without it, there is no detail view, so `WPSplitViewController` automatically creates one. Which was fine on iPad. However, on an iPhone that has single and split view, `collapseSecondary` adds it to the top of the view stack, covering the primary view.

If there's a solution _I'm_ missing, I'll take suggestions. But just not showing the dimmed view didn't work.

Gio2018 (approved):
Tested on iPhone 11 Pro Max 13.7 and iPad Pro 11 inch 2nd Gen 13.7 and worked as expected!

:shipit:!!

ScoutHarris:
Thanks @Gio2018 !

frosty (commented):
Sorry, I misunderstood! In that case, no problem :)

PR #15059, opened 2020-10-06
Bug fix to stock photo selection

### LINKED ISSUE #14646: Unable to deselect images uploaded using the free photo library
### Expected behavior
Each new instance of the free photo library picker should be reset with no images selected. 

### Actual behavior
After uploading a set of images using the free photo library, the free photo library picker keeps the images from the previous set selected with no way to deselect them. 

1) Open media library, tap + to add media
2) Select "Free Photo Library"
3) Search and select media to upload, upload media
4) After media is uploaded, select + to add more media
5) Select "Free Photo Library" 
6) The previously uploaded images are shown in the preview option and the other action is to add the previously uploaded media with no way to deselect or reset it 

Here's a video of the behavior: https://photos.app.goo.gl/rChC2gpz6QCVB8vVA

kduncanwelke:
Fixes #14646 

To test: Go to the Media Library and select + to add media. Choose the Free Photo Library. Search for any photo and upload it, then return to search the free photos again. Note that no items are selected, and when you search for media and upload another image, the previous image is not uploaded in its place, as before.

PR submission checklist:

I've updated the branch so this should merge ok now. @guarani would you be able to take a look, since you were checking over the initial PR for me?

### DIFF
--- a/WordPress/Classes/ViewRelated/Media/MediaLibraryMediaPickingCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Media/MediaLibraryMediaPickingCoordinator.swift
@@ -7,14 +7,13 @@ final class MediaLibraryMediaPickingCoordinator {
     private weak var delegate: PickersDelegate?
     private var tenor: TenorPicker?
 
-    private let stockPhotos = StockPhotosPicker()
+    private var stockPhotos: StockPhotosPicker?
     private let cameraCapture = CameraCaptureCoordinator()
     private let mediaLibrary = MediaLibraryPicker()
 
     init(delegate: PickersDelegate) {
         self.delegate = delegate
 
-        stockPhotos.delegate = delegate
         mediaLibrary.delegate = delegate
     }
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Media/MediaLibraryMediaPickingCoordinator.swift
```diff
@@ -90,7 +89,12 @@ final class MediaLibraryMediaPickingCoordinator {
     }
 
     private func showStockPhotos(origin: UIViewController, blog: Blog) {
-        stockPhotos.presentPicker(origin: origin, blog: blog)
+        let picker = StockPhotosPicker()
+        // add delegate conformance, allow release of picker in the same manner as the tenor picker
+        // in order to prevent duplicated uploads and botched de-selection on second upload
```
frosty, line 94:
```suggestion
```

frosty: I think we don't necessarily need the comment in this instance as it's specific to the bug we're fixing :)

frosty (approved):
Thank you for fixing this issue! Your changes look great, and in testing I can see that the issue is fixed.

WordPress/Classes/ViewRelated/Media/MediaLibraryMediaPickingCoordinator.swift
frosty: Hmm, I was hoping I could commit this change via Github, but I don't think I can because it's on a fork. No worries, I'll go ahead and merge as-is.

Issue #11684, opened 2019-05-13
Auto save doesn't work as expected for Jetpack sites Blocked by: Automattic/wp-calypso#20265


yaelirub:
This issue came up in code review for the implementation of [Auto save published post upon preview and show preview url](https://github.com/wordpress-mobile/WordPress-iOS/pull/11638) and will be visible in 12.5
We assume that something is not working correctly with the autosave endpoint for Jetpack sites. (possibly related to: https://github.com/Automattic/wp-calypso/issues/20265)
Per @aerych's comment:

> This time round when testing I included a self-hosted site connected via Jetpack. Previewing drafts seemed to work correctly but when previewing changes to an already published post the changes were not appearing for me in the preview. Not sure if this is a case were we need to include a force=wpcom parameter to query the cache site or if its just a limitation with this sort of site and we'd want to show the "no preview available" screen

### Expected behavior
When editing a published post and then choosing to preview it, the preview should show the post with the latest changes.

### Actual behavior
For Jetpack sites, the post preview doesn't show the latest changes.

### Steps to reproduce the behavior
1. Open a published post
2. Make changes
3. Tap preview
4. See the post as published without changes 

##### Tested on [device], iOS [version], WPiOS [version]
SImulator iPhone Xs 12.2 and device iPhone 7 12.2

designsimply:
I tested 3 sites using WP Internal 12.4.0.20190506 on iPhone 6S iOS 12.2:

* WP 5.2 Jetpack 7.3.0 `architectural-worm.jurassic.ninja`
* WP 5.1.1 Jetpack 7.2.1 `serious-snake.jurassic.ninja`
* WP 5.2 Self-hosted `open-bear.jurassic.ninja`

And I was able to successfully preview changes made to published posts in all three sites with both the Aztec editor and the block editor, however, the preview looks different depending on whether or not an update has been made tot he post. ([4m46s](https://cloudup.com/cpfYo99xbZv))

| Preview before making a change | Preview after making a change |
|:---:|:-----:|
| <img src="https://user-images.githubusercontent.com/1119271/57658709-1b4f2e00-759d-11e9-8c7e-806d2c641cc1.png" width="270" alt="preview-before-update"> | <img src="https://user-images.githubusercontent.com/1119271/57658895-a6302880-759d-11e9-9aff-9e29c9e56969.png" width="270" alt="preview-after-update"> |

Possibly related: https://github.com/WordPress/gutenberg/issues/13232

[...]

PR #15061, opened 2020-10-07
Previews: Fix for previewing posts on sites with mapped domains.

### LINKED ISSUE #14220: Editor: cannot preview drafts on a WordPress.com site with a custom domain

aerych:
Fixes #14220 

This PR adds new logic to preview URLs to ensure that previewing on a site with a mapped domain will use the site's unmapped URL rather than the mapped URL.  This will let our preview browser use existing auth cookies rather than having to make a complicated series of requests to obtain an authenticated session.

@ScoutHarris could I trouble you for a review?  Also ccing @guarani who has looked into this previously.

** Merge after landing related changes to WordPressKit.

To test:
Prep:
- Start from a fresh install so there are no cookies lingering in the cookie jar.   For an extra level of paranoia, delete and reinstall for each scenario.

Scenario 1: Mapped Domain Draft Preview
- [ ] Create a new draft post on a wpcom site with a mapped domain. 
- [ ] Preview the draft. 
- [ ] Confirm you see the remote preview and not an error.

Scenario 2: Mapped Domain Published Preview
- [ ] Edit an already published post on a wpcom site with a mapped domain. 
- [ ] Preview your changes. 
- [ ] Confirm you see the remote preview and not an error.

Scenario 3: Unmapped Site Draft Preview
- [ ] Repeat Scenario 1 for this site type. Confirm there is no regression.

Scenario 4: Unmapped Published Preview
- [ ] Repeat Scenario 2 for this site type. Confirm there is no regression.

Scenario 5: Jetpack Site Draft Preview
- [ ] Repeat Scenario 1 for this site type. Confirm there is no regression.

Scenario 6: Jeptack Site Published Preview
- [ ] Repeat Scenario 2 for this site type. Confirm there is no regression.

Scenario 7: Self-hosted Site Draft Preview
- [ ] Repeat Scenario 1 for this site type. Confirm there is no regression.

PR submission checklist:


WordPress/Classes/ViewRelated/Post/Preview/PreviewNonceHandler.swift
```diff
@@ -22,4 +25,26 @@ struct PreviewNonceHandler {
         components.queryItems = queryItems
         return components.url
     }
+
+    private static func unmapURL(post: AbstractPost, url: URL) -> URL? {
+        guard
+            let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
```
ScoutHarris: Suggestion - if this was a `var` you wouldn't need `comps` below.

ScoutHarris:
Hey @aerych .

Here are my testing results.

**Successes:**
✅ Scenario 1: Mapped Domain Draft Preview 
✅ Scenario 3: Unmapped Site Draft Preview
✅ Scenario 4: Unmapped Published Preview
✅ Scenario 5: Jetpack Site Draft Preview
✅ Scenario 7: Self-hosted Site Draft Preview


**Failures:**
🚫 Scenario 2: Mapped Domain Published Preview - the preview doesn't show edits until after the post is saved.

Both of these behaved the same:
🚫 Scenario 6: Jetpack Site Published Preview - if it matters, I was using a self-hosted site with Jetpack installed. (via jurassic.ninja)
🚫 Not mentioned: Self-hosted Published Preview - I was using a self-hosted site without Jetpack. (via jurassic.ninja)

- On a published post > Edit > Preview > shows preview successfully.
- On a published post > Edit > edit the post > Preview > shows the `Preview Unavailable` error. Also, this was in the logs:

`Error while trying to save post before preview: Optional(Error Domain=PostServiceErrorDomain Code=0 "Previews are unavailable for this kind of post." UserInfo={NSLocalizedDescription=Previews are unavailable for this kind of post.})`

ScoutHarris (changes requested):
See testing results.

WordPress/Classes/ViewRelated/Post/Preview/PreviewNonceHandler.swift
aerych: Oh ha! I never realized `guard var` was allowed! TIL.  Thanks!

aerych:
Thank you @ScoutHarris !

> Scenario 2: Mapped Domain Published Preview - the preview doesn't show edits until after the post is saved

For this one I found two scenarios where the preview wouldn't show.  One was when proxied 😬.  The other depended on the site's permalink format.  I've pushed an update that should take care of this case. Would you mind trying again?

> Scenario 6: Jetpack Site Published Preview - if it matters, I was using a self-hosted site with Jetpack installed. (via jurassic.ninja)

Arg! I even [had a recording](https://github.com/wordpress-mobile/WordPress-iOS/issues/11684#issuecomment-704665842) of this working. :(   Sadly, it looks like I was [testing this under special circumstances](https://github.com/wordpress-mobile/WordPress-iOS/issues/11684#issuecomment-705207599) so my expectations were off.   The preview is _not_ expected to be shown in this case so, what you saw was probably expected and not a regression, but please confirm!

 > Not mentioned: Self-hosted Published Preview - I was using a self-hosted site without Jetpack

Yeah. I omitted this because these previews were not expected to work in this scenario. (No autosave post type.) :)

Ready for another peek!

ScoutHarris:
Hey @aerych .

> For this one I found two scenarios where the preview wouldn't show. One was when proxied 😬. The other depended on the site's permalink format. I've pushed an update that should take care of this case. Would you mind trying again?

I'm not sure what's wrong but my .blog site doesn't show unsaved edits in the preview. I tried on a real device, so I wasn't proxied.

| post_edit | post_preview |
|--------|-------|

aerych:
Hmm... I wonder if we might be taking different steps?  This is what I'm doing:
preview-published
Lemme know if you spot a difference and I'll try it and see if I can repo what you're seeing.

aerych:
Just in case using a device mattered, I repeated the above test on my iPhone 11 Pro (iOS 13.7) and I'm seeing the same result. (I did use different test text just to be sure.) The previous test was in the iPhone 11 Pro (iOS 14) simulator. 

I'm wondering if the issue you're seeing might be due to a glitch with the cookie auth (not sure yet how to test this).  I can look into this angle tomorrow.

aerych:
That was a merry little debugging adventure :) 

So happy you happened to test with an Atomic site!  After logging in to your test site as an editor (thank you again!) I was able to reproduce the issue there.   To recap our chat conversation, the issue turned out to be that for Atomic sites, the value of their `unmapped_url` is still their mapped domain name, not a wpcom subdomain like for my test site.  This meant that the wpcom auth cookies weren't valid for the atomic site, so the published version of the post was shown rather than the autosaved version. This makes sense considering the way Atomic sites are configured.

Some investigation revealed the trick Calypso uses is to load wp-login.php with a redirect to the desired page.  In the case of Atomic sites with SSO enabled the redirect is fairly seamless.  In cases where SSO is not enabled, there is still a prompt for credentials as a semi-graceful fallback.  It's not perfect, but better than the current situation. 

I've updated the PR with some logic that checks for public atomic sites and rewrite the request URL to proxy through wp-login.php. I tested with your test site and it was working for me there.  I also created an ephemeral atomic site to test with but, interestingly, these sites do not report themselves as being atomic (?!) via the REST API and so regular wpcom auth is attempted.  This is something that needs to be addressed server-side, not in the app, so 🤷 for now.  I'll open an Android issue regarding Atomic published post previews. 

Ready for another peek!

cc @diegoreymendez in case you're interested in this as well.

ScoutHarris (approved):
Excellent debugging skillz!! It is in fact working as advertised. Well done!

:shipit:

WordPress/Classes/Utility/Networking/RequestAuthenticator.swift
```diff
@@ -182,6 +191,36 @@ class RequestAuthenticator: NSObject {
...
+            queryItems.append(URLQueryItem(name: "redirect_to", value: url.absoluteString))
+            components?.queryItems = queryItems
+            let requestURL = components?.url ?? url
+
+            let request = URLRequest(url: requestURL)
+            completion(request)
+        }
+
+        authenticationService.loadAuthCookiesForWPCom(into: cookieJar, username: username, authToken: authToken, success: {
+            done()
```
diegoreymendez, line 211: Fairly minor, but this jumped out at me while looking at the code.  I think you can simplify the success param to just be:

```swift
success: done,
```

(Without the `{}`s and `()`s)

aerych: True. If you don't mind I think I'll leave it as is so it's consistent with the other methods in that class.   Thanks for taking a peek!

aerych:
Thank you @ScoutHarris and @diegoreymendez !

PR #15072, opened 2020-10-08
Reader Improvements Detail: Address design review feedback

### LINKED ISSUE #15052: Reader Improvements: Reader Detail Design Review
Project Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14768

An issue to track design feedback items for the Reader Detail feature.

emilylaguna:
Fixes #15052

### To test:
1. Launch the app
2. Tap on Reader Detail
3. Make sure the bottom toolbar icons are all the same color in light and dark modes 
4. Tap on a topic
5. Make sure the back button says back
4. Tap back
5. Tap back again to go back to the feed
5. Make sure the Search icon is not black

This does not add any options to the more menu which will need to be addressed in a future task. 

### PR submission checklist:


WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
```diff
@@ -236,17 +236,7 @@ class ReaderDetailViewController: UIViewController, ReaderDetailView {
...
-        let titleView = UILabel()
-
-        titleView.attributedText = NSAttributedString(string: title ?? placeholder, attributes: UINavigationBar.standardTitleTextAttributes())
-        navigationItem.titleView = titleView
-        titleView.isHidden = true
-
-        // Allow the title to appear in the back button tap and hold in iOS 14+
-        self.title = title
-    }
+    func show(title: String?) { }
```
leandroalonso: If this method does nothing, why keep it?

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailCommentsView.swift
```diff
@@ -59,7 +59,10 @@ class ReaderDetailCommentsView: UIView, NibLoadable {
 
     // MARK: - Private: Helpers
     private func commentTitle() -> String {
-        let format = commentCount != 1 ? Strings.commentFormatPlural : Strings.commentFormat
+        var format = Strings.noCommentFormat
+        if commentCount > 0 {
```
leandroalonso: This is a little bit hard to read, you can either add multiple `else ifs` or use `switch`:

```
        switch commentCount {
        case let count where count == 1:
            format = Strings.commentFormat
        case let count where count > 0:
            format = Strings.commentFormatPlural
        default:
            format = Strings.noCommentFormat
        }
```

I guess this is one of the cases that writing less lines of code makes it hardto understand. :p

leandroalonso (changes requested):
The described steps work like a charm!

Left 2 comments to improve the code.

mbshakti:
Looking good and works as expected! Thank you for addressing Emily! 🙏

yaelirub:
Great work, @emilylaguna ! 
I followed the test steps and it works as described.

As noted in DM, I do see an issue with the header when going back from the topic feed : 
RI_3

yaelirub (changes requested):
Issue with header not animating when going back to view

emilylaguna:
> As noted in DM, I do see an issue with the header when going back from the topic feed :

Fixed in: https://github.com/wordpress-mobile/WordPress-iOS/pull/15072/commits/3de66fd58706c85bdbafb8be5ded670927f30c51

yaelirub (approved):
LGTM! 👏

Issue #14963, opened 2020-09-22
NSInvalidArgumentException: -[__NSCFType encodeWithCoder:]: unrecognized selector sent to instance 0x1032f2600


jkmassel:
**Sentry Url**: https://sentry.io/share/issue/e8f83bc0ac2648e696b58ada13f6e31b/
**User Count**: 93
**Count**: 149
**First Release**: 15.3.0.3
**First Seen**: 2020-08-18T21:08:38Z
**Last Seen**: 2020-09-21T18:02:10Z
**24 Hours**: 3
**30 Days**: 148
**LogID:** BA1FBB6A-00EE-41B8-A189-757FD8A38A22

```
NSInvalidArgumentException: nil is not a valid object ID
  File "MediaProgressCoordinatorNoticeViewModel.swift", line 169, in MediaProgressCoordinatorNoticeViewModel.blogInContext.getter
  File "<compiler-generated>", line null, in MediaProgressCoordinatorNoticeViewModel.blogInContext.getter
  File "MediaProgressCoordinatorNoticeViewModel.swift", line 78, in MediaProgressCoordinatorNoticeViewModel.notificationInfo.getter
  File "MediaProgressCoordinatorNoticeViewModel.swift", line 65, in MediaProgressCoordinatorNoticeViewModel.failureNotice.getter
  File "MediaProgressCoordinatorNoticeViewModel.swift", line 43, in MediaProgressCoordinatorNoticeViewModel.notice.getter
  File "MediaCoordinator.swift", line 681, in MediaCoordinator.mediaProgressCoordinatorDidFinishUpload
  File "<compiler-generated>", line null, in MediaCoordinator.mediaProgressCoordinatorDidFinishUpload
  File "MediaProgressCoordinator.swift", line 284, in MediaProgressCoordinator.refreshMediaProgress
  File "MediaProgressCoordinator.swift", line 275, in MediaProgressCoordinator.observeValue
  File "<compiler-generated>", line null, in @callee_guaranteed
  File "main.swift", line 7, in main
...
(12 additional frame(s) were not displayed)
```

bummytime:
@diegoreymendez @frosty @guarani, another `nil is not a valid object ID` this time in `MediaProgressCoordinator`. 😢

designsimply:
@bummytime would you like to assign this one since you marked it as high priority and it's a new crash? Alternately, you could de-escalate it if you feel the important Media issues already have attention right now.

[...]

Issue #14960, opened 2020-09-22
NSInvalidArgumentException: nil is not a valid object ID


jkmassel:
**Sentry Url**: https://sentry.io/share/issue/fa5a0d83563c4b85a92aaf6cfbe1c21d/
**User Count**: 114
**Count**: 157
**First Release**: 15.3.0.3
**First Seen**: 2020-08-24T16:42:37Z
**Last Seen**: 2020-09-22T02:40:15Z
**24 Hours**: 8
**30 Days**: 157
**LogID:** 170A3571-0A7B-4AA3-987E-38C133C38B02

```
NSInvalidArgumentException: nil is not a valid object ID
  File "Classes/Services/MediaService.m", line 581, in __56-[MediaService syncMediaLibraryForBlog:success:failure:]_block_invoke
...
(10 additional frame(s) were not displayed)
```

designsimply:
I see that this issue was given high priority but was not assigned. @bummytime may I loop you back in to get help assigning this one since it's a new crash? 

**Events in the last 90d:** 235
**Users affected in the last 90d:** 162
<sub>WORDPRESS-IOS-38VV: https://sentry.io/share/issue/fa5a0d83563c4b85a92aaf6cfbe1c21d/</sub>

<a href="https://sentry.io/organizations/a8c/discover/results/?display=daily&field=timestamp&field=release&field=device&field=os&field=culprit&name=All+Events&project=1438083&query=issue.id%3A1871198051&sort=-timestamp&statsPeriod=90d&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1">![image](https://user-images.githubusercontent.com/1119271/94750911-70334600-0344-11eb-93e4-3f7bd1705e46.png)</a>

bummytime:
Current impact → First Release: 15.3.0.3; 186 Users (5-10 events/day). I am removing the high-pri status from this issue. We will potentially add this issue to a "Media improvements" project starting in the near term.

designsimply:
Fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/15075. Props @diegoreymendez. 🎉

<sub>WORDPRESS-IOS-38VV: https://sentry.io/share/issue/fa5a0d83563c4b85a92aaf6cfbe1c21d/</sub>

[...]

PR #15075, opened 2020-10-08
Small changes to try and fix a Media issue.

diegoreymendez:
This change aims fo fix:
- https://github.com/wordpress-mobile/WordPress-iOS/issues/14963
- https://github.com/wordpress-mobile/WordPress-iOS/issues/14960

We're now retrieving the media object from the main context (which is consistent with what we were doing with the blog object).

I'm also removing some logging code I added for 15.8, since no reports came up in Sentry.  I'm instead assuming that the call to `objectWithID` should be replaced with `existingObjectWithID` to make sure the blog exists, and reporting an error if it doesn't.

I'm optimistic this could be the culprit behind a large number of bugs we've been seeing, where `media.blog` was `nil`.

## To test:

1. Launch the App from a new installation.
2. Go into the Media Library for one of your sites.  Make sure it loads perfectly.
3. Upload media.
4. Make sure the upload doesn't fail.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/MediaService.m
+++ b/WordPress/Classes/Services/MediaService.m
@@ -574,25 +574,18 @@ NSErrorDomain const MediaServiceErrorDomain = @"MediaServiceErrorDomain";
 - (void)syncMediaLibraryForBlog:(Blog *)blog
                         success:(void (^)(void))success
                         failure:(void (^)(NSError *error))failure
-{
-    // This has been added temporarily to try and track the source of a nil object exception
-    // Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/14960
-    //
-    // It's already crashing below inside performBlock.  This assertion just makes sure it crashes
-    // a bit earlier so we can get a better stack trace.
-    //
-    // - Diego Rey Mendez, 2 October 2020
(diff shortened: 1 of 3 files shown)


WordPress/Classes/Services/MediaService.m
```diff
@@ -574,25 +574,18 @@ - (void)getMediaURLFromVideoPressID:(NSString *)videoPressID
...
-        DDLogInfo(@"BlogID: %@", blog.objectID);
-        DDLogInfo(@"Blog MOC: %@", blog.managedObjectContext);
-        @throw NSInternalInconsistencyException;
-    }
-    
+{    
     __block BOOL onePageLoad = NO;
     NSManagedObjectID *blogObjectID = [blog objectID];
     [self.managedObjectContext performBlock:^{
-        Blog *blogInContext = (Blog *)[self.managedObjectContext objectWithID:blogObjectID];
```
diegoreymendez, line 595: I think this could be behind some of the `media.blog == nil` issues we were seeing, since this method doesn't fail if the blog object doesn't exist.

frosty (approved):
Media library and upload works as expected for me 👍

PR #15081, opened 2020-10-09
[iPad] No Sites: fix view sizes on rotation

### LINKED ISSUE #14342: Media: Free Photo Library no results view is cut off when the keyboard is active
Steps to reproduce:

1. Go to Media > + > Free Photo Library or Free GIF Library.
1. Tap the search field to make that field active (if it isn't already).
1. Observe that the NoResults view background graphic is cut off when the keyboard is present.

**Result:** Free GIF/Photo Library search screen no results background gets cut off when the keyboard is open.

IMG_4266 IMG_4265 IMG_4267 IMG_4268

<p style="clear:both;"><sup>Tested with WPiOS 15.1.0.0 TestFlight beta on iPhone 11 iOS 13.5.1.</sup></p>

ScoutHarris:
Ref: #14547 
Fixes: #14342

**TL;DR:** 
This puts in a workaround so the 'No Sites' views are sized correctly on rotation. ([master issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/14547), [example](https://github.com/wordpress-mobile/WordPress-iOS/issues/14547#issuecomment-667410591) of incorrect views, [discussion](https://github.com/wordpress-mobile/WordPress-iOS/issues/14547#issuecomment-702217917).)

**Details:**
This issue is a direct result of the split view widths not updating when the primary width is set to `full`. After some headbanging, the only way I could "fix" it was to just not update the column widths for `full`. _This as a workaround, and should be removed when the split view width issue is resolved._ I just figured until the root cause is fixed, we could at least get the views to look decent.

I put the workaround in the `WPSplitViewController` directly, instead of each VC that uses `full` width. Because those VCs have other logic around the `full` width, that would require more hacking and possible regressions. This was the most straightforward workaround, and the easiest to undo later.

@frosty - if this is an acceptable workaround, I'll add a note to https://github.com/wordpress-mobile/WordPress-iOS/issues/14547 to remove it later.

In addition, because of this workaround, I was also able to fix the NRV being vertically off center. This is a permanent fix, and should continue to work correctly once the split view widths are fixed. (This is what fixes #14342, and every other instance of the NRV being too high.)

**To test:**

---

No Sites view:
- On an iPad, log in with an account that has no sites.
- Rotate the device. 
  - Verify both views are sized correctly (i.e. there are no grey areas).
  - Verify the No Sites view is vertically centered in the left pane.

| Before | After |
|--------|-------|
| ipad_before | ipad_after |

---

NRV vertical placement:
- On an iPhone, log in with an account that has no sites.
  - Verify the No Sites view is vertically centered in the view.
- On an iPhone, log in with an account that has sites.
  - Go to Media > + > Free Photo Library or Free GIF Library.
  - Tap the search field to make that field active.
  - Verify the NRV no longer appears under the search bar.

| Before | After |
|--------|-------|
| iphone_no_sites_before | iphone_no_sites_after |
| iphone_media_before | iphone_media_after |

---

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
@@ -241,7 +241,7 @@ static NSInteger HideSearchMinSites = 3;
     
     // Ensure No Results VC is not shown. Will be shown later if necessary.
     [self.noResultsViewController removeFromView];
-    [self.tableView.refreshControl setHidden: false];
+    [self.tableView.refreshControl setHidden: NO];
     
     // If the user has sites, but they're all hidden...
     if (count > 0 && visibleSitesCount == 0 && !self.isEditing) {
@@ -290,7 +290,7 @@ static NSInteger HideSearchMinSites = 3;
                                            subtitleImage:nil
                                            accessoryView:nil];
 
-        [self.tableView.refreshControl setHidden: true];
+        [self.tableView.refreshControl setHidden: YES];
         [self addNoResultsToView];
     }
 }
@@ -335,7 +335,7 @@ static NSInteger HideSearchMinSites = 3;
                                            accessoryView:nil];
     }
 
-    [self.tableView.refreshControl setHidden: true];
+    [self.tableView.refreshControl setHidden: YES];
     [self addNoResultsToView];
     
 }
--- a/WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
@@ -241,6 +241,7 @@ static NSInteger HideSearchMinSites = 3;
     
     // Ensure No Results VC is not shown. Will be shown later if necessary.
     [self.noResultsViewController removeFromView];
+    [self.tableView.refreshControl setHidden: false];
     
     // If the user has sites, but they're all hidden...
     if (count > 0 && visibleSitesCount == 0 && !self.isEditing) {
@@ -288,6 +289,8 @@ static NSInteger HideSearchMinSites = 3;
                                                    image:@"mysites-nosites"
                                            subtitleImage:nil
                                            accessoryView:nil];
+
+        [self.tableView.refreshControl setHidden: true];
         [self addNoResultsToView];
     }
 }
@@ -332,6 +335,7 @@ static NSInteger HideSearchMinSites = 3;
                                            accessoryView:nil];
     }
 
+    [self.tableView.refreshControl setHidden: true];
     [self addNoResultsToView];
     
 }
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -241,6 +241,7 @@ - (void)updateViewsForCurrentSiteCount
     
     // Ensure No Results VC is not shown. Will be shown later if necessary.
     [self.noResultsViewController removeFromView];
+    [self.tableView.refreshControl setHidden: false];
```
frosty: `false` -> `NO`

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -288,6 +289,8 @@ - (void)showNoResultsViewForSiteCount:(NSUInteger)siteCount
                                                    image:@"mysites-nosites"
                                            subtitleImage:nil
                                            accessoryView:nil];
+
+        [self.tableView.refreshControl setHidden: true];
```
frosty: `true` -> `YES`

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -332,6 +335,7 @@ - (void)showNoResultsViewForAllSitesHidden
                                            accessoryView:nil];
     }
 
+    [self.tableView.refreshControl setHidden: true];
```
frosty: `true` -> `YES`

WordPress/Classes/ViewRelated/System/WPSplitViewController.swift
```diff
@@ -62,13 +62,19 @@ class WPSplitViewController: UISplitViewController {
             preferredPrimaryColumnWidthFraction = UISplitViewController.automaticDimension
         case .narrow:
             let columnWidth = WPSplitViewControllerNarrowPrimaryColumnWidth.width(for: size)
-
             minimumPrimaryColumnWidth = columnWidth
             maximumPrimaryColumnWidth = columnWidth
-            preferredPrimaryColumnWidthFraction = size.width / columnWidth
+            preferredPrimaryColumnWidthFraction = columnWidth / size.width
```
frosty, line 67: Huh, how did this 'work' for so long the wrong way round?!

frosty (approved):
Great that you fixed the NRV issues too – thank you! Tests great for me, just left a few tiny code comments. After that, :shipit:!

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
ScoutHarris: Bah! ObjC! Thank you. 😄

WordPress/Classes/ViewRelated/System/WPSplitViewController.swift
ScoutHarris: I dunno. And correcting it didn't seem to make a difference either. 🤷

ScoutHarris:
Thanks @frosty !

PR #15085, opened 2020-10-13
Enforce use of bundle exec in Fastfile+Podfile

AliSoftware:
## Description

This adds checks in the `Fastfile` and `Podfile` so that it stops you from accidentally run `fastlane` or `pod` without prefixing them with `bundle exec`.

If you forget to prepend `bundle exec` when invoking `fastlane` or `pod`, this will now error and stop, telling you to use `bundle exec` – so you run the versions of those tools specified in the repo's `Gemfile.lock` – instead of continuing to run with the system-wide version – which might not be the same and behave differently than when ran under bundle exec.

## To test:

1️⃣  Run `pod install` and see it fail.

<details><summary>Expected pod failure when running without <tt>bundle exec</tt></summary>

```
[!] Invalid `Podfile` file: Please run fastlane via `bundle exec`.

 #  from /Users/olivier/Documents/Dev/WordPress-iOS/Podfile:4
 #  -------------------------------------------
 #  unless ['BUNDLE_BIN_PATH', 'BUNDLE_GEMFILE'].any? { |k| ENV.key?(k) }
 >    raise 'Please run fastlane via `bundle exec`'
 #  end
 #  -------------------------------------------
```

(Note: Sadly there is no way to avoid the stacktrace from showing, which might make you think there is a syntax error in the Podfile while the syntax is correct and it just shows where the exception was raised)
</details>

2️⃣ Run `bundle exec pod install` (or `be pod install` if you created an alias for `bundle exec` in your `.zshrc`, which I recommend) and check that it runs as expected

3️⃣ Run `(cd Scripts; fastlane lanes)`, and check it fails with the following output

<details><summary>Expected fastlane failure when running without <tt>bundle exec</tt></summary>

```
[✔] 🚀 
[12:50:09]: fastlane detected a Gemfile in the current directory
[12:50:09]: However, it seems like you didn't use `bundle exec`
[12:50:09]: To launch fastlane faster, please use
[12:50:09]: 
[12:50:09]: $ bundle exec fastlane lanes
[12:50:09]: 
[12:50:09]: Get started using a Gemfile for fastlane https://docs.fastlane.tools/getting-started/ios/setup/#use-a-gemfile
[12:50:10]: Installing Ruby gem 'fastlane-plugin-wpmreleasetoolkit'...
[12:50:27]: Found gem "fastlane-plugin-release_helper" instead of the required name "fastlane-plugin-wpmreleasetoolkit"
[12:50:28]: Successfully installed 'fastlane-plugin-wpmreleasetoolkit'
[12:50:28]: Error loading plugin 'fastlane-plugin-wpmreleasetoolkit': cannot load such file -- fastlane/plugin/wpmreleasetoolkit
+-----------------------------------+-----------+-------------------------------------------------------------------------+
|                                                      Used plugins                                                       |
+-----------------------------------+-----------+-------------------------------------------------------------------------+
| Plugin                            | Version   | Action                                                                  |
+-----------------------------------+-----------+-------------------------------------------------------------------------+
| fastlane-plugin-wpmreleasetoolkit | undefined | No actions found                                                        |
| fastlane-plugin-sentry            | 1.6.0     | sentry_upload_dsym sentry_set_commits sentry_upload_sourcemap           |
|                                   |           | sentry_finalize_release sentry_upload_file sentry_upload_proguard       |
|                                   |           | sentry_create_release                                                   |
| fastlane-plugin-appcenter         | 1.8.0     | appcenter_fetch_version_number appcenter_fetch_devices appcenter_upload |
| fastlane-plugin-test_center       | 3.14.2    | suppressed_tests suppress_tests tests_from_xcresult                     |
|                                   |           | suppress_tests_from_junit tests_from_xctestrun                          |
|                                   |           | quit_core_simulator_service collate_test_result_bundles                 |
|                                   |           | tests_from_junit multi_scan test_options_from_testplan                  |
|                                   |           | collate_html_reports collate_junit_reports collate_json_reports         |
|                                   |           | testplans_from_scheme collate_xcresults                                 |
+-----------------------------------+-----------+-------------------------------------------------------------------------+

[!] No actions were found while loading one or more plugins
    Please use `bundle exec fastlane` with plugins
    More info - https://docs.fastlane.tools/plugins/using-plugins/#run-with-plugins

[12:50:28]: ------------------------------
[12:50:28]: --- Step: default_platform ---
[12:50:28]: ------------------------------

[!] Please run fastlane via `bundle exec`
```
</details>

4️⃣ Run `(cd Scripts; bundle exec fastlane lanes)` and check that it runs without issues and lists the available lanes


Podfile
```diff
@@ -1,5 +1,9 @@
 source 'https://cdn.cocoapods.org/'
 
+unless ['BUNDLE_BIN_PATH', 'BUNDLE_GEMFILE'].any? { |k| ENV.key?(k) }
+  raise 'Please run fastlane via `bundle exec`'
```
mokagio:
```suggestion
  raise 'Please run CocoaPods via `bundle exec`'
```

mokagio (approved):
Verified that calling without `bundle exec` fails the command and prints an error

<table>
<tr>
<td>image
</td>
<td>image
</td>
</tr>
</table>

I'm curious about that "`Error loading plugin...`" message. It's not something we _need_ to fix, since it works when running with Bundler, but I'd never seen something like that. 🤔 

There's a little fix to make, approving it in advance because it's not related to the functionality. Good to merge once sorted.

---

> or `be pod install` if you created an alias for bundle exec in your `.zshrc`, which I recommend

[Me too](https://www.mokacoding.com/blog/terminal-aliases/). I use [`be` for `bundle exec` and `bf` for `bundle exec fastlane`](https://github.com/mokagio/dotfiles/blob/480245e0f91998be9477341a17efdd567fc3b754/aliases#L121-L123). 😄 🙌

AliSoftware:
@mokagio I was surprised by the "Error loading plugin..." message when not using bundle exec too, but I got it on `develop` even before my changes, so I figured it was unrelated.

It seems like a check built into fastlane, and only triggered when we use plugins, which makes sense because the fastlane plugins you want to use are defined in the `Gemfile` (well via the `fastlane/Pluginfile` file, which is imported from the `Gemfile`), so no surprise fastlane itself would check and warn about that if you don't run via `bundle exec`

PR #15087, opened 2020-10-13
Use a default timestamp for custom Sentry event

jkmassel:
This fixes a potential crash that could happen if the `Event` object is later used for encrypted logging

PR submission checklist:


shiki (approved):
Yup. This looks like the only place that the crash could happen. Thank you! :shipit:

PR #15099, opened 2020-10-14
Fix screenshot test for Russian & Thai

### LINKED ISSUE #15074: Testing: UI tests fail with device set to certain languages
While generating app store screenshots for https://github.com/wordpress-mobile/WordPress-iOS/pull/15010, I realized it fails entirely (no screenshots generated on any device) when the device is set to `th` and `ru`. The test runs fail with this error:

> testGenerateScreenshots, Failed to synthesize event: Timed out while synthesizing event.
> LoginUsernamePasswordScreen.swift:41

However, the failures aren't specific to that PR and likely affect login for all of our UI tests in those languages.

rachelmcr:
Fixes #15074

As noted in the issue, for some reason typing text works in most text fields and languages _except_ secure text entry fields in Russian and Thai (of the languages we run the test in).

That issue prevented the test from logging in when the device is set to those languages. To work around that, this PR checks the language code on the device and pastes (instead of typing) the password in those languages. I didn't switch to using a paste method for all languages because in repeated test runs I found pasting to be a bit flaky, so this limits it to the languages where it's really needed.

We also have a small mocks improvement in https://github.com/wordpress-mobile/WordPressMocks/pull/21 so I've updated the Podfile here so we can test that change.

To test:

* Run `rake mocks`.
* Set your simulator to Russian and Thai and confirm the WordPressScreenshotGeneration test passes.
* Confirm the WordPressScreenshotGeneration test continues to pass in another language (e.g. English).

(You can run the test at least once on an iPad Pro, 12.9 inch, 2nd or 3rd gen to test the mocks change.)

PR submission checklist:


rachelmcr:
I see there's a Podfile.lock conflict; I'll fix that once the mocks PR is merged and I update the Podfile to use the published pod.

WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
```diff
@@ -39,6 +39,17 @@ extension XCUIElement {
         self.typeText(text)
     }
 
+    /**
+     Pastes text from clipboard to the field
+     Useful for scenarios where typing is problematic, e.g. secure text fields in Russian.
+     - Parameter text: the text to paste into the field
+     */
+    func pasteText(_ text: String) -> Void {
```
jkmassel, line 47: One suggestion – [here's the implementation of this for Woo](https://github.com/woocommerce/woocommerce-ios/blob/9b177373c1a47500127f2a58e73a05a78458424e/WooCommerce/WooCommerceUITests/Utils/XCTest%2BExtensions.swift#L42-L52). It preserves the contents of the clipboard to prevent data loss on developer machines.

Not critical I think, but could be nice to have 😄

jkmassel (approved):
I left a note about a small change for your consideration.

I tested in English, Russian, and Thai.

English worked the first time; but for whatever reason Russian and Thai both failed the first time but worked on subsequent runs.

Additionally, it doesn't seem to work on iOS 14, just on 13.5.

I don't think either of these things are blockers – `snapshot` will auto-retry if there's a failure so worst-case the generation just takes slightly longer.

WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
rachelmcr: Aha! Before I did this I could have sworn we already had an implementation. I must have been thinking of Woo but forgot to check there — that's nice and an easy change, good call.

rachelmcr:
Thanks for the review!

> English worked the first time; but for whatever reason Russian and Thai both failed the first time but worked on subsequent runs.

Yeah, if it failed on the paste action I'm not too surprised because it was pretty flaky when I was trying it out. It's the most straightforward workaround I came up with for now but we can think about more stable options to try (maybe using language-specific text so we can type instead of paste).

> Additionally, it doesn't seem to work on iOS 14, just on 13.5.

Good point. We haven't made any changes for tests to run on iOS 14 but it's on my radar. Agreed that it isn't a blocker for this but I'll see if I can follow up on it soon.

I made the changes mentioned above (the suggested change to the paste implementation and updated the WordPressMocks pod) so I'll go ahead and merge this PR once tests are passing.

PR #15106, opened 2020-10-14
Pin CropViewController to 2.5.3

leandroalonso:
Pin CropViewController in `2.5.3` (this time in `Podfile` too) to avoid warning outputs after pod install.


emilylaguna (approved):
Deja vú 😕 

🚢 :shipit:

PR #15108, opened 2020-10-14
Update app store screenshots to new design

rachelmcr:
This updates our screenshot composition to match the new designs:

image
image
image
image

Internal design ref: pbArwn-13u-p2

## Testing

1. Open the fastlane directory: `cd Scripts/fastlane`
2. Generate a full set of screenshots from the app, if needed (If you already have a set of app screenshots, you can skip to the next step and the composition command will use your existing screenshots):
   1. Start the mock server: `rake mocks`
   2. Generate the screenshots: `bundle exec fastlane screenshots`
3. Add the screenshot composition: `bundle exec fastlane create_promo_screenshots`

## Review

Please confirm code changes look reasonable (in particular, changes to the composition config in `screenshots.json`).

After running the screenshot composition, confirm each screenshot looks as expected:

* Appropriate filename
* Correct background
* Correct string with appropriate spacing/line breaks
* Correct screenshot, well placed within device frame

The English screenshot composition did get a design review already (internal ref: paaHJt-1nL-p2#comment-3146), so the background and screenshot (including placement within the device frame) should be ok. However, I had to make some adjustments to the string positioning once we added translated strings, so I'd appreciate attention on the strings and their positioning in particular.

Note: We don't have translations for Norwegian (`no`) or Thai (`th`) at this point, so the composition won't be correct for those languages.

This PR only needs one reviewer, but @jkmassel and @JavonDavis pinging you as an FYI in case you want to take a look.

## PR submission checklist:


startuptester (commented):
Everything looks great, just a nitpick about line breaks. 
These line breaks in the strings vary from the original mocks. 

Screen Shot 2020-10-15 at 3 34 53 PM
Screen Shot 2020-10-15 at 3 35 13 PM
Screen Shot 2020-10-15 at 3 35 44 PM

rachelmcr:
Yeah, the way the text wraps is a bit tricky with our current setup. Here are the parameters we're working with:

* The strings are on a single line (so we can use the same strings for the iPad screens that are wider and the iPhone screens where we wrap them to fit).
* The text area is the same size for every language.
* The text area can be adjusted per screen (so for example the Media screen can have a different text area size than the other screens if needed).

I had the English text wrapping exactly matching the designs before we got the translated strings back, but I had to make [these adjustments](https://github.com/wordpress-mobile/WordPress-iOS/pull/15108/files/7f122385b1965086ccd878fe433dc9c13db0e528..6102d59dbed2b9ef56761737d428f62035da2691) to make the text areas work for translated strings. (In some cases, e.g. German and French, a single word was getting split onto two lines because the text area was too narrow.)

Here's an example of that issue on the iPhone XS Max Media screenshot in French:



Unfortunately, the adjustments to fix that kind of issue mean the English text doesn't wrap exactly as it did before. (And the fixes I made don't entirely fix the text wrapping in all languages. Especially in languages that don't have spaces between words, e.g. Japanese, there are some cases where words are getting wrapped in weird ways.)

If we really want to make it work well in all languages, some options could be:

* Add line breaks to the strings so we can control where the break happens per language, instead of using text wrapping to manage it. If we do this, we'll either need to use separate strings for iPhone and iPad (so iPad can keep the strings all on one line) or change the design to use multi-line strings even on the iPad screens. (In the last version of app store screenshots we had line breaks like this, but it worked because we also included the line breaks in the iPad screenshots.)
* Change how we do the composition to allow for language-specific composition configurations (or at least exceptions). This would be a bigger change.

@megsfulton What do you think? At least for the 15.9 release, would you prefer we revert to the version I shared with you before where it all matches the designs in English (but words are wrapped onto two lines in some other languages) or is it ok to change the English layout (as in the examples above) to improve the text wrapping in other languages?

jkmassel:
Can we add `override_status_bar: true` to the `capture_ios_screenshots` call? It'll ensure the date and time are consistent for all screenshots (and that the cellular bars are full as well, instead of disabled)

jkmassel (dismissed):
Can we add `override_status_bar: true` to the `capture_ios_screenshots` call? It'll ensure the date and time are consistent for all screenshots (and that the cellular bars are full as well, instead of disabled)

megsfulton:
> At least for the 15.9 release, would you prefer we revert to the version I shared with you before where it all matches the designs in English (but words are wrapped onto two lines in some other languages) or is it ok to change the English layout (as in the examples above) to improve the text wrapping in other languages?

I think at least for this release it's okay to have the English layout changed for the sake of improving text wrapping in other languages. 

For a later release, out of the two options outlined above, I think adding a line break to the strings is probably the right approach. The iPad screens should still work okay with two lines of text. 

Another option could be reducing the text size for languages other than English to better accommodate the longer strings. Which I think might be a language specific configuration like you described? In that case it might worth seeing what the screens look like with a slightly smaller text size across all languages.

startuptester (approved):
Problems discussed with @jkmassel seem to be caused by the tools (time in status bar displaying as '41'). Decided as a follow up task if necessary.

rachelmcr:
> Problems discussed with @jkmassel seem to be caused by the tools (time in status bar displaying as '41'). Decided as a follow up task if necessary.

Weird, I'm seeing that with the status bar override too. Sounds good to investigate it separately as a followup to this PR; I opened https://github.com/wordpress-mobile/WordPress-iOS/issues/15265 to keep track.

I went with this fix for the text wrapping, as it's a much more straightforward change at this point:

> For a later release, out of the two options outlined above, I think adding a line break to the strings is probably the right approach. The iPad screens should still work okay with two lines of text.

I also went ahead and (carefully) added line breaks for non-English languages, to confirm that works without any other layout changes (like text size). If translators submit any changes to those line breaks we can regenerate the screenshots and confirm they still look good at that point.

startuptester (approved):
Thanks for doing this update @rachelmcr!

PR #15115, opened 2020-10-15
Reader: Do not override images that already have an anchor link

### LINKED ISSUE #15009: Reader: Image links are ignored
### Expected behavior

If an image in a post has a link attached, I expect tapping that image in the Reader detail view to open the link.

### Actual behavior

To make all images clickable, [this code](https://github.com/wordpress-mobile/WordPress-iOS/blob/3a26653502ce63186b0a9539a5f852ac52d0edf6/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderWebView.swift#L73) wraps them in a hyperlink. When an image already has a link, this ends up going in between the image tag and its original hyperlink tag, and disabling the original link.

```html
<body class="reader-full-post reader-full-post__story-content">
  <figure class="wp-block-image size-large">
    <a href="https://en.wikipedia.org/wiki/Tree"> // Original link from post
      <a href="https://image.jpg"> // Inserted by ReaderWebView
        <img src="https://image.jpg">
      </a>
    </a>
  </figure>
</body>
```

1. On a site you're following in the reader, create a new post
2. Add an image to the post, and set the link property of that image to a custom URL
3. Open that post in the iOS Reader
4. Notice that tapping the image opens the full-size image viewer, instead of opening the link

leandroalonso:
Fixes #15009 

### To test

1. Open a post in Reader that contains an image with a link (eg.: https://charlietest12345.blog/2020/10/15/image-with-link/)
2. Tap the first image and checks that it opens https://wordpress.org/
3. Tap the 2nd image and checks that it shows itself fullscreen

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderWebView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/WebView/ReaderWebView.swift
@@ -98,8 +98,11 @@ class ReaderWebView: WKWebView {
 
             // Make all images tappable
             // Exception for images in Stories, which have their own link structure
+            // and images that already have a link
             document.querySelectorAll('img:not(.wp-story-image)').forEach((el) => {
-                el.outerHTML = `<a href="${el.src}">${el.outerHTML}</a>`
+                if (el.parentNode.nodeName.toLowerCase() !== 'a') {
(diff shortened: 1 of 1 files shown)


emilylaguna (approved):
LGTM! Works as expected!

PR #15127, opened 2020-10-19
Merge 16.0 code freeze into develop

jkmassel:
Merges 16.0 code freeze into `develop`


.circleci/cache-version
```diff
@@ -1 +1,2 @@
-000010
+# Replace this with your own UUID – can be generated on the command line using uuidgen
+9A83906D-78D0-41F4-8B4C-7E6F06AC615C
```
jkmassel: Using a UUID prevents issues where the same number may be used between branches, and this needed to be changed because of RNCMaskedView issues

.circleci/config.yml
```diff
@@ -174,7 +174,6 @@ jobs:
       - run: 
           name: Install other tools
           command: |
-            brew unlink python@2
```
jkmassel, line 177: Python 2 isn't installed on the XC12 images

.circleci/cache-version
```diff
@@ -1 +1,2 @@
-000010
+# Replace this with your own UUID – can be generated on the command line using uuidgen
```
AliSoftware: nit: felt like the current wording might make people think that each developer was supposed to paste their own UUID here and keep the change local or something (like if this was a local config file similar to `~/.wpios-env.default`) 🧐  Maybe something like this instead?

```suggestion
# To invalidate the cache, generate a new UUID using `uuidgen` on the command line then paste it here
```

AliSoftware (approved):
Apart from comment nit, LGTM!

mokagio (approved):
LGTM 👍 

I never noticed the `release_notes.txt` files being deleted, although it's quite possible I missed it all this time.

They'll be [recreated by `download_metadata.swift`](https://github.com/wordpress-mobile/WordPress-iOS/blob/64fe05e026899d78785f3947f400a7b7353aec96/Scripts/fastlane/download_metadata.swift#L116), so that's all good. I wasn't able to find the code that would have deleted them, though.

.circleci/cache-version
jkmassel: Yep, good call!

jkmassel:
> I never noticed the release_notes.txt files being deleted, although it's quite possible I missed it all this time.

Yeah I wasn't 100% sure why this happened, but I don't think it's a blocker here either for the same reason. Thanks @mokagio!

PR #15129, opened 2020-10-19
Adjust PageTemplateCategories to be associated to the Blog

chipsnyder:
### Description
This is a change to associate the fetched starter page templates to be associated with the current blog. Fetching of blog layouts currently come from one of two API endpoints:
WPCOM: `/wpcom/v2/sites/\(blogID)/block-layouts` <- includes `blogID`
Self-Hosted: `/wpcom/v2/common-block-layouts`

This means that self-hosted sites, and different WPCOM sites could in the future be provided with different sets of starter page templates and categories. The previous implementation would fetch all categories when loading a site. So a user could end up with a mix-match of layouts. However, this is a bit of an edge case because the API call is made every time the picker loads, so the picker could potentially load with the wrong data then would update to the correct set of layouts when the call completes.

### To test:
📓 This is best to use a user that has multiple sites associated with their WPCom username. It's also easiest to use a fresh install to prevent any contamination of cached data from any previous testing. 
<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
            
</details>

1. On site `A` navigate to the Modal Layout Picker
    - Wait for layouts to load 
1. Disconnect the device from the internet
1. On site `B` navigate to the Modal Layout Picker
    - **Expect** The page to show the Ghost cells
    - 📓 If you perform these steps on a build from `develop` you'll notice that layouts will load here.
1. On site `A` navigate to the Modal Layout Picker
    - **Expect** The page to show with the layouts that were previously fetched. 

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Blog.h
+++ b/WordPress/Classes/Models/Blog.h
@@ -13,6 +13,7 @@ NS_ASSUME_NONNULL_BEGIN
 @class Role;
 @class QuickStartTourState;
 @class UserSuggestion;
+@class PageTemplateCategory;
 
 extern NSString * const BlogEntityName;
 extern NSString * const PostFormatStandard;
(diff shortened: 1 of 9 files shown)


antonis (approved):
Great work Chip! 🎉
The app works as expected in the test scenarios and the code LGTM 👍

PR #15130, opened 2020-10-19
Big Titles: implement base

diegoreymendez:
Implements the base changes for working on Big Titles.

## Testing:

TBD

## PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15131, opened 2020-10-19
New AddSiteAlertFactory convenience class.

diegoreymendez:
This PR implements a new `AddSiteAlertFactory` class to modularize some of the code that was previously located in `BlogListViewController.swift`.

The reason for this change is because of some upcoming changes, that will require this action sheet to be shown from within another VC (`BlogDetailsViewController`).

This is part of [a larger upcoming change](https://github.com/wordpress-mobile/WordPress-iOS/pull/15130), and was broken up to make the review and testing easier.

## To test:

Please remember to test this both in iPhone and iPad to make sure I haven't broken the UI.

### Test 1:

1. Log into your WP.com account
2. Go to My Sites.
3. Tap “Switch Sites” in the top left corner.
4. Tap “+”

An action sheet with 2 buttons should come up, looking like this.

actionSheet

5. Tap “Create WordPress.com Site”

Make sure the WP.com Site creation Wizard is launched.

Screen Shot 2020-10-19 at 19 48 40

You can tap “Cancel” and proceed with the second test.

- [ ] iPhone
- [ ] iPad

### Test 2:

1. Log into your WP.com account
2. Go to My Sites.
3. Tap “Switch Sites” in the top left corner.
4. Tap “+”
5. Tap “Add self-hosted site”

Make sure the Self-Hosted Site Address VC is shown.  It looks like this:

Screen Shot 2020-10-19 at 19 50 08

Feel free to “Cancel”

- [ ] iPhone
- [ ] iPad

## PR submission checklist:


WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
```diff
@@ -23,6 +24,8 @@ enum FeatureFlag: Int, CaseIterable, OverrideableFlag {
         }
 
         switch self {
+        case .bigTitlesWhiteHeaders:
```
frosty: Can you use the `newNavBarAppearance` feature flag, or are you differentiating these changes for some reason?

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -1011,18 +981,25 @@ - (void)showAddSiteAlertFrom:(id)source
     if (self.dataSource.allBlogsCount > 0 && self.dataSource.visibleBlogsCount == 0) {
         [self setEditing:YES animated:YES];
     } else {
-        UIAlertController *addSiteAlertController = [self makeAddSiteAlertController];
+        AddSiteAlertFactory *factory = [AddSiteAlertFactory new];
```
frosty, line 984: How about we move this method to a Swift extension on BlogListViewController? Then we can make `AddSiteAlertFactory` a struct and drop the @objc annotations?

frosty (commented):
Changes look good, tested for wpcom and self-hosted, iPhone and iPad ✅

Just left two code comments for you.

frosty:
Oh, also, fwiw I noticed on this branch that the pull to refresh indicator in the blog list is showing in front of the content:

image

WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
diegoreymendez: Actually, I wasn't aware that `newNavBarAppearance` was created for this purpose.

diegoreymendez: Removed.

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
diegoreymendez: As discussed via DM, since there are a few ObjC private members accessed in this class (`defaultWordPressComAccount`, `showLoginControllerForAddingSelfHostedSite`, `addSiteAlertController`), I'm leaving this as-is for now.

I do share the thought we should make more of our code Swift, but in this case it's forcing me to do more changes than I'd like to.

I'll try to dig deeper into this idea when implementing the same logic for the blog details, so we can hopefully have shared presentation code.

diegoreymendez: Marking as resolved for now.

diegoreymendez:
@frosty - I just checked in `develop` and was able to reproduce the pull-to-refresh issue you mentioned.  I was expecting this was not introduced in this PR considering the nature of the changes.

image

diegoreymendez:
Ready for another look.

diegoreymendez:
The fix for pull-to-refresh is up in another PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15136

WordPress/Classes/ViewRelated/Blog/Add Site/AddSiteAlertFactory.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+
+/// This class takes care of constructing our "Add Site" action sheets.  It does not handle any presentation logic and does
+/// not know any external data sources - all of the data is received as parameters.
+@objc
+class AddSiteAlertFactory: NSObject {
+
+    @objc
+    func make(
```
Gio2018: maybe we could use a more specific name here (e.g. makeAddSiteAlert or something like that)?

Gio2018 (approved):
Hey @diegoreymendez ! This works great, I only had one small, totally optional, code comment.
:shipit: !!

WordPress/Classes/ViewRelated/Blog/Add Site/AddSiteAlertFactory.swift
diegoreymendez: Changed.

Issue #15103, opened 2020-10-14
Today Widget - UI


Gio2018:
Build the Today Widget UI with SwiftUI. At this stage, we can use a static model (to be removed in the future) to test the UI.

Designs:

**Small size**
![variations-small-3](https://user-images.githubusercontent.com/34376330/96767592-582e6f80-13a2-11eb-9b6c-cc25b899101b.png)

**Medium size**
![variations-medium-5](https://user-images.githubusercontent.com/34376330/96767659-7005f380-13a2-11eb-9217-9e7c8a2a10f7.png)

PR #15132, opened 2020-10-19
iOS 14 Today Widget: Small Size UI

Gio2018:
Ref #15103
This adds the small size Today Widget for iOS 14

Notes: 
- this PR only adds the UI for the small size Widget.
- this PR also contain additional commits that fix provisioning profiles and allow to cut installable builds

Known issues:
- The content of the widget is just an example
- The medium size widget is not implemented yet
- The displayed number of views does not align with the designs yet, specifically the decimal separator does not currently encroach into the padding area
- The vertical spacing between title lines, title and "Today" and "Views" and views number is automatically set and not explicitly set to 8 pts for now.

To test:

- install the app on a device or simulator with iOS 14
- add a new widget, scroll down to find WordPress and then tap to add
- make sure the added widget is consistent with the designs in #15103. See also examples below
<p align=center>
  
</p>

PR submission checklist:


diegoreymendez:
Hey @Gio2018, when looking at the commit history for this PR I'm seeing a few commits from other people.  Any chance that `feature/15103-today-widget-small-size-ui` includes more than the changes you intended?  Alternatively, any chance the target branch for this PR needs an update?

Gio2018:
> Hey @Gio2018, when looking at the commit history for this PR I'm seeing a few commits from other people.  Any chance that `feature/15103-today-widget-small-size-ui` includes more than the changes you intended?  Alternatively, any chance the target branch for this PR needs an update?

Hey @diegoreymendez , yes there are some commits to fix provisioning profiles for the widget. I added a note in the description to clarify this

WordPress/WordPressHomeWidgetToday/Model/TodayWidgetContent.swift
```diff
@@ -0,0 +1,12 @@
+import WidgetKit
+
+
+struct TodayWidgetContent: TimelineEntry {
+    var date = Date()
```
frosty: Is there a reason these properties couldn't be `let`s?

WordPress/WordPressHomeWidgetToday/Supporting Files/WordPressHomeWidgetTodayRelease-Alpha.entitlements
```diff
@@ -0,0 +1,14 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.org.wordpress.alpha</string>
+	</array>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)3TMU3BH3NK.org.wordpress</string>
```
frosty: I think the bundle prefix is different for alpha and internal builds (check the existing widget entitlements files) – `99KV9Z6BKV`. So not sure if it needs to match here and in the Internal entitlements?

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
```diff
@@ -1,70 +1,52 @@
...
+
+
+    func placeholder(in context: Context) -> TodayWidgetContent {
+        staticModel
     }
 
-    func getSnapshot(for configuration: ConfigurationIntent, in context: Context, completion: @escaping (SimpleEntry) -> ()) {
-        let entry = SimpleEntry(date: Date(), configuration: configuration)
+    func getSnapshot(in context: Context, completion: @escaping (TodayWidgetContent) -> ()) {
+        let entry = staticModel
```
frosty: Is there a reason for creating a separate variable here, instead of passing `staticModel` directly in `completion`?

WordPress/WordPressHomeWidgetToday/View/TodayWidgetView.swift
```diff
@@ -0,0 +1,63 @@
...
+            Spacer()
+        }
+        .padding(.all, Constants.padding)
+    }
+
+    private enum Constants {
+        // TODO - TODAYWIDGET: SwiftUI should be able to automatically localize strings,
+        // so let's not use NSLocalizedString for now and check how this will fit in our existing system
+        // Titles
+        static let todayTitle = "Today"
```
frosty: I _think_ we'll need to declare these as `LocalizedStringKey` for localization to work?

WordPress/WordPressHomeWidgetToday/View/TodayWidgetView.swift
```diff
@@ -0,0 +1,63 @@
...
+        static let commentsTitle = "Comments"
+
+        // Overall Appearance
+        static let padding: CGFloat = 16
+
+        // Site Title
+        static let siteTitleLines = 2
+
+        // Views
+        static let viewsTitleColorName = "Blue50"
```
frosty: How about storing the `Color` reference itself here, rather than just the name?

frosty (commented):
Hey, looking great! Just a few comments for you :)

WordPress/WordPressHomeWidgetToday/Model/TodayWidgetContent.swift
```diff
@@ -0,0 +1,12 @@
...
+    // TODO - TODAYWIDGET: see if this needs to be LocalizedStringKey
```
diegoreymendez: If this is the site title as chosen by the user, there's no need to localize (and I think we can't even if we wanted to).

diegoreymendez (commented):
Works as described.  I agree with @frosty 's observations.

WordPress/WordPressHomeWidgetToday/Model/TodayWidgetContent.swift
Gio2018: Nope. Fixed

WordPress/WordPressHomeWidgetToday/Supporting Files/WordPressHomeWidgetTodayRelease-Alpha.entitlements
Gio2018: I think so too and good catch! Before messing up with entitlements, I'd kindly ask @jkmassel and @AliSoftware if there's anything specific that needs to be done here?

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
Gio2018: Nope. Fixed

WordPress/WordPressHomeWidgetToday/View/TodayWidgetView.swift
Gio2018: 👍

WordPress/WordPressHomeWidgetToday/View/TodayWidgetView.swift
Gio2018: 👍

WordPress/WordPressHomeWidgetToday/Model/TodayWidgetContent.swift
Gio2018: 👍

WordPress/WordPressHomeWidgetToday/Supporting Files/WordPressHomeWidgetTodayRelease-Alpha.entitlements
jkmassel: Yep – good catch @frosty. This is fixed in 1cdff2b0!

Gio2018: Thank you @jkmassel !!

AliSoftware: Great catch!!

WordPress/WordPressHomeWidgetToday/View/TodayWidgetView.swift
```diff
@@ -0,0 +1,63 @@
...
+                    .foregroundColor(Color(.secondaryLabel))
+
+                Spacer()
+
+                Text(Constants.viewsTitle)
+                    .font(.caption)
+                    .fontWeight(.semibold)
+                    .foregroundColor(Constants.viewsTitleColor)
+
+                Text("\(content.views)")
```
AliSoftware, line 28: I'm impressed that SwiftUI is able to auto-localize this, and that formatting of numbers works transparently there 👌 
_(I'm guessing because `Text` takes a `LocalizedStringKey` and this type has a custom `StringInterpolation` overload for number types which applies formatting to it? 🤯 )_

I've tested that on a simulator set to French and it's properly formatted with a thin space – as expected in French locale – instead of a comma like English does, so 👍

AliSoftware:
Just tested the widget with VoiceOver on (maybe it's a bit too early to test that at that stage but just wanted to still drop that feedback somewhere).

_To give a reference example, when tapping e.g. Apple's Music widget (in size small), VoiceOver enounces something like "Music; Widget. {Current song and artist}. Button. swipe up or down to select custom action, Then double tap to activate". So, in short, the widget is seen as a single element (single tap target) and it reads out the whole content displayed by the widget (including current song info) immediately when selected._

By contrast, when selecting the WP widget, it only says "WordPress; Widget." and that's it. And then you can tap to select each individual element inside of the widget (like the title, or the static text "Views", or the view count) each individually.

I think the WP Widget should instead be seen as a single element group by VoiceOver (like Music), with the whole widget being the element (and not allowing to select individual views that compose the widget content), and when selected read something like "WordPress; Widget. Around the world with Pam, 1752 views. Button. […]"

Gio2018:
> Just tested the widget with VoiceOver on (maybe it's a bit too early to test that at that stage but just wanted to still drop that feedback somewhere).
> 
> _To give a reference example, when tapping e.g. Apple's Music widget (in size small), VoiceOver enounces something like "Music; Widget. {Current song and artist}. Button. swipe up or down to select custom action, Then double tap to activate". So, in short, the widget is seen as a single element (single tap target) and it reads out the whole content displayed by the widget (including current song info) immediately when selected._
> 
> By contrast, when selecting the WP widget, it only says "WordPress; Widget." and that's it. And then you can tap to select each individual element inside of the widget (like the title, or the static text "Views", or the view count) each individually.
> 
> I think the WP Widget should instead be seen as a single element group by VoiceOver (like Music), with the whole widget being the element (and not allowing to select individual views that compose the widget content), and when selected read something like "WordPress; Widget. Around the world with Pam, 1752 views. Button. […]"

Hey @AliSoftware , thanks for being so thorough in looking into the widget. The accessibility is indeed yet to be implemented, but this is valuable feedback to take into account for that. Thank you!

frosty (approved):
Looks great! Thanks for making the changes!

PR #15136, opened 2020-10-20
Fixes spinning wheel on top of blog list.

diegoreymendez:
@frosty noticed an issue with our blog list, that's making the pull-to-refresh spinning wheel to come up on top of our table view.

This PR fixes the issue.

## To test:

1. Launch the app.
2. Tap "Switch Site" in the top right corner.

Make sure you don't see this:

## PR submission checklist:


leandroalonso (approved):

leandroalonso:
@diegoreymendez given is such a simple fix, I wonder if we should target `release/16.0` instead of `develop`.

diegoreymendez:
> @diegoreymendez given is such a simple fix, I wonder if we should target release/16.0 instead of develop.

The thing is I don't think this is high priority enough to warrant a hotfix.  It's not blocking the UX.

Gio2018 (approved):
:shipit:!

PR #15137, opened 2020-10-20
[Modal Layout Picker] Retry view

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2740

## To test:
<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
            
</details>

📓 If you've played around with this feature recently then it would be best to start from a fresh build to make sure you aren't seeing cached data
 - Layouts are cached per site. So to help testing if you fetch the data on site A, you can put the device in airplane mode again and test on site B.

### No Network Connection Displays
1. Make sure your device is not connected to the internet
1. Navigate to Modal Layout Picker
    - **Expect** to see the no network connection appear

<kbd>Screen Shot 2020-10-20 at 4 01 02 PM</kbd>

### Retry works
1. Make sure your device is not connected to the internet
1. Navigate to Modal Layout Picker
    - **Expect** to see the no network connection appear
1. Re-establish a network connection
1. Select "Retry"
    - **Expect** to see the data load

### Quick Sanity
1. Make sure your device is not connected to the internet
1. Navigate to Modal Layout Picker for a site where you have opened the picker before
    - **Expect** to see the data load

### PR submission checklist:


chipsnyder:
@iamthomasbishop I put together a similar no network screen that @antonis implemented. I did do a bit of a tweak here from the options we had mentioned in https://github.com/wordpress-mobile/gutenberg-mobile/issues/2563#issuecomment-712297796. It seemed like the norm on iOS was a bit different so I tried to follow that standard. 

I did add the "Retry" button where most of the other screens that use this would have defaulted to a "Create a Blank Page" instead so let me know if you like it or not and I can make some tweaks.

iamthomasbishop:
@chipsnyder looks solid! Matching other similar offline views across WPiOS is a good call. I'm curious whether we should use the same heading Text ("unable to load...") as we're using on Android. Although consistency with other screens is prob a safe bet. 🤔

One other thing: the separator at the bottom of the navigation bar feels a bit...floaty. Can we bring it up so that the spacing above/below the navigation bar Title is even? If not, we can just hide it here.

chipsnyder:
> the separator at the bottom of the navigation bar feels a bit...floaty.

Ah yeah, I see what you mean. I'll make that adjustment this morning. 

> I'm curious whether we should use the same heading Text ("unable to load...")

We could, although the text "Layouts not available while offline" isn't exactly true. It's probably a small use case but there are flows where I could have cached layouts for one site but not the other, so then it becomes obvious that Layouts could be available while offline they just aren't right now. I thought about that way too much 😅

iamthomasbishop:
> We could, although the text "Layouts not available while offline" isn't exactly true

Ahh okay, good and fair points 😃 Let's roll with this then.

chipsnyder:
Hey @iamthomasbishop, 

Made that update the top and bottom spacing is now equal.
<kdb></kdb>

iamthomasbishop:
@chipsnyder Awesome, much better. Thank you! 🙏

WordPress/Classes/Extensions/UIViewController+NoResults.swift
```diff
@@ -33,6 +33,7 @@ extension NoResultsViewHost where Self: UIViewController {
     func configureAndDisplayNoResults(on view: UIView,
                                       title: String,
                                       subtitle: String? = nil,
+                                      noConnectionSubtitle: String? = nil,
```
mkevins, line 37: Should this be added to the documented parameters?

WordPress/Classes/ViewRelated/Gutenberg/CollapsableHeader/CollapsableHeaderViewController.swift
```diff
@@ -46,9 +46,15 @@ protocol CollapsableHeaderContentsDelegate: class {
 
     /// A public interface to notify the container that the content view is loading content still
     func loadingStateChanged(_ isLoading: Bool)
+
+    /// A public interface to notify the container that the content view is loading content still
```
mkevins: Does this comment description apply? (I.e. I guess the content is not loading at this point?)

WordPress/Classes/ViewRelated/Gutenberg/CollapsableHeader/CollapsableHeaderViewController.swift
```diff
@@ -46,9 +46,15 @@ protocol CollapsableHeaderContentsDelegate: class {
...
+    func displayNoResultsController(title: String, subtitle: String?, resultsDelegate: NoResultsViewControllerDelegate?)
+
+    /// A public interface to notify the container that the content view is loading content still
```
mkevins: Same

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -112,7 +112,24 @@ class GutenbergLayoutPickerViewController: UIViewController, CollapsableHeaderDa
...
+                    self?.headerContentsDelegate?.dismissNoResultsController()
+                case .failure(let error):
+                    self?.handleErrors(error)
+                }
+            }
+        }
+    }
+
+    private func handleErrors(_ error: Error) {
+        guard resultsController?.isEmpty() ?? true else { return }
```
mkevins, line 128: Not suggesting a change here, just curious about the case where there is an error, but `resultsController` is still populated, can such a state be reached (if so, is it unexpected / worth logging)?

mkevins (commented):
Nice work Chip! I tested this via the described flows, and everything is working as expected. It's looking good! :tada: I just had some minor notes about some code comments.

antonis (commented):
Great work Chip! The code LGTM and the test scenarios work as expected 🎉 

Noticed two minor UI issues I'm not sure it is worth tackling.

|There is big bottom margin after a successful retry|The subtitle reappears on rotation|
|---|---|
|||

WordPress/Classes/ViewRelated/Gutenberg/CollapsableHeader/CollapsableHeaderViewController.swift
chipsnyder: Oops nope, this doesn't really apply. I must have planned to just tweak some wording from what was here then got distracted. I'll update this with a more accurate description.

WordPress/Classes/ViewRelated/Gutenberg/CollapsableHeader/CollapsableHeaderViewController.swift
chipsnyder: Same :)

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
chipsnyder: Good question. The data is cached here so the most common use case would be:
1. A user has a good network connection
1. They create a page 
    - This would fetch and cache the results
1. They then go offline
1. They create a second page
    - This would pull from the cache so `resultsController?.isEmpty()` should return false and the user can select from the stored templates.

chipsnyder:
@antonis @mkevins Thank you both for your callouts on this! I made some changes to fix the rotation issue and the added content area at the bottom. This should be ready again when you get time.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
mkevins: Ah, that makes sense. Thanks for the explanation :+1:

mkevins (commented):
I tested again with the recent changes, and it seems to have resolved the bottom margin issue after a successful retry. I was unable to reproduce the rotation issue mentioned (without the latest commits), so I'll defer to Antonis to check if that is resolved by the latest commits.

antonis (approved):
Retested and everything looks good 🎉 
Thank you for fixing those @chipsnyder :)

PR #15143, opened 2020-10-21
Improves MySitesCoordinator.

diegoreymendez:
This PR is part of a [larger upcoming PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15130) that was broken up to make the review easier.

Improves MySitesCoordinator so that it:

- Implements the logic to show a blog's details screen.
- Doesn't really know or care about `WPTabBarController` so that we can avoid a circular dependency, which is good for maintainability of the code.

Also removes the code to show a blog's details from `WPTabBarController` (since it's now handled by `MySiteCoordinator` as mentioned above).

## To test:

When testing keep in mind that you should notice no visible differences between develop and this branch.  The changes in this PR are in preparation of upcoming work, and we should just make sure nothing is broken in the process.

### Test 1:

Use an account with existing blogs.

1. Launch the app and log in if needed.
2. Make sure "My Sites" is the first active tab, and that you can see the blog details for your default / selected blog.

### Test 2:

1. Using iOS's spotlight, search for the name of one of your blogs.
2. Spotlight should show you a result for your blog.  Tap on it.
3. The app will be  launched and you should see the details screen for the blog you searched.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
@@ -22,19 +22,15 @@ class MySitesCoordinator: NSObject {
         super.init()
     }
 
-    private func prepareToNavigate() {
+    func showMySites() {
         becomeActiveTab()
 
         mySitesNavigationController.viewControllers = [blogListViewController]
     }
 
-    func showMySites() {
-        prepareToNavigate()
-    }
-
     @objc
     func showBlogDetails(for blog: Blog) {
-        prepareToNavigate()
+        showMySites()
 
         blogListViewController.setSelectedBlog(blog, animated: false)
     }
--- a/WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
@@ -18,7 +18,7 @@ class MySitesCoordinator: NSObject {
         self.mySitesNavigationController = mySitesNavigationController
         self.blogListViewController = blogListViewController
         self.becomeActiveTab = becomeActiveTab
-        
+
         super.init()
     }
 
--- a/WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
@@ -27,11 +27,7 @@ class MySitesCoordinator: NSObject {
     private func prepareToNavigate() {
         becomeActiveTab()
 
-        if Feature.enabled(.newNavBarAppearance) {
-            mySitesNavigationController.viewControllers = [blogDetailsViewController]
-        } else {
-            mySitesNavigationController.viewControllers = [blogListViewController]
-        }
+        mySitesNavigationController.viewControllers = [blogListViewController]
     }
 
     func showMySites() {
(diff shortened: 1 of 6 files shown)


WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
```diff
@@ -5,20 +5,25 @@ class MySitesCoordinator: NSObject {
     let mySitesSplitViewController: WPSplitViewController
     let mySitesNavigationController: UINavigationController
     let blogListViewController: BlogListViewController
+    let becomeActiveTab: () -> Void
```
frosty, line 8: It might be good for this to also be named `onBecomeActiveTab`, otherwise it sounds a little like it's a closure that will cause this to become the active tab.

frosty (approved):
Works for me! Just one small comment.

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
Gio2018: or `tabWillBecomeActive`, since it seems to happen right before the tab switching?

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
```diff
@@ -27,10 +32,15 @@ class MySitesCoordinator: NSObject {
         prepareToNavigate()
     }
```
Gio2018: seems that this method is only calling another one, perhaps we can just use the inner one?

Gio2018 (approved):
Hey @diegoreymendez , tested on iPad Pro 11 2nd Gen iPadOS 14.0 and iPhoneX iOS 13.7 and worked as expected!
Just left a couple of minor code comments.

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: > otherwise it sounds a little like it's a closure that will cause this to become the active tab.

It actually does just that! 😄 

It requests the coordinator's "owner" (that is `WPTabBarController`) to make "My Sites" the active tab, which is why I chose the name `becomeActiveTab`.

The original code (which is shown as removed in line 21) was:

```swift
WPTabBarController.sharedInstance().showMySitesTab()
```

What this block does isn't much.  It just decouples this class from `WPTabBarController`, so that in theory we could change the tab bar implementation completely without making any change to this class.  It's quite a minor change but I think it's always cleaner to avoid the circular dependency (ie: `WPTabBarController` depending on `MySitesCoordinator` and `MySitesCoordinator` depending on `WPTabBarController`.

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: Addressed in my latest commit.

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: I'll go ahead and merge for now, but if there's any further feedback about this, do let me know here.  I'll still be working around this area of the code, and I'm open to making this better - I'm not fully married to my current solution.

One thing I've been asking myself when writing this code is around ownership.

Which class owns activating a tab?  Should it be `WPTabController` or the coordinators?

If we make a call on that I can shift the code accordingly and document our decision in the code.

PR #15144, opened 2020-10-21
Fix reader_article_detail_more_tapped event

leandroalonso:
This PR fixes the event `reader_article_detail_more_tapped`.

In some circumstances, this event was not being triggered.

### To test

1. Open a post in Reader
2. Tap the More button
3. Check that the event is firing


yaelirub (approved):
👍

PR #15150, opened 2020-10-22
Removes some duplicated methods.

diegoreymendez:
This PR removes some duplicated methods.

We need to test that nothing has broken due to these changes.

### Test 1:

1. Launch the app
2. Go to a test blog
3. Create a test post, and save it as draft.  Use a unique title.
4. Switch to the reader tab.
5. Go to iOS's spotlight search, and type the title of your post.
6. The post should be listed as one of the results.  Tap on it.

The App should launch, and you should see the tab transition to "My Sites", open the list of posts for the blog that owns the post, and the editor should be launched showing the post you selected.

### Test 2:

DISCLAIMER: I could not test this scenario because for some really strange reason, local notifications are NOT coming up for me.  I've spent an enormous amount of time trying to figure out why, but then decided to wrap up with the effort since it was a clear distraction from publishing this PR.  @frosty mentioned he's able to see local notifications so hopefully he can verify my changes.

I'm pretty sure I didn't break the notifications with my changes.  I did try to resolve the issue but quite honestly it's been a big time sink, and I'm quite sure none of my changes caused the breakage.

1. Go to "My Sites"
2. Select your test blog
3. Create a new post.
4. At this point, without closing the editor, go to iOS' Settings > Developer > Network Link Conditioner.
5. Enable the conditioner, and chose the Edge preset.
6. Go back to the editor, type in a title, add an image to the post and publish it right away.  Confirm publishing.
7. As soon as you confirm publishing, switch to the reader tab and background the app.
8. Go back to link conditioner and disable it.
9. Make sure yo DO NOT go back into the App... you should see a "Post Published" notification eventually.  Tap on the notification.

The App should launch, and you should see the tab transition to "My Sites", open the list of posts for the blog that owns the post, and the editor should be launched showing the post you selected.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Spotlight/SearchManager.swift
+++ b/WordPress/Classes/Utility/Spotlight/SearchManager.swift
@@ -402,9 +402,9 @@ fileprivate extension SearchManager {
     func openListView(for apost: AbstractPost) {
         closePreviewIfNeeded(for: apost)
         if let post = apost as? Post {
-            WPTabBarController.sharedInstance().switchTabToPostsList(for: post)
+            WPTabBarController.sharedInstance().mySitesCoordinator.showPosts(for: post.blog)
         } else if let page = apost as? Page {
-            WPTabBarController.sharedInstance().switchTabToPagesList(for: page)
+            WPTabBarController.sharedInstance().mySitesCoordinator.showPages(for: page.blog)
         }
     }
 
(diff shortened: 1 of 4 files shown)


frosty (approved):
I ran through the tests and I think this is fine. The second scenario, launching from a local notification, didn't navigate me to the right place but I don't think that's due to these changes (this code doesn't even get executed).

To fix that I think we may need to either tweak the code that sends the notification (to send a site ID in the notification userInfo), or tweak the code that receives it to look up the post using its core data ID (and hence determine which site to navigate to).

Gio2018 (approved):
Same exact results as @frosty . I also did the same test on `develop` and run the second scenario, and even in that case after tapping the local notification, the app opened but did not navigate to the right place, so that issue it's definitely not related to this PR.

diegoreymendez:
Thanks both!

frosty:
FYI I opened an issue for the navigation problem we uncovered: https://github.com/wordpress-mobile/WordPress-iOS/issues/15163

PR #15151, opened 2020-10-22
Adds the medium size iOS 14 Today Widget

Gio2018:
Ref #15103 

Notes:

- The content of the widget is just a static sample
- At the moment, the distances between the labels are automatically set by the system.

To test:

- install the app on a device or simulator with iOS 14
- add a new widget, scroll down to find WordPress, tap to add and then select the medium size widget.
- make sure the added widget is consistent with the designs in #15103. See also the example below
<p align=center>
 
</p>

PR submission checklist:

- [x] ~~I have considered adding unit tests where possible.~~ in a future PR.
- [x] ~~I have considered adding accessibility improvements for my changes.~~ in a future PR.
- [x] ~~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~~ Not yet released.


WordPress/WordPressHomeWidgetToday/Views/Cards/VerticalCard.swift
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// MARK: - Appearance
+extension VerticalCard {
+
+    private enum Appearance {
+
+        static let titleFont = Font.caption
+        static let titleFontWeight = Font.Weight.semibold
+        static let titleColor = Color("Blue50")
```
frosty, line 34: Should we just store the color itself here?

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetMediumView.swift
```diff
@@ -0,0 +1,41 @@
+ import SwiftUI
+
+ struct TodayWidgetMediumView: View {
+    let content: TodayWidgetContent
+    let siteNameTitle: LocalizedStringKey
```
frosty: I know I suggested we use LocalizedStringKey in the previous PR, but I'm now curious whether this is actually necessary or not 🤔 Does it 'just work' if we just use String?

frosty: Actually, in this _specific_ case I don't think we need it for the site name anyway, as that will never be localized.

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetMediumView.swift
```diff
@@ -0,0 +1,41 @@
...
+                    rightValue: "\(content.visitors)")
+            Spacer()
+            makeRow(leftTitle: likesTitle,
+                    leftValue: "\(content.likes)",
+                    rightTitle: commentsTitle,
+                    rightValue: "\(content.comments)")
+        }
+    }
+    /// Constructs a two-card row for the medium size Today widget
+    private func makeRow(leftTitle: LocalizedStringKey,
```
frosty: Same as above comment, maybe we just need strings here? Not sure

frosty (commented):
Looking great! Just a few small comments :)

WordPress/WordPressHomeWidgetToday/Views/Cards/VerticalCard.swift
Gio2018: Can you clarify? `Color("Blue50")` is already a color?

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetMediumView.swift
Gio2018: So the way it should work is that, if you directly use a string in a `Text`(like `Text("Hello world")`), that would be automatically localized, but if you use a variable, that would need to be `LocalizedStringKey` (some context here: https://developer.apple.com/forums/thread/650492). 
I have a specific issue in the project for the localization, as this is kinda new and I'd like to test it thoroughly before we release.

Gio2018: the site name (which is contained in `TodayWidgetContent`) would not, but the name of the field (`siteNameTitle = "Today"`), would

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetView.swift
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+}
+
+// MARK: - Constants
+extension TodayWidgetView {
+
+    private enum Constants {
+        // Titles
+        static let siteNameTitle: LocalizedStringKey = "Today"
```
ScoutHarris: Can I ask that `siteNameTitle` be renamed? Given the current name, I expect the value to be "Site Name" (i.e. site name title). But it's not, it's actually the widget name. Can it be something like `widgetName` or `widgetTitle`?

Gio2018: yup, done!

mattmiklic (approved):
This looks really good @Gio2018! I'm sorry I'm late reviewing it.

I noticed the Comments section doesn't quite vertically align with the Visitors section above it. Here's a screenshot comparing a screenshot from my 11 Pro Max (left) and the design mockups (right). It looks like it's about 5pt off, if you could adjust that. Otherwise good to go!
Screen Shot 2020-10-27 at 11 14 04 PM

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetView.swift
frosty: Good catch!

WordPress/WordPressHomeWidgetToday/Views/Cards/VerticalCard.swift
frosty: Sorry, I don't know what I was thinking... Disregard this comment! :homerdisappear:

Gio2018:
> This looks really good @Gio2018! I'm sorry I'm late reviewing it.
> 
> I noticed the Comments section doesn't quite vertically align with the Visitors section above it. 

> This looks really good @Gio2018! I'm sorry I'm late reviewing it.
> 
> I noticed the Comments section doesn't quite vertically align with the Visitors section above it. 

Thanks for reviewing @mattmiklic !

The new installable build should have the comments alignment right.

ScoutHarris (approved):
Nice!

:shipit:

mattmiklic:
> The new installable build should have the comments alignment right.

Alignment looks good now, thanks! 🚀

PR #15172, opened 2020-10-27
[TESTING] Unit Controls

lukewalczak:
Fixes #

To test:

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -152,7 +152,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => '46cd6b36d31787a5ac69aeb77fa660f105968a02'
+    gutenberg :commit => 'e156e5ae2d1869f7daab5fe8f4233f1bda687ec7'
 
     ## Third party libraries
     ## =====================
--- a/Podfile
+++ b/Podfile
@@ -152,7 +152,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => '012f51bd3d802b4d7bb20ab80fb4d0a379d6134b'
+    gutenberg :commit => '46cd6b36d31787a5ac69aeb77fa660f105968a02'
 
     ## Third party libraries
     ## =====================
--- a/Podfile
+++ b/Podfile
@@ -152,7 +152,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => '275c0c0fcdd59ad5047bb5cfabb18353da18d5dc'
+    gutenberg :commit => '012f51bd3d802b4d7bb20ab80fb4d0a379d6134b'
 
     ## Third party libraries
     ## =====================
--- a/Podfile
+++ b/Podfile
@@ -152,7 +152,7 @@ target 'WordPress' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :tag => 'v1.39.0'
+    gutenberg :commit => '275c0c0fcdd59ad5047bb5cfabb18353da18d5dc'
 
     ## Third party libraries
     ## =====================
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -450,14 +450,14 @@ DEPENDENCIES:
   - CocoaLumberjack (~> 3.0)
   - CropViewController (= 2.5.3)
   - Down (~> 0.6.6)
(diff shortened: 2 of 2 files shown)


iamthomasbishop:
@lukewalczak This is feeling really nice! Visually, it matches the design pretty much identically. Some quick notes that I wrote down while reviewing:

- Can we append labels to the `unit` ActionSheet/Menu? For example: `px` would become `Pixels (px)`, `vw` would become `Viewport width (vw)`, etc.
- I'm not sure if there's much we can do about this, because AFAIK we're just inheriting web logic, but it's a bit confusing to me that you can set `vw` to > 100%. I assume this is the same situation as discussed previously, where `%` can be > 100% on web. Here's an example: I can end up with something like [this](https://cloudup.com/cZ5AdoTZlsa), which doesn't make a lot of sense to me.
- If it's not too complex, would it make sense to round `px` unit steps to whole numbers? Example: if I have `%` set to `33.333`, change the `unit` to `px` and tap `(-)`, the value is adjusted to `23.333`(px), which seemed kinda weird. I expected it to go in whole number steps from `33.333` to `33`, `32`, `31`, and so on. I think this is maybe in a gray area for other units, like `vw`, where users may still want a decimal value, but `px` seems like a good case for whole numbers. It's not a deal-breaker, just feels a little odd to me.
- Curious: How is the column width on the column preview thumbnails determined (and in turn visualized) for units like `px` and `em`? What I mean is they're fixed values, so how is the width of the column in the preview determined — is it relative to an assumed fixed container size like Xpx? Or is it dynamic based on device screen size? Or some other formula?

One more thing, as a request — and let me know if this is too complex to try quickly because I wouldn't consider it a blocker. I'd like to try an alternate layout for the Stepper (w/ Unit) cells, which would be a two-row layout similar to the Slider cells. Essentially this would bump the column preview, unit input, and stepper buttons to the next line. This should provide more breathing room and also a nicer vertical rhythm in cases where Slider cells and Stepper w/ Unit cells are intermixed. Here's a [quick mockup](https://cloudup.com/c2Kh6FdhOOc) of what that'd look like.

lukewalczak:
Hello @iamthomasbishop I've resolved your comments, however need more info about the second point

> How is the column width on the column preview thumbnails determined (and in turn visualized) for units like px and em? What I mean is they're fixed values, so how is the width of the column in the preview determined — is it relative to an assumed fixed container size like Xpx? Or is it dynamic based on device screen size? Or some other formula?

The value defined in different units is translated to the unit which RN understands and then flexbox is dividing the container width into proportional parts.

**UPDATES**

Android:

parent | children | menu
--- | --- | ---
Screenshot 2020-10-28 at 15 41 21 | Screenshot 2020-10-28 at 15 41 38 | Screenshot 2020-10-28 at 15 41 29

IOS:

parent | children | actionsheet
--- | --- | ---
Screenshot 2020-10-28 at 15 41 45 | Screenshot 2020-10-28 at 15 41 56 | Screenshot 2020-10-28 at 15 41 47

What's more I'm generating new build so please check it and let me know what else should be improved ✌🏼

iamthomasbishop:
@lukewalczak nice work. I'm only seeing 1 issue, 1 minor spacing refinement to the Unit/Stepper cells, and a clarification to my previous second point.

- Tapping `(-)` button on the Stepper is actually _increasing_ the value
- Tiny request: can we add 4px of padding on both top and padding of cells that contain both Stepper + UnitControl so that they're the same height as Slider? ([example](https://cloudup.com/cCZ-q7gHaKB))

> however need more info about the second point

What I meant is that if I set the `vw` value to > 100 (greater than the full width of the viewport), it would be confusing that the preview didn't show as much. I don't know if there's much we can do about it because of the core Columns logic, but wanted to mention it as a potential point of confusion w/ users.

> The value defined in different units is translated to the unit which RN understands and then flexbox is dividing the container width into proportional parts.

Ok, thanks for the context. 👍

lukewalczak:
> Tapping (-) button on the Stepper is actually increasing the value

🤦🏼‍♂️ 

> Tiny request: can we add 4px of padding on both top and padding of cells that contain both Stepper + UnitControl so that they're the same height as Slider? 

Sure, will adjust it

> What I meant is that if I set the vw value to > 100 (greater than the full width of the viewport), it would be confusing that the preview didn't show as much. I don't know if there's much we can do about it because of the core Columns logic, but wanted to mention it as a potential point of confusion w/ users.

I see, but currently on web you can also set the value higher than 100% or 100vw :( However displays values properly - 100% is equal 100vw:

Screenshot 2020-10-28 at 21 56 43

twstokes:

PR #15183, opened 2020-10-28
Track Cards API response

leandroalonso:
This PR tracks if:

1) The API response has arrived
2) The user is still waiting for the content/error to be presented (aka didn't left Discover)

The event fired is intended to be used in a funnel, to check how many users are leaving Discover due to the API taking too long to respond.

### To test

Launch the app:

1. Open Discover
2. Wait for the cards to appear
3. Check that `reader_discover_content_presented` is fired

Launch the app again (to clear cache):

1. Open Discover
2. Immediately switch to other section (like Following)
3. Check that `reader_discover_content_presented` is *not* fired

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -25,7 +25,9 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
     private var selectInterestsViewController: ReaderSelectInterestsViewController = ReaderSelectInterestsViewController()
 
     /// Whether the current view controller is visible
-    private var isVisible = false
+    private var isVisible: Bool {
+        return isViewLoaded && view.window != nil
+    }
 
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -39,16 +41,6 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         displaySelectInterestsIfNeeded()
-    }
-
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-        isVisible = true
-    }
-
-    override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-        isVisible = false
     }
 
     // MARK: - TableView Related
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -115,7 +115,10 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         cardsService.fetch(isFirstPage: true, refreshCount: refreshCount, success: { [weak self] cardsCount, hasMore in
             self?.trackApiResponse()
             success(cardsCount, hasMore)
-        }, failure: failure)
+        }, failure: { [weak self] error in
+            self?.trackApiResponse()
+            failure(error)
+        })
     }
 
     override func loadMoreItems(_ success: ((Bool) -> Void)?, failure: ((NSError) -> Void)?) {
@@ -154,7 +157,7 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
             return
         }
 
-        WPAnalytics.track(.readerCardsFetched)
+        WPAnalytics.track(.readerCardsDidRespond)
     }
 
     // MARK: - TableViewHandler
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -24,6 +24,9 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
 
     private var selectInterestsViewController: ReaderSelectInterestsViewController = ReaderSelectInterestsViewController()
 
+    /// Whether the current view controller is visible
+    private var isVisible = false
```
yaelirub: I recommend turning this into a computed property
```
private var isVisible: Bool {
 return isViewLoaded && view.window != nil
}
```

leandroalonso: I've used `isViewOnScreen()` but it didn't work, this one actually works. Done!

leandroalonso:
@yaelirub ready for another 👀

yaelirub (approved):
Thank you so much for working on this so quickly, @leandroalonso!

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -99,7 +104,13 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         page = 1
         refreshCount += 1
 
-        cardsService.fetch(isFirstPage: true, refreshCount: refreshCount, success: success, failure: failure)
+        cardsService.fetch(isFirstPage: true, refreshCount: refreshCount, success: { [weak self] cardsCount, hasMore in
+            self?.trackApiResponse()
+            success(cardsCount, hasMore)
```
ParaskP7, line 109: 👋 @leandroalonso !

The track event name is named `READER_DISCOVER_CONTENT_PRESENTED` but the way you are using the tracking suggests that you want to track the API response, **NOT** the `trackContentPresented`.

With that in mind, and in order to align our tracking logic with how the tracking is called maybe we want to do something like the below:

```
        cardsService.fetch(isFirstPage: true, refreshCount: refreshCount, success: { [weak self] cardsCount, hasMore in
            success(cardsCount, hasMore)
            self?.trackContentPresented()
            failure(error)
            self?.trackContentPresented()
        })
```

Here, I am assuming that the `success(...)` and `failure(...)` is doing the UI update. Is my assumption correct? If it is correct, maybe we should consider the suggestion above.

ParaskP7:
👋 again @leandroalonso !

I am currently implementing the solution on the Android side.

In addition to my above single comment, I have some questions that I want to clarify before concluding on the solution:
1) What exactly do we want to track?
2) What conclusions do we want to derive from the data and how?
3) Are we only interested in the first response or even load more? PS: In my solution I am assuming that we are only interested in the first response.
4) If we show data from cache and send a request for more up to date data, should we track the event or not? PS: In my solution I am assuming yes.
5) If the user leaves before the request finishes and comes back later, should we track the event or not? PS: In my solution I am assuming yes, but this is a tricky question. For example, if the user leaves the screen and comes back again, quickly, before the data is fetched then the tracking will still occur.

Apologies for the very loaded comment. 😊 

Let me know your thoughts!

ParaskP7:
👋 another time @leandroalonso !

This is now done, at least the naive implementation. 😊 

I have added you and @malinajirka as a reviewer to it: [WordPress-Android#13613](https://github.com/wordpress-mobile/WordPress-Android/pull/13613)

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
leandroalonso: Yup! You're right, the title should be "track content being presented".

We're doing this track to see if the API being slow is making the user leave Discover. I guess we can keep it as it is (even tho because I find it a little bit weird to track API responses) also the iOS event is already being used in funnels.

ParaskP7: Thanks for the reply Leandro! 🌟 

Actually, the title is one of my concerns, my main concern is the order, triggering `track` before or after the `success/failure` functions. Thus, I rewrote the above code snippet by changing the title and order.

What do you think about the order?

leandroalonso: > What do you think about the order?

I'm not very worried about the order, tbh. The only problem I see is that if we have a crash, the track will still be done. But moving to below the `success/failure` won't change much so I can make this change + the renaming. :)

ParaskP7: > I'm not very worried about the order, tbh. The only problem I see is that if we have a crash, the track will still be done. But moving to below the success/failure won't change much so I can make this change + the renaming. :)

For Android the order matters a bit as there is a bit of computation happening after the API response arrives. Thus, I put it below the corresponding `success/failure` and make sure to name it accordingly (`trackContentPresented`), so that I can be as much precise as possible, contextual and tracking wise.

PR #15185, opened 2020-10-28
Today widgets: fix translations

ScoutHarris:
Fixes #n/a

When the `All Time` and `This Week` widgets were implemented, the strings were never actually translated. Per @jkmassel "The problem (I think) is that while we have a way to combine .strings files (ie – to translate strings in a dependency), we don't have the ability to handle more than one target". Since the Today widget already existed, this is the one target that was handled and the strings translated.

After discussing solutions with @jkmassel , for now we'll share the WP `Localizable.strings` with the widgets so they are translated. Even though Today was being translated, for consistency, I've removed the widget translations and it's also now using the WP `Localizable.strings`. 

Note this only affects the strings displayed in the widgets, not the actual widget names.

To recap, the changes:
- `WordPressTodayWidget` - removed localization files.
- `WordPress\Resources\Localizable.strings` - added to 3 widget targets.

Screen Shot 2020-10-28 at 2 54 25 PM

---
_Tip_: sometimes the widgets get... confused... and will display an 'unable to load' message. You can fix this by either:
- Build the widget scheme first, then run the `WordPress` scheme.
- Run the widget scheme directly.

Screen Shot 2020-10-28 at 3 21 02 PM

**To test:**

- Run the app.
- Add the widgets to the Today View.
- Change the device language.
- Verify the text is translated.

| unconfigured-english | unconfigured-spanish |
|--------|-------|

- In the app, set a site to use for the Widgets. (Stats > Widgets > Use this site)

widget_site

- Return to the Today View. 
- Verify the widget text is translated.

| configured-english | configured-spanish |
|--------|-------|

---
PR submission checklist:


Gio2018 (approved):
Thanks for fixing this @ScoutHarris !! Worked as expected
<p align=center>
 
</p>

:shipit: !!

ScoutHarris:
Thanks @Gio2018 !

PR #15188, opened 2020-10-29
Workaround for a crashing bug with Bulgarian keyboards.

### LINKED ISSUE #15187: Crash in iOS 14 when using Bulgarian or Icelandic keyboards.
# Description

The App is crashing when focusing a text field using the Bulgarian or Icelandic keyboards.

Reference: p4a5px-2F1-p2

# Crash reproduction

## Prerequisites

- Make sure your device is running iOS 14.
- Make sure you're logged out before running this test.  It's not specific to the login flow but the steps below use the login flow.

## Steps

1. Add the Bulgarian or Icelandic language as the preferred one in your device.  Make sure the corresponding keyboard is also added.
2. Launch WPiOS
3. Tap "Continue with WordPress.com"
4. Tap on the email field to start writing.  If the Bulgarian / Icelandic keyboard is the selected one the App should crash right away.
5. If the selected keyboard is not Bulgarian / Icelandic, tap on the globe icon in your keyboard until you switch to it.

The app will crash.

diegoreymendez:
Fixes #15187 

Adds a workaround to avoid the crash by replacing text field's `returnKeyType == .continue` with `returnKeyType == .next`.

# Reproducing the original issue

If you want to reproduce the original issue, comment the call to `UITextField.activateWorkaroundForBulgarianKeyboardCrash()` and following the testing steps below.

# To test the fix

Be logged out (this issue can be reproduced in any text field, but being logged out is necessary for the proposed reproduction steps below).

1. Add the Bulgarian or Icelandic keyboard in your devices settings.
2. Launch WPiOS
3. Tap "Continue with WordPress.com"
4. Tap on the email field to start writing. If the Bulgarian / Icelandic keyboard is the selected one you should be able to use it normally and the App should not crash.
5. If the selected keyboard is not Bulgarian / Icelandic, tap on the globe icon in your keyboard until you switch to it.  Make sure it works fine and doesn't crash the App.

# PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/UITextField+WorkaroundContinueIssue.swift
+++ b/WordPress/Classes/Extensions/UITextField+WorkaroundContinueIssue.swift
@@ -1,0 +1,49 @@
+import Foundation
+
+@objc
+extension UITextField {
+
+    /// This method takes care of resolving whether the iOS version is vulnerable to the Bulgarian / Icelandic keyboard crash issue
+    /// by Apple.  Once the issue is resolved by Apple we should consider setting an upper iOS version to limit this workaround.
+    ///
+    /// Once we drop support for iOS 14, we could remove this extension entirely.
+    ///
+    public class func shouldActivateWorkaroundForBulgarianKeyboardCrash() -> Bool {
+        if #available(iOS 14.0, *) {
+            return true
+        }
+
+        return false
+    }
+
+    /// We're swizzling `UITextField.becomeFirstResponder()` so that we can fix an issue with
+    /// Bulgarian and Icelandic keyboards when appropriate.
+    ///
+    /// Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/15187
+    ///
+    @objc
+    class func activateWorkaroundForBulgarianKeyboardCrash() {
+        let original = class_getInstanceMethod(
+            UITextField.self,
+            #selector(UITextField.becomeFirstResponder))!
+        let new = class_getInstanceMethod(
+            UITextField.self,
+            #selector(UITextField.swizzledBecomeFirstResponder))!
+
+        method_exchangeImplementations(original, new)
+    }
+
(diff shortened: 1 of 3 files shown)


ScoutHarris:
Hey @diegoreymendez . I'm still reviewing this, but I was wondering if this should target the frozen branch (i.e. be included in the 16.0 release)?

WordPress/Classes/Extensions/UITextField+WorkaroundContinueIssue.swift
```diff
@@ -0,0 +1,49 @@
...
+    /// Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/15187
+    ///
+    @objc
+    class func activateWorkaroundForBulgarianKeyboardCrash() {
+        let original = class_getInstanceMethod(
+            UITextField.self,
+            #selector(UITextField.becomeFirstResponder))!
+        let new = class_getInstanceMethod(
+            UITextField.self,
+            #selector(UITextField.swizzledBecomeFirstResponder))!
```
ScoutHarris: Can these force unwraps be replaced with `guard`s?

ScoutHarris (changes requested):

diegoreymendez:
> I was wondering if this should target the frozen branch

I've run this through Eli and we agreed it's not clear that this is high impact enough to warrant a hotfix (or frozen-fix) since we've only received one report we could trace back to this.

WordPress/Classes/Extensions/UITextField+WorkaroundContinueIssue.swift
diegoreymendez: Changed.

diegoreymendez:
Ready for another look.

ScoutHarris (approved):
Thanks @diegoreymendez !

Confirmed the app no longer crashes.

:shipit:

diegoreymendez:
Thanks for the review!

Issue #15104, opened 2020-10-14
Today Widget - Adapt existing data for Today Widget to the new architecture


Gio2018:
We want to reuse the existing `TodayWidgetStats` to feed data into the new Today Widget. This will likely require some adaptation to the new SwiftUI architecture.

PR #15195, opened 2020-10-29
Feature/15104 today widget data

Gio2018:
Ref #15104

This PR only prepares the `WordPressHomeWidgetToday` target to use the existing types used for the old Today widget. Also removes unnecessary additional entitlements file.

It does not (yet):
- Refresh data after the first load
- Allow to configure the site
- request fresh data from the backend
- Separate data from the old Today Widget

To test:

- Build run and install either size of the new Today widget
- You should see Today data from whatever site you had set for the old widgets (see example below)

<p align=center>

</p>

PR submission checklist:


frosty (approved):

ScoutHarris:
Hey @Gio2018 .

This is something that can be addressed later, but if a site has no title nothing is displayed. You may want to discuss with @mattmiklic , but maybe we could show the URL if there is no title? Pretty sure we do that in other places in the app.

no_site_title

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetMediumView.swift
```diff
@@ -16,15 +16,15 @@
...
-                           lowerValue: "\(content.likes)")
+                           lowerValue: "\(content.stats.likes)")
                 Spacer()
                 Spacer()
                 makeColumn(upperTitle: visitorsTitle,
-                           upperValue: "\(content.visitors)",
+                           upperValue: "\(content.stats.visitors)",
                            lowerTitle: commentsTitle,
-                           lowerValue: "\(content.comments)")
+                           lowerValue: "\(content.stats.comments)")
```
ScoutHarris: Just a reminder - all the stats values need `.abbreviatedString()` appended. You can either do this now or later, up to you.

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetSmallView.swift
```diff
@@ -11,7 +11,7 @@ struct TodayWidgetSmallView: View {
                 FlexibleCard(axis: .vertical, title: widgetTitle, value: content.siteTitle)
 
                 Spacer()
-                VerticalCard(title: viewsTitle, value: "\(content.views)", largeText: true)
+                VerticalCard(title: viewsTitle, value: "\(content.stats.views)", largeText: true)
```
ScoutHarris: Same here - `.abbreviatedString()` is needed.

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetView.swift
```diff
@@ -44,6 +44,9 @@ extension TodayWidgetView {
 
 struct TodayWidgetView_Previews: PreviewProvider {
     static var previews: some View {
+        TodayWidgetView(content: staticModel)
+            .previewContext(WidgetPreviewContext(family: .systemSmall))
+
```
ScoutHarris, line 49: 👍

ScoutHarris (approved):
Just a couple suggestions. But otherwise, :shipit: !

mattmiklic:
> maybe we could show the URL if there is no title?

Good catch, and good idea @ScoutHarris. I think this is what we should do.

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetMediumView.swift
Gio2018: Thanks for catching this!! Fixed

WordPress/WordPressHomeWidgetToday/Views/TodayWidgetSmallView.swift
Gio2018: Fixed

Gio2018:
Thanks for reviewing @frosty @ScoutHarris @mattmiklic !!

the blank site name should be fixed now, using the site url in case it's empty

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
```diff
@@ -30,9 +30,18 @@ struct Provider: TimelineProvider {
...
-        UserDefaults(suiteName: WPAppGroupName)?.string(forKey: WPStatsTodayWidgetUserDefaultsSiteNameKey) ?? "Site name not found"
+        let defaults = UserDefaults(suiteName: WPAppGroupName)
+
+        if let title = defaults?.string(forKey: WPStatsTodayWidgetUserDefaultsSiteNameKey),
+           !title.isEmpty {
+            return title
+        } else if let url = defaults?.string(forKey: WPStatsTodayWidgetUserDefaultsSiteUrlKey),
+                  !url.isEmpty {
+            return url
+        } else {
```
ScoutHarris: Nitpicky, but these `else`s aren't necessary since each case `return`s.

Gio2018: Yup! Done

ScoutHarris:
> the blank site name should be fixed now, using the site url in case it's empty

Confirmed. Thanks!

Issue #15158, opened 2020-10-23
Today Widget - Tracks


Gio2018:
Add tracks to the Today Widget

ScoutHarris:
Hey @Gio2018 .

One thing I remembered while working on https://github.com/wordpress-mobile/WordPress-iOS/pull/15198. 

The widget tracks don't appear in the Xcode console. However, they do appear in the Tracks Live View. For example, if you search for an existing widget's Event Name, you'll see the results. You can verify the new widget's tracking the same way.

PR #15198, opened 2020-10-30
iOS14 Today widget: add Tracks extension

ScoutHarris:
Ref #15158 

This adds a `Tracks+TodayHomeWidget.swift` extension to contain method to track events.

I added two events and methods that I think will be needed.
- `statsLaunched` - User taps widget to view Stats in the app
- `loginLaunched` - User taps widget to login to the app

I tried to get `statsLaunched` to trigger via `onTapGesture`, but it doesn't seem to work. It may be broken (https://developer.apple.com/forums/thread/652258). So there's nothing to test here. However, since this tracking extension works for the existing widgets, I'm assuming it will for the new one.

To test:
- Review the changes, see if it makes sense.

PR submission checklist:


Gio2018 (approved):
LGTM!

PR #15200, opened 2020-10-31
Merge 16.0 final into develop

jkmassel:
Merges final beta changes and string updates into `develop`.


loremattei (approved):
I looks ok. 
What happened with the version number @jkmassel? I see you bumped it many times during the finalisation of the release.

jkmassel:
Just for future reference:

>  What happened with the version number?

I had to re-finalize the release a few times – there was an issue with an existing tag, then something else (sorry – I don't remember what and I don't have the scrollback in terminal anymore), but it was being really weird. Likely due to settings changed as part of the code freeze to fix CI issues. Shouldn't be a problem moving forward.

PR #15204, opened 2020-11-02
Zendesk: Update to 5.1.1

ScoutHarris:
Fixes #n/a

This updates the `ZendeskSupportSDK` to version 5.1.1. 

To test:

I perused the [release notes](https://developer.zendesk.com/embeddables/docs/ios-unified-sdk/release_notes) between 5.0.0 and 5.1.1, and tested functionality. I saw nothing worth noting. So just some smoke testing should do it.
- Go to Me > Help & Support.
- If you've set a `Contact Email`, verify it's the one you want to use.
- Select `Contact Support`.
- If prompted (if you've not set a `Contact Email`), enter your contact info.
- Submit a request.
- Go to Zendesk, and verify a ticket was created.
- In the app, go to `My Tickets`.
- Verify your newly created issue appears there.

Note: This is listed as a ZD [Known Issue](https://developer.zendesk.com/embeddables/docs/ios-unified-sdk/known_issues): `Missing dSYMs for iOS simulator architecture causes a warning to be emitted for CocoaPods integrators. Note: This warning is emitted only when building apps on iOS Simulators.` I didn't see any warnings, but if you do that's why.

PR submission checklist:


aerych (approved):
Looks good to me! :shipit: 

There were a couple of things I spotted but I'm thinking they aren't related to this change.  Documenting here for now:

The first was, from a fresh install, on the Me > Help & Support screen, if I tapped on the empty Contact Email field I would get a prompt to add an email address, but this version of the prompt would only accept the A-Za-z/ character set.  No numbers, "@", "." or ":" would be accepted.
Screen Shot 2020-11-02 at 4 40 25 PM
Tapping Contact Support prompted me to enter a name and email address as expected. After configuring an email address this way, I could then go back, tap on the Contact Email address row, and then the field would accept all the expected characters.

The second think I noticed was, from a fresh install, after configuring my name and email address and submitting a ticket, the Contact Email field remained empty.  I'd have expected it to refresh.  I needed to tap back to the Me screen, then back to Help & Support before it was populated.  Maybe we're missing a call to update a datasource and reload a tableView after submitting a ticket?

ScoutHarris:
Thanks @aerych ! Yea, the issues you noted aren't related to the SDK. I'll look at those separately. Thanks for finding them!

PR #15211, opened 2020-11-04
Add initial A/B testing support

leandroalonso:
Part of https://github.com/wordpress-mobile/WordPress-iOS-Shared/issues/278

This PR adds a basic API to retrieve experiments and their variations from the ExPlat.

(TTL, Anonymization, Deanonymization are coming in subsequent PRs)

### To test

There are no visual changes, so you can rely on the tests (or debugging) to make sure it's working and please propose any changes you have in mind. :)

Also, you can just add `ExPlat().refresh()` in `WordPressAppDelegate`.

### Usage

My idea is to use it like:

```swift
switch abTesting.experiment("reader") {
    case .treatment:
        // Show treatment option
    case .control:
        // Show control option
    case .unknown:
        // Experiment doesn't exist or wasn't retrieved
    case .other(let value) where value == "my-value":
        // any other value 
}
```

Overall, usage should be way more simpler, like:

```swift
switch abTesting.experiment("reader") {
    case .treatment:
        // Show treatment option
    default:
        // Show control option
}
```

### Wait, shouldn't that be on WordPress-iOS-Shared/Kit?

Yes. My plan is to have that in a separate library so both WPiOS and WooiOS (and any other app) can make use of it. But we have a few issues:

- Tracks protocol is on Shared
- `WPAnalyticsTrackerAutomatticTracks` on the other hand is in WPiOS (and Woo have its own implementation)
- The logic to request a V2 endpoint is in another library, WordPress-Kit
- I don't want to create a dependency between Shared and Kit neither rely on the implementation of the `WPAnalyticsTrackerAutomatticTracks` here

Given this would require a lot more discussion I decided to just go ahead and implement it in WPiOS, then we can manage the better way to extract it. (@astralbodies I could really use your expertise here)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/ABTesting/Assignments.swift
+++ b/WordPress/Classes/Utility/ABTesting/Assignments.swift
@@ -1,0 +1,6 @@
+import Foundation
+
+struct Assignments: Decodable {
+    let ttl: Int
+    let variations: [String: String?]
+}
--- a/WordPress/Classes/Utility/ABTesting/Assignments.swift
+++ b/WordPress/Classes/Utility/ABTesting/Assignments.swift
@@ -1,6 +1,11 @@
 import Foundation
 
+/// Model that contains experiments variations and TTL
+///
 struct Assignments: Decodable {
+    /// Time in seconds until the `variations` should be considered stale.
     let ttl: Int
+
+    /// Mapping from experiment name to variation name.
     let variations: [String: String?]
 }
--- a/WordPress/Classes/Utility/ABTesting/ExPlat.swift
+++ b/WordPress/Classes/Utility/ABTesting/ExPlat.swift
@@ -3,6 +3,8 @@ import Foundation
 class ExPlat: ABTesting {
     let service: ExPlatService
 
+    private let assignmentsKey = "ab-testing-assignments"
+
     init(service: ExPlatService = ExPlatService.withDefaultApi()) {
         self.service = service
     }
@@ -15,13 +17,13 @@ class ExPlat: ABTesting {
             }
 
             let validVariations = assignments.variations.filter { $0.value != nil }
-            UserDefaults.standard.setValue(validVariations, forKey: "ab-testing-assignments")
+            UserDefaults.standard.setValue(validVariations, forKey: self.assignmentsKey)
             completion?()
         }
     }
 
     func experiment(_ name: String) -> Variation {
-        guard let assignments = UserDefaults.standard.object(forKey: "ab-testing-assignments") as? [String: String?],
+        guard let assignments = UserDefaults.standard.object(forKey: assignmentsKey) as? [String: String?],
               case let variation?? = assignments[name] else {
             return .unknown
         }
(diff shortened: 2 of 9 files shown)


WordPress/Classes/Utility/ABTesting/Assignments.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+struct Assignments: Decodable {
```
emilylaguna, line 5: Perhaps add some documentation?

WordPress/Classes/Utility/ABTesting/ExPlat.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    func refresh(completion: (() -> Void)? = nil) {
+        service.getAssignments { assignments in
+            guard let assignments = assignments else {
+                completion?()
+                return
+            }
+
+            let validVariations = assignments.variations.filter { $0.value != nil }
+            UserDefaults.standard.setValue(validVariations, forKey: "ab-testing-assignments")
```
emilylaguna: What about moving this key into a constant?

WordPress/Classes/Utility/ABTesting/ExPlat.swift
```diff
@@ -0,0 +1,38 @@
...
+            completion?()
+        }
+    }
+
+    func experiment(_ name: String) -> Variation {
+        guard let assignments = UserDefaults.standard.object(forKey: "ab-testing-assignments") as? [String: String?],
+              case let variation?? = assignments[name] else {
```
emilylaguna, line 27: 😮 First time I've seen this.

WordPress/Classes/Utility/ABTesting/ExPlatService.swift
```diff
@@ -0,0 +1,42 @@
...
+    let assignmentsPath = "wpcom/v2/experiments/0.1.0/assignments/calypso"
+
+    init(wordPressComRestApi: WordPressComRestApi) {
+        self.wordPressComRestApi = wordPressComRestApi
+    }
+
+    func getAssignments(completion: @escaping (Assignments?) -> Void) {
+        wordPressComRestApi.GET(assignmentsPath,
+                                parameters: nil,
+                                success: { responseObject, httpResponse in
```
emilylaguna: what about this since you're not using it? 

```suggestion
                                success: { responseObject, _ in
```

WordPress/WordPressTest/Test Data/explat-malformed-assignments.json
```diff
@@ -0,0 +1,3 @@
+{
+    malformed
```
emilylaguna, line 2: hey this is malformed 😆

emilylaguna (changes requested):
It seems good, left some comments

WordPress/Classes/Utility/ABTesting/Assignments.swift
leandroalonso: Done in aba35ab9e1b49edd938c2d9b844345af479e796e

WordPress/Classes/Utility/ABTesting/ExPlat.swift
leandroalonso: Done in 08b2b2a22520b25d8ae21a16f25b01573df46b61

emilylaguna (approved):
👍 it looks good to me.

PR #15221, opened 2020-11-04
Remove the Story Post option from Blog Posts list screen with feature flag is off

bjtitus:
The story post option was showing in the Blog Posts screen even when the feature flag was off or the blog didn't support it. These two criteria were added in this PR.

#### Feature Flag
- Navigate to the Blog Posts screen (Blog Details / My Site > Blog Posts) 
- Verify that the Story Post option shows upon tapping the FAB when the stories feature flag is enabled (default in development builds)
Screen Shot 2020-11-04 at 3 32 53 PM

- Turn off the Feature Flag from App Settings > Debug > Stories
Screen Shot 2020-11-04 at 3 28 26 PM

- Return to the Blog Posts screen
- Check that tapping the FAB opens the Post Editor.

#### Blog Support
- Use a site with Jetpack < 8.9 (ping me if you need a site to test with)
- Navigate to the Blog Posts screen (Blog Details / My Site > Blog Posts) 
- Verify that tapping the FAB opens the Post Editor.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostListViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostListViewController.swift
@@ -164,18 +164,23 @@ class PostListViewController: AbstractPostListViewController, UIViewControllerRe
     }
 
     private lazy var createButtonCoordinator: CreateButtonCoordinator = {
-        return CreateButtonCoordinator(self, actions: [
+        var actions: [ActionSheetItem] = [
             PostAction(handler: { [weak self] in
                     self?.dismiss(animated: false, completion: nil)
                     self?.createPost()
-            }),
-            StoryAction(handler: { [weak self] in
-                guard let self = self else {
-                    return
-                }
-                (self.tabBarController as? WPTabBarController)?.showStoryEditor(blog: self.blog, title: nil, content: nil, source: "post_list")
             })
-        ])
+        ]
+        if Feature.enabled(.stories) && blog.supports(.stories) {
+            actions.append(
+                StoryAction(handler: { [weak self] in
+                    guard let self = self else {
+                        return
+                    }
+                    (self.tabBarController as? WPTabBarController)?.showStoryEditor(blog: self.blog, title: nil, content: nil, source: "post_list")
+                })
+            )
+        }
+        return CreateButtonCoordinator(self, actions: actions)
     }()
 
     override func viewDidAppear(_ animated: Bool) {


bjtitus:
@wordpress-mobile/release-managers We'll need to get this into the 16.1 release. Should I add `release-managers` as a reviewer?

aerych (approved):
Heya @bjtitus 👋 
Looks good.  I tested as described and stories were hidden by the flag,  For the Jetpack version I just changed the min supported version in Blog.m when testing. 
I spotted one super edge case: 
- I had tapped on the FAB on the Blog Details screen to confirm Stories was an option in the alert control.
- I went to Me > App Settings > Debug and disabled Stories
- I went back and tapped the Fab again and I still saw the option there. 
- After tapping back to the blog list then back to the blog details the alert control options were updated.

The super edge case is that a user could downgrade Jetpack (or disable it maybe?) and still see that option, if they had left the app running on the blog details screen.
I'm not sure if it's worth looking at too closely (it's really so unlikely) but I thought I'd mention it if there was a quick fix you might have in mind. 
Otherwise :shipit:

aerych:
> Should I add release-managers as a reviewer?

I'm not sure how that label is used actually.  Couldn't hurt?

aforcier:
Just chiming in about this:

> The super edge case is that a user could downgrade Jetpack (or disable it maybe?) and still see that option, if they had left the app running on the blog details screen.
> I'm not sure if it's worth looking at too closely (it's really so unlikely) but I thought I'd mention it if there was a quick fix you might have in mind.

Agreed that it's probably okay to leave it since it's such an edge case - also we'll still end up uploading valid story block markup in that case, so that if they eventually re-update/enable Jetpack the story will automatically work as uploaded. 🤷‍♂️

bjtitus:
@aforcier Do you think it's worth creating an issue to track this? I'll have to do a little more digging to figure out exactly how much mitigation we can do.

AliSoftware:
> Should I add release-managers as a reviewer?

@aerych @bjtitus FYI, see this internal post about how those teams are supposed to be used: paaHJt-1wO-p2
I think this is ok to add the team as reviewer when targeting release branch indeed, as the goal is to have it merged ASAP

aforcier:
> @aforcier Do you think it's worth creating an issue to track this? I'll have to do a little more digging to figure out exactly how much mitigation we can do.

I'd say it's okay to leave it, it's a super edge case and probably not the only thing that's wonky if you downgrade/disable Jetpack with the app running.

PR #15226, opened 2020-11-05
AB testing: schedule a refresh based on TTL

leandroalonso:
Part of wordpress-mobile/WordPress-iOS-Shared#278

This PR:

* Adds the feature to schedule a refresh given the `tll` that the ExPlat response returns
* Uses `URLSession` instead of `WordPressComRestApi`
* Has an object configuration, so it is completely decoupled from WPiOS

The latter two will help to extract it to https://github.com/Automattic/Automattic-Tracks-iOS 

### To test

Again, rely on the code/tests. :)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/ABTesting/ExPlat.swift
+++ b/WordPress/Classes/Utility/ABTesting/ExPlat.swift
@@ -14,11 +14,12 @@ class ExPlat: ABTesting {
         return ttlDate.timeIntervalSinceReferenceDate - Date().timeIntervalSinceReferenceDate
     }
 
-    private(set) var scheduleTimer: Timer?
+    private(set) var scheduledTimer: Timer?
 
     init(configuration: ExPlatConfiguration,
          service: ExPlatService? = nil) {
         self.service = service ?? ExPlatService(configuration: configuration)
+        subscribeToNotifications()
     }
 
     /// Only refresh if the TTL has expired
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Utility/ABTesting/ExPlat.swift
```diff
@@ -37,4 +70,28 @@ class ExPlat: ABTesting {
...
+        if ttl > 0 {
+            scheduleTimer?.invalidate()
+
+            /// Schedule the refresh on a background thread
+            DispatchQueue.global(qos: .background).async { [weak self] in
+                guard let `self` = self else {
+                    return
+                }
+
+                self.scheduleTimer = Timer.scheduledTimer(withTimeInterval: self.ttl, repeats: true) { [weak self] timer in
```
emilylaguna: I've had a lot of issues with timers in the past, I think you should invalidate and reset when the app goes to the background or comes to the foreground.

emilylaguna (changes requested):
Seems good, left a concern about timers though.

WordPress/Classes/Utility/ABTesting/ExPlat.swift
leandroalonso: Good point!

Changed in 629db96b2fd19e90bb70da9e652c421eeb521c7e

emilylaguna (approved):
👍 🚢 Looks good to me.

PR #15199, opened 2020-10-30
Improvements to My Site tab.

diegoreymendez:
Swaps the list of sites and blog details when the `newNavBarAppearance` flag is on.

Should not change anything when the flag is off.

## Known limitations:

When the `newNavBarAppearance` feature flag is ON there's currently no way to switch sites.  I've disabled it temporarily until I can implement the site switching.

## To test:

The most important testing that can be done is when the feature flag `newNavBarAppearance` is off, since that's the experience that our users will have in the release version of the App.

Please take a moment to thoroughly test and review the code in this PR.

### Test 1:

Feature flag `newNavBarAppearance` ON.
Use an account that has at least 1 blog.

1. Tap "My Site" in the tab bar
2. Make sure the first VC shown the blog details for your last used blog, or your main blog.
3. Make sure there's no option to go to the list of sites just yet (coming up in my next PR).

### Test 2:

Feature flag `newNavBarAppearance` ON.
Use an account that has no blogs.

1. Tap "My Site" in the tab bar
2. Make sure you see a VC that says you have no blogs.
3. Create a site, make sure once you're done you see the blog details for the blog you created.
4. Delete the site, make sure once you're done you see a no-results view stating you have no blogs.

### Test 3:

Feature flag `newNavBarAppearance` OFF.
Use an account that has at least 1 blog.

1. Tap "My Site" in the tab bar
2. Make sure the first VC shown the blog details for your last used blog, or your main blog.
3. Make sure you can tap on the back button in the navigation bar to pick a different site.
4. Switch to another site and make sure it works fine.

### Test 4:

Feature flag `newNavBarAppearance` OFF.
Use an account that has no blogs.

1. Tap "My Site" in the tab bar
2. Make sure the first VC shown is a no-results saying you have no sites.
3. Create a site.
4. Make sure you see the list of blogs with the site you created.
5. Delete the site.
6. Make sure you see the no results view again.

### Additional notes:

In all tests, make sure you also tap in the user's icon, and that the Me VC comes up just fine.

## PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15227, opened 2020-11-05
Some changes to BlogDetailsViewController+MeButtonTests for reusability

diegoreymendez:
This is part of [a much larger upcoming PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15199) that was split to make the reviews more atomic.

Makes some changes to BlogDetailsViewController+MeButtonTests.swift for reusability.

The reason I'm doing these changes is because I'll be needing to show the Me scene from a new view controller, and I wanted to avoid having to duplicate this snippet a third time:

https://github.com/wordpress-mobile/WordPress-iOS/blob/00f6ac4ee88b2631e0302e644acf44b4f2ed9e9b/WordPress/Classes/ViewRelated/Blog/Blog%20%2B%20Me/UIBarButtonItem%2BMeBarButton.swift#L5-L32

To do so, I just made a generic extension for any `UIViewController` to be able to call `addMeButtonToNavigationBar` instead.

## Important:

Please read my inline comment in the sources.

### Test 1:

Make sure you're logged in with an account that has at least 1 blog.

1. Launch the App
2. Go to the details for a blog

Make sure the user's icon is loaded and visible in the top-right corner as always.

3. Tap on the user's icon.

Make sure the Me scene is shown.

### Test 2:

Make sure you're logged with an account that has no blogs.

1. Launch the App
2. Go to the list of sites.

Make sure the user's icon is loaded and visible in the top-right corner as always.

3. Tap on the user's icon.

Make sure the Me scene is shown.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
@@ -1,39 +1,22 @@
 import Gridicons
 import UIKit
 
-
 /// Add a UIBarButtonItem to the navigation bar that  presents the Me scene.
-extension BlogDetailsViewController {
+extension UIViewController {
     @objc
-    func presentHandler() {
-        meScenePresenter.present(on: self, animated: true, completion: nil)
-    }
-
-    @objc
-    func addMeButtonToNavigationBar() {
-        navigationItem.rightBarButtonItem = UIBarButtonItem(email: blog.account?.email,
-                                                            target: self,
-                                                            action: #selector(presentHandler))
+    func addMeButtonToNavigationBar(email: String?, meScenePresenter: ScenePresenter) {
+        navigationItem.rightBarButtonItem = UIBarButtonItem(
+            email: email,
+            action: {
+                meScenePresenter.present(on: self, animated: true, completion: nil)
+            })
     }
 }
-
-extension BlogListViewController {
-    @objc
-    private func presentHandler() {
-        meScenePresenter.present(on: self, animated: true, completion: nil)
-    }
-
-    @objc
-    func addMeButtonToNavigationBar(with email: String) {
-        navigationItem.rightBarButtonItem = UIBarButtonItem(email: email,
-                                                            target: self,
-                                                            action: #selector(presentHandler))
-    }
-}
-
 
 /// methods to set the gravatar image on the me button
 private extension UIBarButtonItem {
+    typealias TapAction = () -> Void
+
     /// gravatar configuration parameters
     struct GravatarConfiguration {
         static let radius: CGFloat = 32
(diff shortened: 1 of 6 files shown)


WordPress/WordPressTest/BlogDetailsViewController+MeButtonTests.swift
```diff
@@ -58,40 +62,4 @@ class BlogDetailsViewControllerTests: XCTestCase {
         XCTAssertEqual(meButton.accessibilityLabel, TestConstants.meButtonLabel)
         XCTAssertEqual(meButton.accessibilityHint, TestConstants.meButtonHint)
     }
-
-    func testPresentMeOnButtonTap() {
```
diegoreymendez, line 62: I know removing a unit test isn't great, but an explanation on this:

1. I really tried to write a test that made sense here but failed.  I spent around 2 hrs today trying to figure this out, because...
2. ...there's no good support in UIKit to programmatically tap on a view or trigger a gesture recognizer.  In fact because of this I also noticed that...
3. ...the original test isn't really testing a tap either.  It's calling `#selector(BlogDetailsViewController.presentHandler)` in a kind of indirect way.

I can still drop the PR but I feel it makes the code a bit more flexible in case we want to show the Me scene from a new view controller, and avoid code duplication.

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
```diff
@@ -1,39 +1,22 @@
...
     @objc
-    func addMeButtonToNavigationBar() {
-        navigationItem.rightBarButtonItem = UIBarButtonItem(email: blog.account?.email,
-                                                            target: self,
-                                                            action: #selector(presentHandler))
+    func addMeButtonToNavigationBar(email: String?, meScenePresenter: ScenePresenter) {
+        navigationItem.rightBarButtonItem = UIBarButtonItem(
+            email: email,
+            action: {
+                meScenePresenter.present(on: self, animated: true, completion: nil)
```
frosty, line 15: We need a weak self here, or we end up with a retain cycle on BlogDetailViewController.

frosty (commented):
Looks good to me! Just one comment regarding a retain cycle.

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
diegoreymendez: 'Tis done.

If you're happy with the PR feel free to approve, I'll wait for @Gio2018 's review anyway since I'm modifying some of his code.

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
```diff
@@ -1,39 +1,26 @@
...
 
-
 /// Add a UIBarButtonItem to the navigation bar that  presents the Me scene.
-extension BlogDetailsViewController {
-    @objc
-    func presentHandler() {
-        meScenePresenter.present(on: self, animated: true, completion: nil)
-    }
-
+extension UIViewController {
```
Gio2018, line 5: totally agree with removing the repeated method; one minor thing is that we are implicitly saying that any `UIViewController` can add a Me button. We could use a protocol and then conform only the viewControllers we want, but then again we would need to implement those methods (unless we use a default implementation, which is less then ideal anyway)

WordPress/WordPressTest/BlogDetailsViewController+MeButtonTests.swift
Gio2018: Agreed. Testing a tap is more a matter of UI testing. I think it's fine to remove this.

Gio2018 (approved):
Hey @diegoreymendez , this looks good to me. Just a couple of comments/observations.

:shipit:

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
diegoreymendez: I kinda tried a similar approach but if felt a bit like an "unnecessary" extra step, since we'd be providing a default implementation of this method, and I don't think the VC would ever override that implementation.

That's not to say I feel too strongly about this.  I'd be game to implement the protocol if that feels cleaner to you.

Thoughts?

WordPress/WordPressTest/BlogDetailsViewController+MeButtonTests.swift
diegoreymendez: Thanks, @Gio2018 !

WordPress/Classes/ViewRelated/Blog/Blog + Me/UIBarButtonItem+MeBarButton.swift
Gio2018: I think it's fine to leave as it is for now. Also, it's pretty clear what the method does so it should not cause any problem.

diegoreymendez:
Thanks @frosty and @Gio2018 for the review!

PR #15234, opened 2020-11-06
Remove ExPlat from WPiOS

leandroalonso:
This PR removes the `ExPlat` from `WPiOS`.

The feature will be available here: https://github.com/Automattic/Automattic-Tracks-iOS/pull/155


emilylaguna (approved):
:shipit:

PR #15235, opened 2020-11-06
Update issue-triage to reference medium priority instead of normal

frosty:
This PR simply updates the references in issue-triage to mention medium priority instead of normal, to match the labels we use.


diegoreymendez (approved):
Looks good!

PR #15236, opened 2020-11-06
Configure ExPlat for use

leandroalonso:
This PR configure `ExPlat` to be used in `WPiOS`.

Note that the service will only start if there are experiments in the `ABTest` enum. (thanks @chipsnyder for the idea!)

### How to check for an experiment

```
switch ABTest.experimentName.variation {
    case .treatment:
        // Show treatment option
    case .control:
        // Show control option
    case .unknown:
        // Experiment doesn't exist or wasn't retrieved
    case .other(let value) where value == "my-value":
        // any other value 
}
```

### To test

There's not feature per se to test. Please, only review the code. You can also use a proxy tool to intercept requests and play with them.

PR submission checklist:


WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -40,6 +40,28 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
...
+            var userAgent: String? {
+                return WPUserAgent.wordPress()
+            }
+
+            var anonId: String? {
+                return UserDefaults.standard.string(forKey: "TracksAnonymousUserID")
+            }
+        }
+
+        return ExPlat(configuration: WordPressExPlatConfiguration())
```
leandroalonso: My idea is to comment this line and only enable it when there are experiments running.

This avoids unnecessary requests.

chipsnyder: I wonder if we can take a small step away from commenting this line on/off but maybe instead use an enum or struct to A) provide a key for the experiment B) maintain a list of running experiments. 

So a base enum could look like this:
```
enum ExPlatExperiment: String, CaseIterable {
    /// This is a default experiment here to support a base enum
    case unknown = "FallbackExperiment"
}

extension ExPlatExperiment {
    static var hasRunningExperiments: Bool {
        return ExPlatExperiment.allCases.count > 1
    }
}
```

And when we're running an experiment it would be:
```
enum ExPlatExperiment: String, CaseIterable {
    /// This is a default experiment here to support a base enum
    case unknown = "FallbackExperiment"

    /// Running experiments
    case myTest = "my-awesome-test"
}

extension ExPlatExperiment {
    static var hasRunningExperiments: Bool {
        return ExPlatExperiment.allCases.count > 1
    }
}
```

In these cases `unknown` is just there because Swift doesn't like an empty enum. We could use `hasRunningExperiments` to turn the network call off if desired.

----
If you like that idea but don't like the `unknown` item we can remove `unknown` by going to a struct like:

```
struct ExPlatExperiment {
    /// Running experiments
    static let myTest = "my-awesome-test"
    static let runnningExperiment = [myTest]
}

extension ExPlatExperiment {
    static var hasRunningExperiments: Bool {
        return ExPlatExperiment.runnningExperiment.count > 0
    }
}
```

chipsnyder (commented):
Hey @leandroalonso, I was taking a look at this PR as we have a gutenberg-mobile [project](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2726) coming up that could benefit from this. I'm not sure if I missed any other conversations on the progress here so I hope you don't mind but I dropped a thought to our comment here. 

If there is any way I can help out let me know :)

leandroalonso:
Hi @chipsnyder!

I loved this idea. There are still a few details I need to address in the main PR (https://github.com/Automattic/Automattic-Tracks-iOS/pull/155) and I'll address this one too! Thank you so much for your thoughts!

leandroalonso:
@chipsnyder I implemented your idea, let me know what do you think. :)

chipsnyder (commented):
I think this looks great @leandroalonso! Thanks for adding those changes.

leandroalonso:
@astralbodies do you want to take a look at this PR or should I assign someone else? Thanks in advance!

astralbodies:
@leandroalonso I'll review! I meant to do this one yesterday too, and then my brain broke again. 😄

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -127,6 +127,8 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
             startObservingAppleIDCredentialRevoked()
         }
 
+        ABTest.start()
```
astralbodies: Should this be called earlier, possibly before the welcome screen would get presented? Just thinking that any UI may be subject to an experiment, unless we're not going to support anonymous/logged-out experiments.

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
```diff
@@ -242,6 +242,15 @@ - (void)setLoggedInID:(NSString *)loggedInID
     [[NSUserDefaults standardUserDefaults] setObject:loggedInID forKey:TracksUserDefaultsLoggedInUserIDKey];
 }
 
+- (NSString *)token
```
astralbodies: We might run into issues here with the default WordPress.com account being fetched out of Core Data ... at least I think we ran into issues and that's why we used the NSUserDefaults key above to cache the username.

I think it might be safe to call the keychain util directly here since we have the username.

https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Models/WPAccount.m#L81

So maybe the method takes a username property and returns the token? Just thinking of a way to make life less complicated. 😄

astralbodies (changes requested):
Had a comment about simplifying the token fetching to not get us in trouble potentially. 😝

WordPress/Classes/System/WordPressAppDelegate.swift
leandroalonso: Moved to before the screen is presented: 4a6484c4eb2c2c25b1c8f1f415f842392982b327

However, an experiment very early (like when the app is just started) is something that we'd need to discuss better. Since it's a request it can take a few ms and the first screen will probably be presented before.

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
leandroalonso: Done in 044668eb40274e8b1e9715985311c5fdfa11f4e7 and 9f8c374089fb916d037c4072f9ec66d3b5f319f5

Let me know if that is what you had in mind. I tested and the request is being made with the token. :)

leandroalonso:
@astralbodies ready for another review! 👀

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
astralbodies: Looks good to me @leandroalonso!

astralbodies (approved):
Ready for merge after the conflict is resolved.

WordPress/Classes/System/WordPressAppDelegate.swift
astralbodies: Good point! I've really had it in my head that ExPlat won't work for a lot of the login screens unless it's later in the flow.

PR #15243, opened 2020-11-06
Fix direct call to CoreDataManager

Gio2018:
CoreDataManager should not be referenced directly until we have the Swift Core Data stack feature flagged, otherwise in a scenario where the feature flag is turned off, there will be duplicated stacks.
Fixes #NA

To test:
- Build with Xcode and verify that the memory graph does not contain more than two `NSManagedObjectModel` instances
- verify that the console does not contain logs like the following one:
`CoreData: warning: Multiple NSEntityDescriptions claim the NSManagedObject subclass 'Notification' so +entity is unable to disambiguate.`

PR submission checklist:


frosty (approved):
Looks good to me!

Gio2018:
Thanks for reviewing @frosty !! @jkmassel , going to merge this one in a few seconds, thank you!!

PR #15245, opened 2020-11-06
Improves some initializers.

diegoreymendez:
@Gio2018 had found a scenario in which the App was crashing when tapping on the Me button.

I could not reproduce the crasher, but I did identify that the default initializer for `BlogListViewController` and `BlogDetailsViewController` are never setting property `meScenePresenter`.

This PR improves the initializers so that it's always set.

## To test:

Couldn't reproduce the original issue, but I suggest attempting to launch the app:

a. With an account with several blogs, and tapping on the user's icon in a site details.
b. With an account with no blogs and tapping on the user's icon in the sites list.
c. Trying to do state restoration in both scenarios A and B.

## PR submission checklist:


frosty (approved):
:shipit:

PR #15231, opened 2020-11-05
[Home Page Picker] Connect Choose and Skip Actions in site creation

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2721

**Related PR**
- `WordPressKit-iOS`: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/299

📓 This handles the Blog flag as well and skip button. There are two backend changes that are pending that I'm waiting on to make sure this is fully testable.
## To test:
<details>
<summary><strong>To disable or enable the development version of Home Page Picker</strong></summary>

- Open the app from the build that allows FeatureFlags such as a PR build or a local development build
    - By default, the modal layout picker will be disabled.
    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
        - Toggle "Site Creation: Home Page Picker" to enable or disable the picker
        - <kdb><a href="https://user-images.githubusercontent.com/3384451/97343011-5f032980-185d-11eb-9030-06c17f732fec.gif"></a></kdb>

</details>

When enabled, the first step of the Site Creation flow will be replaced with the Home Page Picker flow.

### Create a Site with Site Design
1. Navigate to the Create WordPress.com site
    - My Sites > ➕ > Create WordPress.com site
1. Select a design
1. Tap on "Choose"
1. Search for and Choose a domain
    - **Expect** to see `.wordpress.com` subdomains only unless you selected an option from the test `Create a Blog`
1. Tap "Create Site"

**Expect** Your site to be created

### Create a Site by skipping site design
1. Navigate to the Create WordPress.com site
    - My Sites > ➕ > Create WordPress.com site
1. Select "Skip"
1. Search for and Choose a domain
1. Tap "Create Site"
**Expect** Your site to be created

### Create a Blog
1. Navigate to the Create WordPress.com site
    - My Sites > ➕ > Create WordPress.com site
1. Select a design that has been flagged as a blog
    <details>
    <summary>Examples</summary>
    <kdb><a href="https://user-images.githubusercontent.com/3384451/98397556-87451200-202d-11eb-92f6-c42e86a1a7d2.png"></a></kdb>
    </details>
1. Tap on "Choose"
1. Search for and Choose a domain
    - **Expect** to see `.wordpress.com` and `.blog` subdomains.
1. Tap "Create Site"
**Expect** Your site to be created

## Validate existing flow:

[...]

PR #15247, opened 2020-11-06
[Home Page Picker] Analytic Events

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2723

**Related PR**
- `WordPress-iOS-Shared`: https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/279

📓 This PR implements the suggested approach mentioned here: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2723#issuecomment-723246016
📓This is built off of https://github.com/wordpress-mobile/WordPress-iOS/pull/15231 so this build can be used to test the changes there as well.

## To test:
<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.

</details>

When enabled, the first step of the Site Creation flow will be replaced with the Home Page Picker flow.

For these test cases in addition to verifying in our internal Analytics tool you can validate the event by filtering the console logs on `🔵 Tracked:` 

### enhanced_site_creation_site_design_viewed
**Expect** to see `enhanced_site_creation_site_design_viewed` fire when the screen loads

### enhanced_site_creation_site_design_selected
1. Select a Design
1. Tap on "Choose"
**Expect** to see `enhanced_site_creation_site_design_selected ` fire with the property `template` filled out with the `slug` for the design.

### enhanced_site_creation_site_design_skipped
1. Tap on "Skip"
**Expect** to see `enhanced_site_creation_site_design_skipped` fire.

### error
1. Disconnect your device from the network
**Expect** to see `enhanced_site_creation_error_shown` with `error_info` populated with a description of the error that occurred.

## Additional Context: 

This is a continuation of the work done in the [Home Page Picker Project](https://github.com/wordpress-mobile/gutenberg-mobile/labels/Project%3A%20Home%20Page%20Picker). 

## PR submission checklist:


WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
```diff
@@ -665,6 +665,24 @@ + (TracksEventPair *)eventPairForStat:(WPAnalyticsStat)stat
...
+            break;
+        case WPAnalyticsStatEnhancedSiteCreationSiteDesignPreviewViewed:
+            eventName = @"enhanced_site_creation_site_design_preview_viewed";
+            break;
+        case WPAnalyticsStatEnhancedSiteCreationSiteDesignPreviewLoading:
+            eventName = @"enhanced_site_creation_site_design_preview_loading";
+            break;
+        case WPAnalyticsStatEnhancedSiteCreationSiteDesignPreviewLoaded:
+            eventName = @"enhanced_site_creation_site_design_preview_loaded";
+            break;
```
chipsnyder, line 685: I included the event names for the preview events in preparation for when we add that screen. This will just help prevent the need to create another library version

mkevins: Good idea :smile:

WordPress/Classes/ViewRelated/Site Creation/DesignSelection/SiteDesignContentCollectionViewController.swift
```diff
@@ -120,10 +122,12 @@ class SiteDesignContentCollectionViewController: CollapsableHeaderViewController
             return
         }
         let design = siteDesigns[selectedIndexPath.row]
+        SiteCreationAnalyticsHelper.trackSiteDesignSelected(design)
```
mkevins, line 132: Not sure if the key could ever not be present, but should we have a guard here?

mkevins (commented):
Great work Chip! I tested, and observed the logs, and everything works as expected. I just had one question about a potential `nil` guard.

Unrelated to the PR, I was wondering, do we want to register these events?

WordPress/Classes/ViewRelated/Site Creation/DesignSelection/SiteDesignContentCollectionViewController.swift
chipsnyder: > Not sure if the key could ever not be present

I'm not 100% sure which key you mean based on where the comment is appearing in GH. So let me try to address both lines and see what you think :) 

If you're talking about the `selectedIndexPath.row` here for the array then we are unwrapping that here on the lines above.

https://github.com/wordpress-mobile/WordPress-iOS/blob/fd8ee82c295a792cd705dd9b7ac6d7346f99c2eb/WordPress/Classes/ViewRelated/Site%20Creation/DesignSelection/SiteDesignContentCollectionViewController.swift#L120-L123

I thought about adding an additional error logging but it would have been a wordsmithed version of: "I have no idea how this could happen but if it does here's a log" 😄 If we hit that scenario we treat it like a skip. If we wanted to track who falls under that flow we would probably still be able to do that since selecting a design and skipping both have an analytics call. So the lack of either of those events could be a separate funnel.  

---

For `design` if the `selectedIndexPath` path is beyond the bounds of the array (which would cause a crash) that would have to mean that a successful network request occurred, and then happened again and received different data. In the current state of the code, the API results are only fetched once and held in local memory until the view is destroyed. So there isn't really a code path that would allow this. `#famousLastWords` lol

mkevins: > "I have no idea how this could happen but if it does here's a log"

:smile: 

My apologies .. I selected the wrong line :man_facepalming: . I meant the `design`, I was wondering if the row was not present, if `trackSiteDesignSelected` would choke on a `nil`. But your explanation makes sense, thanks for clarifying! LGTM!

mkevins (approved):
Nice work Chip!

antonis (approved):
Nice work @chipsnyder 👍 
The code LGTM and behaves as expected!

PR #15252, opened 2020-11-06
[Home Page Picker] Poll while thumbnail is not generated

chipsnyder:
**Related PR**
- `WordPressUI-iOS`: https://github.com/wordpress-mobile/WordPressUI-iOS/pull/75

📓 This is based off of https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/279

## Description
The backend service can sometimes take a bit of time to generate thumbnail images if the image either A) hasn't been accessed in a while or B) the combination of screen size and language hasn't been cached yet. 

This implements a polling mechanism to retry fetching the thumbnail images in order to not show the "generating images" screenshot. 

## Testing notes
I _believe_ the images are cached based on a CDN server so you might be able to reuse some of the languages I used for testing. But just in case here are the ones where I already generated a cache for: `ar`, `en`, `es`, `de`, `fr`, `he`, `id`, and `it`. To make sure it's a supported language I started with the list of supported languages based on the [localized strings](https://github.com/wordpress-mobile/WordPress-iOS/tree/develop/WordPress/Resources) then tested the code against the URL: 

`https://public-api.wordpress.com/rest/v1.1/nux/starter-designs?type=mobile&scale=2&preview_width=160&language=is`

So if it helps: `ja` and `ko` are two language codes I haven't used yet that return screenshot URLs with the same language code.

## To test:
<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.

</details>

When enabled, the first step of the Site Creation flow will be replaced with the Home Page Picker flow.

### Test
**Expect** to see the ghost images for a short time then see the images load

## Additional Context: 

This is a continuation of the work done in the [Home Page Picker Project](https://github.com/wordpress-mobile/gutenberg-mobile/labels/Project%3A%20Home%20Page%20Picker). 

## PR submission checklist:


WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/Collapsable Header Collection View Cell/CollapsableHeaderCollectionViewCell.swift
```diff
@@ -21,6 +21,9 @@ class CollapsableHeaderCollectionViewCell: UICollectionViewCell {
         }
     }
 
+    /// The throttle the requests to the imageURL polling if needed.
+    private let throttle = Scheduler(seconds: 1)
```
chipsnyder, line 25: I set the throttle to 1 second here. This has been resulting in about 5 or 6 attempts before the image successfully loads. I'm not sure what the right balance here is though, I can see bumping this up to maybe 2 or 3 seconds. 

Does anyone have a strong opinion here?

mkevins: > resulting in about 5 or 6 attempts

If I understand correctly, this means around 5-6 seconds before a successful response (i.e. 5-6 tries, 1 per second)? I think it sounds reasonable, though I'm curious about whether there are multiple kinds of errors fall into the category. Also, curious about the generation step and / or cache miss. Is it a timeout issue? If so, can anything be done on the other end to keep it alive?

chipsnyder: > there are multiple kinds of errors fall into the category.

One error that could trigger these conditions would be if the CDN or service is malfunctioning. We saw that with one server while testing. They added some uptime logging to better catch that case. From the return result though it looked exactly the same as the "expected" failure case. 

>Is it a timeout issue? If so, can anything be done on the other end to keep it alive?

That's a good question, @enejb was investigating other improvements to this service still but I'm not sure which paths he's explored. A hope would be that if we discovered those paths after a release then we could turn it on and use this as a backup in case that didn't work well.

chipsnyder: > One error that could trigger ...

So maybe we should add a retry cap at say 10 tries which would be excessive for what I'm seeing in testing but wouldn't keep pinging for a URL that won't load. Since this is done in a collection view cell we would probably also need some parent data object to keep track of the retry attempts, that way the counter doesn't restart if you scroll away and come back. WDYT?

mkevins: That seems like a reasonable way to mitigate hammering the server, but I wonder if it would be worth the added complexity in practice. I think the way you have it makes sense, and 1 second seems reasonable, IMO.

One thing I noticed while testing is that it seems _all_ the images load at once, which was surprising. First, a single ghost image "pulsed", then afer a few seconds, several ghost images "pulsed" for several seconds (maybe close to 10 seconds total), and finally, all the images loaded nearly at once (excepting the ones off-screen, which loaded only once scrolled into view).

My hunch is that the API side is where to focus for improving the timing, and I think your throttle amount is a good start, and can be adjusted if needed as usage (in testing) ramps up.

mkevins (approved):
Nice work Chip, LGTM!

WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/Collapsable Header Collection View Cell/CollapsableHeaderCollectionViewCell.swift
chipsnyder: > I wonder if it would be worth the added complexity in practice.

I agree I don't think this would be worth the effort either. When we found the issue in testing our backend teams added monitoring on the server to catch it. So if that issue were to occur again we'd be able to address it there. 

> One thing I noticed while testing is that it seems all the images load at once, which was surprising.

What you're seeing here is actually the progression of the API calls. I can break down what's happening if you're interested:

>First, a single ghost image "pulsed"

This is the app trying to call and get the list of available designs. We use a single cell as a placeholder to help indicate "loading"

> several ghost images "pulsed" for several seconds (maybe close to 10 seconds total)

This is we know what designs are available but we don't have the thumbnails yet. For 10 seconds that would mean the thumbnail was still be generated on the server. Then that's when you see the images load.

If we see issues while testing we could look into limiting the number of active image downloads but that's another one that I don't think is worth the effort IMO.

mkevins: Right, that confirms what I thought about the sequence of api calls. :+1: I guess I expected that the thumbnails would each load at varying times, progressively replacing the ghost placeholders, and wondered whether it was a coincidence that they all loaded at once (i.e. I'd have thought that once the URLs were requested in parallel, the generation step would have had varying response times for the different designs). In any case, certainly not a blocker here :smile:

antonis (approved):
Great work @chipsnyder 👍 
The code LGTM and the app performs as expected 🎉

PR #15254, opened 2020-11-07
Adding Categories to Prepublishing bottom sheet 

yaelirub:
This PR adds categories row to prepublishing bottom sheet.

Previously the Prepublishing navigation controller determined a fixed height `.contentHeight(Constants.height)` in the bottomsheet. That worked well for screens that mostly maintain their height (like PostTagPickerViewController) but since the categories table view has a dynamic height, we have to update the container view whenever the height of the presented changes. 
Because of that I updated the navigation collapsedHeight to be calculated (.intrinsicHeight) based on the content of the presented and thats why I added the preferredContentSize override in the navigation's view controllers . 

To test:
1. Go to the editor 
2. Open a draft post 
3. Click More (3 dots) 
4. Tap Publish 
5. See categories row in bottom sheet
6. Tap Categories 
7. See Post Categories screen
8. Tap + 
9. See add categories screen 
10. Try adding, removing, checking, unchecking categories
11. See `editor_post_category_changed <via: prepublishing_nudges>` event fired
11. Repeat 2-10 on iPad
12. tap on Tags and other rows 

prepublishing_categories

PR submission checklist:


leandroalonso (approved):
@yaelirub works as described!

One question tho, Tags and Categories are slightly different and I think this might led to some confusion: while Tags have a "Done" button, Categories have a "+".

I don't think this is a blocker but we might want to change that later and make them similar.

yaelirub:
> but we might want to change that later and make them similar

Yes. The current implementation does not follow the proposed designs but like you said, not a blocker :)

PR #15262, opened 2020-11-10
Fix crash where UIBarButtonItem is not a UIView

yaelirub:
Sentry crash - https://sentry.io/share/issue/dcf3dbd9e962429d92512da13c9256f9/

UiBarButtonItem does not subclass a UIView hence the crash when trying to set the presentation controller source view to it.

Reverting the barItem's image initialization with a button one

To test:
1. Open a reader article on iPad 
2. Tap on the More menu - 3 vertical dots
3. See the popover and no crash

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailViewController.swift
@@ -576,7 +576,6 @@ private extension ReaderDetailViewController {
 
         let button = barButtonItem(with: icon, action: #selector(didTapMenuButton(_:)))
         button.accessibilityLabel = Strings.moreButtonAccessibilityLabel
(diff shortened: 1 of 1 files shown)


emilylaguna (changes requested):
I tested on iOS and iPad and did not have any crashing issues.
✅  iPhone 11 Pro - iOS 14
✅  iPad Air - iOS 14
✅  iPad Air - iOS 12.4
✅  iPhone 5s - 11.4

I noticed this does cause a minor issue with the alignment of the images as seen below:

Screen Capture on 2020-11-10 at 11-04-18

What do you think about trying to adjust the alignment? I'm okay with it either way.

yaelirub:
Since it's a hot fix I'm ok with the current alignment

emilylaguna (approved):
Sounds good to me! :shipit:

PR #15270, opened 2020-11-11
Fix translator builds

jkmassel:
Fixes an issue with translation release builds – it still tried to remove python 2 before starting, which caused it to fail.

**To Test:**
- Ensure that the translation review build worked

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -232,7 +232,6 @@ jobs:
       - run: 
           name: Install other tools
           command: |
-            brew unlink python@2
             brew install imagemagick
             brew install ghostscript
             curl -sL https://sentry.io/get-cli/ | bash
--- a/Scripts/fastlane/Fastfile
+++ b/Scripts/fastlane/Fastfile
@@ -464,11 +464,11 @@ import "./ScreenshotFastfile"
     ios_merge_translators_strings(strings_folder:"../WordPress/Resources/")
     sh("cd ../.. && ./Scripts/extract-framework-translations.swift")
 
-    xcbuild(
-        workspace: "../WordPress.xcworkspace",
+    gym(
         scheme: "WordPress",
+        workspace: "../WordPress.xcworkspace",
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -212,7 +212,6 @@ jobs:
       - run: 
           name: Install other tools
           command: |
-            brew unlink python@2
```
loremattei, line 235: This looks good. This has been probably broken for a while 🤔

Scripts/fastlane/Fastfile
```diff
@@ -464,11 +464,11 @@ import "./ScreenshotFastfile"
...
     sh("cd ../.. && ./Scripts/extract-framework-translations.swift")
 
-    xcbuild(
-        workspace: "../WordPress.xcworkspace",
+    gym(
         scheme: "WordPress",
+        workspace: "../WordPress.xcworkspace",
         configuration: "Debug",
-        xcargs: "-sdk iphonesimulator SYMROOT='/var/tmp/'"
+        destination: "platform=iOS Simulator,name=iPhone 11",
```
loremattei, line 471: Hmm... I'm not sure I understand this change. We need the x86 build in /var/tmp because we upload it to Appetize (see the next line in the script).

loremattei: Or, we have to update the next line if we build to a different folder.

loremattei (changes requested):

Scripts/fastlane/Fastfile
jkmassel: Heh sorry – left in the middle of this – I thought the python change was the only one required here. Need to move it to `gym` or else it fails the same way the other one did. I'll take a look at this again for review.

Sorry for the spam – I'll move it back to `draft` until it's good to go 😓

jkmassel:
Replaced by https://github.com/wordpress-mobile/WordPress-iOS/pull/15706

PR #15281, opened 2020-11-11
Large Titles: new blog details header and site switcher.

diegoreymendez:
This PR depends on: https://github.com/wordpress-mobile/WordPress-iOS/pull/15687

This PR implements some changes for Large Titles.  It's targetting the feature branch.

Changes:

- Partially implements the new blog details header view.  This is definitely not pixel perfect and some functionality (such as tapping on the site link) is missing.  The main reason this was done was to introduce the chevron that allows for the actual site switching.
- Implements the site switcher.

## Known issues:

- The new blog details header view isn't complete.  It looks particularly off in iPad.  I've decided to leave that work out of this PR to work more atomically, and since this PR already introduces so much.

## To test:

The most important thing is to test that these changes don't break anything for the end users when the `newNavBarAppearance` flag is turned off.

All tests should be performed both in iPhone and iPad.

### Test 1:

`newNavBarAppearance` - Feature flag OFF
Account with no sites.

Make sure on launch you see the list of your sites with no results.

### Test 2:

`newNavBarAppearance` - Feature flag OFF
Account with 1 site.

Make sure on launch you see the blog details for your only site.

### Test 3:

`newNavBarAppearance` - Feature flag OFF
Account with multiple sites.

Make sure on launch you see the list of your sites.  Tapping on any site should open it.

### Test 4:

`newNavBarAppearance` - Feature flag ON
Account with no sites.

Make sure on launch you see a no results view.

### Test 5:

`newNavBarAppearance` - Feature flag ON
Account with 1 site.

Make sure on launch you see the blog details for your only site and the stats are shown on the details pane.

### Test 6:

`newNavBarAppearance` - Feature flag ON
Account with multiple sites.

Make sure on launch you see the selected blog details for your only site and the stats are shown on the details pane.
Tap on the site switcher.  Test the site picker it in any way that could potentially break things.

### Test 7:

`newNavBarAppearance` - Feature flag ON
Account with 1 site

Delete the site.
Recreate a site.
Make sure the whole flow looks OK, ui-wise.

## PR submission checklist:


diegoreymendez:
Listing this as a reminder of what needs to be fixed.

Currently in iPad having no sites looks really bad with the flag on.

Screen Shot 2021-01-14 at 17 23 53

EDIT: this looks actually pretty similar to what we have in `develop`.  Thoughts on this @osullivanchris ?

Screen Shot 2021-01-14 at 18 08 35

WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/ActionRow.swift
```diff
@@ -89,8 +89,9 @@ class ActionRow: UIStackView {
 
         self.init(arrangedSubviews: buttons)
 
-        distribution = .equalCentering
-        spacing = Constants.minimumSpacing
+        distribution = .equalSpacing
+        //spacing = Constants.minimumSpacing
+        //backgroundColor = .blue
```
Gio2018: can we remove the commented out code?

WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/SiteIconView.swift
```diff
@@ -73,13 +73,20 @@ class SiteIconView: UIView {
         }
     }
 
-    override init(frame: CGRect) {
+    init(frame: CGRect, padImage: Bool = true) {
```
Gio2018: how about passing the `UIEdgeInsets` instance directly here? You could also have one single `UIEdgeInsets` constant that wraps the four existing and the initializer would just be

`init(frame: CGRect, insets: UIEdgeInsets) {`
  `button.addSubview(imageView)`
  `button.pinSubviewToAllEdges(imageView, insets: insets)`
`button.addTarget(self, action: #selector(touchedButton), for: .touchUpInside)
         button.translatesAutoresizingMaskIntoConstraints = false`
`}`

or, if you don't want callers to bother with `UIEdgeInsets.zero`

`init(frame: CGRect, insets: UIEdgeInsets? = nil)`

and

`button.pinSubviewToAllEdges(imageView, insets: insets ?? UIEdgeInsets.zero)`

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListDataSource.swift
```diff
@@ -82,6 +82,10 @@ class BlogListDataSource: NSObject {
 
     @objc let recentSitesMinCount = 11
```
Gio2018, line 83: I know this is pre-existing, but since we are here, can we add some comment on why this magic number is 11?

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListDataSource.swift
```diff
@@ -338,6 +342,11 @@ extension BlogListDataSource: UITableViewDataSource {
             case .loggedIn:
                 cell.accessoryType = .none
             default:
+                guard showsDisclosureIndicator else {
```
Gio2018: how about `cell.accessoryType = showDisclosureIndicator ? .none : .disclosureIndicator`

Gio2018 (commented):
Hey @diegoreymendez , the tests work as expected; I had noticed though that the site picker on iPhone (11 Pro Max Device, in my case) it's somewhat hard to tap: would it be possible to increase the tappable area a bit?

I also left a couple of code comments.

osullivanchris:
> EDIT: this looks actually pretty similar to what we have in develop. Thoughts on this @osullivanchris ?

Hey I can think of lots of better ways to display this, as I'm sure you can too, but it depends if we're ok diverging from the page structure in iPhone. For example I think it would be better if
- we moved the empty state to the right hand panel
- we removed the left/right split and displayed the message centred in the full width content area

But both require some specific treatment for iPad. So depends if we are ok with that? Think @frosty and I discussed this before, so pinging him in case he had another solution...I can't remember as it was a while ago.

frosty:
@diegoreymendez @osullivanchris Regarding the empty sites list looking bad on iPad – when no sites are present, the left pane is supposed to take up the full width of the screen. However, a previous iOS update broke that functionality, and neither myself nor @ScoutHarris have been able to fix it. The only thing I can think at this stage to improve it would be to replace the splitview entirely in this situation with just a navigation controller, and swap it back when sites are present. I know this probably isn't particularly straightforward, though.

frosty (commented):
Code looks good to me, and site switching is working great. But I'm seeing some slight functional discrepencies when logging in and out with the feature flag disabled.

1. I logged out of an account with multiple sites and into an account with no sites – I still saw the previous detail view for a site that no longer exists in the app:

Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-01-18 at 14 51 15

2. Probably the same cause, I logged out of an account with no sites and into an account with multiple sites:

Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-01-18 at 14 44 10

In this situation I would expect the primary site to be selected and the stats screen for it to be visible.

WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/ActionRow.swift
diegoreymendez: Removed!

WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/SiteIconView.swift
diegoreymendez: Good observations.  Changed!

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListDataSource.swift
diegoreymendez: Honestly I'm not sure what to add here - the thing is a lot of the code around my changes could be better documented / improved / etc but the balance to avoid scope creep is a hard one.

If there's anything specific you'd like me to write, can you expand a bit?

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListDataSource.swift
diegoreymendez: Changed.

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListDataSource.swift
diegoreymendez: (The thing is that even by looking at the code, the meaning of that variable is not immediately clear to me...)

Gio2018: that's ok, no big deal since it was already there 😄

diegoreymendez:
@Gio2018 - Addressed most of your feedback, and added a question in the remaining observation you raised.

@frosty - Good catch, I've worked to try and resolve those issues today but they're proving to be tricky to manage.  I expect to be able to have an update tomorrow.

diegoreymendez:
@frosty - I pushed some code to fix the refreshing issues on logout / login.

Can I ask you (and @Gio2018) to test this PR again?

diegoreymendez:
@osullivanchris - I've triggered an installable build of this PR in case you want to check it out.

osullivanchris:
> @osullivanchris - I've triggered an installable build of this PR in case you want to check it out.

Great thanks! I've had a look. Generally starting to look good! Along with the things discussed on slack, here are a couple things I think we'll need to tweak. But fine to get to them whenever in your process:

- the nav bar needs a divider or something on scroll. Maybe worth trying the system material background with the blur here too.
- some padding needed between the bottom of the header area with Quick Actions, and the start of the list.

frosty (commented):
I've taken another look, and I would say on the whole things are looking improved – the issues I mentioned before are resolved, and I can't see any issues with the feature flag turned off (I tested both iPhone and iPad).

I am however seeing some issues with the flag on, but as long as we merge with the flag disabled, I think we could address those separately.

**Safe areas**

One smaller one – the quick start items and new site switcher don't respect the safe area inset on a notched device in landscape:

Simulator Screen Shot - iPhone 12 Pro - 2021-01-27 at 13 28 38

(hard to see in a screenshot, but these would be partially covered by the notch)

**Split view issues on iPad**

There are a couple of fairly major split view issues on iPad, when resizing the split view.

- Start with the app full screen, and go to My Site
- Add another app in multitasking, and resize the app so that WP is 1/2 or 1/3 of the screen (so the split view is collapsed)
- Expand the split view again
- You should see that the secondary pane remains collapsed into the primary pane. Here's a video: https://cln.sh/oHgBdw

- Also, follow the same steps as above, but while the split view is collapsed switch sites
- When you expand the split view, you'll see that the right pane is dimmed (initially no view controller is selected, either). Selecting a different entry from the site detail screen will switch the view controller, but it will remain dimmed and there's no way to restore it. Here's a video: https://cln.sh/sFLaVx

Regarding the dimming, I think we can completely disable the dimming when this feature flag is enabled. That was in place to dim the detail view while you were on the site list ion the left pane and switching site, but that's not possible with the new picker.

Gio2018 (commented):
Hey @diegoreymendez , I saw the same issues mentioned by @frosty with the feature flag on (also tested on iPhone 11 Pro Max and saw the dimming on the right when in landscape).

I agree that we could merge with the feature flag off for now.

Issue #15238, opened 2020-11-06
EXC_BREAKPOINT: account > blog > initWithManagedObjectContext: > sharedInstance


jkmassel:
**Sentry Url**: https://sentry.io/share/issue/aacb3ad1dcc14e3cb9ffc9fb49e74cd2/
**User Count**: 1827
**Count**: 3031
**First Release**: 15.9.0.1
**First Seen**: 2020-10-13T06:48:50Z
**Last Seen**: 2020-11-06T16:00:46Z
**24 Hours**: 123
**30 Days**: 2984
**LogID:** 9E7A7E30-EA6C-46B5-A42E-FA1F92E112FA

```
EXC_BREAKPOINT: account > blog > class > initWithManagedObjectContext: > sharedInstance
  File "once.h", line 84, in _dispatch_once
  File "Classes/ViewRelated/System/WPTabBarController.m", line 83, in +[WPTabBarController sharedInstance]
  File "BlogDetailHeaderView.swift", line 82, in BlogDetailHeaderView.toggleSpotlightOnSiteIcon
  File "BlogDetailHeaderView.swift", line 51, in BlogDetailHeaderView.blog.didset
  File "<compiler-generated>", line null, in BlogDetailHeaderView.blog.didset
  File "<compiler-generated>", line null, in BlogDetailHeaderView.blog.setter
  File "<compiler-generated>", line null, in BlogDetailHeaderView.blog.setter
  File "Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m", line 369, in -[BlogDetailsViewController viewDidLoad]
  File "Classes/ViewRelated/Blog/Blog List/BlogListViewController.m", line 809, in -[BlogListViewController setSelectedBlog:animated:]
  File "Classes/ViewRelated/Blog/Blog List/BlogListViewController.m", line 786, in -[BlogListViewController setSelectedBlog:]
  File "Classes/ViewRelated/System/WPTabBarController.m", line 193, in -[WPTabBarController blogListNavigationController]
  File "Classes/ViewRelated/System/WPTabBarController.m", line 339, in -[WPTabBarController blogListSplitViewController]
  File "Classes/ViewRelated/System/WPTabBarController.m", line null, in -[WPTabBarController tabViewControllers]
  File "Classes/ViewRelated/System/WPTabBarController.m", line 111, in -[WPTabBarController init]
[...]

PR #15282, opened 2020-11-12
Replace instance of QuickStartTourGuide inside WPTabBarController with separate singleton

Gio2018:
Ref #15238

the above referenced crash shows two calls to `[WPTabBarController sharedInstance]` in the same stack trace. The crash is possibly caused because the second call happens when the singleton is not yet fully initialized.

The second call to `[WPTabBarController sharedInstance]` is made by `QuickStartTourGuide`, which has an instance in the `WPTabBarController` singleton.

To avoid this, and in general to reduce calls to `[WPTabBarController sharedInstance]`, this PR separates `QuickStartTourGuide` in its own singleton. Singletons are less than ideal and should be avoided when possible, but separating `QuickStartTourGuide` at this stage might make simpler a future (quasi)singleton-free redesign.

To test:

- Add a new site
- when prompted for help, tap "Yes help me"
- Go through all the steps of the quick start guide and make sure they work as expected

PR submission checklist:


diegoreymendez (approved):
Works and looks good to me.

Issue #15285, opened 2020-11-13
Reader Comments: Comment cell can sometimes show an mismatched avatar


aerych:
We've received a report and an example of a reader comment cell showing a mismatched avatar.  See p1605192953097200-slack-wpmobile.  Per the report it was not able to be reproduced. 

One theory is this is an edge case where a cell is being reused while there is an inflight image download that, when complete, replaces a correct image that had been previously retrieved from the image cache. 

### Steps to reproduce the behavior
Unknown, but likely requires scrolling the comment list with a particular sequence of cells, having those cells be reused in a certain order, with corresponding images cached and not cached. 

We might try canceling any inflight downloads when a cell is reused as a fix and use this issue to track future occurrances.

leandroalonso:
@aerych this was fixed?

Given it's quite an edge case and we haven't been hearing complaints about it I think we might want to close this issue.

PR #15286, opened 2020-11-13
Reader: As a precaution, cancel previous downloads when configuring cells

aerych:
Refs #15285 

This PR attempts to address the curious issue reported in #15285 by canceling any inflight image downloads when configuring a reader cell for a new comment.  I remain somewhat suspicious of [the download code](https://github.com/wordpress-mobile/WordPressUI-iOS/blob/98b454a912e5f2b7e39525bc7442efaafbeb04a2/WordPressUI/Extensions/UIImageView%2BNetworking.swift#L82) in the UI pod but haven't been able to suss out the issue with its logic (assuming there is one).  We should leave #15285 open to track future reports in case this patch doesn't do the trick. 

To test:
Apply the patch. 
Smoke test reader comments.  Confirm avatars still load as expected.

@ScoutHarris could I trouble you for a quick peek at this one? 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentCell.swift
@@ -160,7 +160,7 @@ class ReaderCommentCell: UITableViewCell {
         guard let comment = comment else {
             return
         }
-
+        avatarImageView.cancelImageDownload()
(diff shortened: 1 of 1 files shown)


ScoutHarris (approved):
LGTM!

:shipit:

aerych:
Thank you!

PR #15290, opened 2020-11-13
ios 14 today widget refresh policy

### LINKED ISSUE #15289: Today Widget - Add refresh policy
Widgets will have to refresh data periodically. 

From the designs:

"Apple’s suggestion for a similar type of data is to update hourly"

We should then set a configurable update interval, that defaults to 1 hour

Gio2018:
Fixes #15289 

Adds a refresh policy to the new Today Widget for iOS 14

To test:
- Install one or more instances of the widget
- make sure that the random number used for testing (`WordPressHomeWidgetToday.swift`, line 54) updates after the set time interval. This will be the only number showing multiple times in the widget (see example below)

<p align=center>

</p>

**Notes**
1. The current refresh interval is set to 1 hour (`WordPressHomeWidgetToday.swift`, line 79). To test more efficiently, set this interval to a reasonable value. Consider that if it's too low (e.g. 1 minute) iOS will probably not update that often. There seems to be a lower limit of a few minutes at least
2. According to the documentation https://developer.apple.com/documentation/widgetkit/timelineprovider, to test more precisely, the widgets should be detached from Xcode

PR submission checklist:

### DIFF
--- a/WordPress/WordPressHomeWidgetToday/Model/TodayWidgetContent.swift
+++ b/WordPress/WordPressHomeWidgetToday/Model/TodayWidgetContent.swift
@@ -2,7 +2,7 @@ import WidgetKit
 
 
 struct TodayWidgetContent: TimelineEntry {
-    let date = Date()
+    let date: Date
     let siteTitle: String
     let stats: TodayWidgetStats
 }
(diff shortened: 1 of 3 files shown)


frosty (approved):
Works for me! I used quite short refresh interval, and it didn't always seem precise as to when it would update, but it _did_ update.

PR #15300, opened 2020-11-15
Fix force-unwrapped optional for SuggestionsTableView in NotificationDetailsViewController

jkmassel:
Fixes https://sentry.io/share/issue/02558ce4c4ee42f6bd3d3936da263663/ – a blocker for App Store release.

To test: This is a bit of a tricky bug – it's rather rare, so difficult to reproduce. However, the stack trace points to the line that's affected, and it seems like a safe bet that somehow the UI state is invalid, so allowing the `SuggestionsTableView` to be a real optional should prevent a crash, possibly at the expense of some missing UI in invalid states.

To be clear – this isn't necessarily a _fix_ for the problem – more of a band-aid to avoid it so that we can ship this release.

When looking at this, I was initially concerned this may be related to one of the IUOs for the constraints (ie – when the view is removed from the superview it's blowing up somewhere internal), but I don't _think_ that's the case – the stack trace doesn't go into UIKit internals which makes me think it's the view that's the problem, not the constraint. But feel free to check my work on that! :) 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
@@ -529,7 +529,7 @@ private extension NotificationDetailsViewController {
 private extension NotificationDetailsViewController {
     func attachSuggestionsViewIfNeeded() {
         guard shouldAttachSuggestionsView, let suggestionsTableView = self.suggestionsTableView else {
-            self.suggestionsTableView?.removeFromSuperview()
+            suggestionsTableView.removeFromSuperview()
             return
         }
 
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
@@ -54,7 +54,7 @@ class NotificationDetailsViewController: UIViewController {
 
     /// Reply Suggestions
     ///
-    @IBOutlet var suggestionsTableView: SuggestionsTableView!
+    @IBOutlet var suggestionsTableView: SuggestionsTableView?
 
     /// Embedded Media Downloader
     ///
@@ -436,7 +436,7 @@ extension NotificationDetailsViewController {
     func setupSuggestionsView() {
         guard let siteID = note.metaSiteID else { return }
         suggestionsTableView = SuggestionsTableView(siteID: siteID, suggestionType: .mention, delegate: self)
-        suggestionsTableView.translatesAutoresizingMaskIntoConstraints = false
+        suggestionsTableView?.translatesAutoresizingMaskIntoConstraints = false
     }
 
     func setupKeyboardManager() {
@@ -528,8 +528,8 @@ private extension NotificationDetailsViewController {
 //
 private extension NotificationDetailsViewController {
     func attachSuggestionsViewIfNeeded() {
-        guard shouldAttachSuggestionsView else {
-            suggestionsTableView.removeFromSuperview()
+        guard shouldAttachSuggestionsView, let suggestionsTableView = self.suggestionsTableView else {
+            self.suggestionsTableView?.removeFromSuperview()
             return
         }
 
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -528,8 +528,8 @@ private extension NotificationDetailsViewController {
 //
 private extension NotificationDetailsViewController {
     func attachSuggestionsViewIfNeeded() {
-        guard shouldAttachSuggestionsView else {
-            suggestionsTableView.removeFromSuperview()
+        guard shouldAttachSuggestionsView, let suggestionsTableView = self.suggestionsTableView else {
+            self.suggestionsTableView?.removeFromSuperview()
```
AliSoftware, line 532: You `guard let` this just above, so better use the local guarded unwrapped binding directly here

```suggestion
            suggestionsTableView.removeFromSuperview()
```

AliSoftware (commented):
Just did _code_ review from phone without actually testing the fix on device (nor comparing it with the crashlog stack trace) so will let someone who actually looked at stacktrace and tested the fix to approve, but LGTM from code POV

guarani:
Regarding the original code (not this PR), it seems odd to me that `NotificationDetailsViewController` — which uses a Storyboard — instantiates the suggestions table view manually instead of letting it be instantiated via the Storyboard: 

https://github.com/wordpress-mobile/WordPress-iOS/blob/eb82431172bad28b67d6a968a5d5556db485b445/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift#L437-L438

@jkmassel I was looking at under what condition `suggestionsTableView` is `nil` and it looks like one scenario is if `note.metaSiteID` is `nil`. Would `metaSiteID`, which comes from the `Notification.meta` object, be something that comes from the notification payload received from APN? If so, could it ever be missing from the payload?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -528,7 +528,7 @@ private extension NotificationDetailsViewController {
 //
 private extension NotificationDetailsViewController {
     func attachSuggestionsViewIfNeeded() {
-        guard shouldAttachSuggestionsView else {
+        guard shouldAttachSuggestionsView, let suggestionsTableView = self.suggestionsTableView else {
             suggestionsTableView.removeFromSuperview()
```
jleandroperez: Note here! we should move the `guard let suggestionsTableView =...` unwrap above.

Why: `removeFromSuperview` might execute against the "non unwrapped" Suggestions TableView, leading to this crash:

Screen Shot 2020-11-16 at 11 54 03 AM

jleandroperez (changes requested):
@jkmassel Looks great!! sending you a tiny comment there.

Thanks for doing this Jeremy!!

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
jleandroperez: The code that comes inside the **else braces** does not reference the guarded unwrapped, actually.

We're gonna need to either move the guard above (suggested below), ooor restore the `?`, otherwise the crash still happens.

Thank you both gentlemen!!

guarani:
@jkmassel what do you think about removing the `IBOutlet` here as well as the view in the `Notifications.storyboard`?

https://github.com/wordpress-mobile/WordPress-iOS/blob/eb82431172bad28b67d6a968a5d5556db485b445/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift#L57

`suggestionsTableView` is already positioned using Autolayout in code:

https://github.com/wordpress-mobile/WordPress-iOS/blob/eb82431172bad28b67d6a968a5d5556db485b445/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift#L539-L542

I can give this a quick test after lunch and a meeting, but wanted to share here before this is merged.

jkmassel:
@Guarani – I looked at this after submitting the PR too and I think that's a good idea. But because this is a release branch change (essentially a hot fix), I'd prefer not to change anything more than we need to in order to solve _this_ crash. However, I think it might be worthwhile to look at a larger refactor of this class (I was also thinking in terms of refactoring for immutability) at some point for sure :)

jleandroperez (approved):
Fix verified! Thanks for doing this @jkmassel !!!

:shipit:

guarani:
SGTM @jkmassel!

> However, I think it might be worthwhile to look at a larger refactor of this class (I was also thinking in terms of refactoring for immutability) at some point for sure :)

I created a "tech debt" issue to track this: https://github.com/wordpress-mobile/WordPress-iOS/issues/15304

PR #15302, opened 2020-11-16
Fix status bar override for app store screenshots

### LINKED ISSUE #15265: UI Tests: Status bar override cuts off first character of date and time
As noted in https://github.com/wordpress-mobile/WordPress-iOS/pull/15108#pullrequestreview-510827441 there's a problem with the status bar override we do in the UI tests (for screenshots).

The `override_status_bar` setting used when generating screenshots should set the simulator date/time to `9:41 AM Tue Jan 9` but it shows `:41 AM ue Jan 9` instead:

iPad Pro (12 9-inch) (2nd generation)-1-dark-Editor

rachelmcr:
Fixes #15265

We were previously using `SimulatorStatusMagic` to control the status bar in app store screenshots. We're now using the `override_status_bar` setting in Fastlane to control the status bar, and our old setup was breaking the status bar overrides. This removes `SimulatorStatusMagic` since it's no longer needed.

### Screenshots

Notice the status bar changes in the before/after screenshots below:

Before|After
-|-
iPad Pro (12 9-inch) (2nd generation)-1-dark-Editor|iPad Pro (12 9-inch) (2nd generation)-1-dark-Editor

### To test

1. Navigate to the fastlane directory: `cd Scripts/fastlane`
2. Generate screenshots for at least one locale: `bundle exec fastlane screenshots language:"en-US"`
3. Confirm the status bar shows the expected details: date & time `9:41 AM Tue Jan 9`, WiFi active, full cellular connection, full battery

### Release

PR submission checklist:


jkmassel (approved):
Nice catch with this – they must've been conflicting.

A few next steps for this though:
- The simulator list in `fastlane` isn't valid for Xcode 12 anymore – we'll likely need to move these to the 12 series devices.
- The screenshots didn't generate for me on iOS 14 devices – the text fields aren't selected.

I don't think these need to happen as part of this, but if we're looking at these again in the near future it might be worth addressing those things? WDYT?

rachelmcr:
Oh, huh. I used iOS 14 with the existing simulator list and it worked fine to generate the screenshots; I wonder what's up there. But I agree about moving to the 12 series devices, so we can work on that when we next get a chance. Thanks!

Issue #15256, opened 2020-11-09
My site: Unable to select first blog in list of My Sites


ctarda:
The first (top) site in the list of sites in the My Sites tab does not react to taps.

<Img src="https://user-images.githubusercontent.com/2722505/98504250-7eed1100-2291-11eb-8142-7bd42490df39.gif" width="350"/>

### Expected behavior
I should be able to select the first (top) site in the list of sites.

### Actual behavior
I can not select the first site in the list of sites. Other sites can be selected.

### Steps to reproduce the behaviour
1. Navigate to My Sites
2. Tap the first site on the list. Nothing happens
3. Tap a different site. The app navigates to the "site details" view (the one that provides access to stats, pages, posts, media...)

##### Tested on [device], iOS [version], WPiOS [version]
Tested on iPhone 12 Pro, iOS 14.1, WPiOS 16.1.0.3
Tested on iPhone 12 Pro simulator, iOS 14.1, `develop`

guarani:
I've experienced this too. This is especially annoying when searching for a site and you want to pick the first search result.
A "workaround" is to pull to refresh, and while refresh takes place, tap the first cell 😅.

guarani:
FWIW, there was recent work on the site list refresher here: https://github.com/wordpress-mobile/WordPress-iOS/pull/15136

diegoreymendez:
If you wait long enough, does the cell become tappable again?

That's what I'm seeing here.

guarani:
> If you wait long enough, does the cell become tappable again?
> 
> That's what I'm seeing here.

It depends, if I'm searching for a site and there is one search result, this first cell never becomes tappable (I waited ~20 seconds). If I cancel the search, the first cell is still not tappable, and if I then pull to refresh, the first cell becomes tappable again.

startuptester:
Checked on my iPhone 12 Pro in 16.0, 16.1 and 16.2 (beta)
iOS 14.2 

I am able to reproduce this, but I had to open site details for a different site first, then tap "Switch To" and come back to the list in order to trigger the bug. Repeating those steps, it happens 1/10 times and the cell does not become tappable after at least 1 minute.

ScoutHarris:
Fixed with https://github.com/wordpress-mobile/WordPress-iOS/pull/15320 and https://github.com/wordpress-mobile/WordPress-iOS/pull/15325.

PR #15320, opened 2020-11-18
Removes a bad workaround for an issue that was introduced in 16.0

diegoreymendez:
This is the first step to fixing: https://github.com/wordpress-mobile/WordPress-iOS/issues/15256

Rolls back [a workaround](https://github.com/wordpress-mobile/WordPress-iOS/pull/15136) I added for an issue with the refresh control in the list of sites.

My workaround was only hiding the refresh wheel, but it wasn't a proper fix for the problem, as the first row in the list of sites couldn't be selected by users.

## Additional information

After rolling back these changes, we still need to fix the issue, which was introduced in this PR:

https://github.com/wordpress-mobile/WordPress-iOS/pull/15081

And specifically by this commit:

https://github.com/wordpress-mobile/WordPress-iOS/pull/15081/commits/3a9543c74ef4c3b340ae53aacd236416c54ab7cb

## To test:

1. Run the App.
2. Switch to the "My Sites" tab.
3. If the refreshing activity wheel isn't visible on top of the first row, just tap on a site to see it's details, and go back to the list of sites.

Make sure you can see the refresh activity wheel on top of the first site (in the list of sites).  As mentioned above, we need another PR to fix the underlying issue.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
@@ -526,16 +526,6 @@ static NSInteger HideSearchMinSites = 3;
     UIRefreshControl *refreshControl = [[UIRefreshControl alloc] init];
     [refreshControl addTarget:self action:@selector(syncBlogs) forControlEvents:UIControlEventValueChanged];
     self.tableView.refreshControl = refreshControl;
-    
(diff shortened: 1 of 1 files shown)


ScoutHarris (approved):
Thanks @diegoreymendez !

:shipit:

diegoreymendez:
Thanks for the review!

PR #15345, opened 2020-11-19
Update stats in the iOS 14 today widget from the app

### LINKED ISSUE #15104: Today Widget - Adapt existing data for Today Widget to the new architecture

Gio2018:
Fixes #15104

Using the existing Stats logic to update the new Today Widget when the stats in the app are updated.

Known issues:

- at the moment, the logic takes as "default" site whatever site has been set for the old widgets (using the "Widgets" button in the Stats scene). If none was set, this will be the default site for the current user
- at the moment the logout and re-login with a different user is not handled, so it should be tested with only one user

To test:
- run the app and login on the testing device with a user that has at least one site, and possibly some daily stats that are not all zero
- background the app, Install the widget (any size would do, but medium size visualizes more info) and verify that the info look correct
- trigger some stats updates (e.g. do some visits, likes, comments from another device or browser, with another user that is following the site you are testing)
- go back to the app, then choose the site that is shown in the widget and then go to Stats, making sure that the updates you triggered appear
- background the app and go to the widget, and make sure that the same updates are reported in the widget.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Stores/StatsInsightsStore.swift
+++ b/WordPress/Classes/Stores/StatsInsightsStore.swift
@@ -991,28 +991,28 @@ private extension InsightStoreState {
 
     private func saveHomeWidgetTodayData(stats: TodayWidgetStats) {
 
-        guard let siteID = SiteStatsInformation.sharedInstance.siteID as? Int else {
+        guard let siteID = SiteStatsInformation.sharedInstance.siteID else {
             return
         }
 
         var homeWidgetTodayCache = HomeWidgetTodayData.read() ?? initializeHomeWidgetTodayData()
 
-        guard let oldData = homeWidgetTodayCache[siteID] else {
+        guard let oldData = homeWidgetTodayCache[siteID.intValue] else {
             DDLogError("HomeWidgetToday: Failed to find a matching site")
             return
         }
         // TODO - TODAYWIDGET: it might be better to move the blog updates in SiteStatsInformation
         let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
 
-        guard let blog = blogService.blog(byBlogId: NSNumber(value: siteID)) else {
+        guard let blog = blogService.blog(byBlogId: siteID) else {
             DDLogError("HomeWidgetToday: the site does not exist anymore")
             // if for any reason that site does not exist anymore, remove it from the cache.
-            homeWidgetTodayCache.removeValue(forKey: siteID)
+            homeWidgetTodayCache.removeValue(forKey: siteID.intValue)
             HomeWidgetTodayData.write(data: homeWidgetTodayCache)
             return
         }
         // refresh stats and update any blog info, if they had changed
-        homeWidgetTodayCache[siteID] = HomeWidgetTodayData(siteID: siteID,
+        homeWidgetTodayCache[siteID.intValue] = HomeWidgetTodayData(siteID: siteID.intValue,
                                                            siteName: blog.title ?? oldData.siteName,
                                                            url: blog.url ?? oldData.url,
                                                            timeZoneName: blogService.timeZone(for: blog).identifier,
@@ -1038,7 +1038,7 @@ private extension InsightStoreState {
                 let title = ($1.title ?? url).isEmpty ? url : $1.title ?? url
                 let timeZoneName = blogService.timeZone(for: blog).identifier
 
-                $0[Int(truncating: blogID)] = HomeWidgetTodayData(siteID: Int(truncating: blogID),
+                $0[blogID.intValue] = HomeWidgetTodayData(siteID: blogID.intValue,
                                                                   siteName: title,
                                                                   url: url,
                                                                   timeZoneName: timeZoneName,
--- a/WordPress/Classes/Stores/StatsInsightsStore.swift
+++ b/WordPress/Classes/Stores/StatsInsightsStore.swift
@@ -1,6 +1,7 @@
 import Foundation
 import WordPressKit
 import WordPressFlux
+import WidgetKit
 
 enum InsightAction: Action {
 
@@ -990,8 +991,7 @@ private extension InsightStoreState {
 
     private func saveHomeWidgetTodayData(stats: TodayWidgetStats) {
 
-        guard let siteID = SiteStatsInformation.sharedInstance.siteID as? Int,
-              let timeZoneName = SiteStatsInformation.sharedInstance.siteTimeZone?.identifier else {
+        guard let siteID = SiteStatsInformation.sharedInstance.siteID as? Int else {
             return
         }
 
@@ -1002,19 +1002,30 @@ private extension InsightStoreState {
             return
         }
 
-        homeWidgetTodayCache[oldData.siteID] = HomeWidgetTodayData(siteID: oldData.siteID,
-                                                                   siteName: oldData.siteName,
-                                                                   url: oldData.url,
-                                                                   timeZoneName: timeZoneName,
-                                                                   date: Date(),
-                                                                   stats: stats)
+        let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
+
+        guard let blog = blogService.blog(byBlogId: NSNumber(value: siteID)) else {
+            DDLogError("HomeWidgetToday: the site does not exist anymore")
+            // if for any reason that site does not exist anymore, remove it from the cache.
+            homeWidgetTodayCache.removeValue(forKey: siteID)
+            HomeWidgetTodayData.write(data: homeWidgetTodayCache)
+            return
+        }
+        // refresh stats and update any blog info, if they had changed
+        homeWidgetTodayCache[siteID] = HomeWidgetTodayData(siteID: siteID,
+                                                           siteName: blog.title ?? oldData.siteName,
+                                                           url: blog.url ?? oldData.url,
+                                                           timeZoneName: blogService.timeZone(for: blog).identifier,
+                                                           date: Date(),
+                                                           stats: stats)
 
         HomeWidgetTodayData.write(data: homeWidgetTodayCache)
+        if #available(iOS 14.0, *) {
+            WidgetCenter.shared.reloadTimelines(ofKind: "WordPressHomeWidgetToday")
+        }
     }
 
-    // TODO - TODAYWIDGET: this isn't the right place for this initialization to happen. Putting it here for now
-    // as it's suitable to test cache updates from the stats in the app and have to figure out exactly how to handle the
-    // initialization overall
+    // TODO - TODAYWIDGET: this method will likely be exposed in the future to handle the creation of the file at login
     private func initializeHomeWidgetTodayData() -> [Int: HomeWidgetTodayData] {
 
         let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
@@ -1032,7 +1043,7 @@ private extension InsightStoreState {
                                                                   url: url,
                                                                   timeZoneName: timeZoneName,
                                                                   date: Date(),
-                                                                  stats: nil)
+                                                                  stats: TodayWidgetStats())
             }
         }
     }
--- a/WordPress/Classes/Stores/StatsInsightsStore.swift
+++ b/WordPress/Classes/Stores/StatsInsightsStore.swift
@@ -1001,7 +1001,7 @@ private extension InsightStoreState {
             DDLogError("HomeWidgetToday: Failed to find a matching site")
             return
         }
-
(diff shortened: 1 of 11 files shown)


WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -981,3 +985,66 @@ private extension InsightStoreState {
...
+        // refresh stats and update any blog info, if they had changed
+        homeWidgetTodayCache[siteID.intValue] = HomeWidgetTodayData(siteID: siteID.intValue,
+                                                           siteName: blog.title ?? oldData.siteName,
+                                                           url: blog.url ?? oldData.url,
+                                                           timeZoneName: blogService.timeZone(for: blog).identifier,
+                                                           date: Date(),
+                                                           stats: stats)
+
+        HomeWidgetTodayData.write(data: homeWidgetTodayCache)
+        if #available(iOS 14.0, *) {
```
frosty: Can we just guard this whole method for iOS 14? Is there any need to run the rest of it if we're not on iOS 14?

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -981,3 +985,66 @@ private extension InsightStoreState {
...
+            WidgetCenter.shared.reloadTimelines(ofKind: "WordPressHomeWidgetToday")
```
frosty: Should we have a constant for `"WordPressHomeWidgetToday"`?

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -981,3 +985,66 @@ private extension InsightStoreState {
     }
 
 }
+
+// MARK: - iOS 14 Widgets Data
+private extension InsightStoreState {
+
+    private func saveHomeWidgetTodayData(stats: TodayWidgetStats) {
```
frosty: Did you consider having this as another function on `TodayWidgetStats` alongside `saveData`?  It feels like this is putting extra responsibilities on the Store, and I'm not sure if it would be better placed elsewhere?

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -981,3 +985,66 @@ private extension InsightStoreState {
...
+        }
+    }
+
+    // TODO - TODAYWIDGET: this method will likely be exposed in the future to handle the creation of the file at login
+    private func initializeHomeWidgetTodayData() -> [Int: HomeWidgetTodayData] {
+
+        let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
+
+        return blogService.blogsForAllAccounts().reduce(into: [Int: HomeWidgetTodayData]()) {
+            if let blogID = $1.dotComID,
```
frosty: As this isn't just a one-liner, it might be slightly clearer to name the variables here, instead of using $0 and $1

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -981,3 +985,66 @@ private extension InsightStoreState {
...
+               let url = $1.url,
+               let blog = blogService.blog(byBlogId: blogID) {
+
+                let title = ($1.title ?? url).isEmpty ? url : $1.title ?? url
```
frosty: Can this just be `$1.title ?? url`? 🤔

frosty (commented):
Works great for me! Just left a few code comments :)

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -947,14 +948,17 @@ extension StatsInsightsStore {
...
-        }
+
 
         let data = TodayWidgetStats(views: todaysStats?.viewsCount,
                                     visitors: todaysStats?.visitorsCount,
                                     likes: todaysStats?.likesCount,
                                     comments: todaysStats?.commentsCount)
+        saveHomeWidgetTodayData(stats: data)
+
+        guard widgetUsingCurrentSite() else {
```
diegoreymendez, line 957: I'm curious about why this guard is needed.  Its purpose is not entirely clear to me.

The reason I'm asking about it is because:

1. We could have any number of widgets showing different sites.
2. Once I push the widget configuration code, the user could swap to a different site and I guess it could be useful to have the stats ready right away.

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: +1

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: +1

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: +1

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: Or alternatively this could be in a constants file shared between the two targets (since this value is also used in the widget code).

Not sure if that'd be good practice or bad practice to share a file like that across projects, though.  It's just a quick thought I had.

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -981,3 +985,66 @@ private extension InsightStoreState {
...
+// MARK: - iOS 14 Widgets Data
+private extension InsightStoreState {
+
+    private func saveHomeWidgetTodayData(stats: TodayWidgetStats) {
+
+        guard let siteID = SiteStatsInformation.sharedInstance.siteID else {
+            return
+        }
+
+        var homeWidgetTodayCache = HomeWidgetTodayData.read() ?? initializeHomeWidgetTodayData()
```
diegoreymendez, line 998: In reference to `HomeWidgetTodayData.read()`, `HomeWidgetTodayData.write()` and `HomeWidgetTodayData.delete()`

I noticed that, at each call of these three methods, we're instantiating `HomeWidgetCache<HomeWidgetTodayData>`.

Any thoughts on making `HomeWidgetTodayData` instantiable (instead of purely static) and initializable with a specific  `HomeWidgetCache<HomeWidgetTodayData>` (since it's not going to change anyway)?

For more context on why I'm suggesting this: I'm not a big fan of static methods like these ones because then you kinda have a dependency with a specific class that's harder to change or break in the future.  It also tends to make it slightly harder to test the code, since it can't be replaced through a dependency injection easily.  Just my 2 cents.

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: So... I don't necessarily feel like `TodayWidgetStats` is a better place to handle the saving because my (very personal) understanding of the responsibilities of a Store, is that it should take care of abstracting the rest of the code from where data is stored.

But let's try to get some value out of this thought... this made me think that maybe what we can improve here is to clearly specify who owns this particular responsibility in our class descriptions, whether it's the store, or `TodayWidgetStats`, and put all the relevant code together.

WordPress/Classes/ViewRelated/Stats/Today Widgets/iOS 14 Widgets/HomeWidgetTodayData.swift
```diff
@@ -0,0 +1,79 @@
...
+    let stats: WidgetStats
+}
+
+
+// MARK: - Local cache
+extension HomeWidgetTodayData {
+
+    static func read(from cache: HomeWidgetCache<Self>? = nil) -> [Int: HomeWidgetTodayData]? {
+
+        let cache = cache ?? HomeWidgetCache<HomeWidgetTodayData>(fileName: Constants.fileName,
```
diegoreymendez, line 38: This is the instantiation that happens at each call, that I referenced in a previous comment.  Just pointing it out so that my previous comment is a bit clearer.

Having to instantiate this each time made me think that maybe `HomeWidgetTodayData` can be instantiable too, and hold one instance of this object during its lifetime.

diegoreymendez (commented):
Works great! :)

Added some comments, maybe a bit nitpicky.  Please let me know if there's anything you disagree with and we can chat - the idea is not to hold back the PR.

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: I'm not a fan of that either, but since we already have that (`Constants.h, Constants.m`) I'm just gonna put it in there

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: Thanks for being so thorough on this. A few considerations:
- there's a TODO in that method, so it might change
- it's probably a bit confusing because of the coexistence of old and new Today Widgets; `TodayWidgetStats` would not be the right place for methods related to the new widgets, `HomeWidgetTodayData` would.
- This method takes the updated info from the store and then updates the Widget accordingly. It's the equivalent of the existing `storeTodayWidgetData` and, for that reason, I realize that the name should change to `storeHomeWidgetTodayData`, so that I'm gonna do for now.

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: I personally like the compactness of implicit arguments and find them quite readable, but since it seems that I am the minority 😄 I'm gonna change them to explicit

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: If I did that, we would not account the case of a title that is not nil, but empty. What we want, instead, is that if the site has an empty title, we display the url as a title

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: So this is a bit confusing, since I added the method that stores the data for the new widget, inside the old one. 
The `guard` you are looking at is for the old widget. The reason for not separating them was to use the same instance of `TodayWidgetStats` but I realize that this causes confusion, so I'm going to refactor this.

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: I don't like static methods either, and I'm open to refactor this, maybe in a future PR, as we are still working on the overall new widget architecture.
A few considerations, for more context:
- `HomeWidgetTodayData` is indeed instantiable, because it's the `TimelineEntry` used by the widgets. Those static methods could move elsewhere though
- `HomeWidgetCache` is basically a wrapper for `Filemanager.default` (a singleton), that only does the job of telling `Filemanager` what is the file to handle and how to encode the specific `HomeWidgetTodayData`. Since the file is shared between the widget and the app, we could try to inject an instance of an object on both `StatsInsightStore` and the widget, I'd just do it in a future PR if you'd agree.
- Still not ideal, but to offer a bit more testability, I made the cache object injectable in the static method.

Gio2018:
Thanks for reviewing @diegoreymendez @frosty !!!

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: While I now can see why it's done this way, I have to admit it wasn't immediately obvious by reading it.

Any thoughts on rewriting this to make it a bit more readable and matinainable, even if it means slightly less compact?

As an example:

```swift
let title = (element.title != nil && !element.title.isEmpty) ? element.title : url
```

diegoreymendez (approved):
I've added [a comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/15345#discussion_r527915985).

Otherwise I'm happy with the replies, and since this is working and aiming at a feature branch, it's :shipit: from me.

WordPress/Classes/Stores/StatsInsightsStore.swift
Gio2018: This is more readable, but does not completely work, as it returns a `String?` instead of a `String`

Gio2018: I added a comment to this line to describe what the code does, hoping that it helps to better understand what it does

Gio2018:
@frosty , merging this so we can push other PRs, but feel free to add more reviews; we are merging into the feature branch anyway

PR #15358, opened 2020-11-20
Localization: translation key exceeds character limit for GlotPress

### LINKED ISSUE #15353: Fix Localization issue with long key
TL;DR: One of the L10n keys we use is too long and causes issues with GlotPress. We need to split it in 2

### Context

I'm currently implementing an automated linting of our `.strings` file to detect any inconsistency between the translation of some of our keys across different locales. [Ref: paaHJt-D6-p2]

So far my tool has found existing inconsistencies:

 - Some are because of translators not using the right `%d/%@/…` placeholders in their translations; for those I will fix this directly in GlotPress.
 - But one of the issue is with one of our localization key whose length exceeds the 255 limit that GlotPress seems to have on key names. That's the focus of this GH issue

### The Bug

The key in question is used in `WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift`

* The full (long!) key is: `"If you want a site but don't want any of the posts and pages you have now, our support team can delete your posts, pages, media, and comments for you.\n\nThis will keep your site and URL active, but give you a fresh start on your content creation. Just contact us to have your current content cleared out."`
* But in practice it gets truncated during the localisation process, right after `Just cont` (256 char limit).

### Expected behavior

The text for this copy should correctly appear translated in the app

### Actual behavior

Because the key name gets truncated during the round trip to GlotPress, when iOS tries to find a translation from the full long key at runtime, it won't find a matching translation in the `Localizable.strings` if your language is not `en` (the only `.strings` file that has the key untruncated). So instead it will fallback to using the key, thus showing the English copy to the user instead of the translated copy.

* Compare line 2822 of `en.lproj/Localizable.strings` with the same from any other `*.lproj/Localizable.strings`
* You could also test it at runtime by setting your phone to a locale other than english and navigate to the Start Over screen to see that text appear in English despite the phone being in another language.

##### Suggested Fix

To work around the 255 character limit, given that this copy seems to basically be 2 separate paragraphs after all, I suggest we split that text in 2 separate keys, one for each paragraph.

_Once the 2 new keys have been created and used in the code – and the old key is removed from the code – one should ideally go to GlotPress to do the same split of the 2 paragraphs in all the translations for that text, so that we keep using the same translation for those paragraphs than before without having our community to re-translate the newly-introduced key_

mindgraffiti:
Fixes #15353 

### To test
It can't really be tested until the new translation strings are available. So the only thing to look for is: ensure the localized strings do not exceed 255 characters in English.

1. Change device language to a language that uses words that have more characters than English, such as German or Dutch. 
2. Navigate to My Site > Settings > Start Over. The detailed instructions text should now to be translated instead of falling back to English.

### Screenshots
Proof of bug:
<a href="https://user-images.githubusercontent.com/1062444/99831606-f38f3d80-2b24-11eb-8d14-6d202045ee59.png"></a>

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
@@ -15,6 +15,9 @@ open class StartOverViewController: UITableViewController, MFMailComposeViewCont
 
     @objc let headerView: TableViewHeaderDetailView = {
         let header = NSLocalizedString("Let Us Help", comment: "Heading for instructions on Start Over settings page")
+        /// GlotPress breaks if iOS keys are longer than 256 characters, and additionally if a string reaches
+        /// 256 characters in English, it does not leave any flexibility for languages with longer words.
+        /// So lets support our translators and keep GlotPress happy :) 
         let detail1 = NSLocalizedString("If you want a site but do not want any of the posts and pages you have now, " +
                                         "our support team can delete your posts, pages, media, and comments for you.",
                                         comment: "Detailed instructions on Start Over settings page. This is the first paragraph.")
--- a/WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
@@ -15,9 +15,10 @@ open class StartOverViewController: UITableViewController, MFMailComposeViewCont
 
     @objc let headerView: TableViewHeaderDetailView = {
         let header = NSLocalizedString("Let Us Help", comment: "Heading for instructions on Start Over settings page")
-        /// GlotPress breaks if iOS keys are longer than 256 characters, and additionally if a string reaches
-        /// 256 characters in English, it does not leave any flexibility for languages with longer words.
-        /// So lets support our translators and keep GlotPress happy :) 
+        /// GlotPress breaks if iOS keys are longer than 256 characters.
+        /// Additionally, if a string reaches 256 characters in English,
+        /// it does not leave any flexibility for languages with longer words.
+        /// So lets support our translators and keep GlotPress happy :) GH: #15353
         let detail1 = NSLocalizedString("If you want a site but do not want any of the posts and pages you have now, " +
                                         "our support team can delete your posts, pages, media, and comments for you.",
                                         comment: "Detailed instructions on Start Over settings page. This is the first paragraph.")
--- a/WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
@@ -15,7 +15,7 @@ open class StartOverViewController: UITableViewController, MFMailComposeViewCont
 
     @objc let headerView: TableViewHeaderDetailView = {
         let header = NSLocalizedString("Let Us Help", comment: "Heading for instructions on Start Over settings page")
-        let detail1 = NSLocalizedString("If you want a site but don't want any of the posts and pages you have now, " +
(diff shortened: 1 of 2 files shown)


AliSoftware:
@mindgraffiti Thanks or the change; did you also made the change in GlotPress? 🤔 

💭 If we let GlotPress in its current state on the next iteration our process will create 2 new keys (for the ones you created for paragraphs 1 and 2) that won't be translated in other languages at first, despite us already having the translation of those 2 paragraphs in the old key that we'll soon delete. This means that our contributors will see 2 new keys not being translated yet and will translate them, potentially with slight differences compared to the current translation we use. Not sure if that's a big deal, but I was thinking reporting those translations ourselves in GlotPress manually once the 2 new keys are created might avoid this gotcha?

mindgraffiti:
@AliSoftware nope. After the localized strings are created in the class, Platform 9 usually takes over and "magic happens" 😀 . If you'll explain how I should run the script that gathers all the localized strings, I can make an attempt.

AliSoftware:
yeah actually while I was writing the comment above that let me thinking, I'll need to investigate a bit deeper what's the best course of action from here:

 - merge the PR without changing GlotPress, and wait for the next code freeze to let the magic happen (that's when all the localizations are merged into a file that GlotPress will pick up / pull regularly); at which point we'd have the new keys in GlotPress and we could finally do the copy/pasting of those translations for all locales manually before any contributor steps in to provide a slightly different one
 - or we trigger the magic sooner than the code freeze, forcing it to happen on demand, to be able to fix it right away. But I'm realizing that it could have unintended consequences, like uploading strings that are in develop while we're not yet sure they'll stay or not change until the next code freeze, so contributors might see new strings to translate before they're supposed to, etc;

It's Friday night already so I'll wait for next week to get some good rest and thing this through, get a grasp at the potential consequences and see what's best… but I'm starting to think option 1 (don't touch GlotPress until next code freeze) might be the best option after all.

WordPress/Classes/ViewRelated/Blog/Site Management/StartOverViewController.swift
```diff
@@ -15,7 +15,15 @@ open class StartOverViewController: UITableViewController, MFMailComposeViewCont
 
     @objc let headerView: TableViewHeaderDetailView = {
         let header = NSLocalizedString("Let Us Help", comment: "Heading for instructions on Start Over settings page")
-        let detail = NSLocalizedString("If you want a site but don't want any of the posts and pages you have now, our support team can delete your posts, pages, media, and comments for you.\n\nThis will keep your site and URL active, but give you a fresh start on your content creation. Just contact us to have your current content cleared out.", comment: "Detail for instructions on Start Over settings page")
+        let detail1 = NSLocalizedString("If you want a site but do not want any of the posts and pages you have now, " +
```
ScoutHarris, line 20: I'm totally lurking here, but I have one request. Can you add a comment explaining why this is split? So the next person who comes along doesn't say "oh this can be one string" and jam it all together again.

mindgraffiti: good point! I'll do that.

AliSoftware: Took the liberty to update the comment, as the 255-chars restriction is only on the key, not the translations.

Since it happens that on iOS we use the English copy as the key, then yes we will need to split that key which ends up splitting the English copy. But the translations for that key can be longer.

AliSoftware (approved):
I'll merge this without waiting for a change in GlotPress after all.

Since on next Monday we will start a new Sprint and I'll do a code freeze, which is what initiates an update of the `Localizable.strings` from scanning `NSLocalizedString(…)` calls in our code, then let GlotPress import that updated `.strings` file, the new key will start being available in GP at that point anyway.

Issue #15192, opened 2020-10-29
Jetpack Section: Activity Log


emilylaguna:
Project: https://github.com/wordpress-mobile/WordPress-iOS/issues/15193

<img width="1779" alt="Screen Shot 2020-10-29 at 4 46 03 PM" src="https://user-images.githubusercontent.com/793774/97630636-1c705700-19ed-11eb-95c6-10ada417ea8c.png">

## Designs
Coming Soon.

#### Network
- [x] Update endpoints to support the new query parameters (https://github.com/wordpress-mobile/WordPressKit-iOS/pull/301)
- [x] Update ActivityStore to support paging (https://github.com/wordpress-mobile/WordPress-iOS/pull/15274)
- [x] Update Activity model to add any missing fields
- [x] Add endpoint to get activity filter types (https://github.com/wordpress-mobile/WordPressKit-iOS/pull/306)
- [x] Map remote filter type model to a local model
- [x] Add a new model to represent a filter group (Activity Type)

#### Date Picker Filter View
- [x] Filter by date (https://github.com/wordpress-mobile/WordPress-iOS/pull/15412)
- [x] Check rotating issue (https://github.com/wordpress-mobile/WordPress-iOS/pull/15537)

#### Activity Type Filter View
- [x] Filter by activity type (https://github.com/wordpress-mobile/WordPress-iOS/pull/15440)

#### Activity Log
- [x] Add designs of the filter bar (https://github.com/wordpress-mobile/WordPress-iOS/pull/15468)

#### Tracking
- [x] Add tracking (https://github.com/wordpress-mobile/WordPress-iOS/pull/15544)

#### Finishing Touches
- [x] Check Accessibility (https://github.com/wordpress-mobile/WordPress-iOS/pull/15534)
- [x] Design Review (RTL, small devices, iPad) (https://github.com/wordpress-mobile/WordPress-iOS/pull/15468)

PR #15359, opened 2020-11-20
Activity Log: Add Year Calendar Component

leandroalonso:
Part of #15192

This PR adds a Year Calendar component to select one range of dates.

| Light Mode  | Dark Mode |
| Simulator Screen Shot - iPhone 11 - 2020-11-20 at 16 12 02 | Simulator Screen Shot - iPhone 11 - 2020-11-20 at 16 11 56 |

There are still missing a feature:

* You can't create a range by selecting a data _before_ the first one you selected. I'm still working on this.

## Scheduling a post calendar

This uses the same calendar component as Scheduling, so please make sure no regressions were introduce in this one.

1. Create a new post
2. Tap Publish.
3. Tap Publish Date
4. Check that the calendar behaves like before

## Fullscreen calendar

1. Go to Activity Log
2. The calendar should appear right away (I will remove that before merging this PR)
3. Test in light/dark mode, portrait, and landscape
4. When you tap Done, check the dates printed in the debug console

Please, also testing using an iPad.


WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+        updateLabels()
+    }
+
+    private func updateLabels() {
+        guard let startDate = startDate else {
+            resetLabels()
+            return
+        }
+
+        let formatter = DateFormatter()
```
emilylaguna: Date formatters can be a bit expensive to create, maybe move this to a property and reuse it?

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+        formatter.dateFormat = "MMM d, yyyy"
```
emilylaguna: What about using `setLocalizedDateFormatFromTemplate` ?

```suggestion
        formatter.setLocalizedDateFormatFromTemplate("MMM d, yyyy")
```

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -0,0 +1,196 @@
+import UIKit
+
+protocol CalendarViewControllerDelegate: class {
+    func didCancel(calendar: CalendarViewController)
+    func didSelect(calendar: CalendarViewController, startDate: Date, endDate: Date)
+}
+
+class CalendarViewController: UINavigationController {
```
emilylaguna: The naming of this is a bit odd to me since it's a navigation controller. I can't think of a better name for it though. Thoughts?

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+    init(delegate: CalendarViewControllerDelegate?) {
+        let yearCalendarViewController = YearCalendarViewController()
+        yearCalendarViewController.delegate = delegate
+        super.init(rootViewController: yearCalendarViewController)
+    }
+
+    init() {
```
emilylaguna: Should you also add a way to init with a start / end date to restore the users selection?

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+                return
+            }
+
+            // Manually center the current date vertically (centeredVertically is not working)
+            self.calendarCollectionView.setContentOffset(CGPoint(x: 0, y: self.calendarCollectionView.contentOffset.y - self.calendarCollectionView.frame.height / 2), animated: false)
+        }
+    }
+
+    private func resetLabels() {
+        startDateLabel.text = "Start Date"
```
emilylaguna: Should these be localized?

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -0,0 +1,196 @@
...
+        separatorDateLabel.textColor = .textSubtle
+
+        endDateLabel.text = "End Date"
+        endDateLabel.textColor = .textSubtle
+    }
+
+    @objc private func done() {
+        guard let startDate = startDate, let endDate = endDate,
+              let calendar = navigationController as? CalendarViewController else {
+            return
```
emilylaguna: If the user opens the calendar and decides to clear their date selection then clicks done nothing will happen. Should we instead send nil dates to the delegate and let it decide what to do?

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
```diff
@@ -75,34 +108,64 @@ extension CalendarDataSource: JTACMonthViewDelegate {
...
 
     func calendar(_ calendar: JTACMonthView, didDeselectDate date: Date, cell: JTACDayCell?, cellState: CellState, indexPath: IndexPath) {
         configure(cell: cell, with: cellState)
     }
 
-    private func configure(cell: JTACDayCell?, with state: CellState) {
-        let cell = cell as? DateCell
-        cell?.configure(with: state)
+    func calendar(_ calendar: JTACMonthView, shouldSelectDate date: Date, cell: JTACDayCell?, cellState: CellState, indexPath: IndexPath) -> Bool {
+        if style == .year, calendar.selectedDates.count > 1 && cellState.selectionType != .programatic || firstDate != nil && !calendar.selectedDates.isEmpty && date < calendar.selectedDates[0] {
```
emilylaguna: `calendar.selectedDates` is a dynamic getter that returns a sorted version of selected dates. Maybe create a variable at the top to reduce the number of calls to that method?

```
let selectedDates = calendar.selectedDates
```

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
```diff
@@ -75,34 +108,64 @@ extension CalendarDataSource: JTACMonthViewDelegate {
...
+            firstDate = nil
+            let retval = !calendar.selectedDates.contains(date)
+            didDeselectAllDates?()
+            calendar.deselectAllDates()
```
emilylaguna: Minor idea, since the method is `did` deselect shouldn't this be called after the deselection happens?
```suggestion
            calendar.deselectAllDates()
            didDeselectAllDates?()
```

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
```diff
@@ -75,34 +108,64 @@ extension CalendarDataSource: JTACMonthViewDelegate {
...
 
-        cell.dateLabel.textColor = textColor
+    func calendarSizeForMonths(_ calendar: JTACMonthView?) -> MonthSize? {
+        return style == .year ? MonthSize(defaultSize: 50) : nil
+    }
 
-        cell.dateLabel.backgroundColor = cellState.isSelected ? WPStyleGuide.wordPressBlue() : .clear
+    func calendar(_ calendar: JTACMonthView, headerViewForDateRange range: (start: Date, end: Date), at indexPath: IndexPath) -> JTACMonthReusableView {
+        let date = range.start
+        let formatter = DateFormatter()
```
emilylaguna, line 189: Same comment about extracting this to a property to reduce cost as above.

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
```diff
@@ -142,26 +213,93 @@ class DateCell: JTACDayCell {
...
+            rightPlaceholder.centerYAnchor.constraint(equalTo: centerYAnchor)
+        ])
+
+        bringSubviewToFront(dateLabel)
     }
 }
 
 extension DateCell {
-    func configure(with state: CellState) {
+    func configure(with state: CellState, hideInOutDates: Bool = false) {
```
emilylaguna: I'm not sure what `hideInOutDates` means. Could you add some documentation here?
```suggestion
    func configure(with state: CellState, hideInOutDates: Bool = false) {
```

emilylaguna (changes requested):
👋 This is looking good so far. I did hit a few issues one of which we discussed in DM but I'll note them here as well:

- Performance is really poor, if I try to select now + 6 months it takes about 5 seconds in simulator. If I select a date a few years in the future, it hangs indefinitely. 

- on iPad when the calendar first comes into view it's set to Jan 1951 and then animate scrolls to the current date. Is there a way to disable that animation?

- I get a crash when running iOS 11 and iOS 12, I don't know about iOS 13 since my simulator isn't working.

```
WordPress/CalendarViewController.swift: 8: 7: Fatal error: Use of unimplemented initializer 'init(nibName:bundle:)' for class 'WordPress.CalendarViewController'
```

Other than that, dark mode and iPad are looking good, and I didn't see any regressions on the post settings view.

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: I removed that.

It's the responsibility of the presenting VC to encapsulate that into a navigation controller.

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: Done!

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: Done

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
leandroalonso: All those delegate methods were removed (given the very poor performance).

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
leandroalonso: Same as above.

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
leandroalonso: Same as above.

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
leandroalonso: Added!

leandroalonso:
@emilylaguna addressed the performance issues and all the comments, except this:

> on iPad when the calendar first comes into view it's set to Jan 1951 and then animate scrolls to the current date. Is there a way to disable that animation?

I'm not sure this is possible because the collection view is loaded and then it scrolls to the correct position. Instead of performing that without animation, I changed to always animate (iPad and iPhone). I'd rather prefer to display the correct part right away but I wasn't able to do it. We can reassess that in the near future, but I don't think this introduces an issue for the user.

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
```diff
@@ -75,34 +136,66 @@ extension CalendarDataSource: JTACMonthViewDelegate {
     }
 
     func calendar(_ calendar: JTACMonthView, didSelectDate date: Date, cell: JTACDayCell?, cellState: CellState, indexPath: IndexPath) {
+        if style == .year {
```
leandroalonso, line 139: Open to suggestions on how to improve this beauty.

jkmassel:
👋 Hey there! We are freezing `16.3` today, so this PR is being bumped to `16.4`. If you need it to be part of the `16.3` release, please merge it to the `release/16.3` branch and send me a DM – I'll be happy to cut a new beta release!

WordPress/Classes/ViewRelated/Post/Scheduling/CalendarCollectionView.swift
emilylaguna: oof, I spent some time on this but didn't come up with anything better, I also looked at some other projects and they're all doing something similarly. 🤷‍♀️

emilylaguna:
> I'm not sure this is possible because the collection view is loaded and then it scrolls to the correct position. Instead of performing that without animation, I changed to always animate (iPad and iPhone). I'd rather prefer to display the correct part right away but I wasn't able to do it. We can reassess that in the near future, but I don't think this introduces an issue for the user.

I gave this a look too but didn't have any success either. We can loop back and look at it again if it's causing issues.

emilylaguna (approved):
Working well for me! 🚢

PR #15303, opened 2020-11-16
[Home Page Picker] Enable Home Page Picker feature

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2727

## To test:
This enables the Home Page Picker for the enhanced site design selection. The old flow can still be tested for comparison reasons by toggling the feature off.
<details>
<summary><strong>To disable or enable the development version of Home Page Picker</strong></summary>

- Open the app from the build that allows FeatureFlags such as a PR build or a local development build
    - By default, the modal layout picker will be disabled.
    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.
        - Toggle "Site Creation: Home Page Picker" to enable or disable the picker
        - <kdb><a href="https://user-images.githubusercontent.com/3384451/97343011-5f032980-185d-11eb-9030-06c17f732fec.gif"></a></kdb>

</details>

## To Test
From a clean install:
1. Log in 
1. Navigate to the Create WordPress.com site
    - My Sites > ➕ > Create WordPress.com site
1. **Expect** To be presented with the Choose a design screen instead of the site Segment selection 

### Create a Site - Selecting a design
<details>
<summary>Steps</summary>

1. Navigate to the Create WordPress.com site
    - My Sites > ➕ > Create WordPress.com site
1.  **Expect** 
    - To see the analytics event `enhanced_site_creation_site_design_viewed`
    - To see a single placeholder while the network request to fetch the available design is processing
        - 📓 You may not see this if you're on a fast network
    - To see placeholders populate for each site design 
    - To see thumbnail images load as their network request finishes
1. Scroll to the bottom of the list
1.  **Expect** 
    - To see the header collapse as you scroll down
    - To see placeholders for thumbnails that weren't on the screen during the initial load.
        - 📓 You may not see the thumbnail placeholders if you're on a fast network and if they were cached on the server
1. Scroll to the top of the list
1.  **Expect** 
    - To see the header expand as you reach the top
    - Thumbnails should not be fetched from the server again unless your device purged the cache (low memory devices)
1. Select a layout
1.  **Expect** 
    - To see the footer slide in from the bottom
    - To see a border and checkmark animate in place on the selected design. 
1. Select Choose
1.  **Expect** 
    - To see the analytics event `enhanced_site_creation_site_design_selected` with the property `template` populated with the selected layout's slug
1. Search for a domain
[...]

PR #15360, opened 2020-11-20
[Home Page Picker] Bug fixes

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/WordPress-iOS/pull/15303#issuecomment-731341928

This takes care of a few bug fixes I found while doing a comprehensive test. I'll spell out each one here.

### Swipe to dismiss should be disabled
On a device running iOS 13 or newer:

1. Try to swipe away the screen on Choose a Design, Choose a Domain, and Site Confirmation
     - **Expect** to not be able to swipe away the screens
1. Try to swipe away the screen on Choose a Design > Preview 
     - **Expect** to be able to swipe away the screens

### Header jumping on Choose a Domain

<details>
<summary>Original issues</summary>

**Noticed on:** 
- iPad Air 11.4

**Steps**
1. Navigate to Choose a Domain
1. Search for results

Notice the header will sometimes expand
<kdb><a href="https://user-images.githubusercontent.com/3384451/99837456-9bad0280-2b35-11eb-9a9f-7dd8bf56af12.png"></a></kdb>

</details>

**Steps**
1. Navigate to Choose a Domain 
1. Scroll to collapse the header
1. Search for results
1. Select clear
1. Search for results
     - **Expect** the header to maintain state

----

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/CollapsableHeaderViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/CollapsableHeaderViewController.swift
@@ -577,12 +577,9 @@ extension CollapsableHeaderViewController: UIScrollViewDelegate {
     }
 
     /// Restores the stashed content offset if it appears as if it's been reset.
-    private func restorContentOffsetIfNeeded(_ scrollView: UIScrollView) {
+    private func restoreContentOffsetIfNeeded(_ scrollView: UIScrollView) {
         guard let stashedOffset = stashedOffset else { return }
-        if scrollView.contentOffset.y == 0 { // Offset has probably been reset
-            scrollView.contentOffset = stashedOffset
-        }
-        self.stashedOffset = nil
+        scrollView.contentOffset = stashedOffset
     }
 
     private func resizeHeaderIfNeeded(_ scrollView: UIScrollView) {
@@ -602,7 +599,10 @@ extension CollapsableHeaderViewController: UIScrollViewDelegate {
     }
 
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
-        restorContentOffsetIfNeeded(scrollView)
+        guard stashedOffset == nil else {
+            restoreContentOffsetIfNeeded(scrollView)
+            return
+        }
 
         guard !shouldUseCompactLayout,
               !isShowingNoResults else {
--- a/WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/CollapsableHeaderViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/CollapsableHeaderViewController.swift
@@ -571,15 +571,24 @@ extension CollapsableHeaderViewController: UIScrollViewDelegate {
         }
     }
 
-    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
-        /// Clear the stashed offset because the user has initiated a change
-        stashedOffset = nil
-    }
-
     /// Restores the stashed content offset if it appears as if it's been reset.
     private func restoreContentOffsetIfNeeded(_ scrollView: UIScrollView) {
-        guard let stashedOffset = stashedOffset else { return }
+        guard var stashedOffset = stashedOffset else { return }
+        stashedOffset = resolveContentOffsetCollisions(scrollView, cachedOffset: stashedOffset)
         scrollView.contentOffset = stashedOffset
+    }
+
+    private func resolveContentOffsetCollisions(_ scrollView: UIScrollView, cachedOffset: CGPoint) -> CGPoint {
+        var adjustedOffset = cachedOffset
+
+        /// If the content size has changed enough to where the cached offset would scroll beyond the allowable bounds then we reset to the minum scroll height to
+        /// maintain the header's size.
+        if scrollView.contentSize.height - cachedOffset.y < scrollView.frame.height {
+            adjustedOffset.y = maxHeaderHeight - headerHeightConstraint.constant
+            stashedOffset = adjustedOffset
+        }
+
+        return adjustedOffset
     }
 
     private func resizeHeaderIfNeeded(_ scrollView: UIScrollView) {
@@ -593,9 +602,14 @@ extension CollapsableHeaderViewController: UIScrollViewDelegate {
         }
     }
 
-    fileprivate func updateTitleViewVisibility(_ animated: Bool = true) {
+    private func updateTitleViewVisibility(_ animated: Bool = true) {
         let shouldHide = (headerHeightConstraint.constant > midHeaderHeight) && !shouldUseCompactLayout
         titleView.animatableSetIsHidden(shouldHide, animated: animated)
+    }
+
+    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
+        /// Clear the stashed offset because the user has initiated a change
+        stashedOffset = nil
     }
 
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
--- a/WordPress/Classes/ViewRelated/Site Creation/WebAddress/AddressCell.xib
+++ b/WordPress/Classes/ViewRelated/Site Creation/WebAddress/AddressCell.xib
@@ -32,7 +32,6 @@
             <viewLayoutGuide key="safeArea" id="njF-e1-oar"/>
             <constraints>
                 <constraint firstItem="hQm-TV-IDW" firstAttribute="leading" secondItem="njF-e1-oar" secondAttribute="leading" constant="16" id="QB9-fd-Zsn"/>
-                <constraint firstItem="hQm-TV-IDW" firstAttribute="centerY" secondItem="njF-e1-oar" secondAttribute="centerY" id="eud-Yr-hp5"/>
                 <constraint firstItem="njF-e1-oar" firstAttribute="trailing" secondItem="hQm-TV-IDW" secondAttribute="trailing" constant="55" id="pJe-WK-HF0"/>
             </constraints>
             <connections>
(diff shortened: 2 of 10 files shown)


WordPress/Classes/ViewRelated/Site Creation/Wizard/SiteCreationWizardLauncher.swift
```diff
@@ -44,13 +44,15 @@ final class SiteCreationWizardLauncher {
         if FeatureFlag.siteCreationHomePagePicker.enabled {
             if #available(iOS 13.0, *) {
                 wizardContent.modalPresentationStyle = .pageSheet
+                wizardContent.isModalInPresentation = true
```
chipsnyder, line 47: This prevents the ability to swipe the modal away.

WordPress/Classes/ViewRelated/Site Creation/WebAddress/AddressCell.xib
```diff
@@ -32,7 +32,6 @@
             <viewLayoutGuide key="safeArea" id="njF-e1-oar"/>
             <constraints>
                 <constraint firstItem="hQm-TV-IDW" firstAttribute="leading" secondItem="njF-e1-oar" secondAttribute="leading" constant="16" id="QB9-fd-Zsn"/>
-                <constraint firstItem="hQm-TV-IDW" firstAttribute="centerY" secondItem="njF-e1-oar" secondAttribute="centerY" id="eud-Yr-hp5"/>
```
chipsnyder, line 35: Saw some conflicting constraint errors in the console

WordPress/Classes/ViewRelated/Gutenberg/Collapsable Header/CollapsableHeaderViewController.swift
```diff
@@ -602,7 +599,10 @@ extension CollapsableHeaderViewController: UIScrollViewDelegate {
     }
 
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
-        restorContentOffsetIfNeeded(scrollView)
+        guard stashedOffset == nil else {
+            restoreContentOffsetIfNeeded(scrollView)
+            return
+        }
```
chipsnyder, line 619: These changes fix the jumping header. stashedOffset will be reset to nil as soon as a user starts to scroll. (This is in an earlier PR in `scrollViewWillBeginDragging`)

mkevins (commented):
Hi Chip, I tested the PR by first reproducing the issues, and then testing the steps with the changes here on both iPhone 11 (physical device) and iPhone Air (simulator). Everything seems to be working correctly now, but I observed a different issue on the domains screen. It seems that sometimes the results can be partially hidden:



I also noticed that sometimes there are no results for a query that other times has results.. not sure if there is a race condition there :thinking: ?

And one minor nit-pick: the modal seems to move a bit (vertically) from one step to the next. Is this expected?

chipsnyder:
>but I observed a different issue on the domains screen. It seems that sometimes the results can be partially hidden:

Good testing! The only way I was able to recreate this is by:
1. Start with blank results 
1. Enter a search that will return results
1. Swipe up so that the results populate while the view is scrolling
1. Notice error

Let me know if you're seeing it with a different set of steps. I have an idea on how to fix this though so I'll investigate that today. 

>I also noticed that sometimes there are no results for a query that other times has results.. not sure if there is a race condition there 🤔 ?

I noticed some inconsistent behavior here as well but I also started seeing that before we made changes to this screen. It's also different between Android and iOS on the number of results and that Android includes a notice if there is an exact match. Outside of the UI that surrounds the search results, I left the actual search behavior unchanged. In my testing, the inconsistencies were matching what's currently live so I didn't want to hold up the feature while investigating that behavior. I'll create a GH issue today so that we can track that work and mark out the differences. I agree though there is probably a race condition going on with the throttle.

> And one minor nit-pick: the modal seems to move a bit (vertically) from one step to the next. Is this expected?

Yeah, I'm not a big fan of this either but this is the system adjusting the modal to make room for the keyboard. Since the Search box is auto-selected it shifts up while animating in. In your video, it looks like the simulator is using the hardware keyboard setting so the soft keyboard doesn't pop, just the toolbar, so the shift is smaller. But it's noticeable about the cause when using a soft keyboard:

<kdb><a href="https://user-images.githubusercontent.com/3384451/99984423-08fda500-2d7b-11eb-8ad5-3081877b5142.gif"></a></kdb>

chipsnyder:
@mkevins I believe I solved the issue RE: the partially hidden results now: https://github.com/wordpress-mobile/WordPress-iOS/pull/15360/commits/566ce5ca3440cea125f93e8e2c336e4cda17e613

I also included the Gutenberg version bump for @geriux's Group block background color [support](https://github.com/wordpress-mobile/gutenberg-mobile/pull/2703)

antonis (approved):
Tested on iPad Air and iPhone 11 (simulators) and everything seems to work as expected :tada:
Nice work Chip 👍

chipsnyder:
Just FYI to cut back on the number of builds to the enable PR I included the wording changes mentioned in https://github.com/wordpress-mobile/WordPress-iOS/pull/15303#issuecomment-731849867 as part of this PR https://github.com/wordpress-mobile/WordPress-iOS/pull/15360/commits/86bf7cff3dc6bcf8169167241ab1c351ce24e131

chipsnyder:
Updated this branch with the changes from https://github.com/wordpress-mobile/WordPress-iOS/pull/15365 and https://github.com/wordpress-mobile/WordPress-iOS/pull/15376

chipsnyder:
I also included a fix here (https://github.com/wordpress-mobile/WordPress-iOS/pull/15360/commits/3d2a785e956c6ab1cb3d58508985a37fa1ed466e) where the results could update after you made a selection but update and not include that selection. 

Steps
1. Type a bit of text slowly to start getting some results
1. add more text 
1. Make a selection before the results update. 
Expect for your selection to be cleared if the new results don't contain your original selection

chipsnyder:
> I also noticed that sometimes there are no results for a query that other times has results.. not sure if there is a race condition there 🤔 ?

Investigating this a bit more, I think there might be some inconsistencies on the backend as well. Possibly related to the size of the requested results? 

If I make the request: https://public-api.wordpress.com/rest/v1.1/domains/suggestions?locale=en&query=test&segment_id=1
I get: 
```
{
    "error": "empty_results",
    "message": "No available domains for that search."
}
```

However, if I add `quantity` like: https://public-api.wordpress.com/rest/v1.1/domains/suggestions?locale=en&query=test&segment_id=1&quantity=20 

I'll get about 19 results. I'm interested in the results you're seeing @mkevins. if you hook up a proxy does the app reflect the results from the last search that was performed?

Here's the issue to track this https://github.com/wordpress-mobile/WordPress-iOS/issues/15378

mkevins:
> Good testing! The only way I was able to recreate this is by:
> 
> 1. Start with blank results
> 2. Enter a search that will return results
> 3. Swipe up so that the results populate while the view is scrolling
> 4. Notice error
> 
> Let me know if you're seeing it with a different set of steps. I have an idea on how to fix this though so I'll investigate that today.

These are the same steps I used too, and I didn't notice it otherwise.

> I noticed some inconsistent behavior here as well but I also started seeing that before we made changes to this screen. It's also different between Android and iOS on the number of results and that Android includes a notice if there is an exact match. Outside of the UI that surrounds the search results, I left the actual search behavior unchanged. In my testing, the inconsistencies were matching what's currently live so I didn't want to hold up the feature while investigating that behavior. I'll create a GH issue today so that we can track that work and mark out the differences. I agree though there is probably a race condition going on with the throttle.

That makes sense, and good catch on the differences between Android and iOS. I think it makes sense to move forward, and resolve the query issues separately.

> Yeah, I'm not a big fan of this either but this is the system adjusting the modal to make room for the keyboard. Since the Search box is auto-selected it shifts up while animating in. In your video, it looks like the simulator is using the hardware keyboard setting so the soft keyboard doesn't pop, just the toolbar, so the shift is smaller. But it's noticeable about the cause when using a soft keyboard:

Ah, thanks for the explanation! In that case, I don't think there is anything to change there :smile: .

mkevins:
> I'll get about 19 results. I'm interested in the results you're seeing

I tried a few times, and I get no results without the `quantity` parameter, same as you. With `quantity=20`, I get mixed results.. sometimes ~15, sometimes ~18, so I guess it changes each time.

> if you hook up a proxy does the app reflect the results from the last search that was performed?

Can you clarify this part? Do you mean, to running the app via a sandboxed API (or any other steps)? To be clear, the above results were on the production API, and I observed that even by merely clicking the links in your comment above.

chipsnyder:
> > if you hook up a proxy does the app reflect the results from the last search that was performed?
> Can you clarify this part? Do you mean, to running the app via a sandboxed API (or any other steps)? To be clear, the above results were on the production API, and I observed that even by merely clicking the links in your comment above.

Hey @mkevins, I was thinking more along the lines of a network monitoring tool against the prod environment. For example, I was using [Charles Proxy](https://www.charlesproxy.com) but any monitoring tool would probably work.

I'm really hoping to confirm if all of the inconsistencies are network-related or not. While I tested I saw that a particular request would return the result set "A", then on a second call it might return a new result set of "B", then on a subsequent call, I might get "A" again. 

So while running the app I might see a sequence like:
1. `query=te` returns result "A"
1. `query=newString` returns result "C"
1. `query=te` returns result "B"
1. `query=newString` returns result "C"
1. `query=te` returns result "A"

But what I noticed in the app is the display would work as expected and return the results of the last request. So in my example above I would see the results "A". I was wondering if you're seeing the same or not. 


> I think it makes sense to move forward, and resolve the query issues separately.

👍 I'll go ahead and quote the reply above and put it on the issue https://github.com/wordpress-mobile/WordPress-iOS/issues/15378 so that we continue the investigation there

PR #15365, opened 2020-11-23
[Home Page Picker] Add Template to Preview analytics

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/2720
📓 This helps align the approach taken on iOS and Android to include the `template` property in the loading preview analytics

📓 This does not include any of the changes in https://github.com/wordpress-mobile/WordPress-iOS/pull/15360 

## To test:
📓 This was branched off of the enable PR so no need to enable HPP unless you have it overridden

<details>

    - From the site page:
        - Click on your Gravatar.
        - Click on App Settings.
        - Click on Debug.

</details>

📓 To help with testing these you can filter the console logs on `Tracked:`

1.  Select a site design
1. Select Preview
    - **Expect** to see the event `enhanced_site_creation_site_design_preview_viewed` fire with the parameter `template` populated with the selected design's `slug`
1. When the URL starts loading
    -  **Expect** to see the event `enhanced_site_creation_site_design_preview_loading` fire with the parameter `template` 
1. When the URL successfully loads
    -  **Expect** to see the event `enhanced_site_creation_site_design_preview_loaded` fire with the parameter `template` 
1. Select Choose
    - **Expect** to see the event `enhanced_site_creation_site_design_selected` fire with the parameter `template` 

## Additional Context: 

This is a continuation of the work done in the [Home Page Picker Project](https://github.com/wordpress-mobile/gutenberg-mobile/labels/Project%3A%20Home%20Page%20Picker) and is part of the [Round 2 requirements](https://github.com/wordpress-mobile/gutenberg-mobile/labels/Project%3A%20Home%20Page%20Picker%20Round%202).  

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Site Creation/DesignSelection/Preview/SiteDesignPreviewViewController.swift
+++ b/WordPress/Classes/ViewRelated/Site Creation/DesignSelection/Preview/SiteDesignPreviewViewController.swift
@@ -107,7 +107,7 @@ class SiteDesignPreviewViewController: UIViewController, NoResultsViewHost {
 extension SiteDesignPreviewViewController: WKNavigationDelegate {
 
     func webView(_ webView: WKWebView, didStartProvisionalNavigation navigation: WKNavigation!) {
-        SiteCreationAnalyticsHelper.trackSiteDesignPreviewLoading()
+        SiteCreationAnalyticsHelper.trackSiteDesignPreviewLoading(siteDesign)
     }
 
     func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
@@ -119,7 +119,7 @@ extension SiteDesignPreviewViewController: WKNavigationDelegate {
     }
 
     func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
-        SiteCreationAnalyticsHelper.trackSiteDesignPreviewLoaded()
+        SiteCreationAnalyticsHelper.trackSiteDesignPreviewLoaded(siteDesign)
         progressBar.animatableSetIsHidden(true)
         removeProgressObserver()
     }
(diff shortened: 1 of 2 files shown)


chipsnyder:
@antonis I didn't include the `template` in the case of if a preview fails to load. Did you include the `template` in that scenario?

antonis:
> @antonis I didn't include the `template` in the case of if a preview fails to load. Did you include the `template` in that scenario?

I didn't think to include this either. It is easy to add it in the error message text though.

There are two cases I report this:
* When the [connection is down](https://github.com/wordpress-mobile/WordPress-Android/blob/0ca3355e85894dcf44ff436b9e4629b624d3d5cd/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt#L164)
* When an [error loading the preview comes up](https://github.com/wordpress-mobile/WordPress-Android/blob/0ca3355e85894dcf44ff436b9e4629b624d3d5cd/WordPress/src/main/java/org/wordpress/android/ui/sitecreation/theme/HomePagePickerViewModel.kt#L175)

I guess if we add this information it would be useful only in the second case. Wdyt?

chipsnyder:
>I guess if we add this information it would be useful only in the second case. Wdyt?

Yeah, I don't we really need it for this flow. We would be able to set up filters to see how much each is erroring without it anyway. So I think we can leave it off

antonis (approved):
The changes LGTM 🎉 
Nice work @chipsnyder 👍

mkevins:
Thanks for adding this!

> We would be able to set up filters to see how much each is erroring without it anyway. So I think we can leave it off

My thoughts exactly.. once we have the `preview_{viewed,loading,loaded}` events, we'll see if there is a problem with a particular template preview loading from that.

PR #15383, opened 2020-11-24
Activity Log: change it to UIViewController and change ImmuTable to accepts VCs with tableView

leandroalonso:
This PR does two changes:

1. Changes `ImmuTableViewHandler` to accepts `UIViewController`s that has a tableView and conforms to `UITableViewDelegate` and `UITableViewDataSource`
2. Changes `ActivityListViewController` to `UIViewController`

The reason for this change is that for the new Activity Log I need to add custom views, like this filter at the top:

`UITableViewController` makes this simple change extremely hard and painfully (even when using `tableViewHeader`).

This change keeps the functionality for all the existent `ImmuTableViewHandler` consumers (with zero code changes in them) while allows it to be used with plain `UIViewController`.

Another solution would be to get rid of `ImmuTable` stuff from the ActivityListViewController, but based on my analysis this would be a little bit more complex. Analyzing the trade-off, it seems that this is the best option (for now)

### To test

1. Tap in any blog on your list
2. Go to Activity Log
3. Scroll down, make sure pagination works

#### Testing other parts that uses `ImmuTableViewHandler`

1. Tap your profile picture
2. Go to Help & Support and make sure the table view appears correctly
3. Close it, open another site that uses Jetpack
4. Tap Jetpack Settings
5. Make sure the table view appears correctly (tap some options)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/ImmuTable.swift
+++ b/WordPress/Classes/Utility/ImmuTable.swift
@@ -234,7 +234,7 @@ public enum ImmuTableCell {
 ///         reference to the handler from your view controller.
 ///
 open class ImmuTableViewHandler: NSObject, UITableViewDataSource, UITableViewDelegate {
-    typealias UIViewControllerWithTableView = ImmuTableView & UITableViewDataSource & UITableViewDelegate & UIViewController
+    typealias UIViewControllerWithTableView = TableViewContainer & UITableViewDataSource & UITableViewDelegate & UIViewController
 
     @objc unowned let target: UIViewControllerWithTableView
     private weak var passthroughScrollViewDelegate: UIScrollViewDelegate?
@@ -453,8 +453,8 @@ extension UITableView: CellRegistrar {
 
 // MARK: - UITableViewController conformance
 
-@objc public protocol ImmuTableView: class {
+@objc public protocol TableViewContainer: class {
     var tableView: UITableView! { get set }
 }
 
-extension UITableViewController: ImmuTableView {}
+extension UITableViewController: TableViewContainer {}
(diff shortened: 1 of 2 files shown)


WordPress/Classes/Utility/ImmuTable.swift
```diff
@@ -448,3 +450,11 @@ extension UITableView: CellRegistrar {
         }
     }
 }
+
+// MARK: - UITableViewController conformance
+
+@objc public protocol ImmuTableView: class {
```
leandroalonso: Accepting ideas for a better protocol name 😅

WordPress/Classes/Utility/ImmuTable.swift
```diff
@@ -320,47 +322,47 @@ open class ImmuTableViewHandler: NSObject, UITableViewDataSource, UITableViewDel
...
+            return target.tableView?(tableView, canEditRowAt: indexPath) ?? false
         }
 
         return false
     }
 
     open func tableView(_ tableView: UITableView, editActionsForRowAt indexPath: IndexPath) -> [UITableViewRowAction]? {
         if target.responds(to: #selector(UITableViewDelegate.tableView(_:editActionsForRowAt:))) {
-            return target.tableView(tableView, editActionsForRowAt: indexPath)
+            return target.tableView?(tableView, editActionsForRowAt: indexPath)
```
leandroalonso, line 365: Because of this line, we need the target to conform to `UITableViewDataSource`. I'd love to remove this (since THIS class is the Data Source), but I couldn't come up with anything :(

WordPress/Classes/Utility/ImmuTable.swift
jleandroperez: Brainstorming!!!

- `ImmuTableParent` ? 
- `ImmuTableContainer` ?
- `ImmuTableContext` ?

Or... since none of the protocol details implies anything ImmuTable specific (it's just "a thing" that contains a TableView), we could go with:

- `TableViewParent`
- `TableViewContainer`
- `TableViewContext`

The name **ImmuTableView** makes me think of a concrete ImmuTableView component (and not the container per se!)

jleandroperez (approved):
✅ Code looks solid
✅ All test scenarios pass

Nice work @leandroalonso!!! :shipit: sir!

PR #15391, opened 2020-11-25
Today Widget - Fetch remote stats

### LINKED ISSUE #15347: Today Widget - Fetch stats from backend
Allow the widget itself to refresh stats from backend periodically.

Gio2018:
Fixes #15347

Notes:
- in `WordPressHomeWidgetToday` there are two time constants:
    - `refreshInterval`: this is the refresh interval of the widget itself, set to 1 hour
    - `minElapsedTimeToRefresh`: this is the minimum elapsed time to request a refresh for a cached item. If, at the moment a cached element is read, it's older than this time, it will be re-fetched, otherwise it will be kept. This takes into account that a cached element can happen in several points. For example, if the app refreshes the cache after 58 minutes from the previous widget refresh, there is no need for the widget to make another network call when the hour passes.

To test:
- Pre-requisites
    - to test this within reasonable amount of times, it's recommended to tweak  refreshInterval` and `minElapsedTimeToRefresh`. 
As an example, you could set:
      - `refreshInterval = 5` (as far as I have seen, iOS would override smaller values anyway.)
      - `minElapsedTimeToRefresh = 1`
  - Use an account with a site that has some stats and to which you can update stats
 
- Testing
  - On either simulator or device, build/run the app and install the widget (alternatively, you can build/run the widget scheme directly)
  - make sure the stats that you see in the widget are what you expect
  - update the stats of that site from another source (add comments likes etc from the web, another device/account etc)
  - wait a time that is at least the time you set for `refreshInterval` (keep in mind that iOS will decide when exactly to refresh the widget anyway), and make sure that the stats get updated in the widget

PR submission checklist:


Gio2018:
@frosty @diegoreymendez please hold on this one, I have found a bug that needs fixing first. Sorry for the confusion!

Gio2018:
@diegoreymendez @frosty , this PR is back in the game!!

frosty (approved):
Works great for me!

diegoreymendez (approved):
This works really well for me.  I haven't found anything to comment on the code, honestly... it looks rather clean to me and I really like that we're using the service.

Gio2018:
Thanks for reviewing @diegoreymendez @frosty !!

PR #15397, opened 2020-11-26
Initial changes to make the Today Home Widget customizable.

diegoreymendez:
First attempt at making our Today Home Widget customizable.

In this PR you can:

- Customize the widget to select which site to show the stats for.
- See the site stats in the widget.

This PR also adds a new target (the Intents Extension) together with provisioning profiles and App ID.

## About the number of lines changed:

The PR is quite large, with around 2000 lines added.  Most of these, though, are changes to add the new intents target and the certificates.  The code changes aren't that many.

## Known limitations:

- This PR doesn't change the stats refresh logic so the info may not be up to date at all times.
- Remote icons for sites aren't loading yet: Apple's default mechanism for downloading remote images in the intents extension is not working at all, and I'm a bit stuck trying to figure out the best solution.
- It seems we're showing hidden sites too.  We should probably filter the list of sites shown to visible ones.

## To test:

1. Long tap on the widget.
2. Tap "Edit Widget"
3. Tap "Choose" in the "Site" row.
4. Select a site from your list.

Make sure the site you selected is properly loaded in the widget.

## PR submission checklist:


WordPress/Info.plist
```diff
@@ -8,96 +8,85 @@
 	<string>${PRODUCT_NAME}</string>
 	<key>CFBundleExecutable</key>
 	<string>${EXECUTABLE_NAME}</string>
-	<key>CFBundleIcons~ipad</key>
+	<key>CFBundleIcons</key>
```
frosty: I _think_ it's just moving items in the plist around automatically, but I'm not sure these icon changes really belong in here?

WordPress/WordPress.xcodeproj/xcshareddata/xcschemes/WordPressNotificationServiceExtension.xcscheme
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
frosty: Any idea why these schemes for other targets were added?

diegoreymendez:
Update: I've pushed more changes into this PR so that the site you select is shown in the widget.

WordPress/Info.plist
```diff
@@ -253,10 +253,22 @@
...
 	<string>Enable microphone access to record sound in your videos.</string>
-	<key>NSPhotoLibraryUsageDescription</key>
-	<string>To add photos or videos to your posts.</string>
 	<key>NSPhotoLibraryAddUsageDescription</key>
 	<string>To add photos or videos to your posts.</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>To add photos or videos to your posts.</string>
+	<key>NSUserActivityTypes</key>
+	<array>
+		<string>SelectSiteIntent</string>
```
frosty: Should this be reverse-dns style like the others?

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
```diff
@@ -67,12 +68,102 @@ private extension Provider {
...
+    private var defaultSiteID: Int? {
+        // TODO - TODAYWIDGET: taking the default site id from user defaults for now.
+        // This would change if the old widget gets reconfigured to a different site than the default.
+        // This will be updated with the configuration intent.
+        UserDefaults(suiteName: WPAppGroupName)?.object(forKey: WPStatsTodayWidgetUserDefaultsSiteIdKey) as? Int
+    }
+
+    private func widgetData(for siteID: String) -> HomeWidgetTodayData? {
+        // TODO - TODAYWIDGET: we might change this, but for now an ID equal to zero should not return any valid data
+        //HomeWidgetTodayData.read()?[siteID ?? 0]
```
frosty: Can we remove this line?

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
```diff
@@ -67,12 +68,102 @@ private extension Provider {
...
+
+    func getSnapshot(for configuration: SelectSiteIntent, in context: Context, completion: @escaping (HomeWidgetTodayData) -> Void) {
+
+        /// - TODO: review this guard... it's crazy we'd have to ever show static content.  Maybe we need to show an error message?
+        ///
+        guard let site = configuration.site,
+              let siteIdentifier = site.identifier,
+              let widgetData = widgetData(for: siteIdentifier) else {
+
+            completion(Self.staticContent)
```
frosty, line 53: Yes, presumably this should show 0 values and some kind of error if there's no site / data? (this can be for a future PR!)

WordPress/WordPressIntents/IntentHandler.swift
```diff
@@ -0,0 +1,73 @@
...
+        return self
+    }
+
+    // MARK: - SelectSiteIntentHandling
+
+    func defaultSite(for intent: SelectSiteIntent) -> Site? {
+        return sitesDataProvider.defaultSite
+    }
+
+    func resolveSite(for intent: SelectSiteIntent, with completion: @escaping (SiteResolutionResult) -> Void) {
```
frosty, line 19: I think this is primarily intended when using intents with SiriKit. For example, if a user spoke the name of a site to use, you would have to try and match it to one of their sites in the app – you could then pass back either a successful match or a failure. See https://developer.apple.com/documentation/sirikit/resolving_and_handling_intents/resolving_the_parameters_of_an_intent

frosty (commented):
Functionally, this is working as expected – I was able to see my sites in the list and select one, and that selection was saved. I've left some code level comments, and I have a few other things:

- I'm not sure we need the site icon on the configuration view – particularly if a site has a longer title, we might not have space anyway 🤔
image

- The sites on the list don't appear to be in any specific order, and it seems to be randomized each time
- In the main app, we show the site's domain as a subtitle on the cell – is that possible here?
- Do we need the IntentsUI target for this? I was under the impression that was only for use with Siri or Maps, but I may not understand it correctly.

diegoreymendez:
I'm temporarily converting this back to a draft and removing the reviewers, as I fix the merge issues.  Please hold any reviews.

diegoreymendez:
The merge conflicts are resolved, but I'll keep this as draft while I address @frosty 's feedback and also check why it is that site icons are NOT loading in the site selection screen for the widget.

diegoreymendez:
Gio - this is ready for review.  I'll be addressing @frosty 's feedback in the meanwhile.

WordPress/Info.plist
diegoreymendez: I've rolled back all changes to this Info.plist since none of these changes done manually.  I believe the one to trigger all the changes was the `NSUserActivityTypes` one.

I've re-tested my PR without that change and it seems to work just fine.  Can I ask you to double check this @frosty ?

WordPress/Info.plist
diegoreymendez: Changes reverted... read my comment above.

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
diegoreymendez: I'll leave this for future evaluation, since I'm not sure it's safe NOT to call the completion closure.

Personally I'd expect the `completion` closure to always be called, one ways or another, since it doesn't specifically signal a "success" situation, but the end of an operation.

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
diegoreymendez: Done.

WordPress/WordPressIntents/IntentHandler.swift
diegoreymendez: Updates this comment to reflect I probably will need to implement that method (in a future PR).

WordPress/WordPress.xcodeproj/xcshareddata/xcschemes/WordPressNotificationServiceExtension.xcscheme
diegoreymendez: I re-ordered the schemes a bit, and I'm not sure why but that seems to have shared the schemes as well.  I un-shared the schemes and the files are now gone.

WordPress/WordPress.xcodeproj/xcshareddata/xcschemes/WordPressIntents.xcscheme
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
diegoreymendez, line 1: Folks, I'm not sure if this scheme should be shared or not.  IMHO there's no reason why schemes shouldn't be shared but if either of you knows better, can you let me know?

PS: Regardless of my doubt, I still haven't shared the others on purpose.  Not sure what happened there!

diegoreymendez:
Just for context, the rendering issue in the main Widget customization screen seems to be [a bug](https://twitter.com/simonbs/status/1309235571093536770?s=20).

~I'll try to work around it.~  That just didn't work.  I'm turning off site icons for now.

WordPress/WordPress.xcodeproj/xcshareddata/xcschemes/WordPressIntents.xcscheme
frosty: I'm not sure either, although it appears we don't currently share any schemes (although it's also not in our gitignore)... @jkmassel any ideas?

diegoreymendez:
This is ready for a full review.  Please make sure to test the Widget is updating properly, following [the steps described here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15391).

diegoreymendez:
Replying to @frosty :

> I'm not sure we need the site icon on the configuration view – particularly if a site has a longer title, we might not have space anyway 🤔

I just removed the images for now since I couldn't fix the bug you reported.  It seems to be an SDK bug.

> The sites on the list don't appear to be in any specific order, and it seems to be randomized each time

Will work on this in another PR.  We also have to remove the hidden sites.

> In the main app, we show the site's domain as a subtitle on the cell – is that possible here?

Done.

> Do we need the IntentsUI target for this? I was under the impression that was only for use with Siri or Maps, but I may not understand it correctly.

Removed.

WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -1014,6 +1014,7 @@ private extension InsightStoreState {
         // refresh stats and update any blog info, if they had changed
         homeWidgetTodayCache[siteID.intValue] = HomeWidgetTodayData(siteID: siteID.intValue,
                                                            siteName: blog.title ?? oldData.siteName,
+                                                           iconURL: blog.icon ?? oldData.iconURL,
                                                            url: blog.url ?? oldData.url,
                                                            timeZone: blogService.timeZone(for: blog),
                                                            date: Date(),
```
Gio2018: nit: the indentation here needs fixing, but I think that comes from the code I merged in earlier 😀

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
```diff
@@ -1,106 +1,129 @@
...
+            provider: SiteListProvider()
+        ) { entry in
+            TodayWidgetView(content: entry)
+        }
+        .configurationDisplayName("Today")
+        .description("Stay up to date with today's activity on your WordPress site.")
+        .supportedFamilies([.systemSmall, .systemMedium])
+    }
+
+    /*
```
Gio2018: can we remove the commented out code? Seems that we won't need it anyway

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
Gio2018: as far as I understood, `getSnapshot` is used in transient situations, like when the user is installing the widget (so it's sort of a preview): in this case I think it's good to show some local data if we have them already. If we don't, we should probably try to understand what kind of information we want to show in that place (the current `staticContent` needs to be changed anyway, or we should use a different view entirely)

Gio2018 (approved):
Looks good to me! I installed the widget, chose different sites and observed stats updating for that site either when the app received new stats or when the widget refreshed the timeline (had to tweak the refresh interval for that).

just left a few minor comments in the code, otherwise :shipit:!

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
diegoreymendez: Marking this as resolved for now as there's no further action required.

WordPress/Classes/Stores/StatsInsightsStore.swift
diegoreymendez: Fixed.

WordPress/WordPressHomeWidgetToday/WordPressHomeWidgetToday.swift
diegoreymendez: Definitely.  I didn't notice this was still there.

diegoreymendez:
Hey @frosty , @Gio2018 - Since this is targetting a feature branch and Giorgio already approved, I'm moving ahead to merge.

Please feel free to add comments on this PR and I'll address them in a separate PR.

@jkmassel - Please let me know what you think about this: https://github.com/wordpress-mobile/WordPress-iOS/pull/15397#discussion_r532963217

PR #15398, opened 2020-11-26
Fix Jetpack header and add Whitelisted IPs tracking

### LINKED ISSUE #15396: Position of the menu for a Jetpack site is lower than expected
### Expected behavior
Menu should be in the same location

### Actual behavior
image

It appears (B header height in pink) that the space is the same for the section headers. So maybe these headers come with built in top padding? If so, for Jetpack sites we could remove the 32px custom padding that we have (‘A Padding’ in green)

image

Login with a Jetpack site and notice the position of the menu

leandroalonso:
Fixes #15396

#### Jetpack header

1. Open a Jetpack site
2. Check that the Jetpack header doesn't have an additional top padding

#### Whitelisted IPs events

1. Open Jetpack Settings
2. Tap "Whitelisted IP addresses" and make sure `jetpack_whitelisted_ips_viewed` is fired
3. Add a new IP (or remove) and make sure `jetpack_whitelisted_ips_changed` is fired

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -89,6 +89,8 @@ import Foundation
     case jetpackManageConnectionViewed
     case jetpackDisconnectTapped
     case jetpackDisconnectRequested
+    case jetpackWhitelistedIpsViewed
+    case jetpackWhitelistedIpsChanged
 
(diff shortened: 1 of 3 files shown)


osullivanchris (approved):
Approving based on screenshots but will also check build later today.

osullivanchris:
Tested the build on both iPhone and iPad on simulator and all looks as expected 👍

PR #15395, opened 2020-11-25
Add xpost data model to Core Data

guarani:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/2602

This PR is part of the effort to implement xposts in the Gutenberg mobile editor.

This step adds the Core Data models that will be used in later PRs. The new data model, `SiteSuggestion` is modeled very similar to the existing `UserSuggestion` which serves @-mentions.

The changes here are:

- Add new Core Data model `WordPress 103` with new entity `SiteSuggestion`
- Add a one-to-many relationship between `Blog` and `SiteSuggestion`

## To test

There are no user-facing changes here, ideally a code review will suffice for now — in a later PR we will add UI that uses these data models to allow the user to xpost from inside the Gutenberg editor.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15402, opened 2020-11-27
Revert "Add xpost data model to Core Data"

guarani:
Reverts wordpress-mobile/WordPress-iOS#15395

I'm reverting this PR since it would have been better if I'd merged to `gutenberg/after_1.42.0` instead of `develop`. The reason is that while the Gutenberg Mobile release 1.42.0 is in progress, any changes to WPiOS that relate to Gutenberg in anyway (even indirectly) should be merged to `gutenberg/after_1.42.0` to avoid being included in the upcoming WPiOS release that includes 1.42.0.

In this case, it would not have broken the Gutenberg release (or the WPiOS release) in any way. However, I need this PR on `gutenberg/after_1.42.0` to continue making changes that depend on it as well as other changes that have been merged to `gutenberg/after_1.42.0`.


ceyhun (approved):
LGTM!

PR #15406, opened 2020-11-30
Quick start tour update: Mark the theme selection step as completed

Tug:
With the release of the home page picker https://github.com/wordpress-mobile/WordPress-iOS/pull/15303
We make users pick a theme during site creation. We thus need to mark the theme selection step as completed in the quick start tour screen.

To test:
- Create a site
- Pick a theme
- Take the tour
- Notice "Choose a theme" is marked as completed

## TODO
- [ ] Don't mark it as completed if user chose "Skip"

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -21,9 +21,12 @@ open class QuickStartTourGuide: NSObject {
 
 
         let createTour = QuickStartCreateTour()
-        let themeTour = QuickStartThemeTour()
         completed(tour: createTour, for: blog)
-        completed(tour: themeTour, for: blog)
+
+        if FeatureFlag.siteCreationHomePagePicker.enabled {
+            let themeTour = QuickStartThemeTour()
+            completed(tour: themeTour, for: blog)
+        }
     }
 
     @objc func remove(from blog: Blog) {
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
@@ -21,7 +21,9 @@ open class QuickStartTourGuide: NSObject {
 
 
         let createTour = QuickStartCreateTour()
+        let themeTour = QuickStartThemeTour()
         completed(tour: createTour, for: blog)
+        completed(tour: themeTour, for: blog)
     }
 
     @objc func remove(from blog: Blog) {


WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift
```diff
@@ -22,6 +22,11 @@ open class QuickStartTourGuide: NSObject {
 
         let createTour = QuickStartCreateTour()
         completed(tour: createTour, for: blog)
+
+        if FeatureFlag.siteCreationHomePagePicker.enabled {
+            let themeTour = QuickStartThemeTour()
+            completed(tour: themeTour, for: blog)
```
Tug, line 28: I'd like to check `blog.currentThemeId` to see if it's not the default theme here, but it doesn't seem like it is set for any selected theme. @chipsnyder do you think you could have a look here? Also I'd be interested in learning how to do it :)

chipsnyder: From what I can find we use `blog.currentThemeId` as a way to track which theme is activated and this isn't populated until we [make the request](https://github.com/wordpress-mobile/WordPress-iOS/blob/0e6ab412ce5fb24effa489ab6d80c01257d49059/WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift#L405) `/rest/v1.1/sites/{blogID}/themes/mine` on the Theme Selection page. This saves the current theme based on the theme id. In the response for that service, the default theme id would be `hever` and the stylesheet would be `pub\/hever` 

When we fetch the blog details via `/rest/v1.1/sites/{blogID}` we get `options > theme_slug` which has the current theme's stylesheet which in the case of the `Skip` option that's `"theme_slug": "pub\/hever"` 

> I'd like to check `blog.currentThemeId` to see if it's not the default theme here
So to check the theme I think we want to split the `theme_slug` and store the last path component. My concern with this approach though is what happens if the user selected the default theme on purpose instead of using skip. Would we still want to show this menu option here? The other thought I had was what if we added a new starter design that uses this theme. I think we would only want to show this option if we hit skip. 

As an alternative approach, when we [assembly the site](https://github.com/wordpress-mobile/WordPress-iOS/blob/0e6ab412ce5fb24effa489ab6d80c01257d49059/WordPress/Classes/ViewRelated/Site%20Creation/FinalAssembly/SiteAssemblyWizardContent.swift#L111) we would know if the user has skipped or not so we could complete the site creation then. I think we could add this snippet to the site creation:

```
let themeTour = QuickStartThemeTour()
blog.completeTour(themeTour.key)
```

WDYT? I can throw together a PR to test out this idea

Tug: > what happens if the user selected the default theme on purpose instead of using skip

Indeed theme `default` can be selected currently in the HPP, let's forget that approach

Yeah either we go with your approach or we just merge it as it is. How long do you think you approach would take?

BTW thank you for enlightening me here 🙇

chipsnyder: >BTW thank you for enlightening me here 🙇

No problem. 😄

I'm testing my approach now. Had to make a small tweak because of some notification observers but should have a PR open in less than an hour

chipsnyder: Here's the PR for the other approach https://github.com/wordpress-mobile/WordPress-iOS/pull/15408

jkmassel:
👋 Hey there! We are freezing `16.3` today, so this PR is being bumped to `16.4`. If you need it to be part of the `16.3` release, please merge it to the `release/16.3` branch and send me a DM – I'll be happy to cut a new beta release!

chipsnyder:
Closing this PR As we went with https://github.com/wordpress-mobile/WordPress-iOS/pull/15408

PR #15410, opened 2020-11-30
Merge 16.2 final into trunk

jkmassel:
Merges all changes from the `16.2` series into `trunk`.


.circleci/config.yml
```diff
@@ -48,52 +66,56 @@ jobs:
           command: rake dependencies
       - run:
           name: Build for Testing
-          command: cd ./Scripts && bundle exec fastlane build_for_testing
+          command: cd ./Scripts && bundle exec fastlane build_for_testing device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
```
AliSoftware, line 69: That's an interesting fix that I missed the PR for; was this applied to other iOS repos as well?

.circleci/config.yml
```diff
@@ -33,9 +44,16 @@ commands:
 
 jobs:
   Build Tests:
+    parameters:
+      device:
+        type: string
+        description: The device (e.g. "iPhone 11") to use when compiling the build.
+      ios-version:
+        description: The iOS deployment target (e.g. "13.0") used to compile the build that will be used for testing.
+        type: string
```
AliSoftware, line 53: I know this is not for this PR to fix, but just putting this here for posterity, and for some feedback when we do a similar update on other iOS repos: since it seems we already use YAML aliases & references for those (on line 17-26), which I love :+1:, we might as well consider DRY-ing those parameters with aliases & references too 😉 

i.e. just before line 20, declare this:
```yaml
device_and_os_params: &device_and_os_params
  device:
    type: string
    description: The device (e.g. "iPhone 11") to use when compiling the build.
  ios-version:
    description: The iOS deployment target (e.g. "13.0") used to compile the build that will be used for testing.
    type: string
```
And then here we can use `parameters: *device_and_os_params` (or `<<: *device_and_os_params` merge operation if we need more params in addition to those 2)

Again, not for this PR to fix, but worth noting for next time we address this.

WordPress/Info.plist
```diff
@@ -240,8 +240,6 @@
 	</array>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
-	<key>MinimumOSVersion</key>
-	<string>4.0</string>
```
AliSoftware, line 244: 🤔  not sure why that got removed (and on which PR)… probably during the migration to Xcode 12? 🤷

AliSoftware (approved):
LGTM! :shipit:

PR #15412, opened 2020-11-30
Activity Log: Date filter

leandroalonso:
Part of #15192

Adds the feature to filter activities by a date range.

**The filter bar/chips design is not part of this task.**

#### Selecting a range

1. Open Activity Log
2. Tap "Date Range" at the top
3. Select two different dates
4. Check that the results match your selected dates

#### Calendar keeps the selected dates

1. Tap the dates you selected at the top
2. Check that the Calendar opens with your previously selection
3. Tap Cancel
4. Nothing happens

#### Removing the date filter

1. Tap the "X" button at the filter bar
2. Checks that the filtering is reset

#### No results

1. Tap "Date Range" at the top
2. Select a range date in December (or any other you're sure won't return results)
3. The screen should display an error message: "No matching events found."

#### Single date

1. Clear the filter
2. Open the calendar
3. Select a single date
4. Check that results are returned

PR submission checklist:


emilylaguna (changes requested):
Awesome, looking good. I hit a few issues when testing the filtering:

If I select single date, I expected to see results only for that date, but I am seeing results from other days as well:


If I select a date range in the past (ex: April 1st 2015 to April 7th 2015) it is not scrolled into view upon closing, then reopening the calendar.


If I select Nov 28th to Nov 30th I see results from Nov 27th, Nov 28th but nothing else. 


If one of the selected dates is not from this year we should show the year for both the dates. 

Screen Shot 2020-12-01 at 2 38 02 PM


I suspect most of the issues can be resolved by including a time in the `after` `before` query fields which is what .com includes. 

`after=2015-11-07 05:00:00&before=2020-11-15 04:59:59`

leandroalonso:
@emilylaguna ready for another review. :)

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
```diff
@@ -117,6 +135,38 @@ class ActivityListViewModel: Observable {
...
+    private func isYearDifferent(between firstDate: Date?, and secondDate: Date?) -> Bool {
+        guard let firstDate = firstDate, let secondDate = secondDate else {
+            return false
+        }
+
+        let firstDateComponents = Calendar.current.dateComponents([.year], from: firstDate)
+        let secondDateComponents = Calendar.current.dateComponents([.year], from: secondDate)
+
+        return firstDateComponents.year != secondDateComponents.year
+    }
```
emilylaguna: Unfortunately this logic doesn't work if I select 2 dates from a different year.  Here's how I envisioned the logic working. 

Also a naming change suggestion. 
```suggestion
    private func shouldDisplayFullYear(with firstDate: Date?, and secondDate: Date?) -> Bool {
        guard let firstDate = firstDate, let secondDate = secondDate else {
            return false
        }

        let currentYear = Calendar.current.dateComponents([.year], from: Date()).year
        let firstYear = Calendar.current.dateComponents([.year], from: firstDate).year
        let secondYear = Calendar.current.dateComponents([.year], from: secondDate).year

        return firstYear != currentYear || secondYear != currentYear
    }
```

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
```diff
@@ -163,12 +163,12 @@ class CalendarViewController: UIViewController {
         return header
     }
 
-    private func scrollToCurrentDate() {
+    private func scrollToFirstSelectedDateOrToday() {
```
emilylaguna: Thoughts?

```suggestion
    private func scrollToVisibleDate() {
```

emilylaguna (changes requested):
Looking good! Made a couple of comments below. 

Also we discussed this but I'll mention it here as well, when rotating the device to/from landscape the view scrolls to the very first date so Dec 2000

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
leandroalonso: Thanks for that! Done!

WordPress/Classes/ViewRelated/Activity/CalendarViewController.swift
leandroalonso: Renamed!

emilylaguna (approved):
🚢

leandroalonso:
> Also we discussed this but I'll mention it here as well, when rotating the device to/from landscape the view scrolls to the very first date so Dec 2000

I've added a task to check that again in #15192

PR #15421, opened 2020-12-01
iOS 14 Widgets: Customizer - removes hidden sites from and orders the sites alphabetically.

diegoreymendez:
Changes to the widget customizer to remove all hidden sites and to order the selectable sites alphabetically.

## To test:

It could be best to remove your existing WP installation so that this is run from scratch.

1. Launch the app to ensure the list of sites is stored in the cache
2. Launch the Intents extension, and select WordPress to run it.
3. Once WPiOS is up, go back to the Springboard.
4. Add the Widget if it's not already added.
5. Long tap on the widget, and tap "Edit Widget"
6. Open the site picker.

- Make sure none of your hidden sites are listed.
- Make sure sites are listed alphabetically.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/Stores/StatsInsightsStore.swift
+++ b/WordPress/Classes/Stores/StatsInsightsStore.swift
@@ -1029,7 +1029,7 @@ private extension InsightStoreState {
 
         let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
 
-        return blogService.blogsForAllAccounts().reduce(into: [Int: HomeWidgetTodayData]()) { result, element in
+        return blogService.visibleBlogsForWPComAccounts().reduce(into: [Int: HomeWidgetTodayData]()) { result, element in
             if let blogID = element.dotComID,
                let url = element.url,
                let blog = blogService.blog(byBlogId: blogID) {
--- a/WordPress/WordPressIntents/SitesDataProvider.swift
+++ b/WordPress/WordPressIntents/SitesDataProvider.swift
@@ -39,7 +39,20 @@ class SitesDataProvider {
                 display: data.siteName,
                 subtitle: siteDomain,
                 image: nil)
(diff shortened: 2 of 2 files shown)


diegoreymendez:
Since this PR is really simple, I'm going to merge after one approval (unless there are pending comments).

frosty (commented):
Works for me! It's a bit of an edge case, but it would be good if we could update the sorting logic to sort by domain name if the displayStrings are the same (I have a bunch of test sites with the same title).

Gio2018 (commented):
I executed the test case and works as expected. However, I realized that there is a problem when the hidden site list (or, in general, the site list) changes, because we are reading that list once and for all in `StatsInsightStore`.
Since this problem predates this PR, we can merge this one and add this issue to the project, as we should find a more dynamic way to update the site list, cause it could change. What do you think @diegoreymendez ?

diegoreymendez:
> Since this problem predates this PR, we can merge this one and add this issue to the project

I agree.  This is not the only relevant issue we can have by using a file - changing a Site name, adding a Site or deleting a Site are also checkpoints that would require an update of the cache file.

Gio2018:
> > Since this problem predates this PR, we can merge this one and add this issue to the project
> 
> I agree. This is not the only relevant issue we can have by using a file - changing a Site name, adding a Site or deleting a Site are also checkpoints that would require an update of the cache file.

Perfect. I think we could still address @frosty 's comment and then merge it. I am going to approve and add a new issue to the project related to the data refresh

Gio2018 (approved):

diegoreymendez:
> Works for me! It's a bit of an edge case, but it would be good if we could update the sorting logic to sort by domain name if the displayStrings are the same (I have a bunch of test sites with the same title).

This is done.

Issue #15190, opened 2020-10-29
Jetpack Section: Scan


emilylaguna:
Project: https://github.com/wordpress-mobile/WordPress-iOS/issues/15193

Add ability to start scans, see status, and manage threats. 

<img width="1785" alt="Screen Shot 2020-10-29 at 4 48 36 PM" src="https://user-images.githubusercontent.com/793774/97630930-7a9d3a00-19ed-11eb-85e7-86bd7df7acb5.png">

#### Network
- [x] Create a new remote service for the scanning APIs: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Start a new scan: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Get the scanning status: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Get the scan results: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Get the scan history: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Fix / Ignore Threats: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Create model for Threat: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
- [x] Create model for Scan Status: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/304
 
#### My Site View
- [x] Update with approved editorial text: https://github.com/wordpress-mobile/WordPress-iOS/issues/15193
- [x] Add logic to determine if the site has access to scanning: https://github.com/wordpress-mobile/WordPress-iOS/pull/15427
- [x] Add a new menu item for scan: https://github.com/wordpress-mobile/WordPress-iOS/pull/15427
- [x] Add action to push to the new Scan view: https://github.com/wordpress-mobile/WordPress-iOS/pull/15606
 
#### Scan View
- [x] Update with approved editorial text: https://github.com/wordpress-mobile/WordPress-iOS/issues/15193
- [x] Create new view controller: https://github.com/wordpress-mobile/WordPress-iOS/pull/15606
- [x] Add Ghost cells to display while loading the data from the network: https://github.com/wordpress-mobile/WordPress-iOS/pull/15726
- [x] Pull current scan status from the network: https://github.com/wordpress-mobile/WordPress-iOS/pull/15606
 
#### Style the Scanning State view
[...]

PR #15427, opened 2020-12-03
Jetpack Section: Adds Scan item to the My Site menu

emilylaguna:
Part of #15190 
Related WordPressKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/309

This PR adds the logic to display the Scan menu item for sites that support it. This does not include an icon or the correct action when tapping on the item yet. 

#### Scan item appears
1. Launch the app
2. Tap on the My Site tab
3. Tap on a blog that has the Jetpack Scan feature enabled

**Expectation:** the 'Scan' menu item appears under the 'Activity Log'

#### Scan doesn't appear for non-Jetpack Scan sites
1. Launch the app
2. Tap on the My Site tab
3. Tap on a blog that does not support Jetpack scan (wp.com site, or a Jetpack site that doesn't have the feature enabled)

**Expectation:** the 'Scan' menu item **DOES NOT** appear

### PR submission checklist:


WordPress/Classes/Services/JetpackScanService.swift
```diff
@@ -0,0 +1,24 @@
...
+
+    @objc func getScanAvailable(for blog: Blog, success: @escaping(Bool) -> Void, failure: @escaping(Error) -> Void) {
+        guard let siteID = blog.dotComID?.intValue else {
+            return
+        }
+
+        service.getScanAvailableForSite(siteID, success: success, failure: failure)
+    }
+
+    private func defaultApi() -> WordPressComRestApi {
```
leandroalonso: We have this in so many places that I wonder if we should extend `WordPressComRestApi` and add that as a `static` func.

We don't need to change the other places where this is used but it would be good to have a centralized place. Up to you anyway!

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -832,6 +834,14 @@ - (BlogDetailsSection *)jetpackSectionViewModel
                                                      }]];
     }
 
+    if ([self.blog supports:BlogFeatureJetpackScan]) {
```
leandroalonso, line 837: Since we're merging that to `develop` we should create a Feature Flag for it, no?

leandroalonso (changes requested):
@emilylaguna it all works as described. But I think we're missing the Feature Flag for the Scan option?

WordPress/Classes/Services/JetpackScanService.swift
emilylaguna: Done in: 7282ac7b7f40c1cfa6577dd48e92bd704016b6b5

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
emilylaguna: Oops 🤦‍♀️, done in: 540bc33727df1aee015fb80e276a57529289720b

emilylaguna:
@leandroalonso Back to you!

leandroalonso (approved):

PR #15212, opened 2020-11-04
[Xcode 12] Update Core Data transformers to use NSSecureUnarchiveFromData

frosty:
Fixes #15167.

This PR makes some changes to remove the `is using a nil or insecure value transformer` warnings when building with Xcode 12.

- Updates all Transformable Core Data properties that use 'top level' classes to specify the `NSSecureUnarchiveFromData` transformer. This only exists on iOS 12+, so this will generate console warnings on iOS 11. However, if the class isn't found, I'm pretty certain that the system will just fallback to the default behaviour.
- Adds custom value transformers for the two Transformables that don't use top level classes – one is our own Coordinate class, and the other is NSError (which does conform to NSSecureCoding).

**To test**

- Note: I should probably add a unit test to verify the migration before we merge this
- Install a build that uses the 102 version of the data model
- In that build, create a post and in Post Settings assign it a location
- Also, begin uploading some media in the Media library and cause it to error - the easiest way is to disable your internet connection during the upload
- Background the app to ensure everything is saved
- Install a build from this branch, and ensure there are no Core Data errors logged on startup
- Ensure that your post and its location are still correct
- Visit the media library, tap on one of your failed media items and check the error message is shown as the title of the alert controller that appears.

PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15432, opened 2020-12-03
[Xcode 12] Fix Core Data transformer warnings

### LINKED ISSUE #15167: [Xcode 12] Fix Core Data transformer warnings
When building on Xcode 12, there are many of these warnings:

```
warning: XYZ is using a nil or insecure value transformer. Please switch to NSSecureUnarchiveFromDataTransformerName or a custom NSValueTransformer subclass of NSSecureUnarchiveFromDataTransformer
```

- [ ] AbstractPost.revisions
- [ ] Blog.capabilities
- [ ] Blog.options
- [ ] Blog.postFormats
- [ ] BlogSettings.commentsBlacklistKeys
- [ ] BlogSettings.commentsModerationKeys
- [ ] BlogSettings.jetpackLoginWhiteListedIPAddresses
- [ ] Media.error
- [ ] MenuItem.classes
- [ ] Notification.body
- [ ] Notification.header
- [ ] Notification.meta
- [ ] Notification.subject
- [ ] Post.disabledPublicizeConnections
- [ ] Post.geolocation
- [ ] Theme.tags

frosty:
Fixes #15167. This is a new version of #15212, which adds a new core data model version, and fixes a crash on iOS 12. I've also added two migration tests. The remainder of this description is the same as the original PR:

This PR makes some changes to remove the is using a nil or insecure value transformer warnings when building with Xcode 12.

 - Updates all Transformable Core Data properties that use 'top level' classes to specify the NSSecureUnarchiveFromData transformer. This only exists on iOS 12+, so this will generate console warnings on iOS 11. However, if the class isn't found, I'm pretty certain that the system will just fallback to the default behaviour.
- Adds custom value transformers for the three Transformables that don't use top level classes – `Coordinate`, `NSError` (which does conform to NSSecureCoding), and `Set`.

**To test**

- Install a build that uses the 103 version of the data model
- In that build, create a post and in Post Settings assign it a location
- For a blog, go to Site Settings > Discussion, and add some entries to Hold For Moderation
- Also, begin uploading some media in the Media library and cause it to error - the easiest way is to disable your internet connection during the upload
- Background the app to ensure everything is saved
- Install a build from this branch, and ensure there are no Core Data errors logged on startup (you will initially see some from the pre-migration version of the database)
- Ensure that your post and its location are still correct
- Ensure that your site's Hold for Moderation settings have persisted
- Visit the media library, tap on one of your failed media items and check the error message is shown as the title of the alert controller that appears.
- Test with the Swift and the obj-c stack (you can disable the feature flag for this)

PR submission checklist:

- [x] I have considered if this change warrants user-facing release notes and have added them to RELEASE-NOTES.txt if necessary.


frosty:
@shiki Just tagging you because you found a crash in the previous PR – if you have a chance to check over this then that'd be great, but no worries if not!

WordPress/Classes/Utility/ContextManager.m
```diff
@@ -49,6 +50,15 @@ + (instancetype)internalSharedInstance
     return _override ?: _instance;
 }
 
+- (void)registerCustomTransformers
```
Gio2018: Maybe instead of having two different methods with the same code we can call the same method from both Swift and Objc? One option could be to extend our `CoreDataIterativeMigrator`, maybe renaming it since its scope expands a bit?

Gio2018 (approved):
Test case worked as expected on iPhone 11 Pro Max iOS 14.2 and iPhone X iOS 12.4. Tested both with Swift and Objective C stacks, and ran unit tests on iPhone X iOS 12.4 and iPhone 12 Pro Max 14.2 (still with both stacks).
I had only one small code comment.

jleandroperez (approved):
Looks solid on my end!!

Verified on Xcode 12 / iOS 14.2 / iPhone X.

All of the entities were preserved on migration, and the console warnings are gone.

:shipit: sir!!

shiki:
It looks like you got great reviews so I don't think you need me anymore. 🙂 Thank you very much for fixing this! 🙏

WordPress/Classes/Utility/ContextManager.m
frosty: I moved this to a common place – a static method on ValueTransformer. I didn't want to put it in IterativeMigrator as it didn't feel like it was the core purpose of that class. Thanks for the suggestion!

PR #15434, opened 2020-12-04
Move persistent store to app group

diegoreymendez:
Moves the persisten store to the App Group so that it can be used by libraries to implement services.

## Testing:

- Please, test all scenarios with the feature flag `.swiftCoreData` ON and OFF.  We need to make sure nothing is broken.
- For all tests I recommend to filter Xcode’s console log by this text “org.wordpress.StoreToAppGroupMigration” 

### Test 1: Fresh install

1. Delete WP from your phone
2. Run the App

Make sure you see this message in the log:

> org.wordpress.StoreToAppGroupMigration: Persistent store does not exist at source location file:///var/mobile/Containers/Data/Application/425CE222-4755-4133-8AE4-F4BBF6662C76/Documents/WordPress.sqlite. We assume this is the first App run.

3. Launch the app again.

Make sure you see this message in the log:

> org.wordpress.StoreToAppGroupMigration: Migration already performed.

### Test 2: Existing install

Do a clean run of the App by deleting it from the device.

1. Check out branch `release/16.0`.  This includes a version of the data model several versions old.
2. Run the app, and login.
3. Kill the App from the device.
4. Check out branch `try/move-persistent-store-to-app-group`.
5. Run the app.

Make sure you see this message in the log.

> org.wordpress.StoreToAppGroupMigration: Persistent store moved successfully to file:///private/var/mobile/Containers/Shared/AppGroup/93E1CD1A-0140-4B75-AC6B-EF138B346F89/WordPress.sqlite

6. Launch the app again.

Make sure you see this message in the log:

> org.wordpress.StoreToAppGroupMigration: Migration already performed.

## PR submission checklist:


diegoreymendez:
@Gio2018 - Before merging this we should probably implement our Widgets service to see if these changes can really be of help, and to avoid having to roll this back in the future.

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
```diff
@@ -0,0 +1,112 @@
+import Dispatch
+import Foundation
```
Gio2018: nit: both these imports can probably be omitted.

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
```diff
@@ -0,0 +1,112 @@
...
+
+/// This is a class that manages the location of the CoreData files.  This is for use in production code.
+/// While testing it's discouraged to use this class.
+///
+@objc
+class CoreDataFileLocationManager: NSObject {
+    private static let baseFileName = "WordPress"
+    private static let sqliteFileName = "\(baseFileName).sqlite"
```
Gio2018, line 8: nit: is there a reason why we need `baseFileName` or can we collapse into one constant?

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
```diff
@@ -0,0 +1,112 @@
...
+
+// MARK: - Production CoreDataFileLocationManager
+
+extension CoreDataFileLocationManager {
+
+    @objc
+    static func production() -> CoreDataFileLocationManager {
+        // This syntax is a bit weird.  What it does is it allows us to execute the code in
+        // `executeProductionStoreMigrations` once.
+        _ = executeProductionStoreMigrations
```
Gio2018, line 34: Can you explain the syntax and why this is different form directly calling `moveStoreFromMainBundleToAppGroup`, given that the method checks for `StoreToAppGroupMigrationCompleteKey` before the execution?

WordPress/Classes/Utility/CoreDataManager.swift
```diff
@@ -1,10 +1,19 @@
 /// Handles the core data stack for the whole app
 class CoreDataManager: CoreDataStack {
 
-    static let shared = CoreDataManager()
+    // MARK: - Singleton
+
+    /// The shared instance.
+    ///
+    static var shared: CoreDataManager = CoreDataManager(fileLocationManager: CoreDataFileLocationManager.production())
```
Gio2018: any reason why this changed from `let` to `var`?

WordPress/Classes/Utility/CoreDataStoreMover.swift
```diff
@@ -0,0 +1,64 @@
...
+        case couldNotLoadModel
+        case couldNotLoadSourceStore(url: URL, error: Error)
+        case sourceStoreLoadedButNotAvailable(url: URL)
+        case couldNotBackupDatabase(error: Error)
+        case couldNotMigrateStore(error: Error)
+    }
+
+    /// Moves the Store from `sourceLocation` to `targetLocation`.
+    ///
+    /// - Returns: on success this method returns a boolean holding `true` if the store was moved, or `false` if it wasn't
```
Gio2018: It seems that on success, an url is returned, not a bool. Maybe the comment needs to be updated?

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -4,11 +4,60 @@ import CoreData
...
     }
 
+    @objc
+    static func migrateStoreToLatestModel(storeURL: URL, modelURL: URL) {
+        guard FileManager.default.fileExists(atPath: storeURL.path) else {
+            DDLogInfo("No store exists at URL \(storeURL).  Skipping migration.")
+            return
+        }
+
+        guard let metadata = try? NSPersistentStoreCoordinator.metadataForPersistentStore(ofType: NSSQLiteStoreType, at: storeURL, options: nil) else {
```
Gio2018: not sure if we want to catch the error here too?

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
```diff
@@ -4,11 +4,60 @@ import CoreData
...
 class CoreDataIterativeMigrator: NSObject {
+    static let versionInfoPlist = "VersionInfo.plist"
+    static let versionHashesKey = "NSManagedObjectModel_VersionHashes"
 
     private static func error(with code: IterativeMigratorErrorCodes, description: String) -> NSError {
         return NSError(domain: "IterativeMigrator", code: code.rawValue, userInfo: [NSLocalizedDescriptionKey: description])
     }
 
+    @objc
+    static func migrateStoreToLatestModel(storeURL: URL, modelURL: URL) {
```
Gio2018: Is this method for future use? Looks like it's not called anywhere.

Gio2018 (commented):
Hey @diegoreymendez , thanks so much for working on this!! I still have to do the test scenarios; left some code comments. Other than those, I wonder if we should add some Sentry logging into this as well, what do you think?

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
diegoreymendez: Nopes.  This was actually part of an approach I was taking to resolve an issue in my PR, but I resolved it another way, so I should remove this code.  Good catch.

WordPress/Classes/Utility/Migrator/CoreDataIterativeMigrator.swift
diegoreymendez: I'll remove the entire method for now.

WordPress/Classes/Utility/CoreDataStoreMover.swift
diegoreymendez: Updated.

WordPress/Classes/Utility/CoreDataManager.swift
diegoreymendez: Just something in need of a cleanup.  Good catch.

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
diegoreymendez: Removed.

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
diegoreymendez: I've modified line 44 to make use of this (that was the original intention).  Check it out.

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
diegoreymendez: I agree this is ugly and not very intuitive.  This code I added is the equivalent of the old `dispatch_once` calls (which IMHO were easier to read).

From https://swift.org/migration-guide-swift3/:

> The free function dispatch_once is no longer available in Swift. In Swift, you can use lazily initialized globals or static properties and get the same thread-safety and called-once guarantees as dispatch_once provided. Example:
>
> `let myGlobal = { … global contains initialization in a call to a closure … }()`
> `_ = myGlobal  // using myGlobal will invoke the initialization code only `

While this is ugly, I'd still prefer to stick with this official suggestion (to replace `dispatch_once`) since it ensures thread-safety and called-once guarantees (as per the description).

diegoreymendez:
> I wonder if we should add some Sentry logging into this as well, what do you think?

Probably, although I'm not extremely familiar with what kinds of logs we send to Sentry.  Could you ping me directly when you're online to discuss this?

Thanks for the review!  I've addressed all current feedback so I'm re-requesting the review.

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
Gio2018: I see it now and makes total sense, thanks for clarifying!!

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
```diff
@@ -0,0 +1,115 @@
...
+    /// Ensures that the CoreData Store is in the App Group, or moves it there if it isn't.
+    ///
+    private static func moveStoreFromMainBundleToAppGroup() {
+        guard !CommandLine.arguments.contains("isTesting") else {
+            // We really don't want any of this code run during unit testing.
+            return
+        }
+
+        guard !UserDefaults.standard.bool(forKey: StoreToAppGroupMigrationCompleteKey) else {
+            DDLogInfo("\(StoreToAppGroupMigrationIdentifier): Migration already performed.")
```
Gio2018: nit: since this condition also occurs at the subsequent runs after a fresh install, should we change the message into something like: "Migration not required"?. In that case, there was no migration previously performed.

Gio2018 (approved):
Performed all test cases with `swiftCoreData` feature flag both ON and OFF, and all worked as expected. Left another very small comment but other than that, LGTM!

WordPress/Classes/Utility/CoreDataFileLocationManager.swift
diegoreymendez: That's a good point.

diegoreymendez:
Thanks for the review!

I'm holding off on merging this for now until we're sure it's what we need.

I'm now working to implement a service making use of the new location for the store.  Once it's up we can review it makes sense, and merge this as well.

diegoreymendez:
After chatting with @jkmassel and @Gio2018 about this... I'm closing the PR since there are concerns about the direction taken here.  There could be concurrency issues, and it's overall not clear if this is the path we want to take.

Thanks to both for your feedback.

PR #15440, opened 2020-12-04
Activity Log: filter by activity type

leandroalonso:
Part of #15192

Adds the feature to filter activities by activity type.

**The filter bar/chips design is not part of this task.**

#### Filtering by any activity type

1. Open Activity Log
2. Tap "Activity Type"
3. Select just one activity type and tap Done
4. Note that the button label changes to match the activity you selected
5. The results should reflect only activities of the type you selected

#### Reopening the Activity type screen

1. Tap the Activity type button again
2. The screen should show the activity types right away, without reloading
3. The activity type you selected should be selected
4. Tap "Done"
5. Nothing should happen since you selected the same filter as before

#### Reseting activity type

1. Click the "X" near the activity type button
2. All activities types should be displayed

#### Activity type with a date range filter

1. Select a date range that you know will have results, tap Done
2. Tap the Activity Type button
3. Activity types that match your selected date range should appear
4. Select one and tap done
5. Check that the results are from your selected activity type and between your selected range

#### Multiple activity types

1. Open the Activity type screen again
2. Select more than one activity type and tap Done
3. Check that the button now displays "Activity Type (X)" where X is the number of activity types you selected

#### Test more!

Please, test other types of combination of date range + activity types. :)

PR submission checklist:


WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
```diff
@@ -156,6 +211,16 @@ class ActivityListViewModel: Observable {
         return formattedDateRanges.joined(separator: " - ")
     }
 
+    func activityTypeDescription() -> String? {
+        if selectedGroups.isEmpty {
+            return NSLocalizedString("Activity Type", comment: "Label for the Activity Type filter button")
+        } else if selectedGroups.count > 1 {
+            return String.localizedStringWithFormat(NSLocalizedString("Activity Type (%d)", comment: "Label for the Activity Type filter button when there are more than 1 activity type selected"), selectedGroups.count)
```
emilylaguna: To aid in translating we should always be using the explicit positional indexes for localized string formats. 
 
```suggestion
            return String.localizedStringWithFormat(NSLocalizedString("Activity Type ( %1$d)", comment: "Label for the Activity Type filter button when there are more than 1 activity type selected"), selectedGroups.count)
```

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
```diff
@@ -168,6 +233,16 @@ class ActivityListViewModel: Observable {
         return firstYear != currentYear || secondYear != currentYear
     }
 
+    private func applyingNewFilter(after: Date? = nil, before: Date? = nil, group: [ActivityGroup]) -> Bool {
```
emilylaguna: Thoughts on prefixing with `is` to make it a bit clearer?

```suggestion
    private func isApplyingNewFilter(after: Date? = nil, before: Date? = nil, group: [ActivityGroup]) -> Bool {
```

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
```diff
@@ -168,6 +233,16 @@ class ActivityListViewModel: Observable {
...
+        let isSameGroup = group.count == self.selectedGroups.count && self.selectedGroups.elementsEqual(group, by: { $0.key == $1.key })
+
+        return applyingDateFilter(after: after, before: before) || !isSameGroup
+    }
+
+    private func applyingDateFilter(after: Date? = nil, before: Date? = nil) -> Bool {
```
emilylaguna:
```suggestion
    private func isApplyingDateFilter(after: Date? = nil, before: Date? = nil) -> Bool {
```

WordPress/WordPressTest/Activity/ActivityListViewModelTests.swift
```diff
@@ -34,7 +34,7 @@ class ActivityListViewModelTests: XCTestCase {
         XCTAssertEqual(activityStoreMock.offset, 3)
     }
 
-    // Check if `loadMore` dispatchs the correct after and before date
+    // Check if `loadMore` dispatchs the correct after/before date and groups
     //
     func testloadMoreAfterBeforeDate() {
```
emilylaguna:
```suggestion
    func testLoadMoreAfterBeforeDate() {
```

emilylaguna (changes requested):
Hi, this is working pretty well! I left a couple comments and saw 1 weird issue:

- After I triggered the no results view by disabling my internet, instead of a loading view I see the error view:

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
leandroalonso: Thanks! Fixed in 32f969d2f5b405529c5262892595bc6439ce6852

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
leandroalonso: Done in 6f4fb96128a48d3948cf96d3b5fc1cdb91a8d36a

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
leandroalonso: Done in 6f4fb96128a48d3948cf96d3b5fc1cdb91a8d36a

WordPress/WordPressTest/Activity/ActivityListViewModelTests.swift
leandroalonso: Done in 00c9ac7bcb30977213cc2b57b5807aa616c302e7

leandroalonso:
> After I triggered the no results view by disabling my internet, instead of a loading view I see the error view:

@emilylaguna, this is a simulator issue. Sometimes when you toggle your connection on/off the simulator does not follow, I tested here and it's working fine. (We had that issue all the time during the Offline Support project)

emilylaguna (approved):
Working well for me! 🚢

PR #15443, opened 2020-12-06
Merge 16.3 release notes into develop

jkmassel:
Also includes some beta fixes


jkmassel:
Admin-merging this due to it being created late. 

/cc @AliSoftware @mokagio in case of concerns.

PR #15446, opened 2020-12-07
Feature/15178 today widget unconfigured view

### LINKED ISSUE #15178: Today Widget - unconfigured view
On the current Today widget, we have an "unconfigured" view that is displayed when the widget is added but a site to use for widgets has not been selected within the app. Example:

no_site

We need the same sort of view for the Home Today widget.

Gio2018:
Fixes #15178

Adds the unconfigured view for Today Widgets. Here are two example screenshots.

 light appearance | dark appearance
-- | --
 | 

To test:
Prerequisite: if you had previously tested PRs from the iOS 14 widgets project, better to start from scratch (delete the app).

1. Install the app
2. Before logging in, install the widget and verify that the unconfigured view is displayed
3. Log in and verify that the default site stats for that account are shown in the widget
4. also verify that long press on the widget and then "Edit widget" allows you to configure the widget
5. Logout and verify that the unconfigured view appears again in the widget
6. Login with another account and verify that the default site stats for that account appear and that you can do the configuration as you did in step n.4.

**Known issues:**

when switching between accounts and trying to configure widgets, before changing the configuration you might see a site in the preview that belongs to the previously logged in account. This is an issue related to `WidgetKit` that we are trying to figure out. It seems related to the TODO at line 68 of WordPressHomeWidgetToday.swift and we'll open a separate issue for that

PR submission checklist:


frosty (commented):
Looks great to me, and works as described. The only thing I would say is that I think we should show a configured placeholder view in the widget selection screen, not the unconfigured view:

diegoreymendez (approved):

Gio2018:
> Looks great to me, and works as described. The only thing I would say is that I think we should show a configured placeholder view in the widget selection screen, not the unconfigured view:
> 

Thanks for finding this @frosty !! Here is the updated placeholder screen, as per discussion we had earlier on Slack:

<p align=center>

</p>

frosty (approved):
- Posted review too early! See review below.

frosty (commented):
Changes look good, however I'm seeing the placeholder view even though I'm logged into the app...

Also, now that we've changed it to dashes, I do wonder if numbers would make it clearer what this is showing? (@mattmiklic?) Although hopefully most users would already be logged in and see their real data anyway, so I'm not sure it's a big issue either way.

I also tested it with SwiftUI placeholders, but I'm not sure it's any better:

mattmiklic:
Yeah, after seeing this, I do think some numbers would be helpful here. I'm not sure that I'd know what I'm looking at here without having seen a working example of this before. @Gio2018 could we try this with the sample numbers I sent you yesterday?

Gio2018:
@frosty @mattmiklic just pushed the change that uses the ghost/placeholder views.

> however I'm seeing the placeholder view even though I'm logged into the app...

I noticed this yesterday too and spent some time investigating it: it looks that `getSnapshot` is not always called to generate the preview, and it's maybe somewhat cached.. what I have noticed is that after some time, `getSnapshot` gets called again when you need a preview, so if you install another widget right after logging in/out, you would see the "wrong" preview, but if you wait a little, you will see the appropriate one. I had used the same workaround that we used in `getTimeline` in `getSnapshot` to fix the issue of `configuration` not updating.

Gio2018:
Just updated the placeholder content with actual numbers. Here are the updated screenshots of the preview

small widget | medium widget
-- | --
 |

mattmiklic (approved):
Everything is looking great here now. Thanks @Gio2018!

PR #15448, opened 2020-12-07
Reader: allow horizontal scroll in tables

leandroalonso:
This PR adds a horizontal scroll to tables that are bigger than the size of the screen.

### To test

1. Open Reader
2. Open a post with a table that exceeds the screen width (you can [use this one](https://atomicalonso12345.blog/2020/12/07/3066/))
3. Make sure you can scroll horizontally

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/WebView/reader.css
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/WebView/reader.css
@@ -28,6 +28,10 @@ table {
   width: 100%
 }
 
+.wp-block-table {
+  overflow-x: auto;
+}
+
 div.feedflare { display: none; }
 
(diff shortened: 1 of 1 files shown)


emilylaguna (approved):
Nice! Works well for me!

nb:
Thanks, @leandroalonso! Are there other blocks that are experiencing the same problem?

leandroalonso:
@nb I don't think so. Overall, if a block works correctly in the mobile version of Calypso's Reader it should work correct on the WPiOS Reader too.

The table was broken because we had a custom internal CSS to pin it at `width: 100%` but without any scroll. Looking at the [internal CSS](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/Detail/WebView/reader.css) the only other component that has the same 100% width is the `iframe`, which should not be an issue.

Anyway, if you find anything else please let me know!

Issue #15343, opened 2020-11-19
Reader: Add P2 section


ScoutHarris:
In the Reader, add a `Followed P2s` section. 

Ref. pbArwn-1q7-p2

The WordPress.com Reader:

<img width="279" alt="Screen Shot 2020-11-19 at 1 38 32 PM" src="https://user-images.githubusercontent.com/1816888/99721486-8c15b700-2a6c-11eb-89a1-37a0b97b6db5.png">

---
Tasks:
- [x] Reader: add P2 menu item and stream. [PR #15442]
- [x] Update post topic type from "team" to "organization". [PR #15449]
- [x] Update to use `organization_id` instead of `is_wpforteams_site`. (Lighthouse [issue](https://github.com/Automattic/p2/issues/1542), P2 [header](https://github.com/wordpress-mobile/WordPress-iOS/pull/15442) issue and post [detail](https://github.com/wordpress-mobile/WordPress-iOS/pull/15447#issuecomment-740269725) issue it should resolve). [PR #15461]
- [x] Universal Links: Add Reader P2 route to `Routes+Reader`. [PR #15455]
- [x] Add `wpios_reader_p2_shown` tracking. [PR #15462]
- [x] Update 'Following' site filter to display cached sites. [PR #15477]
- [x] Add site filter to P2 and A8C streams. [PRs #15484, #15494]
- [x] Site filter: update view title to match the stream name. [PR  #15497]

PR #15449, opened 2020-12-08
Reader: A8C and P2: update type to "organization"

ScoutHarris:
Ref: #15343 

With v1.3 of the endpoint `read/menu`, the `type` for A8C and P2 are `organization`. Previously, A8C was `team`. This updates the app to match the new type.

There should be no visible change, so smoke testing the `Automattic` and `P2s` Reader streams should do it. 

To test:

---
No P2s:
- Log in with an account that is not subscribed to any P2s or A8C P2s.
- Verify the Reader does not have an `Automattic` or `P2s` menu item.

---
P2s:
- Log in with an account that is following/a user on a new P2.
- Verify the Reader does have a `P2s` menu item.

---
Automattic:
- Log in with an account that is following A8C P2s.
- Verify the Reader does have an `Automattic` menu item.

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/ReaderTeamTopic.swift
+++ b/WordPress/Classes/Models/ReaderTeamTopic.swift
@@ -4,7 +4,7 @@ import Foundation
     @NSManaged open var slug: String
 
     override open class var TopicType: String {
-        return "team"
+        return "organization"
     }
 
     @objc open var icon: UIImage? {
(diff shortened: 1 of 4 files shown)


aerych (approved):
Hi @ScoutHarris 👋 
Code looks shiny. :)  
I tested as described and everything worked as expected.  I also confirmed that the topic was the ReaderTeamTopic and showed the updated type:
Screen Shot 2020-12-08 at 3 09 59 PM
:shipit: !

ScoutHarris:
Thanks @aerych !

Issue #15438, opened 2020-12-04
Auto-paste alerts appear too often (and when nothing should have been pasted)


rachelmcr:
### Expected behavior

I only expect to see the "WordPress pasted from ..." alert when I expected something to be pasted in the app (e.g. a 2FA code if I'm logging in with 2FA).

### Actual behavior

I constantly get this alert when I open or close the app:

<img src="https://user-images.githubusercontent.com/8658164/101171228-61645a80-3637-11eb-8f3a-141b0a6211d1.jpeg" width="300px">

This was discussed a while back (internal ref: pbArwn-Iv-p2) and @bjtitus noted we might be able to improve this behavior:

> It looks like there’s a new number detection pattern in beta 2 that could be used along with the contents changed check: https://developer.apple.com/documentation/uikit/uipasteboard/detectionpattern/3650233-number?changes=latest_minor

This feels important to fix, since it's really unclear what content was in the clipboard and where it would have been pasted (or if something even was pasted anywhere).

##### Tested on [iPhone XS], iOS [14.2], WPiOS [16.3.0.1 TestFlight beta]

designsimply:
@frosty pinging you because we discussed this in Slack+P2 and decided it's a blocker for the 16.3 release and it needs a beta fix.

bjtitus:
This was reported by a user in 3549132-zen.

@frosty I think I have this fixed with changes in WordPressAuthenticator but I haven't taken a look at the React and Aztec pasteboard usage yet.

frosty:
Thanks Brandon, Authenticator is definitely the main issue here. Will review now!

designsimply:
@frosty @bjtitus just checking in to see if will we be able to get this in for 16.3?

frosty:
Hopefully – I've just done a re-review now 👍

bjtitus:
Should have it in today! I ran into some CI issues yesterday but they seem to be resolved now. 👍

PR #15454, opened 2020-12-08
WordPress Authenticator to pasteboard fix branch

bjtitus:
Addresses most of #15438

To test:

* Enter username and password for an account with 2FA enabled.
* On the 2FA screen leave the app - You should _not_ see any pop ups about pasteboard usage when leaving the app.
* Copy a URL from Safari.
* Come back to the app - You should _not_ see any alert for pasteboard usage.
* Leave the app and copy a 2FA code.
* Return to the app - You _should_ see an alert for pasteboard usage and the code should be pasted into the field.

PR submission checklist:


bjtitus:
Closed in favor of the 16.3 PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/15478

PR #15455, opened 2020-12-08
Reader: Add Reader P2 deep link.

ScoutHarris:
Ref: #15343 

This adds a deep link for the Reader P2 section.

To test:
- Run the app in a simulator.
- Log in with an account that is following a non-A8C P2.
- In the command line, enter `xcrun simctl openurl booted https://wordpress.com/read/p2`.
- Back in the simulator, verify the app switched to the Reader > P2 view. Rejoice.

PR submission checklist:


WordPress/Classes/ViewRelated/System/Coordinators/ReaderCoordinator.swift
```diff
@@ -24,9 +24,15 @@ class ReaderCoordinator: NSObject {
         WPTabBarController.sharedInstance().navigateToReaderSearch()
     }
 
-    func showA8CTeam() {
```
aerych, line 27: Digging the renaming.

aerych (approved):
Howdy @ScoutHarris 👋 
Looks good!  I tested as described and was taken to the p2 menu in the reader. 
I also tested with an account that followed no p2s. It opened the reader but that's all. No other issues. Not sure if that's the desired behavior but probably consistent with other paths? 
:shipit:

ScoutHarris:
Hey @aerych .

> I also tested with an account that followed no p2s. It opened the reader but that's all. No other issues. Not sure if that's the desired behavior but probably consistent with other paths?

I believe so. I did the same with the A8C path on a non-A8C account, and it just showed the reader. So I assumed that was as expected.

PR #15463, opened 2020-12-10
Tags and Categories description in Prepublishing bottom sheet 

yaelirub:
The goal of this PR is to add a small text describing the difference between categories and tags in the prepublishing bottom sheet. 

Since the font and color attributes of the description text are different I had to create an NSAttributedString. 

The way the code is written now is not great. There are duplications in the code and settings of String and attributed string. Please help me improve this PR by:
1. Offering a better way of having the title as attributed string 
2. Possibly finding a better phrase for tags 

To test:
1. Start a new post 
2. Tap publish 
3. See the prepublishing bottom sheet 
4. See tags description and categories description 

Simulator Screen Shot - iPhone 11 Pro - 2020-12-09 at 18 27 47

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
@@ -30,11 +30,43 @@ class PrepublishingViewController: UITableViewController {
     }()
 
     private let completion: (AbstractPost) -> ()
-
+    
+    private func rowTitle(rowID: PrepublishingIdentifier) ->  NSAttributedString {
+        switch rowID {
+        case .visibility:
+            return NSAttributedString(string:NSLocalizedString("Visibility", comment: "Label for Visibility"), attributes: nil)
+        case .schedule:
+            return NSAttributedString(string: Constants.publishDateLabel, attributes: nil)
+        case .tags:
+            let font = UIFont.systemFont(ofSize: 12.0)
+            let attributes: [NSAttributedString.Key: Any] = [
+                .font: font,
+                .foregroundColor: UIColor.textSubtle,
+            ]
+            
+            let firstString = NSMutableAttributedString(string: "Tags ", attributes: nil)
+            let secondString = NSAttributedString(string: "(Grow Your audience)", attributes: attributes)
+            firstString.append(secondString)
+            
+            return firstString
+        case .categories:
+            let font = UIFont.systemFont(ofSize: 12.0)
+            let attributes: [NSAttributedString.Key: Any] = [
+                .font: font,
+                .foregroundColor: UIColor.textSubtle,
+            ]
+            let firstString = NSMutableAttributedString(string: "Categories ", attributes: nil)
+            let secondString = NSAttributedString(string: "(Group your posts)", attributes: attributes)
+            firstString.append(secondString)
+            
+            return firstString
+        }
+    }
+    
     private let options: [PrepublishingOption] = [
         PrepublishingOption(id: .visibility, title: NSLocalizedString("Visibility", comment: "Label for Visibility")),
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -31,10 +31,42 @@ class PrepublishingViewController: UITableViewController {
 
     private let completion: (AbstractPost) -> ()
 
+    private func rowTitle(rowID: PrepublishingIdentifier) ->  NSAttributedString {
```
momo-ozawa, line 34: Instead of using a `NSAttributedString` to style the color/fontsize, what do you think about subclassing `WPTableViewCell` and making a custom cell?

I also wonder how the cells would look if we put the description label right below the title label.

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -31,10 +31,42 @@ class PrepublishingViewController: UITableViewController {
...
+            return NSAttributedString(string: Constants.publishDateLabel, attributes: nil)
+        case .tags:
+            let font = UIFont.systemFont(ofSize: 12.0)
+            let attributes: [NSAttributedString.Key: Any] = [
+                .font: font,
+                .foregroundColor: UIColor.textSubtle,
+            ]
+
+            let firstString = NSMutableAttributedString(string: "Tags ", attributes: nil)
+            let secondString = NSAttributedString(string: "(Grow your audience)", attributes: attributes)
```
momo-ozawa, line 48: Some other possibilities besides _**Grow your audience**_ :

**_Make your content discoverable_**
**_Allow others to discover your content_**

I like "discoverable" because it ties in to the Discover tab. Also, an excerpt from our [support page on tags](https://wordpress.com/support/posts/tags/):

> Tags also make it easier for people to find your content.

**_Increase engagement_**

Short and simple. I wonder if "engagement" is too jargon-y though.

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
emilylaguna: > Instead of using a NSAttributedString to style the color/fontsize, what do you think about subclassing WPTableViewCell and making a custom cell?
+1 

> I also wonder how the cells would look if we put the description label right below the title label.
I like that idea, I did a quick mock to see how that would look:

Simulator Screen Shot - iPhone 11 Pro - 2020-12-11 at 14 55 00

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -118,12 +150,16 @@ class PrepublishingViewController: UITableViewController {
...
+            cell.textLabel?.attributedText = rowTitle(rowID: .tags)
             configureTagCell(cell)
         case .visibility:
+            cell.textLabel?.attributedText = rowTitle(rowID: .visibility)
             configureVisibilityCell(cell)
         case .schedule:
+            cell.textLabel?.attributedText = rowTitle(rowID: .schedule)
             configureScheduleCell(cell)
         case .categories:
+            cell.textLabel?.attributedText = rowTitle(rowID: .categories)
```
emilylaguna, line 162: I think we could add a `subTitle` field to the `PrepublishingOption` struct and use a custom cell view, then we can 
just do:

```
cell.textLabel?.text = option.title
cell.subTitleLabel?.text = option.subTitle
```

emilylaguna (changes requested):
Nice, looking good. I agree with @momo-ozawa 's comments on using a custom cell, and adding the label below the title field. I included a mock up pic below.

twstokes:

PR #15467, opened 2020-12-10
Update fastlane plugins

jkmassel:
I've landed this in the release branch in order to test it when doing the final build, but we'll test it here too.

**To test:**
- Ensure installable build was successful

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -33,7 +33,7 @@ GEM
       json (>= 1.5.1)
     atomos (0.1.3)
     aws-eventstream (1.1.0)
-    aws-partitions (1.403.0)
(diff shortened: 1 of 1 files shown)


mokagio (approved):
Screenshot of successful installable build step

👍

mokagio:
This is great to see 👌 

I'm guessing you did it because of the reminder message in the Fastlane output, because there's no other change in the code using them. Or was there something that didn't work and required the update?

Asking so I know whether to do this sooner rather than later on my end

jkmassel:
@mokagio – yep that's right – I figured I might as well do a little housekeeping while I was waiting for other things 😄

PR #15475, opened 2020-12-11
WPTableViewHandler: fix an iOS 14 crash due to operations with negative rows

### LINKED ISSUE #15368: [NotificationsViewController.selectRow] NSInternalInconsistencyException: Attempted to perform update with invalid index path: <NSIndexPath: 0x283a1b9a0> {length = 2, path...
**Sentry Url**: https://sentry.io/share/issue/0fdb6fedd3d44625bf944ffb7c306193/
**User Count**: 616
**Count**: 803
**First Release**: 14.8.0.3
**First Seen**: 2020-09-04T04:34:54Z
**Last Seen**: 2020-11-23T18:22:01Z
**24 Hours**: 216
**30 Days**: 763

```
NSInternalInconsistencyException: Invalid indexPath
  File "NotificationsViewController.swift", line 910, in NotificationsViewController.selectRow
  File "NotificationsViewController.swift", line 211, in NotificationsViewController.traitCollectionDidChange
  File "<compiler-generated>", line null, in NotificationsViewController.traitCollectionDidChange
...
(70 additional frame(s) were not displayed)
```

leandroalonso:
Fixes #15368 

After iOS 14 we've been experiencing a lot of crashes with messages similar to this one:

```
Attempted to perform update with invalid index path: <NSIndexPath: 0x2804eedc0> {length = 2, path = 2 - 18446744073709551614}
```

They're happening inside `WPTableViewHandler.m` in two different methods:

* `-[WPTableViewHandler controller:didChangeObject:atIndexPath:forChangeType:newIndexPath:]`
* `-[WPTableViewHandler controllerDidChangeContent:]`

Also, it pretty much only happens in iOS 14.

After digging a little bit I found this repo, you can use it to reproduce the exact same crash: https://github.com/jianghui1/TestReloadRowOfTableView (and it only happens in iOS 14).

This PR changes `WPTableViewHandler.m` to not call update methods in the `UITableView` if the `row` is negative.

### To test

Unfortunately, I still don't know how to reproduce the crash. So please, test the places that use `WPTableViewHandler.m`:

* Comments
* Notifications
* Pages
* Posts
* Reader

Pull to refresh to update them and etc. :)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/WPTableViewHandler.m
+++ b/WordPress/Classes/Utility/WPTableViewHandler.m
@@ -636,10 +636,12 @@ static CGFloat const DefaultCellHeight = 44.0;
 - (void)controllerDidChangeContent:(NSFetchedResultsController *)controller
 {
     [self.tableView endUpdates];
-    
-    if (self.indexPathSelectedAfterUpdates) {
+
+    // Prevent crashes in iOS 14 if the row is negative
+    // See http://git.io/JIKIB
+    if (self.indexPathSelectedAfterUpdates && self.indexPathSelectedAfterUpdates.row > 0) {
         [self.tableView selectRowAtIndexPath:self.indexPathSelectedAfterUpdates animated:NO scrollPosition:UITableViewScrollPositionNone];
-    } else if (self.indexPathSelectedBeforeUpdates) {
+    } else if (self.indexPathSelectedBeforeUpdates && self.indexPathSelectedBeforeUpdates.row > 0) {
         [self.tableView selectRowAtIndexPath:self.indexPathSelectedBeforeUpdates animated:NO scrollPosition:UITableViewScrollPositionNone];
     }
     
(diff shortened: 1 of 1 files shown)


WordPress/Classes/Utility/WPTableViewHandler.m
```diff
@@ -657,6 +659,12 @@ - (void)controller:(NSFetchedResultsController *)controller
      forChangeType:(NSFetchedResultsChangeType)type
       newIndexPath:(NSIndexPath *)newIndexPath
 {
+    // Prevent crashes in iOS 14 if the row is negative
+    // See http://git.io/JIKIB
+    if (indexPath.row < 0 || newIndexPath.row < 0) {
```
leandroalonso, line 664: I'm not 100% sure about this change here.

In theory, iOS 13 only ignores the operations in the tableView. The rest of the code (like clearing cached row heights) is still executed though.

I'm just afraid that preventing the execution here will lead to weird states in the tableView.

Another option would be to create an extension for the `UITableView` and do the row checking there, and use those methods here.

aerych: > I'm just afraid that preventing the execution here will lead to weird states in the tableView

In that case I wonder if would be good to [self.tableView reloadData] before returning?

leandroalonso: @aerych I'm more worried about `[self clearCachedRowHeightsBelowIndexPath:newIndexPath];` not being called, for example.

_In theory_, a negative row shouldn't make any difference for the `UITableView`. I guess that some inconsistencies happen when the indexPath is positive and not present, in this case, a `reloadData` might help.

Anyway, a minor issue with the TableView (that can be solved by a pull to refresh) seems to have a smaller impact than the app crashing.

Wdyt?

yaelirub: >Anyway, a minor issue with the TableView (that can be solved by a pull to refresh) seems to have a smaller impact than the app crashing.

I agree, @leandroalonso

leandroalonso: I just looked at it again and I don't think that clearing (or invalidating) the cache of a "negative row" is necessary. :)

yaelirub (approved):
👍

leandroalonso:
@jleandroperez @aerych we're going ahead and merging the fix, but please let me know if you find anything. :)

PR #15477, opened 2020-12-11
Reader: update 'Following' site filter to display cached sites

ScoutHarris:
Ref: #15343

This updates the 'Following' site filter to display cached sites before re-fetching.
- When initially shown, the list will be what's in Core Data.
- After the sites are re-fetched, the list is updated with any changes.

To test:
- Go to Reader > Following.
- Select the site filter.
  - The list now populates immediately.
- Follow and/or unfollow a site.
  - Return to the filter.
  - After a few seconds (when the re-fetching is complete), the list will update with any changes.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
@@ -140,42 +140,17 @@ extension ReaderSiteTopic {
             completion(.success(siteService.allSiteTopics()))
         }, failure: { error in
             DDLogError("Could not sync sites: \(String(describing: error))")
-            completion(.failure(error ?? FollowedSitesConstants.remoteServiceError))
+            let remoteServiceError = NSError(domain: WordPressComRestApiErrorDomain, code: -1, userInfo: nil)
+            completion(.failure(error ?? remoteServiceError))
         })
     }
 
     /// Fetch sites from Core Data
     ///
     private static func fetchStoredFollowedSites(completion: @escaping (Result<[ReaderSiteTopic], Error>) -> Void) {
-        do {
-            let context = ContextManager.sharedInstance().mainContext
-            guard let sites = try context.fetch(sitesFetchRequest) as? [ReaderSiteTopic] else {
-                DDLogError(FollowedSitesConstants.fetchRequestError + FollowedSitesConstants.objectTypeError.localizedDescription)
-                return completion(.failure(FollowedSitesConstants.objectTypeError))
-            }
-            completion(.success(sites))
-        } catch {
-            DDLogError(FollowedSitesConstants.fetchRequestError + error.localizedDescription)
-            return completion(.failure(error))
-        }
-    }
-
-    /// Fetch request to extract followed sites from Core Data
-    ///
-    private static var sitesFetchRequest: NSFetchRequest<NSFetchRequestResult> {
-        let fetchRequest = NSFetchRequest<NSFetchRequestResult>(entityName: ReaderAbstractTopic.classNameWithoutNamespaces())
-        fetchRequest.predicate = NSPredicate(format: FollowedSitesConstants.predicateFormat, FollowedSitesConstants.objectType)
-        fetchRequest.sortDescriptors = [NSSortDescriptor(key: FollowedSitesConstants.sortByKey, ascending: true)]
-        return fetchRequest
-    }
-
-    private struct FollowedSitesConstants {
-        static let predicateFormat = "type = %@"
-        static let objectType = "site"
-        static let sortByKey = "title"
-        static let fetchRequestError = "There was a problem fetching followed sites."
-        static let objectTypeError = NSError(domain: "ReaderFilterProviderDomain", code: -1, userInfo: nil)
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
```diff
@@ -114,30 +115,69 @@ extension ReaderSiteTopic {
...
+            completion(.success(siteService.allSiteTopics()))
         }, failure: { error in
-            let unknownRestAPIError = NSError(domain: WordPressComRestApiErrorDomain, code: -1, userInfo: nil)
-            completion(.failure(error ?? unknownRestAPIError))
             DDLogError("Could not sync sites: \(String(describing: error))")
+            completion(.failure(error ?? FollowedSitesConstants.remoteServiceError))
         })
     }
+
+    /// Fetch sites from the Core Data
```
mindgraffiti:
```suggestion
    /// Fetch sites from Core Data
```

WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
```diff
@@ -114,30 +115,69 @@ extension ReaderSiteTopic {
...
+            guard let sites = try context.fetch(sitesFetchRequest) as? [ReaderSiteTopic] else {
+                DDLogError(FollowedSitesConstants.fetchRequestError + FollowedSitesConstants.objectTypeError.localizedDescription)
+                return completion(.failure(FollowedSitesConstants.objectTypeError))
+            }
+            completion(.success(sites))
+        } catch {
+            DDLogError(FollowedSitesConstants.fetchRequestError + error.localizedDescription)
+            return completion(.failure(error))
+        }
+    }
```
mindgraffiti: yaaas

Thank you for fixing this annoyance.

mindgraffiti (approved):
✅  code
✅  ui smoke tested

:shipit:

WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
```diff
@@ -114,30 +115,69 @@ extension ReaderSiteTopic {
...
+        do {
+            let context = ContextManager.sharedInstance().mainContext
+            guard let sites = try context.fetch(sitesFetchRequest) as? [ReaderSiteTopic] else {
+                DDLogError(FollowedSitesConstants.fetchRequestError + FollowedSitesConstants.objectTypeError.localizedDescription)
+                return completion(.failure(FollowedSitesConstants.objectTypeError))
+            }
+            completion(.success(sites))
+        } catch {
+            DDLogError(FollowedSitesConstants.fetchRequestError + error.localizedDescription)
+            return completion(.failure(error))
```
aerych: Maybe omit the return here?

WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
```diff
@@ -114,30 +115,69 @@ extension ReaderSiteTopic {
...
-            completion(.failure(error ?? unknownRestAPIError))
             DDLogError("Could not sync sites: \(String(describing: error))")
+            completion(.failure(error ?? FollowedSitesConstants.remoteServiceError))
         })
     }
+
+    /// Fetch sites from Core Data
+    ///
+    private static func fetchStoredFollowedSites(completion: @escaping (Result<[ReaderSiteTopic], Error>) -> Void) {
+        do {
```
aerych: It looks like the following is doing pretty much the same work as `siteService.allSiteTopics()` which is called by above in `fetchFollowedSites`.  Am I missing something?  It might be nice if `fetchFollowedSites` and `fetchStoredFollowedSites` could get the list of sites the same way.

aerych (approved):
Heya @ScoutHarris 👋   Just two suggestions from me.  Otherwise :shipit: :)

aerych:
Er... wait.  I missed this at first.  It looks like the list reorders itself after a moment.  Check it out: 
list reordering
Can we look into the cause of that?  Looks like the ordering might be case sensitive in one case but not in the other?

WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
ScoutHarris: Woops. Thank you.

WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
ScoutHarris: 🤦 yea.... I didn't notice what `allSiteTopics()` was doing. I'll update it. Thanks!

ScoutHarris:
Hey @aerych .

Thanks for noticing `allSiteTopics()`. I've updated `fetchStoredFollowedSites` to use that. 

>  Looks like the ordering might be case sensitive in one case but not in the other?

Since `fetchStoredFollowedSites` is now using the same fetch request, the ordering should be the same. Could you try again please?

aerych:
That did the trick!  :shipit: for realz!

PR #15484, opened 2020-12-12
Reader: add site filter to Automattic and P2s

### LINKED ISSUE #15343: Reader: Add P2 section

ScoutHarris:
Ref: #15343 

For Automattic and P2 reader streams, the 'Filter' tab bar option is now available to filter by site.

To test:

---
- Go to Reader > Following.
- Select `Filter`.
- Verify nothing has changed. That is, all followed sites and P2s appear in the `Sites` list, and filtering by `Topics` is available.

<kbd>following</kbd>

---
- Go to Reader > P2s.
- Select `Filter`.
- Verify the `Sites` list only contains non-A8C P2s. 
- Verify filtering by `Topic` is not available.

<kbd>p2</kbd>

---
- Go to Reader > Automattic.
- Select `Filter`.
- Verify the `Sites` list only contains A8C P2s. 
- Verify filtering by `Topic` is not available.

<kbd>a8c</kbd>

 ---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/ReaderTeamTopic.swift
+++ b/WordPress/Classes/Models/ReaderTeamTopic.swift
@@ -11,6 +11,10 @@ import Foundation
         return slug == ReaderTeamTopic.a8cSlug ? .readerA8CShown : .readerP2Shown
     }
 
+    var organizationType: SiteOrganizationType {
+        return slug == ReaderTeamTopic.a8cSlug ? .automattic : .p2
+    }
+
     static let a8cSlug = "a8c"
     static let p2Slug = "p2"
 }
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+16.5
+-----
+* [*] Reader 'P2s': added ability to filter by site. [#15484]
+
 16.4
 -----
 * [internal] Removed unused Reader files. Should be no functional changes. [#15414]
--- a/WordPress/Classes/Models/ReaderSiteTopic.swift
+++ b/WordPress/Classes/Models/ReaderSiteTopic.swift
@@ -30,9 +30,12 @@ import Foundation
         }
     }
 
+    var organizationType: SiteOrganizationType? {
+        SiteOrganizationType(rawValue: organizationID.intValue)
+    }
+
     var isP2Type: Bool {
-        let orgType = SiteOrganizationType(rawValue: organizationID.intValue)
-        return orgType == .p2 || orgType == .automattic
+        return organizationType == .p2 || organizationType == .automattic
     }
 
     @objc open var blogNameToDisplay: String {
--- a/WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
@@ -32,7 +32,7 @@ class FilterProvider: Observable, FilterTabBarItem {
         case .loading, .error:
             return []
         case .ready(let items):
-            return items
+            return FilterProvider.filterItems(items, siteType: siteType)
         }
     }
 
@@ -44,6 +44,7 @@ class FilterProvider: Observable, FilterTabBarItem {
     let emptyTitle: String
     let emptyActionTitle: String
     let section: ReaderManageScenePresenter.TabbedSection
+    let siteType: SiteOrganizationType?
 
     private let titleFunc: (State?) -> String
     private let provider: Provider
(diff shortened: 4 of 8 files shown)


WordPress/Classes/Models/ReaderTeamTopic.swift
```diff
@@ -11,6 +11,10 @@ import Foundation
         return slug == ReaderTeamTopic.a8cSlug ? .readerA8CShown : .readerP2Shown
     }
 
+    var organizationType: SiteOrganizationType {
```
aerych, line 14: Wfm.  However, if we wanted to have consistency with ReaderSiteTopic we could use the organization_id here also.  It's already returned by the /read/menu endpoint, we'd just need to add it to RemoteReaderTopic and then to the ReaderTeamTopic model. Not sure if it's worth the effort.  Up to you!

aerych (changes requested):
Howdy @ScoutHarris 👋 
Looking good! I had one suggestion code wise but I'm not sure if it's worth the effort.
I spotted a couple of things while testing that we should look at. 
First, the title of the bottom sheet always says "Following".  Should it match the stream being filtered? 
Simulator Screen Shot - iPhone 11 Pro - 2020-12-14 at 15 14 31

Second, if I'm one of the filterable streams (e.g. P2s) and select a site, I can then choose another filterable stream (e.g. Following) but the filter still shows the title of the site I previously chose.  Should it reset to "Filter"? 
Simulator Screen Shot - iPhone 11 Pro - 2020-12-14 at 15 15 41

Back to you!

ScoutHarris:
Hey @aerych .

>  the title of the bottom sheet always says "Following". Should it match the stream being filtered?

Should it? This is the view title, which to me is still accurate, i.e. every filter list is a list of Following. I can look into changing it if we like. Do we want to exactly match the stream name ("Following", "Automattic", "P2s", or do we want to append the specific stream name in some manner, or something similar ("Following", Following - Automattic", "Following - P2s")? cc @mattmiklic 

> the filter still shows the title of the site I previously chose. Should it reset to "Filter"?

Yes, yes it should. 🤦 I'll fix this.

WordPress/Classes/Models/ReaderTeamTopic.swift
ScoutHarris: In this particular case, I don't think it's worth it. But it probably wouldn't hurt for consistency, clarity, etc. I'll do it in a follow up PR.

ScoutHarris:
Hey @aerych .

I fixed the filter not resetting when the tabs change.

I'll follow up on these separately:
- 'Following' title (once we decide what it should be).
- Add `organization_id` to `ReaderTeamTopic`.

aerych (approved):
Looks good!  
Sounds good on the follow up PR for the org id.
As far as the title... I had a moment of confusion thinking it was referring to the Following menu item and that it didn't update for the P2s menu item. I guess my expectation was it matched the label in the reader menu.  Curious what @mattmiklic thinks.  
Anyway, good to go!  :shipit: from me.

ScoutHarris:
Thanks @aerych !

mattmiklic:
I think updating the title of the sheet to match the view being shown probably makes sense. That said, I think it'd be good to remove the tab bar from the P2 and Automattic views, as when there's only one tab it's redundant.

ScoutHarris:
Thanks @mattmiklic 

> I think updating the title of the sheet to match the view being shown probably makes sense. 

Just to clarify - to what? Match the stream name as Eric suggested?

> I think it'd be good to remove the tab bar from the P2 and Automattic views, as when there's only one tab it's redundant.

I see your point, but that might be a significant refactor. This view is very much designed to have a tab bar. 

Anyway, I'll update the view titles to match the stream name. I'll backburner removing the tab bar for now and consider exactly how to do that.

mattmiklic:
> Just to clarify - to what? Match the stream name as Eric suggested?

Yep! And understood on the issues on removing the tab bar.

PR #15486, opened 2020-12-14
Test WPAuthenticator's update to handling of display of TOS

ctarda:
Ref: wordpress-mobile/WordPressAuthenticator-iOS#543

This PR points WPAuthenticator to the branch that implements some changes to the logic that decides how to present the TOS button when logging in with a WordPress.com account.

| Before | After |
| ---- | ---- |
|  |  |

This draft PR is for testing purposes only, and it will be closed after the corresponding PR in WPAuthenticator is closed.

## How to test
* Checkout the branch, run `bundle exec pod install`.
* Check that the TOS button is still visible and works as expected

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -198,9 +198,9 @@ target 'WordPress' do
 
     pod 'Gridicons', '~> 1.1.0-beta'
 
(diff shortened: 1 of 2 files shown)


mindgraffiti:
Tested and verified accurate. Thank you!

PR #15490, opened 2020-12-14
Merge 16.3 final into develop

jkmassel:
Lands all remaining changes from the `16.3` series in `develop`.

Includes two release finalizations because of localization validation


Scripts/fastlane/Fastfile
```diff
@@ -209,7 +209,7 @@ import "./ScreenshotFastfile"
     ios_finalize_prechecks(options)
     unless ios_current_branch_is_hotfix
       ios_update_metadata(options)
-      ios_lint_localizations(input_dir: 'WordPress/Resources')
+      #ios_lint_localizations(input_dir: 'WordPress/Resources')
```
AliSoftware: ❓ 😭  why weren't the problematic localizations – that I assume led you to disable the linter – at least rejected in GlotPress to make the linter pass (and more importantly, to avoid releasing an app that will definitively crash at runtime on that problematic string/translation in that problematic locale)?

In any case I don't think we should merge this in develop – it should at least be re-enabled before we merge this PR for it to still be enabled for the next release, otherwise I'm sadly confident that we will forget to re-enable it…

AliSoftware (changes requested):

Scripts/fastlane/Fastfile
jkmassel: Yep – agreed. Fixed in e5b956c34d0d0bfdf545d746d44d20981d51c0cf

AliSoftware (approved):

PR #15491, opened 2020-12-14
Merge 16.3 final to trunk

jkmassel:
Fixes #

To test:

PR submission checklist:


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,17 @@
+16.4
+-----
```
AliSoftware, line 2: I might be too tired and confused, but isn't this entry supposed to only be added after `code_freeze`, so only on once the `release/16.4` branch is cut from `develop` and we merge it immediately after version bump and first beta build? Seems odd that it's present in the `release/16.3` branch innit?

AliSoftware (approved):
Overall looks good so approving, but find it odd that `RELEASE-NOTES.txt` already has the `16.4` new empty entry, so might be worth double-checking that this doesn't accidentally borrowed a commit from the 16.4 code freeze cut?!

RELEASE-NOTES.txt
jkmassel: In this case `16.4` exists because of a late-added change to `16.3` that was too late for release notes.

AliSoftware: Not sure how that explains the entry to be there? Late additions to release/16.3 after issues found in the betas are common and indeed are to late to update the release notes, so the `RELEASE-NOTES.txt` file should not be changed by those late additions and from those PRs that were targeting `release/16.3` during beta phase – we even recently added a Peril rule to warn about this – so that still doesn't really explain why we'd have a `16.4` entry in here 🤔 

But hey if you're confident that this is OK and it has a reasobale explanation then 👍 👌

PR #15494, opened 2020-12-14
Add organizationID to ReaderTeamTopic

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/15343, https://github.com/wordpress-mobile/WordPress-iOS/pull/15484#discussion_r542790416
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/316

This adds `organizationID` to `ReaderTeamTopic`. That is now used to determine the `organizationType` (instead of the `slug`).

To test:
There should be no visible changes from https://github.com/wordpress-mobile/WordPress-iOS/pull/15484 so smoke testing the site filters should be sufficient.

- Go to Reader > P2s.
  - Select `Filter`.
  - Verify the `Sites` list only contains non-A8C P2s. 
- Go to Reader > Automattic.
  - Select `Filter`.
  - Verify the `Sites` list only contains A8C P2s. 

PR submission checklist:


WordPress/Classes/Models/ReaderTeamTopic.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 
 @objc open class ReaderTeamTopic: ReaderAbstractTopic {
     @NSManaged open var slug: String
+    @NSManaged open var organizationID: NSNumber
```
aerych: In the datamodel, it looks like this is set to use the scalar value. Maybe we could just make this an `Int` vs an `NSNumber`?  This would let us avoid having to constantly call .intValue later.
Fwiw, in the data model properties we could also uncheck "optional" since a default of 0 is provided.

WordPress/Classes/Models/ReaderTeamTopic.swift
```diff
@@ -11,8 +12,8 @@ import Foundation
         return slug == ReaderTeamTopic.a8cSlug ? .readerA8CShown : .readerP2Shown
     }
 
-    var organizationType: SiteOrganizationType {
-        return slug == ReaderTeamTopic.a8cSlug ? .automattic : .p2
+    var organizationType: SiteOrganizationType? {
+        return SiteOrganizationType(rawValue: organizationID.intValue)
```
aerych: Maybe `return SiteOrganizationType(rawValue: organizationID) ?? .none` to avoid the optional?

aerych (commented):
Heya @ScoutHarris 👋 
I had a couple of nitpicky suggestions.  What do ya think?

WordPress/Classes/Models/ReaderTeamTopic.swift
ScoutHarris: I'm probably being thick here, but I don't see "optional" on the Attributes. What am I missing?

Screen Shot 2020-12-14 at 7 08 58 PM

aerych: It's in the sidebar's details for the organizationID attribute.  For me, sometimes Xcode is a little fussy. If the side panel doesn't show the attribute properties when organizationID is selected try tapping to the side.  It's an annoying bug that's been around for a while. :/ 
Screen Recording 2020-12-15 at 9 32 07 AM 2020-12-15 09_35_33

ScoutHarris: > It's in the sidebar's details for the organizationID attribute.

Wowza. I _was_ being thick. I keep the sidebar closed, and it just didn't occur to me to look there. Sorry for the trouble! I've unchecked `optional` for the three instances of `organizationID` I added it to recently (`ReaderTeamTopic`, `ReaderSiteTopic`, `ReaderPost`).

ScoutHarris:
OK @aerych . I believe I got everything if you wanna give it another go. Thank you!!

aerych (approved):
Looks good to me! :shipit:!

PR #15513, opened 2020-12-16
Reader: fix a glitch in the following tab

yaelirub:
Fixing an issue where the following tab is being initially laid out out off to the left of the screen

Gif of the issue:

following_tab

To test:
1. Launch the app 
2. tap the Reader
3. see that the "Following" tab is not jumping from the left 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabItemsStore.swift
@@ -87,6 +87,8 @@ extension ReaderTabItemsStore {
         }
         state = .loading
 
+        // Return the tab bar items right away to avoid waiting for the request to finish
+        fetchTabBarItems()
 
         // Sync the reader menu
         service.fetchReaderMenu(success: { [weak self] in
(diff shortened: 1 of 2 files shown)


leandroalonso:
@yaelirub this call was added on purpose.

Mainly, to prevent this issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14412

_I think_ that if we remove this call and follow the steps described in the issue (fresh install -> Reader) the issue will happen again — and a small glitch is less problematic than waiting (sometimes seconds) for the screen to appear.

leandroalonso:
@yaelirub I managed to fix.

We need to change this guard: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/System/FilterTabBar.swift#L427

Basically, it should be like this:

```
guard scrollView.contentSize.width > scrollView.frame.width, bounds.width > 0 else {
```

The issue is happening because this method is trying to do some calculations while the `bounds` haven't been calculated yet. This solves the issue. :)

yaelirub:
Thank you so so much, @leandroalonso ! I updated to your recommended. Ready of another look

leandroalonso (approved):

yaelirub:
Done in https://github.com/wordpress-mobile/WordPress-iOS/pull/15536

PR #15517, opened 2020-12-16
Log Out: fix a crash in ReaderDetailToolbar.swift

### LINKED ISSUE #15509: Logging out of the app sometimes results in a crash
### Expected behavior

Logging out of the app should take me to the logged out screen.

### Actual behavior

Logging out of the app causes it to crash sometimes (I only experienced this once). Although re-opening the app correctly shows me the logged out screen.

I don't have steps to reproduce, couldn't reproduce after a few tries. What I did when this crash happened was:

1. Downloaded 16.4 beta from TestFlight
2. Had opened navigated to multiple places in the app including Notifications and the Reader
3. Went to log out and the app crashed when I confirmed the log out

### Additional information

The crash shown in TestFlight:

```
0   WordPress                     	0x0000000100ca86e4 Swift runtime failure: Unexpectedly found nil while implicitly unwrapping an Optional value + 0 (ReaderDetailToolbar.swift:186)
```

leandroalonso:
Fixes #15509

### To test

1. Log in into your A8C account in the simulator
2. Enable notifications*
3. Log out
4. You shouldn't see any crash

* I'm really not sure why this step is needed, but I couldn't reproduce the crash in `develop` without enabling notifications.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailToolbar.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailToolbar.swift
@@ -183,7 +183,8 @@ class ReaderDetailToolbar: UIView, NibLoadable {
             return
         }
 
-        likeButton.isEnabled = (ReaderHelpers.isLoggedIn() || post.likeCount.intValue > 0) && !post.isExternal
+        let postLikeCount = post.likeCount?.intValue ?? 0
+        likeButton.isEnabled = (ReaderHelpers.isLoggedIn() || postLikeCount > 0) && !post.isExternal
         // as by design spec, only display like counts
         let likeCount = post.likeCount()?.intValue ?? 0
         let title = likeLabel(count: likeCount)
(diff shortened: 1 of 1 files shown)


WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailToolbar.swift
```diff
@@ -183,7 +183,8 @@ class ReaderDetailToolbar: UIView, NibLoadable {
             return
         }
 
-        likeButton.isEnabled = (ReaderHelpers.isLoggedIn() || post.likeCount.intValue > 0) && !post.isExternal
+        let postLikeCount = post.likeCount?.intValue ?? 0
```
emilylaguna: Is this the same value as line `189`?

`let likeCount = post.likeCount()?.intValue ?? 0`

leandroalonso: It is.

Changed in 63403d140120d5f21c39890a0caa12b8209efd93

emilylaguna (approved):
I confirmed the crash happens using `develop` on an iPhone 11 Pro, and confirmed it is fixed after applying this PR.

PR #15520, opened 2020-12-16
Refresh feature announcements cache

### LINKED ISSUE #15153: Refresh Feature Announcements cache
### Expected behavior
If Announcements have changed for a specific version, the app should reflect that change in "App Settings/What's New in WordPress", even if the user has already updated the app

### Actual behavior
When users update the app, they see Feature Announcements (if available) and those announcements are stored in the local cache and don't get refreshed, even if the announcements on the backend have changed for that same version

Gio2018:
Fixes #15153

To test:

- Delete the app from the device you are using
- Checkout and Install a previous version (e.g `develop`)
- Checkout this branch
- Manually change the version to 50.3
- Verify that the announcement screen (see below) appears at startup, and it's also available in "App Settings/What's New in WordPress"
- Go to Mission Control, and from the mobile announce tool, make some edits to announce 141
- Wait some time to allow the edits to propagate (I have experienced waits up to 10 minutes)
- Navigate to a different tab (e.g. "Reader"), then go back to "My Site" and then "App Settings/What's New in WordPress"
- Make sure the changes appear in the announcements.

**Important notes**
1. **When in the mobile announce tool, be sure to only edit the announce 141, to avoid changing production announcements**
2. To conveniently test this PR, you will need to tweak the property `cacheExpirationTime` in `CachedAnnouncementsStore` (unless, of course, you'd prefer to wait 24 hours to complete the test). Some options are:
    - temporarily set it to 0: this will cause the cache to always update
    - tweak the `cacheExpired` computed properties by using `.minute` instead of `.hour` and then set `cacheExpirationTime` to a low enough value (e.g. 1 minute).
  

example screenshot of announcements

<p align=center>

</p>

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/What's New/Data store/AnnouncementsStore.swift
+++ b/WordPress/Classes/ViewRelated/What's New/Data store/AnnouncementsStore.swift
@@ -84,6 +84,7 @@ class CachedAnnouncementsStore: AnnouncementsStore {
         state = .loading
         if let announcements = cache.announcements, cacheIsValid(for: announcements) {
             state = .ready(announcements)
+            updateCacheIfNeeded()
             return
         }
         // clear cache if it's invalid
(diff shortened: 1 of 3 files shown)


frosty (approved):
I followed the steps, and this worked as expected! :shipit:

Issue #15483, opened 2020-12-11
Reader: Add unread count for posts


mindgraffiti:
Ref. pbArwn-1p9-p2#comment-2018 

As a user, I want to see the unread counts for posts in the Reader section of the app. 
Unread count -> currently returned for followed sites under `/v1.2/read/following/mine` as `unseen_count`.

### Tasks
- [x] Reader: add `.unseenPostCount` feature flag. [#15508]
- [x] Reader: map `unseen_count` to `RemoteReaderSiteInfo` and add `unseenCount` property to `ReaderSiteTopic`. [#15527] and [wordpress-mobile/WordPressKit-iOS/pull/319]
- [x] Followed Sites Filter: generate a new view with the unread post count and attach it to the accessory view in sites filter bottom sheet. [#15581]
- [x] Right-to-Left support. [#15581]
- [x] VoiceOver testing. [#15581]
- [x] Remove Feature Flag. [#15586]

### Context on the web

In the new P2 design, there are unread counts for posts, comments, and mentions. 

| P2 on the web | Reader on the web |
| --- | --- |
| <img width="300" alt="Screen Shot 2020-11-20 at 8 58 55 AM" src="https://user-images.githubusercontent.com/1062444/99814586-c9328580-2b0e-11eb-81c8-488e052bc4f0.png"> | <img width="244" alt="Screen Shot 2020-12-15 at 4 06 33 PM" src="https://user-images.githubusercontent.com/1062444/102278532-9c944280-3eef-11eb-8b85-1849661b1201.png"> |

### Designs for mobile:

![unread-counts-1](https://user-images.githubusercontent.com/1062444/100663430-df64f080-331b-11eb-9893-bd650dca8ea9.png)
![unread-counts-specs](https://user-images.githubusercontent.com/1062444/100663439-e2f87780-331b-11eb-87f0-161e85b66b43.png)

PR #15527, opened 2020-12-16
Reader: Map and store `unseen_count` for unread post counts

mindgraffiti:
Ref. #15483 
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/319

This PR takes the `unseenCount` value from `RemoteReaderSiteInfo` and stores it in core data as the `unseenCount` attribute in the `ReaderSiteTopic` entity. 

### To test

1. Set a breakpoint in ReaderTopicService.m L55.
2. Check the output in the variables inspector panel. Example: 

`_unseenCount	__NSCFNumber *	(long)3	0xfa871a3e58ac89af`

PR submission checklist:


WordPress/Classes/Models/ReaderSiteTopic.swift
```diff
@@ -18,6 +18,7 @@ import Foundation
     @NSManaged open var siteID: NSNumber
     @NSManaged open var siteURL: String
     @NSManaged open var subscriberCount: NSNumber
+    @NSManaged open var unseenCount: NSNumber
```
ScoutHarris: It would be nice to declare this as an `Int` here so when it's referenced it doesn't have to be converted (ex `organizationID`). It can be converted sooner, in `ReaderTopicService`.

WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>WordPress 108.xcdatamodel</string>
+	<string>WordPress 109.xcdatamodel</string>
```
ScoutHarris, line 6: Since a default value for `unseenCount` is provided in WPKit, it'd be good to uncheck `optional` in the data model for it.

Screen Shot 2020-12-16 at 5 01 36 PM

ScoutHarris (approved):
Just a couple minor requests. Otherwise :shipit: .

mindgraffiti:
Thanks for the review @ScoutHarris !

Issue #14787, opened 2020-08-28
SIGABRT: Cannot form weak reference to instance (0x281511310) of class WordPressKit.WordPressOrgXMLRPCApi....


jkmassel:
**Sentry Url**: https://sentry.io/share/issue/8b016bafccf045cebde095bc0ce0525d/
**User Count**: 14
**Count**: 121
**First Release**: 15.4.0.0
**First Seen**: 2020-07-28T08:36:13Z
**Last Seen**: 2020-08-28T15:57:19Z
**24 Hours**: 29
**30 Days**: 118
**LogID:** 2D143B8C-8563-445B-A9A2-69AB925E6F86

```
SIGABRT: Cannot form weak reference to instance (0x281511310) of class WordPressKit.WordPressOrgXMLRPCApi....
  File "WordPressOrgXMLRPCApi.swift", line 53, in WordPressOrgXMLRPCApi.makeSessionManager
  File "WordPressOrgXMLRPCApi.swift", line 31, in WordPressOrgXMLRPCApi.sessionManager.getter
  File "WordPressOrgXMLRPCApi.swift", line 29, in WordPressOrgXMLRPCApi.sessionManager.getter
  File "WordPressOrgXMLRPCApi.swift", line 90, in WordPressOrgXMLRPCApi.__deallocating_deinit
  File "<compiler-generated>", line null, in WordPressOrgXMLRPCApi.__deallocating_deinit
  File "Classes/Models/Blog.m", line 109, in -[Blog didTurnIntoFault]
...
(17 additional frame(s) were not displayed)
```

designsimply:
Noticed there's a similar-sounding error for Gutenberg Mobile at https://github.com/wordpress-mobile/WordPress-iOS/issues/13758.

bummytime:
@bjtitus I see [you updated some refs](https://github.com/wordpress-mobile/WordPressKit-iOS/commit/b84d27ac10b0690493edd9dd9a96d4ebab4fab12) in `WordPressKit/WordPressOrgXMLRPCApi.swift` on July 20th. Any chance this is related? 😄

bjtitus:
[...]

PR #15543, opened 2020-12-18
Replace lazy Gutenberg var with local variable

### LINKED ISSUE #13758: SIGABRT: Cannot form weak reference to instance (0x15c295c00) of class WordPress.GutenbergViewController. ...
https://sentry.io/share/issue/412d8f3542594abc829b4aeef91b681f/

```
SIGABRT: Cannot form weak reference to instance (0x15c295c00) of class WordPress.GutenbergViewController. It is possible that this object was over-released, or is in the process of deallocation. > objc[7526]: Cannot form weak reference to instance (0x15c295c00) of class WordPress.GutenbergViewController. It is possible that this object was over-released, or is in the process of deallocation.

  File "Gutenberg.swift", line 73, in Gutenberg.init
  File "<compiler-generated>", in Gutenberg.__allocating_init
  File "GutenbergViewController.swift", line 208, in GutenbergViewController.gutenberg.getter
  File "GutenbergViewController.swift", line 207, in GutenbergViewController.gutenberg.getter
  File "GutenbergViewController.swift", line 255, in GutenbergViewController.__deallocating_deinit
...
(35 additional frame(s) were not displayed)
```

bjtitus:
Fixes #13758

A similar crash was fixed earlier which occurs when the class (`GutenbergViewController`) calls `deinit` before using the `gutenberg` property. This causes the lazy property to be created _during_ deinit when we want to treat it as an optional.

#### On `develop`
- Apply [this patch](https://gist.github.com/bjtitus/f0392902dd32699a95965be861df60da).
- Open the Post Editor and verify that the crash occurs.

#### On `fix/13758-gutenberg-lazy-var`
- Apply [this patch](https://gist.github.com/bjtitus/f0392902dd32699a95965be861df60da).
- Open the Post Editor and verify no crash occurs.
- Ensure that the Gutenberg editor is running and behaves properly
- Ensure that media uploads work as expected (this should directly exercise the Gutenberg property).

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
@@ -267,9 +267,15 @@ class GutenbergViewController: UIViewController, PostEditor {
         return GutenbergImageLoader(post: post)
     }()
 
-    private lazy var gutenberg: Gutenberg = {
-        return Gutenberg(dataSource: self, extraModules: [gutenbergImageLoader])
-    }()
+    private var _gutenberg: Gutenberg?
+    private var gutenberg: Gutenberg {
+        guard let gutenberg = _gutenberg else {
+            let gutenberg = Gutenberg(dataSource: self, extraModules: [gutenbergImageLoader])
+            _gutenberg = gutenberg
+            return gutenberg
+        }
+        return gutenberg
+    }
 
     private var requestHTMLReason: RequestHTMLReason?
     private(set) var mode: EditMode = .richText
@@ -321,7 +327,7 @@ class GutenbergViewController: UIViewController, PostEditor {
     deinit {
         tearDownKeyboardObservers()
         removeObservers(fromPost: post)
-        gutenberg.invalidate()
+        _gutenberg?.invalidate()
         attachmentDelegate.cancelAllPendingMediaRequests()
     }
 


jkmassel:
👋 We're freezing WPiOS `16.5` today, so this PR is being bumped to `16.6`. 

If you need it to be part of the `16.5` release, please DM me – I'll be happy to cut a new beta release once this lands!

SergioEstevao:
@guarani or @ceyhun do you mind giving a look at this one?

guarani (commented):
Thanks for the PR, @bjtitus! Were steps to reproduce this crash consistently ever found?

My understanding is that this PR changes the `gutenberg` property from a lazy non-optional variable to an optional singleton type variable. If the goal is to ensure `gutenberg` is not instantiated when its `invalidate()` method is called, could it just be converted to a lazy _optional_  variable? This produces about 33 compile errors though, so we'd need to change a lot of places in `GutenbergViewController` to handle `gutenberg` being optional.

I'm looking to better understand under what conditions is `gutenberg` is never accessed, so that we can either keep it non-optional or make it optional. Is there a user flow you're aware of that reproduces this crash?

On the other hand, do you know why accessing `gutenberg` caused the crash in the first place? It seems to me like accessing an uninitiated lazy variable in `deinit` shouldn't cause a crash.

jkmassel:
👋 Howdy! We are cutting the `16.6` release today.

Because of that, this PR will be bumped to `16.7`. If you need it to be part of `16.6`, please merge it into the `release/16.6` branch and DM me – I'll be happy to cut a new beta release!

bjtitus:
@Guarani Thanks for taking a look!

> Were steps to reproduce this crash consistently ever found?

I don't have _true_ reproduction steps but it is possible to reproduce this with a small code change to instantiate and quickly deallocate a `GutenbergViewController` instance. I've uploaded [a patch](https://gist.github.com/bjtitus/f0392902dd32699a95965be861df60da) and updated the PR details with testing instructions.

This [sample code](https://gist.github.com/bjtitus/5140023db8fd42a8aeb96c67ad74a5fd) may prove more useful in debugging and reviewing the core cause of the crash, though.

> could it just be converted to a lazy optional variable

It doesn't look like this solves the issue from my testing with the [sample code](https://gist.github.com/bjtitus/5140023db8fd42a8aeb96c67ad74a5fd). I think it could be possible to use a property behavior (as [suggested](https://forums.swift.org/t/lazy-var-and-deinit/2291/2) by a Swift core member) to achieve this but I spent longer than I probably should have trying to get that to work. I couldn't get past some of the property behavior limitations with referencing `self`.

> It seems to me like accessing an uninitiated lazy variable in deinit shouldn't cause a crash

The crash here is caused by the attempt to create a weak reference to `GutenbergViewController` during deinitialization. As the crash error message indicates, the `GutenbergViewController` instance is in the process of deallocation which seems to cause a crash in `NSObject` subclasses.

```
Cannot form weak reference to instance (0x102b1b800) of class WordPress.GutenbergViewController. It is possible that this object was over-released, or is in the process of deallocation.
```

Further background from the original hunch are [here](https://github.com/wordpress-mobile/WordPress-iOS/issues/14787#issuecomment-685162840).

guarani:
> I don't have _true_ reproduction steps but it is possible to reproduce this with a small code change to instantiate and quickly deallocate a `GutenbergViewController` instance. I've uploaded [a patch](https://gist.github.com/bjtitus/f0392902dd32699a95965be861df60da) and updated the PR details with testing instructions.

I could reproduce the crash with that patch, very useful – thanks!

> This [sample code](https://gist.github.com/bjtitus/5140023db8fd42a8aeb96c67ad74a5fd) may prove more useful in debugging and reviewing the core cause of the crash, though.

I could also reproduce with this sample code.

> > It seems to me like accessing an uninitiated lazy variable in deinit shouldn't cause a crash
> 
> The crash here is caused by the attempt to create a weak reference to `GutenbergViewController` during deinitialization. As the crash error message indicates, the `GutenbergViewController` instance is in the process of deallocation which seems to cause a crash in `NSObject` subclasses.
> 
> ```
> Cannot form weak reference to instance (0x102b1b800) of class WordPress.GutenbergViewController. It is possible that this object was over-released, or is in the process of deallocation.
> ```


TIL 🙇. I came across this [great SO question](https://stackoverflow.com/questions/35991363/why-setting-object-that-is-undergoing-deallocation-to-weak-property-results-in-c) which I think covers the issue we're seeing here. The TL;DR; is that (I think) it's not well understood if that's intentional behavior or not that a crash is triggered when an object being deallocated is assigned to a weak property. 

One alternative to the solution proposed here is removing the `lazy` attribute to ensure the weak reference to `GutenbergViewController` is established well before the view controller is deallocated. The relevant code that sets the weak reference is:

https://github.com/WordPress/gutenberg/blob/5ced23696aa74ecbb7f20e6d1c7af9593221353a/packages/react-native-bridge/ios/Gutenberg.swift#L105

The only potential downside to removing lazy is a degraded performance. However, as you mentioned offline, `gutenberg` although lazy, is accessed in `viewDidLoad`. So if we convert it to a non-lazy property performance should stay the similar and we can ensure that the circumstances leading to this particular crash won't occur. There could be a difference in performance because we'd now be initializing `gutenberg` in the view controller's initializer, but I gave it a quick test and couldn't feel the difference using a simulator. I'd be happy to check with a real device tomorrow though.

jkmassel:
👋 We're freezing `16.7` today, so this PR is being bumped to `16.8`. If you need this to be part of the `16.7` release, please merge it into the `release/16.7` branch and DM me – I'll be happy to cut a new beta release!

bjtitus:
@Guarani Sorry for the delay on this one. I keep pushing it off to future releases as the impact is relatively small.

> One alternative to the solution proposed here is removing the lazy attribute to ensure the weak reference to GutenbergViewController is established well before the view controller is deallocated.

Are you proposing the following?

```patch
+++private var gutenbergImageLoader: GutenbergImageLoader!
+++private var gutenberg: Gutenberg!
```

```patch
required init(
    post: AbstractPost,
    loadAutosaveRevision: Bool = false,
    replaceEditor: @escaping (EditorViewController, EditorViewController) -> (),
    editorSession: PostEditorAnalyticsSession? = nil) {

    self.post = post
    self.loadAutosaveRevision = loadAutosaveRevision

    self.replaceEditor = replaceEditor
    verificationPromptHelper = AztecVerificationPromptHelper(account: self.post.blog.account)
    self.editorSession = PostEditorAnalyticsSession(editor: .gutenberg, post: post)

+++    let imageLoader = GutenbergImageLoader(post: post)
+++    self.gutenbergImageLoader = imageLoader
+++    self.gutenberg = Gutenberg(dataSource: self, extraModules: [imageLoader])

    addObservers(toPost: post)

    PostCoordinator.shared.cancelAnyPendingSaveOf(post: post)
    navigationBarManager.delegate = self
}
```

[GutenbergInitializer.patch.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/6400238/GutenbergInitializer.patch.zip)

It seems something like this would work, although the safety seems lacking with the implicit unwraps.

LMK if I misunderstood your suggestion.

jkmassel:
👋 We're freezing 17.3 today, so this PR is being bumped to 17.4

guarani:
Hi @bjtitus, I think I had in mind something simpler (that avoided the implicitly unwrapped optionals), but I can't think of what it was right now. Will need to give this a bit more thought.

jkmassel:
👋 This has been pushed several times, so I'm going to move it to the `Someday` milestone 🙂

twstokes:

PR #15545, opened 2020-12-18
Merge 16.4 release notes to develop

jkmassel:
Merges 16.4 release notes into `develop`.

Because of some weirdness with the initial code freeze merge, the changes from the second beta are already in `develop`, but the version numbers aren't (I had to re-create the merge branch a few times to get the merge to work correctly).

That's the reason for the version change being part of this PR


mokagio (approved):
👍 

I can see that some (all?) the `release_notes.txt` source files for `deliver` are gone, which is what I'd expect given that this PR is updating the release notes and those files are now obsolete.

PR #15547, opened 2020-12-18
Move WordPressFlux to a Swift Package

jkmassel:
This has several benefits, namely:

- We don’t need to deal with maintaining a separate set of build settings for it, which will help with the transition to Apple Silicon, and will be one less thing to manage.
- Swift Packages are just represented by the files on disk, rather than using a manifest as in `.xcproj` files. This lets us delete a lot of not-so-useful configuration that's merge error-prone.
- No code signing or release/internal/debug configurations are required anymore.

Unfortunately, this PR necessitates a re-formatting of the `.xcodeproj` file by Xcode (which is where the vast majority of the additions in this PR come from).

**To test:**
- Ensure all tests pass

PR submission checklist:


frosty (approved):
This looks good to me! I'm seeing one Swift compiler warning in WordPressFlux that I didn't see before, but it's not a dealbreaker:

```
QueryStore.swift:120:49: 'index(where:)' is deprecated: renamed to 'firstIndex(where:)'
```

I also have two tests failing, but that're unrelated and I'll report them elsewhere.

jkmassel:
Thanks @frosty!

I've fixed the deprecation warning in 6e33103c3b198fb7bac34d57093c4e2197f3ff27 and I'll put this on hold until the unit test is fixed! 😄

frosty:
@jkmassel Looks like the tests have been fixed in develop: https://github.com/wordpress-mobile/WordPress-iOS/commit/8b38b17745864a050bb101a66dbaca1e022eb0f2

PR #15558, opened 2020-12-24
Add 60s throttle to suggestion requests

guarani:
Previously the contents of the suggestions UI (Xpost and Mention options) were cached until the app was restarted. Now a 60 second thottle limits the number of calls made to the server, by using cached results if available and if not older than the duration of the throttle.

Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/2602

### To test

**What you'll need**: A WordPress.com site that is capable of xposting to one or more other sites

_Note: https://wordpress.com/p2/ can be used to create xpost-capable sites._

**Xpost in Gutenberg**

1. Open the block editor
2. In a Paragraph or any other rich text block, type `+` and verify that the list of Xpost suggestions is shown after a slight network delay
3. Dismiss the suggestions UI, type `+` again and verify that the list of suggestions is shown instantly
4. After slightly more than 60 seconds has elapsed since Step 2, type `+` again and ensure that a slight delay is apparent before the UI loads again

**Mentions in Gutenberg**
Repeat steps 1 to 4 above, but use `@` instead so as to trigger the Mentions UI

**Smoke test Mentions outside of Gutenberg**
1. Go to a post in Reader and make sure Mentions work in comments
2. Go to a sites Comments section (My Site → Comments), select a comment, and make a reply that includes a Mention
3. Check anywhere else Mentions are used (e.g. Notification replies)

Note: There's room here to reduce code duplication and improve naming between `SuggestionService` and `SiteSuggestionService`: https://github.com/wordpress-mobile/gutenberg-mobile/issues/2932

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/SiteSuggestionService.swift
+++ b/WordPress/Classes/Services/SiteSuggestionService.swift
@@ -4,6 +4,7 @@ import Foundation
 class SiteSuggestionService {
 
     private var blogsCurrentlyBeingRequested = [Blog]()
+    private var requests = [Blog: Date]()
 
     static let shared = SiteSuggestionService()
 
@@ -15,7 +16,16 @@ class SiteSuggestionService {
     */
     func suggestions(for blog: Blog, completion: @escaping ([SiteSuggestion]?) -> Void) {
 
-        if let suggestions = retrievePersistedSuggestions(for: blog), suggestions.isEmpty == false {
+        let throttleDuration: TimeInterval = 60 // seconds
+        let isBelowThrottleThreshold: Bool
+        if let requestDate = requests[blog] {
+            isBelowThrottleThreshold = Date().timeIntervalSince(requestDate) < throttleDuration
+        } else {
+            isBelowThrottleThreshold = false
+        }
+
+
+        if isBelowThrottleThreshold, let suggestions = retrievePersistedSuggestions(for: blog), suggestions.isEmpty == false {
             completion(suggestions)
         } else if ReachabilityUtils.isInternetReachable() {
             fetchAndPersistSuggestions(for: blog, completion: completion)
@@ -62,6 +72,8 @@ class SiteSuggestionService {
 
             // Save the changes
             try? ContextManager.shared.mainContext.save()
+
+            self.requests[blog] = Date()
 
             completion(suggestions)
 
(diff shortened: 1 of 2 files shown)


WordPress/Classes/Services/SiteSuggestionService.swift
```diff
@@ -15,7 +16,16 @@ class SiteSuggestionService {
...
     func suggestions(for blog: Blog, completion: @escaping ([SiteSuggestion]?) -> Void) {
 
-        if let suggestions = retrievePersistedSuggestions(for: blog), suggestions.isEmpty == false {
+        let throttleDuration: TimeInterval = 60 // seconds
+        let isBelowThrottleThreshold: Bool
+        if let requestDate = requests[blog] {
+            isBelowThrottleThreshold = Date().timeIntervalSince(requestDate) < throttleDuration
+        } else {
+            isBelowThrottleThreshold = false
+        }
```
guarani: @jleandroperez, I wish I knew a Swifty way of cleaning this up into a one-liner.

WordPress/Classes/Services/SuggestionService.swift
```diff
@@ -32,16 +41,18 @@ class SuggestionService {
     */
     private func fetchAndPersistSuggestions(for blog: Blog, completion: @escaping ([UserSuggestion]?) -> Void) {
 
+        guard let blogId = blog.dotComID, let hostname = blog.hostname else { return }
```
jleandroperez:
```suggestion
        guard let blogId = blog.dotComID, let _ = blog.hostname else { return }
```

jleandroperez (approved):
Looks good @guarani !!

:shipit: when ready (one tiny suggestion added!)

WordPress/Classes/Services/SiteSuggestionService.swift
jleandroperez: I think this is the most readable approach! Anything else would probably involve using `map`, but you'd need to deal with the `nil` scenario as well (which is not cool!!)

```
let isBelowThrottleThreshold = requests[blog].map { date in
      Date().timeIntervalSince(requestDate) < throttleDuration
}  ?? false
```

guarani: Yeah, using map like this is an alternative. Thanks for the suggestion 👍

WordPress/Classes/Services/SuggestionService.swift
guarani: Thanks for catching! This ended up being an unused variable I'd forgotten to delete 🤦

guarani:
Thanks a lot for the review, @jleandroperez! Much appreciated!

jleandroperez:
Anytime @guarani !!! remember I owe you a million reviews, keep the PRs coming please!!

🔥

PR #15139, opened 2020-10-20
Xpost Suggestions for Gutenberg

guarani:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/2602
Related PRs:
- https://github.com/wordpress-mobile/gutenberg-mobile/pull/2746
- https://github.com/WordPress/gutenberg/pull/26211
- https://github.com/wordpress-mobile/WordPress-Android/pull/13184

This PR adds to Gutenberg the ability to xpost to another site.

### To test

**What you'll need**: A WordPress.com site that is capable of xposting to one or more other sites
- Make sure some of the destination sites have a blavatar photo (site photo) set
- The ability to verify that the destination sites were successfully xpost'ed to

_Note: https://wordpress.com/p2/ can be used to create xpost-capable sites._

**Xpost in Gutenberg**

- Open the block editor
- In a Paragraph or any other rich text block, type <code>+</code> and verify that the list of suggestions is shown
- Make sure that the blavatar images load for sites which have a blavatar image
- Select one of the suggestions to add it to the post
- Expect the site name (not its URL) to be added to the post
- Publish the post
- Verify that the destination site received the xpost

Please note: There are improvements coming to the Xpost feature before it will be shipped, this PR aims to bring Xposts the same level of functionality Mentions currently has, nothing more, nothing less. (Improvements include a request throttle, prefetching, analytics, etc.)

### PR submission checklist

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15562, opened 2020-12-29
Prefetch xpost suggestions in Gutenberg

guarani:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/2602

This PR prefetches the suggestion data that populates the Xpost suggestion UI. Without this, the user experiences a network delay when first inserting an Xpost suggestion in the block editor.

#### Expected behavior 
https://user-images.githubusercontent.com/1898325/103297740-b3e72b80-49d7-11eb-8abe-8ca42dd9b382.MP4

Note: The throttle introduced in https://github.com/wordpress-mobile/WordPress-iOS/pull/15558 is a 60 second throttle. This means that the user will still experience the network delay if they wait for more than 60 seconds between opening the editor and inserting an Xpost suggestion.

### To test

**What you'll need**: Two or more WordPress.com sites that are capable of xposting (https://wordpress.com/p2/ can be used to create xpost-capable sites). It's best if you can test with a site that has lots of Xpost suggestions (10 to 20), to ensure that the network delay would be noticeable if present.

**Xpost in Gutenberg**

1. Open the block editor
2. In a Paragraph or any other rich text block, type `+` and verify that the list of Xpost suggestions is shown immediately (no network delay should be noticeable)
3. Switch to a different site from within the editor
4. Type `+` (or long press the `@` toolbar button and select Xpost) and expect the list of suggestions to be shown immediately

Note: There is an issue which @antonis first reported in https://github.com/wordpress-mobile/WordPress-iOS/pull/15139#discussion_r548485770 where sites that don't have Xposts suggestions (e.g. non-P2 sites) show the suggestions UI for an instant while the network loads. I didn't address that here because I had difficulty updating the Gutenberg capability when switching between sites while in the the editor. I created https://github.com/wordpress-mobile/gutenberg-mobile/issues/2942 to address this in a later PR.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/SiteSuggestionService.swift
+++ b/WordPress/Classes/Services/SiteSuggestionService.swift
@@ -35,6 +35,17 @@ class SiteSuggestionService {
     }
 
     /**
+     If no suggestions already in Core Data, fetch them from the network and store them in Core Data.
+     @param blog The blog/site to prefetch suggestions for
+     */
+    func prefetchSuggestions(for blog: Blog) {
+        let persistedSuggestions = retrievePersistedSuggestions(for: blog)
+        if persistedSuggestions == nil || persistedSuggestions?.isEmpty == true {
+            fetchAndPersistSuggestions(for: blog, completion: { _ in})
+        }
+    }
+
+    /**
     Performs a REST API request for the given blog.
     Persists response objects to Core Data.
 
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
@@ -345,6 +345,8 @@ class GutenbergViewController: UIViewController, PostEditor {
         }, failure: { (error) in
             DDLogError("Error syncing JETPACK: \(String(describing: error))")
         })
+
+        SiteSuggestionService.shared.prefetchSuggestions(for: self.post.blog)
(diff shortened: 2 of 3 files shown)


guarani:
Adding you here @fluiddot since you mentioned you'd like to review/test this. If you're not comfortable reviewing the code, a test is still much appreciated 🙇 and you can just leave a comment instead of an approval.

jleandroperez (approved):
✅ Suggestions delay is gone
✅ Switching to another site also prefetches suggestions

Looks good! :shipit: sir!!

fluiddot (approved):
It works perfect, nice work 🎊 !

**Minor issue:** The only thing I noted which I don't know if it's by design is that when you have only the symbol `+`/`@` in the text input and press backspace it hides the x-post UI but doesn't remove the character from the paragraph block. So if you want to trigger the UI again you have to remove the symbol and input it again.

guarani:
Thank you!

> Minor issue: The only thing I noted which I don't know if it's by design is that when you have only the symbol +/@ in the text input and press backspace it hides the x-post UI but doesn't remove the character from the paragraph block. So if you want to trigger the UI again you have to remove the symbol and input it again.

Thanks for reporting that @fluiddot! I think this is a bug that's already in production (tested on iOS and on the Android beta) so I created https://github.com/wordpress-mobile/gutenberg-mobile/issues/2943.

PR #15555, opened 2020-12-23
Jetpack Section: Present action sheet for backup and restore

momo-ozawa:
Part of #15191 

- Added `.jetpackBackupAndRestore` feature flag 
- Updated `ActivityTableViewCell`
    - Updated gridicon to ellipsis 
    - Present an action sheet when tapping gridicon

Activity Log | Action Sheet
--- | ---
Screen Shot 2020-12-24 at 9 16 45 | Screen Shot 2020-12-24 at 9 16 51

 
#### `.jetpackBackupAndRestore` feature flag ON

1. My Site > Activity Log > rewindable activity cell
    - ✅ Rewindable activity cells should display an ellipsis icon instead of a rewind icon
2. Tap on ellipsis icon
    - ✅ Should present action sheet
    - ⚠️ Tapping on `Restore`, `Download Backup` won't do anything atm
3. Go back to activity log
4. Tap on activity cell
    - ✅ Should navigate to activity detail

#### `.jetpackBackupAndRestore` feature flag OFF

1. My Site > Activity Log > rewindable activity cell
    - ✅ Rewindable activity cells should display a rewind icon
2. Tap on rewind icon
    - ✅ Should navigate to activity detail
3. Go back to activity log
4. Tap on activity cell
    - ✅ Should navigate to activity detail

### PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #15191, opened 2020-10-29
Jetpack Section: Backup and Restore detail


emilylaguna:
Project: https://github.com/wordpress-mobile/WordPress-iOS/issues/15193

![Screen Shot 2021-01-04 at 13 19 42](https://user-images.githubusercontent.com/6711616/103500816-901c5d80-4e8f-11eb-9cc8-183cd005459b.png)

#### Network
- [x] Add endpoints (https://github.com/wordpress-mobile/WordPressKit-iOS/pull/321)

#### General
- [x] Replace occurrences of Rewind with Restore to be inline with current branding (#15817)

#### Activity Log
- [x] Update backup table cell to use ⋮ ellipsis icon (https://github.com/wordpress-mobile/WordPress-iOS/pull/15555)
- [x] Tap ⋮ ellipsis icon > present action sheet (https://github.com/wordpress-mobile/WordPress-iOS/pull/15555)
- [x] Tap `Restore` > present Restore Options (https://github.com/wordpress-mobile/WordPress-iOS/pull/15555)
- [x] Tap `Download backup` > present Backup Options (https://github.com/wordpress-mobile/WordPress-iOS/pull/15555)
- [x] Display restore progress cell after returning to activity log from restore flow (https://github.com/wordpress-mobile/WordPress-iOS/pull/15689)
- [x] Display backup progress cell after returning to activity log from backup flow (#15776)

#### Activity Details
- [x] Route `Restore` to Restore Options (https://github.com/wordpress-mobile/WordPress-iOS/pull/15598)
- [x] Add `Download backup` cell (https://github.com/wordpress-mobile/WordPress-iOS/pull/15683)
- [x] Route `Download backup` to Backup Options (https://github.com/wordpress-mobile/WordPress-iOS/pull/15683)

#### Download / Restore Options View
- [x] Create a reusable view controller for both `Restore` and `Download Backup` actions (https://github.com/wordpress-mobile/WordPress-iOS/pull/15566)
- [x] Tap `Cancel` > dismiss (https://github.com/wordpress-mobile/WordPress-iOS/pull/15566)
- [x] Tap `Download` button > push Backup Status  (https://github.com/wordpress-mobile/WordPress-iOS/pull/15566)
- [x] Tap `Restore` button > push Restore Warning (https://github.com/wordpress-mobile/WordPress-iOS/pull/15566)
[...]

PR #15566, opened 2021-01-04
Jetpack Section: Add Backup/Restore Reusable VC

momo-ozawa:
Part of #15191

⚠️ Please review https://github.com/wordpress-mobile/WordPress-iOS/pull/15555 first... thanks!

- Added a new resusable VC for Download Backup / Restore actions
- _Switches/buttons will be hooked up later_
- _Styling will be updated later_

Restore | Download Backup
--- | ---
Screen Shot 2021-01-04 at 15 31 01 | Screen Shot 2021-01-04 at 15 31 08

#### Restore
1. My Site > Activity Log > rewindable activity cell
2. Tap on ellipsis icon
3. Tap on `Restore`
    - ✅ Should display `Restore` view modally
4. Tap on `Cancel`
    - ✅ Should dismiss VC 

#### Download Backup
1. My Site > Activity Log > rewindable activity cell
2. Tap on ellipsis icon
3. Tap on `Download Backup`
    - ✅ Should display `Download Backup` view modally
4. Tap on `Cancel`
    - ✅ Should dismiss VC 
    
### PR submission checklist:


WordPress/Classes/ViewRelated/Jetpack/Jetpack Restore/JetpackRestoreViewController.swift
```diff
@@ -0,0 +1,169 @@
...
+        )
+        let rootRow = SwitchRow(
+            title: NSLocalizedString("WordPress root", comment: "Downloadable/Restorable items: WordPress root"),
+            value: true,
+            onChange: { _ in }
+        )
+
+        let headerText: String
+        switch restoreAction {
+        case .restore:
```
leandroalonso, line 125: We already have this `switch` statement in a few different parts. This makes me wonder if we shouldn't break this VC (either by using inheritance or composition).

Right now, as it is, is not that complex. But I fear that this might grow into something huge and messy in the future.

We could have something like `JetpackRestoreViewController` and `JetpackBackupViewController`.

What do you think?

(An example of usage of inheritance (although it has it's own problems too tbh) is the `ReaderCardsStreamViewController`)

leandroalonso (approved):
@momo-ozawa works as described!

When I tap the ellipsis icon I'm seeing an AutoLayout warning:

```
2021-01-04 11:09:31.953733-0300 WordPress[27317:1108158] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x7fa764033130 UIView:0x7fa764037e20.width == - 16   (active)>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x7fa764033130 UIView:0x7fa764037e20.width == - 16   (active)>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
```

Also left an architectural suggestion. Let me know what do you think, you can address that in a future task if you like the idea. Up to you! :)

WordPress/Classes/ViewRelated/Jetpack/Jetpack Restore/JetpackRestoreViewController.swift
momo-ozawa: @leandroalonso 

> Right now, as it is, is not that complex. But I fear that this might grow into something huge and messy in the future.

That's a good point. I'll address this in a future PR!

momo-ozawa:
@leandroalonso The autolayout warning seems to be an iOS bug, so I'm going to go ahead and merge this PR! 
```
<_UIAlertControllerActionView: 0x7fe4c2814e90; frame = (0 0; 374 57); Action = <UIAlertAction: 0x7fe4c265d4a0 Title = "Download Backup" Descriptive = "(null)" Image = 0x0>>
2021-01-06 11:56:41.545558+0900 WordPress[2858:2826191] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x7fe4b2adb360 UIView:0x7fe4c291edc0.width == - 16   (active)>"
)
```

https://stackoverflow.com/questions/55372093/uialertcontrollers-actionsheet-gives-constraint-error-on-ios-12-2-12-3

PR #15581, opened 2021-01-05
Reader site filter: show unseen post count

### LINKED ISSUE #15483: Reader: Add unread count for posts

ScoutHarris:
Ref #15483

On the Reader's site filter, a unseen post count is now displayed for each site.

To test:
- Go to the Reader.
- On any feed that has a filter (Following, Automattic, P2s), tap `Filter`.
- Verify:
  - For sites that have unseen posts, the count is displayed.
  - For sites that have no unseen posts, the count is not displayed.
  - The counts match those in the web Reader.
  - Numbers above 1000 are abbreviated with multipliers(K, M, B, etc.)

To verify large numbers are abbreviated, you can hack the displayed count. In `FilterProvider:addUnseenPostCount`, change the displayed value: `countLabel.text = topic.unseenCount.abbreviatedString()`

Example:
```
let count = 876875666978
countLabel.text = count.abbreviatedString()
```

---
| unseen_count | abbrevs |
|--------|-------|

| ipad_light | ipad_dark |
|--------|-------|

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
@@ -148,6 +148,10 @@ extension ReaderSiteTopic {
                     return TableDataItem(topic: topic, configure: { cell in
                         cell.textLabel?.text = topic.title
                         cell.detailTextLabel?.text = topic.siteURL
+
+                        if FeatureFlag.unseenPostCount.enabled {
+                            addUnseenPostCount(topic, with: cell)
+                        }
                     })
                 }
             }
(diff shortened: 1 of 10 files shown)


mattmiklic (approved):
These are looking good. Thanks @ScoutHarris!

ScoutHarris:
Hey @mattmiklic . Sorry for the ping, but I just noticed something. Is the multiplier for thousands supposed to be upper or lower case "K"? I made it lowercase to [match Android](https://github.com/wordpress-mobile/WordPress-Android/pull/13625), but I see now the designs have uppercase. Thank you!

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
```diff
@@ -43,7 +43,7 @@ extension Double {
                 static let units: [Unit] = {
                     var units: [Unit] = []
 
-                    units.append(Unit(abbreviationFormat: NSLocalizedString("%@K", comment: "Label displaying value in thousands. Ex: 66.6K."), accessibilityLabelFormat: NSLocalizedString("%@ thousand", comment: "Accessibility label for value in thousands. Ex: 66.6 thousand.")))
+                    units.append(Unit(abbreviationFormat: NSLocalizedString("%@k", comment: "Label displaying value in thousands. Ex: 66.6k."), accessibilityLabelFormat: NSLocalizedString("%@ thousand", comment: "Accessibility label for value in thousands. Ex: 66.6 thousand.")))
```
aerych: Was going to ask if we wanted the upper case K to match the other abbreviations, but I see you already asked :)

WordPress/Classes/ViewRelated/Reader/Filter/FilterProvider.swift
```diff
@@ -180,6 +184,48 @@ extension ReaderSiteTopic {
         completion(.success(sites))
     }
 
+    /// Adds a custom accessory view displaying the unseen post count.
+    ///
+    private static func addUnseenPostCount(_ topic: ReaderSiteTopic, with cell: UITableViewCell) {
+
```
aerych, line 190: Wondering if it would be better to set the accessory view to nil here.  Refs my comment in prepareForReuse.

WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
```diff
@@ -43,6 +43,10 @@ class SiteTableViewCell: UITableViewCell, GhostableView {
         fatalError("init(coder:) has not been implemented")
     }
 
+    override func prepareForReuse() {
+        accessoryView = nil
```
aerych: I think we'd probably want to call `super` here. :) 

However, I'm wondering if this would be better done elsewhere.  Per the [docs](https://developer.apple.com/documentation/uikit/uitableviewcell/1623223-prepareforreuse): 
> For performance reasons, you should only reset attributes of the cell that are not related to content, for example, alpha, editing, and selection state.

Changes that modify the view hierarchy seem like they would be content related.

aerych (changes requested):
Heya @ScoutHarris ! 
I tested and the changes look good. 
I had one nitpick code wise. Apart from that and the question about `k` vs `K` this should be good to go. :)

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
ScoutHarris: I went ahead and reverted the change (i.e. make it uppercase again) since I think we all agree on it. If @mattmiklic rebels, I can change it later. 😄

WordPress/Classes/ViewRelated/Reader/Filter/FilterTableData.swift
ScoutHarris: > I think we'd probably want to call super here.

Every. Time. 🤦 

> I'm wondering if this would be better done elsewhere.

Thanks for that! Tis been moved.

ScoutHarris:
Thanks @aerych ! Ready for another pass please.

aerych (approved):
Shiny! :shipit:

WordPress/Classes/ViewRelated/Stats/Extensions/Double+Stats.swift
mattmiklic: No rebellion here, I agree with making this uppercase. :)

PR #15585, opened 2021-01-05
Reader Following site filter: don't show P2 or A8C sites.

### LINKED ISSUE #15578: Reader: update Following to exclude A8C and P2
The web Reader `Followed Sites` now only shows what's not in `Automattic` or `Followed P2s`. It looks like it's been updated to `organization_id = 0` instead of showing everything. We should update the app to match.

ScoutHarris:
Fixes #15578 

This removes P2 and A8C sites from the `Following` site filter to match the web Reader.

The Following stream itself was already filtering them out, so no changes were needed there.

To test:
- Go to Reader > Following.
- Select `Filter`.
- Verify only non-P2/A8C sites are displayed (i.e. the list matches `Followed Sites` in the web Reader).

PR submission checklist:


aerych (approved):
Hola @ScoutHarris 👋 
Change looks good. Confirming that my p2s no longer show under following, but still appear under their respective tabs. 
:shipit: !

Issue #15193, opened 2020-10-29
Jetpack Section: Project Issue


emilylaguna:
This is the main project issue for the Jetpack Section project

### Tasks
- [x] Jetpack My Site Menu Improvements: https://github.com/wordpress-mobile/WordPress-iOS/issues/15275
- [x] Activity Log Filters: https://github.com/wordpress-mobile/WordPress-iOS/issues/15192
- [x] Download Backup/Restore: https://github.com/wordpress-mobile/WordPress-iOS/issues/15191
- [x] Scan: https://github.com/wordpress-mobile/WordPress-iOS/issues/15190
- [x] Update with latest editorial text: https://github.com/wordpress-mobile/WordPress-iOS/issues/15741
- [x] Final Design Review

### Deferred Tasks
- [ ] ~Server Credentials: https://github.com/wordpress-mobile/WordPress-iOS/issues/15505~

emilylaguna:
Closing as this as all the tasks are complete and this will be released in the next version: https://github.com/wordpress-mobile/WordPress-iOS/pull/15813

PR #15590, opened 2021-01-06
Jetpack: adds Backup screen

leandroalonso:
Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/15193

This PR adds the Backup screen under the Jetpack section.

@momo-ozawa I'm using the same feature flag as backup/rewind. Let me know if you think this is a problem. :) 

### To test

1. Open any site detail
2. Tap Backup
3. You should see a list containing ONLY items that can be downloaded or rewinded
4. Scroll down and compare your list to the one in wordpress.com

Filtering by date:

1. Open any site detail
2. Tap Backup
3. Select a single date or a date range
4. You should see only items that can be downloaded or rewinded in the range or day you selected

This PR only covers showing the Backup screen. This will be covered in future PRs:

* Checking if the site has the Backup feature
* Changing the error messages
* Refactoring `ActivityListViewController` and `BackupListViewController` to make the inheritance more obvious
* Updating the events

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Stores/ActivityStore.swift
+++ b/WordPress/Classes/Stores/ActivityStore.swift
@@ -72,6 +72,8 @@ class ActivityStore: QueryStore<ActivityStoreState, ActivityQuery> {
 
     private let activityServiceRemote: ActivityServiceRemote?
 
+    var onlyRewindableItems = false
+
     override func queriesChanged() {
         super.queriesChanged()
         processQueries()
@@ -236,9 +238,16 @@ private extension ActivityStore {
             after: afterDate,
             before: beforeDate,
             group: group,
-            success: { [actionDispatcher] (activities, hasMore) in
+            success: { [weak self, actionDispatcher] (activities, hasMore) in
+                guard let self = self else {
+                    return
+                }
+
                 let loadingMore = offset > 0
-                actionDispatcher.dispatch(ActivityAction.receiveActivities(site: site, activities: activities, hasMore: hasMore, loadingMore: loadingMore))
+                actionDispatcher.dispatch(ActivityAction.receiveActivities(
+                                            site: site,
+                                            activities: self.onlyRewindableItems ? activities.filter { $0.isRewindable } : activities,
+                                            hasMore: hasMore, loadingMore: loadingMore))
         },
             failure: { [actionDispatcher] error in
                 actionDispatcher.dispatch(ActivityAction.receiveActivitiesFailed(site: site, error: error))
--- a/WordPress/Classes/Stores/ActivityStore.swift
+++ b/WordPress/Classes/Stores/ActivityStore.swift
@@ -245,10 +245,13 @@ private extension ActivityStore {
                 }
 
                 let loadingMore = offset > 0
-                actionDispatcher.dispatch(ActivityAction.receiveActivities(
-                                            site: site,
-                                            activities: self.onlyRestorableItems ? activities.filter { $0.isRewindable } : activities,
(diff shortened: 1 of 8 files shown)


WordPress/Classes/Stores/ActivityStore.swift
```diff
@@ -72,6 +72,8 @@ class ActivityStore: QueryStore<ActivityStoreState, ActivityQuery> {
 
     private let activityServiceRemote: ActivityServiceRemote?
 
+    var onlyRewindableItems = false
+
```
momo-ozawa: I wonder if we should call this `onlyRestorableItems` instead to keep in line with the `rewind` -> `restore` rebrand?

WordPress/Classes/Stores/ActivityStore.swift
```diff
@@ -236,9 +238,16 @@ private extension ActivityStore {
...
+                guard let self = self else {
+                    return
+                }
+
                 let loadingMore = offset > 0
-                actionDispatcher.dispatch(ActivityAction.receiveActivities(site: site, activities: activities, hasMore: hasMore, loadingMore: loadingMore))
+                actionDispatcher.dispatch(ActivityAction.receiveActivities(
+                                            site: site,
+                                            activities: self.onlyRewindableItems ? activities.filter { $0.isRewindable } : activities,
+                                            hasMore: hasMore, loadingMore: loadingMore))
```
momo-ozawa: Alignment for this method seems to be off 🙏

momo-ozawa:
@leandroalonso I noticed that we should be using the site's timezone for the date filter calendar. It seems like the date filter calendar is set to use your device timezone.
  
I'm testing a site where the site's timezone doesn't match my device timezone, so the date date filtering results come back differently on the app vs WordPress.com.

WordPress/Classes/Stores/ActivityStore.swift
leandroalonso: Agree, good catch! Changed in 21cc34bbef8b495a06abe72f18f36b73e9428a89

WordPress/Classes/Stores/ActivityStore.swift
leandroalonso: Done in a6345392498dcc192b2bc99a45d41356bf61a15b

leandroalonso:
@momo-ozawa I tested a site here using your timezone (Asia/Tokyo), both in the app and in WordPress.com. I selected 1-7 Jan and yet there is a Dec 31 in the results in both platforms.

I did an investigation and the issue I'm finding is that I'm not able to correctly retrieve the TimeZone for the blog, thus there's no way to adjust the date formatting:

```
(lldb) po blog.settings?.timezoneString
▿ Optional<String>
  - some : ""
```

In order to fix that I wonder that perhaps we should call service and then update this value... which is a little bit complex 😬 (not to take into account that TimeZones are quite a rabbit hole).

Plus, I'm seeing this banner on the web version:



I was talking to @emilylaguna that perhaps we can leave that outside of this MVP. The assumption is that not that many users will be in such different timezones. What do you think?

Also, please let me know if you have any idea on how to fix that with an easier approach.

leandroalonso:
@momo-ozawa Just an update on the above, the timezone was empty because no timezone was specified. 😅

But even with this change, I'm still seeing 31 Dec results (on the web and in the app — plus, applying the site's timezone in the app logic):

momo-ozawa (approved):
@leandroalonso Works as described!

Thanks for investigating the timezone issue!!! 🙏 
Strange that Dec 31 shows up in results, but I don't think it's a blocker.

PR #15591, opened 2021-01-06
Hide Xpost UI on unsupported sites

### LINKED ISSUE #2942: Xposts UI shown momentarily on sites that don't support Xpost suggestions
**Describe the bug**
There are a number of scenarios where Xpost suggestion UI is shown on sites that don't support Xposts (e.g. non-P2 sites):
- After typing `+`, the suggestion UI flashes for a moment while the network request loads (first reported in https://github.com/wordpress-mobile/WordPress-iOS/pull/15139#discussion_r548485770)
- The editor `@` toolbar button incorrectly allows the long-press gesture which allows users to insert an Xpost suggestion
- When switching from a site that supports Xposts to a different one (one which doesn't support Xposts), the Xpost UI is not dismissed

**To Reproduce**
Open the block editor on a site that does not support suggestions (e.g. a Simple WP.com site)
  1. Type `+`and notice the suggestion UI flash before it disappears
  2. Long-press the `@` toolbar button and notice that the Xpost option is there (even though the site doesn't support it)
  3. Go to a site that _does_ support Xposts, then switch to a site that doesn't support it – notice the Xpost UI is not dismissed

**Expected behavior**
We prefetch the suggestions list when the editor is opened (see https://github.com/wordpress-mobile/WordPress-iOS/pull/15562), so we should be able to hide all Xpost UI if the site doesn't support it.

**Smartphone (please complete the following information):**
 - Device: iPhone 11
 - OS: iOS 14.2
 - Version WPiOS `develop` branch

**Additional context**
First reported by @antonis in https://github.com/wordpress-mobile/WordPress-iOS/pull/15139#discussion_r548485770.

guarani:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/2942

Previously, sites that don't support Xposts (e.g. Simple WordPress.com sites) showed the suggestions UI for a few seconds while the network loaded when the user typed a `+` character. This PR fixes this by disabling the suggestions UI as soon as the Xpost prefetch request returns a result from the network that indicates that Xposts are not available.

If after the app determines that a site has Xposts disabled, the site then becomes Xpost-enabled remotely, the app will be able to pick up this change only once the app is restarted (that is, when the `SiteSuggestionService` singleton is cleared from memory).

### To test

**What you'll need**: One WP.com site that is capable of xposting (https://wordpress.com/p2/ can be used to create xpost-capable sites) and one WP.com site that is **not** capable of xposting.

**Xpost in Gutenberg**

1. Open the block editor on the site capable of Xposting
2. In a Paragraph or any other rich text block, type `+` and verify that the list of Xpost suggestions is shown
3. Switch to the site that is not xpost capable
4. Type `+` and ensure that no suggestion UI is shown
5. Verify that long-pressing the `@` toolbar button does not show Xpost options (expect the long-press gesture to be disabled)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/SiteSuggestionService.swift
+++ b/WordPress/Classes/Services/SiteSuggestionService.swift
@@ -5,7 +5,7 @@ class SiteSuggestionService {
 
     private var blogsCurrentlyBeingRequested = [NSNumber]()
     private var requests = [NSNumber: Date]()
-    var disabledBlogs = [NSNumber]()
+    private var disabledBlogs = [NSNumber]()
 
     static let shared = SiteSuggestionService()
 
--- a/WordPress/Classes/Services/SiteSuggestionService.swift
+++ b/WordPress/Classes/Services/SiteSuggestionService.swift
@@ -5,6 +5,7 @@ class SiteSuggestionService {
 
     private var blogsCurrentlyBeingRequested = [NSNumber]()
     private var requests = [NSNumber: Date]()
+    var disabledBlogs = [NSNumber]()
 
     static let shared = SiteSuggestionService()
 
@@ -38,10 +39,12 @@ class SiteSuggestionService {
      If no suggestions already in Core Data, fetch them from the network and store them in Core Data.
      @param blog The blog/site to prefetch suggestions for
      */
-    func prefetchSuggestions(for blog: Blog) {
+    func prefetchSuggestions(for blog: Blog, completion: @escaping () -> Void) {
         let persistedSuggestions = retrievePersistedSuggestions(for: blog)
         if persistedSuggestions == nil || persistedSuggestions?.isEmpty == true {
-            fetchAndPersistSuggestions(for: blog, completion: { _ in})
+            fetchAndPersistSuggestions(for: blog, completion: { _ in
+                completion()
+            })
         }
     }
 
@@ -86,13 +89,17 @@ class SiteSuggestionService {
 
             self.requests[blogId] = Date()
 
+            self.disabledBlogs.removeAll { $0 == blogId }
             completion(suggestions)
 
             // remove blog from the currently being requested list
             self.blogsCurrentlyBeingRequested.removeAll { $0 == blogId }
-        }, failure: { [weak self] error, _ in
+        }, failure: { [weak self] error, response in
             guard let `self` = self else { return }
 
+            if response?.statusCode == 400 {    // blog/site does not have Xposts available
+                self.disabledBlogs.append(blogId)
+            }
             completion([])
 
             // remove blog from the currently being requested list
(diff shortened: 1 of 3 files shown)


illusaen:
Out of curiousity, is this an iOS only bug?

WordPress/Classes/Services/SiteSuggestionService.swift
```diff
@@ -5,6 +5,7 @@ class SiteSuggestionService {
 
     private var blogsCurrentlyBeingRequested = [NSNumber]()
     private var requests = [NSNumber: Date]()
+    var disabledBlogs = [NSNumber]()
```
illusaen: Why not private here?

WordPress/Classes/ViewRelated/Post/PostEditor+BlogPicker.swift
```diff
@@ -28,10 +28,6 @@ extension PostEditor where Self: UIViewController {
             }
             self.recreatePostRevision(in: blog)
             self.mediaLibraryDataSource = MediaLibraryPickerDataSource(post: self.post)
-
```
illusaen, line 31: Out of curiousity, if you kept this here but added your changes, would it be able to also check if the new blog had xpost suggestions and thus show them if it did?

WordPress/Classes/Services/SiteSuggestionService.swift
guarani: Good catch! I missed that. Thanks!

WordPress/Classes/ViewRelated/Post/PostEditor+BlogPicker.swift
guarani: I'd move these lines to [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15591/files#diff-beacfffc3f8a87edf94895da980a81e356f1b5aa7998f70791bc7baec9a11731R436-R439) where they will get triggered both when a user switches blogs from within the editor as well as when the user navigates to the editor (e.g. from the My Site screen).

So as far as I know, the move ensures that Xposts are always prefetched. Does that answer your question?

guarani:
> Out of curiousity, is this an iOS only bug?

Yes, it's only iOS.

WordPress/Classes/ViewRelated/Post/PostEditor+BlogPicker.swift
illusaen: Interesting, it does! That's why you only had the one call vs the original two in different places. Thanks!

illusaen (approved):

WordPress/Classes/ViewRelated/Post/PostEditor+BlogPicker.swift
guarani: Thanks for checking 👍

PR #15594, opened 2021-01-06
Reader filters: retain selected filter

### LINKED ISSUE #15568: Reader filters: retain selected filter
As requested by @mattmiklic : the filter should be kept in place if you just switch tabs, or switch to a different part of the app and come back. But it shouldn't be so persistent as to remain after restarting the app.

Ref https://github.com/wordpress-mobile/WordPress-iOS/pull/15484

ScoutHarris:
Fixes #15568 

When a Reader stream is filtered, the selected filter is now saved.
- The filter is saved for each stream that can be filtered.
- The filter is restored when changing Reader tabs.
- The filters persist while in the app.

To test:
- Go to the Reader.
- Apply filters to streams that can be filtered.
- Change Reader tabs.
  - Verify the selected filters persist for each stream.
  - Verify the stream content is still filtered.
- Clear a filter.
  - Verify the filter is cleared and remains cleared when switching Reader tabs.
  - Verify the stream content is no longer filtered.
- Muck about in other areas of the app.
- Return to the Reader.
  - Verify applied filters persist.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -13,6 +13,7 @@ class ReaderTabView: UIView {
 
     private let viewModel: ReaderTabViewModel
 
+    private var filteredTabs: [(index: Int, topic: ReaderAbstractTopic)] = []
 
     init(viewModel: ReaderTabViewModel) {
         mainStackView = UIStackView()
@@ -199,6 +200,13 @@ extension ReaderTabView {
                   let self = self else {
                 return
             }
+
+            let selectedIndex = self.tabBar.selectedIndex
+
+            // Remove any filters for selected index, then add new filter to array.
+            self.filteredTabs.removeAll(where: { $0.index == selectedIndex })
+            self.filteredTabs.append((index: selectedIndex, topic: selectedTopic))
+
             self.resetFilterButton.isHidden = false
             self.setFilterButtonTitle(selectedTopic.title)
         }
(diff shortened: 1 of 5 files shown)


ScoutHarris:
Sorry @aerych . Please don't review this yet. Looks like I broke some tests.

ScoutHarris:
OK @aerych . The overlords have been appeased. Proceed at will.

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -13,6 +13,7 @@ class ReaderTabView: UIView {
 
     private let viewModel: ReaderTabViewModel
 
+    private var filteredTabs: [(index: Int, topic: ReaderAbstractTopic)] = []
```
aerych, line 16: I'm wondering if there's ever a situation where the topic entity could be deleted, while a reference to it is still stored here.  I tried to test this in the app by following/unfollowing a site and by login out and back in, both while the filter was set.  There were no issues as far as I could tell (yay!) but this is till a question nagging me. 

For paranoia sake, would it be safer to store the objectID rather than the entity or am I worrying too much?

aerych (approved):
Heya @ScoutHarris 👋 
Looking good! I tested as described and everything seems to work.  I had one nitpicky / paranoid question about the code but otherwise 👍. 
Lemme know what you think.  Good to :shipit: if you think its fine. :)

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
ScoutHarris: I can't think of reason to do so. Since you stated the cases you thought of, the only other one I came up with was deleting a site. So I filtered by a site then deleted the site. The filter did remain and showed content, so there were no adverse effects. The site did disappear from the site filter list, so after I cleared the filter I wasn't able to see the site anymore. As far as I can tell it works fine as is.

Tangentially, one thing I did notice is that if you filter by a site or tag, then go to Manage on the nav bar and unfollow the site/tag, the filter is not cleared. I'll see what I can do about that.

aerych: Nice catch!

PR #15583, opened 2021-01-05
Feature/ios13 min build version

aerych:
See pbArwn-1ua-p2 for the background on this change.

This PR sets the app's minimum deployment target to iOS 13.0, and acts as the base branch for fixing deprecation warnings.

To test:
Smoke test the app and make sure there are no obvious regressions.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15599, opened 2021-01-07
iOS 13: remove keyWindow deprecation warnings

leandroalonso:
Refs #15583.

This PR fixes the `keyWindow` deprecation warnings.

```
'keyWindow' was deprecated in iOS 13.0: Should not be used for applications that support multiple scenes as it returns a key window across all connected scenes
```

A little bit of context:

`keyWindow` was deprecated because it's not safe for apps that support multiple scenes. AFAIK, we don't use this (see https://developer.apple.com/documentation/uikit/app_and_environment/scenes/supporting_multiple_windows_on_ipad#3663342).

So I just went ahead and added an extension that contains a `mainWindow`, which works just like `keyWindow`.

### To test

Open the app and use parts that make use of `mainWindow` like: Reader, trigger a Notice, etc.

Please, test using an iPad too.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/UIApplication+mainWindow.swift
+++ b/WordPress/Classes/Extensions/UIApplication+mainWindow.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+extension UIApplication {
+    @objc var mainWindow: UIWindow? {
+        return UIApplication.shared.windows.filter {$0.isKeyWindow}.first
+    }
+}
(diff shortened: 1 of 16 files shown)


WordPress/Classes/Extensions/UIApplication+mainWindow.swift
```diff
@@ -0,0 +1,7 @@
+import UIKit
+
+extension UIApplication {
+    @objc var mainWindow: UIWindow? {
+        return UIApplication.shared.windows.filter {$0.isKeyWindow}.first
```
aerych, line 5: Nice :)

aerych (approved):
Looks good to me.  Tested on an iPhone and iPad simulator. 
:shipit: !

PR #15607, opened 2021-01-07
Reverts: Allow updatable statusbars on each tab bar view

### LINKED ISSUE #15506: Status bar icons invisible on some screens
### Expected behavior

The status bar icons should always be visible.

### Actual behavior

The status bar icons are invisible in some parts of the app (e.g. when tapping on a notification for a post that I'm subscribed to).

1. Go to the Notifications tab
2. Tap on a Notification for a post on a site you're subscribed to
3. Notice the lack of status bar icons

### Additional information

This is similar to https://github.com/wordpress-mobile/WordPress-iOS/issues/15458 but covers any part of the app including Notifications screens, whereas that issue was solely focused on Reader screens.

emilylaguna:
I've reverted this change that fixed #15506

### To test:
In order to test in the simulator you need to change line 47 in `FeatureFlag.swift` to `false` to disable the white nav bar appearance.
```
        case .newNavBarAppearance:
            return false
```

1. Launch app
2. Tap on the My Site tab
3. Verify the status bar icons are white
4. Tap on Reader, and verify the status bar icons are white
5. Tap on Notifications, and verify the status bar icons are white
6. Verify the original issue exists (Invisible status bar icons) by navigating to the Reader Detail in the Notifications tab

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/WPTabBarController+ReaderTabNavigation.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/WPTabBarController+ReaderTabNavigation.swift
@@ -10,12 +10,13 @@ extension WPTabBarController {
     }
 
     override open var childForStatusBarStyle: UIViewController? {
-        var selectedController = selectedViewController
-        if let navController = selectedController as? UINavigationController {
-            selectedController = navController.topViewController
+        guard
+            let topViewController = readerNavigationController?.topViewController,
+            ((topViewController as? DefinesVariableStatusBarStyle) != nil)
+        else {
+            return nil
         }
-
-        return selectedController
+        return topViewController
     }
 }
 


aerych:
Thanks for the patch :)  I tested and it works as described. 
Just my two cents here, but if we're trading one glitch for another, it seems like being able to see the status bar text an icons would be preferable.  Maybe we can live with the reader's status bar being a different color for one release and follow up with a permanent solution in the next one?  Just a thought.

aerych (approved):

emilylaguna:
@aerych Heya, I thought about that but I feel that the notifications reader detail issue is a lot less likely to occur for the majority of users while the status bar icons color will happen for everyone who uses the reader. 

The notifications issue has also existed in the app for a number of versions at this point and hasn't generated any support traffic as far as I can tell. 

That being said, I feel more comfortable living with the existing issue than introducing a new higher visibility one.

Thoughts?

aerych:
Honestly, I'm fine either way.  No strong opinion here. :)

emilylaguna:
👍 For tracking I've reopened the [original issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15506) and I'm planning on looking into it again next sprint.

PR #15609, opened 2021-01-07
Reader filters: clear filter when topic is unfollowed

ScoutHarris:
Ref: #15568

When a site or tag is unfollowed via the Reader Manage view:
- Any corresponding saved filter is removed.
- If the current stream is filtered by the site/tag, the stream is reloaded.

To test:
- Go the Reader.
- Filter a stream by a site/tag.
- Go to the Manage view (gear icon on the nav bar).
- Unfollow the same site/tag.
- Verify the filter bar is reset.
- Verify the stream is no longer filtered.

no_more_cats

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -1,5 +1,12 @@
 import UIKit
 
+// NSNotification sent when a site or a tag is unfollowed via Reader Manage screen.
+extension NSNotification.Name {
+    static let ReaderTopicUnfollowed = NSNotification.Name(rawValue: "ReaderTopicUnfollowed")
+}
+let topicUserInfoKey = "topic"
+
+
 class ReaderTabView: UIView {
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -1,5 +1,12 @@
 import UIKit
 
+// NSNotification sent when a site or a tag is unfollowed via Reader Manage screen.
```
ScoutHarris, line 3: FWIW, I did try to use delegates. But it got messy due to all the classes that are involved just to display the tags and sites view controllers. This ended up being much cleaner/less convoluted.

aerych: I really like this approach :)  Totes agree delegates be more complicated.

aerych (approved):
Hola @ScoutHarris 👋 
Works great!  :shipit: !

ScoutHarris:
Thanks @aerych !

PR #15622, opened 2021-01-11
Merge 16.4 final into develop

jkmassel:
Merges the last beta fix, release notes, an update to the copyright year (and fastlane), and some release version shenanigans caused by an incomplete release finalization.


Scripts/fastlane/Deliverfile
```diff
@@ -11,7 +11,7 @@ privacy_url({
   'default' => 'https://automattic.com/privacy/',
 })
 
-copyright('2020 Automattic Inc.')
+copyright('2021 Automattic Inc.')
```
AliSoftware, line 14: 👍 
Made me realize that for Woo I updated the value in ADC but didn't think about updating it in Deliverfile

config/Version.public.xcconfig
```diff
@@ -1,4 +1,4 @@
 VERSION_SHORT=16.4
 
 // Public long version example: VERSION_LONG=9.9.0.0
-VERSION_LONG=16.4.0.2
+VERSION_LONG=16.4.0.5
```
AliSoftware, line 4: Does this mean we forgot to merge after betas .3, .4 and .5? pfew!

Scripts/fastlane/Deliverfile
jkmassel: Oh weird – `fastlane deliver` yelled at me in the `precheck` stage about this, which was how I figured out that it was in the `deliverfile` at all 😄 

Personally I'd like to get rid of the deliver file and just put these in .txt files instead but 🤷‍♂️

AliSoftware (approved):
LGTM! :shipit:

Scripts/fastlane/Deliverfile
AliSoftware: Yeah `fastlane deliver` also yelled at me for it which is what prompted me asking about it to Legal; but at that point I thought it was pulling the data from ADC and putting it into the `copyright.txt` file and that's where the warning was coming from, which is why I updated the info in ADC directly, thinking it was pulled from the `.txt`.

I didn't know we also mentioned it in the `Deliverfile`, and I wonder if it's still useful to keep it given that we also have it in the `copyright.txt` file anyway? 🤔

config/Version.public.xcconfig
jkmassel: Haha no – the release finalization failed the first time (burning `.3`) because of the duplicate milestone issue you'd mentioned (I didn't think this would cause it to fail, but guess I was wrong!)

So I fixed that and ran it again, and it failed because the frozen tag wasn't applied to the milestone anymore (burning `.4`). 

So I fixed _that_ and it ran almost all the way though (using `.5`), then finished it from there by hand by closing up the milestone.

Ideally I think this process would be idempotent, but this whole process only took 5 minutes, so it may not be worth the time to change it 😄

AliSoftware: Ah makes sense, thanks for the explanation!

Scripts/fastlane/Deliverfile
jkmassel: Yeah for WPiOS it's specified in the `deliverfile` which is fine, but given that all our other metadata is stored in the fastlane-specified `.txt` files IMHO it'd be nice to have that be consistent.

Might open a PR to that effect later today

PR #15624, opened 2021-01-11
Update PageListViewController deletion UI to match PostListViewController

frosty:
Refs p2-pb3aDo-Ff#comment-1176. This PR updates the logic of the `PageListViewController` page to match that of the posts list:

* Deletion / Trash options in the ... menu of posts now uses the destructive style
* Tapping one of these options shows a confirmation alert before the action takes place
* I also removed the Publish Immediately option from the ... menu on the Trashed posts list, as it seemed odd to have it right next to the 'delete permanently' option, and it's not present on the posts list.

trash-options

I considered pulling out a component that encapsulates this logic, or moving it to the AbstractPostListViewController to save on duplication (I took the method straight from PostListViewController) – however, I opted just to duplicated in this instance as a) It's only used in these two places, and b) The text in the UI needs to distinguish between Posts and Pages, so this seemed the simplest way.

**To test:**

* Build and run
* Navigate to the Pages screen for a site, and in either Published, Draft, or Scheduled, check that the option to Trash is in red and shows you a warning if you tap it.
* Check that if you cancel the warning nothing happens, and if you confirm it, the post is trashed.
* In the Trashed posts list, check that the Delete Permanently option in the ... menu is read, and the same logic as above applies to the confirmation alert.

PR submission checklist:


WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
```diff
@@ -874,6 +865,39 @@ class PageListViewController: AbstractPostListViewController, UIViewControllerRe
...
+        let messageText: String
+        let titleText: String
+
+        if post.status == .trash {
+            cancelText = NSLocalizedString("Cancel", comment: "Cancels an Action")
+            deleteText = NSLocalizedString("Delete Permanently", comment: "Delete option in the confirmation alert when deleting a page from the trash.")
+            titleText = NSLocalizedString("Delete Permanently?", comment: "Title of the confirmation alert when deleting a page from the trash.")
+            messageText = NSLocalizedString("Are you sure you want to permanently delete this page?", comment: "Message of the confirmation alert when deleting a page from the trash.")
+        } else {
+            cancelText = NSLocalizedString("Cancel", comment: "Cancels an Action")
```
Gio2018: nit: `cancelText` does not need to be repeated

Gio2018 (approved):
This works great!! Just one very minor comment that you can totally ignore.

:shipit:!

frosty:
@iamthomasbishop I'm merging this one, but do let me know if you spot anything that needs changing :)

iamthomasbishop:
Thanks, @frosty! The labels look good. 

One other thing, and I'm not sure if this contradicts what I had mentioned previously, but after looking at this ticket I'm curious if it would make more sense to show the confirmation dialog after the "delete permanently" action is taken instead of after the user moves the post/page to trash?

In other words, we would keep the existing behavior or trashing (where we show a message that the post/page has been moved to trash), then if the user attempts to delete permanently from the Trashed screen, *at that point* show a confirmation dialog.

If that's out of scope for this ticket, feel free to merge and we can discuss the confirmation placement separately. 😀

frosty:
Thanks for the suggestion, @iamthomasbishop – I can definitely remove the confirmation from the trash step, and just leave it for deleting permanently.

PR #15625, opened 2021-01-11
Backups screen: add strings and events

leandroalonso:
Part of https://github.com/wordpress-mobile/WordPress-iOS/issues/15193

This PR:

* Adds two new VCs: `BackupListViewController` and `JetpackActivityLogViewController` that inherits from `BaseActivityListViewController`
* Adds the correct events and strings for each different screen

#### Events

1. Open the Backups screen
2. Check that `backup_list_opened` is fired
3. Tap the date range filter button and check that `backup_filterbar_range_button_tapped` is fired
4. Select a date or a range, tap "Done" and check that `backup_filterbar_select_range` is fired
5. Tap "X" in the date ranger filter button and make sure `backup_filterbar_reset_range` is fired

Test the same flow above in Activity Log, but the events should be prepended by an `activitylog_` instead of `backup_`

#### Strings

1. Open the Backups screen for a site without backups
2. You should see a text saying "Loading Backups..." and then "Your first backup will be ready soon" (this screen would only be visible to users with the Backup capability in the feature)
3. Select a date range and tap Done
4. You should see a "No matching backups found." message

PR submission checklist:


momo-ozawa (approved):
@leandroalonso Works as described! 👍

PR #15631, opened 2021-01-13
Update metadata scripts + existing metadata

jkmassel:
Downloads all of the latest metadata from App Store Connect and updates WPiOS scripts to update App Store Connect for two new locales: `es-MX` and `ar-SA` in a similar manner as https://github.com/woocommerce/woocommerce-ios/pull/3352.

**To Test:**
- Run `cd Scripts/fastlane && ./download-metadata.swift`. Note that the release notes have been deleted (because they're not present for `16.5` yet).
- Change the `glotPressWhatsNewKey` to `v16.4-whats-new`. Note that far fewer release notes are missing (just the ones that weren't part of `16.4`.


Scripts/fastlane/Deliverfile
```diff
@@ -7,12 +7,6 @@ app_identifier "org.wordpress"
 # Make sure to update these keys for a new version
 app_version "16.5"
 
-privacy_url({
-  'default' => 'https://automattic.com/privacy/',
-})
-
-copyright('2021 Automattic Inc.')
```
AliSoftware, line 14: So does that mean you confirm we can remove that from our `Deliverfile` in all our repos now that we have the `copyright.txt` in `metadata`? That's great, avoids having 2 sources of truth! 🎉

.gitignore
```diff
@@ -77,6 +77,7 @@ Scripts/Preview.html
 Scripts/fastlane/test_output/
 Scripts/fastlane/google_cloud_keys.json
 Scripts/fastlane/promo_screenshots
+/Scripts/fastlane/metadata/review_information
```
AliSoftware, line 80: 👍 🔒

Scripts/fastlane/download_metadata.swift
```diff
@@ -106,21 +108,19 @@ func downloadTranslation(languageCode: String, folderName: String) {
...
-        /// This sort of hack-ey – we're not actually downloading anything here, but the idea is
-        /// to keep all of the metadata generation code in the same place. By keeping this in the
-        /// same file, it should be easier to find, and update if necessary.
-        let marketingURL = "https://apps.wordpress.com/mobile/"
-
         do {
+
+            let releaseNotesPath = "\(languageFolder)/release_notes.txt"
+
+            /// Remove existing release notes in case they weren't translated for this release (that way `deliver` will fall back to the `default` locale)
```
AliSoftware, line 115: Is the `default` locale/folder created by `fastlane deliver` automatically? If so, I'm a bit lost as to why we created it by copying the content of `en_US` [in the other PR](https://github.com/woocommerce/woocommerce-ios/pull/3352/files/41e1543f718e8f25c1f8291538c48cdcdaa48644#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R587) 🤔

AliSoftware (commented):
I'm surprised to discover we don't have a `download_app_store_translations` lane for `WordPress-iOS` like we do for `woocommerce-ios`, why is that? 🤔 

As a result, I can see that NONE of the metadata is actually translated in any language. Neither the privacy policies, nor the app subtitle… This is kinda worrying if that's the state of our metadata L10n on AppStore?!

Scripts/fastlane/Deliverfile
jkmassel: Yep that's right – we should move this to `.txt` files instead of the `Deliverfile` in all repos (which we're now doing in WCiOS too!)

Scripts/fastlane/download_metadata.swift
jkmassel: > I'm a bit lost as to why we created it by copying the content of en_US in the other PR

Yeah good question – the screenshots are a separate glotpress project and aren't downloaded by `download_metadata.swift`, so these processes work separately.

Realistically, it'd make a lot of sense to symlink `default` to `en-US`, for sure. Though in the past we've avoided symlinks in repos. Though I could see a good argument for relaxing this restriction for iOS repos since there'd be no reason to check them out on Windows.

> Is the default locale/folder created by fastlane deliver automatically
I don't think so, because `default` could be any locale – it's not guaranteed to be an `en` variant (if the developer is using ZH, as a for-instance).

AliSoftware: > I don't think so, because default could be any locale – it's not guaranteed to be an en variant (if the developer is using ZH, as a for-instance).

Yeah but I thought that when _downloading_ metadata from ASC (which you did to resync the metadata and make the repo become the new source of truth) `deliver` would create the `default` folder based on the default locale it saw selected for the app in ASC…

AliSoftware (approved):
Overall looks good to me.

_At first it seemed to me that this change failed to synchronize the `fastlane/metadata/default`  folder with `en-US` like the other PRs do, but then I saw that this was solved by having `"default": "en-us"` in the `languages` map after all (which means we download the metadata for `en-us` twice, in practice, but I don't think that's a big deal at all.)_

Side note that we might want ensure we use a similar logic (on how we handle `default`) in other app repos for consistency (esp resync what you did on the woo PR with those changes here). Though at some point, that part might fall into the release-toolkit refactor project (See https://github.com/wordpress-mobile/release-toolkit/issues/195).

_I'd love your help on this since you seem to be quite at ease with how our various scripts handle the various metadata with your recent PRs_ 😄

Issue #15355, opened 2020-11-20
Reader: Mark posts as seen / unseen


mindgraffiti:
As a user on mobile, I would like the app to mark a post as seen when I visit it in Reader. I would also like to mark a post as seen / unseen at will, in the "more" menu.

Calypso alignment: Do not release until 16.4 or later.

Tasks:
- [x] Add ability to mark posts seen/unseen. [wordpress-mobile/WordPressKit-iOS/pull/327, wordpress-mobile/WordPressKit-iOS/pull/336, wordpress-mobile/WordPressKit-iOS/pull/341]
- [x] Add seen/unseen menu option.
  - [x] Post details. [#15638, #15718, #15745]
  - [x] Reader card. [#15645, #15718, #15745]
- [x] ~~Only [posts from followed feeds](https://github.com/Automattic/wp-calypso/pull/49028) should have the mark as seen option. [#15670]~~ (negated by #15745)
- [x] Mark post as seen when viewed in the Reader. [#15676, #15718]
- [x] Add Tracks. [#15671]
- [x] Error handling. [#15681]
- [x] Enable `unseenPosts` feature flag. [#15753]

---
Example from the web:

<img width="667" alt="Screen Shot 2020-11-20 at 9 03 16 AM" src="https://user-images.githubusercontent.com/1062444/99815358-c8e6ba00-2b0f-11eb-8a9a-97a1af6e83cd.png">

More menu in Reader > post:

<a href="https://user-images.githubusercontent.com/1062444/99820186-08b0a000-2b16-11eb-990e-da56a412ed90.PNG"><img src="https://user-images.githubusercontent.com/1062444/99820186-08b0a000-2b16-11eb-990e-da56a412ed90.PNG" width="350" /></a>

Design:

<img width="1237" alt="Screen Shot 2021-01-12 at 3 40 50 PM" src="https://user-images.githubusercontent.com/1816888/104383173-a5773480-54ec-11eb-9dbc-b37a9fdd5c24.png">

ScoutHarris:
From Slack convo: Mark Seen/Unseen -> works as described in [this](https://wp.me/pbArwn-ZE-p2#comment-1910) comment.

PR #15638, opened 2021-01-13
Reader post details: add ability to mark as seen/unseen.

ScoutHarris:
Ref #15355 
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/327

This adds the ability to mark a post seen/unseen from post details. (The same functionality will be added to the Reader cards in a follow-up PR.)

  - The `isSeen` property is now parsed and added to `ReaderPost`.
  - That property is used to determine which option to display in the menu (mark as seen/ mark as unseen).
  - When the menu option is selected, the corresponding endpoint is called to update the post's seen status.

To test:
- Go to the Reader.
- Tap on a post.
- In a browser, view the same post in the Reader.
- In the app, tap the menu button on the nav bar.
- On the menu, verify the seen/unseen option matches that in the Reader.

| app_mark_seen | web_mark_seen |
|--------|-------|

- Select the menu option.
- Redisplay the menu. Verify the option is now the opposite.
- Refresh the post in the browser. Verify the option is now the opposite.
  - NOTE: Sometimes it take a minute/a few refreshes for the status to be updated.

| app_mark_unseen | web_mark_unseen |
|--------|-------|

PR submission checklist:


WordPress/Classes/Models/ReaderPost.h
```diff
@@ -34,6 +34,7 @@ extern NSString * const ReaderPostStoredCommentTextKey;
 @property (nonatomic) BOOL isReblogged;
 @property (nonatomic) BOOL isWPCom;
 @property (nonatomic) BOOL isSavedForLater;
+@property (nonatomic) BOOL isSeen;
```
aerych, line 37: Maybe add a new entry in MIGRATIONS.txt about the new field?

WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>WordPress 110.xcdatamodel</string>
+	<string>WordPress 111.xcdatamodel</string>
```
aerych, line 6: Screen Shot 2021-01-14 at 3 11 03 PM
What do you think about making this a scalar value, providing a default, and making it non-optional?

WordPress/Classes/ViewRelated/Reader/ReaderPostMenu.swift
```diff
@@ -48,19 +50,27 @@ open class ReaderPostMenu {
...
+        alertController.addActionWithTitle(post.isFollowing ? ReaderPostMenuButtonTitles.unfollow : ReaderPostMenuButtonTitles.follow,
             style: .default,
             handler: { (action: UIAlertAction) in
                 if let post: ReaderPost = self.existingObject(for: post.objectID, context: post.managedObjectContext) {
                     self.toggleFollowingForPost(post, viewController)
                 }
         })
 
+        // Seen
+        alertController.addActionWithTitle(post.isSeen ? ReaderPostMenuButtonTitles.markUnseen : ReaderPostMenuButtonTitles.markSeen,
```
aerych: Should we put this behind the unread counts feature flag so we can flip a switch and turn both features on at once?

aerych (changes requested):
Hola @ScoutHarris 👋 
Looking good!  I had a few questions/suggestions. 
Lemme know what you think!

WordPress/Classes/ViewRelated/Reader/ReaderPostMenu.swift
ScoutHarris: We could. I went back and forth, but ultimately didn't because I figured the ability to toggle seen could be useful standalone and released without the count. If you prefer otherwise, I can do it in my next PR (which I'm working on now to add this functionality to reader cards.)

WordPress/Classes/Models/ReaderPost.h
ScoutHarris: Ugh, I totally forgot. Thank you!

WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
ScoutHarris: I thought about that, specifically providing a default value. But I didn't want to potentially have a conflict with Calypso. How do we know if a default of seen or unseen is correct?

WordPress/Classes/ViewRelated/Reader/ReaderPostMenu.swift
ScoutHarris: Nevermind. I'll go ahead & do it.

WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
aerych: I suppose we don't, but I'm not sure if it matters?  We would want to set it to whatever value was returned when syncing a post.  Have a scalar, non-optional, default is more of a data integrity thing--avoiding nil values that make implicitly unwrapped optionals go :boom:   So, in the cases of a faulted and deleted ReaderPost, if for some reason the object was still being held in memory, and the `isSeen` field was inspected, it would return a boolean rather than nil. 

It's also entirely possible I'm being overly paranoid about it.  CoreData has broken me.

ScoutHarris:
Hey @aerych . Ready for another go please. Thanks!

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -435,6 +437,63 @@ - (void)toggleSavedForLaterForPost:(ReaderPost *)post
...
+                    }
+                }];
+            }];
+        };
+        
+        // Call the remote service.
+        ReaderPostServiceRemote *remoteService = [[ReaderPostServiceRemote alloc] initWithWordPressComRestApi:[self apiForRequest]];
+        [remoteService markPostSeenWithSeen:seen
+                                     feedID:readerPost.feedID
+                                 feedItemID:readerPost.feedItemID
```
aerych, line 492: I was doing some more testing and I'm seeing something new. The value of feedItemID is turning up nil for the sites I've tested.  I haven't dug in to figure out why but it's happened on a private site of mine, an internal p2, and a public blog.   
Just dropping a quick note.  I'll take another look first thing tomorrow.

aerych: I think what might have happened is I updated my local branch after bd38f6f but before 73dc7be and 25d4f77 were pushed up.  I've updated and tried a clean build and now things work as expected. :)

aerych (approved):
Looks good!  I tested as described and everything worked.
Two final questions:
- Will the feature flag be added in a follow up? 
- Would it make sense to add test coverage for the `toggleSceneForPost` service call? Maybe also in a follow up? 

Otherwise :shipit: :)

ScoutHarris:
Hey @aerych .

> Will the feature flag be added in a follow up?

I added the `FeatureFlag.unseenPostCount.enabled` check when adding the menu option. (https://github.com/wordpress-mobile/WordPress-iOS/pull/15638/commits/070ef0525c7a510587e6b7022443ae5ebaf2d3ed)

> Would it make sense to add test coverage for the toggleSceneForPost service call? Maybe also in a follow up?

Yeppa. Adding tests is on my TODO list once I get everything working.

Thanks for your help and patience. 😄

WordPress/Classes/Services/ReaderPostService.m
ScoutHarris: Sorry! That's what I get for assuming you wouldn't re-review so soon! 😄

PR #15645, opened 2021-01-14
Reader post card: add ability to mark as seen/unseen

ScoutHarris:
Ref #15355 

This adds the ability to mark a post seen/unseen from a Reader post card.

To test:
- Go to the Reader.
- In a browser, view the same Reader stream.
- In the app, tap the menu button on a card.
- On the menu, verify the seen/unseen option matches that in the Reader for the same post.

| app_mark_seen | web_mark_seen |
|--------|-------|

- Select the menu option.
- Redisplay the menu. Verify the option is now the opposite.
- Refresh the stream in the browser. Verify the option is now the opposite.

| app_mark_unseen | web_mark_unseen |
|--------|-------|

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderSeenAction.swift
```diff
@@ -0,0 +1,7 @@
+/// Encapsulates a command to toggle a post's seen status
+final class ReaderSeenAction {
+    func execute(with post: ReaderPost, context: NSManagedObjectContext, completion: (() -> Void)? = nil) {
```
aerych: This method accepts a completion block, that can be nil, but then the block is never called?

WordPress/Classes/ViewRelated/Reader/ReaderSeenAction.swift
```diff
@@ -0,0 +1,7 @@
...
+        let postService = ReaderPostService(managedObjectContext: context)
+        postService.toggleSeen(for: post, success: nil, failure: nil)
```
aerych: Would it make sense to define a failure block to notify the user in the case of failure?

aerych (changes requested):
Howdy @ScoutHarris 👋 
I just had a couple of questions on this one. 
Back to you!

WordPress/Classes/ViewRelated/Reader/ReaderSeenAction.swift
ScoutHarris: 🤦 You know when you look at a dumb mistake and you flash back to exactly the moment you made it and know exactly what you did? Yea, that.

WordPress/Classes/ViewRelated/Reader/ReaderSeenAction.swift
ScoutHarris: Yea, I'll add a failure block.

Regarding `notify the user in the case of failure?` - it looks like the various failures in this menu are not consistent. At least one shows an alert, some others silently fail. I really didn't want to show an alert and disrupt the user just for this. I'm going to look at failures here a bit more and see if they all can show _something_ .

ScoutHarris:
OK @aerych give it another go please kind sir.

aerych (approved):
Lookin' good!
:shipit:

ScoutHarris:
Thanks @aerych !

PR #15652, opened 2021-01-15
Jetpack: adds logic for when to display Activity Log filter + small fixes

leandroalonso:
Part of #15193

* Displays the Activity Log filters only if the site has a paid plan OR has the backup capability
* Remove dots from some strings
* Fixes an event property
* Changes the number of items per page for Backups to 100

### To test

Review the strings and the event change. :)

Then:

1. Run the app
2. Tap Activity Log for a free site
3. Check that the filters do not appear
4. Open Activity Log in a paid or Jetpack site with backup
5. Check that the filters do appear
6. Tap Backup
7. Check that more items are loaded initially

PR submission checklist:


WordPress/Classes/Models/JetpackSiteRef.swift
```diff
@@ -31,4 +37,8 @@ struct JetpackSiteRef: Hashable, Codable {
         return lhs.siteID == rhs.siteID
             && lhs.username == rhs.username
```
momo-ozawa, line 38: WDYT of adding checks for `hasBackup` and `hasPaidPlan` for `==(lhs: rhs:)` just to be safe?

WordPress/Classes/ViewRelated/Activity/ActivityListViewModel.swift
```diff
@@ -18,7 +18,7 @@ class ActivityListViewModel: Observable {
     private let noResultsTexts: ActivityListConfiguration
     private var storeReceipt: Receipt?
 
-    private let count = 20
+    var numberOfItemsPerPage = 20
```
momo-ozawa: Better name, thanks! 👍

momo-ozawa (commented):
Works as described 👍

WordPress/Classes/Models/JetpackSiteRef.swift
leandroalonso: Sure, good catch!

Done in 03b314dd274da2303f0c40d8379f7e6c418ab4bf

momo-ozawa (approved):

PR #15659, opened 2021-01-18
Pages List: Add notice after page parent is set

enejb:
This PR add a notice that shows up after a user changes the post parent.

## To test:
1. Go to pages.
2. From the 3 dot menu select "Set parent"
3. Choose a parent. 
4. Notice the pill notice at the bottom.

Screen Shot 2021-01-18 at 9 38 15 AM

5. From the 3 dot menu select "Set parent"
6. Choose "Cancel"
7. Notice that the pill doesn't show up at the bottom.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
+++ b/WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
@@ -758,8 +758,15 @@ class PageListViewController: AbstractPostListViewController, UIViewControllerRe
         let parentPageNavigationController = ParentPageSettingsViewController.navigationController(with: pages, selectedPage: selectedPage) {
             self._tableViewHandler.isSearching = false
             self._tableViewHandler.refreshTableView(at: index)
+            self.handleSetParentSuccess()
         }
         present(parentPageNavigationController, animated: true)
+    }
+
+    private func handleSetParentSuccess() {
+        let setParentSuccefullyNotice =  NSLocalizedString("Page Parent updated.", comment: "Message informing the user that their pages parent has been set successfully")
+        let notice = Notice(title: setParentSuccefullyNotice, feedbackType: .success)
+        ActionDispatcher.global.dispatch(NoticeAction.post(notice))
     }
 
     fileprivate func pageForObjectID(_ objectID: NSManagedObjectID) -> Page? {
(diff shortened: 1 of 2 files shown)


enejb:
I am not sure this is the correct way to do this. 
One thing that I notice is that the notice shows up even after the user selects "Cancel" which to me is unexpected. 

But I wasn't sure how to fix that. 

This PR is more of a way for me to learn, so please consider this a low priority item but if you have any pointers how to best proceed with this I would be happy to try to implement it. :)

kyleaparker:
I would suggest making a minor wording update so that this notice is consistent with the others: 

**Parent page successfully updated**

WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
```diff
@@ -755,13 +755,19 @@ class PageListViewController: AbstractPostListViewController, UIViewControllerRe
         let selectedPage = pageAtIndexPath(index)
         let newIndex = _tableViewHandler.index(for: selectedPage)
         let pages = _tableViewHandler.removePage(from: newIndex)
-        let parentPageNavigationController = ParentPageSettingsViewController.navigationController(with: pages, selectedPage: selectedPage) {
+        let parentPageNavigationController = ParentPageSettingsViewController.navigationController(with: pages, selectedPage: selectedPage, onClose: {
             self._tableViewHandler.isSearching = false
             self._tableViewHandler.refreshTableView(at: index)
-        }
+        }, onSuccess: { self.handleSetParentSuccess() } )
```
antonis: I would suggest to use a [weak reference](https://www.avanderlee.com/swift/weak-self/) here and reformat the code like the following:
```
        let parentPageNavigationController = ParentPageSettingsViewController.navigationController(with: pages, selectedPage: selectedPage, onClose: { [weak self] in
            self?._tableViewHandler.isSearching = false
            self?._tableViewHandler.refreshTableView(at: index)
        }, onSuccess: { [weak self] in
            self?.handleSetParentSuccess()
        })
```

antonis (approved):
Great work @enejb 👍 
I tested the app on an iPhone SE with iOS 14.2 and it works as expected 🎉 
I just left a [minor comment/suggestion](https://github.com/wordpress-mobile/WordPress-iOS/pull/15659/files#r562508561).

WordPress/Classes/ViewRelated/Pages/PageListViewController.swift
enejb: Thanks for the info about weak reference. ❤️

PR #15660, opened 2021-01-18
Reader: add a source property when following a topic

### LINKED ISSUE #15636: Reader: track topic followed in Discover section
### Expected behavior

1. Given the user followed a topic/interest in the Discover section
2. Track the topic/interest that was followed

### Additional information

As for now, we have the `reader_reader_tag_followed` being fired. However, it is fired from multiple places, so there's no easy way to check if it came up from the Discover section.

We can either:

* Add a property `source` in `reader_reader_tag_followed`
* Create a new event and fire it only in Discover

leandroalonso:
Fixes #15636

#### Following from Discover

1. Remove all the followed topics you might have
2. Go to Reader -> Discover
3. Select a few topics and tap "Done"
4. Check that `reader_reader_tag_followed` is fired with a `source: discover` property

#### Following from somewhere else

1. Tap the cog icon (Reader navigation bar, on the right side)
2. Add a topic
3. Check that  `reader_reader_tag_followed` is fired with a `source: unknown` property

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
@@ -1471,6 +1471,7 @@ NSString *const TracksUserDefaultsLoggedInUserIDKey = @"TracksLoggedInUserID";
             break;
         case WPAnalyticsStatReaderTagFollowed:
             eventName = @"reader_reader_tag_followed";
+            eventProperties = @{ @"source" : @"unknown" };
             break;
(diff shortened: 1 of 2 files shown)


emilylaguna (approved):
✅ Followed topics on the Discover view and confirmed the source is correct:
`🔵 Tracked: reader_reader_tag_followed <source: discover, tag: activism>`

✅ Followed topics using the Manage section and confirmed the source is unknown:
`🔵 Tracked: reader_reader_tag_followed <source: unknown, subscription_count: 41, tag: hello>`

PR #15668, opened 2021-01-19
Reader: Reset content when switching from Discover to filtered stream

### LINKED ISSUE #15649: Reader filters cause Discover tab to take over Following tab
### Expected behavior
Reader filters should persist while in the app

### Actual behavior
If you filter on the Following tab, then change to the Discover tab, the Discover tab persists on the Following tab. 

https://cloudup.com/c4UJjpDpdSu

1. In Reader, filter for a certain site
2. Note that the filter works correctly
3. Switch to the Discover tab and then back to the Following tab
4. Notice that the Following tab has the filter you applied, plus the Discover tab content

##### Tested on iPhone 12 Pro, iOS 14.3, WPiOS 16.5.0.0

ScoutHarris:
Fixes #15649

When switching from the Discover tab to a filtered stream, the stream content is now correctly restored.

To test:

---
- Go to a Reader tab that can be filtered, and select a filter.
- Switch to the Discover tab, and back to the filtered tab.
- Verify:
  - The content matches the filter.
  - The stream header does not appear.

| filtered | discover |
|--------|-------|

---
- Tap a site's card header.
- Verify the stream header does appear.

| card_header | site_header |
|--------|-------|

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -21,6 +21,11 @@ class ReaderTabView: UIView {
     private let viewModel: ReaderTabViewModel
 
     private var filteredTabs: [(index: Int, topic: ReaderAbstractTopic)] = []
+    private var previouslySelectedIndex: Int = 0
+
+    private var discoverIndex: Int? {
+        return tabBar.items.firstIndex(where: { $0.title == NSLocalizedString("Discover", comment: "Discover tab name") })
+    }
 
     init(viewModel: ReaderTabViewModel) {
         mainStackView = UIStackView()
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -21,6 +21,11 @@ class ReaderTabView: UIView {
     private let viewModel: ReaderTabViewModel
 
     private var filteredTabs: [(index: Int, topic: ReaderAbstractTopic)] = []
+    private var previouslySelectedIndex: Int = 0
+
+    private var discoverIndex: Int? {
+        return tabBar.items.firstIndex(where: { $0.title == NSLocalizedString("Discover", comment: "Discover tab name") })
```
aerych, line 27: The string comparison is something we were doing before but it worries me a little.  Translations could differ between the web and the apps.  I don't really have a better suggestion tho.

aerych (approved):
Howdy @ScoutHarris 👋 
Looks good to me! 
:shipit: !

WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
ScoutHarris: Yea, I just copied it from the original implementation. It made me uncomfortable as well, but it seems to work. Since this is a patch, I didn't want to make _too_ many changes. I've added it to my list to look at later.

ScoutHarris:
Thanks @aerych !

PR #15669, opened 2021-01-19
Generalize Today Widget to Stats Widgets

Gio2018:
This refactor will allow to use the existing extension of Today Widgets for all iOS 14 Stats Widgets.

The most important thing in this review (other than the testing that nothing is broken) is to look at the code as a whole to determine if the overall design needs improvements; the main goal being to host all three stats widgets in one extension.

Fixes #NA

To test:
The main goal of the testing is to make sure this did not break anything for the existing Today Widget.
- build/run and make sure you can install today widget and it works as expected.

PR submission checklist:


WordPress/WordPressStatsWidgets/Remote service/StatsWidgetsService.swift
```diff
@@ -0,0 +1,95 @@
+import WordPressKit
+
+/// Type that wraps the backend request for new stats
+class StatsWidgetsService {
```
diegoreymendez, line 4: In order to avoid one of the mistakes we made with WPiOS initially, I wonder if we could create a local Swift package to store the services, instead of including them directly in the Widget target.

I realize that right now it means having to do a bit more work, but I see a few advantages in doing so:

1. It forces decoupling the Widget UI logic from the services and model logic in a very healthy way (and avoid repeating the mistakes we did in WPiOS).
2. If we ever want to separate the logic, it's usually much easier to do so at the beginning, than once the codebase has grown (again, thinking about WPiOS today).
3. It's easier to share the service and model with other targets, if we ever need.
4. Kind of the same as point 3: A future migration to newly designed widgets would be much easier.

WordPress/WordPressStatsWidgets/Remote service/StatsWidgetsService.swift
```diff
@@ -0,0 +1,95 @@
...
+        do {
+            let token = try SFHFKeychainUtils.getPasswordForUsername(WPStatsTodayWidgetKeychainTokenKey,
+                                                                     andServiceName: WPStatsTodayWidgetKeychainServiceName,
+                                                                     accessGroup: WPAppKeychainAccessGroup)
+
+            let wpApi = WordPressComRestApi(oAuthToken: token)
+            let service = StatsServiceRemoteV2(wordPressComRestApi: wpApi, siteID: widgetData.siteID, siteTimezone: widgetData.timeZone)
+
+            // handle fetching depending on concrete type
+            // we need to do slike this as there is no unique service call
```
diegoreymendez: Typo "slike"

diegoreymendez (approved):
Works great for me.  Just added one comment that could entail more changes, but since my comment is open to discussion and not really in the scope of this PR, I'm approving your changes.

WordPress/WordPressStatsWidgets/Remote service/StatsWidgetsService.swift
Gio2018: Great suggestion @diegoreymendez and I agree 100%!!

frosty (approved):
This works for me! Tested both widget types on various sites, and all looks good 👍

PR #15670, opened 2021-01-19
Reader > mark as seen: Show menu option only for posts that are followed

ScoutHarris:
Ref #15355 , https://github.com/Automattic/wp-calypso/pull/49028

To test:
- Go to the Reader.
- Search for a Post or Site that you are not following.
  - Verify the card and post details options do not include the seen option.
- Search for a Post or Site that you are following.
  - Verify the card and post details options include the seen option.
- Filter `Followed` by a topic.
  - Verify the card and post details options do not include the seen option.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15676, opened 2021-01-20
Reader post details: mark post as seen when viewed

ScoutHarris:
Ref #15355 

This does two things:
1. When a post is viewed from a reader stream, it is automatically marked `seen`.
2. When a post's seen status changes from either of these locations, the post is updated in the stream so the card's menu has the correct seen option.
  - From viewing the post (point 1 above).
  - From manually toggling the seen status from post detail's options menu.

I also centralized the reader's NSNotification definitions into `ReaderHelpers`, hence the seemingly unrelated changes to the `ReaderTopicUnfollowed` notification.

To test:

---
Automatically marking posts as `seen`:
- In the Reader, tap on a post.
- Tap on the menu button in the nav bar.
  - Verify the seen option is `Mark as unseen` (since the post was automatically marked as `seen`).
- Don't select anything. Close the post details & return to the reader stream.
- Tap the menu button on the card.
  - Verify the seen option is `Mark as unseen`.

| Post details | Stream after post viewed  |
|--------|-------|
| post_details | stream |

---
Manually toggling seen:
- In the Reader, tap on a post.
- Tap on the menu button in the nav bar.
- Select `Mark as unseen`.
- Close the post details & return to the reader stream.
- Tap the menu button on the card.
  - Verify the seen option is `Mark as seen`.

Feel free to toggle seen in post details to whatever you like. The point here is that the card's menu reflects the final state in post details.

| Post details | Stream after toggling |
|--------|-------|
| post_details2 | stream2 |

---
PR submission checklist:


ScoutHarris:
Hey @aerych .

I [added](https://github.com/wordpress-mobile/WordPress-iOS/pull/15676/commits/0591f4f4cac276e786cb7860d6bad86bcc39cdbb) a check for `feedItemID` to `ReaderPostService`, which stops the crashing when attempting to automatically mark unfollowed posts as seen. With https://github.com/wordpress-mobile/WordPress-iOS/pull/15670, all the crashing should be fixed.

Ready for another go please!

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
```diff
@@ -241,6 +241,26 @@ class ReaderDetailCoordinator {
...
+    }
+
+    private func markPostAsSeen() {
+        guard let post = post,
+              let context = post.managedObjectContext else {
+            return
+        }
+
+        // Ensure post is unseen before toggling it so it end's up as seen.
+        post.isSeen = false
```
aerych: If the post is already marked as seen should we return early rather than making another call to mark it seen?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -1168,6 +1170,23 @@ import WordPressFlux
...
+
+        guard let userInfo = notification.userInfo,
+              let post = userInfo[ReaderNotificationKeys.post] as? ReaderPost,
+              let indexPath = content.indexPath(forObject: post),
+              let cellPost: ReaderPost = content.object(at: indexPath) else {
+            return
+        }
+
+        cellPost.isSeen = post.isSeen
+        tableView.reloadRows(at: [indexPath], with: UITableView.RowAnimation.fade)
```
aerych: Could omit `UITableView.RowAnimation` for brevity.

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -441,6 +441,18 @@ - (void)toggleSeenForPost:(ReaderPost *)post
                   success:(void (^)(void))success
                   failure:(void (^)(NSError *error))failure;
 {
+    if (post.feedItemID == nil) {
+        NSString *description = NSLocalizedString(@"Could not toggle Seen: missing feedItemID attribute.", @"An error description explaining that Seen could not be toggled due to a missing feedItemID attribute.");
+        NSDictionary *userInfo = @{ NSLocalizedDescriptionKey : description };
+        NSError *error = [NSError errorWithDomain:ReaderPostServiceErrorDomain code:0 userInfo:userInfo];
```
aerych, line 448: It might be nice to add a `DDLogError` here to capture this in the logs?

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -441,6 +441,18 @@ - (void)toggleSeenForPost:(ReaderPost *)post
                   success:(void (^)(void))success
                   failure:(void (^)(NSError *error))failure;
 {
+    if (post.feedItemID == nil) {
```
aerych, line 444: Could be a `guard` but meh. I'm not a stickler about such things ;)

aerych (commented):
Howdy @ScoutHarris 👋 
Looking good! I just had a few questions.  Lemme know what you think.

WordPress/Classes/Services/ReaderPostService.m
ScoutHarris: Oh if only ObjC had `guard`...

ScoutHarris:
OK @aerych one more time 🤞 !

WordPress/Classes/Services/ReaderPostService.m
aerych: 🤦

aerych (approved):
:shipit: !

ScoutHarris:
Thanks @aerych !

PR #15678, opened 2021-01-20
Update/page layout api query paramaters

enejb:
This PR adds 'type=mobile and is_beta=true to the query parameters of the page layout api calls. 

This make it so that the api will be able to return type ( mobile or web ) specific page layouts as well as beta versions.

To test:
Do the tests pass. 

D55810-code implements the API side of things to make it work as expected. 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/PageLayoutService.swift
+++ b/WordPress/Classes/Services/PageLayoutService.swift
@@ -74,13 +74,8 @@ class PageLayoutService {
 
     private static let type = "mobile"
 
-    private static let isBeta: String = {
-        let isDevMode = BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
-        if isDevMode == true {
-            return "true"
-        }
-        return "false"
-    }()
+    // Return "true" or "false" for isBeta that gets passed into the endpoint.
+    private static let isBeta = String(BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest])
 
 }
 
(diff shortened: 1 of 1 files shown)


WordPress/Classes/Services/PageLayoutService.swift
```diff
@@ -66,6 +71,17 @@ class PageLayoutService {
...
+
+    private static let type = "mobile"
+
+    private static let isBeta: String = {
+        let isDevMode = BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+        if isDevMode == true {
+            return "true"
+        }
+        return "false"
+    }()
```
antonis: Since the `~=` [does not return an optional](https://github.com/wordpress-mobile/WordPress-iOS/blob/50e1647c52304eef6b84fd330729d03be9796040/WordPress/Classes/Utility/BuildInformation/BuildConfiguration.swift#L26) I would simplify this like:
```
private static let isBeta = String(BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest])
```
This might be opinionated and your solution might be better since it is easier to see the actual values passed.

antonis (approved):
Nice work @enejb 👍 
The change works as expected and the code LGTM 🎉 
I just left a minor suggestion.

ps. Thank you for this additions in the API. I think they are quite useful 🙇

WordPress/Classes/Services/PageLayoutService.swift
enejb: Great idea. Thanks for the suggestion to simplify it.

chipsnyder (approved):
LGTM! :shipit:

PR #15681, opened 2021-01-21
Reader: Show message when toggling seen fails.

ScoutHarris:
Ref #15355 

This shows a "toast" message when toggling seen fails from:
- Reader post card options menu.
- Post details options menu.

To test:

**Reader card:**
- Go the Reader.
- Disable network.
- Tap the options button on a post card, and toggle the seen status.
- Verify a toast message is displayed with the correct message (`Unable to mark post seen` / `Unable to mark post unseen`).

**Post details:**
- Tap on a post card in the Reader.
- Disable network.
- Tap the options button in the post details, and toggle the seen status.
- Verify a toast message is displayed with the correct message (`Unable to mark post seen` / `Unable to mark post unseen`).

| seen_fail | unseen_fail |
|--------|-------|

PR submission checklist:


ScoutHarris:
Hey @mattmiklic .

I just wanted to get your input here. As you can see in the description, when selecting `Mark as seen` / `Mark as unseen` for a post, I chose to display a toast message if it fails. I did this instead of an alert as I didn't think it was worth blocking the user over. In addition, the other menu options are inconsistent. Most don't show anything on failure. Following/unfollowing a site does show an alert:

follow_fail


So, two questions.
1. Is showing a toast message acceptable?
2. If so, I would like to implement that for the other fail cases where applicable (in a follow-up PR). Including following/unfollowing a site, which would nix the alert.

Thoughts?

aerych (approved):
Hola @ScoutHarris 👋 
Code looks good!
I ran into some odd behavior while testing that I'm pretty sure is unrelated but I'm curious if you can reproduce.   
Basically, on the post list I tried to repeatedly mark a post seen/unseen while offline to spam notices.  Then I'd tap to view the reader detail while the notice(s) was visible.  Then I'd try to mark the reader detail seen/unseen.    One time I locked up the simulator with a bunch of calls to tableView.reloadData (not sure what that was about). One time the notices just stopped appearing (got out of sync somehow?).   
I'd guess it's probably something more to do with how the dispatched notice actions are handled.  Curious what you think. 
Anyhow, I think this is fine to :shipit: and we can figure out the above later.

ScoutHarris:
> I ran into some odd behavior while testing that I'm pretty sure is unrelated but I'm curious if you can reproduce.
Basically, on the post list I tried to repeatedly mark a post seen/unseen while offline to spam notices. Then I'd tap to view the reader detail while the notice(s) was visible. Then I'd try to mark the reader detail seen/unseen. One time I locked up the simulator with a bunch of calls to tableView.reloadData (not sure what that was about). One time the notices just stopped appearing (got out of sync somehow?).

Hrm. I tried it on a simulator and an actual iPhone. I spam toggled seen several times in both the Reader and post details. I didn't see either problem. 

Simulator: iPhone 12 Pro Max (14.3)
Device: iPhone 11 Pro Max (14.0)

aerych:
>  I didn't see either problem.

I'll blame the ghost in the machine. Thanks for checking!

ScoutHarris:
Hey @mattmiklic .

I'm going to go ahead and merge this. If you have any thoughts on my [questions](https://github.com/wordpress-mobile/WordPress-iOS/pull/15681#issuecomment-764069798) please let me know. Thanks!

mattmiklic (commented):
Sorry for the wait @ScoutHarris; I think a toast message is a good way to handle these error states, and I agree that it's more appropriate than a blocking modal. I'd be in favor of converting any other blocking alerts with this same solution.

PR #15687, opened 2021-01-21
Show SignIn as root VC.

### LINKED ISSUE #15399: Tab bar briefly visible when app starts
### Expected

When opening the app after a clean install, the splash/launch screen should be followed by a smooth animation to the logged out welcome screen.

### Actual

When first opening the app after a clean install, the tab bar is briefly shown before the app shows the welcome screen.

The same issue is present in WooCommerce for iOS (https://github.com/woocommerce/woocommerce-ios/issues/3107), and there is existing discussion there. A solution might be found in the Authenticator shared library so separate fixed won't be needed for each app.

### Steps to reproduce

1. Perform a clean install or log out of the app and kill it (swipe up from the app switcher)
2. Open the app and notice that after the splash screen, a tab bar is shown for a split second before the welcome screen is shown

|   |
| - |

Tested on WPiOS 16.1.1, iOS 14.2

diegoreymendez:
This PR changes our logic to show the sign in flow as our root VCs.

This also fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/15399 (thanks @jaclync for pointing this out!)

## The short explanation

This is an example of what this PR resolves:

startupGlitch

## The long explanation

While working in [this PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281) I noticed the App's UI was sometimes showing up for a moment before the login UI was visible.  Something similar to what @jaclync reported [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/564#pullrequestreview-573716784).  My fix is actually (kind of) similar to what Jaclyn is pushed [here](https://github.com/woocommerce/woocommerce-ios/issues/3429).

Additionally, [in this PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281), I'm having a bit of trouble with some view controllers that are being updated behind the login UI while there's no user.  This is a scenario that we've discussed in the past with @frosty.

This PR is probably not enough to kill the full UI while logged out, since WPiOS uses a shared `WPTabBarController` which is never really deleted.  But it's a very strong (IMHO) first step towards simplifying out UI updates while there's no user signed in.

In short this aligns the lifecycle of the UI with the authentication state of the user.

## Animation Changes:

While I didn't intend to change anything animation-wise, I found that the previous transitions we used for swapping between full-screen sign in and the regular App UI require some custom code with the changes I'm introducing in this PR.  We can implement them as they were but they'll require some extra coding.

In weighting this, I also noticed the new animation looks kinda good IMHO - but I'm not sure if it's just me so I'm posting a demo to get direction on it from our designers.

@mbshakti can I ask you to let me know if it'd be acceptable to use this transition, or if you want me to roll back to the regular slide-down animation we previously had?

newSignInTransition

## To test:

1. Test logging in from scratch:
   - [ ] Try the standard case where you just log in and dismiss the login epilogue directly.
   - [ ] Try the variant where in the login epilogue you add another site by address.
   - [ ] Try the variant where the account you use to log in has no sites - this should show a different epilogue.
2. Test signing up:
   - [ ] Include the standard case in which you just sign up and dismiss the epilogue.
   - [ ] Include the variant in which you create a site in the signup epilogue.
   - [ ] Include the variant in which you add site by address in the signup epilogue.
3. Test adding a self hosted site in the "My Sites" tab, when you're already logged in.
   - [ ] Test you're able to add a self hosted site just fine.
   - [ ] When using an iPad the login UI should be shown "inline" instead of full screen.
4. Test disconnecting the WPiOS session from your WP.com Security Settings to see how the App reacts to it.
   - [ ] Test you're able to reconnect your session just fine.
   - [ ] When using an iPad the login UI should be shown "inline" instead of full screen.

Make sure you test everything in both iPad and iPhone.

## PR submission checklist:


jaclync:
Hi @diegoreymendez! This PR probably also fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/15399?

WordPress/Classes/System/WindowManager.swift
```diff
@@ -0,0 +1,86 @@
...
+
+        show(WPTabBarController.sharedInstance())
+    }
+
+    /// Shows the initial UI for unauthenticated users.
+    ///
+    func showUIForUnauthenticatedUsers() {
+        isSignInVisible = true
+
+        guard let loginViewController = WordPressAuthenticator.loginUI() else {
```
jaclync, line 69: we might also want to track `.openedLogin` like in the previous `WordPressAuthenticator.showLogin`?

diegoreymendez: Thanks!

diegoreymendez:
Note: this PR is referencing a commit hash until the pod version is updated - I'll prolly update that tomorrow once the release is out.

diegoreymendez:
> Hi @diegoreymendez! This PR probably also fixes #15399?

Most definitely @jaclync !

WordPress/Classes/System/WindowManager.swift
diegoreymendez: I've added the tracking call.

I was thinking though that maybe this tracking call should be in the VC's `viewWillAppear` method in the Authenticator pod?  I think it could be a less  "fragile" approach in sight of future changes.

Definitely not for the scope of this PR though, but let me know what you think.

WordPress/Classes/System/WindowManager.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    // MARK: - Initial App UI
+
+    /// Shows the initial UI for the App to be shown right after launch.  This method will present the sign-in flow if the user is not
+    /// authenticated, or the actuall App UI if the user is already authenticated.
+    ///
+    func showUI() {
+        if AccountHelper.isLoggedIn {
+            showUIForAuthenticatedUsers()
```
Gio2018: nit: these two names are very similar could create some confusion; can we call them differently, like `shouLoggedInUI` and `showLoggedOutUI` or `showAppTabs` and `showLoginScreen` or.. ?

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -18,6 +18,11 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
 
     var window: UIWindow?
 
+    @objc
+    lazy var windowManager = {
+        WindowManager(window: window!)
```
Gio2018: clearly if this window is not available the crash is unavoidable.. but maybe we can add an error message if this happens? Something like

``
lazy var windowManager: WindowManager = {
        guard let window = window else {
            fatalError("message here")
        }
        return WindowManager(window: window)
    }()
``

Gio2018 (commented):
Hey @diegoreymendez , this looks really good!

Had just a couple of minor comments, and, as discussed on slack, it seems that the login screen is still full screen in the case of self hosted login.

WordPress/Classes/System/WindowManager.swift
jaclync: Moving the tracking call to the login VC's `viewWillAppear` sounds like a good idea! I can make this change after code freeze for the authenticator library later this week 👍 just created an issue: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/567

diegoreymendez: Thanks!

diegoreymendez:
I've changed the milestone to 16.7 as I'd rather this be tested a bit more.

WordPress/Classes/System/WindowManager.swift
diegoreymendez: Changed!

WordPress/Classes/System/WordPressAppDelegate.swift
diegoreymendez: Changed.

diegoreymendez:
@Gio2018 - I've addressed the code comments.  I've also checked in `develop`, and the self-hosted UI is also shown full screen.  It looks the same.

The PR is ready for another look.

Gio2018 (approved):
Thanks for this @diegoreymendez !! Looks good!! 

:shipit:

diegoreymendez:
@mbshakti - I've restored the original transition, so there's no need to review :)

PR #15683, opened 2021-01-21
Jetpack Section: Activity Detail Download Backup Button

momo-ozawa:
Part of #15191

### Description:
- Added `Download backup` button to Activity Detail VC for rewindable activities
- Added button action for `Download file` button in Restore Complete VC

### Merge instructions:
- ⚠️ Merge after https://github.com/wordpress-mobile/WordPress-iOS/pull/15673

### To test:
1. My Site > Activity Log
2. Tap on a rewindable cell to view Activity Detail
    - ✅ `Restore` and `Download backup` buttons
3. Tap on `Download backup` button
    - ✅ Should navigate to Backup Options VC 
4. Tap on `Create downloadable file` and wait until your backup completes
5. Tap on `Download file`
    - ✅ Should open Safari with download prompt

Feature Flag OFF | Feature Flag ON | Download file -> Safari
--- | --- | ---
IMG_1136 | IMG_1137 | IMG_1140

### PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15689, opened 2021-01-22
Jetpack Section: Backup / Restore Already Running

momo-ozawa:
Part of #15191

### Description:
- Added error handling for "restore already running" / "backup already running"
- Updated styling for `RestoreStatusView`

### Merge instructions:
- ⚠️ Merge after https://github.com/wordpress-mobile/WordPress-iOS/pull/15683

#### Restore
1. My Site > Activity Log > rewindable item
2. Tap `Restore`
3. Tap `Restore to this point`
4. Tap `Confirm`
    - ✅ Progress value should be displayed on the left side
    - ✅ Progress title should be displayed on the right side
5. Tap `OK, notify me!`
    - ✅ Should return to the activity log
    - ✅`RewindStatusRow` should display the progress
5. Start another restore by repeating steps 2~4
    - ✅ `There's a restore currently in progress, please wait before starting the next one` should be displayed

#### Backup
Since the create downloadable backup process is so quick, I recommend testing this process by using both the app and Calypso.

1. [App] My Site > Activity Log > rewindable item
2. [App] Tap `Download backup`
3. [Web] Go to wordpress.com > Activity log > rewindable item
4. [Web] Kick off a download backup process
4. [App] Before the backup on wordpress.com finishes, tap `Create downloadable file`
    - ✅ `There's a backup currently being prepared, please wait before starting the next one` should be displayed
 

Updated styling | Restore already running | Backup already running 
--- | --- | ---
IMG_1148 | Simulator Screen Shot - iPhone 12 Pro - 2021-01-25 at 20 06 44 | Simulator Screen Shot - iPhone 12 Pro - 2021-01-25 at 20 09 30

### PR submission checklist:


leandroalonso:
@momo-ozawa I tested the described steps, but I find it a little bit odd that:

* When starting from the web I wasn't able to see the progress (even after refreshing the list)
* When dismissing the screen, I also wasn't able to see the progress (only if I tapped on the "Notify me later" button)

I'd like to propose a change that perhaps might fix that:

* If you call `store.fetchRewindStatus(site: site)` from `ActivityListViewModel.refresh` both the aforementioned cases would be fixed
* It still doesn't cover the case when you return from a backup/restore, but there are a few ways you can fix that: by using a completion block and/or delegate you can trigger the `fetchRewindStatus` and voilá!
* This will also remove the need to inject the store into the coordinators/VCs, removing the coupling and simplifying the code

Let me know what do you think. :)

momo-ozawa:
@leandroalonso 

> If you call store.fetchRewindStatus(site: site) from ActivityListViewModel.refresh both the aforementioned cases would be fixed

Fixed! 98e59cb

> It still doesn't cover the case when you return from a backup/restore, but there are a few ways you can fix that: by using a completion block and/or delegate you can trigger the fetchRewindStatus and voilá!

> This will also remove the need to inject the store into the coordinators/VCs, removing the coupling and simplifying the code

Yeah a delegate feels better, thanks for the suggestion! a152ede

momo-ozawa:
### Backup / Restore FYI

How to tell if there’s a backup / restore already in progress:
- When there’s a restore already in progress, and you send a second POST restore request, the `job_id` in the response is equal to `0`
- When there’s a backup already in progress, and you send a second POST prepare backup request, the `rewindID` you sent in the request doesn’t match the `rewindID` returned in the response. The response `rewindID` matches the `rewindID` of the backup that's already in progress.

You can test this on Calypso by having two windows open.

leandroalonso (approved):
It all works as described!

I'm still with this issue tho:

> When dismissing the screen, I also wasn't able to see the progress (only if I tapped on the "Notify me later" button)

I don't think it's a blocker, we can address it in a subsequent task.

PR #15692, opened 2021-01-22
iOS 14 Today Widget - Refresh site list

Gio2018:
This PR refreshes the site list when `syncBlogs` in `BlogListViewController` is called.

Note that this requires `BlogListViewController` to appear on screen. Another approach could be to call `BlogService.syncBlogsForAccount` every time some information about a site is updated (e.g. creation, title change etc), but considering that this operations would not occur so frequently, and they would not cover 100% of the cases anyway (e.g. add a site from another platform), I chose to avoid additional write calls to CoreData for now. If this proves to be a problem we can always add these behaviors.

Fixes #NA

To test:

- Install Today Widget and visualize existing stats in it
- Add a site in the app, then navigate back to the site list
- Go back to the widget, long press then edit widget, make sure the site appears in the available sites list
- Go back to the app and delete a site, this should redirect you to the sites list
- Go back to the widget, long press then edit widget and make sure the site you just deleted is not available in the list anymore. Note: if you delete a site that was selected in the widget, the widget should fall back to the default site after deletion

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Today Widgets/iOS 14 Widgets/BlogListViewController+StatsWidgets.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Today Widgets/iOS 14 Widgets/BlogListViewController+StatsWidgets.swift
@@ -10,14 +10,14 @@ extension BlogListViewController {
         let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
         let updatedSiteList = blogService.visibleBlogsForWPComAccounts()
 
-        let newData = updatedSiteList.reduce(into: [Int: HomeWidgetTodayData]()) { result, element in
-            guard let blogID = element.dotComID else {
+        let newData = updatedSiteList.reduce(into: [Int: HomeWidgetTodayData]()) { sitesList, site in
+            guard let blogID = site.dotComID else {
                 return
             }
             let existingSite = currentData[blogID.intValue]
 
-            let siteURL = element.url ?? existingSite?.url ?? ""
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Stats/Today Widgets/iOS 14 Widgets/BlogListViewController+StatsWidgets.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    @objc func refreshStatsWidgetsSiteList() {
+
+        guard let currentData = HomeWidgetTodayData.read() else {
+            return
+        }
+        let blogService = BlogService(managedObjectContext: ContextManager.shared.mainContext)
+        let updatedSiteList = blogService.visibleBlogsForWPComAccounts()
+
+        let newData = updatedSiteList.reduce(into: [Int: HomeWidgetTodayData]()) { result, element in
```
diegoreymendez: This is very nitpicky, as the code is good as it is.  I was just thinking that replacing `element` with something like `site` could make the code slightly more readable.

Obviously, feel free to disregard if you disagree or simply prefer the current term.

diegoreymendez (approved):
Works as described.  Nice job!

WordPress/Classes/ViewRelated/Stats/Today Widgets/iOS 14 Widgets/BlogListViewController+StatsWidgets.swift
Gio2018: totally agree, changed!!

PR #15699, opened 2021-01-23
Merge 16.5 final into develop

jkmassel:
Merges beta changes, translation merge, and 16.5 release note translations into `develop`


mokagio (approved):
Looks good 👍 

I was surprised by the size of this PR. The Czech translations (`cs`) a responsible for 2,000+ lines, as it looks like they were previously not translated, so that explains it.

PR #15706, opened 2021-01-26
Move Fastlane to Project Root

jkmassel:
Aligns this project with others to put the `fastlane` directory at the root of the project.

**To Test:**
Note that the following CI jobs are passing for this PR:
- Build Tests
- Unit Tests
- UI Tests (both)
- Installable Build
- Translation Build (note that the build in appetize.io is the one associated with the CI run for this PR)

**To Review:**
This is (necessarily) a _huge_ PR because it moves every file in the `fastlane` directory. I'd recommend reviewing this by commit – the first just moves all of the files into the new location – subsequent ones alter them to make them compatible with the new location.

---

It's possible that there might be other breaking changes I haven't encountered while testing, but I'd like to land this in the release branch (then in `develop`) as soon as possible in order to avoid conflicts with other changes – I can fix other broken tooling if/when I run into it – this should be the vast majority of it.

--

Reviews: I'd like review from both @AliSoftware and @loremattei for this – @loremattei because he's most familiar with all of it and @AliSoftware because I streamlined a bunch of the pathing and variables and I'd love your input on the changes.

Thank you!!


.gitignore
```diff
@@ -70,17 +70,20 @@ WordPress/Images.xcassets/AppIcon-Internal.appiconset
...
+Preview.html
+fastlane/test_output/
+fastlane/google_cloud_keys.json
+fastlane/promo_screenshots
+fastlane/metadata/review_information
+default.profraw
 derived-data/
 
+# CI Artifacts Location
+/Artifacts
```
AliSoftware:
```suggestion
Artifacts
```

And actually similar for other existing entries below like `/WordPress/…` lines and such

Scripts/extract-framework-translations.swift
```diff
@@ -3,22 +3,30 @@
 import Foundation
 
 let fileManager = FileManager.default
-let cwd = fileManager.currentDirectoryPath
-let script = CommandLine.arguments[0]
 
-let base = cwd
-let projectDir = base.appending("/WordPress")
+let projectRoot = findProjectRoot()
+let projectDir = projectRoot.appending("/WordPress")
```
AliSoftware: Can't we use `Foundation`'s `appendingPathComponent` here (to let it append the `/` as needed or not for us etc)? Or does that require an explicit cast to `NSString` (which would make it uglier)?

Scripts/update-translations.rb
```diff
@@ -124,27 +119,44 @@
...
+  # Step 2 – Download the new strings
+  if File.exist? destination
+    FileUtils.copy destination, backup_destination
+  end
 
   url = "#{download_url}/#{code}/default/export-translations?#{strings_filter}format=strings"
-  destination = "#{lang_dir}/Localizable#{strings_file_ext}.strings"
 
-  system "curl -fLso #{destination} \"#{url}\"" or begin
+  system "curl -fgLo #{destination} \"#{url}\"" or begin
```
AliSoftware: Why this change?

Scripts/update-translations.rb
```diff
@@ -124,27 +119,44 @@
...
+    puts "\e[31mFatal Error: #{destination} appears to be empty. Exiting.\e[0m"
+    abort()
   end
 
-  system "./Scripts/fix-translation #{lang_dir}/Localizable#{strings_file_ext}.strings"
-  system "plutil -lint #{lang_dir}/Localizable#{strings_file_ext}.strings" and system "rm #{lang_dir}/Localizable#{strings_file_ext}.strings.bak"
+  fix_script_path = File.join(script_root, 'fix-translation')
+
+  system "#{fix_script_path} #{lang_dir}/Localizable#{strings_file_ext}.strings"
+  system "plutil -lint #{lang_dir}/Localizable#{strings_file_ext}.strings"
```
AliSoftware: iinm, `system`, like many similar methods dealing with executing shell commands, has a safer version which accepts taking each shell parameter as separate method parameters. Which is safer because you don't need to be sure not to forget to shell-escape everything or put each parameter inside quotes manually.

```ruby
# unsafe version: will fail if e.g. either `fix_script_path` or `lang_dir` happens to contain spaces
system "#{fix_script_path} #{lang_dir}/Localizable#{strings_file_ext}.strings"
# manual, safer version: quoting everything manually. Could fail if one of the variable expansion itself contains quotes, and still makes it open to injection and security vulnerabilities
system "\"#{fix_script_path}\" \"#{lang_dir}/Localizable#{strings_file_ext}.strings\""
# Safer way, letting system make everything secure and properly escaped (actually, avoids involving the shell interpreter internally to let it split the string into parameters, and instead popen the execution process with the right argvs directly)
# and thus doesn't risk any issue or security vulnerability with spaces, unescaped-quotes, or env variable injection
system(fix_script_path, "#{lang_dir}/Localizable#{strings_file_ext}.strings")
```

(only commenting there, but ofc apply same for every other call to `system`)

fastlane/Fastfile
```diff
@@ -7,8 +7,16 @@ unless FastlaneCore::Helper.bundler?
   UI.user_error!('Please run fastlane via `bundle exec`')
 end
 
+# Paths that are re-used across multiple lanes
+PROJECT_ROOT_FOLDER=File.dirname Dir.pwd
```
AliSoftware: Not sure if it's safe to base the logic on `pwd` – i.e. the current directory from where the user runs the script. What if the user runs `fastlane` from the root dir vs from inside the `fastlane/` folder?

I think that given how fastlane runs its Fastfile, in some way it guarantees that the `pwd ` is always inside the `fastlane/` subfolder when executing the `Fastfile`, but still, feels a bit brittle compared to basing that on the `File.dirname(File.dirname(__FILE__))` or something like that? (needs some testing tho, didn't check that `__FILE__` would work in fastlane's context)

fastlane/Fastfile
```diff
@@ -57,10 +65,9 @@ Dotenv.load(USER_ENV_FILE_PATH)
 Dotenv.load(PROJECT_ENV_FILE_PATH)
 ENV[GHHELPER_REPO="wordpress-mobile/wordpress-iOS"]
 ENV["PROJECT_NAME"]="WordPress"
-ENV["PUBLIC_CONFIG_FILE"]="../config/Version.Public.xcconfig"
-ENV["INTERNAL_CONFIG_FILE"]="../config/Version.internal.xcconfig"
+ENV["PUBLIC_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.Public.xcconfig")
+ENV["INTERNAL_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.internal.xcconfig")
```
AliSoftware, line 69: 👍

fastlane/Fastfile
```diff
@@ -130,11 +137,10 @@ import "./ScreenshotFastfile"
   #####################################################################################
   desc "Updates the AppStoreStrings.po file with the latest data"
   lane :update_appstore_strings do | options |
-    prj_folder = Pathname.new(File.join(Dir.pwd, "../..")).expand_path.to_s
-    source_metadata_folder = File.join(prj_folder, "Scripts/fastlane/appstoreres/metadata/source")
+    source_metadata_folder = File.join(PROJECT_ROOT_FOLDER, "fastlane/appstoreres/metadata/source")
 
     files = {
-      whats_new: File.join(prj_folder,  "/WordPress/Resources/release_notes.txt"),
+      whats_new: File.join(PROJECT_ROOT_FOLDER,  "/WordPress/Resources/release_notes.txt"),
```
AliSoftware: nit:
```suggestion
      whats_new: File.join(PROJECT_ROOT_FOLDER, "WordPress/Resources/release_notes.txt"),
```
or even better:
```suggestion
      whats_new: File.join(PROJECT_ROOT_FOLDER, "WordPress", "Resources", "release_notes.txt"),
```

fastlane/Fastfile
```diff
@@ -469,30 +469,52 @@ import "./ScreenshotFastfile"
...
-    sh("cd ../.. && ./Scripts/update-translations.rb review waiting")
-    sh("cd ../.. && ./Scripts/update-translations.rb review fuzzy")
+    puts "Building version: #{options[:version_name]}"
 
-    ios_merge_translators_strings(strings_folder:"../WordPress/Resources/")
-    sh("cd ../.. && ./Scripts/extract-framework-translations.swift")
+    update_translations_script_path = File.join( PROJECT_ROOT_FOLDER, 'Scripts', 'update-translations.rb' )
+    sh("#{update_translations_script_path} review current")
+    sh("#{update_translations_script_path} review waiting")
+    sh("#{update_translations_script_path} review fuzzy")
```
AliSoftware: Same remark as with `system` about spaces-in-path + shell parsing skip performance + injection/security vulnerability:

```suggestion
    sh(update_translations_script_path, "review", "current")
    sh(update_translations_script_path, "review", "waiting")
    sh(update_translations_script_path, "review", "fuzzy")
```

AliSoftware (commented):
Only initial review for now (by just reading diff from GitHub per commit; not tested locally yet)

Scripts/update-translations.rb
jkmassel: Ah good question:

- Removed `s` because it silences warnings, such as the one that was being emitted about glob syntax, which required
- Adding `g` to tell `curl` to ignore `[` and `]` in the URL rather than interpreting them as glob syntax

Once this moves into the toolkit, hopefully we can avoid using `curl` directly at all – that's been a source of _many_ issues in the past

Scripts/update-translations.rb
jkmassel: Oh nice – I like this quite a lot – will run around and audit for it

jkmassel: Applied in bb46c203005ff89ffdeadf292ebe194207943663

fastlane/Fastfile
jkmassel: I actually went back and forth on this – I originally had it the other way, but figure this was more idiomatic 😅 

Changed in 32241c00ae722b6cfce096c3e298a6b53d44af87

fastlane/Fastfile
jkmassel: Love this – addressed in 868f344b5083164afb48d5460be0c660cfec23cb to handle other instances as well (in two cases, I ditched the use of `sh` entirely since it was trivial to do the work in Ruby instead)

Scripts/extract-framework-translations.swift
jkmassel: Fixed in an IMHO reasonably clean way in 9ecb9f5e0b9a874350d729080578891f92c8f927

Scripts/update-translations.rb
loremattei: Hmm... removing `s` will add quite a lot of noise to the output (I'm thinking about all the progress bars). 
If we are only interested in errors, can we use `sS`?
Otherwise, we may probably try `--no-progress-bar`, even though this doesn't work well in some versions iirc.

Scripts/update-translations.rb
```diff
@@ -124,27 +119,44 @@
...
   system "grep -a '\\x00\\x20\\x00\\x22\\x00\\x22\\x00\\x3b$' #{lang_dir}/Localizable#{strings_file_ext}.strings"
+
+  # Clean up after ourselves
+  FileUtils.rm destination if File.exist? destination
```
loremattei: [**] I'm not sure about this. I think this has been added to replace the call to `system "rm #{lang_dir}/Localizable#{strings_file_ext}.strings.bak"` which has been removed from above, right?

If so, I think we should add `.bak` to `destination`, otherwise it looks like to me that this will delete the `strings` files instead of the temporary backups.

fastlane/Fastfile
```diff
@@ -57,10 +65,9 @@ Dotenv.load(USER_ENV_FILE_PATH)
 Dotenv.load(PROJECT_ENV_FILE_PATH)
 ENV[GHHELPER_REPO="wordpress-mobile/wordpress-iOS"]
 ENV["PROJECT_NAME"]="WordPress"
-ENV["PUBLIC_CONFIG_FILE"]="../config/Version.Public.xcconfig"
-ENV["INTERNAL_CONFIG_FILE"]="../config/Version.internal.xcconfig"
+ENV["PUBLIC_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.Public.xcconfig")
+ENV["INTERNAL_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.internal.xcconfig")
 ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
-ENV["PROJECT_ROOT_FOLDER"]="../"
```
loremattei, line 63: [**] I'm not sure we can get rid of passing `PROJECT_ROOT_FOLDER` as an env variable 🤔 IIRC this is used in the `release-toolkit` as well.

fastlane/Fastfile
```diff
@@ -308,22 +314,22 @@ import "./ScreenshotFastfile"
 
     gym(
       scheme: "WordPress Alpha",
-      workspace: "../WordPress.xcworkspace",
+      workspace: WORKSPACE_PATH,
       export_method: "enterprise",
       clean: true,
-      output_directory: "../build/",
-      derived_data_path: "../derived-data/alpha/",
```
loremattei, line 315: With that we're moving the `derived_data_path` to `DerivedData` for both the Internal and the Beta 🤔 
I can't think about issues with this, but wanted to highlight to make sure this is wanted.

fastlane/Fastfile
```diff
@@ -308,22 +314,22 @@ import "./ScreenshotFastfile"
...
+      output_directory: BUILD_PRODUCTS_PATH,
+      derived_data_path: DERIVED_DATA_PATH,
       export_team_id: ENV["INT_EXPORT_TEAM_ID"],
-      export_options: { method: "enterprise" })
-
-    sh("mv ../../build/WordPress.ipa \"../../build/WordPress Alpha.ipa\"")
+      export_options: { method: "enterprise" }
```
loremattei, line 327: This looks good to me on CI, but I found it useful to be able to run the lane locally to build both the versions and have both of them available at the end.

fastlane/Fastfile
```diff
@@ -368,33 +374,30 @@ import "./ScreenshotFastfile"
...
       export_method: "enterprise",
       clean: true,
-      output_directory: "../build/",
-      derived_data_path: "../derived-data/internal/",
+      output_directory: BUILD_PRODUCTS_PATH,
+      derived_data_path: DERIVED_DATA_PATH,
       export_team_id: get_required_env("INT_EXPORT_TEAM_ID"),
-      export_options: { method: "enterprise" })
-
-    sh("mv ../../build/WordPress.ipa \"../../build/WordPress Internal.ipa\"")
```
loremattei, line 379: Same as above

loremattei (commented):
I gave a first pass and left a few comments. 

I'm going to be AFK for the rest of this week, but I don't want to block the work on this PR, so I've put a [**] on the two that I think can lead to bugs and should be addressed or carefully tested before merging. 

For the others, feel free to do what you prefer.

Scripts/update-translations.rb
AliSoftware: tbh at that point, why not use `net/http` and something like this (not tested, might contain typos):

```ruby
File.write(destination, Net::HTTP.get(URI.parse(url)))
```

_Note: There might be a potential gotcha with this solution as it reads the whole downloaded content in memory before writing it to disk, as opposed to streaming the write as the response body comes; but those files are <800KB so should not be that much of an issue here._

fastlane/Fastfile
AliSoftware: It is indeed, well spotted.

Though now that all our projects are aligned to have the `fastlane/` folder at root after this change, we could consider make the `release-toolkit` default that value to `../`, or better yet, compute it – by move up the dir hierarchy in a `while` loop until we find `.git/` folder, or by just using `git rev-parse --show-toplevel`. But that's for yet another bullet point that I'll have to add to my long refactoring todo list, so we're not there yet in release-toolkit and we should keep the env var for now.

Scripts/update-translations.rb
jkmassel: >Hmm... removing s will add quite a lot of noise to the output (I'm thinking about all the progress bars).
If we are only interested in errors, can we use sS?

Yes, this could certainly be worth a try! I can test it against the old code and see if it fails nicely for us.

As for using `net/http`, that's possible but as you mentioned @AliSoftware, potentially problematic – one big issue is that we have to follow redirects, which IIRC `Net::HTTP` doesn't do all that well?

IMHO ideally this code would live in the release toolkit (or a dependency of it?), and from there we could use something like the [HTTP gem](https://github.com/httprb/http) to do this work at a higher level that makes it easier to handle gotchas. 

I think for the purposes of this PR though I'd prefer to change it as little as possible 😅

fastlane/Fastfile
jkmassel: Ah ok – easy enough to fix – I'll just set it for now then we can remove it in the Great Refactor ™

ref: 38c660e23f00934ba8bd0e29f2f9bdb3b519fede

fastlane/Fastfile
jkmassel: Yeah, I thought it was weird that we were using different directories for each - especially locally it sort of makes the project directory messy.

I don't foresee any issues with this (especially given that we have `clean: true` there – it'll do cleanup prior to the build locally.

fastlane/Fastfile
jkmassel: Hmm good point – I might make it so that this just moves them to the `Artifacts` directory that we already have – that way it's consistent across the whole project – if you want the `.ipa` file locally, it'll always be in `Artifacts` (which is what we use for the translation lane)

fastlane/Fastfile
AliSoftware: Tracked in https://github.com/wordpress-mobile/release-toolkit/issues/205

Scripts/update-translations.rb
jkmassel: Resolved in 7f6cdede4b58aea894088c46bd06646719893123!

Scripts/update-translations.rb
jkmassel: Resolved the `s` issue in f090c57efcedd83ea6ede1205719e41410a44e0c – I've just added it back for now – I feel like in the near-ish future we'll be replacing this anyway, so no point over-stressing about it.

fastlane/Fastfile
jkmassel: This is adjusted in b15122eac37aa7cb79263daa86d1995c8012112a, which puts the build artifacts for all lanes in the `Artifacts` directory.

fastlane/Fastfile
jkmassel: This is adjusted in b15122eac37aa7cb79263daa86d1995c8012112a, which puts the build artifacts for all lanes in the `Artifacts` directory.

.gitignore
jkmassel: Cleaned up in bf2ccf60b8bbed5ebe02fe1cb1d8dd5558db2e5e

jkmassel:
@AliSoftware this is ready for another look when you are!

.circleci/config.yml
```diff
@@ -232,28 +228,20 @@ jobs:
...
             curl -sL https://sentry.io/get-cli/ | bash
       - run:
           name: Build
-          working_directory: Scripts
           command: | 
-            orig_version=`cat ../config/Version.internal.xcconfig | grep VERSION_SHORT | cut -d'=' -f2`
-            sed -i '' "$(awk '/^VERSION_SHORT/{ print NR; exit }' "../config/Version.internal.xcconfig")s/=.*/="${orig_version}"-`date +"%Y%d%m"`-"${CIRCLE_BUILD_NUM}"/" "../config/Version.internal.xcconfig"
-            bundle exec fastlane build_for_translation_review
+            orig_version=`cat config/Version.internal.xcconfig | grep VERSION_SHORT | cut -d'=' -f2`
+            sed -i '' "$(awk '/^VERSION_SHORT/{ print NR; exit }' "config/Version.internal.xcconfig")s/=.*/="${orig_version}"-`date +"%Y%d%m"`-"${CIRCLE_BUILD_NUM}"/" "config/Version.internal.xcconfig"
```
AliSoftware, line 238: Feedback for another, separate PR: shouldn't this automatic version "bump" / edit be done by a dedicated release-toolkit action and/or integrated within the `build_for_translation_review` lane? 🤔

Scripts/update-translations.rb
```diff
@@ -124,27 +121,52 @@
...
+    FileUtils.copy destination, backup_destination
+  end
 
   url = "#{download_url}/#{code}/default/export-translations?#{strings_filter}format=strings"
-  destination = "#{lang_dir}/Localizable#{strings_file_ext}.strings"
 
-  system "curl -fLso #{destination} \"#{url}\"" or begin
+  system "curl", "-fgsLo", destination, url or begin
     puts "Error downloading #{code}"
   end
```
AliSoftware: For readability I'd add parentheses around arguments in that case, and/or just avoid the `begin/end` block given it's a one-liner
```suggestion
  system("curl", "-fgsLo", destination, url) || puts "Error downloading #{code}"
```

Though not sure what's `#{code}` is supposed to be, I think we won't capture the error code as expected (I think we'd need to resort to using `Open3.popen3` to capture the error). So overall and given we know we'll refactor that sonn-ish anyway, maybe just use the url in the error message instead.

```suggestion
  system("curl", "-fgsLo", destination, url) || puts "Error downloading #{url}"
```

Scripts/update-translations.rb
```diff
@@ -124,27 +121,52 @@
...
-  system "plutil -lint #{lang_dir}/Localizable#{strings_file_ext}.strings" and system "rm #{lang_dir}/Localizable#{strings_file_ext}.strings.bak"
-  system "grep -a '\\x00\\x20\\x00\\x22\\x00\\x22\\x00\\x3b$' #{lang_dir}/Localizable#{strings_file_ext}.strings"
+  fix_script_path = File.join(script_root, 'fix-translation')
+
+  # References a file like: "#{lang_dir}.Localizable-old.strings" where strings_file_ext == "old"
+  strings_file_path = File.join(lang_dir, "Localizable#{strings_file_ext}.strings")
+
+  system fix_script_path, strings_file_path
+  system "plutil", "-lint", strings_file_path
+  system "grep", "-a", "\\x00\\x20\\x00\\x22\\x00\\x22\\x00\\x3b$", strings_file_path
```
AliSoftware, line 160: Might be worth adding a comment to explain wth this odd escape sequence is for. From me trying to read in the matrix, it seems like the UTF-16 sequence for `<space><quote><quote><semicolon>` aka ` "";`, so I'm guessing it's to print any potential empty strings that would be remaining in the file even after running the fix script that is supposed to get rid of all of them (and replacing those with the key as a fallback for missing translations)?

```suggestion
  # The fix-translations script is supposed to have replaced all strings with empty translations, to use the key (aka English copy)
  # as the translation instead, as a fallback. So after that pass we should not have any more empty translations left,
  # but just as a control, let's grep for lines ending with the UTF-16 sequence ` "";` to ensure there's none left.
  system "grep", "-a", "\\x00\\x20\\x00\\x22\\x00\\x22\\x00\\x3b$", strings_file_path
```

Scripts/update-translations.rb
```diff
@@ -124,27 +121,52 @@
...
+
+  # Clean up after ourselves
+  FileUtils.rm_f backup_destination
+
+  # Delete the downloaded file if we're using a strings filter
+  if !strings_file_ext.empty?
+        FileUtils.rm_f destination
```
AliSoftware: nit: indentation inconsistency (too many spaces)
```suggestion
    FileUtils.rm_f destination
```

fastlane/Fastfile
```diff
@@ -0,0 +1,690 @@
...
+fastlane_require 'xcodeproj'
+fastlane_require 'dotenv'
+fastlane_require 'open-uri'
+
+unless FastlaneCore::Helper.bundler?
+  UI.user_error!('Please run fastlane via `bundle exec`')
+end
+
+# Paths that are re-used across multiple lanes
+PROJECT_ROOT_FOLDER=File.dirname File.expand_path __dir__
```
AliSoftware: function composition like that seem harder to read without using parentheses around function parameters imho
```suggestion
PROJECT_ROOT_FOLDER=File.dirname(File.expand_path(__dir__))
```

fastlane/Fastfile
```diff
@@ -0,0 +1,690 @@
...
+      "app_store_screenshot-4" => File.join(source_metadata_folder, "promo_screenshot_4.txt"),
+      "app_store_screenshot-5" => File.join(source_metadata_folder, "promo_screenshot_5.txt"),
+      "app_store_screenshot-6" => File.join(source_metadata_folder, "promo_screenshot_6.txt"),
+      "app_store_screenshot-7" => File.join(source_metadata_folder, "promo_screenshot_7.txt"),
+    }
+
+    ios_update_metadata_source(po_file_path: PROJECT_ROOT_FOLDER + "/WordPress/Resources/AppStoreStrings.po",
+      source_files: files,
+      release_version: options[:version])
+  end
```
AliSoftware: Because the indentation broke my brain (and I initially thought that the closing `end` on the next line was misaligned), and because we ❤️  `File.join` too
```suggestion
    ios_update_metadata_source(po_file_path: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources', 'AppStoreStrings.po'),
      source_files: files,
      release_version: options[:version]
    )
  end
```

fastlane/Fastfile
```diff
@@ -0,0 +1,690 @@
...
+      ios_bump_version_beta()
+    end
+    ios_final_tag(options)
+
+    # Wrap up
+    version = ios_get_app_version()
+    removebranchprotection(repository:GHHELPER_REPO, branch: "release/#{version}")
+    setfrozentag(repository:GHHELPER_REPO, milestone: version, freeze: false)
+    create_new_milestone(repository:GHHELPER_REPO)
+    close_milestone(repository:GHHELPER_REPO, milestone: version)
```
AliSoftware: Super-nitpick:
```suggestion
    removebranchprotection(repository: GHHELPER_REPO, branch: "release/#{version}")
    setfrozentag(repository: GHHELPER_REPO, milestone: version, freeze: false)
    create_new_milestone(repository: GHHELPER_REPO)
    close_milestone(repository: GHHELPER_REPO, milestone: version)
```

fastlane/Fastfile
```diff
@@ -0,0 +1,690 @@
...
+  # bundle exec fastlane build_and_upload_release [skip_confirm:<skip confirm>] [create_gh_release:<create release on GH>]
+  #
+  # Example:
+  # bundle exec fastlane build_and_upload_release
+  # bundle exec fastlane build_and_upload_release skip_confirm:true
+  # bundle exec fastlane build_and_upload_release create_gh_release:true
+  #####################################################################################
+  desc "Builds and updates for distribution"
+  lane :build_and_upload_release do | options |
+    final_tag = (is_ci() == true) ? ios_validate_ci_build() : false
```
AliSoftware:
```suggestion
    final_tag = is_ci ? ios_validate_ci_build() : false
```
Or alternatively:
```suggestion
    final_tag = is_ci() && ios_validate_ci_build()
```

fastlane/Fastfile
```diff
@@ -0,0 +1,690 @@
...
+    create_release = (final_tag && is_ci()) || options[:create_gh_release]
```
AliSoftware: `final_tag` can only possibly be `true` if `is_ci` (if `is_ci` is false, the previous line will always define `final_tag` to be `false`), so `final_tag && is_ci` can be simplified to just `final_tag`… right?
```suggestion
    create_release = final_tag || options[:create_gh_release]
```

fastlane/Fastfile
```diff
@@ -0,0 +1,690 @@
...
+      result_bundle: true
+    )
+  end
+
+########################################################################
+# Helper Lanes
+########################################################################
+  desc "Get a list of pull request from `start_tag` to the current state"
+  lane :get_pullrequests_list do | options |
+    get_prs_list(repository:GHHELPER_REPO, start_tag:"#{options[:start_tag]}", report_path:"#{File.expand_path('~')}/wpios_prs_list.txt")
```
AliSoftware: super-nit:
```suggestion
    get_prs_list(repository: GHHELPER_REPO, start_tag:"#{options[:start_tag]}", report_path:"#{File.expand_path('~')}/wpios_prs_list.txt")
```

fastlane/README.md
```diff
@@ -0,0 +1,132 @@
+fastlane documentation
```
AliSoftware: Don't we want to git-ignore that file? And same for `fastlane/Preview.html`?

Those files are shown as entirely brand new files, while they are properly listed in the `.gitignore`… so maybe those were `git add` by accident _before_ the paths being updated in the `.gitignore`?

Also, side-note, I see we also have a `fastlane/.gitignore` in addition too the `.gitignore` at the root… probably merging the two (I'm not even sure the files listed in `fastlane/.gitignore` are still relevant anymore anyway?)

.gitignore
```diff
@@ -63,27 +63,29 @@ WordPress/Derived Sources/
...
-Scripts/fastlane/promo_screenshots
-/Scripts/fastlane/metadata/review_information
-/Scripts/default.profraw
+fastlane/Preview.html
+fastlane/report.xml
+fastlane/README.md
+Preview.html
+fastlane/test_output/
+fastlane/google_cloud_keys.json
+fastlane/promo_screenshots
```
AliSoftware, line 78: Those `fastlane/promo_screenshots` were added and commited as part of this PR, _before_ you updated the path to `.gitignore`. Which means that all the promo screenshot PNGs are now erroneously into the repo 😱

.gitignore
```diff
@@ -63,27 +63,29 @@ WordPress/Derived Sources/
...
-Scripts/fastlane/test_output/
-Scripts/fastlane/google_cloud_keys.json
-Scripts/fastlane/promo_screenshots
-/Scripts/fastlane/metadata/review_information
-/Scripts/default.profraw
+fastlane/Preview.html
+fastlane/report.xml
+fastlane/README.md
+Preview.html
+fastlane/test_output/
```
AliSoftware, line 76: The directory was commited to git before you fixed the `.gitignore` path, so this PR currently have the `test-output/report.{html,junit}` files

AliSoftware (changes requested):
The recurring theme when reviewing this PR from the GitHub UI is clearly this 😅 
image

Joke aside, added a couple of nits, but also saw **a couple of files that were commited while they really shouldn't**.

It can easily be seen by looking at the diff between `16.6.0.1` (the tag from which you started to initially cut the branch from at that time, iinm) and `move/fastlane` – after cloning the branches locally (in Terminal, because a diff in GitHub UI will be swamped by the diff size) – like so: ` git diff --summary 16.6.0.1 move/fastlane | grep -v rename`

AliSoftware:
Note: I've created a Draft P2 post in +iosp2 (edit link sent in DM – ref: paNNhX-c0-p2) – just so we don't forget to announce the change made by that PR – that is ready to be published once we're ready to approve and merge this PR.

Scripts/update-translations.rb
loremattei: I'm not sure I understand this 🤔 I _think_ we should never delete the `destination` file.
But I may be missing something here, tbh. Why are you doing this?

loremattei (changes requested):
This looks generally good to me, other than the accidentally added files that Olivier pointed out. 
I've still a doubt about the refactoring of `update-translations.rb`, so I left a question there. 

I've tried to run a couple of modified lanes and they worked as expected anyway :-)

.circleci/config.yml
jkmassel: Yes, IMHO this would make a lot of sense – also that would help make this code more portable 😏

Created https://github.com/wordpress-mobile/release-toolkit/issues/209 for this.

fastlane/Fastfile
jkmassel: Yeah, this is where having Rubocop or something would help me – I need something to yell at me when I get the rules wrong 😁 

Thanks!!

fastlane/README.md
jkmassel: Yeah, good question:

Addressed in:
d9e55ae027c828c2be41dade8a11ced887546638
e2f41e5ba8ecbb8abe50e040df5d9bda3cc34513

.gitignore
jkmassel: Shoot, yep. Fixed in d76a8e6ae4450360a8842f3dd0c1e0e85c1ee591

fastlane/Fastfile
jkmassel: Fixed in 6848df39976d84edbfbd0a0a11eb22949a24876d

fastlane/Fastfile
jkmassel: Addressed in 821f0745fc3aea246108d9d6f2512f559eec86c2

fastlane/Fastfile
jkmassel: Addressed in 821f0745fc3aea246108d9d6f2512f559eec86c2

fastlane/Fastfile
jkmassel: That sounds right to me, yep!

Addressed in 821f0745fc3aea246108d9d6f2512f559eec86c2

fastlane/Fastfile
jkmassel: Addressed in 821f0745fc3aea246108d9d6f2512f559eec86c2

.gitignore
jkmassel: Fixed in d9e55ae027c828c2be41dade8a11ced887546638, thanks!

Scripts/update-translations.rb
jkmassel: Ah, thanks for following up on this. I thought this script would merge them, but it looks like that actually happens in `ios_merge_translators_strings`, so you're right – it shouldn't be removed. 

Fixed in 6f0d1894c27577a6d3ea8a6a53853b1f2fa169f0

jkmassel:
Thanks for the feedback @loremattei and @AliSoftware – it should all be addressed.

I did need to adjust one thing – Ruby wasn't happy with:

```ruby
system("curl", "-fgsLo", destination, url) || puts "Error downloading #{url}"
```

but it was fine with

```ruby
  system("curl", "-fgsLo", destination, url) or raise "Error downloading #{url}"
```

for reasons I don't entirely understand.

That said, I think I prefer this, because IMHO if it can't download the strings it should fail the build instead of carrying on with missing strings.

loremattei (approved):
LGTM!
I briefly tested `update-translation.rb` and it worked as expected. I recommend double checking the diff it generates on the first actual release.
:shipit:

AliSoftware (approved):
The merge of release/16.7 into that branch made it a bit hard to ensure the last diff was right, but looking at the incremental diffs from last time (537 commits behind 😱) it LGTM.
```
…
Switched to branch 'move/fastlane'
Your branch is behind 'origin/move/fastlane' by 537 commits, and can be fast-forwarded.
```

So I say :shipit: and:

 - Don't forget to publish the draft P2 post I prepared [paNNhX-c0-p2] to announce the change (even if we MIEs might be more frequent users of those lanes than the code wranglers)
 - As Lore said, we might want to remember to double-check the diff during next release just to be sure. (It will always be time to fix anything we might have missed later anyway if needed, so better finally merge it ASAP)

PR #15715, opened 2021-01-26
Update WPKit

leandroalonso:
Updates WPKit to the latest beta version and update the required callbacks.

### To test

A green build should be enough.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -43,7 +43,7 @@ def wordpress_ui
 end
 
 def wordpress_kit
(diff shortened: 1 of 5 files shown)


leandroalonso:
@momo-ozawa I tagged you so you can be aware of this change.

#15689 depends on another PR so I thought it would be good to just update `develop`, since the change is quite small.

ScoutHarris (approved):
Verified WPiOS now builds successfully. Thanks @leandroalonso !

:shipit:

PR #15730, opened 2021-01-28
Get the index from the scanner's string, not the parsed string.

aerych:
This PR fixes an out of index issue I spotted while testing a patch for @leandroalonso.  This PR uses the correct string when retrieving an index from the scanner.

To test:
- Test in the reader with comments that contain images and videos. 
- Ensure that the images and videos continue to display correctly. 

Note there is a different, non iOS 13 specific, out of index error addressed by @leandroalonso's patch. 

@leandroalonso would you be game to review after the changes in the other PR are merged to develop?

PR submission checklist:


leandroalonso (approved):
Applied the changes from #15735, tested, and no crash! :shipit:

aerych:
Thank you! :)

PR #15734, opened 2021-01-28
Open stats in the app when tapping on Today Widget

Gio2018:
Fixes an issue that causes tapping on the widget to just open the app, without showing the current stats
Fixes #NA

To test:

- build/run and install the widget
- configure the widget with any site
- tap on the widget and make sure it opens a modal in the app displaying the stats of that same site

PR submission checklist:

### DIFF
--- a/WordPress/Classes/System/WordPressAppDelegate+openURL.swift
+++ b/WordPress/Classes/System/WordPressAppDelegate+openURL.swift
@@ -11,6 +11,11 @@ import AutomatticTracks
             return true
         }
 
+        if UniversalLinkRouter.shared.canHandle(url: url) {
+            UniversalLinkRouter.shared.handle(url: url, shouldTrack: true)
+            return true
+        }
+
         guard url.scheme == WPComScheme else {
             return false
         }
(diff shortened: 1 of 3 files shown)


frosty (commented):
This works as expected, however I wonder if we could make a few alterations...

* This URL scheme only takes us to the root stats screen for the site, but it would be best if we could go directly to the Day view (as this comes from the Today widget).
* We could achieve that by using the existing Day stats deep link. This is usually triggered by `application(_:continue:restorationHandler:)` in the app delegate, but it seems that the widgetURL calls `application(_:open:options:)` instead. I've put together [a small gist](https://gist.github.com/frosty/21010244fb0148f380ac611bf82db3d9) that would pass this over to our normal link router if the URL doesn't have the special `wordpress://` or `wpdebug://` protocol.
* I've also made a tweak to the analytics handling there, so we can track these deep links originating from the widget.

Take a look at the gist and let me know what you think 🙂

Gio2018:
> This works as expected, however I wonder if we could make a few alterations...
> 
> * This URL scheme only takes us to the root stats screen for the site, but it would be best if we could go directly to the Day view (as this comes from the Today widget).
> * We could achieve that by using the existing Day stats deep link. This is usually triggered by `application(_:continue:restorationHandler:)` in the app delegate, but it seems that the widgetURL calls `application(_:open:options:)` instead. I've put together [a small gist](https://gist.github.com/frosty/21010244fb0148f380ac611bf82db3d9) that would pass this over to our normal link router if the URL doesn't have the special `wordpress://` or `wpdebug://` protocol.
> * I've also made a tweak to the analytics handling there, so we can track these deep links originating from the widget.
> 
> Take a look at the gist and let me know what you think 🙂

Thank you @frosty ! This is a much better experience; just pushed those changes, ready for a second pass whenever you can.

frosty (approved):
Looks great! :shipit:

PR #15737, opened 2021-01-28
Reader: show 'Followed P2s' tab when first P2 is followed

ScoutHarris:
Fixes #n/a
Ref: #15343 

When a user follows their first P2, the `Followed P2s` tab now appears immediately. Previously, it would require an app restart for it to show.

To test:
On an account that is not following any P2s:
- Go to Reader > Manage (the gear icon in the nav bar).
- Enter the URL of a P2 site to follow.
- Close the Manage view.
- Verify the `Followed P2s` tab appears.

Notes:
- If you unfollow all P2s, the menu item is still present. It will be removed on app restart. It got... hinky when I tried to auto-remove it. The empty state works fine, so this should be sufficient for now.
- If you unfollow a P2, the stream is not automatically refreshed. This is the behavior of the other streams, so no change here. Pull to refresh will of course refresh the stream.
- Unfollowing an external site from a reader card or post details doesn't seem to work. I checked the app store version (16.5) and it's broken there as well. I'll look at this separately as it is unrelated to this particular issue.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/ReaderSiteService.m
+++ b/WordPress/Classes/Services/ReaderSiteService.m
@@ -4,7 +4,6 @@
 #import "ContextManager.h"
 #import "ReaderPostService.h"
 #import "ReaderPost.h"
-#import "ReaderTopicService.h"
 #import "WPAccount.h"
 #import "WordPress-Swift.h"
 #import "WPAppAnalytics.h"
@@ -201,6 +200,21 @@ NSString * const ReaderSiteServiceErrorDomain = @"ReaderSiteServiceErrorDomain";
         if (failure) {
             failure(error);
         }
+    }];
+}
+
+- (void)topicWithSiteURL:(NSURL *)siteURL success:(void (^)(ReaderSiteTopic *topic))success failure:(void(^)(NSError *error))failure
+{
+    WordPressComRestApi *api = [self apiForRequest];
+    ReaderSiteServiceRemote *service = [[ReaderSiteServiceRemote alloc] initWithWordPressComRestApi:api];
+    
+    [service findSiteIDForURL:siteURL success:^(NSUInteger siteID) {
+        ReaderTopicService *topicService = [[ReaderTopicService alloc] initWithManagedObjectContext:self.managedObjectContext];
+        NSNumber *site = [[NSNumber alloc] initWithUnsignedLong:siteID];
+        ReaderSiteTopic *topic = [topicService findSiteTopicWithSiteID:site];
+        success(topic);
+    } failure:^(NSError *error) {
+        failure(error);
     }];
 }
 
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
```diff
@@ -259,6 +260,18 @@ private extension ReaderTabView {
...
+
+    @objc func siteFollowed(_ notification: Foundation.Notification) {
+        guard let userInfo = notification.userInfo,
+              let site = userInfo[ReaderNotificationKeys.topic] as? ReaderSiteTopic,
+              site.organizationType == .p2 else {
+            return
+        }
+
+        // Refresh the Reader menu to ensure Followed P2s is displayed.
+        viewModel.fetchReaderMenu()
```
ScoutHarris, line 274: I wanted to add a check for `Followed P2s` in the Reader menu before refreshing it. However, the only way I could see to do that was by the title string. Similar to how the Discover tab is found - `tabBar.items.firstIndex(where: { $0.title == NSLocalizedString("Discover", comment: "Discover tab name") })`. I didn't particularly want to do that, so I'll look into this further later. But for now this will do.

WordPress/Classes/Services/ReaderSiteService.m
```diff
@@ -204,6 +203,21 @@ - (void)flagSiteWithID:(NSNumber *)siteID asBlocked:(BOOL)blocked success:(void(
...
 }
 
+- (void)topicWithSiteURL:(NSURL *)siteURL success:(void (^)(ReaderSiteTopic *topic))success failure:(void(^)(NSError *error))failure
+{
+    WordPressComRestApi *api = [self apiForRequest];
+    ReaderSiteServiceRemote *service = [[ReaderSiteServiceRemote alloc] initWithWordPressComRestApi:api];
+    
+    [service findSiteIDForURL:siteURL success:^(NSUInteger siteID) {
+        ReaderTopicService *topicService = [[ReaderTopicService alloc] initWithManagedObjectContext:self.managedObjectContext];
+        NSNumber *site = [[NSNumber alloc] initWithUnsignedLong:siteID];
```
aerych: Purely a matter of stylistic choice, there's `[NSNumber numberWithUnsignedLong:]` if you prefer to not use `alloc`.

aerych (approved):
Howdy @ScoutHarris 👋 
Looks good to me! Tested as described and it worked as expected. :) 
:shipit: when ready!

WordPress/Classes/Services/ReaderSiteService.m
ScoutHarris: Oh thanks! alloc init is pure habit with ObjC. 😄

PR #15745, opened 2021-01-29
Reader: only show seen toggle if seen is supported

ScoutHarris:
Ref: #15355
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/341

This adds a check to the post menu options to show the `Mark as seen` / `Mark as unseen` option based on `isSeenSupported`.

Data model changes:
- The default value of `isSeen` is now `true` to align with Calypso.
- `isSeenSupported` is added to `ReaderPost`.

To test:

The Calypso [change](https://code.a8c.com/D56224) to exclude `is_seen` from older posts is not yet merged, so some posts will not be correct. Since the `read/search` endpoint doesn't return `is_seen`, we can use that to test this logic.

- On any post card in the Reader, tap the options menu button.
  - Verify `Mark as seen` / `Mark as unseen` is present.
- Tap the post card. In the post details, tap the options menu button.
  - Verify `Mark as seen` / `Mark as unseen` is present.
- Tap the search button on the nav bar.
  - Enter a search term.
  - On a post card and post details options menus (same as above), verify `Mark as seen` / `Mark as unseen` is not present.

| Post card | Search card |
|--------|-------|
| post_card | search |

PR submission checklist:

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -3,6 +3,12 @@
 This file documents changes in the data model. Please explain any changes to the
 data model as well as any custom migrations.
 
+## WordPress 112
+
+@scoutharris 2021-01-29
+
+- Added `isSeenSupported` attribute to  `ReaderPost` entity.
+
 ## WordPress 111
 
 @scoutharris 2021-01-14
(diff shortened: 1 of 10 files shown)


aerych (commented):
Hola @ScoutHarris 👋 
Tested as described and the changes look good to me. :) 
:shipit: !

aerych (approved):

ScoutHarris:
Thanks @aerych !

PR #15740, opened 2021-01-29
Jetpack Section: Fix/Ignore Threat

momo-ozawa:
Part of #15190 
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/339

### Decription:
- Update fix threat button to trigger fix threat request
- Update ignore threat button to trigger ignore threat request

#### Fix threat
1. Launch the app, tap on the My Site, tap on a site with Jetpack Scan and also has Threats
2. Tap on a threat cell
3. Tap `Fix threat` button
4. ✅ Fix threat alert is displayed
5. Tap `Cancel` button on alert
6. ✅ Alert is dismissed
7. Tap `Fix threat` button, then tap `OK` button on alert
4. ✅ Should navigate back to the Scan screen
5. After about a minute has passed, pull to refresh
6. ✅ The fixed threat no longer shows up
7. Tap on History > Fixed tab
8. ✅ The fixed threat shows up in the Fixed tab 

#### Ignore threat
1. Launch the app, tap on the My Site, tap on a site with Jetpack Scan and also has Threats
2. Tap on a threat cell
3. Tap `Ignore threat` button
4. ✅ Ignore threat alert is displayed
5. Tap `Cancel` button on alert
6. ✅ Alert is dismissed
7. Tap `Ignore threat` button, then tap `OK` button on alert
4. ✅ Should navigate back to the Scan screen
5. After about a minute has passed, pull to refresh
6. ✅ The ignored threat no longer shows up
7. Tap on History > Ignored tab
8. ✅ The ignored threat shows up in the Ignored tab 

### Example screenshots:

Fix threat | Ignore threat
--- | ---
Simulator Screen Shot - iPhone 12 Pro - 2021-01-29 at 18 56 02 | Simulator Screen Shot - iPhone 12 Pro - 2021-01-29 at 18 56 08

### PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15749, opened 2021-02-01
Jetpack Section: Threat Details File Context

momo-ozawa:
Part of #15190

### Description:
- Added an attributed string for file context in Threat Details

### Merge instructions:
- ⚠️ Merge after https://github.com/wordpress-mobile/WordPress-iOS/pull/15740

### To test:
1. Launch the app, tap on the My Site, tap on a site with Jetpack Scan and also has Threats
2. Tap on a threat cell for a threat that includes a file context
3. ✅ The file context can scroll horizontally
4. ✅ The file context styling should match the design specs  

Screen Shot 2021-02-01 at 15 48 46

### PR submission checklist:


WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
     }
 }
 
+extension JetpackScanThreatViewModel {
+
+    private static func attributedString(for context: JetpackThreatContext?) -> NSAttributedString? {
```
emilylaguna: How do you feel about extracting this into an extension on `JetpackThreatContext` that also takes a struct called something like `JetpackThreatContextRendererConfig`? Then the config struct can return the dictionaries for the NSAttributedString attributes?

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+        guard let context = context else {
+            return nil
+        }
+
+        let lineFont = UIFont.monospacedSystemFont(ofSize: 13, weight: .semibold)
+        let numberFont = UIFont.monospacedSystemFont(ofSize: 13, weight: .bold)
+        let attrString = NSMutableAttributedString()
+
+        let biggestLen = String(context.lines.last?.lineNumber ?? 0).count
```
emilylaguna: This name is pretty confusing. We could probably rename it to `lastNumberLength` which is slightly better, then add some documentation for it:

```        
        // Calculates the "longest" number string length
        // this will be used to pad the number column to make sure it's all the same size regardless of
        // the line number length
        //
        // ie: last line number is 1000, which is 4 chars
        // when processing line 50 we'll append 2 spaces to the end so it ends up being equal to 4
```

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+
+        let longestLine = context.lines.sorted { $0.contents.count > $1.contents.count}.first!
```
emilylaguna: This line causes a crash if the context.lines is an empty array. We could probably guard or default instead of using a forced unwrap.

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+        let longestLen = longestLine.contents.count
+
+        let paragraph = NSMutableParagraphStyle()
```
emilylaguna: I think we can remove this since we don't actually change the paragraph style in the attributes.

```suggestion
```

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+
+        for line in context.lines {
+
+            // Number attributed string
+
+            var numberStr = String(line.lineNumber)
+            let numberPadding = "".padding(toLength: biggestLen - numberStr.count,
```
emilylaguna:
```suggestion
            let numberPadding = String().padding(toLength: biggestLen - numberStr.count,
```

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+            }
+
+            numberAttr.setAttributes([
+                NSAttributedString.Key.font: numberFont,
+                NSAttributedString.Key.foregroundColor: Constants.normal.numberTextColor,
+                NSAttributedString.Key.backgroundColor: numberBackgroundColor,
+            ], range: NSRange(location: 0, length: numberStr.count))
+
+            attrString.append(numberAttr)
+            attrString.append(NSAttributedString(string: ""))
```
emilylaguna: I don't think this is needed anymore. 

```suggestion
```

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+            var numberStr = String(line.lineNumber)
+            let numberPadding = "".padding(toLength: biggestLen - numberStr.count,
+                                           withPad: Constants.noBreakSpace, startingAt: 0)
+            numberStr = numberPadding + numberStr
+            let numberAttr = NSMutableAttributedString(string: numberStr)
+
+            // Contents attributed string
+
+            let contents = line.contents
+            let contentsPadding = "".padding(toLength: longestLen - contents.count,
```
emilylaguna:
```suggestion
            let contentsPadding = String().padding(toLength: longestLen - contents.count,
```

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+                                             withPad: Constants.noBreakSpace, startingAt: 0)
+
+            let padding = " "
```
emilylaguna: I forgot what this was for, we can probably rename it to something like this or .. ?

We should also move this to Constants

```suggestion
            let columnSpacer = " "
```

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -309,8 +311,128 @@ struct JetpackScanThreatViewModel {
...
+            static let textColor = UIColor(rgb: 0x2C3337)
+            static let backgroundColor = UIColor(rgb: 0xF6F7F7)
+            static let numberTextColor = UIColor(rgb: 0x646970)
+            static let numberBackgroundColor = UIColor(rgb: 0xC3C4C7)
+        }
+
+        struct highlighted {
+            static let textColor = UIColor.white
+            static let backgroundColor = UIColor(rgb: 0xD63638)
+            static let numberBackgroundColor = UIColor(rgb: 0xF1DADA)
```
emilylaguna: I think we can just convert these to normal UIColors so we don't need to add the hex to UIColor extension below. 

I normally use https://www.uicolor.io/ to help with this.

emilylaguna (changes requested):
Heya!! Nice work, I hit a few snags and left a few comments below.

Let me know your thoughts!

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
momo-ozawa: Extracted to a private `JetpackThreatContext` extension! ✨  de51d96

I decided to define a private extension inside `JetpackScanThreatViewModel` instead of WPKit, because attributed strings feel closer to a viewmodel than a data model.

momo-ozawa:
@emilylaguna Thanks for reviewing! 
I've addressed your suggestions -- ready for another look! 😄

WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/View Models/JetpackScanThreatViewModel.swift
```diff
@@ -49,7 +50,37 @@ struct JetpackScanThreatViewModel {
...
+                NSAttributedString.Key.backgroundColor: Constants.colors.highlighted.numberBackground
+            ],
+            highlightedSectionAttributes: [
+                NSAttributedString.Key.font: Constants.monospacedFont,
+                NSAttributedString.Key.foregroundColor: Constants.colors.highlighted.text,
+                NSAttributedString.Key.backgroundColor: Constants.colors.highlighted.background
+            ]
+        )
+
+        attributedFileContext = threat.context?.attributedString(with: contextConfig) ?? nil
```
emilylaguna: Since `attributedString(with:`) already returns nil I don't think we need to default here. 

```suggestion
        attributedFileContext = threat.context?.attributedString(with: contextConfig)
```

emilylaguna (changes requested):
Heya! Looking awesome! I left 1 comment below, and I had 2 suggestions:

1. Not exactly related to this so we can touch on it later but can we change the file name font to the monospaced one? 
2. When loading my very large threats list on my site I am getting some stutters due to the attributed string. Can we defer the loading of this until we need it on the Threat Details view?

momo-ozawa:
@emilylaguna Back to you ✨ 

> Not exactly related to this so we can touch on it later but can we change the file name font to the monospaced one?

Fixed! https://github.com/wordpress-mobile/WordPress-iOS/pull/15749/commits/3da4bda7a3d5216871b6eb8ba55b424532a7c6b3

> Since attributedString(with:) already returns nil I don't think we need to default here.

Fixed! c00bc17

> When loading my very large threats list on my site I am getting some stutters due to the attributed string. Can we defer the loading of this until we need it on the Threat Details view?

Fixed! 06069c9

emilylaguna (approved):
LGTM 🚢

Issue #14765, opened 2020-08-26
Reader Improvements: Discover: Sorting Posts


emilylaguna:
Project Issue: https://github.com/wordpress-mobile/WordPress-iOS/issues/14768

### Goal
When the user is on the new Discover screen they are given the option to sort the posts by popularity or most recent.

### Screenshots

<img width="983" alt="Screen Shot 2020-08-26 at 3 07 03 PM" src="https://user-images.githubusercontent.com/793774/91345740-b0773580-e794-11ea-98d1-8a328e12d92d.png">

### Designs
- [Explorations](https://www.figma.com/file/gVXXG9V6AiMkXbBvvy6Ggw/Sorting-Posts?node-id=2%3A387)
- [Specs](https://www.figma.com/file/1I8UKvncCSTRjUtJuLp2xd/Sorting-UX?node-id=1%3A172)

High level asks:

- [ ] Add feature flag
- [ ] Ensure views are correct for both dark and light appearance.
- [ ] Add Accessibility support.
- [ ] Add Tracks.
- [ ] Enable feature flag for internal.
- [ ] Enable feature flag for all.
- [ ] Remove feature flag.

kean:
You can now do that with "topics":

<img width="320" alt="Screenshot 2024-11-15 at 9 55 02 AM" src="https://github.com/user-attachments/assets/fc3ad202-8dd7-404c-846c-dac0deeccc0b">

PR #15754, opened 2021-02-01
Discover posts sorting control

bozidarsevo:
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/340

Basic setup for sorting option with showing the sorting option control in the Reader tab (Discover section) and passing the sorting option when fetching the cards (code added in WPKit PR https://github.com/wordpress-mobile/WordPressKit-iOS/pull/340)

### To test:

Open the Reader tab and check if the control is showing properly.

Things left to check/do:
- [ ] if used strings are localized
- [ ] use proper icons
- [ ] how should the control behave when touched/highlighted
- [ ] set accessibility correctly
- [ ] check if the styling is correct

Simulator Screen Shot - iPhone 12 Pro - 2021-01-31 at 22 48 55

PR submission checklist:


Podfile
```diff
@@ -43,9 +43,9 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.26-beta'
+    # pod 'WordPressKit', '~> 4.26-beta'
     # pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :tag => ''
-    # pod 'WordPressKit', :git => 'https://github.com/wordpress-mobile/WordPressKit-iOS.git', :branch => ''
+    pod 'WordPressKit', :git => 'https://github.com/bozidarsevo/WordPressKit-iOS.git', :branch => 'feature/fetching-cards-sort-option'
```
bozidarsevo: Using feature branch from my forked repository with added sort option parameter

WordPress/Classes/Services/ReaderCardService.swift
```diff
@@ -28,7 +28,7 @@ class ReaderCardService {
         self.siteInfoService = siteInfoService ?? ReaderTopicService(managedObjectContext: coreDataStack.mainContext)
     }
 
-    func fetch(isFirstPage: Bool, refreshCount: Int = 0, success: @escaping (Int, Bool) -> Void, failure: @escaping (Error?) -> Void) {
+    func fetch(isFirstPage: Bool, refreshCount: Int = 0, sortingOption: String? = nil, success: @escaping (Int, Bool) -> Void, failure: @escaping (Error?) -> Void) {
```
bozidarsevo: Added optional `sortingOption` parameter that will be used for fetching cards.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,120 @@
+import UIKit
+
+enum ReaderSortingOption: String, CaseIterable {
```
bozidarsevo: Enum that hold possibly sorting options. Currently only two, `date` and `popularity`.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,120 @@
...
+            // TODO: use proper icon
+            return .gridicon(.calendar)
+        case .popularity:
+            // TODO: use proper icon
+            return .gridicon(.lineGraph)
+        }
+    }
+}
+
+class ReaderSortingOptionButton: UIControl {
```
bozidarsevo, line 31: Control that will be used above the cards in the UI to show current sorting option and when user tap it we will show the bottom sheet for picking sorting options.

bozidarsevo: The control has 3 UI elements, an icon view, label and the chevron view. Icon view and label have different content depending the `sortingOption` value

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -10,6 +10,19 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
     /// Refresh counter used to for random posts on pull to refresh
     private var refreshCount = 0
 
+    /// Sorting option used for fetching cards
+    private var sortingOption: ReaderSortingOption = .popularity {
```
Gio2018: if `sortingButton` has an accessible property `sortingOption`, why do we need another one here that just mirrors that one?

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -31,10 +44,19 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
...
+
         ReaderWelcomeBanner.displayIfNeeded(in: tableView)
         tableView.register(ReaderTopicsCardCell.self, forCellReuseIdentifier: readerCardTopicsIdentifier)
         tableView.register(ReaderSitesCardCell.self, forCellReuseIdentifier: readerCardSitesIdentifier)
 
+        tableView.tableHeaderView = sortingButton
+
+        NSLayoutConstraint.activate([
+            sortingButton.widthAnchor.constraint(equalTo: tableView.widthAnchor),
+        ])
```
Gio2018: can we group the sorting option tasks in a method called here, to leave `viewDidLoad` cleaner?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,124 @@
...
+
+extension ReaderSortingOption {
+
+    var localizedDescription: String? {
+        // TODO: check if strings are localized
+        switch self {
+        case .date:
+            return NSLocalizedString("Recent", comment: "Sorting option description")
+        case .popularity:
+            return NSLocalizedString("Popular", comment: "Sorting option description")
```
Gio2018: can we use comments that clarify a little better what the strings actually are? Like "Description of the popularity sorting option in the Discover tab"

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,124 @@
...
+        view.textColor = UIColor(light: .black, dark: .white)
+        view.font = WPFontManager.systemSemiBoldFont(ofSize: 15.0)
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+
+    private lazy var chevronView: UIImageView = {
+        let view = UIImageView()
+        view.image = .gridicon(.chevronDown)
+        view.tintColor = UIColor(light: UIColor(hexString: "4D4D4D"), dark: UIColor(hexString: "BFBFBF"))
```
Gio2018: can we group the "magic numbers" (e.g. `15.0`) together(e.g. in an `enum`) so they are easily identifiable and modifiable?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,124 @@
...
+
+    var sortingOption: ReaderSortingOption? {
+        didSet {
+            label.text = sortingOption?.localizedDescription
+            iconView.image = sortingOption?.image
+        }
+    }
+
+    // TODO: check how should it behave when touched/highlighted
+    override var isHighlighted: Bool {
```
Gio2018, line 83: would you mind checking if we already have a type in the app that already implements default highlighting behaviors? I'm specifically thinking of the filter button in the "Following" tab

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,124 @@
...
+}
+
+class ReaderSortingOptionButton: UIControl {
+    var sourceView: UIView {
+        return chevronView
+    }
+
+    private lazy var iconView: UIImageView = {
+        let view = UIImageView()
+        view.tintColor = UIColor(light: UIColor(hexString: "4D4D4D"), dark: UIColor(hexString: "BFBFBF"))
```
Gio2018: is it necessary to define custom colors here? I would think we can rely on system colors.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,124 @@
...
+        super.init(coder: coder)
+
+        setup()
+    }
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+
+        setup()
+    }
```
Gio2018: I don't think we need to support the usage of this component from Interface Builder for now

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -0,0 +1,124 @@
...
+            iconView.leadingAnchor.constraint(equalTo: safeAreaLayoutGuide.leadingAnchor, constant: 16.0),
+            label.centerYAnchor.constraint(equalTo: centerYAnchor),
+            label.leadingAnchor.constraint(equalTo: iconView.trailingAnchor, constant: 6.0),
+            label.topAnchor.constraint(equalTo: topAnchor, constant: 16.0),
+            label.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -16.0),
+            chevronView.centerYAnchor.constraint(equalTo: centerYAnchor),
+            chevronView.leadingAnchor.constraint(equalTo: label.trailingAnchor, constant: 6.0),
+            chevronView.trailingAnchor.constraint(lessThanOrEqualTo: safeAreaLayoutGuide.trailingAnchor, constant: -16.0),
+            chevronView.heightAnchor.constraint(equalToConstant: 24.0),
+            chevronView.widthAnchor.constraint(equalToConstant: 24.0),
```
Gio2018: again can we group the magic numbers away?

Gio2018 (changes requested):
Hey @bozidarsevo , this looks good; just had a few changes to consider. Also, I know that we have the gray background in the specs for the sorting button (although the screenshots in the issue are different, sorry for the confusion), I wonder if that should be white, similarly to the filter button in the following page. @mbshakti what do you think?

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
bozidarsevo: Good point, I did not clean that after I added the button, tnx!

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
bozidarsevo: I used the hex values from Figma. Do you know what system colors would those be?
I tried some of the system gray colors and they do not match the ones in Figma.
In Figma for light mode it says "Button / Black Fill / Gray Outline" and hex 4D4D4D, for dark "Button / White Fill / Gray Outline" and hex BFBFBF.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
bozidarsevo: Filter button sets specific colors for normal, selected and highlighted state since it is a subclass of `UIButton`.
I found `AnimatedControl` but that looks a bit too much for this simple case.
Since my control has 2 icons and a label I went with subclassing `UIControl` and with basic override of `isHighlighted`. That is why I also added `TODO` to check how should I handle highlighting

Gio2018: > Filter button sets specific colors for normal, selected and highlighted state since it is a subclass of UIButton.
Do you think it would be possible to use that same subclass then, since it's already there?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
Gio2018: Maybe @mbshakti can give us some guidance? Not a blocker at all, but I think it would be better to use system colors as much as we can.

Gio2018: I'd say it's ok for now to leave them as they are, let's maybe just put a TODO to double check if we can use system colors?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
Gio2018: Sounds good. Can you please add the alpha values in the `Constants` as well?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
Gio2018: Also can you add the values to `Constants`?

Gio2018 (commented):
Just a couple more comments then we are ready to merge this one!

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
bozidarsevo: Sure! For future "proper" solution we can go with a similar patter as `UIButton` has and have special colors/style for highlighted, disabled and selected state which would be applied depending in what state this control is.

bozidarsevo: If/when we have the styles for those states, we can update it to use them

Gio2018 (approved):
This looks good @bozidarsevo !

bozidarsevo:
> This looks good @bozidarsevo !

Should we merge it and then open new PR for next parts of the task?

PR #15758, opened 2021-02-02
Add device modes for View Site previews

mkevins:
### Description

This PR adds device modes to the View Site preview screen.

I utilized the existing `PreviewWebKitViewController`, which subclasses `WebKitViewController`. I also added an optional boolean (default false) to the factory method of the parent class. Since the site preview isn't associated with a particular `post`, it was necessary to make that optional in the subclass.

I''ve added a button for the device modes here: https://github.com/wordpress-mobile/WordPress-iOS/pull/15758/files#diff-8515b18aa430fd6d7e25e540843b387c5b0ef97acdaedf383c39e7b653c235d9R125-R127, but I'm not sure whether I've tested all cases for when `linkBehavior == .all`. @bjtitus :wave: :smile: , do you know whether there are cases where this subclass is being used where the setting is left unimplemented? . Fwiw, when I tried making `post` private, I noticed a build failure from a required visibility in `SearchManager`, so I tried a viewing a few posts from the post list search screen, and everything I tried seems to work as expected.

Ideally, we could enable the device modes everywhere we use these webviews for previews, for consistency, so please feel free to let me know if I've missed a case. On a related note, I'd considered moving the device modes to the parent class, but I think it makes sense to defer to a later PR, as I'd want to have a more comprehensive understanding of the impact before introducing it there (there may be some cases / paths where the feature does not make sense, for example). Wdyt?

To test:

* On the My Site screen, scoll to the bottom and under "External" tap "View Site"
* When the preview opens, expect a device modes button to appear in the lower right corner
* Tap the device modes button (the currently selected device should reflect the default for the device you are on (i.e. mobile or tablet)
* Tap a different device mode (e.g. desktop)
* Expect the preview to reload with the selected device mode's viewport size

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
@@ -4,7 +4,7 @@ import WebKit
 /// An augmentation of WebKitViewController to provide Previewing for different devices
 class PreviewWebKitViewController: WebKitViewController {
 
-    let post: AbstractPost
+    let post: AbstractPost?
 
     private let canPublish: Bool
 
@@ -75,6 +75,8 @@ class PreviewWebKitViewController: WebKitViewController {
     }
 
     func trackOpenEvent() {
+        guard let post = post else { return }
+
         let eventProperties: [String: Any] = [
             "post_type": post.analyticsPostType ?? "unsupported",
             "blog_type": post.blog.analyticsType.rawValue
@@ -162,7 +164,7 @@ class PreviewWebKitViewController: WebKitViewController {
 
         alertController.addCancelActionWithTitle(cancelTitle)
         alertController.addDefaultActionWithTitle(publishTitle) { [unowned self] _ in
-            PostCoordinator.shared.publish(self.post)
+            PostCoordinator.shared.publish(self.post!)
 
             if let editorVC = (self.presentingViewController?.presentingViewController as? EditPostViewController) {
                 editorVC.closeEditor(true, showPostEpilogue: false, from: self)
--- a/WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
@@ -122,7 +122,9 @@ class PreviewWebKitViewController: WebKitViewController {
                 space,
                 shareButton,
                 space,
-                safariButton
+                safariButton,
+                space,
+                previewButton
             ]
         case .hostOnly:
             if canPublish {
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
```diff
@@ -162,7 +172,7 @@ class PreviewWebKitViewController: WebKitViewController {
 
         alertController.addCancelActionWithTitle(cancelTitle)
         alertController.addDefaultActionWithTitle(publishTitle) { [unowned self] _ in
-            PostCoordinator.shared.publish(self.post)
+            PostCoordinator.shared.publish(self.post!)
```
antonis: This is more a suggestion/personal preference. Thought I don't see a case where `self.post!` can be `nil` I would use a `guard` statement to avoid force unwrapping.

WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
```diff
@@ -70,11 +70,19 @@ class PreviewWebKitViewController: WebKitViewController {
         super.init(configuration: configuration)
     }
 
+    @objc override init(configuration: WebViewControllerConfiguration) {
```
antonis, line 73: Since the new constructor adds one more use case we can update the `PreviewWebKitViewControllerTests` and add a test case for the site preview validating the buttons that are expected to be present.

antonis (approved):
Great job @mkevins 👍 
I tested this on an iPhone SE 2020 with iOS 14.2 and everything works as expected.
Overall the code is consistent and LGTM 🎉 
I just added two comment/suggestions.

bjtitus:
@mkevins I did a quick scan of the `PreviewWebKitViewController` references and it looks like every reference should be a post or page preview. So I think it's safe if I'm understanding your question correctly.

I don't think I'm any more sure than you are, though. 😄

WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
mkevins: Good suggestion, it is definitely better with a guard!

WordPress/Classes/ViewRelated/Post/Preview/PreviewWebKitViewController.swift
mkevins: Good idea. I've added a test for this case as well.

mkevins:
> it looks like every reference should be a post or page preview

Thanks for being another set of eyes on this. :smiley:

mkevins:
Thanks Antonis and Brandon for reviewing / testing this one.

PR #15759, opened 2021-02-02
Show post/page updated notification if this is not a first time publish

### LINKED ISSUE #3018: Updating a Published Page/Post notifies "Page Published" instead of updated
**Describe the bug**
Updating a Published Page/Post notifies "Page Published" instead of "Page Updated."

**To Reproduce**
1. Go to a published Page / Post
2. Make a change
3. Save the change
4. The toast says, “Page Published”; however, “Page Updated” is more accurate to the action taken.

**Expected behavior**
Update the toast to say, “Page Updated.” 

**Screenshots**
<kbd><a href="https://user-images.githubusercontent.com/3384451/104965324-81b06480-59ac-11eb-92e2-aeafd4b11445.gif"></a></kbd>

**Additional context**
This was called out in one of our recent Empathy Challenge.

antonis:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3018

## Description

Shows post/page updated notification if this is not a first time publish

## To test:

**Note**: The tests should be executed **both for .com and self-hosted sites**

### Published Page / Post
1. Open a published Page / Post
2. Make a change
3. Save the change
4. Verify that the toast says “Page/Post Updated”

### Draft Page / Post
1. Go to the draft Page / Post list
2. Open it in the editor
3. Select Publish
3. Verify that the toast says “Page/Post Published”

### Draft Page / Post List
1. Open a draft Page / Post for edit
2. Tap on the editor option (...) at the top-right of the screen
3. Select Publish
4. Notice that the toast says “Page/Post Updated” which is wrong

### New Page / Post
1. Create a new Page / Post
2. Publish it
3. Verify that the toast says “Page/Post Published”

PR submission checklist:


antonis:
As noted in the [equivalent Android issue](https://github.com/wordpress-mobile/WordPress-Android/pull/13942#discussion_r569559780) the provided solution is a workaround.

WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
```diff
@@ -102,8 +106,23 @@ struct PostNoticeViewModel {
...
+                return NSLocalizedString("Post updated", comment: "Title of notification displayed when a post has been successfully updated.")
+            }
+        }
+    }
+
+    private func isFirstTimePublish(_ post: AbstractPost) -> Bool {
+        guard let publishDate = post.dateCreated, let modifiedDate = post.dateModified else {
+            return false
+        }
+        let publishDateWithThreshold = Calendar.current.date(byAdding: .second, value: 5, to: publishDate) ?? publishDate
```
antonis: This threshold seemed necessary only for self-hosted sited

mkevins: Is this due to system clocks being slightly out of sync, or something else? Maybe we can add a comment to explain the need for the threshold, wdyt?

mkevins:
> * There is a know issue with the approach taken
> 
> 1. Open a draft Page / Post for edit
> 2. Tap on the editor option (...) at the top-right of the screen
> 3. Select Publish
> 4. Notice that the toast says “Page/Post Updated” which is wrong

Are these the steps for the second test? I assumed this part in the description was meant for the editor:

> Go to the draft Page / Post list

and I encountered the issue described. The other three flows worked as described.

antonis:
Hey @mkevins 👋 

Thank you for looking into this 🙇 

> Are these the steps for the second test? I assumed this part in the description was meant for the editor:
> 
> > Go to the draft Page / Post list
> 
> and I encountered the issue described. The other three flows worked as described.

You are right. I edited the description thinking that I hadn't added a specific case (Draft Page / Post List) for this. The second test fails.
May be I should revisit this. The problem is that many functional changes are needed to make this work.

WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift
antonis: In my understanding this happens because the publishDate is updated when the remote post is received. In any case I updated the implementation to avoid such workarounds.

antonis:
Hey @mkevins 👋 
I updated the implementation basing the message on a new boolean field added to the `BasePost` model. This is set to `true` wherever the post publishing action is triggered and [set again when the remote post is received](https://github.com/wordpress-mobile/WordPress-iOS/pull/15759/files#diff-427e9e088f6cb54084a18b28f4841c71720a2afbaf9fc1b0edb66d6b956bec8aR282). I feel that this approach, though the `isFirstTimePublish` is used as a special flag, is much cleaner. Wdyt?

mkevins (approved):
> I feel that this approach, though the isFirstTimePublish is used as a special flag, is much cleaner. Wdyt?

I agree, this is cleaner. Nice work! I tested all the flows again, and everything is working as expected. LGTM!

PR #15724, opened 2021-01-27
Quick Start: Cancel Edit homepage when homepage is removed

chipsnyder:
Fixes this report from the 16.6 testing

> I got stuck on “Select Homepage” step on the Pages screen, since I had no homepage. I wasn’t able to dismiss the prompt, and it obstructed some of the actions on that page. If you don’t have a Homepage and create one using “Home” layout it will not be recognized by Quick Start logic.

### Remove Homepage
1. Create a site. 
2. Select "Yes, Help Me" in the quick start prompt
3. Remove your homepage. You can do this by:
    - Going to Pages and deleting your homepage (I'll create an issue for this because Web blocks you from deleting your HomePage)
-or-
    - Going to Site Settings > Homepage Settings and select Classic Blog
4. Return to Quick Start 
5. **Expect** Edit Your homepage to be marked as completed

### Abandon flow
1. Create a site. 
2. Select "Yes, Help Me" in the quick start prompt
3. Start the "Edit your homepage" tour
4. Select Pages
5. Select a different page 
6. **Expect** The notice to dismiss 

## PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15760, opened 2021-02-02
Fix a scenario where quick start prompts when marking a step as complete

chipsnyder:
### Description
Fixes an issue introduced in https://github.com/wordpress-mobile/WordPress-iOS/pull/15724 where the quick start menu will pop-up while trying to update completed steps. 

The change here will only attempt to update the quick start steps if the menu option to display quick start is being displayed. This will prevent the unintentional activation of the quick start menu.

### To test:
Using a site that is set with the HomePage to a Classic Blog:
1. Navigate to My Site Page
2. **Expect** the Homepage to not present. 

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+FancyAlerts.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+FancyAlerts.swift
@@ -103,7 +103,7 @@ extension BlogDetailsViewController {
     }
 
     @objc func cancelCompletedToursIfNeeded() {
-        if blog.homepagePageID == nil {
+        if shouldShowQuickStartChecklist() && blog.homepagePageID == nil {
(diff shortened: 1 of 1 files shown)


frosty (approved):
This fixes the issue for me!

However, I feel like ideally we shouldn't be calling this every single time on viewDidAppear. I think it would be best if we could just update it whenever we actually change the homepage?

Of course right now we just want to resolve this issue, but we could reinvestigate when we revamp Quick Start on iOS.

chipsnyder:
Thanks @frosty 

> However, I feel like ideally we shouldn't be calling this every single time on viewDidAppear. I think it would be best if we could just update it whenever we actually change the homepage?

I agree I had started with it on viewDidLoad. However, this resulted in issues where I would make the change on another endpoint and return to the app, and not get the expected updates. So there probably is a better place to trigger this check that I didn't track down.

> but we could reinvestigate when we revamp Quick Start on iOS.

This sounds like a good plan.

PR #15771, opened 2021-02-03
iOS 14 All Time Widget

Gio2018:
Fixes #NA

Adds the "All Time" widget to the iOS 14 Stats widgets (see example below)

<p align=center>

</p>

To test:

- build/run and make sure that you can install both "Today" and "All Time" widgets
- Trigger some stats updates, open the app, check those stats and make sure they are the same in the widgets
- Trigger other stats updates, and without opening the app, make sure they are reflected in the widgets when the timeline updates

tip: the normal timeline update is 1 hour; so either start the review, then have lunch, a workout or a nap and come back, or, if you don't want to wait that long, you can tweak the properties `refreshInterval = 5` and `minElapsedTimeToRefresh = 1`in `SiteListProvider.swift` to speed up things (lower values would not work).

PR submission checklist:


WordPress/Classes/Stores/StatsInsightsStore.swift
```diff
@@ -989,60 +990,86 @@ private extension InsightStoreState {
...
+        var widgetKind = ""
+        if widgetType == HomeWidgetTodayData.self, let stats = stats as? TodayWidgetStats {
 
-        HomeWidgetTodayData.write(items: homeWidgetTodayCache)
-        WidgetCenter.shared.reloadTimelines(ofKind: WPHomeWidgetTodayKind)
-    }
+            widgetKind = WPHomeWidgetTodayKind
 
-    private func initializeHomeWidgetTodayData() -> [Int: HomeWidgetTodayData] {
+            homeWidgetCache[siteID.intValue] = HomeWidgetTodayData(siteID: siteID.intValue,
```
frosty, line 1018: As we discussed on Slack, it would be great if we can avoid needing to know about the concrete data types here, but perhaps something for a future enhancement :)

WordPress/WordPressStatsWidgets/Views/TodayWidgetSmallView.swift
```diff
@@ -11,7 +11,7 @@ struct TodayWidgetSmallView: View {
                 FlexibleCard(axis: .vertical, title: widgetTitle, value: .description(content.siteName))
 
                 Spacer()
-                VerticalCard(title: viewsTitle, value: content.stats.views, largeText: true)
+                VerticalCard(title: viewsTitle, value: (content as? HomeWidgetTodayData)?.stats.views ?? (content as? HomeWidgetAllTimeData)?.stats.views ?? 0, largeText: true)
```
frosty: How about a computed var of `views` to make this a bit cleaner?

frosty (approved):
:shipit: 

As we discussed, there's definitely scope to consolidate the different data types at some point, as we seem to be having to check the type too often – for example in the today widget small view, we shouldn't need to know if it's Today or All Time data, as both have a views property on their stats.

PR #15776, opened 2021-02-03
Jetpack Backup: show progress

leandroalonso:
Part of #15191

This PR shows the backup progress in Activity Log and Backups screen.

#### Creating a backup from the app

1. Go to Activity Log or Backups screen
2. Tap the 3 dots in the first available cell
3. Tap "Download Backup"
4. Tap "Create downloadable file"
5. Tap "Ok notify me"
6. Check that the progress cell is displayed and disappear once the backup is completed

#### Creating a backup from the site

1. Go to your site and trigger a backup there
2. In the app, open Activity Log or Backups Screen
3. The progress bar should be displayed and disappear once the backup is completed

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Stores/ActivityStore.swift
+++ b/WordPress/Classes/Stores/ActivityStore.swift
@@ -28,12 +28,15 @@ enum ActivityAction: Action {
 enum ActivityQuery {
     case activities(site: JetpackSiteRef)
     case restoreStatus(site: JetpackSiteRef)
+    case backupStatus(site: JetpackSiteRef)
 
     var site: JetpackSiteRef {
         switch self {
         case .activities(let site):
             return site
         case .restoreStatus(let site):
+            return site
+        case .backupStatus(let site):
             return site
         }
     }
@@ -51,8 +54,12 @@ struct ActivityStoreState {
     var rewindStatus = [JetpackSiteRef: RewindStatus]()
     var fetchingRewindStatus = [JetpackSiteRef: Bool]()
 
+    var backupStatus = [JetpackSiteRef: JetpackBackup]()
+    var fetchingBackupStatus = [JetpackSiteRef: Bool]()
+
     // This needs to be `fileprivate` because `DelayStateWrapper` is private.
     fileprivate var rewindStatusRetries = [JetpackSiteRef: DelayStateWrapper]()
+    fileprivate var backupStatusRetries = [JetpackSiteRef: DelayStateWrapper]()
 }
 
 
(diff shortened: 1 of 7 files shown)


WordPress/Classes/Stores/ActivityStore.swift
```diff
@@ -451,6 +540,16 @@ private extension ActivityStore {
...
     }
 
+    func backupStatusUpdated(site: JetpackSiteRef, status: JetpackBackup) {
+        state.backupStatus[site] = status
+
+        if let progress = status.progress, progress > 0 {
+            delayedRetryFetchBackupStatus(site: site)
+        } else {
+            // TODO: show download cell
+        }
```
emilylaguna, line 550: TODO coming soon?

WordPress/Classes/ViewRelated/Activity/BaseActivityListViewController.swift
```diff
@@ -499,7 +499,8 @@ extension BaseActivityListViewController: JetpackBackupStatusViewControllerDeleg
             guard let self = self else {
                 return
             }
-            // TODO: fetch backup status
+
+            self.viewModel.refresh()
```
emilylaguna: Should we remove the guard and just do:

```suggestion
            self?.viewModel.refresh()
```

emilylaguna (changes requested):
Heya! Looking good, I left some comments below. Also when I start the backup on .com, the once it's finished the status never updates on the app (see video)


https://user-images.githubusercontent.com/793774/106791452-ec0b0b00-6622-11eb-9f75-d23260db779f.mov

WordPress/Classes/Stores/ActivityStore.swift
leandroalonso: Yes!

WordPress/Classes/ViewRelated/Activity/BaseActivityListViewController.swift
leandroalonso: Done in ed80398f57db02ca9ca2d3862dbb439b436563ca

leandroalonso:
@emilylaguna I'm not being able to reproduce this behavior here. :( (Atomic site, triggering the backup from the web)

momo-ozawa (approved):
@leandroalonso @emilylaguna 
I was able to reproduce the error Emily was seeing exactly once, but that's it. I think I saw it on a JN self-hosted site. I tried again with the same site but wasn't able to reproduce the error.

The PR works as described otherwise, so I'm going to approve it. 
I'm thinking we can keep an eye out for when this happens again, then address it in a future PR.

emilylaguna (approved):
LGTM! 🚀

Issue #15761, opened 2021-02-02
Reader: align options menus on post card and post details


ScoutHarris:
Currently, the options menu on the Reader post cards and in post details are created and handled separately, with much of the code being duplicated, but also not quite the same. This has led to:
- Options being different
- Success messages being different
- Error messages being different

We should align the two so that they use the same code and have the same functionality.

- [x] Show both menus using the same code. [#15778]
- [x] Both menus have the same options. [#15778]
- [x] Verify both menus use the same code for all actions.
- [x] Show success messages for all applicable actions.
  - [x] Follow/unfollow a site. [#15784, #15786, #15800, #15819]
  - [x] Mark as seen/unseen. [#15793]
  - [x] Turn on/off notifications. [#15796]
  - [x] Block site [#15846]
- [x] Show error messages for all applicable actions.
  - [x]  Follow/unfollow a site. [#15800]
  - [x] Turn on/off notifications. [#15796]
  - [x] Block site [#15846]
- [x] Update Discover stream when site is blocked from Discover > post details. [#15845]
- [x] Dismiss post details when site is blocked. [#15845]
- [x] Remove `ReaderPostMenu.swift`. [#15892]

| Post Card | Post Details |
|--------|-------|
| ![card](https://user-images.githubusercontent.com/1816888/106651738-ceba3c00-6551-11eb-9396-a7302b27e184.png) | ![details](https://user-images.githubusercontent.com/1816888/106651793-dd085800-6551-11eb-8817-a3ee81b7fd71.png) |

PR #15778, opened 2021-02-03
Reader: align options menus on post card and post details

ScoutHarris:
Ref #15761

This change does two things:
- The options menu in post details now uses the same as post card. 
- Post details options now has Block and Report when applicable.

Notes: 
- Block and Report where the only differences I found with the two menus. i.e. all the post card and post details options should now match. 
- I have not yet aligned the success/failure messages, so those still differ.
- Following from post details no longer shows a toast message since post card never did. I'll fix this when I align the messaging.
- Just a tip. If you happen to block a site and want to unblock it, https://wordpress.com/me/site-blocks is the place to be.

To test:

Menu change: 
- There should be no visible or functional change. All the menu options should perform the same action as before.

Block / Report:
- Discover > post details now has Block and Report options.
- Discover > You might like > topic > post details now has Report option.
- Following > filter by topic > post details now has Report option.

| Discover post | Discover topic post | Following topic post |
|--------|-------|------|
| discover | discover_topic | followed_topic |

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderShowMenuAction.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderShowMenuAction.swift
@@ -6,7 +6,13 @@ final class ReaderShowMenuAction {
         isLoggedIn = loggedIn
     }
 
-    func execute(with post: ReaderPost, context: NSManagedObjectContext, topic: ReaderSiteTopic? = nil, readerTopic: ReaderAbstractTopic?, anchor: UIView, vc: UIViewController) {
+    func execute(with post: ReaderPost,
+                 context: NSManagedObjectContext,
+                 siteTopic: ReaderSiteTopic? = nil,
+                 readerTopic: ReaderAbstractTopic? = nil,
+                 anchor: UIView,
+                 vc: UIViewController) {
+
         // Create the action sheet
         let alertController = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
         alertController.addCancelActionWithTitle(ReaderPostMenuButtonTitles.cancel, handler: nil)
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderShowMenuAction.swift
```diff
@@ -6,7 +6,13 @@ final class ReaderShowMenuAction {
         isLoggedIn = loggedIn
     }
 
-    func execute(with post: ReaderPost, context: NSManagedObjectContext, topic: ReaderSiteTopic? = nil, readerTopic: ReaderAbstractTopic?, anchor: UIView, vc: UIViewController) {
+    func execute(with post: ReaderPost,
```
aerych, line 9: Yay for clarifying the argument names. :)

aerych (commented):
Howdy @ScoutHarris 👋 

> Post details options now has Block and Report when applicable

I like this change.  Iirc, we excluded the block option from the detail because we figured content that was offensive wouldn't be tapped into.  It's possible that offensive content might not be apparent until clicking into the post, so having the option available makes sense.

>  If you happen to block a site and want to unblock it, https://wordpress.com/me/site-blocks is the place to be.

Thanks :)  I follow the sports tag just for testing site blocks ;)

Question:  If I'm on the post detail and I block a site, should the menu update with the option to unblock that site, or should we automatically dismiss the detail?  Currently the details remains and the menu still says "Block This Site". Tapping the option a second time yields an error.

ScoutHarris:
Hey @aerych .

> It's possible that offensive content might not be apparent until clicking into the post

That. :smile:

>  If I'm on the post detail and I block a site, should the menu update with the option to unblock that site

I'll look into it. 

> should we automatically dismiss the detail?

🤔 I don't think so. That would assume the user meant to tap that option. And since there is no way to unblock sites in the app, it might be better to give them a chance to undo it.

aerych:
> And since there is no way to unblock sites in the app, it might be better to give them a chance to undo it.

Sounds good 👍

ScoutHarris:
Hey @aerych .

> menu still says "Block This Site". Tapping the option a second time yields an error.

The same thing happens on post card, so this just inherited existing behavior. How would you feel about me looking into that separately so I can look at both scenarios and see what's best? And... I'm not quite sure how to fix it yet. 😄

aerych:
Works for me!

aerych (approved):
:shipit: !

ScoutHarris:
Thanks much @aerych !

PR #15779, opened 2021-02-03
Jetpack Section: Scan: Push to the scan view from the Notifications

emilylaguna:
Project: #15190 
---

Tapping on the threat notification will now push to the scan results view. 

### To test:
1. Launch the app
2. Tap the Notifications tab
3. Locate a "Jetpack Scan" notification
4. Tap on the "sites's scan Results" link
5. You should be pushed to the correct sites, scan results views

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationContentRouter.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationContentRouter.swift
@@ -85,6 +85,9 @@ struct NotificationContentRouter {
             try coordinator.displayFollowersWithSiteID(range.siteID, expirationTime: expirationFiveMinutes)
         case .user:
             try coordinator.displayStreamWithSiteID(range.siteID)
+        case .scan:
+            try coordinator.displayScanWithSiteID(range.siteID)
+
         default:
             throw DefaultContentCoordinator.DisplayError.unsupportedType
         }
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationContentRouter.swift
```diff
@@ -85,6 +85,9 @@ struct NotificationContentRouter {
             try coordinator.displayFollowersWithSiteID(range.siteID, expirationTime: expirationFiveMinutes)
         case .user:
             try coordinator.displayStreamWithSiteID(range.siteID)
+        case .scan:
+            try coordinator.displayScanWithSiteID(range.siteID)
+
```
momo-ozawa, line 90: ❓ Do we need to use a feature flag for this?

momo-ozawa (commented):
Works as described!
LGTM with one question.

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationContentRouter.swift
emilylaguna: 🤦 I totally forgot to do that, I went back and added the checking if scan is available on the site with the intention of feature flagging it too but nope forgot.. 

Done in: a27feaea7095dcfc9ce89490acbe0749407df79e

momo-ozawa (approved):
✨

PR #15755, opened 2021-02-01
Jetpack Section: Scan: Adds threat fixing status

emilylaguna:
Project: #15190 
---

Related: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/342

Adds a new state that displays the threats that are actively being fixed. 

### Demo
https://user-images.githubusercontent.com/793774/106516007-7842f400-64a4-11eb-8f8a-46d6bf8caeb8.mov

### To test:

**Note:** You can generate a fixable threat by installing these plugins: 
- [calendar-1.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/5906694/calendar-1.zip)
- [wordpress-seo.zip](https://github.com/wordpress-mobile/WordPress-iOS/files/5906708/wordpress-seo.zip)

1. Launch the app, tap on the My Site, tap on a site with Jetpack Scan
2. Tap Scan
3. Tap the 'Fix All' button
4. 👁️ You should immediately see the fixing status with all the fixable threats displayed
5. The threat description should inform you as to how Jetpack will auto fix it
5. The view will auto refresh after 5 seconds, and should appear to be the same
6. Pull to refresh to verify you're still in the fixing status
7. Wait a few seconds for the threat to be fixed
8. If it's not the last threat you should see it removed from the list
9. If it is the last threat the state should not longer be in the fixing state

### PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15780, opened 2021-02-03
Jetpack Section: Scan: Updates fix / ignore flow

emilylaguna:
Project: #15190
---

This updates the fix / ignore flow to use the new flow introduced in: https://github.com/wordpress-mobile/WordPress-iOS/pull/15755

### Demos
https://user-images.githubusercontent.com/793774/106806752-54171c80-6636-11eb-88f2-534223723e97.mov

https://user-images.githubusercontent.com/793774/106806571-203bf700-6636-11eb-912f-2c5256274e62.mov

### To test:
1. Launch the app
2. Tap My Site
3. Tap a site with Jetpack Scan enabled
4. Tap on a threat to view its details
5. Tap the Fix button
6. You should be brought back to the main view and the state should display as fixing
7. After a few seconds it should be updated to the idle state
8. Tap another threat to view its details
9. Tap the Ignore button
10. You should see an async spinner
11. Once it's ignored you are brought back to the main view and a message informing you it was fixed will be displayed

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/JetpackScanCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Jetpack/Jetpack Scan/JetpackScanCoordinator.swift
@@ -10,9 +10,7 @@ protocol JetpackScanView {
 
     func presentAlert(_ alert: UIAlertController)
 
-    func showFixThreatSuccess(for threat: JetpackScanThreat)
     func showIgnoreThreatSuccess(for threat: JetpackScanThreat)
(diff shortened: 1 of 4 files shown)


momo-ozawa (commented):
Works as described! 

❓ Should we show a success / error notice for the fix threat flow? Wondering because I think the web shows a notice and maybe Android as well?

emilylaguna:
@momo-ozawa I don't think we need it. The "fix all" flow already kind of handles this by displaying which item is being fixed and updates once its done.

leandroalonso (approved):

PR #15774, opened 2021-02-03
Jetpack Section: Add new restore/backup fetch status failed views

momo-ozawa:
Part of #15191 

### Description:
- Added a new view for restore / backup fetch status failed
- Updated to new copy from editorial

Restore | Backup
--- | ---
IMG_1177 | IMG_1176

#### Restore
1. My Site > Activity Log > rewindable cell
2. Start a restore
3. When restore is in progress, turn on airplane mode
4. ✅ New restore fetch status failed view should be displayed

#### Backup
1. My Site > Activity Log > rewindable cell
2. Start a backup
3. When backup is in progress, turn on airplane mode
4. ✅ New backup fetch status failed view should be displayed

### PR submission checklist:

- [ ] I have considered adding unit tests where possible.
- [ ] I have considered adding accessibility improvements for my changes.
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15787, opened 2021-02-04
Jetpack Section: prevent backup status failed screen from being presented multiple times

momo-ozawa:
Part of #15191
Fixes https://github.com/wordpress-mobile/WordPress-iOS/pull/15774#pullrequestreview-582464033

### Description:
- Updated the backup polling interval to 3 seconds to fix an error where the status failed screen was being presented multiple times

_This behavior only seems to happen while testing on the simulator. It seems that this behavior is caused by the polling interval length. A polling interval of 1 second seems a bit short anyway, so I decided to increase the interval to 3 seconds to solve this issue._

#### Should work on both a device and the simulator
1. Create a backup from My Site > Backup 
2. While a create a downloadable backup is in progress, turn off wifi
3. Wait
4. ✅ The status failed screen is presented only once

### Simulator demo:

https://user-images.githubusercontent.com/6711616/106835773-45367580-66db-11eb-8e1e-9ec104b49bb7.mov

### PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Jetpack/Jetpack Restore/Restore Status/Coordinators/JetpackBackupStatusCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Jetpack/Jetpack Restore/Restore Status/Coordinators/JetpackBackupStatusCoordinator.swift
@@ -94,7 +94,7 @@ class JetpackBackupStatusCoordinator {
 extension JetpackBackupStatusCoordinator {
 
     private enum Constants {
-        static let pollingInterval: TimeInterval = 1
+        static let pollingInterval: TimeInterval = 3
         static let maxRetryCount: Int = 3
     }
 }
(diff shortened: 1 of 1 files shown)


leandroalonso (changes requested):
@momo-ozawa the issue is still happening here. But now the screen is presented every 3 secs.

What happens is that for every request that is made, the error callback is being called over and over.

This is an edge case, but I think it might happen in the device (with a very poor connection, for example). Perhaps we can check if the error screen is already presented before presenting it?

emilylaguna (approved):
Oops ignore me.

emilylaguna:
@leandroalonso I was able to reproduce this issue on my end so I added a flag to restrict only 1 refresh request to be triggered at the same time. This fixed it for me.

leandroalonso (approved):

PR #15770, opened 2021-02-02
Update Tracks to 0.8.0

jkmassel:
Upgrades Sentry and cleans up associated code.

To test:
- Download the installable build from this PR onto a test device
- Go to Me > App Settings > Debug, and tap "Send Log Message", then "Send Test Crash". Relaunch the app.
- Go to https://sentry.io/organizations/a8c/discover/results/ and enter `user.email:$your_email`, and note that the events show up (might take up to 5 minutes)

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15789, opened 2021-02-04
AB testing: implements an A/A experiment

leandroalonso:
Ready for review after #15770 is merged

Automattic Tracks: https://github.com/Automattic/Automattic-Tracks-iOS/pull/172

This PR implements the A/A experiment to validate that our ExPlat implementation is not biased.

### To test

1. Add a breakpoint in `ExPlatService.swift:70`
2. Add a breakpoint in `WPTabBarController+Swift.swift:89`
3. Run the app
4. The first breakpoint in `ExPlatService` should be hit
5. Tap the Reader tab
6. The second breakpoint should be hit and the property `variation` should be `control`

### Closing and reopening the app

1. Stop the app
2. Run it again
3. The first breakpoint in `ExPlatService` should be hit

### Coming back from the background

1. Put the app in the background
2. Put the app in the foreground
3. The first breakpoint in `ExPlatService` should not be hit

PR submission checklist:


leandroalonso:
@emilylaguna can I bother you to review this one?

leandroalonso:
@renanferrari JFYI, here's the iOS PR for the A/A experiment.

WordPress/Classes/ViewRelated/System/WPTabBarController+Swift.swift
```diff
@@ -84,7 +84,13 @@ extension WPTabBarController {
                 return false
         }
 
-        WPAppAnalytics.track(stat)
+        if tabType == .reader {
+            let variation = ABTest.biasTest.variation == .control ? "control" : "treatment"
+            WPAppAnalytics.track(stat, withProperties: ["experiment_id": 20086, "variation": variation])
```
aaronyan: The event with extra props should only fire when the API returns a variation. If it fails to retrieve an experiment assignment, no extra props should be returned even if the `tabType == .reader`.

aaronyan (commented):
Does everyone that open an app get assigned an experiment variation? If so, that's okay, I just want to understand how this works. :)

leandroalonso:
@aaronyan It works like this:

* If there's any active experiment (aka an [enum added here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15789/files#diff-967be67716e482e146e14315ec4b853f60884754d0f9f01e8516e6de2aa2d94aR5)), the request is made
* The returned experiments values are saved
* When the variation for an experiment is consulted, we either return "treatment" (if the API returned treatment) or "control" (in any other case: missing value, null, etc - [as per suggested here](https://github.com/Automattic/Automattic-Tracks-iOS/pull/155#discussion_r560243072))

WordPress/Classes/ViewRelated/System/WPTabBarController+Swift.swift
leandroalonso: @aaronyan Knowing if an experiment "exists" is something that we should worry about or is it specific for this case?

We used to have an `unknown` property for events that didn't exist (or expired), but we changed the implementation to always return `control` in this case ([see here](https://github.com/Automattic/Automattic-Tracks-iOS/pull/155#discussion_r560243072)).

jkmassel:
👋 We're freezing `16.7` today, so this PR is being bumped to `16.8`. If you need this to be part of the `16.7` release, please merge it into the `release/16.7` branch and DM me – I'll be happy to cut a new beta release!

WordPress/Classes/ViewRelated/System/WPTabBarController+Swift.swift
aaronyan: Yeah, we don't need to know this anymore. We've decided to drop the eventprop requirement, so I'm going to mark this thread as resolved!

aaronyan:
> - If there's any active experiment (aka an enum added here), the request is made
> - The returned experiments values are saved
> - When the variation for an experiment is consulted, we either return "treatment" (if the API returned treatment) or "control" (in any other case: missing value, null, etc - as per suggested here)

This looks great!

emilylaguna (approved):
LGTM. Works as described.

leandroalonso:
I tested the scenarios for the A/A testing as described in `pbxNRc-Fx-p2#comment-1656`

PR #15791, opened 2021-02-04
Jetpack Section: add events for backup / restore

leandroalonso:
Part of #15191

### To test

- [ ] Tap "Restore" from Activity Log and Backups screen and make sure the events are fired with the correct `source` property
- [ ] Restore your site to any point and make sure the restore confirmed event is fired with the selected types
- [ ] When restoring, tap the "notify me" button and make sure the correlated event is fired
- [ ] Restore your site and turn off Wi-fi, make sure an event is fired with the `cause` property
- [ ] Tap "Download Backup" from Activity Log and Backups screen and make sure the events are fired with the correct `source` property
- [ ] Confirm a backup and make sure an event is fired with the selected types
- [ ] Turn off wi-fi while a backup is being generated make sure an event is fired with the `cause` property
- [ ] Tap the "notify me" button wile the backup is being generated and make sure the event is fired
- [ ] When the backup is done, tap the download button and make sure an event is fired
- [ ] Tap the share button and make sure the event is fired

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -106,6 +106,7 @@ import Foundation
     case backupFilterbarRangeButtonTapped
     case backupFilterbarSelectRange
     case backupFilterbarResetRange
+    case restoreOpened
 
     // Comments
     case commentViewed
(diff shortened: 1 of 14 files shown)


emilylaguna (approved):
Followed the steps and confirmed each of them triggered the correct event and properties!

🚀

PR #15795, opened 2021-02-04
Fix CocoaPods warning

jkmassel:
Fixes a CocoaPods warning:

```
[!] The `WordPress [Debug]` target overrides the `SWIFT_INCLUDE_PATHS` build setting defined in `../Pods/Target Support Files/Pods-WordPress/Pods-WordPress.debug.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.
```

To test:
- On the `develop` branch, run `bundle && bundle exec pod install` – note that the error appears
- Switch to this branch and run the same command. Note that the error does not appear.

PR submission checklist:

### DIFF
--- a/WordPress/WordPress.xcodeproj/project.pbxproj
+++ b/WordPress/WordPress.xcodeproj/project.pbxproj
@@ -15808,7 +15808,6 @@
 				PRODUCT_BUNDLE_IDENTIFIER = org.wordpress;
 				PRODUCT_NAME = WordPress;
 				PROVISIONING_PROFILE_SPECIFIER = "WPiOS Development Profile";
-				SWIFT_INCLUDE_PATHS = "";
 				SWIFT_OBJC_BRIDGING_HEADER = "Classes/System/WordPress-Bridging-Header.h";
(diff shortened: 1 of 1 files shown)


aerych (approved):
Tested.  No more warnings. :) 
:shipit: !

AliSoftware (approved):
FWIW, this will work, but it's usually even better/cleaner to completely *delete* the `SWIFT_INCLUDE_PATHS` build settings (select the build setting at the appropriate level in the the project/target, then hit the Delete key) than using `$(inherited)` (which feels a bit like overriding a method just to only call `super` and nothing else 😂 )

But that's not really a big deal, and using `$(inherited)` will work too there, so 👍

jkmassel:
> which feels a bit like overriding a method just to only call super and nothing else

Wait, you mean you _don't_ do this when you subclass something? :trollface: 

I agree that that's a better solution, so I've adjusted the PR to work that way instead – same diff, so don't think it requires a re-review. I'll land it when 🟢

PR #15820, opened 2021-02-05
Remove Swift-based Core Data Stack

### LINKED ISSUE #15818: Remove existing Core Data Stack split
We should align the Core Data Stack between development and production versions prior to starting the project

jkmassel:
Removes the Swift-based Core Data Stack currently only used in development builds. Production builds haven't been using this stack to the point, so there should be no adverse affects caused by the change.

We shouldn't keep both this and the `ContextManager` object (especially given #15830 and #15831, and the fact that this one was experiencing some bugs in testing), so I'm favouring the one that's currently used in production.

Fixes #15818 

To test:
- Ensure tests pass

PR submission checklist:


leandroalonso (approved):
🚢

jkmassel:
Thanks @leandroalonso!!

PR #15824, opened 2021-02-06
Display loader when site title is updating

### LINKED ISSUE #15717: When changing Site Title. there is no indication that something is updating
### Expected behavior
When I change my site title, if it doesn't happen immediately, there should be indication that it is updating. 

### Actual behavior
There is no indication. 
While checking Android, the change was instant so no indication of title changing seems necessary. 
On iOS, it took less than 2 seconds but wasn't instant. 
I have seen it take about 5 seconds before on a slower connection with no indication that it's changing.

on the My Site page:
1. Tap the title of your site under the icon
2. A Site Title modal will open where you can edit.
3. Make some change and Save

yaelirub:
Fixes #15717 

To test:
on the My Site page:

1. Tap the title of your site under the icon
2. A Site Title modal will open where you can edit.
3. Make some change and Save
4. See loader 
5. See loader removed when stopped loading 
6. See title updated 

https://user-images.githubusercontent.com/1335657/107131474-6b027c80-688b-11eb-98bb-d21833a9bc69.mov

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
+++ b/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
@@ -7,6 +7,7 @@ protocol Spotlightable: UIView {
 class SpotlightableButton: UIButton, Spotlightable {
 
     var spotlight: QuickStartSpotlightView?
+    var originalTtile: String?
 
     var shouldShowSpotlight: Bool {
         get {
@@ -24,12 +25,14 @@ class SpotlightableButton: UIButton, Spotlightable {
     }
 
     func startLoading() {
+        originalTtile = titleLabel?.text
         setTitle("", for: .normal)
         activityIndicator.startAnimating()
     }
 
     func stopLoading() {
         activityIndicator.stopAnimating()
+        setTitle(originalTtile, for: .normal)
     }
 
     private lazy var activityIndicator: UIActivityIndicatorView = {
--- a/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
+++ b/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
@@ -23,6 +23,28 @@ class SpotlightableButton: UIButton, Spotlightable {
         }
     }
 
+    func startLoading() {
+        setTitle("", for: .normal)
+        activityIndicator.startAnimating()
+    }
+    
+    func stopLoading() {
+        activityIndicator.stopAnimating()
+    }
+
+    private lazy var activityIndicator: UIActivityIndicatorView = {
+        let activityIndicator = UIActivityIndicatorView(style: .large)
+        activityIndicator.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(activityIndicator)
+        
+        NSLayoutConstraint.activate([
+            activityIndicator.centerYAnchor.constraint(equalTo: self.centerYAnchor),
+            activityIndicator.centerXAnchor.constraint(equalTo: self.centerXAnchor)
+        ])
+        
+        return activityIndicator
+    }()
+
     private func setupSpotlight() {
         spotlight?.removeFromSuperview()
 
--- a/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
+++ b/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
@@ -32,7 +32,7 @@ class SpotlightableButton: UIButton, Spotlightable {
 
     func stopLoading() {
         activityIndicator.stopAnimating()
-        setTitle(originalTtile, for: .normal)
+        setTitle(originalTitle, for: .normal)
     }
 
     private lazy var activityIndicator: UIActivityIndicatorView = {
--- a/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
+++ b/WordPress/Classes/Utility/Spotlight/Spotlightable.swift
@@ -25,7 +25,7 @@ class SpotlightableButton: UIButton, Spotlightable {
     }
 
     func startLoading() {
-        originalTtile = titleLabel?.text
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/BlogDetailHeaderView.swift
```diff
@@ -92,6 +92,10 @@ class BlogDetailHeaderView: UIView, BlogDetailHeader {
         siteIconView.spotlightIsShown = QuickStartTourGuide.shared.isCurrentElement(.siteIcon)
     }
 
+    @objc func updateLoadingTitle(isLoading: Bool) {
```
frosty: How about `setTitleLoading(_ isLoading: Bool)`? (Would prevent the duplication of 'loading' when calling it, and I think `setTitleLoading` may be a little clearer than the existing about what it's doing?)

WordPress/Classes/Utility/Spotlight/Spotlightable.swift
```diff
@@ -23,6 +23,28 @@ class SpotlightableButton: UIButton, Spotlightable {
         }
     }
 
+    func startLoading() {
+        setTitle("", for: .normal)
+        activityIndicator.startAnimating()
+    }
+
+    func stopLoading() {
+        activityIndicator.stopAnimating()
```
frosty, line 34: Should we restore the title when we stop loading?

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
```diff
@@ -77,6 +78,7 @@ extension BlogDetailsViewController {
                                 feedbackType: .success)
             ActionDispatcher.global.dispatch(NoticeAction.post(notice))
 
+            self?.headerView.updateLoadingTitle(isLoading: false)
             self?.headerView.refreshSiteTitle()
         }, failure: { [weak self] error in
             self?.blog.settings?.name = existingBlogTitle
```
frosty, line 84: The header view is currently not updated if there's a failure with the request to update the settings, which results it showing the loading indicator indefinitely. We should also stop loading here and restore the original title.

frosty (commented):
Mostly looking good! 🙂 Just one or two small comments, mainly handling the error case.

There's also a kind-of-edge-case which could happen if the user has a poor connection – if you navigate to another site and then back while the request is ongoing, when you come back there's nothing to indicate that it's happening. But I think it's fine to leave that case.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
yaelirub: Absolutely! Thank you!

WordPress/Classes/ViewRelated/Blog/Blog Details/Detail Header/BlogDetailHeaderView.swift
yaelirub: Done in 1e49f70d7941a7b87ab0c07795ec71643224fcac

WordPress/Classes/Utility/Spotlight/Spotlightable.swift
yaelirub: Done in 1e49f70d7941a7b87ab0c07795ec71643224fcac

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
yaelirub: Done in 1e49f70d7941a7b87ab0c07795ec71643224fcac

yaelirub:
@frosty , thank you for reviewing! Ready for another look

WordPress/Classes/Utility/Spotlight/Spotlightable.swift
```diff
@@ -7,6 +7,7 @@ protocol Spotlightable: UIView {
 class SpotlightableButton: UIButton, Spotlightable {
 
     var spotlight: QuickStartSpotlightView?
+    var originalTtile: String?
```
frosty: Typo:

```suggestion
    var originalTitle: String?
```

WordPress/Classes/Utility/Spotlight/Spotlightable.swift
```diff
@@ -23,6 +24,30 @@ class SpotlightableButton: UIButton, Spotlightable {
         }
     }
 
+    func startLoading() {
+        originalTtile = titleLabel?.text
```
frosty:
```suggestion
        originalTitle = titleLabel?.text
```

WordPress/Classes/Utility/Spotlight/Spotlightable.swift
```diff
@@ -23,6 +24,30 @@ class SpotlightableButton: UIButton, Spotlightable {
...
+        setTitle("", for: .normal)
+        activityIndicator.startAnimating()
+    }
+
+    func stopLoading() {
+        activityIndicator.stopAnimating()
+        setTitle(originalTtile, for: .normal)
```
frosty:
```suggestion
        setTitle(originalTitle, for: .normal)
```

frosty (commented):
Thanks for making the changes! Just one minor typo – I've added some suggestions so you should be able to just amend it via Github.

Other than that, looking great 👍

yaelirub:
Good catch, @frosty! I committed your suggestions

frosty (approved):
Looks great! ✅

PR #15828, opened 2021-02-08
Merge 16.6 final into trunk

jkmassel:
Merges all changes from the `16.6` release series into `trunk`.


mokagio (approved):
Approving to speed up the process, but I wonder why GitHub says this branch is out of sync with `trunk` 🤔

Issue #15821, opened 2021-02-05
Introduce Updated Core Data Write API


jkmassel:
Introduce an API that abstracts away the complexity of saving the Core Data Store

PR #15837, opened 2021-02-08
Remove obtainPermanentIDForObject

jkmassel:
A little bit of cleanup on the way to #15821.

This PR removes the `obtainPermanentIDForObject` method from the `ContextManager` for a few reasons:
1. It's only called from a single place.
2. When it's called, the return value is currently ignored.
3. The Core Data Stack shouldn't be responsible for this – if anything, this should be an extension of `NSManagedObjectContext` (though because of the single use, I'm not sure it's worth converting).

For safety, I've added an `NSAssert` to the the `PostService`, but I could see a strong argument for making it crash in production as well – if a post isn't set up correctly, IMHO it's better to crash than potentially lose user data.

To test:
- Ensure tests pass
- Ensure making an empty post/page works correctly in a `debug` build.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -55,8 +55,9 @@ const NSUInteger PostServiceDefaultNumberToSync = 40;
     post.postFormat = blog.settings.defaultPostFormat;
     post.postType = Post.typeDefaultIdentifier;
 
-    [[ContextManager sharedInstance] obtainPermanentIDForObject:post];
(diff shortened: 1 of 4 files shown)


jleandroperez (approved):
✅ Changes look good
✅ Unit Tests pass
✅ Verified: Empty Post / Page still work great!

:shipit: sir!

jkmassel:
Thanks @jleandroperez!! 🙇

PR #15839, opened 2021-02-09
File Block Media Source label improvement

jd-alexander:
## Description

This change simply updates the label on the File Block's `media-upload` options sheet. Previously the label was `Choose from device` but as stated by @ceyhun in the [comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/15839#issuecomment-778340310) this label is used to open the Media Picker UI in other blocks, eg. Image block. 

This change aligns the behavior of this block with the other media blocks by using the `Other apps` label that represents the action being committed. 

## Testing 

1. Create a new post. 
2. Add a File Block. 
3. Notice that the verbiage is now `Other apps` instead of `Choose from device.` when the media options are shown.

Before | After 
--------|-------
  <kbd></kbd>   |  <kbd></kbd>

## Regression Notes

There are no areas of impact based on my understanding of this change. A string is being removed and another string is being utilized in multiple places.

I relied on the smoke testing I did of the app with this PR. 

I ran the optional tests to verify they will pass. 


jd-alexander:
For more clarity here, I am seeing that there's an option for "Other apps" and "Choose from device" in the Video, and Image block on iOS but on the File Block the "Choose from device" option triggers the "Other Apps" behavior/opening the File selector instead of staying in-app like the Image and Video blocks.

ceyhun:
I think maybe we should update the File block to show "Other Apps" instead. I had in my mind this mapping:
Other Apps -> Opens File Picker UI -> You can pick any file or pick an audio file
Choose from device -> Opens Media Picker UI -> You can pick an image or a video

What do you think @iamthomasbishop?

jd-alexander:
> I think maybe we should update the File block to show "Other Apps" instead. I had in my mind this mapping:
Other Apps -> Opens File Picker UI -> You can pick any file or pick an audio file
Choose from device -> Opens Media Picker UI -> You can pick an image or a video

Thanks for making this connection @ceyhun I like this approach. I am happy I brought it up 😄

jd-alexander:
Ignore the name of the branch. This PR was updating an Audio block-related label incorrectly at first due to using the File Block as the source of truth when the File Block was the label that needed updating.

WordPress/Classes/ViewRelated/Media/StockPhotos/StockPhotosStrings.swift
```diff
@@ -6,11 +6,7 @@ extension String {
     }
 
     static var otherApps: String {
-        return NSLocalizedString("Other Apps", comment: "Menu option used for adding media from other applications.")
-    }
-
-    static var allFiles: String {
-        return NSLocalizedString("Choose from device", comment: "Menu option used for adding media from other applications.")
+        return NSLocalizedString("Other apps", comment: "Menu option used for adding media from other applications.")
```
fluiddot: I noticed that the text changed from `Other Apps` to `Other apps` (note the downcase of `apps`), is it correct?

fluiddot (approved):
LGTM 🎊 ! (just added a minor comment regarding the text).

_Tested on iPhone 11 (iOS 14.2)._

WordPress/Classes/ViewRelated/Media/StockPhotos/StockPhotosStrings.swift
jd-alexander: Thanks for spotting that. I will utilize the upper case variant :)

Issue #15829, opened 2021-02-08
Remove `newMainContextChildContext` and other unneeded API from `CoreDataStack`


jkmassel:
Our current model allows two ways to accomplish the same task – making a new write context. These can be accessed via `newMainContextChildContext` and `newDerivedContext`. `newMainContextChildContext `creates contexts that are writable from the main thread (ie – they have `mainQueueConcurrencyType`), while `newDerivedContext` creates background contexts (ie – `privateQueueConcurrencyType`).

We will remove `newMainContextChildContext`, which will align our data store with how `NSPersistentContainer` works. This will allow us to finish implementing it.

PR #15845, opened 2021-02-09
Reader > Discover: update UI when site is blocked

ScoutHarris:
Ref: #15761 

This updates relevant UI when a site is blocked from Discover > post card and post details.
- Post card: the Discover stream is updated to automatically remove blocked posts.
- Post details:
  - The Discover stream is updated to automatically remove blocked posts.
  - The post details is automatically dismissed.

Notes:

These changes relate to a conversation on https://github.com/wordpress-mobile/WordPress-iOS/pull/15778. 

> menu still says "Block This Site". Tapping the option a second time yields an error.

I realized the Discover stream was not refreshed to remove blocked site posts, and probably should have been to mimic the behavior when blocking via tag filter. This has been fixed, so you no longer have the opportunity to re-block a site.

> should we automatically dismiss the detail?

Once I realized the stream should automatically remove posts, this made more sense. As noted above, this now happens.

To test:

---
- Go to Reader > Discover.
  - On a post card, `Block This Site`.
  - Verify the stream is refreshed, and the post card is removed.

https://user-images.githubusercontent.com/1816888/107437659-91732280-6aec-11eb-9766-13387077461f.mp4

---
- Tap a Discover post card.
  -  In post details, `Block This Site`.
  - Verify the post details is dismissed.
  - Verify the stream is refreshed, and the post card is removed.

https://user-images.githubusercontent.com/1816888/107437855-da2adb80-6aec-11eb-8b6c-fc336bbfb9c4.mp4

---
PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4225, opened 2015-08-28
Issues/4093 reader list refresh

aerych:
Refs #4093 

This PR adds the refreshed reader post list to the app, but does not yet replace the current reader post list pending additional work. 

The new `ReaderStreamViewController` ports the current `ReaderPostsViewController` to Swift. Other changes include:
- Adds header views for site, tag, and some list topics. 
- Consolidates the functionality of the (now deprecated) `ReaderBrowseSiteViewController` 
- Adopts storyboards an xibs for layout
- Improvements to the refresh mechanism. 
- Convenience methods for identifying a topic type.  These are subject to change when refreshing the topic model
- Allows sharing posts from the list
- Updates the (now deprecated) `ReaderPostsViewController` to reconcile other changes made to the `ReaderPostService` and `WPTableViewHandler`.

#### Please note: 
- Card layout is pending a refresh to be full screen on the iPhone. 
- Some features in the new header views (e.g. following, some data fields) are not implemented pending improvements to the `ReaderTopic` model in a later.

Of particular interest is this PR changes the `NSManagedObjectContext` used for fetching and displaying content in the new list, and in the old list, to be a child context of our main context, but also running on the main queue.  Syncing is still performed on a private queue, derived context. This is done for better separation of concerns and encapsulation of the mechanism to refresh the table view while preserving the visible content.  cc @astralbodies

#### General:
- Does the new list load?
- Does pull to refresh function?
- Does load more function? 

#### Headers: 
- Freshly Pressed, Posts I Like, and Following feeds should not show a header. 
- Tags should show a tag header (following is not implemented)
- Browsing a site should show a site header. Some content is a placeholder. Following is not enabled. 
- Custom lists should show a list header.  Some content is a placeholder.

#### Cell Menus:
- You should not see an option to block a site while browsing a site, or browsing Following.
- You should be able to block a site, and undo a block.  Pull to refresh should clear an undo cell. 
- You should be able to share a post via the share menu option. 

#### Cell Features: 
- You should be able to like or unlike a post by tapping the like button in a cell
- You should be able to view comments by tapping on the comment button in a cell
- You should be able to view a post detail by tapping on a cell

#### Site Browsing:
- Tapping the site name or blavatar should browse a site. 
- When browsing a site, the site name should be disabled. Tapping the site name or blavatar should be like any other tap on a cell. 

#### Old list:
Pull to refresh should still work. 
Infinite scroll / load more should still work. 
You should still be able to like or unlike a post from a cell

### How to test
[...]

PR #15849, opened 2021-02-10
Remove newMainContextChildContext in ReaderStreamViewController

jkmassel:
Works towards #15829

**To test:**
- Launch the app, use the reader stream
- Navigate from the reader stream to My Site to test dealloc behaviour
- Block a site from within the feed, and note that it disappears. (I'm not sure how you would unblock a site from the feed given that it disappears, but given that it works exactly the same I have no reason to believe it wouldn't work if it's possible)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
@@ -12,8 +12,6 @@ import WordPressFlux
 ///   - This controller uses MULTIPLE NSManagedObjectContexts to manage syncing and state.
 ///     - The topic exists in the main context
 ///     - Syncing is performed on a derived (background) context.
-///     - Content is fetched on a child context of the main context.  This allows
-///         new content to be synced without interrupting the UI until desired.
 ///   - Row heights are auto-calculated via UITableViewAutomaticDimension and estimated heights
 ///         are cached via willDisplayCell.
 ///
@@ -62,6 +60,12 @@ import WordPressFlux
         return ReaderPostService(managedObjectContext: syncContext)
     }()
 
+    /// An alias for the apps's main context – temporarily replaces  `newMainContextChildContext` until we have `NSPersistentContainer` support
+    ///
+    private var viewContext: NSManagedObjectContext {
+        ContextManager.sharedInstance().mainContext
+    }
+
     private(set) lazy var footerView: PostListFooterView = {
         return tableConfiguration.footer()
     }()
@@ -100,10 +104,6 @@ import WordPressFlux
     private let cellConfiguration = ReaderCellConfiguration()
     /// Actions
     private var postCellActions: ReaderPostCellActions?
-
-    /// Used for fetching content.
-    private lazy var displayContext: NSManagedObjectContext = ContextManager.sharedInstance().newMainContextChildContext()
-
 
     private var siteID: NSNumber? {
         didSet {
@@ -358,10 +358,6 @@ import WordPressFlux
 
         dismissNoNetworkAlert()
 
-        // We want to listen for any changes (following, liked) in a post detail so we can refresh the child context.
-        let mainContext = ContextManager.sharedInstance().mainContext
-        NotificationCenter.default.addObserver(self, selector: #selector(ReaderStreamViewController.handleContextDidSaveNotification(_:)), name: NSNotification.Name.NSManagedObjectContextDidSave, object: mainContext)
-
         ReaderTracker.shared.stop(.filteredList)
     }
 
@@ -576,9 +572,6 @@ import WordPressFlux
         // Rather than repeatedly creating a service to check if the user is logged in, cache it here.
         isLoggedIn = AccountHelper.isDotcomAvailable()
 
-        // Reset our display context to ensure its current.
-        managedObjectContext().reset()
-
         configureTitleForTopic()
         configureShareButtonIfNeeded()
         hideResultsStatus()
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -12,8 +12,6 @@ import WordPressFlux
 ///   - This controller uses MULTIPLE NSManagedObjectContexts to manage syncing and state.
 ///     - The topic exists in the main context
 ///     - Syncing is performed on a derived (background) context.
-///     - Content is fetched on a child context of the main context.  This allows
```
jkmassel, line 15: I've removed the `newMainContextChildContext` from this class, replacing it with the stack's main context for now.

I've used this on my iPhone 6s+ and not noted any performance hiccups from it – I suspect that's because this approach may not work anymore? If I'm following this correctly, it looks like this class was listening for when the `mainContext` is saved, and when that happened, it would update the child context. But then it's unclear to me what would cause the `mainContext` to be saved. AFAICT, that would most likely happen in the `ContextManager` as a result of an `NSManagedObjectContextDidSave` notification. But if that's the case, wouldn't changes in the sync context propagate directly to the child context, just with the middle step of the `mainContext` in the middle?

If that's the case, it'd explain why there was no performance difference removing the child context, but I'm not entirely certain I'm understanding this correctly.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -62,6 +60,12 @@ import WordPressFlux
         return ReaderPostService(managedObjectContext: syncContext)
     }()
 
+    /// An alias for the apps's main context – temporarily replaces  `newMainContextChildContext` until we have `NSPersistentContainer` support
+    ///
+    private var viewContext: NSManagedObjectContext {
```
jkmassel, line 65: This is a shortcut to ensure that it breaks properly once we migrate this class to using a DI-ed Core Data Stack – if we referenced the `sharedInstance` directly it'd be easy to miss an instance.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -576,9 +572,6 @@ import WordPressFlux
         // Rather than repeatedly creating a service to check if the user is logged in, cache it here.
         isLoggedIn = AccountHelper.isDotcomAvailable()
 
-        // Reset our display context to ensure its current.
```
jkmassel, line 579: Given that we'd be operating directly on the main context now, it's IMHO important not to destroy its cache as navigating back to "My Sites" could be quite slow in that case. I suspect that perhaps this was due to odd `NSManagedObjectContext` cache invalidation issues or some other bug?

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -1415,7 +1404,6 @@ extension ReaderStreamViewController: WPTableViewHandlerDelegate {
 
     func tableViewHandlerWillRefreshTableViewPreservingOffset(_ tableViewHandler: WPTableViewHandler) {
         // Reload the table view to reflect new content.
-        managedObjectContext().reset()
```
jkmassel, line 1395: Same thing here as the other `reset` call

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -1392,7 +1380,8 @@ extension ReaderStreamViewController: WPTableViewHandlerDelegate {
     // MARK: - Fetched Results Related
 
     func managedObjectContext() -> NSManagedObjectContext {
-        return displayContext
+        assert(Thread.isMainThread, "WPTableViewHandler should use Core Data on the main thread")
```
jkmassel, line 1360: Quick safety check – I don't _think_ `WPTableViewHandler` is going to act up, but if it does I'd love to catch it in action while working on this project.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
aerych: I'm guessing a lot has changed since that mechanism was first introduced [six years ago](https://github.com/wordpress-mobile/WordPress-iOS/pull/4225). I'm struggling to remember the exact problem it was supposed to solve.  As long as we're not seeing any obvious visual glitch related to syncing and reloading the list (while scrolling) it's probably fine. :)

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
aerych: Nice

aerych (approved):
Hola @jkmassel 👋 

I took the branch for a spin and spent some time testing different flows and features.  I didn't see anything that was negatively impacted by the change. Actually, when the list refreshes there's a nice transition instead of an abrupt change now. 

Something (probably unrelated) I noticed was different since the last time I looked at the readed.  When returning to a topic with cached posts the reader fetches newer posts automatically (no user action required).  Previously in the reader, when the new posts were loaded, the list would update in a way that preserved the appearance of the list. The visible posts would remain in their place and newer content would be inserted above.    This no longer seem to be the case -- older content is now pushed down.  I see it in the current beta so maybe it is intentional.  Either way, it's unrelated to the changes here. Just making a note.

:shipit:  from me.

startuptester (approved):
Used the Reader: 
- Spent most of my time in refreshing and scrolling due to the identified potential for failure there
- Liked, commented, saved, followed, blocked, unfollowed
- Deleted published post and viewed update
- Published/edited post and viewed update
- Manage (followed/unfollowed sites and topics)

The only issues I ran into were not due to this change:
1. Blocking a site does not remove it from the feed until after refresh. This causes a double block action that fails. I will log this. @ScoutHarris just fixed this yesterday in #15845 
2. Blocking only appears in the post card menu in the Discover tab, not in the Following tab, not sure if that's intentional? @ScoutHarris is working on this in #15761

ScoutHarris:
Thanks @startuptester .

> Blocking only appears in the post card menu in the Discover tab, not in the Following tab, not sure if that's intentional?

This is intentional (and AFAIK has always been this way). You cannot block a site you are following.

PR #15866, opened 2021-02-11
Discover posts sorting bottom sheet

bozidarsevo:

To test:

Open Reader tab, switch to Discover screen, tap sorting option button on top of the posts, switch sorting option and see if the posts reload properly depending on the sorting option.

https://user-images.githubusercontent.com/6242034/107681320-303f8080-6c9f-11eb-99fe-48bd644fb9bd.mov

Things left to check/do:
- [ ] if used strings are localized
- [ ] use proper icons
- [ ] set accessibility correctly
- [ ] check if the styling is correct
- [ ] tracking support

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -80,6 +80,9 @@ import Foundation
     case readerDiscoverContentPresented
     case readerPostMarkSeen
     case readerPostMarkUnseen
+    case readerDiscoverSortingOptionButtonTapped
+    case readerDiscoverSortingOptionPopularitySelected
+    case readerDiscoverSortingOptionDateSelected
 
     // What's New - Feature announcements
     case featureAnnouncementShown
@@ -277,6 +280,12 @@ import Foundation
             return "reader_mark_as_seen"
         case .readerPostMarkUnseen:
             return "reader_mark_as_unseen"
+        case .readerDiscoverSortingOptionButtonTapped:
+            return "reader_discover_sorting_option_button_tapped"
+        case .readerDiscoverSortingOptionDateSelected:
+            return "reader_discover_sorting_option_date_selected"
+        case .readerDiscoverSortingOptionPopularitySelected:
+            return "reader_discover_sorting_option_popularity_selected"
 
         // What's New - Feature announcements
         case .featureAnnouncementShown:
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -81,8 +81,7 @@ import Foundation
     case readerPostMarkSeen
     case readerPostMarkUnseen
     case readerDiscoverSortingOptionButtonTapped
-    case readerDiscoverSortingOptionPopularitySelected
-    case readerDiscoverSortingOptionDateSelected
+    case readerDiscoverSortingOptionSelected
 
     // What's New - Feature announcements
     case featureAnnouncementShown
@@ -282,10 +281,8 @@ import Foundation
             return "reader_mark_as_unseen"
         case .readerDiscoverSortingOptionButtonTapped:
             return "reader_discover_sorting_option_button_tapped"
-        case .readerDiscoverSortingOptionDateSelected:
-            return "reader_discover_sorting_option_date_selected"
-        case .readerDiscoverSortingOptionPopularitySelected:
-            return "reader_discover_sorting_option_popularity_selected"
+        case .readerDiscoverSortingOptionSelected:
+            return "reader_discover_sorting_option_selected"
 
         // What's New - Feature announcements
         case .featureAnnouncementShown:
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -15,6 +15,7 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
     private lazy var sortingButton: ReaderSortingOptionButton = {
         let view = ReaderSortingOptionButton()
         view.addTarget(self, action: #selector(didTapSortingButton), for: .touchUpInside)
+        view.accessibilityHint = NSLocalizedString("Tap to change sorting option", comment: "Accessibility hint for sorting option button.")
         return view
     }()
 
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -140,7 +140,7 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         let availableSortingOptions: [ReaderSortingOption] = [.popularity, .date]
         let viewController = ReaderSortingOptionViewController(options: availableSortingOptions, preselectedOption: sortingButton.sortingOption) { [weak self] option in
             if let trackingEvent = option.trackingEvent {
-                WPAnalytics.track(trackingEvent)
+                WPAnalytics.track(trackingEvent, properties: ["sortingOption": option.rawValue])
             }
             self?.updateSortingOption(option)
             if self?.presentedViewController != nil {
@@ -366,13 +366,9 @@ extension ReaderCardsStreamViewController {
 
 extension ReaderSortingOption {
     var trackingEvent: WPAnalyticsEvent? {
-        switch self {
-        case .date:
-            return .readerDiscoverSortingOptionDateSelected
-        case .popularity:
-            return .readerDiscoverSortingOptionPopularitySelected
-        case .noSorting:
+        if self == .noSorting {
             return nil
         }
+        return .readerDiscoverSortingOptionSelected
     }
 }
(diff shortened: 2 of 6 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -122,12 +122,31 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         sortingButton.sortingOption = sortingOption
 
         if optionChanged, reloadCards {
+            showGhost()
```
bozidarsevo, line 134: Showing ghost animation when we are fetching new cards after user changes sorting option.

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -122,12 +122,31 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
...
             super.syncIfAppropriate(forceSync: true)
         }
     }
 
     @objc func didTapSortingButton() {
-        // TODO: show bottom sheet
+        let availableSortingOptions: [ReaderSortingOption] = [.popularity, .date]
```
bozidarsevo, line 141: Creating bottom sheet and presenting it.

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -137,9 +156,11 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
         refreshCount += 1
 
         cardsService.fetch(isFirstPage: true, refreshCount: refreshCount, sortingOption: sortingButton.sortingOption, success: { [weak self] cardsCount, hasMore in
+            self?.hideGhost()
             self?.trackContentPresented()
             success(cardsCount, hasMore)
         }, failure: { [weak self] error in
+            self?.hideGhost()
```
bozidarsevo, line 176: Hiding ghost animation after cards are fetched.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -0,0 +1,277 @@
+import UIKit
+
+struct ReaderSortingActionSheetOption {
```
bozidarsevo, line 3: Simple struct to keep the data for the sorting option in the sheet.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+
+class ReaderSortingActionSheetOptionControl: ClosureControl {
+    enum Constants {
+        static let iconsHeight: CGFloat = 24.0
+        static let iconsWidth: CGFloat = 24.0
+        static let iconLeading: CGFloat = 16.0
+        static let labelLeading: CGFloat = 6.0
+        static let minimalTappableHeight: CGFloat = 44.0
+        // TODO: check if we should color from assets or from hex
+        static let checkTintColor: UIColor = UIColor(light: UIColor(named: "Blue50") ?? UIColor(hexString: "2271B1"), dark: UIColor(named: "Blue30") ?? UIColor(hexString: "5198D9"))
```
bozidarsevo, line 27: In Figma it says to use Blue50 and Blue30 for colors, but since using `UIColor(named:)` returns optional color I had to use hex colors as a fallback.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -0,0 +1,277 @@
...
+    // MARK: - setup
+
+    convenience init(option: ReaderSortingActionSheetOption, closure: @escaping () -> Void) {
+        self.init(frame: CGRect.zero, minimalTappableHeight: Constants.minimalTappableHeight, closure: closure)
+
+        self.option = option.sortingOption
+        isSelected = option.checked
+        imageView.image = option.image
+        label.text = option.title
+        accessibilityIdentifier = option.identifier
```
bozidarsevo, line 72: I plan to go thru all UI components in final PR to use proper accesibility identifiers and labels.

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
```diff
@@ -80,6 +80,9 @@ import Foundation
     case readerDiscoverContentPresented
     case readerPostMarkSeen
     case readerPostMarkUnseen
+    case readerDiscoverSortingOptionButtonTapped
```
Gio2018, line 83: rather than track three different events here, I believe we should track one event (sorting button tapped) with two different properties for `sortingOption` to be sent (e.g. `popular` and `recent`).

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -314,3 +339,35 @@ extension ReaderCardsStreamViewController: ReaderSitesCardCellDelegate {
         }
     }
 }
+
```
Gio2018, line 348: can you add a MARK that briefly describes what this extension is about?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -0,0 +1,278 @@
+import UIKit
```
Gio2018, line 1: Is there a reason why we cannot reuse `ActionSheetViewController` like the create button does, instead of creating a whole new object that seems to do a very similar job?

Gio2018 (changes requested):
Hey @bozidarsevo, I left a couple of comments in the code for you to consider.
Other than that, on iPad, I noticed that the ghost triggered from the sorting option looks different from the one that is presented in the reader in all other situations. Not a huge deal, but I would not expect it to be different (see images below):

Existing ghost | Ghost for select a sorting option
-- | --
 |

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
bozidarsevo: Should we have 2 events, 1 when user taps the sorting option button that opens the bottom sheet and 1 when user changes the sorting option by tapping on the item in the bottom sheet?
My understanding was that I should track those things, when user taps button to open the sheet and when user taps the option that changes the sorting.
Maybe to have `readerDiscoverSortingOptionButtonTapped` (plain) and `readerDiscoverSortingOptionSelected` (with `sortingOption` property)?

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
bozidarsevo: I was thinking of using `ActionSheetViewController` but it has a bit different behaviour and UI. To be able to use it we would need to refactor `ActionSheetViewController` to add ability to have different items/options, have option to change the spacing, etc. Basically I would need to make `ActionSheetViewController` a bit more generic.
I can look into it and maybe make a proposal on what would I refactor to be able to use `ActionSheetViewController` for sorting option. Should I do that?

bozidarsevo:
> Hey @bozidarsevo, I left a couple of comments in the code for you to consider.
> Other than that, on iPad, I noticed that the ghost triggered from the sorting option looks different from the one that is presented in the reader in all other situations. Not a huge deal, but I would not expect it to be different (see images below):
> 
> Existing ghost	Ghost for select a sorting option
> 	

Tnx! Regarding the ghost animation, it appears that first time it shows in that "wide" way and after that in "narrow" way. I also checked "Following" tab and it looks like there is the same issue. I will look a bit more into it and try to see what is happening.

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
Gio2018: Good point. I agree it makes sense since those are two different events, let's do this.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
Gio2018: as discussed on slack, it's ok to use the new component.

Gio2018:
> Tnx! Regarding the ghost animation, it appears that first time it shows in that "wide" way and after that in "narrow" way. I also checked "Following" tab and it looks like there is the same issue. I will look a bit more into it and try to see what is happening.

good catch @bozidarsevo ! Since it's a pre-existing problem, we don't need to address this here. Let me know when you have those tracks events updated and I'll give a second pass!

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -139,4 +142,8 @@ class ReaderSortingOptionButton: UIControl {
         label.text = sortingOption.localizedDescription
         iconView.image = sortingOption.image
     }
+    
+    public func setLabelBottomCompensation(_ compensation: CGFloat) {
```
bozidarsevo, line 146: Added this method to set bottom compensation regarding to those empty spaces at the top of the cells. I was thinking of setting it directly when creating constraints, but this way we can set it programmatically depending on the setup in the app and this button can be centered by default and compensated if needed.
Previously I did the compensation in `containerInsets` constant in `ReaderTableCardCell.swift` but this way it is much cleaner and we see better what is happening and why.

Gio2018 (approved):
With the caveat of addressing the accessibility issues discussed on slack, this one looks good to me @bozidarsevo !
:shipit: !

Issue #15056, opened 2020-10-06
Prologue Carousel - Project Issue


mindgraffiti:
A project ticket to track prologue carousel changes. 
Authenticator project ticket: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/issues/494
Feature branch: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/tree/feature/unified-prologue-carousel

### Tasks
- [ ] Remove feature flag
- [x] Remove Lottie

### In Progress
- [ ] Implement parallaxed background
- [ ] Scale text depending on device size

### Done
- [x] Create feature flag #15062
- [x] Update headlines #15064 
- [x] Extend background behind the button view #15073
- [x] Move carousel dots to the bottom of the container
- [x] Add placeholder views for each slide, and reposition header text
- [x] Create individual views for each slide, use real text that can be localized instead of images as text
- [x] Check dark mode support
- [x] Check iPad support

### Designs
![i5-ios-1](https://user-images.githubusercontent.com/1062444/95364944-14624300-0897-11eb-844e-aa5f2dc63401.png)
![i5-ios-dark](https://user-images.githubusercontent.com/1062444/95364953-175d3380-0897-11eb-9247-7d311e9ff3f1.png)

kean:
This screen was removed during the Jetpack migration but is now making a comeback on iPad:

<img width="800" alt="Screenshot 2024-09-11 at 2 16 56 PM" src="https://github.com/user-attachments/assets/5424c912-b112-4b4a-9e5c-00455d58a48b">

PR #15867, opened 2021-02-11
Initial replacement prologue carousel

frosty:
Refs #15056. This PR implements a basic replacement login prologue carousel in WPiOS, which overrides the usual one that is bundled with WordPressAuthenticator. This first version simply adds the paging carousel, and pages with titles. Full design treatment and each new page will be coming later.

**To test**

- Set the `unifiedPrologueCarousel` feature flag to `true`.
- Build and run.
- If you're not already signed out, sign out so you can see the login prologue.
- Swipe through each page and ensure you see a title.
- Build and run again with the feature flag disabled and check that you see the original carousel.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/UnifiedProloguePages.swift
+++ b/WordPress/Classes/ViewRelated/NUX/UnifiedProloguePages.swift
@@ -41,9 +41,8 @@ class UnifiedProloguePageViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    override func viewDidLoad() {
-        super.viewDidLoad()
-
+    override func loadView() {
+        view = UIView()
         view.backgroundColor = .clear
 
         configureStackView()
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/NUX/UnifiedProloguePages.swift
```diff
@@ -0,0 +1,78 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.backgroundColor = .clear
+
+        configureStackView()
+        configureTitle()
```
Gio2018, line 49: since we are not using IB, I wonder if we should build/configure the view in `loadView` instead of `viewDidLoad`? What do you think?

Gio2018 (changes requested):
Hey @frosty , I had one small comment, but other than that, I noticed that the prologue does not look right in dark mode (see screenshot below):

<p align=center>

</p>

WordPress/Classes/ViewRelated/NUX/UnifiedProloguePages.swift
frosty: Sure, happy to do either – I just kind of default to `viewDidLoad` 🙂

frosty:
@Gio2018 I added some tweaks for your comments!

Gio2018 (approved):
LGTM!

PR #15879, opened 2021-02-11
Remove external access to the Core Data Stack’s persistentStoreCoordinator and managedObjectModel

jkmassel:
Continues to work toward #15829 by removing external access to the Core Data Stack's `persistentStoreCoordinator` (PSC) and `managedObjectModel` (MOM).

The PSC and MOM should be an implementation detail of the Core Data Stack. Right now, the PSC is used indirectly for state restoration or similar using `managedObjectID forURIRepresentation:`.

In the future, this could probably be an extension on `NSManagedObject` that could simplify external usage with something like:

```swift
static func from(URIRepresentation url: URL, in context: NSManagedObjectContext) -> Self? {
    guard
        let persistentStoreCoordinator = context.persistentStoreCoordinator,
        let objectID = persistentStoreCoordinator.managedObjectID(forURIRepresentation: url)
    else {
        return nil
    }

    return context.object(with: objectID) as? Self
}
```

For now, removing the PSC from the `CoreDataStack` protocol helps prepare it for replacement.

The MOM is not used from anywhere in the app, which IMHO helps to validate its removal as a property.

The MOM and PSC are both used by the test target on some very old migration tests. While I don't want to remove those tests (in fact, I'd love to expand them further), reworking them entirely is IMHO outside the scope of this PR given that part of this project will be to finish the work started by @Gio2018 to adopt the WooCommerce iterative migrator. To resolve this cyclical dependency, for now I've just swizzled the methods into the test class for use until they can be replaced.

Additionally, some tests previously relied on the persistent store coordinator for `ContextManager.sharedInstance` – this can be resolved by relying on a full mock stack while under test, so this PR adjusts the affected tests to work that way instead (and removes the static initializer for `MockContext`, moving it to the `TestContextManager` instead to avoid re-introducing this usage in future tests).

@Gio2018 The `MockContext` mostly has your name on it, so would you be willing to take a look at this one? 😃 

**To test:**
- Ensure tests pass – there was only one use of this method that was trivially replaced in app code, but there's a (relatively straightforward) swizzling of stack internals in the `ContextManagerMock.m` file now.

**PR submission checklist:**

### DIFF
--- a/WordPress/Classes/Utility/ContextManager.h
+++ b/WordPress/Classes/Utility/ContextManager.h
@@ -5,7 +5,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol CoreDataStack
 @property (nonatomic, readonly, strong) NSManagedObjectContext *mainContext;
-@property (nonatomic, readonly, strong) NSManagedObjectModel *managedObjectModel;
 - (NSManagedObjectContext *const)newDerivedContext;
 - (NSManagedObjectContext *const)newMainContextChildContext;
 - (void)saveContextAndWait:(NSManagedObjectContext *)context;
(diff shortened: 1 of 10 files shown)


Gio2018 (approved):
:shipit:!

jkmassel:
Thanks @Gio2018!

Issue #15832, opened 2021-02-08
Replaces instances of `NSManagedObjectContext` with `ContextManager`


jkmassel:
We currently pass around `NSManagedObjectContext` objects in many places – namely View Controllers and services.

In these cases, we often make changes directly to the `mainContext`, [which is incompatible with how `NSPersistentContainer` is designed to work](https://developer.apple.com/documentation/coredata/nspersistentcontainer/1640622-viewcontext).

We'll replace instances of a bare context being passed around with reads from the `viewContext` and writes using [the new Core Data write API](https://github.com/wordpress-mobile/WordPress-iOS/issues/15830).

dangermattic:
Thanks for reporting! 👍

jkmassel:
Pretty sure we did this.

PR #15880, opened 2021-02-12
Move blog time zone lookup into blog object

jkmassel:
Works toward #15832 by reducing the need to create as many `BlogService` objects. Fewer `BlogService` objects will result in fewer places we need to thread the `CoreDataStack` through, which will make #15832 far simpler.

For this PR, we're moving the `timeZoneForBlog:` method into the `Blog` object, which allows us to remove several ephemeral `BlogService` instances (and as a bonus, a couple of references to `ContextManager.sharedInstance()`).

@aerych – wondering if you can take a look at this, since I see you did similar work in 3ba5cda.

@leandroalonso – wondering if you'd be able to take a look at the tests for this – the approach to using `PostBuilder.setUpInMemoryManagedObjectContext()` in the `PostBuilder` constructor means that the managed object context is destroyed when `PostBuilder` passes out of scope because `NSManagedObject` only holds a weak reference to its context. I went around and fixed this (and added debug assertions to prevent regression) in the failing tests caused by my changes – hope that's ok!!

**To test:**
- The change is covered by tests, so ensure those pass in CI
- If you look at the commits from oldest to newest, they spell out the process in small mechanical steps that should help verify that this code is functionally identical to what came before.
- Ensure the new tests look ok – they should be identical – the change just makes it so the context is owned by the test, not the `Post` object.

**PR submission checklist:**


leandroalonso (approved):
@jkmassel from the tests perspective, it all looks good to me!

Thank you for making this change. :)

Moving the context to the test itself was something that @shiki mentioned a lot of months ago. :)

aerych (approved):
Hola @jkmassel 👋 
Changes look good to me.  I ran the tests locally for paranoia sake and everything passed. A quick smoke test of the app looked good. 
:shipit: from me!

jkmassel:
Thanks @aerych and @leandroalonso!!!

PR #15869, opened 2021-02-11
Fix automatic session tracking

jkmassel:
Enables automatic session tracking via Sentry in WPiOS to provide info on sessions vs crashes

To test:
- Ensure tests pass
- Do a quick smoke test on an installable build

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15848, opened 2021-02-10
Reader: fix post options menu tracks

ScoutHarris:
Ref: #15761 

This fixes some tracks that got lost when I merged the Reader post options menus on https://github.com/wordpress-mobile/WordPress-iOS/pull/15778.

To test:

Post card > tap more button:
`🔵 Tracked: post_card_more_tapped <subscription_count: xxx>`

Post card > toggle seen:
`🔵 Tracked: reader_mark_as_unseen <source: post_card>`
`🔵 Tracked: reader_mark_as_seen <source: post_card>`

Post details > tap more button: 
`🔵 Tracked: reader_article_detail_more_tapped <subscription_count: xxx>`

Post details > toggle seen:
`🔵 Tracked: reader_mark_as_unseen <source: post_details>`
`🔵 Tracked: reader_mark_as_seen <source: post_details>`

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15882, opened 2021-02-12
Merge 16.7 beta 2 into develop

jkmassel:
Merges changes between 16.7 beta 1 and beta 2 into `develop`

Does not include release notes.


AliSoftware:
~Side note: correct me if I'm wrong but you seem to have cut the `merge-16.7-beta-2-into-develop` branch… from the `release/16.7` (as opposed to cutting it from `develop` and merging `release/16.7` into it, before making the PR from that branch to `develop`).~

~That's not an issue since in practice that intermediate branch is not needed here as we don't have conflicts, and as in practice currently both `merge-16.7-beta-2-into-develop` and `release/16.7` are actually pointing to the same commit, that's similar to asking to merge `release/16.7`  to develop`, but just wanted to check if it was intentional (and point out for next time) since personally I usually do it the other way around~

Actually scratch that, now that I think about it I realize that your way works too… and is simpler than mine 😅 
It's just that I've always been doing it in the other direction, and never realized that both solutions work 😄  Since yours is simpler, I'll probably do it your way from now on 😛

AliSoftware (approved):
Seems to match the diff of the fixes that were included in 16.7.0.1, i.e. #15848, #15842 and #15869, and version numbers looks good too, so :shipit: !

PR #15884, opened 2021-02-12
Merge the "Move fastlane into project root" changes into develop

jkmassel:
Because this is a huge change, I'd like to land this in `develop` independent of other changes (instead of trying to sort through a beta fix with all this stuff cluttering up the PR).

It was already approved in #15706 and these changes should be identical – immediately following the PR merge, I cut this branch from `release/16.7`.


AliSoftware (approved):
I have to admit that I didn't review this one as in-depth as the original one, but the diff and git graph makes sense to me, so we should be good.

(I noticed that the PR re-adds the `fastlane/metadata/*/release_notes.txt` files needlessly, but I don't think that's a big deal so 👍 )

PR #15887, opened 2021-02-14
Move `blogByBlogId` out of blog service

jkmassel:
Moves blog look-ups out of the `BlogService` object.

## Why?

The current approach to looking up a single `Blog` object either has insufficient context for proper threading, or requires state leaking into the service. It also encourages the creation of many intermediate `BlogService` objects that are strictly used for looking up `Blog` objects. This complicates a refactor to avoid global access to the Core Data Stack, because we'd need to thread the stack to these instances. If we instead 

Consider the following pseudo-code examples:

```swift
// An example why we can't just use a single context in a service
class BlogService
{
  init(context: NSManagedObjectContext) {
    self.context = context
  }

  func syncBlogs() {
    let blogs = self.allBlogs()
    let remote = BlogRemote(blog.wpcomAPIRemote)
    remote.fetchAllBlogs { blogs in
      // If we want to perform writes here on a context other than the main one, we need a 
      // temporary context. For that we need the stack.
    }
  }
}
```

```swift
// This is better, but now the BlogService doesn't know how to reliably determine where it 
// should perform operations

class BlogService
{
  init(stack: CoreDataStack) {
    self.coreDataStack = stack
  }

  func updateBlogName(forSiteID siteID: Int, name: String) {
    let context = coreDataStack.newDerivedContext()

    context.performAndWait { context in
      // How do we tell Core Data which context to use for this lookup?
      // We could say "always fetch on the view context", but that precludes lookups in a background context
      let blog = self.blog(withID: siteID)
      blog.name = name
      self.coreDataStack.saveContext(context)
    }
  }
}
```

```swift
// Instead, let's pass the `context` in which we're performing the lookup into the service:
class BlogService
{
  init(stack: CoreDataStack) {
    self.coreDataStack = stack
  }

  func updateBlogName(forSiteID siteID: Int, name: String) {
    let context = coreDataStack.newDerivedContext()

    context.performAndWait { context in
      // How do we tell Core Data which context to use for this lookup?
      // We could say "always fetch on the view context", but that precludes lookups in a background context
      let blog = self.blog(withID: siteID, in: context)
      blog.name = name
      self.coreDataStack.saveContext(context)
    }
  }
  
  func blog(withID siteID: Int, in context: NSManagedObjectContext) -> Blog? {
    let fetchRequest = NSFetchRequest<Self>(entityName: Blog.entityName())
    fetchRequest.predicate = NSPredicate(format: "blogID == %@", siteID)
    return try? context.fetch(fetchRequest).first
  }
}
```

Let's pause here. We've fixed the background problem, but our API is very strange now. Consider trying to use it externally:

```swift
let service = BlogService(stack: self.coreDataStack)
let blog = service.blog(withID: 4, in: self.coreDataStack.viewContext) /// why do I need to provide a context _and_ a stack?!
```

Taking a step back, let's look at the problem this API is trying to solve – we want to be able to fetch `Blog` objects without needing to write the `NSFetchRequest` by hand each time. We see this in other services:

```swift
class PostService 
{
  // [...snip...]

  func downloadPosts(forSite siteID: Int) {
     let api = BlogService(context: self.coreDataStack.viewContext).blog(withID: siteID).wpcomAPIRemote
     let remote = PostRemote(api)
     // [...snip...]
  }
}
```

and even in helpers

```swift
class StatsInsightsHelper 
{
   func getStatsCountForBlog(withID siteID: Int) -> [Stat] {
     return BlogService(content: ContextManager.shared.mainContext).blog(withID: siteID).stats 
   }
}
```

We can improve all of these examples by moving the fetch logic into a static method on the `Blog` object:

```swift
extension Blog {
     static func lookup(withID id: Int, in context: NSManagedObjectContext) -> Blog? {
       let fetchRequest = NSFetchRequest<Self>(entityName: Blog.entityName())
       fetchRequest.predicate = NSPredicate(format: "blogID == %d", id)
       return try? context.fetch(fetchRequest).first
     }
  }
}
```

And here is how those examples would look with the new lookup API:

```swift
class BlogService
{
  init(stack: CoreDataStack) {
    self.coreDataStack = stack
  }

  func updateBlogName(forSiteID siteID: Int, name: String) {
    let context = coreDataStack.newDerivedContext()

    context.performAndWait { context in
      // How do we tell Core Data which context to use for this lookup?
      // We could say "always fetch on the view context", but that precludes lookups in a background context
      let blog = Blog.lookup(withID: siteID, in: context)
      blog.name = name
      self.coreDataStack.saveContext(context)
    }
  }
}

```

```swift
/// no need to provide a stack or have an intermediate object
let blog = Blog.lookup(withID: 4, in: self.coreDataStack.viewContext) 
```

```swift
class PostService 
{
  // [...snip...]

  func downloadPosts(forSite siteID: Int) {
    let api = Blog.lookup(siteID, in: self.coreDataStack.viewContext).wpcomAPIRemote
    let remote = PostRemote(api)
    // [...snip...]
  }
}
```

```swift
class StatsInsightsHelper 
{
   func getStatsCountForBlog(withID siteID: Int) -> [Stat] {
      return Blog.lookup(siteID, in: ContextManager.shared.mainContext).stats
   }
}

// One benefit here is that this method becomes far easier to clean up – it now feels an awfully
// lot like something that could go in a `StatsInsightsService`.
//
// So let's refactor it into one

class StatsInsightsService
{
  init(stack: CoreDataStack) {
    self.coreDataStack = stack
  }

  func getStatsCountForBlog(withID siteID: Int) -> [Stat] {
    // Service methods should only ever be called from the main thread since they're accessed from the UI layer
    assert(Thread.isMainThread) 
    return Blog.lookup(withID: siteID, in: self.coreDataStack).stats
  }
}
```

Requesting review from @aerych and @frosty, as y'all have been around this code a lot for a long time. If both of you would be willing to 👍 it, I'll take that as a sign that it's a good refactoring 😅. Happy to answer questions and/or discuss more – there's a lot of explanation I left out to try to keep this from becoming a novel.

**To test:**
Ensure test pass. To review this PR, if you go commit-by-commit, it should be very clear that the change is mechanical, with no new functionality (except tests)

**PR submission checklist:**

### DIFF
--- a/WordPress/Classes/Models/Blog+Lookup.swift
+++ b/WordPress/Classes/Models/Blog+Lookup.swift
@@ -1,0 +1,46 @@
+import Foundation
+
+/// An extension dedicated to looking up and returning blog objects
+public extension Blog {
+
+    /// Lookup a Blog by ID
+    ///
+    /// - Parameters:
+    ///   - id: The ID associated with the blog.
+    ///
+    ///     On a WPMU site, this is the `blog_id` field on the [the wp_blogs table](https://codex.wordpress.org/Database_Description#Table:_wp_blogs).
+    ///   - context:  An NSManagedObjectContext containing the `Blog` object with the given `blogID`.
+    /// - Returns: The `Blog` object associated with the given `blogID`, if it exists.
+    static func lookup(withID id: Int, in context: NSManagedObjectContext) throws -> Blog? {
+        return try lookup(withID: Int64(id), in: context)
+    }
+
+    /// Lookup a Blog by ID
+    ///
+    /// - Parameters:
+    ///   - id: The ID associated with the blog.
+    ///
+    ///     On a WPMU site, this is the `blog_id` field on the [the wp_blogs table](https://codex.wordpress.org/Database_Description#Table:_wp_blogs).
+    ///   - context:  An NSManagedObjectContext containing the `Blog` object with the given `blogID`.
+    /// - Returns: The `Blog` object associated with the given `blogID`, if it exists.
+    static func lookup(withID id: Int64, in context: NSManagedObjectContext) throws -> Blog? {
+        let fetchRequest = NSFetchRequest<Self>(entityName: Blog.entityName())
+        fetchRequest.predicate = NSPredicate(format: "blogID == %ld", id)
+        return try context.fetch(fetchRequest).first
+    }
+
+    /// Lookup a Blog by ID
+    ///
+    /// - Parameters:
+    ///   - id: The NSNumber-wrapped ID associated with the blog.
+    ///
+    ///     On a WPMU site, this is the `blog_id` field on the [the wp_blogs table](https://codex.wordpress.org/Database_Description#Table:_wp_blogs).
+    ///   - context:  An NSManagedObjectContext containing the `Blog` object with the given `blogID`.
+    /// - Returns: The `Blog` object associated with the given `blogID`, if it exists.
+    @objc
+    static func lookup(withID id: NSNumber, in context: NSManagedObjectContext) -> Blog? {
+        let fetchRequest = NSFetchRequest<Self>(entityName: Blog.entityName())
+        fetchRequest.predicate = NSPredicate(format: "blogID == %@", id)
+        return try? context.fetch(fetchRequest).first
+    }
+}
--- a/WordPress/Classes/Models/WPAccount+AccountSettings.swift
+++ b/WordPress/Classes/Models/WPAccount+AccountSettings.swift
@@ -12,8 +12,7 @@ extension WPAccount {
         case .displayName(let value):
             self.displayName = value
         case .primarySite(let value):
-            let service = BlogService(managedObjectContext: managedObjectContext!)
-            defaultBlog = service.blog(byBlogId: NSNumber(value: value))
+            defaultBlog = try? Blog.lookup(withID: value, in: managedObjectContext!)
         default:
             break
         }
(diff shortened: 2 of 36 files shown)


aerych (commented):
Howdy @jkmassel :wave:

I'm a fan of the idea and approach, and I'd like to suggest an addition.

With the current implementation every look up will generate a new fetch request. This is fine when only a blog's ID is known. However, when there is already a Blog instance, it might be nice to have methods that could (potentially) avoid a fetch by calling either `NSManagedObjectContext.existingObjectWithID` or `NSManagedObjectContext.objectWithID` and passing the extant Blog instance's `objectID`.  

The changes to InsightStoreState might benefit from this.  

It's a (minor?) optimization but since this patch is about optimizing I thought I'd throw the idea out there. :)

jkmassel:
Yeah good suggestion @aerych – I've added it in https://github.com/wordpress-mobile/WordPress-iOS/pull/15887/commits/5fa8997f1ccac46b06a8058fc06c0e250f78e96d / https://github.com/wordpress-mobile/WordPress-iOS/pull/15887/commits/a8c5e92de174375444d804cb94436f71f90ae14c.

What's nice is that by doing it as an extension on `NSManagedObject`, we get the functionality on every model right away for free.

Tests aren't included in this PR – I have a branch where I'm building out the updated stack, and [they'll live there](https://github.com/wordpress-mobile/WordPress-iOS/commit/6bb63b129b9606bd405302014c017fee1456d5df) for now until the new stack launches.

aerych:
> What's nice is that by doing it as an extension on NSManagedObject, we get the functionality on every model right away for free.

Super cool 🙇‍♂️

aerych (approved):
:shipit: from me!

frosty (approved):
This looks good to me too! I love the amount of unnecessary BlogService objects we can clean up here!

jkmassel:
Thanks @aerych and @frosty!!

PR #15890, opened 2021-02-15
Update release-toolkit to add new actions to trigger CircleCI jobs via API calls

loremattei:
This PR update `release-toolkit` to a new version which adds a `circleci_trigger_job` action. 
It also sets up a test workflow in CircleCI which will be used as a base for the next steps in a follow-up PR. 

Since testing changes to the release tooling is always complex, I'm splitting this feature in multiple PRs which should make testing easier. Because of this, this PR targets a feature branch.

**To test**
The test workflow is set up to post a message on Slack.

1. `bundle install`
2. Run `bundle exec fastlane test_release_build` and verify that the `A new Release build has been triggered via API!` message has been posted on the build channel on Slack.
3. Run `bundle exec fastlane test_beta_build` and verify that the `A new Beta build has been triggered via API!` message has been posted on the build channel on Slack.

PR submission checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,7 @@ jobs:
       - store_artifacts:
           path: Artifacts
           destination: Artifacts
-  Beta Build:
+  beta-build:
     executor:
       name: ios/default
       <<: *xcode_version
@@ -349,7 +349,7 @@ workflows:
   Beta Build:
     when: or [ << pipeline.parameters.beta_build >>, << pipeline.parameters.release_build >> ]
     jobs: 
-      - Beta Build
+      - beta-build
   Translation Review Build:
     when: << pipeline.parameters.translation_review_build >>
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,7 @@ jobs:
       - store_artifacts:
           path: Artifacts
           destination: Artifacts
-  beta-build:
+  Beta Build:
     executor:
       name: ios/default
       <<: *xcode_version
@@ -347,9 +347,9 @@ workflows:
             tags:
               only: /^\d.\d?(.\d)?(.\d)?(.\d{1,4})$/
   Beta Build:
-    when: or [ << pipeline.parameters.beta_build >>, << pipeline.parameters.release_build >> ]
+    when: << pipeline.parameters.beta_build >>
     jobs: 
-      - beta-build
+      - Beta Build
   Translation Review Build:
     when: << pipeline.parameters.translation_review_build >>
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,12 @@ parameters:
   translation_review_lang_id:
     type: string
     default: all-lang
+  beta_build:
+    type: boolean
+    default: false 
+  release_build:
+    type: boolean 
+    default: false
 
 xcode_version: &xcode_version
   xcode-version: "12.1.0"
@@ -168,6 +174,31 @@ jobs:
       - store_artifacts:
           path: Artifacts
           destination: Artifacts
+  Beta Build:
+    executor:
+      name: ios/default
+      <<: *xcode_version
+    steps:
+      - when:
+          condition: << pipeline.parameters.beta_build >>
+          steps:
+            - run:
+                name: Beta Build Test
+                command: |
+                  echo "export SLACK_SUCCESS_MESSAGE=':tada: A new Beta build has been triggered via API!'" >> $BASH_ENV 
+      - when:
+          condition: << pipeline.parameters.release_build >>
+          steps:
+            - run:
+                name: Release Build Test
+                command: |
+                  echo "export SLACK_SUCCESS_MESSAGE=':tada: A new Release build has been triggered via API!'" >> $BASH_ENV
+      - slack/status:
+          include_job_number_field: false
+          include_project_field: false
+          include_visit_job_action: true
+          webhook: '${SLACK_BUILD_WEBHOOK}'
+          success_message: '${SLACK_SUCCESS_MESSAGE}'
   Release Build:
     executor:
       name: ios/default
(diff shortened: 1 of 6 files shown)


fastlane/Fastfile
```diff
@@ -571,6 +572,24 @@ import "./ScreenshotFastfile"
     )
   end
 
+  lane :test_beta_build do | options |
+    circleci_trigger_job(
+      circle_ci_token: ENV["CIRCLE_CI_AUTH_TOKEN"], 
+      repository:PROJECT_NAME, 
+      branch:"feature/trigger-ci-via-api",
```
AliSoftware: My OCD is triggered 😄 
```suggestion
      repository: PROJECT_NAME, 
      branch: "feature/trigger-ci-via-api",
```
(same for `test_release_build`)

fastlane/Pluginfile
```diff
@@ -6,7 +6,9 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 3.2.0'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.12.0'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', ref: '24db1b638b964ded7ea8a908a765ed8c60728722'
+#gem 'fastlane-plugin-wpmreleasetoolkit', path: '../../../wordpress-mobile/release-toolkit'
```
AliSoftware: 🎗️ It is obvious that you plan to revert this later in other incremental PRs, but just leaving a comment to highlight it so we don't forget later 🙂

fastlane/Fastfile
```diff
@@ -69,6 +69,7 @@ ENV["PUBLIC_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.Publ
 ENV["INTERNAL_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.internal.xcconfig")
 ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
 ENV["PROJECT_ROOT_FOLDER"]=PROJECT_ROOT_FOLDER
+PROJECT_NAME="WordPress-iOS"
```
AliSoftware: Nit: Maybe name this var `REPOSITORY_NAME` instead, to not confuse it with the Xcode project's name.

AliSoftware:
Missing steps to test this PR:

 - Go to https://app.circleci.com/settings/user/tokens
 - Create a Personal API Token
 - Update your existing `~/.wpios-env.default` file with `CIRCLE_CI_AUTH_TOKEN=<Your Token>` – see the updated `fastlane/env/user.env-example` for template.

AliSoftware (approved):
Tested with test beta and release builds and worked as expected. See the 2 Slack messages in p1613412222084100-slack-C5ALGJYEP.

Only left a nit about variable name but overall LGTM 👍

fastlane/Fastfile
AliSoftware: Alternative idea (take it or leave it): determine this dynamically, using:
```ruby
REPOSITORY_NAME=File.basename(`git remote get-url origin`.chomp, '.git')
```

loremattei:
Thank you @AliSoftware!
I addressed your comments and updated `release-toolkit` to the latest.

AliSoftware:
Just saw an occurrence of empty message in p1613515616086800-slack , triggered by a build linked to this branch, which seemed suspicious and probably worth investigating 🧐

loremattei:
Good catch @AliSoftware!
It turned out that `or` conditions must be put on a new line in order to work as expected. Otherwise it looks like the workflow is triggered on API calls as expected, but it's also triggered on every commit.
Single conditions can be added on the same line of the `when` keyword, instead.

AliSoftware:
FWIW it's not about newline but about YAML syntax: here `or` is expected to be a YAML (or JSON, as JSON is just a subset of YAML) hash/dictionary but in your earlier commits when you tried on a single line, you didn't put the `:` between the `or` key and its value 🙃

Anyway, looking better now 👍

.circleci/config.yml
```diff
@@ -168,6 +174,31 @@ jobs:
       - store_artifacts:
           path: Artifacts
           destination: Artifacts
+  Beta Build:
```
AliSoftware, line 177: Nitpick: since this job handles both Beta and Release builds, maybe use a more explicit name… though I'm drawing a blank about ideas (naming is hard)… so maybe that name is ok after all 😅

AliSoftware (approved):
I had to apply the fix we did in https://github.com/wordpress-mobile/WordPress-iOS/pull/15895 on this PR to be able to run `bundle install` when trying to test this PR (as I've updated to bundler `2.2.10`). So I figured I'd commit the fix here (8da16877f1ecaf069815c64c71f121cb881a2fec), which was a good occasion to double-check that new commits don't trigger the "Beta Build" job anymore.

But I noticed that while this works, now builds triggered by API calls trigger the _whole_ workflow (so they do add and run the "Beta Build" job, but that API call still triggers the other workflows.

Shouldn't we prevent especially the `wordpress_ios` workflow to run in these situations – like we do when the `translation_review_build` parameter is passed? Or were you planning to make the real "Beta Build" jobs ultimately depend on `wordpress_ios` and wait for it to finish building to then make the "Beta Build" job finish by adding the GH release and tag?

Approving the PR nonetheless, assuming that ☝️ is what you plan to go with in your next PR, but if not, then probably worth fixing (here or in the next one)

.circleci/config.yml
loremattei: No, as the name of the lanes suggest, this is just a test setup to test the changes in `release-toolkit` and how to use a combination of parameters to trigger the same workflow. 
We'll update the `Release build` workflow in the final version.

loremattei: And yes, at that point we'll update the configuration to trigger only the workflows that we want to. 👍

AliSoftware: Perfect 👍

PR #15898, opened 2021-02-17
Show loader while activating theme

### LINKED ISSUE #15329: Activating a theme does not close the preview or disable the "Activate" button
### Expected behavior
I expect activation to look like something is installing and then the preview to close.

### Actual behavior
1. During activation, there is no user feedback, so I do not actually know anything is happening and I am likely to keep tapping the Activate button. (The video is a bit faster, actual time of seemingly no action is about 7 seconds). 
2. It stays open and the Activate button is still enabled, so even though I got the message thanking me for choosing the theme, I still feel like it didn't work and I need to tap the button again. 

https://cloudup.com/cPSa1glnLMw

yaelirub:
Fixes #15329 

To test: 
1. Go to themes 
2. Got to a specific theme 
3. If its active, notice the bar item tittle is "Active" , if it's not active notice title = "Activate"
4. If it's active see Customizer, if it's not active see theme
5. Tap "Activate" and notice spinner 
6.  Notice spinner is hidden once theme is active
7. Notice title turns to disabled "Active" once the theme is activated 

* Anything else I didn't think of 

PR submission checklist:

theme_loader

### DIFF
--- a/WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift
+++ b/WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift
@@ -182,6 +182,12 @@ public protocol ThemePresenter: class {
     @objc func resumingSearch() -> Bool {
         return !suspendedSearch.trim().isEmpty
     }
+    
+    fileprivate lazy var activityIndicator: UIActivityIndicatorView = {
+           let indicatorView = UIActivityIndicatorView(style: .medium)
+           indicatorView.translatesAutoresizingMaskIntoConstraints = false
+           return indicatorView
+       }()
 
     open var filterType: ThemeType = ThemeType.mayPurchase ? .all : .free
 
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift
```diff
@@ -183,6 +183,15 @@ public protocol ThemePresenter: class {
         return !suspendedSearch.trim().isEmpty
     }
 
+    fileprivate var activityIndicator: UIActivityIndicatorView = {
+        let indicatorView = UIActivityIndicatorView(style: .medium)
+        indicatorView.frame = CGRect(x: 0.0, y: 0.0, width: 40.0, height: 20.0)
```
Gio2018: nit: since we already have grouped constants in this class [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/163292dcb88b606e51d5ec325be072fda7461c9f/WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift#L250), maybe we can add these numbers (or even just the `CGRect` itself) to these constants?

WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift
```diff
@@ -780,6 +793,12 @@ public protocol ThemePresenter: class {
                 let successMessage = String(format: successFormat, theme?.name ?? "", theme?.author ?? "")
                 let manageTitle = NSLocalizedString("Manage site", comment: "Return to blog screen action when theme activation succeeds")
                 let okTitle = NSLocalizedString("OK", comment: "Alert dismissal title")
+
+                self?.activityIndicator.stopAnimating()
+                self?.activateButton?.customView = nil
+                self?.activateButton?.isEnabled = false
+                self?.activateButton?.title = ThemeAction.active.title
```
Gio2018: nit: totally not a big deal, but maybe we can group these calls in an helper method for readability? Same with the two calls to start the animation above.

Gio2018 (approved):
Tested on devices: iPhone 11 Pro Max - iOS 14.4 and iPad Pro 11 2nd gen, iPadOS 14.4 and it worked as described.

One thing that I noticed, not directly related to this PR, is that if I activate a theme from the "meatball menu" (see image below), there is no similar feedback for the users before the theme is actually activated. This is not a huge deal, but I can open/handle a separate issue if it's the case.

Other than that, I left a couple of minor comments that can be safely ignored.

:shipit:!

<p align=center>

</p>

WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift
yaelirub: Done in 3a57070fa24d7e0913ae182564a7da9e800d56fb

WordPress/Classes/ViewRelated/Themes/ThemeBrowserViewController.swift
yaelirub: Done in 3a57070fa24d7e0913ae182564a7da9e800d56fb

yaelirub:
Thank you so much for the review, @Gio2018 🙏 
Would you please open an issue to handle loading from activate in the bottom sheet. I think we need design solution there 
cc: @iamthomasbishop

PR #15899, opened 2021-02-17
Comments: provide status when fetching

### LINKED ISSUE #15893: Comments: update remote to allow specifying type
Currently, `CommentServiceRemoteREST:getCommentsWithMaximumCount` requests `status:all` when fetching comments. To support filtering, we need to be able to specify the `status`.

- WPKit: update comment method(s) to accept `status`.
- WPiOS: update to use new/updated WPKit methods, ensuring the current Comment list does not change.

ScoutHarris:
Fixes: #15893 
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/350

This uses the WPKit changes to explicitly provide a Comment status when fetching. Right now it's still fetching `all`, but this will change when Comment filters are added.

The eagle-eyed reviewer (😉)  will notice it is not necessary to pass `commentStatusAll` to `syncCommentsForBlog` and `loadMoreCommentsForBlog` as I've changed it to do because the remotes add `status:all` if `status` is not provided. I did this for two reasons:
1. I'll be using this method later when I add filters.
2. It conveniently provides an easy way to test different statuses. You'll see. 😄 

To test:

There should be no visible change in the Comments displayed for WP or self-hosted sites. So for both site types:
- Go to Comments.
- Verify all Comments are listed as before, that is all Pending first followed by all Approved.

To test these changes actually work for different `status` types:
- In `CommentService`, change `commentStatusAll` to a different status. 
  - The `status` parameters for REST and XMLRPC endpoints differ for approved and unapproved.
  - To test REST, use:
    - `approved`
    - `unapproved`
  - To test XMLRPC, use:
    - `approve`
    - `hold`
  - `trash` and `spam` are the same for both.
- The Comments view does not refresh when first viewed, so pull to refresh.
  - The UI is not equipped to display all the statuses correctly, so I've added temporary log messages for the Comment content and status.
- Verify the status logged out matches the value you set for `commentStatusAll`.
```
🔴 comment content: <content>
🔴 comment status: <status>
```

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/CommentService.h
+++ b/WordPress/Classes/Services/CommentService.h
@@ -2,6 +2,7 @@
 #import "LocalCoreDataService.h"
 
 extern NSUInteger const WPTopLevelHierarchicalCommentsPerPage;
+extern NSString *commentStatusAll;
 
 @class Blog;
 @class Comment;
@@ -28,6 +29,11 @@ extern NSUInteger const WPTopLevelHierarchicalCommentsPerPage;
                     success:(void (^)(BOOL hasMore))success
                     failure:(void (^)(NSError *error))failure;
 
+- (void)syncCommentsForBlog:(Blog *)blog
+                 withStatus:(NSString *)status
+                    success:(void (^)(BOOL hasMore))success
+                    failure:(void (^)(NSError *error))failure;
+
 // Determine if a recent cache is available
 + (BOOL)shouldRefreshCacheFor:(Blog *)blog;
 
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Services/CommentService.h
```diff
@@ -2,6 +2,7 @@
 #import "LocalCoreDataService.h"
 
 extern NSUInteger const WPTopLevelHierarchicalCommentsPerPage;
+extern NSString *commentStatusAll;
```
ScoutHarris, line 5: I need to define all the endpoint `status` types, but I'm not sure yet where/how that will be. This will do for now since we're just using `all`.

WordPress/Classes/Services/CommentService.m
```diff
@@ -14,6 +14,7 @@
 NSUInteger const WPTopLevelHierarchicalCommentsPerPage = 20;
 NSInteger const  WPNumberOfCommentsToSync = 100;
 static NSTimeInterval const CommentsRefreshTimeoutInSeconds = 60 * 5; // 5 minutes
+NSString *commentStatusAll = @"all";
```
ScoutHarris, line 17: I need to define all the endpoint `status` types, but I'm not sure yet where/how that will be. This will do for now since we're just using `all`.

aerych (approved):
Hola @ScoutHarris 👋 
Patch looks good :).  I tested as described and everything worked.  I was able to fetch and view a specific set of comments depending on the value I used for status. 
:shipit: !

ScoutHarris:
Thanks @aerych !

PR #15906, opened 2021-02-17
Adds a "Confetti" view to be used anywhere in the app.

Gio2018:
This PR also adds a feature flag for Milestone Notifications

Fixes #NA

**Note**
At the moment, the confetti view is hooked to the existing achievement system (that is why it's still in draft status)

To test:
- Before building
    - set the `milestoneNotifications` feature flag to `true`
    - in `NotificationsDetailViewController`, edit the `isViewMilestone` property to match an existing notification type, e.g. a likes or "first post" milestone. In this case,  in that property, you can replace `note.type == "view_milestone"` with `note.type == "like_milestone_achievement"` or with `note.type == "post_milestone_achievement"` (or both, if you want to test it on more notifications).

- Build and run
    - log in with an account that have existing achievement notifications (e.g. likes), or trigger an achievement notification from another source
    - Go to "Notifications" and open the achievement notifications, and make sure the confetti view is displayed (see example below)

<p align=center>

</p>

PR submission checklist:


WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -159,14 +163,23 @@ class NotificationDetailsViewController: UIViewController {
...
+        super.viewDidAppear(animated)
+        showConfetti()
+    }
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         keyboardManager?.stopListeningToKeyboardNotifications()
         tearDownNotificationListeners()
         storeNotificationReplyIfNeeded()
         dismissNotice()
+
```
emilylaguna:
```suggestion
```

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -1221,7 +1234,46 @@ extension NotificationDetailsViewController: SuggestionsTableViewDelegate {
     }
 }
 
+// MARK: - Milestone notifications
+//
+private extension NotificationDetailsViewController {
+
+    /// Determines if the notification content is a views milestone
```
emilylaguna:
```suggestion
    /// Determines if the notification content is a view milestone
```

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -146,11 +148,13 @@ class NotificationDetailsViewController: UIViewController {
         setupKeyboardManager()
 
         Environment.current.appRatingUtility.incrementSignificantEvent(section: "notifications")
+
+
```
emilylaguna:
```suggestion
```

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -1221,7 +1234,46 @@ extension NotificationDetailsViewController: SuggestionsTableViewDelegate {
...
+// MARK: - Milestone notifications
+//
+private extension NotificationDetailsViewController {
+
+    /// Determines if the notification content is a views milestone
+    var isViewMilestone: Bool {
+        if let type = note.type, type.contains("view_milestone"), FeatureFlag.milestoneNotifications.enabled {
+            return true
+        }
+        return false
```
emilylaguna: I think we can simplify this a bit, what do you think?

We can also change the `contains` to an equals since we know the exact type the notification will be. 

```suggestion
        return FeatureFlag.milestoneNotifications.enabled && note.type == "view_milestone"
```

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -1221,7 +1234,46 @@ extension NotificationDetailsViewController: SuggestionsTableViewDelegate {
...
+    func addConfettiViewIfNeeded() {
+        guard isViewMilestone else {
+            return
+        }
+        let newConfettiView = ConfettiView()
+
+        confettiView = newConfettiView
+        guard let navigationController = navigationController else {
+            return
+        }
```
emilylaguna: Should we move this to the guard at the top of the method since we don't show the confetti if the nav controller check fails?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -159,14 +163,23 @@ class NotificationDetailsViewController: UIViewController {
...
         tearDownNotificationListeners()
         storeNotificationReplyIfNeeded()
         dismissNotice()
+
     }
 
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        removeConfettiView()
+    }
```
emilylaguna:
```suggestion
    }
    
```

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -159,14 +163,23 @@ class NotificationDetailsViewController: UIViewController {
         setupNotificationListeners()
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        showConfetti()
+    }
```
emilylaguna, line 168:
```suggestion
    }
    
```

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -1233,6 +1285,9 @@ extension NotificationDetailsViewController {
 
         onSelectedNoteChange?(previous)
         note = previous
+        removeConfettiView()
+        addConfettiViewIfNeeded()
+        showConfetti()
```
emilylaguna: Since this block of methods is used in a few places how do you feel about  combining `removeConfettiView()` and `addConfettiViewIfNeeded()` into the `showConfetti()` method and rename it to `showConfettiIfNeeded()` ?

emilylaguna (changes requested):
Awesome job and working well with the "real" data! I left a few comments below. 

I also noticed when I tapped on the 'take me there' option and then tapped back the confetti was replayed, I'm thinking it may be a bit too much. How about restricting it to just when the notification object changes or the view loads for the first time?

Gio2018:
Thanks for reviewing @emilylaguna !! I have applied the requested changes.

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -1221,7 +1234,50 @@ extension NotificationDetailsViewController: SuggestionsTableViewDelegate {
...
+        FeatureFlag.milestoneNotifications.enabled && note.type == "view_milestone"
+    }
+
+    func addConfettiViewIfNeeded() {
+        guard let navigationController = navigationController else {
+            return
+        }
+        let newConfettiView = ConfettiView()
+
+        confettiView = newConfettiView
```
frosty: Can we just assign `confettiView = ConfettiView()` directly here, and then set the frame etc on `confettiView` below?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -91,13 +91,17 @@ class NotificationDetailsViewController: UIViewController {
...
+            confettiWasShown = false
             router = makeRouter()
             refreshInterface()
             markAsReadIfNeeded()
         }
     }
 
+    var confettiView: ConfettiView?
+
+    var confettiWasShown = false
```
frosty: Rather than introducing a property here just to track whether confetti has been shown, we could instead add a property to `ConfettiView` to track if it has been fired or not. Then in the `note` setter above we could call something like `confettiView.reset()` to reset the property? It just slightly lessens the confetti-specific state added to the details view controller.

WordPress/Classes/ViewRelated/Notifications/Milestone Notifications/ConfettiView.swift
```diff
@@ -0,0 +1,193 @@
...
+        layer.removeFromSuperlayer()
+    }
+}
+
+
+extension ConfettiView {
+
+    func emitConfetti() {
+        // Images
+        let star = UIImage(named: "confetti-star")!
```
frosty: Are we okay with these being force unwrapped?

WordPress/Classes/ViewRelated/Notifications/Milestone Notifications/ConfettiView.swift
```diff
@@ -0,0 +1,193 @@
...
+        let circle = UIImage(named: "confetti-circle")!
+        let hotdog = UIImage(named: "confetti-hotdog")!
+
+        // Colors
+        let purple = UIColor(red: 0.75, green: 0.35, blue: 0.95, alpha: 1.00)
```
frosty, line 180: Should we translate these to muriel colors?

frosty (commented):
Looking good! I've left a few comments in the code.

I also wonder if we should consider decoupling the confetti view from the notification detail view controller lifecycle (or more explicitly, from `viewDidDisappear`). The main reason is that when we navigate away from a notification detail screen, the confetti just vanishes in an instant once the VC has disappeared:

https://user-images.githubusercontent.com/4780/108502330-8ec1bb80-72aa-11eb-81d8-b1ffc0cbdf01.mov

I think it would be nice in the circumstance if the confetti could complete its descent to the bottom of the screen. Perhaps the confetti view should instead be owned by the navigation controller or the tab bar controller?

WordPress/Classes/ViewRelated/Notifications/Milestone Notifications/ConfettiView.swift
emilylaguna: I don't think these colors are in the muriel set. cc @mbshakti ?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
Gio2018: we can, but then we have to unwrap `confettiView` before adding it to the view?

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
Gio2018: I totally see your point here. I did not want `ConfettiView` to hold a state, being just something that you can add anywhere and throw confetti. Does this make sense?

WordPress/Classes/ViewRelated/Notifications/Milestone Notifications/ConfettiView.swift
Gio2018: Not okay 😀. Changed.

Gio2018:
Thanks for reviewing @frosty !

> when we navigate away from a notification detail screen, the confetti just vanishes in an instant once the VC has disappeared

That should be solved now

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
```diff
@@ -1221,7 +1229,32 @@ extension NotificationDetailsViewController: SuggestionsTableViewDelegate {
...
+              !confettiWasShown,
+              let view = UIApplication.shared.mainWindow,
+              let frame = navigationController?.view.frame else {
+
+            return
+        }
+
+        // removing this instance only after the animation completes, will prevent
+        // the animation to suddenly stop if users navigate away from the note
+        ConfettiView.cleanupAndAnimate(on: view, frame: frame) { $0.removeFromSuperview() }
```
frosty: How about if the ConfettiView just removes itself from the superview in its own animation completion method? 🤔

frosty (commented):
Changes look good, but I have one question – just added a comment!

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationDetailsViewController.swift
Gio2018: You mean in the convenience method? Sure!

frosty: So I mean that we don't need to pass `$0.removeFromSuperview` here, and in `ConfettiView`'s `animationDidStop` we just add a call to `removeFromSuperview`?

Gio2018: As discussed on slack, I added some clarifications and changed the implicit parameter with an explicit one, to better clarify the method.

frosty (approved):
:shipit:

emilylaguna (approved):

PR #15910, opened 2021-02-17
ABTest: fix a nil crash

leandroalonso:
This PR fixes a crash that was happening if the user has opted out for Analytics tracking. This makes the ExPlat to not be configurated but the app was still trying to use it. The changes are:

1. Make the request only if ExPlat was already configured
2. Always resort to `control` if the user opted out

### To test

1. Run the app
2. Disable the "Collect information" option under Privacy
2. Run the app again
3. No crashes :)

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/AB Testing/ABTest.swift
+++ b/WordPress/Classes/Utility/AB Testing/ABTest.swift
@@ -6,7 +6,7 @@ enum ABTest: String, CaseIterable {
 
     /// Returns a variation for the given experiment
     var variation: Variation {
-        return ExPlat.shared.experiment(self.rawValue)
+        return ExPlat.shared?.experiment(self.rawValue) ?? .control
     }
 }
 
@@ -18,6 +18,6 @@ extension ABTest {
             return
         }
 
-        ExPlat.shared.refresh()
+        ExPlat.shared?.refresh()
     }
 }


leandroalonso:
@fluiddot can I bother you to review this one?

fluiddot (commented):
I tested in the same conditions where I experienced the crash and now is working 🎊 , thanks!

fluiddot:
After reading the description of the PR, I enabled the "Collect information" option in the build where I experienced the crash and it didn't happen. So for testing this PR I think it would be ideal to have that option disabled to assure the crash it not reproducible.

fluiddot:
@leandroalonso I didn't approve the PR because I'm not very familiar with iOS native code. If that's already checked let me know and I'll go ahead and  approve it.

jkmassel (approved):
This seems ok from a "prevent a crash" perspective.

Not sure if we want to default to control – that's more of a data science concern 😁

leandroalonso:
@fluiddot good point about the testing steps! And don't worry, thanks for reviewing the crash itself!

> Not sure if we want to default to control – that's more of a data science concern 😁

Good question @jkmassel! I was 99% sure that we should default to `control` and @aaronyan helped me with the remaining 1%, so yeah, that's correct. :)

PR #15915, opened 2021-02-18
Quick Start: Automatically check off steps if the user completes them outside of Quick Start

### LINKED ISSUE #15712: Quick Start: Completing a step on your own does not automatically remove it from the list
### Expected behavior
If I change the site title or complete any other step in Quick Start, the step should no longer be on the list. 

### Actual behavior
The step remains in the list of things to do

1. Create a new WP.com site in the app
2. Accept help when prompted to start Quick Start
3. Change the site title or upload an icon on your own and see that those items are still in the Customize list

https://cloudup.com/cwDobmUpO01

frosty:
Fixes #15712. This PR adds some extra calls to complete Quick Start checklist items if the user performs those items while not actively completing their related Quick Start tour.

As an example, if Quick Start is active on a user's site and the user changes their site title, then the Site Title Quick Start checklist item should be checked off, regardless of whether that is the current 'tour' or not.

The logic here is slightly different from normal Quick Start. Ordinarily, Quick Start just navigates a user to the place in the app where they would complete an action – for example, for the Site Title or Site Icon tours, the user just has to tap the site title or the site icon in the blog detail header, and the tour is complete. For the change in this PR, however, we wanted to only check the items off if the user actually completes them – if they actually tap Save to change their site title, or site icon.

I couldn't reuse the existing `visited` method, as this expects:

- The visited element to be part of the current Quick Start tour, otherwise it's ignored.
- It follows the flow outlined above where a user might not have to necessarily complete an action.

Instead, I added a new `complete` method that simply marks off an item and doesn't post a notification about it (so that it doesn't interrupt an existing tour).

Finally, this PR adds completion actions for _most_ Quick Start tours but not _all_:

- Set your site title
- Upload a site icon
- View your Site
- Edit your homepage
- Enable post sharing
- Publish a post

If doesn't include these items because they don't really have explicit actions that the user can perform – the user could just be looking around the app and not actively trying to find those areas:

- Review site pages
- Check your site stats

It also doesn't include "Follow other sites", because if the Reader tour isn't underway, there's currently no easy way to tell which site should have its Quick Start tour checked off from the Reader (elsewhere in My Site we always have a reference to a blog that we can pass).

**To test**

- Build and run
- Either create a new site and accept Quick Start, or use the debug menu to enable Quick Start for a site
- Ignore any prompts from Quick Start from now on
- For each of these items, complete that action in the app and then look on the Quick Start lists to check that it has been marked as complete:

- [x] Set your site title
- [x] Upload a site icon
- [x] View your Site
- [x] Edit your homepage (just tap the row for your homepage in Pages)
- [x] Enable post sharing (link your site to Twitter or similar using the Sharing entry in My Site)
- [x] Publish a post

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,7 @@
 -----
 * [**] Prevent deleting published homepages which would have the effect of breaking a site. [#15797]
 * [**] Prevent converting published homepage to a draft in the page list and settings which would have the effect of breaking a site. [#15797]
+* [*] Quick Start: Completing a step outside of a tour now automatically marks it as complete. [#15712]
 
 16.7
 -----
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
@@ -69,6 +69,9 @@ extension BlogDetailsViewController {
         blog.settings?.name = title
         headerView.setTitleLoading(true)
 
+        QuickStartTourGuide.shared.complete(tour: QuickStartSiteTitleTour(),
+                                                    silentlyForBlog: blog)
+
         let service = BlogService(managedObjectContext: context)
         service.updateSettings(for: blog, success: { [weak self] in
             NotificationCenter.default.post(name: NSNotification.Name.WPBlogUpdated, object: nil)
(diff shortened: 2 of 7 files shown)


Gio2018 (approved):
tested on device iPad Pro 11 2nd gen, iPadOS 14.4 and worked as expected.
:shipit:

frosty:
Thanks Giorgio!

PR #15931, opened 2021-02-19
Update secrets in develop

mokagio:
Run via `bundle exec fastlane run configure_update`

To test: If CI passes we're fine.


mokagio:
I originally opened PRs against `develop` and the release branch both to avoid having to merge the release branch into `develop` to get this change.

I then realized that we'll be merging the release branch into develop soon as a result of the release cycle finalization. The release branch is the only one that consumes the values changed in this update.

PR #15933, opened 2021-02-19
Extract filterable categories class and use for layouts and site designs

### LINKED ISSUE #3125: Initial design picker screen: add categories
### Why are we doing this?

We want to make it easier for our users to find the design that they are looking for by presenting them in a categorized list. Since we are planning to add a lot more designs in the future.

### Goals

- Update the current design picker to behave in the same way as the current categorized page picker (Starter Page Template Picker). 
- The user should be able to select a category. Once selected only the selected categories design should be available to be chosen.
- Once a design is selected the user should be presented with the previous “Preview” and “Choose” buttons. 
- If the user selects preview they should be able to preview the designs. The same way they were able to before.
- If the user selects the  “Choose” button they should be presented with the choose the domain screen. (The same as before)

### Changes that will need to happen to the apps.

1. Update the copy of the description on the screen from "Pick your favourite homepage layout. You can edit and customize it later." to
"Get started with a pre-made layout and customize it to fit your needs"
2. The new categorized design picker will be very similar to the Page picker design. 
3. Our new layouts will be grouped horizontally in the same way as the Starter Page Picker.
4. Note that the aspect ratio of the preview image is staying the same. 

#### Endpoint Changes
We are also implementing a new endpoint. The endpoint is being implemented in D56541-code.

The endpoint will be very similar to the common-block-layouts and will return a list of design + categories. The URL that we are planning to use will be. We have decided that for Android we will return the new "Single page designs + the original designs. While for iOS we will just return the original designs for now. 
https://public-api.wordpress.com/wpcom/v2/common-starter-site-designs/

- Query Parameters:
    - `type=mobile` - to signify this is the mobile apps
    - `group` - You will be able to pass in a group through as a query parameter in order to return the correct group of layouts.
        - By default the endpoint will ____?____ 
        - **iOS** should set this to equal `original` to exclude the old layouts.
        - **Android** should set this to equal `original,single-page`
    - `scale` - This works the same as it does for the common-block-layouts
    - `preview_width` - This works the same as it does for the common-block-layouts

<details>
<summary>Screenshot of the new designs.</summary>

image

</details>

### Additional context 

This Issue is part of the single page layout project. pb3aDo-IP-p2
Internal proposal and discussion.  pb3aDo-Lz-p2

mkevins:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3125

#### Related PR:

`WordPressKit-iOS`: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/352

### Description

This PR extracts a `FilterableCategoriesViewController` class from the implementation previously used in the modal layout picker so that the site design collection screen can use the same filterable categories view controller.

Since there are differences in the sibling subclasses (and the types used to represent their data), I've also introduced a `CategorySection` protocol which allows each subclass to continue to benefit from the strong type bindings, but also utilize the more general implementations of the superclass they inherit from for most of the filtering and display logic.

#### :construction: Device Mode Selection

Since the device mode selection feature has not yet been added to the modal layout picker, the extracted class does not support this yet. <s>This will need to be implemented in the superclass to retain the feature for the site design screen </s> (and it will subsequently be enabled for the modal layout screen as well).

**Update:** This has been updated to re-enable device mode selection on the site design screen. The feature can be enabled for the modal layout picker in a subsequent PR.

# Testing steps

The steps to test are mostly the same as flows we've already tested, e.g. the category filters should work the same way in the site design screen as they do in the layout picker. To be thorough, I've enumerated many of the flows below.

<s>

#### :desert: Sandbox

In order to test this PR, the API will need to be sandboxed until the API changes are deployed.
</s>

**Update:** The API changes have been deployed. This PR can now be tested without sandboxing :tada:

## Modal Layout Picker

The layout picker for new pages should behave the same as before. Test to make sure no regressions were introduced during the refactoring.

<details>
<summary>Create / preview a page tests</summary>

### Create a Blank Page
1. Create a new page for a Gutenberg site.
1. Select "Create a Blank Page"
     - **Expect** to see the Editor Load and not present the old Starter Page Template Picker Bar

### Create a Page from a layout
1. Create a new page for a Gutenberg site.
1. Select a layout
1. Select "Create Page"
     - **Expect** to see the Editor Load with the contents of the layout

### Preview a layout
1. Create a new page for a Gutenberg site.
1. Select a layout
1. Select "Preview"
     - **Expect** to see a Read-Only Editor load with the contents of the layout
1. Select "Create Page"
     - **Expect** to see the Editor load with the contents of the layout

</details>

<details>
<summary>Filter bar tests</summary>

### Select a Category
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Deselect the category
    - **Expect** to see the other categories return

### Select Multiple-Categories
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Select a different category
    - **Expect** to see the new section added to the list
1. Deselect a category
    - **Expect** the list to update again to be only the selected categories
1. Deselect the rest of the categories
    - **Expect** the all of the categories to return

### Selecting a Layout then filter on the same category
1. Select a layout in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the selected layout to still be selected.
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout to still be selected.

### Selecting a Layout then filter on the same category deselect layout
1. Select a layout in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the selected layout to still be selected.
1. Deselect the layout
    - **Expect** the bottom bar to revert to "Create a Blank Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for no layout to be selected

### Selecting a Layout then filter on a different category
1. Select a layout in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout to still be selected.

### Selecting a Layout then filter on a different category select a different layout
1. Select a layout in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected layout
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Select a different layout
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected layout from step 3 to still be selected.

</details>

## Site Design Picker

Navigate to the Create WordPress.com site - My Sites > ➕ > Create WordPress.com site

<details>
<summary>Create / preview site tests</summary>

### Create a Site - Selecting a Design
<details>
<summary>Steps</summary>

1. Select a design
1.  **Expect**
    - To see the footer slide in from the bottom
    - To see a border and checkmark animate in place on the selected design.
1. Select Choose
1.  **Expect**
    - To see the analytics event `enhanced_site_creation_site_design_selected` with the property `template` populated with the selected layout's slug
1. Search for and select a domain
1. Select Create Site
1.  **Expect**
    - To see a temporary screen as the site is being created
    - To see the site confirmation screen.

</details>

### Preview then Create a Site

<details>
<summary>Steps</summary>

1. Select a design
1. Select Preview
1.  **Expect** the preview to show for the selected design
1. Select Choose
1. **Expect** to see the preview modal dismiss and for the Choose a Domain screen to be "pushed" on to the navigation flow.
1. Search for and select a domain
1. Select Create Site
1.  **Expect**
    - To see a temporary screen as the site is being created
    - To see the site confirmation screen.

</details>
</details>

##### Categories

The site design collection screen should now show a category filter bar that functions the same as that used in the layout picker.

<details>
<summary>Filter bar tests</summary>

### Select a Category
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Deselect the category
    - **Expect** to see the other categories return

### Select Multiple-Categories
1. Select a category
    - **Expect** to see the sections condensed to be only the selected category filter
1. Select a different category
    - **Expect** to see the new section added to the list
1. Deselect a category
    - **Expect** the list to update again to be only the selected categories
1. Deselect the rest of the categories
    - **Expect** the all of the categories to return

### Selecting a design then filter on the same category
1. Select a design in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected design
    - **Expect** to see the section updated to only show the selected category, and for the selected design to still be selected.
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected design to still be selected.

### Selecting a design then filter on the same category deselect design
1. Select a design in a category such as Blog
1. Select the category in the filter bar that matches the category of the selected design
    - **Expect** to see the section updated to only show the selected category, and for the selected design to still be selected.
1. Deselect the design
    - **Expect** the bottom bar to revert to "Create a Blank Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for no design to be selected

### Selecting a design then filter on a different category
1. Select a design in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected design
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected design to still be selected.

### Selecting a design then filter on a different category select a different design
1. Select a design in a category such as Blog
1. Select a category in the filter bar that does not match the category of the selected design
    - **Expect** to see the section updated to only show the selected category, and for the bottom bar to remain as "Preview" and "Create Page"
1. Select a different design
1. Deselect the category in the filter
    - **Expect** to see the other categories return, and for the selected design from step 3 to still be selected.

</details>

### Site design thumbnails

<details>
<summary>Tests</summary>

#### Initial Mode on a Phone
1. Start the Home Page Picker on a phone
1. **Verify** that **mobile** thumbnails are displayed

#### Initial Mode on a Tablet
1. Start the Home Page Picker on a tablet
1.  **Verify** that **tablet** thumbnails are displayed

#### Thumbnail options availability
1. Start the Home Page Picker on a phone
1.  **Verify** that there is a **thumbnail button** at the left of skip
1. Press the thumbnail button
1.  **Verify** that three options are displayed: **Mobile, Tablet, Desktop**

#### Change to Tablet
1. Start the Home Page Picker on a phone
1. Press the thumbnail button (at the left of skip)
1. Select **Tablet**
1.  **Verify** that tablet thumbnails are displayed

#### Change to Desktop
1. Start the Home Page Picker on a phone
1. Press the thumbnail button (at the left of skip)
1. Select **Desktop**
1.  **Verify** that desktop thumbnails are displayed

#### Change to Mobile
1. Start the Home Page Picker on a phone
1. **Change to Desktop** as above
1. Press the thumbnail button (at the left of skip)
1. Select **Mobile**
1.  **Verify** that mobile thumbnails are displayed
</details>

### Site design previews

<details>
<summary>Tests</summary>

#### Preview options availability
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. **Verify** that there is a preview mode button at the top right of the screen
1. Press the **preview mode button**
1. Verify that three options are displayed: Mobile, Tablet, Desktop

#### Mobile preview
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. Press the **preview mode button**
1. Select **Mobile**
1. **Verify** that a mobile preview is displayed

#### Tablet preview
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. Press the **preview mode button**
1. Select **Tablet**
1. **Verify** that a tablet preview is displayed

#### Desktop preview
1. Start the Home Page Picker
1. Select a design
1. Press the **Preview** button
1. Press the **preview mode button**
1. Select **Desktop**
1. **Verify** that a desktop preview is displayed

</details>

### Mode between steps

<details>
<summary>Tests</summary>

#### Change mode on the preview screen

1. Start the Home Page Picker on a phone
1. Select a design
1. Press the **Preview** button
1. Press the preview mode button
1. Select **Desktop**
1. **Verify** that desktop mode is displayed
1. Close the preview screen (back or swipe down)
1. **Verify** that desktop thumbnails are displayed

#### Change mode from the design screen
1. Start the Home Page Picker on a phone
1. Press the preview mode button
1. Select **Desktop**
1. Select a design
1. Press the **Preview** button
1. **Verify** that a desktop preview is displayed
</details>

---

PR submission checklist:


WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
```diff
@@ -0,0 +1,237 @@
...
+        didSet {
+            if !(oldValue != nil && selectedItem != nil) {
+                itemSelectionChanged(selectedItem != nil)
+            }
+        }
+    }
+    private let filterBar: CollapsableHeaderFilterBar
+
+    /* This should be overriden by the subclass to provide a conforming collection of categories */
+    internal var categorySections: [CategorySection] { get { [] } }
```
mkevins: I'm not thrilled with this pattern (as it offers no compiler signaling to the subclass). For example, this would have been nicer as protocol getter (along with the suggested snippet the subclass implementation would get for free). But in my explorations, I didn't find a valid inheritence heirarchy to accomplish that. This seemed acceptable to me, all things considered, but open to suggestions if you have any :smiley:

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
```diff
@@ -1,33 +1,52 @@
 import UIKit
 import Gutenberg
 
-protocol LayoutPickerSectionTableViewCellDelegate: class {
-    func didSelectLayoutAt(_ position: Int, forCell cell: LayoutPickerSectionTableViewCell)
-    func didDeselectItem(forCell cell: LayoutPickerSectionTableViewCell)
-    func accessibilityElementDidBecomeFocused(forCell cell: LayoutPickerSectionTableViewCell)
+protocol CategorySectionTableViewCellDelegate: class {
+    func didSelectItemAt(_ position: Int, forCell cell: CategorySectionTableViewCell, slug: String)
```
chipsnyder, line 5: Nitpick: I'm thinking about how this delegate could be used in the future if we applied other use cases. So what follows might be premature optimization.

I'm wondering why we decided to add the `slug` to the delegate signature when we already reference the position. My original thinking with this delegate was to provide the position, which could then be used to get the section data from the data provider. The delegate then gets to decide what information is important to it, and this UI only needs to know enough information to display the data. 

If the goal is not to perform that lookup, I wonder if we should pass the full object to let the delegate decide what it needs.  WDYT?

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -2,11 +2,20 @@ import UIKit
...
     var layouts: [PageTemplateLayout]
     var scrollOffset: CGPoint
 
+    var title: String { section.title }
+    var emoji: String? { section.emoji }
+    var categorySlug: String { section.slug }
+    var description: String? { section.desc }
+    var thumbnails: [Thumbnail] {
+        // TODO: pass device different modes after data model supports them
+        layouts.map { Thumbnail(urlDesktop: $0.preview, urlTablet: $0.preview, urlMobile: $0.preview, slug: $0.slug) }
```
chipsnyder: Instead of iterating through each `layout` to transform  `PageTemplateLayout` to `Thumbnail` could we instead make the `Thumbnail` a `protocol` and have `PageTemplateLayout` conform to that?

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
```diff
@@ -2,11 +2,20 @@ import UIKit
 import Gridicons
 import Gutenberg
 
-class GutenbergLayoutSection {
+class GutenbergLayoutSection: CategorySection {
```
chipsnyder, line 12: We were previously passing the `PageTemplateCategory` as the protocol for the filter bar. Out of curiosity what was the reasoning behind moving the protocol conformance here instead? It seems like this object is just acting as a pass-through to the `PageTemplateCategory` So I'm wondering why we would add that layer of abstraction? 

I'm wondering if this change was an attempt to clean up calls like `sections[index].section` which I agree is confusing. If that is the goal could we reduce the confusion by renaming `section` in `GutenbergLayoutSection` to `category`?

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
```diff
@@ -1,33 +1,52 @@
...
+    var title: String { get }
+    var emoji: String? { get }
+    var description: String? { get }
+    var thumbnails: [Thumbnail] { get }
+    var scrollOffset: CGPoint { get set }
+}
+
+class CategorySectionTableViewCell: UITableViewCell {
+
+    struct Thumbnail {
```
chipsnyder: I feel like this could be a protocol instead of a struct to save work in mapping between objects. We could then define that protocol alongside or inside `CategorySection`

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
```diff
@@ -1,33 +1,52 @@
...
     static let estimatedCellHeight: CGFloat = 310.0
 
     @IBOutlet weak var categoryTitle: UILabel!
     @IBOutlet weak var collectionView: UICollectionView!
 
-    weak var delegate: LayoutPickerSectionTableViewCellDelegate?
+    weak var delegate: CategorySectionTableViewCellDelegate?
 
-    private var layouts = [PageTemplateLayout]() {
+    private var thumbnails = [Thumbnail]() {
```
chipsnyder, line 39: I like the rename to `thumbnails` 👍

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
```diff
@@ -76,23 +95,24 @@ extension LayoutPickerSectionTableViewCell: UICollectionViewDelegate {
     }
 
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-        delegate?.didSelectLayoutAt(indexPath.item, forCell: self)
+        guard let slug = section?.categorySlug else { return }
```
chipsnyder, line 97: Nitpick: Related to my other comment I feel like this decision belongs to the delegate so that the delegate can decide what's important to it and how it should behave if the slug is missing.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
chipsnyder: I think there are a few options here:

1. Don't make this a computed property and have it be the base for `sections` in the subclasses. This wouldn't provide the compiler hint but would reduce the duplicated variables in the subclasses.

2. We could have this be:
```
private let dataSource: FilterableCategoriesDataSource
```
We can then pass the dataSource in through the `init` and that dataSource could be something like:

```
protocol FilterableCategoriesDataSource {
     var categorySections: [CategorySection] { get }
}
```
This provides the compiler hint and in the terms of the subclasses, `SiteDesigns` and `GutenbergLayoutSection` could conform to `FilterableCategoriesDataSource` 

3. The getter could provide a fatal error at runtime like `fatalError("This should be overridden by the subclass to provide a conforming collection of categories")`. This would force the implementation in the subclass. 

Out of these options, I think I like `#2` the best.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
```diff
@@ -0,0 +1,237 @@
...
+
+    private var filteredSections: [CategorySection]?
+    private var visibleSections: [CategorySection] { filteredSections ?? categorySections }
```
chipsnyder, line 22: ❤️

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
```diff
@@ -0,0 +1,237 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        tableView.register(CategorySectionTableViewCell.nib, forCellReuseIdentifier: CategorySectionTableViewCell.cellReuseIdentifier)
+        filterBar.filterDelegate = self
+        tableView.dataSource = self
+        configureCloseButton()
+        navigationItem.backButtonTitle = backButtonTitle
```
chipsnyder: Nitpick: Since this is meant as a base class, I feel like this setting could be done in the subclasses. If we do decide to keep this here, I think we should make the variable optional. That way, we can provide a mechanism not to override the default back button behavior if desired. WDYT?

WordPress/Classes/ViewRelated/Site Creation/DesignSelection/SiteDesignContentCollectionViewController.swift
```diff
@@ -1,67 +1,78 @@
 import UIKit
 import WordPressKit
 
-class SiteDesignContentCollectionViewController: CollapsableHeaderViewController, UIPopoverPresentationControllerDelegate {
+class SiteDesignSection: CategorySection {
```
chipsnyder, line 10: This has the same structure as `GutenbergLayoutSection` I'm wondering if instead of repeating this object if we could have `CategorySection` be a `struct` and then have `RemoteSiteDesignCategory` and `PageTemplateCategory` conform to one protocol and then have `RemoteSiteDesign` and `PageTemplateLayout` conform to a different protocol.

WordPress/Classes/ViewRelated/Site Creation/DesignSelection/SiteDesignContentCollectionViewController.swift
```diff
@@ -1,67 +1,78 @@
...
+    }
+}
+
+class SiteDesignContentCollectionViewController: FilterableCategoriesViewController, UIPopoverPresentationControllerDelegate {
     let completion: SiteDesignStep.SiteDesignSelection
     let itemSpacing: CGFloat = 20
     static let aspectRatio: CGFloat = 0.75
     static let cellWidth: CGFloat = 160
     static let cellHeight: CGFloat = cellWidth / aspectRatio
     let cellSize = CGSize(width: cellWidth, height: cellHeight)
```
chipsnyder: nitpick: I know this is a WIP so apologies if this causes noise that you already have covered. 

`itemSpacing`, `aspectRatio`, `cellWidth`,  `cellHeight`, and `cellSize` were all added as something for the collection view, so these can likely all go away now.

chipsnyder (commented):
This is a great start @mkevins! 

Left a few notes mostly related to some nitpick preferences and preferences suggestions on other options where we can add or adjust some protocols that I think could help add clarity.

enejb:
Nice work @mkevins. 
I did some early testing and notice a couple of things. (Also a couple of things on the endpoint that I have now fixes as well. 

1. The spacing around the screenshots needs some padding. See
Screen Shot 2021-02-19 at 2 25 48 PM
2. You probably know this already, We should be display the mobile screenshot and be able to switch between the designs. :) 

3. Also we are planning to just query the original group on the iOS implementation. 
You can do that by making sure that the api calls 
`https://public-api.wordpress.com/wpcom/v2/common-starter-site-designs/?type=mobile&group=original` which should only return the original designs.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
mkevins: Thanks for providing some ideas here Chip!

> 1. Don't make this a computed property and have it be the base for `sections` in the subclasses. This wouldn't provide the compiler hint but would reduce the duplicated variables in the subclasses.

I had initially tried this, but it won't work because generics are not covariant:

`Property 'sections' with type '[GutenbergLayoutSection]' cannot override a property with type '[CategorySection]'`

so, I introduced a computed property to alias the sections subtypes and perform the implicit upcasting.

> 2. We could have this be:
> ```
> private let dataSource: FilterableCategoriesDataSource
> ```
> We can then pass the dataSource in through the `init` and that dataSource could be something like:
> 
> ```
> protocol FilterableCategoriesDataSource {
>      var categorySections: [CategorySection] { get }
> }
> ```
> This provides the compiler hint and in the terms of the subclasses, `SiteDesigns` and `GutenbergLayoutSection` could conform to `FilterableCategoriesDataSource` 

I began investigating this approach, and I agree that this one seems the most promising, but it isn't as straightforward as it may at first seem. `GutenbergLayoutSection` in it's current form represents a _single_ section, and conformance to `FilterableCategoriesDataSource` would require it to _have a_ collection of `CategorySections`.

I also investigated making `GutenbergLayoutPickerViewController` conform to `FilterableCategoriesDataSource` (seemingly more appropriate regarding the plural _has a_  relationship with the `sections` array),__ and a different problem arises: it must pass `self` as the `FilterableCategoriesDataSource` when calling its `super.init`, but it cannot do so from within its own `init`, since `self` is not yet defined. We could make it `var dataSource` (instead of `let`) and assign in subclass `init` after the call to `super.init`, but that feels awkward: `self.dataSource = self`.

Generally, the collection must be variable since we assign asyncronously via remote callbacks, and I believe this restricts us from binding at init, and as a consequence, I think we forfeit that kind of compiler signalling. E.g. in the latter approach, there is nothing stopping someone from omiting the `dataSource` assignment while subclassing `FilterableCategoriesViewController` in `init`.

It's possible that I'm missing something in your suggestion, so if what I tried is not what you had in mind, please elaborate further.

> 3. The getter could provide a fatal error at runtime like `fatalError("This should be overridden by the subclass to provide a conforming collection of categories")`. This would force the implementation in the subclass. 

Would it make sense to include this behind a debug flag?

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
mkevins: Good idea! I moved this while moving so many things to the super class, but now that you mention it, it is already in the super super class :smile: so, no need for the extra middle layer here. I've pushed a change to put it back in the subclasses. :+1:

WordPress/Classes/ViewRelated/Site Creation/DesignSelection/SiteDesignContentCollectionViewController.swift
mkevins: For this, we were still using `cellWidth`, so I [removed these](https://github.com/wordpress-mobile/WordPress-iOS/pull/15933/commits/4827a06f86031fa017622d6b8c6743b1f5975700), but also [refactored the remote service a bit](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/352/commits/8ca2c7617d1de11073c6d6dd036fac20051635dd) to begin aligning it with the remote layouts service.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
mkevins: I meant to leave a comment here when I pushed the draft. I originally tried the protocol approach you mentioned, but was surprised that it would not work the way we want, since `var thumbnails: [Thumbnail] { get }` is required by the adopted `CategorySection` protocol in the subclasses' collections. Specifically, what I ran into was that Swift does not support conformance to read-only property requirements as a covariant ([this answer on SO](https://stackoverflow.com/a/42716340) was informative).

I used this computed mapping as a stop-gap for now, but I'm still looking for a better pattern to improve this.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
mkevins: I had this same thought. This is one area where the degree of generality could be increased. My initial goal was to move as much logic as was _in common_ between the two implementations to the superclass I was extracting, and as it currently stands, this includes _the delegate itself_. I.e. currently, `FilterableCategoriesViewController` has the `CategorySectionTableViewCellDelegate` extension, and not the subclasses. My rationale for using the `slug` was that it is currently part of the `CategorySection` protocol, as a non-optional, so it should unambiguously identify the section, and it was intenionally minimal (only a `String`) to move in the direction of decoupling entirely.

I think that as we massage this into shape for re-use, we could relieve both `CategorySectionTableViewCellDelegate` and `CategorySection` from their knowledge requirement about `slug`s, but it's hard to know what kind of shape it should take without knowing the other specific uses. E.g. there is still a fair degree of coupling at the data model level in the protocol.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
mkevins: I had originally tried this approach, but encountered some issues ([described here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15933#discussion_r580070089)).

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
mkevins: Technically, this is really an indication that the `section` is missing, as the `slug` is non-optional. I think this could be written a bit better to make that more clear.. maybe `guard let section = section else { return }` and dereference on the next line would be more clear. :thinking:

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
mkevins: > Out of curiosity what was the reasoning behind moving the protocol conformance here instead?

The rationale for moving the conformance here (and later merging the protocols) was to provide the necessary abstraction as a collaborator with the extracted base class. I've tried to structure my commits to be as buildable and atomic as was reasonable, and following the changes in commit order might help elucidate the issue.

In the initial state, the base class extraction was fairly strongly coupled to the underlying types through a reference to the collection of sections (which itself holds references to its specifically typed collaborators). In order to make the base class general while maintaining the benefits to the strong type bindings of the former implementations within their new subclasses, the type heirarchy of view controller classes was mirrored in their collaborators.

WordPress/Classes/ViewRelated/Site Creation/DesignSelection/SiteDesignContentCollectionViewController.swift
mkevins: I had this in mind at the start of the refactor, but if I understand correctly, I think some other things need to be addressed before we can do this, because we rely on the more specific type bindings throughout the subclass implementations.

For example, if we renamed `designs` and `layouts` to some common name in a struct, with a protocol type, and eliminated the type heirarchy of the `sections`, we can no longer safely dereference the needed properties / methods from the data model instances. We could address that by merging the data models too (these endpoints are now looking nearly identical), but that feels inflexible with respect to API changes, and also limits the usage to things conforming to that particular unified data model.

mkevins:
> This is a great start @mkevins!
> 
> Left a few notes mostly related to some nitpick preferences and preferences suggestions on other options where we can add or adjust some protocols that I think could help add clarity.

Thank you for taking a look and providing very thoughtful feedback! :smiley: I've addressed some of the improvements you suggested, and pushed changes. For some other suggestions / questions, I've described some of the challenges I encountered while refactoring, and my thought process and rationale for the choices I made. Please let me know if you have any additional thoughts or suggestions.

mkevins:
Thanks for testing it out @enejb !

> Nice work @mkevins.
> I did some early testing and notice a couple of things. (Also a couple of things on the endpoint that I have now fixes as well.
> 
> 1. The spacing around the screenshots needs some padding. See
> 
> Screen Shot 2021-02-19 at 2 25 48 PM

I recently made some small changes with the parameters passed, though I'm not sure whether this addresses what your mentioning here. I'll take a closer look at this once the device modes are re-enabled.

> 2. You probably know this already, We should be display the mobile screenshot and be able to switch between the designs. :)

Yes, this is a known issue in the current state. I'll push changes to address it.

> 1. Also we are planning to just query the original group on the iOS implementation.
>    You can do that by making sure that the api calls
>    `https://public-api.wordpress.com/wpcom/v2/common-starter-site-designs/?type=mobile&group=original` which should only return the original designs.

Ah, thanks for the reminder. :smile:  I had originally left that parameter off just to see the complete list, but I'll be sure to add it before making this ready to land.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
chipsnyder: Ah, I see the issue you were running into, I didn't consider that limitation. I wonder if it would still be beneficial to set up a protocol and use the `thumbnails` getter as a wrapper for `layouts`. I'm thinking of something like https://github.com/wordpress-mobile/WordPress-iOS/commit/d50ba9ad24337b81d0ad00970f8b2cf079d561eb

To be clear though I thinking the effort of mapping to a struct is minimal so not a show stopper.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
chipsnyder: No this is fine. My confusion really started with the delegate's signature but I don't have a strong feeling here.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
chipsnyder: > Would it make sense to include this behind a debug flag?

We could although for this use case it would closely align to our other [instances](https://github.com/wordpress-mobile/WordPress-iOS/search?q=fatalError) where this is would be a developer error and wouldn't be executed in `develop` unless a change to the screen was made and not tested. So I don't think a dev flag would be necessary.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/GutenbergLayoutPickerViewController.swift
mkevins: Nice, thank you for the suggestion! I pulled the changes from your branch to make it a protocol and use the wrapper property in the implementation. :+1:

mkevins: The only thing that we lose is the namespacing for Thumbnail, but I don't think that's a huge issue, and we could prepend it in the future if the need arises.

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/CategorySectionTableViewCell.swift
mkevins: I've merged your suggestion in to extract Thumbnail as a protocol. :+1:

mkevins:
@enejb :wave: :smile: 

> 1. The spacing around the screenshots needs some padding. See
> 
> Screen Shot 2021-02-19 at 2 25 48 PM

Waiting for the cache to clear, and hopefully this will be resolved by the update you deployed.

> 2. You probably know this already, We should be display the mobile screenshot and be able to switch between the designs. :)

The preview device mode selection should be re-enabled now.

> 3. Also we are planning to just query the original group on the iOS implementation.
>    You can do that by making sure that the api calls
>    `https://public-api.wordpress.com/wpcom/v2/common-starter-site-designs/?type=mobile&group=original` which should only return the original designs.

I've updated the request to include `group=original,single-page` for iOS (per the discussion). Also, I've added a mechanism to change this from WordPress-iOS without an additional WordPressKit-iOS PR.

@chipsnyder and @antonis , this is aready for another pass. I've updated the description with test-cases, mostly the ones we are familiar with already.

antonis (approved):
Great job @mkevins 👍 
I tested this on an iPhone SE 2020 (iOS 14.2) and on an iPad simulator. Everything worked as expected! The code LGTM too 🎉

chipsnyder:
@kyleaparker @mbshakti If you'd like to check out the new layouts, the new designs are now available as part of this [build](https://install.appcenter.ms/orgs/automattic/apps/WPiOS-One-Offs/).

chipsnyder (approved):
Very nicely done @mkevins code looks good and is working as expected 👍

mkevins:
> Waiting for the cache to clear, and hopefully this will be resolved by the update you deployed.

It seems that this was not (or at least _not only_) a back-end issue after all. The aspect for the thumbnails did not match (~0.7 instead of 0.75), so I've pushed an update to address that.


Thumbnail type | Before (ratio ~= 0.7) | After (ratio == 0.75)
-|-|-
Site Designs | site-design-thumbnail-aspect-7 | site-design-thumbnail-aspect-75
Page Layouts | page-layout-thumbnail-aspect-7 | page-layout-thumbnail-aspect-75

WordPress/Classes/ViewRelated/Gutenberg/Layout Picker/FilterableCategoriesViewController.swift
mkevins: I've pushed an update to add the message as a fatalError in the base class.

enejb:
Thanks for fixing the ration it looks much better now! 🥳

kyleaparker:
@chipsnyder thanks for sharing! Is this ready for a design review at this point?

chipsnyder:
@mkevins is this good to go for a design review now?

mkevins:
@kyleaparker  and @chipsnyder :wave: :smile: 

> @chipsnyder thanks for sharing! Is this ready for a design review at this point?
> @mkevins is this good to go for a design review now?

I've updated the branch to include `stable` designs (from the latest API change). This is ready for design feedback. Thank you!

kyleaparker:
Looking really good! 

I'd like to make a few changes to the categories: 
- Remove blank category
- Reorder categories to: About, Blog, Links, Highlights, Splash, Business, Professional (We can experiment with this order)

Some other questions: 
- We had talked about updating the aspect ratio of the thumbnails to look more mobile centric. Is this still a possibility?
- Is there a way we can auto-resize the thumbnails so that two thumbnails always take up the width of the screen? For example: 

Screen Shot 2021-03-01 at 11 22 05 AM

Ideally the left and right margins are always the same for those first two thumbnails and the thumbnails would scale up on larger devices.

chipsnyder:
> Remove blank category

cc @enejb

antonis:
Hey @kyleaparker 👋 
just jumping in this conversation since I'll [align the implementation on Android](https://github.com/wordpress-mobile/WordPress-Android/pull/14144).

>We had talked about updating the aspect ratio of the thumbnails to look more mobile centric. Is this still a possibility?

If we agree on this. Will it apply on all designs and layouts (since the the implementation tries to reuse the interface on both use cases)?

>Is there a way we can auto-resize the thumbnails so that two thumbnails always take up the width of the screen?

We did something similar in the old Home Page Picker with the grid layout. In that case there were extra constraints for tablet devices, maximum container width etc (ref: `pb3aDo-xN-p2`). Do these apply here too?

Thank you for the feedback 🙇

kyleaparker:
>If we agree on this. Will it apply on all designs and layouts (since the the implementation tries to reuse the interface on both use cases)?

Yep all layouts! We should make sure that we are optimizing for mobile

>We did something similar in the old Home Page Picker with the grid layout. In that case there were extra constraints for tablet devices, maximum container width etc (ref: pb3aDo-xN-p2). Do these apply here too?

I'm not familiar with the implementation here, is there an example of how it worked on the old homepage picker on larger devices? Past a certain width we could switch to 3 visible per row to account for a larger screen size.

enejb:
It looks like the "blank" category is due to the "Slate" design. 
What category should that design have instead?

cc: @chipsnyder and @kyleaparker.

chipsnyder:
>  Is this still a possibility?
> Yep all layouts! We should make sure that we are optimizing for mobile

Yup this should still be easy enough. It looks like there was just some confusion from the GH issue and the design docs. @enejb can we update the aspect ratio on the backend? 

This will affect the screenshots for the existing layouts for the starter page templates. So we might want to leave the current aspect ratio as the default and either pass in the desired Aspect ratio to the two endpoints or include the height. WDYT @enejb, @mkevins, and @antonis?

---

> Is there a way we can auto-resize the thumbnails so that two thumbnails always take up the width of the screen? 

This might become a bigger change to make this dynamic. My performance on this would be to adjust the size so that it meets the ask for the iPhone then work on making it dynamic as a PR that follows. I'm afraid that change would delay this PR from merging this week. 

As a follow-up, I'm concerned that users won't think to scroll horizontally if the width isn't set to leave a hint of the offscreen thumbnail visible. @iamthomasbishop I know you intentionally made that offscreen thumbnail partially visible so not sure if you had explored this route at all. 


> I'm not familiar with the implementation here, is there an example of how it worked on the old homepage picker on larger devices? Past a certain width we could switch to 3 visible per row to account for a larger screen size.

@kyleaparker here's a sample of what it used to look like with three items per row.

<kbd><a href="https://user-images.githubusercontent.com/3384451/109555657-d06d1600-7aa3-11eb-973c-5b5539e7778c.png"></a></kbd>

enejb:
> @enejb can we update the aspect ratio on the backend?

We sure can. Since this is a new endpoint we could make it so that we have a new default ration. If we want to have a different ration between iOS and Android I would suggest we pass in a ratio query parameter in the endpoint. 

Would the "Page Layout selector" also have a different ratio? If we are changing that as well. We should probably update that as well. 

What is the new ratio that we would like to see?

mkevins:
@chipsnyder and @enejb , I was wondering about the ratios. To clarify, are we talking about passing an aspect ratio parameter in the fetch request that will be used to generate the mshots images? If so, one concern I have is that we will not know in advance what the ratios are for every device (since the ratios would be based on the devices' widths), and so a higher number of users will encounter cache misses.

Would an alternative be to have a fixed ratio from the service, and instead, adjust the width to meet Kylea's criteria, but also adjust the height to match the aspect of the thumbnails?

antonis:
Hey @kyleaparker, @chipsnyder, @mkevins, @enejb 

>Would an alternative be to have a fixed ratio from the service, and instead, adjust the width to meet Kylea's criteria, but also adjust the height to match the aspect of the thumbnails?

I agree with that. Having a different ratio per device will complicate things. 

>What is the new ratio that we would like to see?

I experimented with a `0.64` (was `0.75`) aspect ratio (as in the [design](https://mobilegutenpagesp2.files.wordpress.com/2021/01/newlayouts.png), ref: `pb3aDo-KV-p2`) and a width so that two thumbnails fit per row (on my device). You can check some screenshots and download a build in a [temporary/draft PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14171) I created. This is just a hackish implementation hardcoding some values and always displays mobile thumbnails (changing the mode does nothing). The changes are applied both in Home Page Picker and Modal Layout Picker.

Implementation wise (at least on Android) it will need some custom UI code to make this work completely dynamically since it is not only the thumbnail size that changes  but also the height of the category rows.

One other issue is that on phone-landscape mode the thumbnails do not fit the screen even with the header collapsed.
<image width="500" src="https://user-images.githubusercontent.com/304044/109657816-987ed500-7b6e-11eb-8ef5-96884c92af9c.png" />

Another question is how this should behave on tablet devices.

kyleaparker:
> It looks like the "blank" category is due to the "Slate" design. What category should that design have instead?

@enejb I don't think we need to show this one, since it's just the default that we use if someone clicks the skip button


>This might become a bigger change to make this dynamic. My performance on this would be to adjust the size so that it meets the ask for the iPhone then work on making it dynamic as a PR that follows. I'm afraid that change would delay this PR from merging this week.

>As a follow-up, I'm concerned that users won't think to scroll horizontally if the width isn't set to leave a hint of the offscreen thumbnail visible. @iamthomasbishop I know you intentionally made that offscreen thumbnail partially visible so not sure if you had explored this route at all.

@chipsnyder let's just focus on iPhone so that we don't delay merging 👍 Users will still see a hint of the next layout, if you take a [look at my example](https://github.com/wordpress-mobile/WordPress-iOS/pull/15933#issuecomment-788080847) you can see the yellow background of the next design. The margins on the screen edges are larger than the margin between designs, and it ensures if there are only two designs they have equal margins on the left and right.

chipsnyder:
@kyleaparker and I were chatting and we landed on a new aspect ratio of `2:3` so for the current width `160` we'd have images that are `160x240`.  We can then discuss the more dynamic widths as a future enhancement. 

Based on our discussions this might still need some tweaks.

> To clarify, are we talking about passing an aspect ratio parameter in the fetch request that will be used to generate the mshots images? If so, one concern I have is that we will not know in advance what the ratios are for every device (since the ratios would be based on the devices' widths), and so a higher number of users will encounter cache misses.

Yeah, the suggestion is to pass a predefined aspect ratio `2:3` to the API in addition to the screen width we send today. This change will move the predefined aspect ratio that is in the API to the apps. So this will have the same behavior as the code does today. 

When we explore more dynamic widths we might want to discuss the cached images more but that shouldn't hold anything up for this design iteration IMO.

> Would an alternative be to have a fixed ratio from the service, and instead, adjust the width to meet Kylea's criteria, but also adjust the height to match the aspect of the thumbnails?

Sorry for the confusion here. Having a fixed aspect ratio (defined by the client) is what I was suggesting. I was also trying to consider the impact on the existing STP (Page Creation) endpoints and apps since those will also need to change and they don't have new endpoints. 

Since the Aspect ratio is a UI item I feel like that should be information passed to the API service anyway instead of having UI definitions live in the APIs. This way the screen layout can change without needing an API change.  

To help add clarity here are the changes I'm thinking of:

**Edit**: This has been updated to supply width and height instead of aspect ratio. after a chat with @enejb 

| endpoint | 16.8 and below | 16.9 and above |
| --- | --- | --- |
| (Site Creation) `/rest/v1.1/nux/starter-designs` | no change | N/A |
| (Site Creation) `/wpcom/v2/common-starter-site-designs` | N/A | add `preview_height` = `240` |
| (Page Creation) `/wpcom/v2/sites/{{blogID}}/block-layouts` | Service defaults to old Aspect Ratio | add `preview_height` = `240` |
| (Page Creation) `/wpcom/v2/common-block-layouts` | Service defaults to old Aspect Ratio | add `preview_height` = `240` |

For `/wpcom/v2/common-starter-site-designs` I would recommend supporting `preview_height` as well since these two endpoints are closely tied and I believe UI specs should be in the client and this still might be tweaked in the future. 

cc @enejb @mkevins @antonis 

-------
**Edit**: This has been updated to supply width and height instead of aspect ratio.

mkevins:
I've pushed an update to add the `preview_height` parameter and adjust the thumbnail sizes for the 2:3 ratio (as 160 / 240). This required testing via the sandbox, since the API changes are not yet deployed, so here are some screenshots to show how it looks:

<details>
<summary>Screenshots</summary>

Thumbnail type | Orientation | Before | After
-|-|-|-
Site Designs | Portrait | site-designs-portrait-before | site-designs-portrait-after
Site Designs | Landscape | site-designs-landscape-before | site-designs-landscape-after
Page Layouts | Portrait | page-layouts-portrait-before | page-layouts-portrait-after
Page Layouts | Landscape | page-layouts-landscape-before | page-layouts-landscape-after

</details>

chipsnyder (approved):
I tested this with the latest changes everything looks good to me 👍

chipsnyder:
https://github.com/wordpress-mobile/WordPress-iOS/pull/15933#issuecomment-788080847 
>I'd like to make a few changes to the categories:
>
> - Remove blank category
> - Reorder categories to: About, Blog, Links, Highlights, Splash, Business, Professional (We can experiment with this order)

I almost missed this ask @enejb can we set this order in the `categories` section then have iOS and Android honor that? 
This way if new categories are added in the future we won't need to update hardcoded ordering.

cc @antonis @mkevins

mkevins:
Thank you all for testing and reviewing! Before merging, I ran the CI installable build once more to generate the IPA file. This can be used for testing. The API ordering will be respected, so to test w/o blank, and with the specified category order, either we can test with the sandbox, or wait for the endpoint changes to be deployed.

PR #15862, opened 2021-02-11
Remove newMainContextChildContext from SharingButtonsViewController

jkmassel:
Like its friends #15849 and #15861, this PR continues to work towards #15829.

It replaces `newMainContextChildContext` with a reference to the global `mainContext` for now (this will be changed once we adopt `NSPersistentContainer`). There was no indication from the code _why_ it used a child context – unlike `PeopleController`, the sharing settings are persisted between runs and there are no calls to `rollback` anywhere. 

To test:
- Ensure tests path
- Pick a test site, then go to "Sharing", then choose "Manage" under the "Sharing Buttons" heading.
- Made adjustments to the available sharing buttons, then compare your adjusted settings with Calypso at WordPress.com > Settings > Sharing.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15861, opened 2021-02-10
Replace newMainContextChildContext in PeopleViewController

jkmassel:
In the style of #15849, this PR continues to work towards #15829.

There's no logic in this file (or in the `PeopleService`) that requires a child context, so swapping this for the `mainContext` should work with no further adjustments – and does in my testing.

To test:
- Launch the app, then navigate to "My Site" > "People" (only available for sites you're an admin on – I'd recommend apps.wordpress.com)
- Try inviting someone to the site
- Try removing someone from the site
- Try changing someone's role on a site
- Any other tests you can think to run 😅

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15935, opened 2021-02-19
Remove no-longer-needed newMainContextChildContext

### LINKED ISSUE #15829: Remove `newMainContextChildContext` and other unneeded API from `CoreDataStack`

jkmassel:
Closes #15829

**To test:**
All the code that depends on this method has been removed in prior PRs (#15849, #15861, #15862), so we just need to ensure that tests pass.

I also added a release note about this work – while it should be identical, this way it'll get added to the CfT.

PR submission checklist:

_This PR just needs review from one person, but I'm tagging a few folks for review so that it can land before the `16.8` code freeze – whoever gets to it first wins. Thanks y'all 💖_


leandroalonso (approved):
Tested Reader, People Management and Share. All good! :shipit:

PR #15936, opened 2021-02-19
Add Prepublishing title field

bjtitus:
Adds a title field to the Prepublishing sheet.

| Expanded | Compact |
|-|-|
| <a href="https://user-images.githubusercontent.com/3250/108464896-37860180-723e-11eb-9338-10704d33ef72.png"></a>|<a href="https://user-images.githubusercontent.com/3250/108464925-42d92d00-723e-11eb-94d8-451ada6777ba.png"></a>|

* Adds a title field which is automatically selected when the sheet shows up. Editors can control which fields are shown using the `prepublishingIdentifiers` variable and it can also be specified in the `PrepublishingViewController`s initializer.
* When the keyboard is shown, the Prepublishing sheet will not allow user transition into the compact state. This prevents the sheet from getting stuck behind the keyboard.

## Testing

* Open the Blog Details page
* Tap on the FAB and select "Story Post"
* Add media and hit the confirm button in the lower right
* Fill out the title field and publish the post

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
@@ -1,12 +1,29 @@
 import UIKit
 import WordPressAuthenticator
+import Combine
 
 private struct PrepublishingOption {
     let id: PrepublishingIdentifier
     let title: String
+    let type: PrepublishingCellType
 }
 
-private enum PrepublishingIdentifier {
+private enum PrepublishingCellType {
+    case value
+    case textField
+
+    var cellType: UITableViewCell.Type {
+        switch self {
+        case .value:
+            return WPTableViewCell.self
+        case .textField:
+            return WPTextFieldTableViewCell.self
+        }
+    }
+}
+
+enum PrepublishingIdentifier {
+    case title
     case schedule
     case visibility
     case tags
@@ -36,12 +53,7 @@ class PrepublishingViewController: UITableViewController {
 
     private let completion: (CompletionResult) -> ()
 
-    private let options: [PrepublishingOption] = [
-        PrepublishingOption(id: .visibility, title: NSLocalizedString("Visibility", comment: "Label for Visibility")),
-        PrepublishingOption(id: .schedule, title: Constants.publishDateLabel),
-        PrepublishingOption(id: .tags, title: NSLocalizedString("Tags", comment: "Label for Tags")),
-        PrepublishingOption(id: .categories, title: NSLocalizedString("Categories", comment: "Label for Categories"))
-    ]
+    private let options: [PrepublishingOption]
 
     private var didTapPublish = false
 
@@ -52,8 +64,11 @@ class PrepublishingViewController: UITableViewController {
         return nuxButton
     }()
 
-    init(post: Post, completion: @escaping (CompletionResult) -> ()) {
+    init(post: Post, identifiers: [PrepublishingIdentifier], completion: @escaping (CompletionResult) -> ()) {
         self.post = post
+        self.options = identifiers.map { identifier in
+            return PrepublishingViewController.option(for: identifier)
+        }
         self.completion = completion
         super.init(nibName: nil, bundle: nil)
     }
@@ -61,6 +76,9 @@ class PrepublishingViewController: UITableViewController {
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    private var cancellables = Set<AnyCancellable>()
+    @Published private var keyboardShown: Bool = false
 
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -1,12 +1,29 @@
...
+private enum PrepublishingCellType {
+    case value
+    case textField
+
+    var cellType: UITableViewCell.Type {
+        switch self {
+        case .value:
+            return WPTableViewCell.self
+        case .textField:
+            return WPTextFieldTableViewCell.self
```
leandroalonso, line 20: The dynamic type is different between `WPTextFieldTableViewCell` and `WPTableViewCell`:



I'd recommend making them equal because we recently received some support requests from issues associated with big font sizes.

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -72,9 +90,25 @@ class PrepublishingViewController: UITableViewController {
...
+
+        configureKeyboardToggle()
+    }
+
+    /// Toggles `keyboardShown` as the keyboard notifications come in
+    private func configureKeyboardToggle() {
+        NotificationCenter.default.publisher(for: UIResponder.keyboardDidShowNotification)
+            .map { _ in return true }
+            .assign(to: \.keyboardShown, on: self)
+            .store(in: &cancellables)
```
leandroalonso, line 109: Noob user of `Combine` here so apologies if my question is dumb: what is the function of the `.store(in: &cancellables)` in this call?

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -164,6 +224,38 @@ class PrepublishingViewController: UITableViewController {
         tableView.reloadData()
     }
 
+    private static func option(for identifier: PrepublishingIdentifier) -> PrepublishingOption {
```
leandroalonso: It seems a bit odd that this is static, I think you can change options to be a var, like this:

```
private var options: [PrepublishingOption] = []
```

And then in the constructor:

```
    init(post: Post, identifiers: [PrepublishingIdentifier], completion: @escaping (CompletionResult) -> ()) {
        self.post = post
        self.completion = completion
        self.options = identifiers.map { identifier in
            return option(for: identifier)
        }
    }
```

This won't affect the lifecycle and will keep the VC encapsulation.

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
```diff
@@ -164,6 +224,38 @@ class PrepublishingViewController: UITableViewController {
...
+    // MARK: - Titls
+
+    private func configureTitleCell(_ cell: WPTextFieldTableViewCell) {
+        cell.textField.text = post.postTitle
+        cell.textField.textColor = .text
+        cell.textField.placeholder = "Title"
+        cell.textField.heightAnchor.constraint(equalToConstant: 40).isActive = true
+        cell.textField.autocorrectionType = .yes
+        cell.textField.autocapitalizationType = .sentences
+        cell.textField.becomeFirstResponder()
```
leandroalonso: The `becomeFirstResponder` makes this field to be selected every time the main screen appears, thus having this effect:


https://user-images.githubusercontent.com/7040243/108557855-e62a5080-72d7-11eb-8fce-200e2853d9c7.mov

I think it's a little bit confusing that the Bottom Sheet is moving around but I wouldn't consider that a block. Up to you to fix that or not!

leandroalonso (approved):
Hey @bjtitus, it all works as described!

I really recommend fixing the dynamic font size before merging this PR. The rest are just suggestions so it's up to you, sir!

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
bjtitus: Ah, good point @leandroalonso. I just hate having these properties be mutable when they don't need to be. I often rely on whether the property is mutable or not when debugging.

What do you think about doing a new initializer on `PrepublishingOption` which takes the identifier? That seems like the cleanest solution to me.

```swift
private extension PrepublishingOption {
    init(identifier: PrepublishingIdentifier) {
        switch identifier {
        case .title:
            self.init(id: .title, title: PrepublishingViewController.Constants.titlePlaceholder, type: .textField)
        case .schedule:
            self.init(id: .schedule, title: PrepublishingViewController.Constants.publishDateLabel, type: .value)
        case .categories:
            self.init(id: .categories, title: NSLocalizedString("Categories", comment: "Label for Categories"), type: .value)
        case .visibility:
            self.init(id: .visibility, title: NSLocalizedString("Visibility", comment: "Label for Visibility"), type: .value)
        case .tags:
            self.init(id: .tags, title: NSLocalizedString("Tags", comment: "Label for Tags"), type: .value)
        }
    }
}
```

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
bjtitus: No problem! I've used RxSwift in the past, so I had some exposure to this a while back.

`assign` returns an `AnyCancellable` which is like the object you would get back when calling `NotificationCenter.addObserver(... using block:)` to hold a reference to the subscription. The lifecycle of this reference determines how long the subscription is around for. So deallocating it will remove the subscription (and cancel any operations), similar to a call to `NotificationCenter.removeObserver` for an object + selector observer.

The `store` just adds the `AnyCancellable` to the set. The set (`cancellables`) is basically just a container which holds references to subscriptions. By adding it to a property on the object it will be deallocated at the same time as `PrepublishingViewController`.

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
bjtitus: I had noticed this but wasn't sure the exact sequence. I've added a property `hasSelectedText` which causes `becomeFirstResponder` to only be called the first time the sheet is shown. I think this solves the issue. 👍

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
leandroalonso: Oh, gotcha! Thank you for explaining it!

WordPress/Classes/ViewRelated/Post/PrepublishingViewController.swift
leandroalonso: I'm ok with this solution. :)

PR #15939, opened 2021-02-19
Discover posts sorting - accessibility

bozidarsevo:

To test:

Enable VoiceOver, open Reader tab, switch to Discover screen. Check if sorting option button on top of the posts is properly setup for VoiceOver. Tap the sorting option button and check if the sorting bottom sheet is properly setup for VoiceOver.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
@@ -15,7 +15,6 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
     private lazy var sortingButton: ReaderSortingOptionButton = {
         let view = ReaderSortingOptionButton()
         view.addTarget(self, action: #selector(didTapSortingButton), for: .touchUpInside)
-        view.accessibilityHint = NSLocalizedString("Tap to change sorting option", comment: "Accessibility hint for sorting option button.")
         return view
     }()
 
--- a/WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
@@ -112,10 +112,6 @@ class ReaderSortingOptionButton: UIControl {
     private func setup() {
         translatesAutoresizingMaskIntoConstraints = false
 
-        // TODO: check if accessibility is set correctly
-        accessibilityIdentifier = "Reader sorting option button"
-        accessibilityLabel = NSLocalizedString("Choose reader sorting option", comment: "Accessibility label for sorting option button")
-
         addSubview(iconView)
         addSubview(label)
         addSubview(chevronView)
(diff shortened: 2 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -69,9 +69,17 @@ class ReaderSortingActionSheetOptionControl: ClosureControl {
...
+        checkView.isAccessibilityElement = false
+        imageView.isAccessibilityElement = false
+        label.isAccessibilityElement = false
+
         accessibilityIdentifier = option.identifier
-        accessibilityLabel = option.identifier
+        accessibilityLabel = option.sortingOption.accessibilityLabel
         accessibilityHint = option.sortingOption.accessibilityHint
+        accessibilityValue = option.sortingOption.accessibilityValue
+        accessibilityTraits = UIAccessibilityTraits.button
```
bozidarsevo: I am thinking maybe to include `option.checked` for setting the accessibility properties, to tell the user that one of the items is selected/checked. Would it be better to have `accessibilityValue` something like "selected" or should I put it directly in `accessibilityLabel`?

Gio2018 (changes requested):
Hey @bozidarsevo , I ran voiceOver on a device (iPhone 11 Pro Max, iOS 14.4) and found a few issues:

- When opening the bottom sheet, the sheet handle is still detected (and announced) as a "dismiss button"
- the sorting option message is a little too verbose. Now it says "Sorting option - popular (or recent) - select to change sorting option". I believe that "Sorting option button - Popular" or similar should suffice
- Similarly in the selection sheet, I believe that mentioning the buttons (and maybe added "Selected" to the selected one) should suffice.

WordPress/Classes/ViewRelated/Reader/ReaderCardsStreamViewController.swift
```diff
@@ -15,7 +15,6 @@ class ReaderCardsStreamViewController: ReaderStreamViewController {
     private lazy var sortingButton: ReaderSortingOptionButton = {
         let view = ReaderSortingOptionButton()
         view.addTarget(self, action: #selector(didTapSortingButton), for: .touchUpInside)
-        view.accessibilityHint = NSLocalizedString("Tap to change sorting option", comment: "Accessibility hint for sorting option button.")
```
bozidarsevo, line 18: Removed the hint for the sorting option button.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionButton.swift
```diff
@@ -155,7 +155,7 @@ extension ReaderSortingOptionButton: Accessible {
 
         // TODO: check if accessibility is set correctly
         accessibilityIdentifier = "Reader sorting option button"
-        accessibilityLabel = NSLocalizedString("Sorting option button", comment: "Accessibility label for sorting option button")
```
bozidarsevo: Removed "button" from the label because it appears that VoiceOver ignores the word "button" when it is speaking, probably because it speaks "button" at the and as a result of `accessibilityTraits = UIAccessibilityTraits.button`

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -76,9 +76,11 @@ class ReaderSortingActionSheetOptionControl: ClosureControl {
         label.isAccessibilityElement = false
 
         accessibilityIdentifier = option.identifier
-        accessibilityLabel = option.sortingOption.accessibilityLabel
-        accessibilityHint = option.sortingOption.accessibilityHint
-        accessibilityValue = option.sortingOption.accessibilityValue
+        accessibilityLabel = NSLocalizedString("Sorting option", comment: "Accessibility label for sorting option")
```
bozidarsevo, line 79: Since VoiceOver speaks in this order:
- label
- value
- traits
- hint

I made it so that label is generic text "Sorting option", value gets to be "Recent" or "Popular", and I use the hint to say if the option is selected or not.

WordPress/Classes/ViewRelated/Reader/Sorting/ReaderSortingOptionSheetViewController.swift
```diff
@@ -259,7 +261,7 @@ class ReaderSortingOptionViewController: UIViewController {
 
 extension ReaderSortingOptionViewController: Accessible {
     func prepareForVoiceOver() {
-        gripButton.accessibilityLabel = NSLocalizedString("Dismiss", comment: "Accessibility label for the grip button which is used to dismiss the dialog.")
+        gripButton.isAccessibilityElement = false
```
bozidarsevo, line 264: @Gio2018 I have set the sheet handle not to be accessibility element, so we do not have VoiceOver for it. Let me know if I should remove the logic for dissmissing the sheet if user taps it (as mentioned on slack)

Gio2018 (approved):
Sounds good!!

PR #15941, opened 2021-02-19
Fix wpios-env.default path in Fastlane error msg

guarani:
This tweaks a Fastlane error message to indicate that `wpios-env.default` lives in `fastlane/env/user.env-example`, not `env/user.env-example`, as it previously suggested.

To test:
You could switch to the `develop` branch, delete your `~/.wpios-env.default` file, run a Fastlane command such as `bundle exec fastlane register_new_device`, and notice the incorrect path in the error message.
Then switch to this branch, do the same steps, and notice the error message is now correct.

PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -44,7 +44,7 @@ before_all do |lane|
   unless lane == :test_without_building
     # Check that the env files exist
     unless is_ci || File.file?(USER_ENV_FILE_PATH)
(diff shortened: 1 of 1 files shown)


jkmassel (approved):
Thanks for doing this @Guarani!!

guarani:
Thanks @jkmassel, I change the milestone to 16.8 since you reviewed it already 👍

PR #15926, opened 2021-02-18
Merge 16.7 beta 5 to develop

jkmassel:
Merges all changes in the `16.7` beta series into `develop`

PR #15948, opened 2021-02-22
Merge 16.7 final into develop

jkmassel:


AliSoftware:
Looks suspicious 🧐 that:
- there's only 5 of the Mag16 having `release_notes.txt` files in that diff
- There's a lot of `.strings` files added in this PR that are related to TodayWidget and ShareExtension that have been added but are empty. (iinm from last time I checked, if a `strings` file exists for a locale it will be loaded and used, and then if a key is not present in that file it will fallback to use the key as the copy. As opposed to if the strings file doesn't exist at all it will try to find another fallback locale that exists.)

mokagio:
> there's only 5 of the Mag16 having release_notes.txt files in that diff

I'd say that's because this release had 5 betas? In #15926 there are other release notes touched.

loremattei:
> there's only 5 of the Mag16 having release_notes.txt files in that diff

Hmm... yes, this looks weird as there are far more in GlotPress.

loremattei:
> I'd say that's because this release had 5 betas? In #15926 there are other release notes touched.

It should not have any effect on this: the `.po` file was not updated, which was the right thing to do because it broke all of the translations otherwise given how late in the cycle we were.

AliSoftware:
If the release notes were updated during the `release/16.7` branch – while the `.po` file wasn't, which is indeed good because it was too late – we might want to check if the Peril rule that's supposed to warn about it worked and if it's enough of a communication for code wranglers to realize that their new entries will be ignored (won't be translated nor make it to the final notes) 🤔

loremattei:
No, the changes aren't in the release branch. They seem to come from merging `develop`.

jkmassel:
There are only 5 release notes changes because I merged #15926 the day before, which contained many metadata localization changes FWIW 😅

AliSoftware:
Cool, that explains it then. What about the **empty** `.strings` files for sharedextensions / widgets though? (which were _also_ added to #15926 apparently)

AliSoftware (approved):
After a chat with @jkmassel in Slack it seems that there's a runtime, manually-implemented fallback mechanism for the strings in our code for the Widget that will explicitly fall back to `en-US` if a translation is not found.

Not ideal to have to hardcode it ourselves, especially since Locale and ICU fallback rules are more complex that that and it's usually nicer to let the OS do the fallback logic for us than try to re-create it. But our tooling is not ready for that, so… this will have to do for now 🤷

PR #15918, opened 2021-02-18
Update lanes to build stable and beta releases

loremattei:
This PR updates `build_and_upload_release` so that it doesn't expect the tag to be already present in the repository. 
The user can now switch between beta and stable builds passing the right value to the new `beta_release` parameter. 

When `create_gh_release` is passed with a `true` value, the lane will create the draft of the release on GitHub associating the HEAD commit of the branch to it. When the release manager releases the draft, GitHub creates the actual tag in the repository. This means that we don't need to give more access right to the CI to be able to push the tag as we get the tag "for free" with out current process. 

The PR also adds two new lanes `build_and_upload_beta_release` and `build_and_upload_stable_release` which invoke `build_and_upload_release` with the correct value of the `beta_release` flag.

Also, this PR updates `Fastlane` to the latest. 

A follow up PR will take care of updating the CI configuration and the related Fastlane lanes.

**To test:**
1. Create a new test branch out of this one. 
2. Update the Internal and Public versions in the xcconfig files. Please, use a version that is not used in production (anything with a minor number > 9 will work).
3. Comment out the code to upload [to iTC](https://github.com/wordpress-mobile/WordPress-iOS/blob/944ac9db29e6be4c5f0519195b3f192bb8b2edc0/fastlane/Fastfile#L440) and [to AppCenter](https://github.com/wordpress-mobile/WordPress-iOS/blob/944ac9db29e6be4c5f0519195b3f192bb8b2edc0/fastlane/Fastfile#L391). These changes don't affect that part of the code, so we don't need to upload dummy stuff, especially to TestFlight where it can be removed. 
4. Run `bundle exec fastlane build_and_upload_beta_release create_gh_release:true` and verify that the build succeeds and that releases are drafted in GitHub for the beta (_Note_: don't release it or a tag will be added and the build started on CI and uploaded)
5. Run `bundle exec fastlane build_and_upload_stable_release create_gh_release:true` and verify that the release is drafted for stable  (_Note_: don't release it or a tag will be added and the build started on CI and uploaded).
6. Cleanup all the dummy releases on GitHub.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #15953, opened 2021-02-22
Switch Fastlane to API calls 

loremattei:
This PR wraps up the changes to trigger builds on CI via API calls on WPiOS. It's built on top of https://github.com/wordpress-mobile/WordPress-iOS/pull/15918 and should be merged after that one. 

The PR removes all the steps which push tags in `Fastlane` and replace them with API calls to the CI. 
CircleCI's `config.yml` is updated with filters which run the build lanes only on the proper calls. 

**To test**:
Testing this one is tricky. I think that it's quite easy to test the critical points, but a complete test of the flow would require many small tricks/adjustments. So: 

1. Verify CI is green. 
2. Double check that, in `Fastfile`, every call to `ios_tag_build()` and `ios_final_tag()` has been replaced with a call to `trigger_beta_build` and `trigger_release_build`.
3.  Create a new `release/test` branch out of this one and push it.
4. Run `bundle exec fastlane trigger_beta_build branch_to_build:"release/test"` (which is the same lane invoked by the other lanes in `Fastfile`) and verify that the build is triggered _and_ that the standard PR jobs are not triggered. Note that the build will likely fail because the built version already exists.
5. Remove the `release/test` branch and clean up.  

PR submission checklist:


jkmassel (approved):
I tested it and this performed exactly as expected for me.

Issue #15955, opened 2021-02-22
Comments: filter by status


ScoutHarris:
Add the ability to filter Comments by status.

Tasks:
- [x] Add `FilterTabBar` to `CommentsViewController`. [#15966]
- [x] Fetch comments for each filter. [wordpress-mobile/WordPressKit-iOS/pull/357, #15972]
- [x] Update UI to display all Comments correctly for each filter. [ #15998]
- [x] Add state restoration. [#16001]
- [x] Show offline NRV. [#16105]
- [x] Save selected filter. [#16108]
- [x] Track filter selection. [#16108]
- [x] Enable feature flag, add release note. [#16130]

Design: (V1 pbArwn-1EF)

![comment_enhancements](https://user-images.githubusercontent.com/1816888/108762951-f7f53900-750d-11eb-9e2b-a55e494f9869.png)

PR #15966, opened 2021-02-23
Comments: add filter tab bar

ScoutHarris:
Ref: #15955

This adds a `FilterTabBar` at the top of the Comments view. 
- `CommentsViewController` is now a `UIViewController` instead of a `UITableViewController`.
- `CommentsViewController` shows the new `CommentsList.storyboard`, which contains a `FilterTabBar` and `UITableView`.
- Selecting different filters doesn't do anything yet.

To test:
- Go to Site > Comments.
- Verify the filter tab bar is not shown.

<kbd>no_filter</kbd>

- Enable the feature flag.
- Go to Site > Comments.
- Verify the filter tab bar is shown.
- To show tab selection actually works, the tab index is logged out. I'll remove this before merging.
`🔴 tab selected:  1`

<kbd>filter</kbd>

PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -1555,8 +1551,7 @@ - (void)scrollToElement:(QuickStartTourElement) element
 - (void)showComments
 {
     [WPAppAnalytics track:WPAnalyticsStatOpenedComments withBlog:self.blog];
-    CommentsViewController *controller = [[CommentsViewController alloc] initWithStyle:UITableViewStylePlain];
-    controller.blog = self.blog;
+    CommentsViewController *controller = [CommentsViewController controllerWithBlog:self.blog];
```
aerych, line 1554: Nice

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -36,20 +42,34 @@ - (void)dealloc
...
+    CommentsViewController *controller = [storyboard instantiateInitialViewController];
+    [controller setupWithBlog:blog];
+    return controller;
+}
+
+- (void)setupWithBlog:(Blog *)blog
+{
+    self.blog = blog;
+    self.restorationClass = [self class];
+    self.restorationIdentifier = NSStringFromClass([self class]);
```
aerych: I want to say whenever we have a storyboard or a xib for a view controller we tend to set the restoration identifier via the identity panel. We could do the same here for consistency now that there's a storyboard.  Or not.  Up to you. :)
Screen Shot 2021-02-24 at 9 54 19 AM

aerych (changes requested):
Howdy @ScoutHarris 👋 
I took the branch for a spin.  Code wise everything looks good, just one nitpicky suggestion that is totes optional. 
I spotted something when testing on the iPhone and rotating between portrait and landscape orientation.  It looks like after rotation the tableview is gets pushed down when the tab bar is featured flagged.  This doesn't happen on the iPad as far as I could tell. Not sure if there is a tweak to the constraint that's needed or if a stackview might be more helpful. Either way back to you! :)
Simulator Screen Shot - iPhone 11 Pro - 2021-02-24 at 09 49 25
Simulator Screen Shot - iPhone 11 Pro - 2021-02-24 at 09 49 34

ScoutHarris:
Hey @aerych .

Thanks for noticing the rotation issue. Yea, I made the constraint business way too complicated. 😄 Tis been fixed. As well as the restoration ID suggestion.

Ready for another go please!

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -36,20 +37,22 @@ - (void)dealloc
...
 {
-    self = [super init];
-    if (self) {
-        self.restorationClass = [self class];
-        self.restorationIdentifier = NSStringFromClass([self class]);
-    }
-    return self;
+    NSParameterAssert([blog isKindOfClass:[Blog class]]);
+    UIStoryboard *storyboard = [UIStoryboard storyboardWithName:@"CommentsList" bundle:nil];
+    CommentsViewController *controller = [storyboard instantiateInitialViewController];
```
aerych, line 44: Cool that we're setting the restoration identifier via the storyboard, but for state restoration to work we'd need to still set the restoration class. 
```
controller.restorationClass = [controller class];
```
However, when I tested this it resulted in a crash due to `self.blog` being nil after restoration.  The methods to encode the blog were apparently never implemented, and since there was no prior crash I'm doubtful state restoration was working correctly. 🤔 

To wire up restoration we'd need to implement something like: 
```
-(void)encodeRestorableStateWithCoder:(NSCoder *)coder
{
    [coder encodeObject:[self.blog hostname] forKey:@"blog-hostname"];
    [super encodeRestorableStateWithCoder:coder];
}

- (void)decodeRestorableStateWithCoder:(NSCoder *)coder
{
    NSString *hostname = [coder decodeObjectForKey:@"blog-hostname"];
    BlogService *service = [BlogService serviceWithMainContext];
    self.blog = [service blogByHostname:hostname];
    [super decodeRestorableStateWithCoder:coder];
}
```
However, this still had an issue as `[self viewDidAppear]` is executed prior to `[self decodeRestorableStateWithCoder]` and `self.blog` was still nil.  

Adding a check for the `blog` in `viewDidAppear` and implementing 
```
- (void)applicationFinishedRestoringState
{
    [super applicationFinishedRestoringState];
    [self refreshNoResultsView];
    [self refreshAndSyncIfNeeded];
}
```
Avoided the crash, however no results were being returned.  I'm assuming because `fetchRequest` captured the nil blog and needed to be reset but this is the point where I stopped tinkering. 

I think since restoration was probably busted already we can just make a note of all of this and address later (as alluded to in Slack). :)

aerych (approved):
Hola @ScoutHarris 👋 
Retested and the rotation issue is resolved :)
I left notes about state restoration being cranky but this seems like an existing issue so please :shipit:  when ready and we'll sort it out later. :)

ScoutHarris:
Thanks @aerych !

PR #15967, opened 2021-02-24
Update WordPressAuthenticator pod to fix prologue tracking issue

frosty:
This PR updates the version of WordPressAuthenticator to fix an issue introduced when we switched to displaying the authenticator as the window's root view controller instead of modally. That change meant that the previous check for `isBeingPresentedInAnyWay` always returned false, so we didn't track the prologue step.

The WordPressAuthenticator update reverts to the previous method of tracking whether this is the first time the view controller has appeared, using a simple property.

**To test**

- Add breakpoints to lines 102 and 105 in `LoginPrologueViewController.swift`

image

- Build and run.
- If you're already logged out, ensure that the breakpoint on line 102 is hit. Otherwise, log out and ensure the same.
- Tap one of the buttons on the prologue screen, and then tap Back to return to the prologue screen.
- Ensure that the other breakpoint is hit.
- You should only see one console log for the prologue unified_login_step tracking:

```
🔵 Tracked: unified_login_step <flow: prologue, source: default, step: prologue>
```

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -198,9 +198,9 @@ target 'WordPress' do
 
     pod 'Gridicons', '~> 1.1.0'
 
-    # pod 'WordPressAuthenticator', '~> 1.35.1'
+    pod 'WordPressAuthenticator', '~> 1.35.2'
     # While in PR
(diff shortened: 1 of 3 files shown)


emilylaguna (approved):
🚀 https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/578#pullrequestreview-597554176

PR #15974, opened 2021-02-25
Stories: Add missing analytics events + Enable Story Block

bjtitus:
* Adds several missing analytics events which are necessary for stories reporting.
* Enables the story block in the Gutenberg editor.

### Story block
1. Log in to a WordPress.com site or site with Jetpack >= 9.1
2. Create a new blog post
3. Tap to add a new block
4. Verify that the Story block appears from the list

### Analytics
1. Open the Stories editor and confirm that the following events are tracked on open:

```
editor_post_created {... editor_source: stories...}
```

2. Verify the following events when publishing a new story:

```
editor_post_publish_tapped
editor_post_publish_dismissed
editor_post_published {"editor_source":"wp_stories_creator" ... "has_wp_stories_blocks":true}
```

3. Open an existing story post and edit a story block and confirm that the following event is tracked:

```
editor_session_start {... editor: stories ...}
```

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/Stories/StoryEditor.swift
+++ b/WordPress/Classes/Services/Stories/StoryEditor.swift
@@ -136,7 +136,7 @@ class StoryEditor: CameraController {
         super.init(settings: settings,
                  mediaPicker: WPMediaPickerForKanvas.self,
                  stickerProvider: nil,
-                 analyticsProvider: KanvasAnalyticsHandler(editorAnalyticsSession: analyticsSession),
+                 analyticsProvider: KanvasAnalyticsHandler(),
                  quickBlogSelectorCoordinator: nil,
                  tagCollection: nil,
                  saveDirectory: saveDirectory)
@@ -198,6 +198,10 @@ class StoryEditor: CameraController {
                 completion(.success(()))
             }
         }
+    }
+
+    func trackOpen() {
+        editorSession.start()
     }
 }
 
--- a/WordPress/Classes/Services/Stories/StoryEditor.swift
+++ b/WordPress/Classes/Services/Stories/StoryEditor.swift
@@ -8,9 +8,7 @@ class StoryEditor: CameraController {
 
     var onClose: ((Bool, Bool) -> Void)? = nil
 
-    lazy var editorSession: PostEditorAnalyticsSession = {
-        PostEditorAnalyticsSession(editor: .stories, post: post)
-    }()
+    var editorSession: PostEditorAnalyticsSession
 
     let navigationBarManager: PostEditorNavigationBarManager? = nil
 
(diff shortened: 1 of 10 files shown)


aforcier:
I gave the analytics a check through, from create sheet (from My Site and from the post list) -> Stories intro -> story editor -> published post, everything looks good there! 👍

I did run into some trouble with editing though: I don't see an `editor_session_start` event when editing a story block from Gutenberg. (No event fires at all once GB has loaded.)

I do see a `editor_session_end <content_type: gutenberg, editor: stories, outcome: cancel, post_type: post>` if I press the X button to dismiss the story editor and return to GB.

(Side note: GB in the simulator behaves pretty oddly for me, the app is frozen for a solid minute when tapping 'Edit' on a post or anything that launches GB before the post finally loads in GB.)

aforcier:
One other thing that might cause issues later:

```
editor_post_created {post_type: story}
```

I think this is potentially misleading since `post_type` is otherwise used by that event to mean the actual (WordPress) post type of the post - which in the stories case is still a `post`. So there's a bit of ambiguity when attempting to segment this event by `post_type`, since `story` isn't a real post type.

What if we kept it as `post_type: post`, but added the `editor: stories` property to this event?

aforcier:
Gave the new changes a try, confirmed that:

* `editor_post_created` now has `post_type: post` and `editor_source: stories` for stories
* The `editor_session_start` event is triggered when editing a story from GB, with the appropriate properties
* The story block now shows up in the GB block picker (unless the current site is a self-hosted non-Jetpack site)
* The `story_block_add_media_tapped` event fires when 'Add media' is tapped in a newly added story block
* Previously checked events are still firing correctly

I do wonder if we should be consistent about naming for the `editor_source` property - sometimes it's `stories`, sometimes it's `wp_stories_creator`. All things being equal `wp_stories_creator` might be best just for consistency with Android, but if it's easier to normalize to `stories` for any reasons that's fine too IMO. WDYT @bjtitus ?

bjtitus:
@aforcier Yep, I had used `wp_stories_creator` in several instances to match Android, but the editors are in an enum which uses single words, which is why other spots use "stories".

There are also several spots where "editor" is used instead of "editor_source" but those seem to be pre-existing so I would leave those.

Gio2018 (approved):
:shipit: !

aforcier:
Just noticed an issue with the `editor_post_published` event - it needs to include the remote post ID of the post as a `post_id` property.

On Android this happens [here](https://github.com/wordpress-mobile/WordPress-Android/blob/4e01062e871ae2031b5cfd5cc0b27d9f5366f7b4/WordPress/src/main/java/org/wordpress/android/ui/uploads/PostUploadHandler.java#L681).

bjtitus:
After talking with @aforcier we may wait to fix the `editor_post_published` event. It looks like it has been broken (not filling out the `post_id`) for some time in the normal Gutenberg editor as well. Basically, we log the event before the post is fully published so we don't have the remote data.

I have a fix which tracks the event when `NoticeAction` occurs — the same event the notice UI uses to notify the user of a published post. I think this should probably apply to both editors but it feels a little weird using a Flux Action intended for UI feedback to handle logging.

Issue #15662, opened 2021-01-18
Like notifications: show all likes when notification tapped


ScoutHarris:
When tapping a Like notification, the app currently show a max of 10 users. Let's show them all.

---
Tasks:
- [x] Add feature flag (`newLikeNotifications`). [#16007]
- [x] Add endpoint support to fetch all post likes. [wordpress-mobile/WordPressKit-iOS/pull/347]
- [x] Add endpoint support to fetch all comment likes. [wordpress-mobile/WordPressKit-iOS/pull/351]
- [x] Add service support to fetch post likes. [#15977]
- [x] Add service support to fetch comment likes. [#15994]
- [x] Show all likes for posts. [#16007]
- [x] Show all likes for comments. [#16007]
- [x] Update WPKit to use v1.2 endpoints to get `preferred_blog`. [#16353, wordpress-mobile/WordPressKit-iOS/pull/389]
- [x] Store like information in Core Data. [#16361, #16383, #16394, #16398, #16399, #16409]
- [x] Update likes list and user profile sheet to use info from Core Data. [#16410]
- [x] Order the likes list newest to oldest. [#16409]
- [x] ~~Show liker's blog title if available, and display blog when cell tapped.~~ (Ref: wordpress-mobile/WordPressKit-iOS/pull/361) (Replaced by #16249)
- [x] Show error when fetching fails. [#16495]
- [x] Add/verify Tracks. [#16327, #16452]

---
The current view:

![limited_likes](https://user-images.githubusercontent.com/1816888/104964778-5cb2f600-599a-11eb-95db-f279ecc1abcf.png)

---
Endpoints:

Currently, we are using the `/notifications` endpoint to get Notifications. Each notification contains a `body` property that contains the content displayed in notification details. This content is limited. To get full lists, we'll probably need to refactor the notification details to use these endpoints:

- Get likes for a post: https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post_ID/likes/
- Get likes for a comment: https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/comments/%24comment_ID/likes/

These endpoints are not implemented, so they'll need to be added to https://github.com/wordpress-mobile/WordPressKit-iOS.

PR #15977, opened 2021-02-25
Implement fetch likes in PostService

bitdc:
Refs #15662 

Add implementation to fetch likes for a post in `PostService`. 

To test:

Confirm the app builds. The rest are covered through unit tests as this is a "backend" change.

PR submission checklist:


WordPress/Classes/Services/PostService.m
```diff
@@ -676,6 +676,26 @@ - (void)restoreRemotePostWithPost:(AbstractPost*)post
     [remote restorePost:remotePost success:successBlock failure:failureBlock];
 }
 
+- (void)getLikesForPost:(AbstractPost *)post
```
ScoutHarris: I don't know how you're going to call this method, but I wonder if this should be postID instead of AbstractPost. NotificationsVC and NotificationsDetailVC get the PostID from the notification. So it looks like you'd have to create an AbstractPost just to fetch likes. But that's not really necessary since fetching only requires the ID.

bitdc: Ah yes, I was actually in two minds on fetching the likes through `AbstractPost` or simply `postID` and `siteID`. My initial thought was, using `AbstractPost` would reuse most of the existing logic like `PostServiceRemoteFactory` that accepts a `Blog` object, and also makes use of the existing mocking functionalities on the unit testing side. 

As for how the caller utilizes this method, I assume the app would synchronize blogs when the app is opened; and information for the `Blog` and `AbstractPost` can be obtained through `BlogService`:
```objc
- (nullable Blog *)blogByBlogId:(NSNumber *)blogID;
``` 
and `PostService` respectively: 
```objc
- (nullable AbstractPost *)findPostWithID:(NSNumber *)postID inBlog:(Blog *)blog;
```

However, after giving it a second thought, going with just `postID` and `siteID` will make the implementation more reusable should we decide to implement the Likes page from the Reader page in the future – since Reader utilizes a different model (i.e. `ReaderPost`). 

Thanks for your input @ScoutHarris, and I'll make the necessary adjustments shortly! 🙂

ScoutHarris:
Hey David.

This branch isn't building. It's failing looking for `CommentStatusFilter`, which was recently added to WordPressKit. Maybe your WPKit `feature/15662-notification_likes_support` is not up to date with our WPKit `feature/15662-notification_likes_support`?

Also, when I installed pods in this branch, I had a local change to `Podfile.lock`. You'll want to make sure this is up-to-date in your branch.

bitdc:
I've updated the `Podfile.lock`. Thanks for notifying!

WordPress/Classes/Services/PostServiceRemoteFactory.swift
```diff
@@ -16,4 +16,36 @@ import WordPressKit
...
+    // MARK: Private methods
+
+    /// Get the api to use for making REST requests.
+    ///
+    /// - Returns: an instance of WordPressComRestApi
+    private func apiForRESTRequest(using context: NSManagedObjectContext) -> WordPressComRestApi? {
+        let accountService = AccountService(managedObjectContext: context)
+
+        guard let account = accountService.defaultWordPressComAccount(),
+              let api = account.wordPressComRestApi else {
```
ScoutHarris, line 37: This could just be one line:
`guard let api = accountService.defaultWordPressComAccount()?.wordPressComRestApi`

ScoutHarris (approved):
I just had one small suggestion. Otherwise, I think this will work.

Let me know what you decide on the suggestion, and I'll merge this accordingly.

ScoutHarris:
Hey David. I'll go ahead and merge this so you can proceed. Please make a note of the suggestion for your next PR. Thanks!

WordPress/Classes/Services/PostServiceRemoteFactory.swift
bitdc: Noted!

PR #15980, opened 2021-02-26
Stories: Fix Media Import bugs

bjtitus:
* Fixes a bug where media was unnecessarily exported (instead of copied/reused) when adding to story.
* Adds a Loading indicator during media import so there is some indicator while media files are being exported.
* Allows video uploads for local URLs so that Stories media can be uploaded to the media library.

### Testing
* Go to the Blog Details page
* Create a new Story Post
* Add several pieces of media (with at least 1 video) from your device
* Ensure that media is imported quickly into the Stories editor
* Publish the post
* Ensure that video is uploaded and published in story

## Code Overview

PR submission checklist:


WordPress/Classes/Utility/Media/MediaURLExporter.swift
```diff
@@ -77,10 +77,9 @@ class MediaURLExporter: MediaExporter {
     ///
     func exportURL(fileURL: URL, onCompletion: @escaping OnMediaExport, onError: @escaping OnExportError) -> Progress {
         // Verify the export is permissible
-        if let urlExportOptions = urlOptions,
-            !urlExportOptions.allowableFileExtensions.isEmpty,
-            let fileExtension = fileURL.typeIdentifierFileExtension {
-            if !urlExportOptions.allowableFileExtensions.contains(fileExtension) {
+        if let fileExtension = fileURL.typeIdentifierFileExtension {
+            // Check the default file types. We want to limit to supported types but mobile is not restricted to only allowed ones.
+            if !MediaImportService.defaultAllowableFileExtensions.contains(fileExtension) {
```
bjtitus, line 82: Checked with @aforcier that Android does not appear to be using the `allowed_file_types` from the blog.

On iOS, we don't check the list for PHAssets but do for media which comes from a URL. This needs to change for Stories since we are uploading video which came from Photos (or WP Media) and then ended up re-exported with modifications.

The `MediaImportService.defaultAllowableFileExtensions` value is still used so that we still check for supported file types.

WordPress/Classes/Services/Stories/WPMediaPicker+MediaPicker.swift
```diff
@@ -161,38 +165,52 @@ extension WPMediaAsset {
...
 
             if let assetURL = urlAsset?.url {
+                let existingExtension = assetURL.pathExtension
+                let exportURL = url.appendingPathExtension(existingExtension)
                 if urlAsset?.url.scheme != "file" {
                     // Download any file which isn't local and move it to the proper location.
                     return URLSession.shared.downloadTaskPublisher(url: assetURL).tryMap { (location, _) -> URL in
                         if let location = location {
-                            try FileManager.default.moveItem(at: location, to: url)
+                            try FileManager.default.moveItem(at: location, to: exportURL)
```
bjtitus, line 235: It is safe to move the item here since we just downloaded it.

WordPress/Classes/Services/Stories/WPMediaPicker+MediaPicker.swift
```diff
@@ -161,38 +165,52 @@ extension WPMediaAsset {
...
+                        } else {
+                            return url
                         }
-                        return url
                     }.eraseToAnyPublisher()
                 }
-                try FileManager.default.moveItem(at: assetURL, to: url)
-                return Just(url).setFailureType(to: Error.self).eraseToAnyPublisher()
+                do {
+                    try FileManager.default.copyItem(at: assetURL, to: exportURL)
```
bjtitus: We must copy the item instead of moving because we don't have permission to move files from the Photos library.

WordPress/Classes/Services/Stories/WPMediaPicker+MediaPicker.swift
```diff
@@ -161,38 +165,52 @@ extension WPMediaAsset {
     func videoURLPublisher() -> AnyPublisher<URL, Error> {
         return videoAssetPublisher().tryMap { asset -> AnyPublisher<URL, Error> in
             let filename = UUID().uuidString
-            let url = try FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true).appendingPathComponent(filename).appendingPathExtension("mp4")
+            let url = try FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true).appendingPathComponent(filename)
             let urlAsset = asset as? AVURLAsset
 
             if let assetURL = urlAsset?.url {
+                let existingExtension = assetURL.pathExtension
+                let exportURL = url.appendingPathExtension(existingExtension)
```
bjtitus: Uses the Media's original extension instead of hardcoding it.

aerych: Maybe `url.appendingPathExtension(assetURL.pathExtension)` since we don't use `existingExtension` for anything else?  Up to you.

WordPress/Classes/Services/Stories/WPMediaPicker+MediaPicker.swift
```diff
@@ -161,38 +165,50 @@ extension WPMediaAsset {
     func videoURLPublisher() -> AnyPublisher<URL, Error> {
         return videoAssetPublisher().tryMap { asset -> AnyPublisher<URL, Error> in
             let filename = UUID().uuidString
-            let url = try FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: true).appendingPathComponent(filename).appendingPathExtension("mp4")
+            let storiesURL = try MediaFileManager.cache.directoryURL().appendingPathComponent("Stories", isDirectory: true)
```
aerych: It doesn't look like "Stories" is used anywhere else but are there plans to?  Maybe make this a constant if so?

aerych (commented):
Heya @bjtitus 👋 

I had one or two nitpicks/questions code wise. I need to brush up on my Combine. :) 

I tested on an iPhone 11 Pro device running 14.4.  There were a few things I noticed that I wanted to share. 

- During one test, I selected three images and two videos. The HUD stayed up for a long time
and I noticed that the preview button was tappable... so I tapped it.  I was then stuck on a different screen with no way back.  I didn't think to get a screenshot and I haven't been able to reproduce.  Just FYI. 
- I noticed that if I'm on the blog details screen when I start a story, when I go to publish the story, it appears to fail silently.  It looks like media is uploaded but not the post. The post never appears in the post list and I don't see the post publish nudges.  However, when I create a story from the Post list, everything works.  I can also create a regular post from the blog detail screen just fine. 
- Video frames load super fast.  Image frames are very slow by comparison. 

Thoughts on these?

WordPress/Classes/Services/Stories/WPMediaPicker+MediaPicker.swift
bjtitus: I decided to reuse this directory name for the `saveDirectory` used in `StoryEditor` in b29af92. This should keep the Stories files separated out in case we need to do any cleanup or rearranging later on.

`saveDirectory` is technically temporary (storing archives before media is uploaded and receives its remote media ID), but we want to be in control so we aren't using the system temp directory.

bjtitus:
Thanks @aerych!

> The HUD stayed up for a long time and I noticed that the preview button was tappable

I've hidden those buttons while the HUD is up and added some error handling if the import does fail for some reason. It may be hard to get it to happen again since the image resizing change (mentioned below) should make import much quicker.

> I noticed that if I'm on the blog details screen when I start a story, when I go to publish the story, it appears to fail silently.

🤔 I'm not sure why this would happen and I'm having trouble reproducing.

> Image frames are very slow by comparison

I've made a modification to resize images to match the screen size. This mirror Android's behavior and the behavior of the camera UI. Should be much quicker now! This changes includes a duplicate of the `image(with size:...` method in `WPMediaPicker`, which I will migrate over in a new PR tomorrow. The resized image doesn't always match the dimensions exactly so we need to wait for the non-degraded version

WordPress/Classes/ViewRelated/Gutenberg/Processors/GutenbergMediaFilesUploadProcessor.swift
```diff
@@ -35,7 +35,11 @@ class GutenbergMediaFilesUploadProcessor: Processor {
             }
 
             do {
-                try FileManager.default.moveItem(at: URL(string: mediaFile.url)!, to: newURL)
+                if let mediaURL = URL(string: mediaFile.url) {
+                    try FileManager.default.moveItem(at: mediaURL, to: newURL)
+                } else {
+                    DDLogError("No Media File URL was present. This is Stories error and should be investigated.")
```
aerych: `This is Stories error and should be investigated` > `This is a Stories error and should be investigated`

aerych (approved):
Heya @bjtitus 👋 
Changes look good from here.  Image frames are much snappier and seem consistent with video loading.  I wasn't able to reproduce the issue with launching the stories editor from the blog details screen.  Not sure what that was about. 
I had one nitpick about logging but otherwise :shipit: from me after resolving the merge conflict. :)

PR #15982, opened 2021-02-26
Stories: Android compatibility

bjtitus:
Fixes cross compatibility issue with Android. Prior to this change, Android stories would not open in the Stories editor.

Additional changes:
- Fixes a leak of `StoryEditor`
- Fix the mime type of uploaded files

## Testing
* Create a story with the Android application
* Open the same post from iOS
* Attempt to edit the Story (tap on the Edit button on the block)
* Story should load with flattened images.

### Story in Reader
* Open a created story in the Reader.
* Verify that the thumbnail image is displayed.

PR submission checklist:


aforcier:
Can't speak to the code changes necessarily (though they make sense to me at a glance), but I tested the functionality and confirm all this works:

* Editing a story created on Android correctly loads the story with flattened media
* Creating a story on iOS, adding text, and then editing it loads the story with layered media
* Editing a story created on Android, adding text, saving, then editing it again on iOS loads the story with the added text layer unflattened

:shipit: from my end

WordPress/Classes/ViewRelated/Gutenberg/Processors/GutenbergMediaFilesUploadProcessor.swift
```diff
@@ -30,7 +30,7 @@ class GutenbergMediaFilesUploadProcessor: Processor {
                 return mediaFile
             }
 
-            guard let newURL = StoryPoster.filePath?.appendingPathComponent("\(self.serverMediaID)") else {
+            guard let newURL = StoryPoster.filePath?.appendingPathComponent(self.serverMediaID.description) else {
```
aerych: Maybe `String(self.serverMediaID)` rather than a call to `description`.  I'm not sure if calling `description` would be a poor practice, but the explicit cast seems more straightforward, and we're not privy to how `description` is implemented under the hood.

WordPress/Classes/ViewRelated/Gutenberg/Processors/GutenbergMediaFilesUploadProcessor.swift
```diff
@@ -42,7 +42,7 @@ class GutenbergMediaFilesUploadProcessor: Processor {
 
             let file = MediaFile(alt: mediaFile.alt,
                                   caption: mediaFile.caption,
-                                  id: Double(self.serverMediaID),
+                                  id: self.serverMediaID.description,
```
aerych: Same here.

aerych (changes requested):
Heya @bjtitus 👋 
I had one very nitpicky think to ask ask about the code changes. 
When testing the branch in the simulator I noticed that the separate frames were showing up but the images appeared black.  I'm not sure if this is a simulator quirk or not.  I also noticed that I couldn't open the story editor a second time after I closed it.  I tried with a newly created story, and with an story I created earlier.  Here's a gif of what I saw.   Thoughts?
Screen Recording 2021-02-26 at 2 19 41 PM 2021-02-26 14_29_01

bjtitus:
> I noticed that the separate frames were showing up but the images appeared black. I'm not sure if this is a simulator quirk or not.

I believe this is a simulator issue. Sometimes the Metal context is not properly created in the simulator. I think it depends on which specific simulator you are using, but it's something I haven't spent too much time on since it works on device.

> I couldn't open the story editor a second time after I closed it

This was a bug! I've fixed it in the last commit on this branch and also a memory leak which was occurring. Entering edit mode multiple times in a row should work now.

aforcier:
Just a note confirming that the mime/file type fix from a4c4a02 is working as expected, and static fallback previews for stories (e.g. in the Reader, or with JS disabled on the site) aren't blank anymore. 👍

bjtitus:
@aerych As Alex mentioned, I made a small tweak to fill out the mime type of the media in a different way. I've added a new testing instruction to test the Story post in the Reader, if you are able to just check that on any _newly created_ story.

Thanks!

aerych:
Heya @bjtitus  I shared a couple of videos in slack of what I'm seeing.  
Code changes look good apart from the `GutenbergViewController.swift` conflict.  Lemme know what you think about what I shared in slack if it's something to address here or in an other PR. :)

aerych (approved):
Tested again.  Everything seems to be working as expected. :) 
:shipit:  from me!

PR #15988, opened 2021-02-26
Fix broken blog lookup from Objective C

jkmassel:
Prior to this commit, Blog lookup from Objective C could crash if passed a `nil` NSNumber, because the Objective-C to Swift bridge doesn’t have enough type information to complain about it being an invalid operation.

To fix it, we convert the `NSNumber` object to a primitive – if the `NSNumber` is nil, calling `int64Value` will return `0` which we can look up using the existing code path.

To test:
- Ensure CI tests pass
- Try logging in and out from this branch

PR submission checklist:

### DIFF
--- a/WordPress/WordPress.xcodeproj/project.pbxproj
+++ b/WordPress/WordPress.xcodeproj/project.pbxproj
@@ -199,9 +199,9 @@
 		17BB26AE1E6D8321008CD031 /* MediaLibraryViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17BB26AD1E6D8321008CD031 /* MediaLibraryViewController.swift */; };
 		17BD4A0820F76A4700975AC3 /* Routes+Banners.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17BD4A0720F76A4700975AC3 /* Routes+Banners.swift */; };
 		17BD4A192101D31B00975AC3 /* NavigationActionHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17BD4A182101D31B00975AC3 /* NavigationActionHelpers.swift */; };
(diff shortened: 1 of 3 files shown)


WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -199,9 +199,9 @@
 		17BB26AE1E6D8321008CD031 /* MediaLibraryViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17BB26AD1E6D8321008CD031 /* MediaLibraryViewController.swift */; };
 		17BD4A0820F76A4700975AC3 /* Routes+Banners.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17BD4A0720F76A4700975AC3 /* Routes+Banners.swift */; };
 		17BD4A192101D31B00975AC3 /* NavigationActionHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17BD4A182101D31B00975AC3 /* NavigationActionHelpers.swift */; };
+		17C2FF0925D4852400CDB712 /* UnifiedProloguePages.swift in Sources */ = {isa = PBXBuildFile; fileRef = 17C2FF0825D4852400CDB712 /* UnifiedProloguePages.swift */; };
```
jkmassel, line 202: Xcode keeps trying to make me do this every time I switch branches...I guess we can let it? Or I could just discard the change 🤷‍♂️

ScoutHarris (approved):
✅ login and logout
✅ unit tests

:shipit: !

WordPress/WordPress.xcodeproj/project.pbxproj
ScoutHarris: Ha! I think it's fine. It happens occasionally. It looks like the file was just moved/reordered, so it should be ok.

PR #15992, opened 2021-02-27
Switch Gutenberg pod reference to use https

frosty:
I was having difficulty installing the Gutenberg pod on my M1 mac – CocoaPods was saying it couldn't find the right version of the dependency:

```
Analyzing dependencies
Pre-downloading: `Kanvas` from `git@github.com:tumblr/kanvas-ios.git`, commit `62e5220e8e19456181f837ba494c5ee048f45926`
[!] CocoaPods could not find compatible versions for pod "Gutenberg":
  In snapshot (Podfile.lock):
    Gutenberg (from `http://github.com/wordpress-mobile/gutenberg-mobile/`, tag `v1.47.0`)

  In Podfile:
    Gutenberg (from `http://github.com/wordpress-mobile/gutenberg-mobile/`, tag `v1.47.0`)

None of your spec sources contain a spec satisfying the dependencies: `Gutenberg (from `http://github.com/wordpress-mobile/gutenberg-mobile/`, tag `v1.47.0`), Gutenberg (from `http://github.com/wordpress-mobile/gutenberg-mobile/`, tag `v1.47.0`)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.
rake aborted!
```

The only difference I could see between Mobile Gutenberg and other pods was that it was using an `http` reference instead of `https`. I updated it, and then (I'm not entirely sure why!) the pod installed just fine.

**To test**

- On both an Intel and an M1 Mac, install dependencies with `rake dependencies`
- Build and run the app (afaik on M1 the app will only build for a device, not simulator)

PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -82,7 +82,7 @@ def shared_with_extension_pods
 end
 
 def gutenberg(options)
-    options[:git] = 'http://github.com/wordpress-mobile/gutenberg-mobile/'
+    options[:git] = 'https://github.com/wordpress-mobile/gutenberg-mobile.git'
     options[:submodules] = true
     local_gutenberg = ENV['LOCAL_GUTENBERG']
     if local_gutenberg
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -454,7 +454,7 @@ DEPENDENCIES:
   - Gifu (= 3.2.0)
   - glog (from `https://raw.githubusercontent.com/wordpress-mobile/gutenberg-mobile/v1.48.0-alpha1/third-party-podspecs/glog.podspec.json`)
   - Gridicons (~> 1.1.0)
-  - Gutenberg (from `http://github.com/wordpress-mobile/gutenberg-mobile/`, tag `v1.48.0-alpha1`)
+  - Gutenberg (from `https://github.com/wordpress-mobile/gutenberg-mobile.git`, tag `v1.48.0-alpha1`)
   - JTAppleCalendar (~> 8.0.2)
   - Kanvas (~> 1.2.2)
   - MediaEditor (~> 1.2.1)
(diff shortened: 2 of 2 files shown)


frosty:
@mokagio I tagged you for this because I thought you'd been building on M1, but let me know if not and I can find someone else to review! 🙂

mokagio (approved):
@frosty you are correct, I own have an M1 Mac Mini test device.

The error you reported is one that we've seen a few times already, usually when we jump on a repo using CocoaPods after not working on it for a long time. In fact, that's exactly what happened to me while checking out your PR on my Intel MacBook Pro. Usually, the fix for that is to call `pod install` with `--repo-update`. I took the liberty of adding that change to this PR as well.

I think using HTTPS for Gutenberg is appropriate 👍 Maybe, it fixed the issue for you because changing the scheme made CocoaPods fetch it from scratch? _But then, why did it fail on my end?_ 🤷‍♂️ 

I wasn't able to verify the fix on my M1, though 😳 The command crashes the Ruby interpreter on my end. I assume that's an issue with my setup, which I'll keep investigating.

In the meantime, this PR does no harm so I'd say _ship it_.

frosty:
Thank for the review @mokagio! I should've added in my PR description that I _did_ run `--repo-update` (many times! 🙂) and it would not get rid of the issue for me. Updating the protocol was the only thing that seemed to kick it into action.

Updating the `install` action feels like a bigger change than I had here initially. I think my preference would be to just tweak the Gutenberg line and we can discuss changing the way `rake dependencies` works separately. What do you think?

mokagio:
> I think my preference would be to just tweak the Gutenberg line and we can discuss changing the way `rake dependencies` works separately. What do you think?

Sounds good to me. If that's okay with you, I'm happy to take this from here, interactive-rebase-dropping my commit, to avoid any extra work on your end 😅 

Even more so because I went back and looked at an earlier conversation we –the platform team– had about it; turns out there's a better way of doing this than I suggested here, ref paaHJt-13M-p2#comment-2321. 🤦‍♂️ 👍

frosty:
Sure, I'm happy for you to go ahead and wrap this up – thank you!

PR #15996, opened 2021-03-02
Stories: 16.8 Bug Fixes

bjtitus:
Related Kanvas PR: https://github.com/tumblr/kanvas-ios/pull/93

- Fixes a bug where the Media Picker could be rotated to landscape and mess up the UI. Notice the size of the view controller + tab bar and the "Add" button in the "before" video below:

https://user-images.githubusercontent.com/3250/109680694-fe049e80-7b39-11eb-91be-a903ef451ce7.mp4

- Fixes a bug when scaling Portrait video to fill the device screen. Note that this does slow down video import a little.

| Before | After |
| ------ | ----- |
| <a href="https://user-images.githubusercontent.com/3250/109605557-94a57100-7ae2-11eb-92e6-2aaa024100d0.PNG"></a> | <a href="https://user-images.githubusercontent.com/3250/109605589-9e2ed900-7ae2-11eb-9372-c07dd938b3b1.PNG"></a> |

- Disables Stories on iPad for the 16.8 release.

### Testing
* Open the Story Editor.
* Open the Media Picker using the button in the bottom left.
* Rotate the device to landscape and ensure that the status bar remains fixed.
* Select a **video** which was taken in **Portrait** by your iPhone.
* Ensure that the video appears in the correct aspect ratio and is not warped.

#### iPad
* Tap on the FAB from the Blog Details or Post list screens
* Ensure that the "Story post" option is not available on iPad
* When attempting to edit a story, ensure that a message appears when tapping the edit button.
* The Story block should not be available in the block picker when adding a new block.

PR submission checklist:


aforcier:
Gave this a try, things seem to work as expected 🎉 Found one possible issue.

* Confirmed that adding a video filmed in portrait mode used to be warped, and now isn't
* Confirmed that nothing breaks when rotating the device (I'm not sure I understood the step - the expected result is that landscape mode is effectively disabled, is that right?)
* On an iPad simulator, ensured that "Story post" was not available as an option in the create sheet from either blog details or post list, and confirmed that tapping a story in an existing post shows a message and prevents the story editor from being loaded.

The issue I found is that the story block still shows up in the Gutenberg block picker list on an iPad. We should probably hide it from there as well. However the message that iPad is not supported shows up correctly when attempting to populate the block, so nothing breaks, and we found this to be rarely used on Android - so if it's for some reason difficult/dangerous to implement it's probably fine to leave it in.

WordPress/Classes/ViewRelated/System/WPTabBarController+ShowTab.swift
```diff
@@ -69,22 +69,26 @@ extension WPTabBarController {
...
-                case .failure(let error):
-                    self?.dismiss(animated: true, completion: nil)
-                    let controller = UIAlertController(title: "Failed to create story", message: "Error: \(error)", preferredStyle: .alert)
-                    let dismiss = UIAlertAction(title: "Dismiss", style: .default) { _ in
-                        controller.dismiss(animated: true, completion: nil)
+            do {
+                let controller = try StoryEditor.editor(blog: blog, context: ContextManager.shared.mainContext, updated: {_ in }, uploaded: { [weak self] result in
+                    switch result {
+                    case .success:
+                        ()
```
aerych: I know this was an existing expression, but what would you think about changing the `()` to a `break` since that's the more expected syntax for a switch/case skip. (Maybe this predates break being added to Swift 🤔)

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -675,7 +675,34 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
         }
 
-        storyEditor = StoryEditor.editor(post: post, mediaFiles: files, publishOnCompletion: false, updated: { [weak self] result in
+        do {
+            try showEditor(files: files)
+        } catch let error {
+            switch error {
+            case StoryEditor.EditorCreationError.unsupportedDevice:
+                let title = NSLocalizedString("Unsupported Device", comment: "Title for stories unsupported device error.")
+                let message = NSLocalizedString("We are still working on the Stories editor for iPad. In the meantime, please try Stories on your iPhone.", comment: "Message for stories unsupported device error.")
```
aerych: I might be a little overly cautious about the wording here, but I'm wondering if it could run afoul of Apple's "no betas" rule if we get a reviewer that's particularly fussy.  Instead of "We are still working on the Stories editor for iPad" would it be safer to say something like "The Stories editor for is not currently available for your iPad"?

WordPress/Classes/Services/Stories/WPMediaPicker+MediaPicker.swift
```diff
@@ -253,17 +267,51 @@ extension WPMediaAsset {
...
             return exportSession.exportPublisher(url: exportURL)
         } else {
             throw WPMediaAssetError.videoAssetExportFailed
         }
     }
+
+    /// Applies the `preferredTransform` of the video track.
+    /// - Returns: Returns both an AVMutableComposition containing video + audio and an AVVideoComposition of the rotate video.
+    private func rotate() -> (AVMutableComposition, AVVideoComposition) {
+        let videoTrack = tracks(withMediaType: .video).first!
```
aerych: Maybe this could be tweaked to avoid the implicitly unwrapped optional?

aerych (approved):
Hola @bjtitus 👋 
Looking good!  Tested on an iPhone device and iPad simulator. I had some very nitpicky comments code wise but nothing that's a blocker (tho getting rid of the optional would be delightful).  :shipit: from me, and if you decide to make any other changes let me know and I'll quickly re-review. :)

PR #15998, opened 2021-03-02
Comments: show Trashed and Spam

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/15955
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/358

When a filter is selected, its `NSPredicate` is sent back to `CommentsViewController`. It is then used when re-fetching Comments. This allows Trashed and Spam comments to be displayed correctly.

Outstanding issues:
- The results are still sorted by status, then date. 
- The selected filter is not yet saved.

To test:

- Go to Site > Comments.
- On the web, go to `Manage Comments` for the same site. (https://wordpress.com/comments/all/yoursite)
- In the app:
  - Verify all the filters show the same Comments as the web.
  - Verify only Pending comments have the yellow dot.
  - Verify the param log messages no longer appear. (`🔴 REST > params`, `🔴 XMLRPC > params`)
- Disable `commentFilters` feature flag.  
  - Verify the filter tab bar is not displayed, and only Pending and Approved comments are shown as before.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
@@ -23,7 +23,6 @@ static NSInteger const CommentsFetchBatchSize                   = 10;
 @property (nonatomic, strong) Blog                      *blog;
 
 @property (nonatomic) CommentStatusFilter currentStatusFilter;
-@property (nonatomic, strong) NSPredicate *statusPredicate;
 @property (weak, nonatomic) IBOutlet FilterTabBar *filterTabBar;
 @property (weak, nonatomic) IBOutlet UITableView *tableView;
 @property (weak, nonatomic) IBOutlet NSLayoutConstraint *tableViewTopConstraint;
@@ -350,7 +349,8 @@ static NSInteger const CommentsFetchBatchSize                   = 10;
     NSFetchRequest *fetchRequest = [NSFetchRequest fetchRequestWithEntityName:[self entityName]];
 
     if ([Feature enabled:FeatureFlagCommentFilters]) {
-        fetchRequest.predicate = self.filteredCommentsPredicate;
+        // CommentService purges Comments that do not belong to the current filter.
+        fetchRequest.predicate = [NSPredicate predicateWithFormat:@"(blog == %@)", self.blog];
     } else {
         fetchRequest.predicate = [NSPredicate predicateWithFormat:@"(blog == %@ AND status != %@)", self.blog, CommentStatusSpam];
     }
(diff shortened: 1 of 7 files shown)


ScoutHarris:
Hey @aerych . There's an issue on iPad where the notifications don't reload.
- Go to Comments. 
- Select a filter other than All.
- Select a different view in the site table (Stats, Activity Log, etc.)
- Return to Comments.
- The filter is reset to All (this is known) but the Comments are not correct (this is broken).

You can pull to refresh, and the Comments are updated correctly.

I believe it is related to 1) the filter not being saved, and 2) the Comments view not updating when shown. I'll look at this separately.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -346,17 +347,33 @@ - (NSManagedObjectContext *)managedObjectContext
...
+    fetchRequest.fetchBatchSize = CommentsFetchBatchSize;
     
     return fetchRequest;
 }
 
+- (NSPredicate *)filteredCommentsPredicate
+{
+    NSMutableArray *subpredicates = [NSMutableArray arrayWithObjects:[NSPredicate predicateWithFormat:@"blog == %@", self.blog], nil];
+
+    if (self.statusPredicate != NULL) {
```
aerych: For objc, if you like, the `!= NULL` could be omitted.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -346,17 +347,33 @@ - (NSManagedObjectContext *)managedObjectContext
...
+
+    NSSortDescriptor *sortDescriptorStatus = [NSSortDescriptor sortDescriptorWithKey:@"status" ascending:NO];
+    NSSortDescriptor *sortDescriptorDate = [NSSortDescriptor sortDescriptorWithKey:@"dateCreated" ascending:NO];
+    fetchRequest.sortDescriptors = @[sortDescriptorStatus, sortDescriptorDate];
+    fetchRequest.fetchBatchSize = CommentsFetchBatchSize;
     
     return fetchRequest;
 }
 
+- (NSPredicate *)filteredCommentsPredicate
```
aerych: It doesn't look like this is called when changing filters. It looks like the intent was to update the fetchRequest's predicate for the selected filter, however the fetch request's predicate isn't being updated and the fetched objects are not being refreshed. 
Try disabling wifi and testing and you'll see the list never updates when there is no internet connection.
The filtering appears to be happening in the service where comments that do not belong to the current filter are purged. (I dig this approach actually.)

aerych (changes requested):
Hola @ScoutHarris 👋 
Code looks good, but I spotted an issue while offline.  Back to you!

ScoutHarris:
Hey @aerych .

- I removed the filter specific predicates as they were unnecessary.
- When offline, all filters will display the comments for the last selected filter.
- The `No internet connection.` toast message will appear when offline and a filter is selected.

Regarding the last two points, as we discussed via Slack, in another PR I'll show the offline NRV instead.

Ready for another review please!

aerych (approved):
LGTM! :shipit:

ScoutHarris:
Thanks @aerych !

PR #15994, opened 2021-03-01
Implement fetch likes in CommentService

bitdc:
Refs #15662

Add implementation to fetch likes for a comment in `CommentService`. There's also a slight refactor that introduces `CommentServiceRemoteFactory` to make `CommentService` injectable in unit tests.

To test:

Confirm the app builds. The rest are covered through unit tests as this is a "backend" change.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16007, opened 2021-03-03
Likes Notification: Display likes using the new endpoints

bitdc:
Refs #15662 

Since the likes data will be fetched from a service, `LikesListController` is introduced to encapsulate the fetching logic. The component mainly serves as a replacement for the table view's delegate and data source in `NotificationDetailsViewController`. Here are a couple of reasons why I chose this approach:

- Minimal impact on the current implementation.
- Minimal effort to implement feature flag (see `setupTableDelegates` method in `NotificationDetailsViewController`).
- Makes it easy to modify later on once we want to implement the Likes page from Reader (e.g. create a thin view controller that uses this component).

By checking out this PR, you should be able to see the look and feel of the new changes directly. However, there are some areas that are not yet covered in this PR:
- [x] Implement feature flag.
- [ ] Add tests for the `LikesListController`. In order to minimize the PR size, I plan to separate the test into another PR.
- [ ] Ensure that the changes do not impact accessibility.
- [ ] Implement Tracks.

Additionally, some review comments from previous PRs were addressed in this change:
- https://github.com/wordpress-mobile/WordPress-iOS/pull/15977#discussion_r585114909
- https://github.com/wordpress-mobile/WordPress-iOS/pull/15994#discussion_r585122810

To test:

- Ensure that the app builds and that you are logged in using an account that already has existing notifications.
- Go to the Notifications tab, and select a notification that's either post likes or comment likes – preferably one that has more than 10 likes. Verify that the list displays all users that liked the content.
- Try to navigate between notifications through the previous/next buttons on the top right navigation bar, and verify that the contents are displayed properly.

PR submission checklist:


WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+
+    private let formatter = FormattableContentFormatter()
+
+    private let dependency: LikesListDependency
+
+    private let content: ContentIdentifier
+
+    let siteID: NSNumber
+
+    let note: Notification?
```
ScoutHarris: I know this is done in other places in the app, but it would be more clear / less confusing if `note` was the full `notification`.

Also, looks like this can be `private`.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+            return
+        }
+
+        // shows the loading cell and prevents double refresh.
+        isLoadingContent = true
+
+        fetchLikes(success: { [weak self] users in
+            if let someUsers = users {
+                self?.likingUsers = someUsers
+            }
```
ScoutHarris: I think this can just be `self?.likingUsers = users ?? []`.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+            self?.isLoadingContent = false
+        }, failure: { [weak self] _ in
+            // TODO: Show error state?
```
ScoutHarris: 👍

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+        let view = UIActivityIndicatorView(style: .medium)
+        view.translatesAutoresizingMaskIntoConstraints = false
+
+        return view
+    }()
+
+    private lazy var loadingCell: UITableViewCell = {
+        let cell = UITableViewCell()
+
+        // add activity indicator
```
ScoutHarris: Nitpick: I don't think this comment is necessary as it's clear what's happening.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+/// This is intended to be used as replacement for table view delegate and data source.
+class LikesListController: NSObject {
+
+    private let formatter = FormattableContentFormatter()
+
+    private let dependency: LikesListDependency
+
+    private let content: ContentIdentifier
+
+    let siteID: NSNumber
```
ScoutHarris: Looks like this can be `private`.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+
+    let note: Notification?
+
+    let tableView: UITableView
```
ScoutHarris: Looks like this can be `private`.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+
+    private var likingUsers: [RemoteUser] = []
+
+    weak var delegate: LikesListControllerDelegate?
```
ScoutHarris: Looks like this can be `private`.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// MARK: - Notification Cell Handling
+
+extension LikesListController {
+
+    func headerCell() -> NoteBlockHeaderTableViewCell {
+        guard let cell = tableView.dequeueReusableCell(withIdentifier: NoteBlockHeaderTableViewCell.reuseIdentifier()) as? NoteBlockHeaderTableViewCell,
+              let group = note?.headerAndBodyContentGroups[Constants.headerRowIndex] else {
+            fatalError()
```
ScoutHarris: We should just `return` here. Maybe add a `DDLogError` before returning.

If you don't want the return value to be optional, you could just return an `NoteBlockHeaderTableViewCell()`.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+
+        // Download the Gravatar
+        let mediaURL = gravatarBlock.media.first?.mediaURL
+        cell.downloadAuthorAvatar(with: mediaURL)
+    }
+
+    func userCell(for indexPath: IndexPath) -> NoteBlockUserTableViewCell {
+        guard indexPath.row < likingUsers.count,
+              let cell = tableView.dequeueReusableCell(withIdentifier: NoteBlockUserTableViewCell.reuseIdentifier()) as? NoteBlockUserTableViewCell else {
+            fatalError()
```
ScoutHarris: We should just `return` here. Maybe add a `DDLogError` before returning.

If you don't want the return value to be optional, you could just return an `NoteBlockUserTableViewCell
()`.

WordPress/Classes/ViewRelated/Likes/LikesListDependency.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+import WordPressKit
+
+
+/// A simple, convenient dependency wrapper for LikesListController.
+class LikesListDependency {
```
ScoutHarris: I'm sorry, I don't quite understand the purpose of this separate class. All its content is only used by `LikesListController`, so why not just define them there?

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        delegate?.didSelectUser(likingUsers[indexPath.row])
+    }
+
+}
+
+// MARK: - Notification Cell Handling
+
+extension LikesListController {
```
ScoutHarris: I think this can be a `private` extension since everything in it should be `private`.

ScoutHarris (changes requested):
Hey David. 

This is looking good. I have a few code comments, suggestions, etc.

Functionally, I did notice one thing. On devices with split view (ex: iPad and iPhone Pro Max), when rotated the table reloads. It should just rotate with the existing content.

WordPress/Classes/ViewRelated/Likes/LikesListDependency.swift
bitdc: Moved the class into `LikesListController`. Thanks for the suggestion!

bitdc:
The latest update should address all of the reviews above, and also:

1. Added a feature flag for the likes notification.
2. Added a simple cell to display when a fetch error occurs.
3. Fixed the reload issue when the device is rotated.

Here's a preview of the error cell:
Screen Shot 2021-03-04 at 15 17 26

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,343 @@
...
+
+        return cell
+    }
+
+}
+
+// MARK: - Dependency Definitions
+
+/// A simple, convenient dependency wrapper for LikesListController.
+class LikesListDependency {
```
ScoutHarris: Thanks for moving this, but I think my question wasn't clear. I'm confused as to why we have this class period. Its content is simply properties used by `LikesListController`. Why not just make them properties of `LikesListController`?

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,343 @@
...
+        NSLayoutConstraint.activate([
+            activityIndicator.safeCenterXAnchor.constraint(equalTo: cell.safeCenterXAnchor),
+            activityIndicator.safeCenterYAnchor.constraint(equalTo: cell.safeCenterYAnchor)
+        ])
+
+        return cell
+    }()
+
+    /**
+     TODO: Update this once we get more updates from the design?
```
ScoutHarris: Yea, we would do something different here. Can you undo the `errorCell` change please? It'll just have to be removed later. If we do it now while it's fresh, there won't be any confusion later. Thanks!

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -0,0 +1,343 @@
...
+              let cell = tableView.dequeueReusableCell(withIdentifier: NoteBlockUserTableViewCell.reuseIdentifier()) as? NoteBlockUserTableViewCell else {
+            DDLogError("Error: couldn't get a user cell or requested row is out of boundary")
+            return NoteBlockUserTableViewCell()
+        }
+
+        let user = likingUsers[indexPath.row]
+        cell.accessoryType = .none
+        cell.name = user.displayName
+
+        // TODO: Re-enable follow functionality once the information is available.
```
ScoutHarris: 🤔 I don't think we are going to re-enable it, correct? Per the design, the follow button on the cell has been removed, and can be accessed via the user's site. So this comment isn't accurate?

ScoutHarris (changes requested):

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
davidchristiandy: Oh alright, noted on this.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
davidchristiandy: Got it. Initially I thought it would be helpful for unit tests, but I guess I can remove this class instead. Thanks for the input!

bitdc: Ah yes, initially I thought it would be helpful for unit testing later on, but I guess I can remove it to simplify the class. Thanks for the input!

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
bitdc: Alright, noted! Sorry for this. I think I misunderstood the comment on previous review. 🙏

ScoutHarris: Oh! Sorry to confuse! Thank you!

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
bitdc: Oh, got it. I assumed that we wanted to match the previous look first, instead of building it towards the new cell designs directly. Will fix this in the next commit!

ScoutHarris: Yea, generally that was the idea. But I think going ahead as-is is worth not having to add endpoint support for something that would be short lived. Thank you!

ScoutHarris (approved):
I think we got it this time! 😄 Thanks David!

PR #16008, opened 2021-03-03
Add classic editor deprecation notice

guarani:
Addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/3209

This PR adds a deprecation notice to the classic editor for all users regardless of site type.

### To test
1. Create a new post using the classic editor. This can be done by changing the site editor preference to classic in My Site > Site Setting > Turn OFF Use Block Editor setting, and then creating a post as you normally would.
2. Expect an alert to be shown notifying users of the future removal of the classic editor for new posts/pages.
3. Smoke test the app to ensure the alert is not shown for existing posts with classic content or any other scenarios.
4. I also tested the iOS share extension to make sure the popup doesn't show there.

### Screenshot

 

PR submission checklist:


cameronvoell (approved):
Changes look good. Notice looks good in dark mode. I did not see it come up after the initial presentation, worked as expected 👍

PR #16019, opened 2021-03-05
Add device mode selection for page layouts

mkevins:
#### Related PR

`WordPressKit-iOS`: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/365

### Description

This PR enables mobile and tablet thumbnail previews for the page layout picker. It also sets the default thumbnail mode to mobile and tablet on mobile and tablet devices, respectively.

To test:

#### Default modes

1. Access the page layout picker (press (+) to create a new page from the page list or my site screen).
2. Observe the default thumbnail mode
    * Mobile: it should display mobile preview thumbnails
    * Tablet: it should display tablet preview thumbnails

##### Screenshots

Device type | Before | After
-|-|-
Phone | page-layout-thumbnail-default-desktop | page-layout-thumbnail-default-mobile
Tablet | page-layout-thumbnail-default-desktop-tablet | page-layout-thumbnail-default-tablet

#### Changing modes

1. Access the page layout picker (press (+) to create a new page from the page list or my site screen).
2. Observe the default thumbnail mode
3. Tap the thumbnail device mode selector
    * The current mode should be indicated via a checkmark
    * 
4. Choose another mode by tapping a selection
5. Observe the thumbnail images: they should change to the selected thumbnail preview type (e.g. desktop, tablet or mobile)

##### Screenshots

Mobile | Tablet
-|-
<video src="https://user-images.githubusercontent.com/8507675/111582256-7ca44180-87b2-11eb-94b0-540123ec553c.mov"> | <video src="https://user-images.githubusercontent.com/8507675/111582290-862da980-87b2-11eb-9850-d328775a9cb8.mov">

PR submission checklist:


antonis:
@mkevins I noticed that the categories ordering (this is not part of this PR) does not align with the Android implementation. On Android I removed the alphabetic ordering reusing the implementation of the [Home Page Picker](https://github.com/wordpress-mobile/WordPress-iOS/pull/15933#issuecomment-790117564). Not sure if my approach is correct since we did not specifically discussed this for this screen.
If we plan to use the server ordering we should align iOS otherwise I'll fix this on the Android side. Wdyt?
cc @chipsnyder, @enejb

antonis (approved):
Great work @mkevins 👍 
Tested this on an iPhone SE 2020 (iOS 14.2) and on an iPad simulator and everything works as expected! The code changes are consistent and LGTM 🎉

chipsnyder:
>If we plan to use the server ordering we should align iOS otherwise I'll fix this on the Android side. Wdyt?

Yeah, we didn't specifically address this screen but I agree using the server ordering seems to be the better approach especially as we consider adding new layouts. 

I didn't test this out but from my scan of the code and memory I believe all that we would need to do in order to achieve this would be to remove `.sorted()` from https://github.com/wordpress-mobile/WordPressKit-iOS/pull/365/files#diff-81546307097bc3c5d77fa313473bb4595a716faaa3564ec433ae70ddd8a467dfR15

antonis:
As part of the [layout picker preview mode refactoring I added a couple of events in this screen on Android](https://github.com/wordpress-mobile/WordPress-Android/pull/14184). Specifically: `layout_picker_thumbnail_mode_button_tapped` and `layout_picker_preview_mode_changed`
I'm adding this as a note and not as a change request specifically for this PR.

mkevins:
Thank you for reviewing and testing!

Regarding the sort order:

> I believe all that we would need to do in order to achieve this would be to remove `.sorted()`

I made this change, as this will be necessary anyway, but it turns out it's not enough because of the caching layer. I think we'll need to explicitly add an ordering parameter to the data model as well, so we can use that as the sort descriptor key in the results controller. I'll defer that to the next PR.

PR #16030, opened 2021-03-05
Avoid tracking widget events in the preview

Gio2018:
Fixes #NA

This PR fixes an issue that was causing widget install events to be sent when the widget were only previewed and not actually installed.

To test:
- Pre-requisites:
    - It's recommended to use a simulator
    - To avoid waiting longer times, in `SiteListProvider` you can tweak:
        - `refreshInterval = 5`
        - `minElapsedTimeToRefresh = 1`
        - In the simulator settings, enable "WidgetKit developer mode"

- Test execution:
    - Build/run the app, then background it
    - in Xcode, Debug/Attach to process and select `WordPressStatsWidgets`
    - Do a variety of widget operations (preview, install, change site, delete, etc)
    - Use breakpoints in `Tracks+StatsWidgets.swift to make sure that:
        - No events are sent when in preview mode
        - events are only sent if a widget is installed, and not when the site in the same widget is changed
        - Wait approx 5 minutes so that iOS triggers a widget update, and make sure no events are sent in that case
    - The method that sends the tracks is `trackExtensionEvent`

PR submission checklist:


frosty (changes requested):
I followed the testing steps, and while I'm not receiving events from previews, I am receiving events when changing sites on an existing widget (I think actually two events get fired in this situation). I just added two widgets of the same type, changed the site on each of them, and received 6 events.

Gio2018:
> I followed the testing steps, and while I'm not receiving events from previews, I am receiving events when changing sites on an existing widget (I think actually two events get fired in this situation). I just added two widgets of the same type, changed the site on each of them, and received 6 events.

Hey @frosty , I tried a variety of combinations installing, removing and changing sites, but did not encounter the above situation. When I change the site, the `trackWidgetUpdated` method is indeed called, but the `trackExtensionEvent` inside that method (which is what actually sends the tracks to the backend) does not get called because the `properties` dictionary does not change. Maybe tomorrow we can have a quick call to see what exactly are your testing condition, so I can better troubleshoot this? Thanks!!

frosty:
Sure, I'll take another test of this and see if it happens again!

frosty (approved):
After our discussion and further investigation, it looks as though the issue I found may well be due to some unexpected WidgetKit behaviour. Let's merge this version as it's still an improvement in filtering out the previews, and we can always iterate on it again if necessary!

PR #16038, opened 2021-03-08
Add APP_STORE_STRINGS_FILE_NAME env constant

jkmassel:
Adds the `APP_STORE_STRINGS_FILE_NAME` expected by the release toolkit.

PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -72,7 +72,7 @@ ENV["PUBLIC_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.Publ
 ENV["INTERNAL_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.internal.xcconfig")
 ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
 ENV["PROJECT_ROOT_FOLDER"]=PROJECT_ROOT_FOLDER + "/"
-
+ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.po"
 
 ########################################################################
 # Screenshots


designsimply (approved):
LGTM :shipit:

fastlane/Fastfile
```diff
@@ -72,7 +72,7 @@ ENV["PUBLIC_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.Publ
 ENV["INTERNAL_CONFIG_FILE"]=File.join(PROJECT_ROOT_FOLDER, "config", "Version.internal.xcconfig")
 ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
 ENV["PROJECT_ROOT_FOLDER"]=PROJECT_ROOT_FOLDER + "/"
-
+ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.po"
```
AliSoftware, line 75: 🤔 I'm surprised that this is injectable, I thought the release toolkit hard-coded that 🤷 

Why did that env var suddenly started to be needed (or, if it was always needed, how did it manage to work until now then?)

mokagio: @jkmassel I'm curious about this, too.

According to GitHub, there's only [one occurrence of `APP_STORE_STRINGS_FILE_NAME` in the release toolkit](https://github.com/wordpress-mobile/release-toolkit/search?q=APP_STORE_STRINGS_FILE_NAME), and it has no fallback if the value is not set.

I had a quick look through the blame history, and I couldn't see a fallback being there in recent the past either.

How has this been working and not crashing when Ruby couldn't find this value 🤔 🤔 🤔

mokagio, line 75: Out of scope.

While looking the code and the history (see [previous comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/16038/files#r589863179)) I noticed in WooCommerce iOS [this file has the `.pot` extension](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Resources/AppStoreStrings.pot).

From what I understand from [Wikipedia's explanation of how `gettext` works for localizations](https://en.wikipedia.org/wiki/Gettext), `.pot` stands for Portable Object Template and "contains a list of all the translatable strings extracted from the sources" while `.po` stands for Portable Object and it's a file that the translator derives "from the template using the `msginit` program, then fills out the translations". `msginit` "initializes the translations".

I'm not 100% sure of what that all means in practical terms for us, but I suspect `.pot` would be the correct extension.

PR #15751, opened 2021-02-01
Merges the latest from the Large Headers feature branch into develop.

diegoreymendez:
Title has it.

What's important in this PR is to test the App with the feature flag `newNavigationBar` OFF, and make sure everything works fine.

Once this is merged, we can incrementally implement and push [what remains to be done](https://github.com/wordpress-mobile/WordPress-iOS/issues/15750) for `newNavigationBar` ON.

## Testing

Make sure the `newNavigationBar` feature flag is OFF by default.

- Make sure the UI in the My Site tab works fine.
- Make sure the Me button works fine.
- Make sure logging in and out between account with multiple sites, one site, and no sites works fine.

## PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16039, opened 2021-03-08
Large Titles / White Headers: Move My Sites VCs to My Sites Coordinator

frosty:
This PR is the first part of pulling the changes from #15751 into develop. We uncovered some issues in that branch, so I'm moving things over piece by piece so we can test each in isolation. This PR implements the changes moving the main My Sites view controllers over to My Sites Coordinator. This will allow us to selectively switch between the new and old style blog pickers in an upcoming PR, from a purely Swift-based class.

**To test**

- Test on both iPad and iPhone.
- Build and run.
- Ensure that you can log in and out and back into the app and the blog details screens show up as expected.
- Try moving the iPad into multitasking to collapse the split view and back out and ensure that a detail view controller is restored in My Site.
- After logging out and back in, use the Memory Graph debugging view to ensure that there is only a single instance of WPTabBarController, 2 WPSplitViewControllers, and 1 BlogListViewController and BlogDetailViewController.
- Try performing an action that automatically navigates into a blog detail section. For example, from the home screen, tap and hold on the WordPress icon and choose Stats, or run a deep link using a command such as: `xcrun simctl openurl booted https://wordpress.com/comments/your_site_domain_here`

PR submission checklist:


WordPress/Classes/ViewRelated/System/WPTabBarController.m
```diff
@@ -154,36 +151,27 @@ - (void)decodeRestorableStateWithCoder:(NSCoder *)coder
...
 
-#pragma mark - Tab Bar Items
+#pragma mark - My Site View Controllers
 
-- (UINavigationController *)blogListNavigationController
-{
-    if (_blogListNavigationController) {
-        return _blogListNavigationController;
-    }
+// These getters are temporary while we extract items from here to the MySiteCoordinator.
```
frosty, line 156: Just a note that I'll be removing these in a follow-up PR, which will move the navigation methods that rely on these properties. I figured this PR was big enough already :)

Gio2018: 👍

Gio2018 (approved):
:shipit: !

PR #16050, opened 2021-03-09
Initial use of MySiteViewController

frosty:
This PR introduces `MySiteViewController` as an alternative to `BlogListViewController` / `BlogDetailViewController` when using the `newNavBarAppearance` feature flag. A few things to note:

- This PR is mainly concerned with testing the My Site section of the app with the feature flag _disabled_, which is now the default.
- As such, there are some known issues when the feature flag is enabled, as the new site picker isn't yet in place. This means that with the feature flag enabled, you are currently unable to switch sites. This will be coming in the next PR.
- I ran into an issue with state restoration once during my testing, where the feature-flagged MySitesViewController was incorrectly restoring a BlogListViewController into the navigation stack before it. If this proves to be a problem for you too, we can address it in a future PR.
- Also note that while `MySiteViewController` contains a lot of code, it is essentially identical to @diegoreymendez's earlier code, which has already been reviewed.

|    |   |
|---|---|
|  IMG_0582 |  IMG_0579 |
| IMG_BE1374C89D0B-1 | IMG_0583 |

**To test:**

- Build and run, and do a smoke test to ensure that the My Site section looks and works as expected on iPhone and iPad.
- Main things to test include logging in and out, logging into an account with no sites and then adding one, and hiding all sites in an account to ensure the no results view is shown.
- Ensure that there is a detail view present when logging into to an account with sites, and that you see a blank (but not grey) detail pane for an account with no sites.
- Also on iPad you can test entering and exiting multitasking mode.
- You can also do a cursory test with the feature flag enabled. Main functionality I've mentioned above should all work, with the exception of switching sites.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
@@ -2,6 +2,8 @@ import WordPressAuthenticator
 
 class MySiteViewController: UIViewController, NoResultsViewHost {
 
+    private let meScenePresenter: ScenePresenter
+
     // MARK: - Initializers
 
     init(meScenePresenter: ScenePresenter) {
@@ -316,10 +318,6 @@ class MySiteViewController: UIViewController, NoResultsViewHost {
 
         self.blog = blog
     }
-
-    // MARK: - Me Bar Button
-
-    private let meScenePresenter: ScenePresenter
 }
 
 extension MySiteViewController: WPSplitViewControllerDetailProvider {
--- a/WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
@@ -175,7 +175,7 @@ class MySiteViewController: UIViewController, NoResultsViewHost {
 
         present(addSiteAlert, animated: true)
     }
-    
+
     private func launchSiteCreation() {
         let wizardLauncher = SiteCreationWizardLauncher()
         guard let wizard = wizardLauncher.ui else {
@@ -213,7 +213,7 @@ class MySiteViewController: UIViewController, NoResultsViewHost {
 
         blogDetailsViewController.view.translatesAutoresizingMaskIntoConstraints = false
         view.pinSubviewToAllEdges(blogDetailsViewController.view)
-        
+
         showInitialDetails()
     }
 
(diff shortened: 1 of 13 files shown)


WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
```diff
@@ -0,0 +1,335 @@
...
+    ///     `AccountService.defaultWordPressComAccount()`
+    ///     > `AccountService.accountWithUUID()`
+    ///     > `NSManagedObjectContext.executeFetchRequest(...)`.
+    ///
+    /// The issue is that `executeFetchRequest` updates the managed object context, thus triggering
+    /// a `NSManagedObjectContextObjectsDidChange` notification, which caused a neverending
+    /// loop in the observer in this VC.
+    ///
+    private func verifyThatBlogsWereInserted(in notification: NSNotification) -> Bool {
+        guard let insertedObjects = notification.userInfo?[NSInsertedObjectsKey] as? Set<NSManagedObject>,
```
Gio2018, line 295: We could probably simplify like so:

```suggestion
        private func verifyThatBlogsWereInserted(in notification: NSNotification) -> Bool {
        (notification.userInfo?[NSInsertedObjectsKey] as? Set<NSManagedObject>)?.contains(where: { $0 as? Blog != nil }) == true
    }
```

WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
```diff
@@ -0,0 +1,335 @@
...
+        guard let blog = blogService.lastUsedOrFirstBlog() else {
+            return
+        }
+
+        self.blog = blog
+    }
+
+    // MARK: - Me Bar Button
+
+    private let meScenePresenter: ScenePresenter
```
Gio2018: nit: since this is a stored property initialized in the init, can we move this at the top for more legibility?

Gio2018 (approved):
Just a couple of minor comments, other than that looks good!

WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
frosty: Thanks for the suggestion – I think I might leave it as-is, as I think it's perhaps a little clearer to read having it split out over a couple of lines 🙂

WordPress/Classes/ViewRelated/Blog/My Site/MySiteViewController.swift
frosty: Definitely!

Issue #15714, opened 2021-01-26
Status Bar while searching Themes


startuptester:
### Expected behavior
The status bar should not turn white with a white search bar

### Actual behavior
https://cloudup.com/c5IWwVEE5Ah

### Steps to reproduce the behavior
1. Tap Themes on the My Site screen
2. Enter the search bar

yaelirub:
Hey @Gio2018 , will this be fixed as part of the white headers project? can you please make sure this is fixed?

Gio2018:
> Hey @Gio2018 , will this be fixed as part of the white headers project? can you please make sure this is fixed?

Sure!!

Gio2018:
Hey @mattmiklic ,pinging you since I understood this came up in the focused site creation, and just to make sure I understand correctly: what would be the exact expected behavior here? 

When we will activate the white headers, if we don't change anything else, the behavior will reflect what is already happening in dark mode, which is: 

the nav bar would still shift up, but be the same color of the background. Is this acceptable or do we want a different behavior for that nav bar?

Thank you!!

mattmiklic:
> the nav bar would still shift up, but be the same color of the background. Is this acceptable or do we want a different behavior for that nav bar?

@Gio2018 I think that's just right. It already looks fine in Dark Mode (other than the issue where the search box is no longer vertically centered on its background when it's active). The issue in light mode is that the header background is switching from blue to white but the status bar text is white on white.

Gio2018:
Thanks for clarifying @mattmiklic !

frosty:
This issue will be fixed by #15714, once large titles / white headers is shipped.

PR #16058, opened 2021-03-10
My Site: New Site Picker

frosty:
This is the final PR pulling the main pieces of https://github.com/wordpress-mobile/WordPress-iOS/pull/15751 into `develop`. It introduces the NewBlogHeaderView, as a new way of picking sites. This makes the My Site (Blog Detail) view controller the top level VC in the My Site navigation stack. The blog list is currently accessed by tapping the down chevron next to the site title.

|    |   |
|---|---|
| IMG_0054 | IMG_0592 |

This should be the final PR that is a little more involved in terms of testing. From here on out all of the major changes to the view hierarchy should be complete, so it's mostly style tweaks.

**Known issues**

- On iPad, when switching site, the correct row may not always be selected in the blog detail list for the currently displayed screen on the right. This is due to the adding of the Jetpack section to the blog details screen – I'll look at addressing it separately.
- There will almost certainly be style things that aren't quite right yet – design related PRs will be coming separately.
- The blog list picker currently doesn't have an explicit Close button – again, that'll come later. For now, you can drag down on the title to dismiss.

**To test**

- Test on iPhone and iPad
- First, test with the feature flag to `false`, that everything looks unchanged in terms of the My Site view hierarchy, switching sites, etc.
- With the feature flag on, ensure that when you log in the left and right panes of the split view are populated on iPad. Check that you can switch between different sections of the My Site screen.
- Try changing sites and check that the blog detail list and the right pane update to reflect that.
- On iPhone, try navigating in and out of sections and check that the large header collapses correctly. I wouldn't treat issues here as a blocker, as that work isn't yet final, but it would be good to have a heads-up if there's anything that needs addressing separately.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
@@ -1099,6 +1099,30 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
     [self showSiteTitleSettings];
 }
 
+- (void)siteSwitcherTapped
+{
+    BlogListViewController* blogListViewController = [[BlogListViewController alloc] initWithMeScenePresenter:self.meScenePresenter];
+    __weak __typeof(self) weakSelf = self;
+    blogListViewController.blogSelected = ^(BlogListViewController* blogListViewController, Blog *blog) {
+        [weakSelf switchToBlog:blog];
+        [blogListViewController dismissViewControllerAnimated:true completion:nil];
+    };
+    
+    UINavigationController* navigationController = [[UINavigationController alloc] initWithRootViewController:blogListViewController];
+    navigationController.modalPresentationStyle = UIModalPresentationFormSheet;
+    [self presentViewController:navigationController animated:true completion:nil];
+}
+
+#pragma mark Site Switching
+
+- (void)switchToBlog:(Blog*)blog
+{
+    self.headerView.blog = blog;
+    self.blog = blog;
+    
+    [self.tableView reloadData];
+}
+
 #pragma mark Site Icon Update Management
 
 - (void)showUpdateSiteIconAlert
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
@@ -1120,7 +1120,17 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
     self.headerView.blog = blog;
     self.blog = blog;
     
+    [self showInitialDetailsForBlog];
     [self.tableView reloadData];
+}
+
+- (void)showInitialDetailsForBlog
+{
+    if ([self splitViewControllerIsHorizontallyCompact]) {
+        return;
+    }
+    
+    [self showDetailViewForSubsection:BlogDetailsSubsectionStats];
 }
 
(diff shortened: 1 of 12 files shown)


WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -1552,6 +1586,7 @@ - (void)showCommentsFromSource:(BlogDetailsNavigationSource)source
 {
     [self trackEvent:WPAnalyticsStatOpenedComments fromSource:source];
     CommentsViewController *controller = [CommentsViewController controllerWithBlog:self.blog];
+    controller.navigationItem.largeTitleDisplayMode = UINavigationItemLargeTitleDisplayModeNever;
```
Gio2018, line 1589: I wonder if we could set `prefersLargeTitles = false` on the navigation bar instead of explicitly set `UINavigationItemLargeTitleDisplayModeNever` for each child vc?

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -467,6 +476,11 @@ - (BOOL)shouldBypassBlogListViewControllerWhenSelectedFromTabBar
 
 - (void)bypassBlogListViewController
 {
+    if ([Feature enabled:FeatureFlagNewNavBarAppearance]) {
```
Gio2018: nit: maybe we can move this condition into `shouldBypassBlogListViewControllerWhenSelectedFromTabBar`?

Gio2018 (approved):
Tested on devices: iPad Pro 2nd gen iPadOS 14.4, iPhone 11 Pro Mac iOs 14.4 and the tests worked as expected (and also observed the known issues).

Just a couple of minor comments in the code.

Also, this PR fixes #15714

:shipit: !

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
frosty: I don't think this works, as setting `prefersLargeDisplayTitles` to `false` will make the navigation bar never display large titles even at the root level.

One other way we could approach this, to avoid having to set it on each child, would be to do something like a custom navigation controller:

```
class LargeTitleNavigationController: UINavigationController {
    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
        viewController.navigationItem.largeTitleDisplayMode = .never
        super.pushViewController(viewController, animated: animated)
    }
}
```

What do you think?

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
frosty: 👍

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
Gio2018: That sounds good to me. Another similar approach, if we don't want to subclass, would be to use a factory method that assigns `largeTitleDisplayMode = .never` to the newly created instance.

frosty: I'm having second thoughts about the navigation controller subclass, as I don't think we should rely on knowing where the navigation controller has come from – for example, the split view controller sometimes creates navigation controllers when the view is collapsed and expanded. I'm going to leave the code as-is for now, and we can always re-visit it later.

PR #16102, opened 2021-03-16
[Tooling] Podfile: Add internal spec repo as source

AliSoftware:
Same as https://github.com/woocommerce/woocommerce-ios/pull/3833

As part of HACK week, I started to mirror our most used internal pods (`WordPressKit`, `WordPressUI`, `WordPressShared`, `WordPressAuthenticator`) into an internal spec repo. (See paaHJt-1SO-p2)

This PR adds the new `source` to the `Podfile`, so that `pod install` searches our internal a8c pods in our internal spec repo mirror in priority before searching them in the CDN. This will hopefully allow us to avoid having to wait for the CDN to sync with CocoaPods trunk after we push new versions of those pods.


Podfile
```diff
@@ -1,3 +1,6 @@
+# For security reasons, please always keep the wordpress-mobile source first and the CDN second.
```
mokagio, line 1: What do you think of adding a link to the rationale?

```suggestion
# For security reasons, please always keep the wordpress-mobile source first and the CDN second.
# See 
# https://github.com/wordpress-mobile/cocoapods-specs#source-order-and-security-considerations
# for more info.
```

mokagio (approved):
CI already shows this, but checked out on my local and it installed everything fine there, too.

👍

Podfile
AliSoftware: 🤔 ~What URL did you have in mind? Not sure there's an official article about this that would be specific to CocoaPods. Or did you mean my P2 post?~ nvm failed to see your diff suggestion at first.

Yeah, good idea! Fixed via https://github.com/wordpress-mobile/WordPress-iOS/pull/16102/commits/a88f968d4623d0802c99e1a3063ab2029b7127ac

PR #16095, opened 2021-03-15
Jetpack app: new target

momo-ozawa:
## Description

- Added a new **Jetpack target**
    - Duplicated WordPress target
    - Duplicated Info.plist, Info-Alpha.plist, Info-Release.plist 
- Added a new **Jetpack scheme**
    - Duplicated WordPress scheme
- Added **pre-action scripts** to WordPress and Jetpack schemes in order to handle App Extensions (ref: https://stackoverflow.com/a/51073669)
    - An app extension needs to use the same bundle ID as the parent target. Otherwise, you'll get an error: `Embedded binary's bundle identifier is not prefixed with the parent app's bundle identifier`. The script changes the bundle ID for the app extensions to match whichever target you're building.
    - ⚠️ The script changes the bundle ID for the app extensions each time we switch schemes, which is not so great
- Updated **Podfile** to define pods for new Jetpack target 
- Added an **asset folder** to the new Jetpack target
    - Changed the AppIcon to the Jetpack Icon
- Added a **Launch Screen** to the new Jetpack target 

### Jetpack target
1. Switch scheme to Jetpack, run `bundle exec pod install`, then run about build the Jetpack app
4. ✅ A Jetpack app with a different icon / splash screen should be running on your device
5. Go to My Site > Switch Site 
6. ✅ A "Welcome to the Jetpack app!" notice should be displayed

### Regression testing (WordPress target)
1. Switch scheme to WordPress, run `bundle exec pod install`, then run about build the WordPress app
4. ✅ A WordPress app with the usual icon / splash screen should be running on your device
5. Go to My Site > Switch Site 
6. ✅ A "Welcome to the Jetpack app!" notice should NOT be displayed

App Icon | Launch Screen | Site List
--- | --- | ---
Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 12 06 08 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 12 06 20 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 12 06 23

### Build Phase Run Script vs Pre-Action Script

- I tried out the [Build Phase Run Script approach](https://stackoverflow.com/a/46722213) that @leandroalonso suggested in  b9c5d93... but like it mentions in the answer, you need to build twice in order for it to build correctly
- Also, this approach has the same problem as the Pre-Action Script approach, where the script changes the bundle ID for the app extension each time we switch schemes

## Pros / Cons

Pros | Cons
--- | ---
[...]

PR #16104, opened 2021-03-16
Jetpack app: add it as a new project

leandroalonso:
This PR is part of an exploration to try different ways of building the Jetpack app.

This differs from https://github.com/wordpress-mobile/WordPress-iOS/pull/16095 because it adds a new project (`pbxproj`), instead of keeping the new app as just a new target/scheme.

This adds more complexity, configuration, and files. But the pros and cons are going to be discussed in the end.

### To test

1. `pod install` (there are a few warnings due to `Generated duplicate UUIDs` that I haven't investigated why it's happening)
2. Build the WordPress scheme, it should work as usual
3. Build the Jetpack scheme, it should build and it should have a 1) different icon 2) different splash screen
4. Also, notice that we have two projects now: Jetpack and WordPress

### Testing different features

1. In the WordPress app, open the site list: nothing should happen
2. In the Jetpack app, open the site list: you should see a message saying "Welcome to the Jetpack app!" at the bottom

The different behavior was done using a preprocessor macro:

```swift
    #if JETPACK
    [self displayNoticeWithTitle:@"Welcome to the Jetpack app!" message:nil];
    #endif
```

### Pros

* Having Jetpack as a separate project helps to keep `WordPress.pbxproj` intact and avoiding possible merge conflicts
* There is no issue with bundle identifiers being changed on build time and having them polluting your git changes (like it happens with the solution in #16095)
* Files added in the Jetpack project will only live there and they will not affect WordPress

### Cons

* If we want to change a screen that already exists in WordPress, we need to add preprocessor macros (`#if JETPACK`) in the original code. Either that or duplicating the file (which might become very complex and costly very quickly)
* If new files get added to the WordPress project, they won't be automatically added to the Jetpack project. We'll have to do it manually or (preferably, and if possible) write a custom script to take care of that for us


momo-ozawa (approved):
Works as described! 👍

PR #16105, opened 2021-03-16
Comments: show No Connection view

ScoutHarris:
Ref: #15955 

This uses the `NoResultsViewController` to display the `Unable to load this content right now` view when there is no connection. It is displayed for filters that do not have cached results. Comments are still displayed for the last fetched filter.

To test:

---
- Go to Site > Comments.
- Disable connection.
- Select a different Comment filter.
- Verify the no connection view is displayed.
- Verify the Comments are still displayed on the original filter.

no_connection

---
- Enable connection.
- Tap `Retry`.
- Verify Comments for the selected filter are displayed.

---
- Select a filter with no Comments.
- Disable connection.
- Select a different Comment filter.
- Verify the no connection view is displayed.
- Verify the no results view is still displayed on the original filter.

| no_results | no_connection |
|--------|-------|

---
PR submission checklist:


WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -406,34 +395,34 @@ - (void)syncHelper:(WPContentSyncHelper *)syncHelper syncContentWithUserInteract
         [commentService syncCommentsForBlog:blogInContext
                                  withStatus:self.currentStatusFilter
                                     success:^(BOOL hasMore) {
-                                        if (success) {
-                                            dispatch_async(dispatch_get_main_queue(), ^{
-                                                success(hasMore);
-                                            });
-                                        }
-                                    }
+            if (success) {
```
aerych, line 397: The way Xcode aligns this. Ugh.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -406,34 +395,34 @@ - (void)syncHelper:(WPContentSyncHelper *)syncHelper syncContentWithUserInteract
...
 }
 
 - (void)syncHelper:(WPContentSyncHelper *)syncHelper syncMoreWithSuccess:(void (^)(BOOL))success failure:(void (^)(NSError *))failure
 {
+    __typeof(self) __weak weakSelf = self;
     NSManagedObjectContext *context = [[ContextManager sharedInstance] newDerivedContext];
     CommentService *commentService  = [[CommentService alloc] initWithManagedObjectContext:context];
+    
     [context performBlock:^{
         Blog *blogInContext = (Blog *)[context existingObjectWithID:self.blog.objectID error:nil];
```
aerych: Not part of these changes, but if we wanted to address this line, a best practice would be to follow the pattern shown elsewhere and `NSManagedObjectID *blogObjectID = self.blog.objectID;` outside of the perform block, then use `blogObjectID` when getting the `blogInContext` in order to avoid referencing the main concurrency `self.blog` on a background thread.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -514,40 +513,107 @@ - (void)refreshPullToRefresh
     }
 }
 
+#pragma mark - No Results Views
+
+- (void)initNoResultsViews
```
aerych: I wonder if there's any chance of naming confusion with this method. Even tho it's return type is `void` and not `instanceType` could it be mistaken for an initializer?

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -514,40 +513,107 @@ - (void)refreshPullToRefresh
...
+    self.noResultsViewController = [NoResultsViewController controller];
+    self.noConnectionViewController = [NoResultsViewController controller];
+}
+
 - (void)refreshNoResultsView
 {
-    [self.noResultsViewController removeFromView];
-    
     if (![self contentIsEmpty]) {
+        [self.tableView setHidden:FALSE];
```
aerych: Seems like our convention is to use `YES` and `NO` vs `TRUE` and `FALSE`.  An ack shows very few uses of `FALSE` in our code (more in third party pods) but lots and lots of uses of `NO`. 
I guess strictly speaking YES and NO are BOOL types and TRUE and FALSE are Boolean types.  NSHipster has a page about it. https://nshipster.com/bool/  🤷

aerych (changes requested):
Hola @ScoutHarris 👋 
I took the branch for a spin.  I had some strange behavior while testing that I've captured video of and will share directly.  Lemme know what you think. 
Code wise I had a few questions.
Back to you!

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
ScoutHarris: > could it be mistaken for an initializer?

Sorry, I'm not seeing what could be confusing. I don't get what one would think this was an initializer of. I can rename it to whatever though. I changed it from `configureNoResultsView` because it's not actually configuring, and _that_ confused _me_.  😄  Would `initializeNoResultsViews` be better? Or is `initialize` the part you take issue with?

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
ScoutHarris: > Seems like our convention is to use YES and NO vs TRUE and FALSE.

Ugh, I always get these backward. Sorry.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
aerych: I was just thinking about ObjC's pattern of `initWithFoo:`, `initWithBar:` for initializers and wondered if the similarity would give someone pause at a glance.  I might just be overthinking it but wanted to ask :)

ScoutHarris:
Hey @aerych .

The weird behavior you reported is a simulator issue. The Comments VC doesn't always get the correct Reachability state, but the NoResults VC does. I noticed the same behavior in other views (ex: Stats fails loading after the connection has been reestablished).

For Comments, this results in:
- The incorrect messaging and image being displayed on the No Connection view (that is, "wp-illustration-empty-results" instead of the no connection "cloud" image).
- Spamming API queries because Comments VC thinks there is a connection when there isn't.

Everything works properly on a real device. I'm not sure there's anything I can do here.

aerych (approved):
Re-tested (on a device this time) and everything works as expected. :) (Yay for simulator bugs?) Out of curiosity I retested on the simulator and the strange (looped?) api requests were no longer happening tho I could reproduce some of the other oddities.  Totes agree it's the simulator at fault and making a note for the future that Simulator + Reachability = 🙃.
:shipit: with great 🚢ing.

ScoutHarris:
Thanks @aerych ! Sorry for all the trouble!

PR #16116, opened 2021-03-17
Fix navigation bar styles for WebKitViewController

### LINKED ISSUE #16078: View Site: the site title and close/reload icons aren’t visible in light mode
When I tap View Site in the My Site menu, the site title and close/reload icons aren’t visible in light mode. This was not present in 16.8.
This is visible in other webviews; i.e. Pages > ... > View and Blog Posts > View

**This is a regression in 16.9**

image
image

iPhone 12 Pro Max, 14.4, 16.9

frosty:
Fixes #16078. WebKitViewController was overriding some styles defined for light navigation bars in `WPStyleGuide+ApplicationStyles`. Ideally, all places in the app should use WebKitViewControllers with light navigation controllers, however that doesn't appear to be the case. As such, I've introduced some conditional code so that this works for both types of navigation bar, which seemed like the simplest fix for this issue for now. I did however update uses in the Reader stream and detail views so that they now use LightNavigationBar.

A good next step would be to audit each use of WebKitViewController (and its subclasses) 

|    |    |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 16 01 10 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 19 50 00 |

**To test**

- Build and run.
- In My Site, scroll to the bottom and choose View Site.
- Check that the title and buttons are visible in the navigation bar.
- Go to Reader.
- Tap the ... on a card in the stream and choose Visit. Check the navigation bar in the web view.
- Dismiss that, and tap into a post in Reader. Tap the globe / Safari icon in the navigation bar, and check the resulting web view.
- If you'd like to test the dark style navigation bars, you can force one to appear by adding the following line of code to `viewDidAppear` in `NotificationDetailViewController`:

`coordinator.displayWebViewWithURL(URL(string: "https://www.google.com")!)`

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/WebKitViewController.swift
+++ b/WordPress/Classes/Utility/WebKitViewController.swift
@@ -88,6 +88,19 @@ class WebKitViewController: UIViewController, WebKitAuthenticatable {
     private var widthConstraint: NSLayoutConstraint?
     private var onClose: (() -> Void)?
 
+    private var useLightStyle: Bool {
+        navigationController is LightNavigationController
+    }
+
+    private var barButtonTintColor: UIColor {
+        useLightStyle ? .listIcon : UIColor(light: .white, dark: .neutral(.shade70))
+    }
+
+    private var navBarTitleColor: UIColor {
+        useLightStyle ? .text : UIColor(light: .white, dark: .neutral(.shade70))
+    }
+
+
     private struct WebViewErrors {
         static let frameLoadInterrupted = 102
     }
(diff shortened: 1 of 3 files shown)


chipsnyder (approved):
> A good next step would be to audit each use of WebKitViewController (and its subclasses)

💯 

Thanks again for taking care of this. Tested it on my iPhone 11 and everything works as expected. :shipit:

Issue #15750, opened 2021-02-01
Large Titles: Main Issue


diegoreymendez:
Implement the new UI for Large Titles, White Headers.

## Feature Availability

The feature branch for this project is `feature/large-headers`.
This feature is also hidden behind feature flag `newNavBarAppearance`.

## New Blog Details Header View

`NewBlogDetailHeaderView` is a new view that replaces `BlogDetailHeaderView`.  It has some UI changed and includes a new site picker, since the list of sites is no longer shown as part of the split view.

Reference design (courtesy of @osullivanchris):

![image](https://user-images.githubusercontent.com/1836005/106460912-cf2fd580-6472-11eb-89dc-b25656c3ce19.png)

### UI / UX

- [x] Change App Heading fonts to New York Large semibold (see image above).
- [x] Make sure Large Titles are set to 34pt (see image above).
- [x] The link (and the icon next to the link) in the new blog details view (right below the site title) should open the site when tapped (see image above).
- [x] [Add a divider to the nav bar when scrolling](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281#issuecomment-768223247).
- [x] [Add padding in between the bottom of the new blog details view and the list below it](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281#issuecomment-768223247).
- [x] Once done with the changes above, iterate with @osullivanchris on the design.

### Known Issues:

- [x] `NewBlogDetailHeaderView` [should not ignore the safe area](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281#pullrequestreview-577465352).
- [x] [Split view issues on iPad](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281#pullrequestreview-577465352).
- [x] [Disable the split view detail view dimming when the feature flag (`newNavBarAppearance`) is ON](https://github.com/wordpress-mobile/WordPress-iOS/pull/15281#pullrequestreview-577465352).
   - This shouldn't happen now that the primary split view pane contains just a BlogDetailViewController.

### Color updates
- [Remove usages or WordPress Blue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15852)
- [Replace usages of Pink with equivalent shades of Blue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15853)
- [Replace app icon WordPress Blue with new Blue](https://github.com/wordpress-mobile/WordPress-iOS/issues/15854)

### Possible future improvements

[...]

PR #16119, opened 2021-03-17
Add large titles to the Notifications screen

Gio2018:
Ref #15750

To test:

- Test on iPhone and iPad
- set the `newNavBarAppearance` feature flag to `false`
- Build, run and make sure that there is no disruption on the existing UI of the Notifications screen
- Discard the changes to the `newNavBarAppearance` feature flag
- Build/run again and go to the Notifications screen
- Make sure the large title appears and that the UI behaves as expected
- Navigate into notifications details and make sure that they still look as they used to.

<p align=center>

</p>

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -464,6 +464,8 @@ private extension NotificationsViewController {
         navigationController?.navigationBar.prefersLargeTitles = true
         navigationItem.largeTitleDisplayMode = .always
         // Don't show 'Notifications' in the next-view back button
+        // we are using a space character because we need a non-empty string to ensure a smooth
+        // transition back, with large titles enabled.
         navigationItem.backBarButtonItem = UIBarButtonItem(title: " ", style: .plain, target: nil, action: nil)
         navigationItem.title = NSLocalizedString("Notifications", comment: "Notifications View Controller title")
     }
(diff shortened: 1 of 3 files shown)


frosty (commented):
Code changes look good!

I noticed the title animation isn't quite 'normal' – the title disappears instantly and then re-animates from the middle of the view instead of into the place of the back button on the following screen:

note-push-1

I think this is because we set a back button with an empty title for the Notifications view controller. I think the fix is either to remove this, or it seems to help by setting the title to a single space rather than an empty string, in `setupNavigationBar` of `NotificationsViewController`:

`navigationItem.backBarButtonItem = UIBarButtonItem(title: " ", style: .plain, target: nil, action: nil)`

Which gives us this animation:

note-push-2

It's still not quite right when navigating to a Reader detail view from Notifications, but I'm not sure what else would fix that... It's also not super obvious at normal speed, so I think we're okay to run with it for now.

note-push-3

WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
```diff
@@ -461,10 +461,10 @@ class NotificationsViewController: UITableViewController, UIViewControllerRestor
 //
 private extension NotificationsViewController {
     func setupNavigationBar() {
-        // Don't show 'Notifications' in the next-view back button
         navigationController?.navigationBar.prefersLargeTitles = true
         navigationItem.largeTitleDisplayMode = .always
-        navigationItem.backBarButtonItem = UIBarButtonItem(title: String(), style: .plain, target: nil, action: nil)
+        // Don't show 'Notifications' in the next-view back button
+        navigationItem.backBarButtonItem = UIBarButtonItem(title: " ", style: .plain, target: nil, action: nil)
```
frosty, line 469: Sorry, can I nitpick and ask that we add a small comment to explain why it's a space character and not an empty string? 🙂

frosty (approved):

PR #16124, opened 2021-03-18
Jetpack App: New Target, No App Extensions

momo-ozawa:
## Description

This PR adds a new Jetpack target and scheme, so we can build and run a Jetpack app 🚀 

After investigating several different approaches to building a new Jetpack app, we decided that a target / scheme approach was the most feasible solution.

## Changes

- Added a new **Jetpack target and scheme**
- Removed all **App Extensions** from the Jetpack target
- Updated the **Podfile** to include the Jetpack target
- Added **Jetpack icons and launch screens**
- Added a **Jetpack flag** to preprocessor macros

## Screenshots

Different app icons | Running both apps | Jetpack launch screen | Jetpack preprocessor flag 
--- | --- | --- | ---
Simulator Screen Shot - iPhone 12 Pro - 2021-03-18 at 15 25 11 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-18 at 15 25 44 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-18 at 15 25 14 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-18 at 15 25 23

## How to test

1. Run `bundle exec pod install`
2. Build the WordPress scheme
3. ✅ Should build as usual
4. Build the Jetpack scheme
5. ✅ Should build and it should have a 1) different icon 2) different splash screen
6. In the Jetpack app, go to My Site > Switch Site
7. ✅ A "Welcome to the Jetpack app!" notice should be printed in the console
8. In the WordPress app, go to My Site > Switch Site
9. ✅ A "Welcome to the Jetpack app!" notice should NOT be printed in the console

## Additional information

* Splash screen and icons are just placeholders and not the final ones

You can find the other explorations here:

* As a new project: https://github.com/wordpress-mobile/WordPress-iOS/pull/16104
* New target + duplicated app extensions: https://github.com/wordpress-mobile/WordPress-iOS/pull/16111
* New target (similar to this one): https://github.com/wordpress-mobile/WordPress-iOS/pull/16095
* We also briefly looking into Tuist and using a script to keep two projects in sync, both alternatives have a high cost


leandroalonso (commented):
All works as described!

A star is born! ⭐ 

@jkmassel can we get your blessing on this one please, sir?

Podfile
```diff
@@ -85,6 +85,73 @@ def shared_with_extension_pods
     pod 'Down', '~> 0.6.6'
 end
 
+def shared_with_app_target_pods
```
jkmassel: This approach works, but is IMHO somewhat fragile – as a for-instance, if you (or the WP devs) need to customize WPAuth or another dependency, how would you (or they) specify a different version without causing overlapping breaking changes?

The better approach here is probably an `abstract_target` (see https://guides.cocoapods.org/using/the-podfile.html and search for "abstract target"). This allows inheritance, which is probably preferable to composability in this case IMHO.

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -313,7 +313,7 @@
 		1ABA150822AE5F870039311A /* WordPressUIBundleTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1ABA150722AE5F870039311A /* WordPressUIBundleTests.swift */; };
 		1AE0F2B12297F7E9000BDD7F /* WireMock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AE0F2B02297F7E9000BDD7F /* WireMock.swift */; };
 		1AED7239229D2E260036C5B8 /* WireMock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AE0F2B02297F7E9000BDD7F /* WireMock.swift */; };
-		1D36FCB53C05724865D41F7A /* Pods_WordPress.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 556E3A9C1600564F6A3CADF6 /* Pods_WordPress.framework */; };
+		1D36FCB53C05724865D41F7A /* (null) in Frameworks */ = {isa = PBXBuildFile; };
```
jkmassel, line 316: I'm not sure this is correct?

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -5519,8 +7231,9 @@
 				93F2E5401E9E5A180050D489 /* libsqlite3.tbd in Frameworks */,
 				FD21397F13128C5300099582 /* libiconv.dylib in Frameworks */,
 				E19DF741141F7BDD000002F3 /* libz.dylib in Frameworks */,
-				1D36FCB53C05724865D41F7A /* Pods_WordPress.framework in Frameworks */,
+				1D36FCB53C05724865D41F7A /* (null) in Frameworks */,
```
jkmassel: Similar issue here

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -18780,6 +22459,279 @@
...
+				ENABLE_BITCODE = NO;
+				FRAMEWORK_SEARCH_PATHS = "$(inherited)";
+				GCC_OPTIMIZATION_LEVEL = 0;
+				GCC_PREFIX_HEADER = WordPress_Prefix.pch;
+				GCC_PREPROCESSOR_DEFINITIONS = (
+					"$(inherited)",
+					"COCOAPODS=1",
+					"NSLOGGER_BUILD_USERNAME=\"${USER}\"",
+					"WPCOM_SCHEME=\\@\\\"${WPCOM_SCHEME}\\\"",
+					"JETPACK=1",
```
jkmassel: If you're specifying this here, you might consider also specifying it under "other swift flags" in order to have it available in Swift as well (without having to do any bridging for this special case).

jkmassel (changes requested):
This looks like a solid approach!

A few suggestions:
- It's probably not worth bringing over all of the different sub-targets at this point – I'd consider just bringing over the main `Jetpack` one for now and getting rid of the `Alpha` and `Internal` builds for now? 
- It's probably a good idea to define both a preprocessor macro for Obj-C and an "Other Swift Flags" entry for Swift
- For your consideration (probably not as part of this PR), but it might make sense to make a Swift module for Jetpack code where all that can live?

WordPress/WordPress.xcodeproj/project.pbxproj
momo-ozawa: Fixed! 923be31

WordPress/WordPress.xcodeproj/project.pbxproj
momo-ozawa: Fixed! 923be31

WordPress/WordPress.xcodeproj/project.pbxproj
momo-ozawa: Fixed! https://github.com/wordpress-mobile/WordPress-iOS/pull/16124/commits/7038bc1af53a62fe106e1a578a43a0a178bf0837

Podfile
momo-ozawa: Good point! Thanks for the tip. Fixed in a905e59

momo-ozawa:
@jkmassel Thanks for reviewing! Ready for another look 🙇‍♀️ 

>  It's probably not worth bringing over all of the different sub-targets at this point – I'd consider just bringing over the main Jetpack one for now and getting rid of the Alpha and Internal builds for now?

Fixed in c7206b8

> It's probably a good idea to define both a preprocessor macro for Obj-C and an "Other Swift Flags" entry for Swift

Great point. Fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/16124/commits/7038bc1af53a62fe106e1a578a43a0a178bf0837

> For your consideration (probably not as part of this PR), but it might make sense to make a Swift module for Jetpack code where all that can live?

Added as [a TODO to our project backlog](https://github.com/orgs/wordpress-mobile/projects/90#card-57397197) 👍

jkmassel (approved):
@momo-ozawa – sorry about the wait – this looks great.

Merge when ready (pending the `Podfile.lock` fix – no need to re-review for it)

PR #16132, opened 2021-03-19
Update NewBlogDetailHeaderView sizing

frosty:
A few tweaks to the new blog detail header sizing:

* TitleView constrains to safe layout guides horizontally
* Action button row either matches titleview or maxes out at 390 wide

| Device | Portrait | Landscape |
|---|---|---|
| iPhone 12 Pro | Simulator Screen Shot - iPhone 12 Pro - 2021-03-19 at 10 57 20 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-19 at 10 57 24 |
| iPhone 12 Pro Max | Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-19 at 11 00 00 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-19 at 11 00 09 |
| iPad Pro 9.7" Split screen | Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-03-19 at 11 11 32 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-03-19 at 11 11 30 |
| iPad Pro 9.7" Full screen | Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-03-19 at 11 11 41 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-03-19 at 11 11 39 |

**Known Issue**

The spacing of the buttons on iPhone Pro Max in landscape is a little off – ideally I'd like the trailing edge of the buttons to match the trailing edge of the title view, but I haven't got that working at this stage. It's currently balancing out the left and right spacing (because the system has large spacing on the left edge to avoid the notch).

**To test**

- Build and run
- Test on a variety of devices, including portrait and landscape

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
@@ -175,7 +175,7 @@ extension UIColor {
     /// Muriel/iOS navigation color
     static var appBarBackground: UIColor {
         if FeatureFlag.newNavBarAppearance.enabled {
-            return UIColor(light: white, dark: .gray(.shade100))
+            return .secondarySystemGroupedBackground
         }
 
         return UIColor(light: .brand, dark: .gray(.shade100))
--- a/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
@@ -63,12 +63,17 @@ extension WPStyleGuide {
         appearance.largeTitleTextAttributes = largeTitleTextAttributes
 
         if FeatureFlag.newNavBarAppearance.enabled {
-            appearance.shadowColor = .clear
+            appearance.shadowColor = .separator
+
+            let scrollEdgeAppearance = appearance.copy()
+            scrollEdgeAppearance.shadowColor = .clear
+            navigationAppearance.scrollEdgeAppearance = scrollEdgeAppearance
+        } else {
+            navigationAppearance.scrollEdgeAppearance = appearance
         }
 
         navigationAppearance.standardAppearance = appearance
-        navigationAppearance.scrollEdgeAppearance = navigationAppearance.standardAppearance
-
+        navigationAppearance.compactAppearance = appearance
 
         // Makes bar buttons visible in "Other Apps" media source picker.
         // Setting title text attributes makes bar button items not go blank when switching between the tabs of the picker.
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
@@ -659,7 +659,7 @@ NSString * const WPCalypsoDashboardPath = @"https://wordpress.com/stats/";
 
 - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)sectionNum {
     BlogDetailsSection *section = self.tableSections[sectionNum];
-    if (section.showQuickStartMenu == true || [Feature enabled:FeatureFlagNewNavBarAppearance]) {
+    if (section.showQuickStartMenu == true || (sectionNum == 0 && [Feature enabled:FeatureFlagNewNavBarAppearance])) {
         return BlogDetailQuickStartSectionHeight;
     } else if (([section.title isEmpty] || section.title == nil) && sectionNum == 0) {
         // because tableView:viewForHeaderInSection: is implemented, this must explicitly be 0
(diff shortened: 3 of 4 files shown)


osullivanchris:
👋  Hello hello and thanks for the build Its looking really good. I reviewed on iPhone 12 Pro. I am having troubles with my iPad so just going off screenshots there for now...

**General notes**

- The pages in white header modals like site switcher and notification settings seem perfect to me. Similarly I can't see any flaws with 'Notifications'. 
- I know that reader has issues which we'll discuss separately.
- Plus what you noted on iPhone Pro Max landscape.
- I think we need dividers below small headers (and when large header collapses) but I know you said its not trivial

**Small tweaks on My Site**

1. Can we add a divider below the header area, with the same colour and style as the list items?
2. Can we add 8px more padding here? It looks like 16px in the build, I had 24px. I think the additional space will make it a little more balanced.
3. A tricky one, but we have a bit of jagged alignment going on down the right hand side between the profile image, chevron, edge of the Quick Action, FAB. I am thinking we could align the chevron vertically centred with the profile image. To do so by my eye would mean moving the chevron 8px further right. However, I know this only applies specifically to iPhone portrait, as the 'me' icon will drift off to the right further on iPad and landscape. so if you think its nonsense thats ok!!

Screenshot 2021-03-19 at 12 36 00


And one more in dark mode, can we make the background colour of that header area the same colour as the list items background?
F4E99900-4A43-4BC8-BA9B-3EDAE1C45325

None of these are huge blockers. If you are prioritising I think 1 and 2 on My Site should hopefully be small at least and tighten things up.

osullivanchris (changes requested):
forgot to use 'add review' button 😄

osullivanchris:
Sorry adding one more that I spotted. In the design, there is less space between the site title and site URL. And it means that the space above and below the two pieces of text is equal (11 and 11 visually).

Don't take these numbers as a rule, as the code might not measure exactly to the top and bottom of the font like this. But is there a space we have defined between the title and URL, and can we reduce it a little bit? If we could trim 4px there, I think it would balance it out. We might need to reduce the space between above the site title 1 or 2px as well to balance it.


Screenshot 2021-03-19 at 15 58 21

frosty:
Thanks @osullivanchris! I've pushed up some changes for your first set of issues:

- On the main My Site screen, when you scroll you now see a separator. This is also present on all subscreens, but I'm looking at if we can remove that when we show a filter bar. This will likely be a change in a future PR though.
- I increased the bottom padding below the quick action buttons.
- I tweaked the background of the My Site header to now use `secondarySystemGroupedBackground` (previously was white in light mode and gray-100 in dark mode). I also changed the site address button to use the `primary` colour, which makes it higher contrast in dark mode.
- I added a separator below the large header in My Site.

frosty:
|   |   |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-19 at 15 28 45 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-19 at 15 29 01 |

frosty:
I took a look at aligning the chevron, but it's a little problematic because it doesn't know anything about the placement of the Me button (they're entirely separate views) – as you mentioned, if I fix it so it looks okay on iPhone, then it's out of place on Max and iPad. Can take another look at this in the future.

osullivanchris (approved):

osullivanchris:
all these improvements look good:
- extra padding
- changed background in dark mode
- divider below the header area
- dividers on all small headers

as discussed, we can later look at

1. the reader large title
2. spacing between the site title and URL
3. removing dividers when there is tabs below them to create one large visual header
4. aligning the site picker chevron with the 'me' icon

I know 3 and 4 are not straight forward. 1 and 2 would be nice to get in for when it is released to users.

Gio2018 (approved):
Tested on iPhone 11 Pro Max and iPad Pro 2nd Gen 11 (devices, iOS 14.4) and looked good!

:shipit: !

osullivanchris:
Was discussing the new titles/headers with @mattmiklic  and we noticed that in dark mode the nav bar has a different background colour to the tabs. When looking at it, Matt was also interested in looking at using system materials for those backgrounds again to get the native blur into play. But just fixing the colours as a bug fix could be a first step.

IMG_0863
IMG_0862


Per my previous list above the outstanding tweaks then are (in a rough priority order):

1. Reader Large Title (we now have it in, but we don't have the collapse on scroll behaviour)
2. Changing the navigation bar header colour on dark mode (issue raised above)
3. Space between site title and URL
4. Removing dividers when there is tabs below them to create one large visual header
5. Aligning the site picker chevron with the 'me' icon

Again, the last two are low priority maybe ones. The others would be good to fix. Wondering if there's somewhere else I should post this now as an improvement ticket or something? Cc @startuptester  who I know was looking at this seeing as it went live in this release.

Thanks 😄

frosty:
@osullivanchris Thanks for rounding up this list – I've made sure they're all captured on the [main issue for the large titles project](https://github.com/wordpress-mobile/WordPress-iOS/issues/15750).

osullivanchris:
@frosty  a-ha! Perfect. Thanks for finding the correct home for it. I knew this didn't feel like the right place. Looks good!

PR #16138, opened 2021-03-22
Merge 16.9 final into develop

jkmassel:
Merges beta changes from 16.9 into `develop`.


Podfile.lock
```diff
@@ -513,10 +513,6 @@ DEPENDENCIES:
   - ZIPFoundation (~> 0.9.8)
 
 SPEC REPOS:
-  https://github.com/wordpress-mobile/cocoapods-specs.git:
-    - WordPressAuthenticator
-    - WordPressShared
-    - WordPressUI
```
mokagio, line 519: Not sure at what point the change to use our own CocoaPods spec repo was reverted. Was it intentional?

fastlane/Fastfile
```diff
@@ -238,7 +238,7 @@ import "./ScreenshotFastfile"
 
     # Wrap up
     version = ios_get_app_version()
-    ios_clear_intermediate_tags(version: version)
+    #ios_clear_intermediate_tags(version: version)
```
mokagio, line 241: Normally, I'd say "could you please leave a comment here?" but I know you're onto this and the change will be short lived 👍

mokagio (commented):
LGTM but holding the ✅ just to make sure the change to the pods spec was intentional (or is reverted, or we agree to merge as-is and followup later).

Podfile.lock
AliSoftware: I think that's due to the glitch I discovered when I worked on the feature. See paaHJt-1SO-p2#what-changes for details.

I think that's a bug in CocoaPods, defaulting to use the CDN as source for _transitive_ dependencies, and only use the declared source order for root dependencies. I yet have to open a GitHub issue about this on CocoaPods' repo.

 - Not sure we can do much about it except reporting and waiting for CP to fix it.
 - One way to workaround this is to manually edit the `Podfile.lock` to switch those back to the `cocoapods-specs.git` entry (then `rm -rf Pods/` for good measure), then re-run `pod install`… not ideal but it seems it worked last time I tried
 - If it doesn't work, in the end that's not that big a deal as long as it finds the podspec and the right version – since the podspecs are the same in both cases anyway – even if it's quite unfortunate.

jkmassel: Took a look at this and chatted about it with @AliSoftware, and we'll land it as-is for now pending a response from the CocoaPods folks. Thanks!

AliSoftware: Issue reported in https://github.com/CocoaPods/CocoaPods/issues/10539

WordPress/WordPressIntents/en.lproj/Sites.strings
```diff
@@ -0,0 +1,12 @@
...
+
+"ILcGmf" = "Site";
+
+"cyajMn" = "Site";
+
+"gpCwrM" = "Select Site";
+
+"s4dJhx" = "Just to confirm, you wanted ‘${site}’?";
+
+"tVvJ9c" = "Select Site Intent";
```
AliSoftware, line 11: There are no way to have nicer, better keys? (I'm assuming those are Xcode-generated, like when you ask Xcode to extract the strings from a XIB or Storyboard, but for Intents?)

Not even sure those keys are even picked up by our localization tools currently, but when we upgrade our tools so that they do, it would be nice to rely on keys with clear intent and naming (reverse-DNS style for example, like we did for Widgets)

WordPress/WordPressIntents/en.lproj/Sites.strings
```diff
@@ -0,0 +1,12 @@
+"BOl9KQ" = "There are ${count} options matching ‘${site}’.";
+
+"ILcGmf" = "Site";
+
+"cyajMn" = "Site";
+
+"gpCwrM" = "Select Site";
+
+"s4dJhx" = "Just to confirm, you wanted ‘${site}’?";
```
AliSoftware, line 9: I didn't know this was possible to inject variables like that via `${site}`. I'm guessing this syntax is Intent-specific and the iOS runtime will automatically pick up the Intent's parameter with that name in those cases?

fastlane/Fastfile
AliSoftware: Tbh I'm not sure why we have that step (while we don't clear intermediate tags in Android), and I'd vote to remove it entirely… 😉

AliSoftware (approved):
I'm ok with merging with the current state of the `Podfile.lock` (i.e. even if we still have that https://github.com/CocoaPods/CocoaPods/issues/10539 issue), as that just means we point to the podspecs in the CDN instead of our mirror, but since both podspecs have the same content that doesn't really matter (that just means we don't take advantage of our mirror like we hoped to, but once the version is sync'd to CDN, that doesn't really matter)

I still left a couple of questions, related to the Intent localizations. That can be addressed in a separate PR though (as this PR is about merging a published version to store before cutting the new code freeze), just would love to have those points not be forgotten and checked at some point.

WordPress/WordPressIntents/en.lproj/Sites.strings
jkmassel: Yes – definitely agree with all of this – for now this was to silence warnings in the project and get the translations moving. Worst-case, we might end up just hand-copying them into position until the tooling improvements are present.

WordPress/WordPressIntents/en.lproj/Sites.strings
jkmassel: I didn't either – this was auto-generated by Xcode though, and like you say, appears to be intent-specific.

PR #16139, opened 2021-03-22
Merge 16.9 final into trunk

jkmassel:
Merges all changes from the `16.9` series into `trunk`.


mokagio (approved):
I guess the same question regarding the `Podfile` [as the `develop` counterpart applies](https://github.com/wordpress-mobile/WordPress-iOS/pull/16138#discussion_r598424981) here, but, because the merges on `trunk` are just a reference of the state of the latest published app, I think that doesn't block the merge here.

mokagio:
Took the liberty of converting this to draft so that we'll merge it only once 16.9 is deployed.

By the way, before doing so I checked and saw that its status is "Pending Developer Release" on ASC. I guess I could have shipped it and then merged this but it felt "out of place". I feel I don't have enough context on the status to confidently hit that button. 

That's probably just me being paranoid, but maybe there's something we can do to clearly communicate whether a build is good to release, so the first person who gets there can do it?

mokagio:
Just noticed the PR state changed to out-of-date with the base branch after I switched it to draft. 🤔 🤔 🤔 Why would that happen?!

image

AliSoftware:
> By the way, before doing so I checked and saw that its status is "Pending Developer Release" on ASC. I guess I could have shipped it and then merged this but it felt "out of place". I feel I don't have enough context on the status to confidently hit that button.

I think it's better to let the Release Manager push the button. Because releasing to the App Store is not just about pushing the button after all, but also about checking Sentry, communicating (in Slack, P2s, etc), checking the box in the scenario to keep track of where they were at, and there can be other things you're not aware of, for example if someone ping'd the RM during the week-end to warn about a last minute emergency bug that required to hold off the release, but the RM didn't get the notification yet during their week-end so didn't invalidate it in ASC yet (but would have first thing in their morning on Monday), etc.

So sure it could help, but since it's not a big deal to wait just a couple of hours more for the RM to get their Monday starting and release the app live anyway, I feel it's better to let the RM own their own process (unless they asked otherwise or are AFK and asked for us to cover, of course)

AliSoftware:
> Just noticed the PR state changed to out-of-date with the base branch after I switched it to draft. 🤔 🤔 🤔 Why would that happen?!

Not sure, especially since there's no reason why it would indeed go out-of-sync given `trunk` haven't changed since. The change of status of the PR to draft probably made GitHub trigger a new check about branches being out of sync, but I don't get why it would conclude that those branches would be out-of-date.

That being said, that's not the first time I see GitHub mis-reporting "out-of-date" status, and if this is the same as my past experiences with that GitHub glitch, I am hopeful that as soon as we'll turn this PR back into "Ready for review" the "Merge" button will still be green and ready to be pushed without any sync required.

jkmassel:
Yeah, I get this message every time I merge to `trunk` in this repo, but it never stops me from doing so 🤔

PR #16142, opened 2021-03-22
Fix UI tests for large title changes

frosty:
This PR fixes UI tests that were broken when enabling the large titles / white headers feature flag – primarily due to the change in view hierarchy in the My Site section of the app. Blog details / My Site is now the the top level screen in the My Site section, rather than the blog list. There is a new site switcher button to change site.

**To test**

- Run the mocks server: `rake mocks`
- Build and run tests on the WordPressUITests target
- If you like, also do a cursory look at running the app yourself. The one change I made there is when removing the last self hosted site in the app, we now return to the login screen.
- To test that, you could check that if you remove the last WordPress.com site in the app while logged in, you stay logged in but see the **Add new site** button on the My Site screen. And also check that if you add a self hosted site without being logged into wpcom, then remove that site, you're returned to login.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
+++ b/WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift
@@ -40,7 +40,7 @@ enum FeatureFlag: Int, CaseIterable, OverrideableFlag {
         case .gutenbergXposts:
             return true
         case .newNavBarAppearance:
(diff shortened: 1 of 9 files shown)


Gio2018 (approved):
UI Tests ran successfully on iPhone 12 Pro simulator with iOS 14.4.
Also tested removing last WP.com site and add/remove a self hosted site while not signed in, and worked as described.

:shipit:

PR #16156, opened 2021-03-24
Respect API ordering for page template categories

mkevins:
This PR adds an ordinal attribute to the `PageTemplateCategory` data model to explicitly use the API ordering while caching results from the remote fetch.

To test:

* Add a new page (from the my sites tab or page list)
* Observe the order of the categories in the filter bar and sections
  * The ordering should reflect the API order (e.g. in my case it was Blog, Highlights, Splash, About, Links, Home, Team, Services, Portfolio, Contact, Gallery, Featured, but generally it should not be strictly alphabetical).

PR submission checklist:


WordPress/Classes/Services/PageLayoutService.swift
```diff
@@ -82,7 +82,7 @@ extension PageLayoutService {
     static func resultsController(forBlog blog: Blog, delegate: NSFetchedResultsControllerDelegate? = nil) -> NSFetchedResultsController<PageTemplateCategory> {
         let context = ContextManager.shared.mainContext
         let request: NSFetchRequest<PageTemplateCategory> = PageTemplateCategory.fetchRequest(forBlog: blog)
-        let sort = NSSortDescriptor(key: "title", ascending: true)
+        let sort = NSSortDescriptor(key: "ordinal", ascending: true)
```
chipsnyder: Not a change request: To share a neat trick I recently came across you can leverage the name of the value we're sorting on by:
```suggestion
        let sort = NSSortDescriptor(key: #keyPath(PageTemplateCategory.ordinal), ascending: true)
```

chipsnyder (changes requested):
Thanks for making these changes @mkevins Overall it looks good except for one issue I came across when attempting a CoreData migration from the previous version. I used `trunk` to install the app and cache some layouts then applied your changes to test that the migration went smoothly. Unfortunately I got this error: 
```
UserInfo={entity=PageTemplateLayout, attribute=demoUrl, reason=Validation error missing attribute values on mandatory destination attribute}";
```

Based on this I believe we just need to add a default value to `demoUrl` or make it optional but I didn't have enough time today to validate that.

This [looks like](https://github.com/wordpress-mobile/WordPress-iOS/blob/release/17.0/WordPress/Classes/WordPress.xcdatamodeld/WordPress%20117.xcdatamodel/contents#L459) It will be a concern for 17.0 as well so we should validate and apply a fix before that goes live.

WordPress/Classes/Services/PageLayoutService.swift
mkevins: Thanks Chip! I really like this (it kinda bothered me that this was just a string literal). :+1:

mkevins:
Thanks for testing this Chip!

> when attempting a CoreData migration from the previous version. I used `trunk` to install the app and cache some layouts then applied your changes to test that the migration went smoothly. Unfortunately I got this error:

Nice catch! I overlooked this when adding the non-optional attributes. I pushed a fix here: https://github.com/wordpress-mobile/WordPress-iOS/pull/16167 for those, and for the `ordinal` attribute attribute introduced in this PR, the default value was already set.

chipsnyder:
>Nice catch! I overlooked this when adding the non-optional attributes.

No problem 👍 

I added a 118 model in https://github.com/wordpress-mobile/WordPress-iOS/pull/16170 but If I get some time today I can help create a PR to this branch that applies the CoreData changes.

mkevins:
Thanks Chip! This one's ready for another look.

MIGRATIONS.md
```diff
@@ -16,7 +23,7 @@ data model as well as any custom migrations.
 
 @mkevins 2021-03-17
 
-- `PageTemplateLayout`: added the entities
+- `PageTemplateLayout`: added the attributes
```
chipsnyder, line 26: Good catch here 👍

chipsnyder (approved):
LGTM! I tested by performing migrations from:
16.9 -> this branch
17.0 -> this branch

PR #16168, opened 2021-03-26
Remove AccountService from all the things

jkmassel:
Disclaimer: this PR description is written by @crazytonyli, not Jeremy. 😸 

----

## Changes

The main change is in [AccountService](https://github.com/wordpress-mobile/WordPress-iOS/pull/16168/files#diff-93852dcd12ccf58883495b486b5f8e5733039063979dea8f9a2f0932d22d3ef0), removing a few account lookup APIs from it. The rest changes are making the migration, using the APIs in [WPAccount+Lookup](https://github.com/wordpress-mobile/WordPress-iOS/blob/93d5bf4cf8f1853cae63a859ba0bf7159fab0070/WordPress/Classes/Models/WPAccount+Lookup.swift) instead. Even though there are lots of files are changed, but the changes are quite similar, I would recommend utilize the "Viewed" checkbox at top right of each file in the PR diff page—it sure helped me reviewing this PR.

## Test Instructions

Making sure features only available for logged in user are working as expected, and vise-versa. i.e. user is able to edit draft in their sites, not in others' sites.

PR submission checklist:

  _N/A. No new API is added in this PR_
  _N/A_
  _Not Needed_


mokagio:
Pushing this PR to the next milestone, 17.9, because I'm running the code-freeze today and this is still a draft.

mokagio:
Pushing this PR to the next milestone, 18.0, because I'm running the code-freeze today and this is still a draft.

@jkmassel is this a change that we can implement in pieces? It might make it easier to get the ball rolling?

mokagio:
Pushing this PR to the next milestone, 18.1, because I'm running the code-freeze today and this is still a draft.

@jkmassel is this a change that we can implement in pieces? It might make it easier to get the ball rolling? How can I help?

mokagio:
Pushing this PR to the next milestone, 18.2, because I'm running the code-freeze today and this is still a draft.

mokagio:
Pushing this PR to the next milestone, 18.3, because I'm running the code-freeze today and this is still a draft.

mokagio:
Pushing this PR to the next milestone, 18.4, because I'm running the code-freeze today and this is still a draft.

mokagio:
Pushing this PR to the next milestone, 18.5, because I'm running the code-freeze today and this is still a draft.

mokagio:
Pushing this PR to the next milestone, 18.6, because I'm running the code-freeze today and this is still a draft.

mokagio:
After bumping this to the next milestone for more than 10 release cycles, I think it's time to park this in _Someday_.

WordPress/Classes/Services/AccountService+MergeDuplicates.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 extension AccountService {
 
     func mergeDuplicatesIfNecessary() {
-        guard numberOfAccounts() > 1 else {
+        guard ((try? WPAccount.lookupNumberOfAccounts(in: managedObjectContext)) ?? 0) > 1 else {
```
mokagio: Nitpick. This code sort of reimplements the logic of the check with that `?? 0`. Have you considered splitting the `guard`?

```swift
guard let count =  try? WPAccount.lookupNumberOfAccounts(in: managedObjectContext)), count > 1 else {
    return
}
```

WordPress/Classes/Services/AccountService.m
```diff
@@ -530,18 +480,19 @@ - (void)purgeAccountIfUnused:(WPAccount *)account
 
 - (void)setVisibility:(BOOL)visible forBlogs:(NSArray *)blogs
 {
+    WPAccount *account = [WPAccount lookupDefaultWordPressComAccountInContext:self.managedObjectContext];
```
mokagio: How do you feel about calling this `defaultAccount` to make it clearer what it is later on when it's used?

WordPress/Classes/Services/AccountSettingsService.swift
```diff
@@ -226,8 +226,8 @@ class AccountSettingsService {
     }
 
     fileprivate func createAccountSettings(_ userID: Int, settings: AccountSettings) {
-        let accountService = AccountService(managedObjectContext: context)
-        guard let account = accountService.findAccount(withUserID: NSNumber(value: userID)) else {
+
+        guard let account = try? WPAccount.lookup(withUserID: Int64(userID), in: context) else {
```
mokagio, line 230: In the future, it would be cool to remove the need for the `Int64` typecast.

WordPress/Classes/Services/WordPressComSyncService.swift
```diff
@@ -48,12 +48,12 @@ class WordPressComSyncService {
...
         }
 
-        if isJetpackLogin && !accountService.isDefaultWordPressComAccount(account) {
+        if isJetpackLogin && !account.isDefaultWordPressComAccount {
             let blogService = BlogService(managedObjectContext: context)
             blogService.associateSyncedBlogs(toJetpackAccount: account, success: onSuccessInternal, failure: onFailureInternal)
 
         } else {
-            if accountService.defaultWordPressComAccount()?.isEqual(account) == false {
+            if !account.isDefaultWordPressComAccount {
```
mokagio, line 56: Nitpick / Personal opinion

I find `!something` to be less clear than `something == false`. That `!` at the start is easy to miss when looking at the code.

I guess you don't think so given the change in this line?

WordPress/Classes/Utility/AccountHelper.swift
```diff
@@ -43,14 +42,13 @@ import Foundation
...
-        let accountService = AccountService(managedObjectContext: context)
         let blogService = BlogService(managedObjectContext: context)
         let allBlogs = blogService.blogsForAllAccounts()
         let blogsByAccount = Dictionary(grouping: allBlogs, by: { $0.account })
 
-        let defaultAccount = accountService.defaultWordPressComAccount()
+        let defaultAccount = try? WPAccount.lookupDefaultWordPressComAccount(in: context)
 
-        let accountCount = accountService.numberOfAccounts()
+        let accountCount = (try? WPAccount.lookupNumberOfAccounts(in: context)) ?? 0
```
mokagio, line 51: After seeing a few of these nil-coalescing to `0`,  I wonder whether we should make returning `0` the method behavior as opposed to throwing an error?

Not a must have for this PR to be clear, just an idea. None of the code I've seen so far uses the error, so I wondered whether we can safely swallow it.

WordPress/Classes/Utility/ZendeskUtils.swift
```diff
@@ -64,8 +63,14 @@ extension NSNotification.Name {
         return Locale.preferredLanguages[0]
     }
 
+    private let contextManager: CoreDataStack
+
     // MARK: - Public Methods
 
+    init(contextManager: CoreDataStack) {
+        self.contextManager = contextManager
+    }
```
mokagio, line 72: 🎉 for Dependency Injection

WordPress/Classes/ViewRelated/Blog/Site Management/DeleteSiteViewController.swift
```diff
@@ -230,8 +230,11 @@ open class DeleteSiteViewController: UITableViewController {
 
                                     self?.updateNavigationStackAfterSiteDeletion()
 
-                                    let accountService = AccountService(managedObjectContext: ContextManager.sharedInstance().mainContext)
-                                    accountService.updateUserDetails(for: (accountService.defaultWordPressComAccount()!),
+                                    let context = ContextManager.shared.mainContext
+                                    // TODO: This force-unwrap was present before refactoring the Core Data Stack – we should remove it
+                                    // at the first opportunity
+                                    let account = try! WPAccount.lookupDefaultWordPressComAccount(in: context)!
```
mokagio, line 236: 👌 seems like an appropriate use case for the "[Type] Tech Debt" label.

Logged to keep track of this https://github.com/wordpress-mobile/WordPress-iOS/issues/19132

WordPress/Classes/ViewRelated/Me/App Settings/PrivacySettingsViewController.swift
```diff
@@ -124,8 +124,13 @@ class PrivacySettingsViewController: UITableViewController {
         let appAnalytics = WordPressAppDelegate.shared?.analytics
         appAnalytics?.setUserHasOptedOut(!enabled)
 
-        let accountService = AccountService(managedObjectContext: ContextManager.sharedInstance().mainContext)
-        AccountSettingsHelper(accountService: accountService).updateTracksOptOutSetting(!enabled)
+        let context = ContextManager.shared.mainContext
+        guard let account = try? WPAccount.lookupDefaultWordPressComAccount(in: context) else {
```
mokagio, line 128: Curious: How come this code defines `context` in a constant where most (all?) of the code I've seen so far accesses it inline when calling `lookupDefault...`?

No an issue and either way is legit. Purely just curious 🤷‍♂️

WordPress/Classes/ViewRelated/Me/My Profile/MyProfileViewController.swift
```diff
@@ -187,11 +187,6 @@ private class MyProfileController: SettingsController {
     // FIXME: (@koke 2015-12-17) Not cool. Let's stop passing managed objects
     // and initializing stuff with safer values like userID
     fileprivate func defaultAccount() -> WPAccount? {
-        let context = ContextManager.sharedInstance().mainContext
-        let service = AccountService(managedObjectContext: context)
-        let account = service.defaultWordPressComAccount()
-        // Again, ! isn't cool, but let's keep it for now until we refactor the VC
-        // initialization parameters.
-        return account
```
mokagio, line 195: These comments made me smile...

1. `FIXME` dating back to 2015/12, 4.5 years ago! Fixme comments get lost easily, but also: it's been working fine all this time, even if the code implementation wasn't ideal
2. The comment about force unwrapping (`! isn't cool`) was outdated. Classic example of how easily comments become noise even when the most well intentioned developers work with the code

WordPress/Classes/ViewRelated/NUX/SignupEpilogueViewController.swift
```diff
@@ -197,11 +197,13 @@ private extension SignupEpilogueViewController {
...
-            let api = account.wordPressComRestApi else {
-                navigationController?.popViewController(animated: true)
-                return
+
+        guard
+            let account = try? WPAccount.lookupDefaultWordPressComAccount(in: context),
+            let api = account.wordPressComRestApi
+        else {
+            navigationController?.popViewController(animated: true)
+            return
```
mokagio, line 206: Wow, this (already existing) UX is far from user friendly... Something went wrong? I'll just dismiss the screen you were onto without giving an explanation.

🤔 

It seems unlikely we'll ever end in this branch because one shouldn't be able to change their username if they don't have an account, but never say never.

I wonder where's the best place to log this. I also think the fix shouldn't be in this method, but in the caller. It should be the caller's responsibility to verify `account` and `api` are available and either pass them to this method or show an error to the user. (Just an idea)

WordPress/Classes/ViewRelated/NUX/SignupEpilogueViewController.swift
```diff
@@ -238,22 +240,32 @@ private extension SignupEpilogueViewController {
...
 
-        let accountSettingService = AccountSettingsService(userID: defaultAccount.userID.intValue, api: restApi)
+            let accountSettingService = AccountSettingsService(userID: account.userID.intValue, api: restApi)
 
-        accountSettingService.updatePassword(newPassword) { (success, error) in
-            if success {
-                WordPressAuthenticator.track(.signupEpiloguePasswordUpdateSucceeded, properties: self.tracksProperties())
-            } else {
-                WordPressAuthenticator.track(.signupEpiloguePasswordUpdateFailed, properties: self.tracksProperties())
+            accountSettingService.updatePassword(newPassword) { (success, error) in
```
mokagio: Nitpick: The parenthesis should be unnecessary, right?

```suggestion
            accountSettingService.updatePassword(newPassword) { success, error in
```

mokagio (commented):
> The main change is in `AccountService`, removing a few account lookup APIs from it. The rest changes are making the migration, using the APIs in `WPAccount+Lookup` instead.

@jkmassel @crazytonyli what's the motivation for this change? My guess would be better layering / Dependency Inversion?

> Making sure features only available for logged in user are working as expected, and vise-versa. 

What would be a good way to test this? I'm guessing something like creating a self-hosted site and checking how it works in the app, given that would work without a WordPress.com account?

WordPress/Classes/Utility/AccountHelper.swift
crazytonyli: Good point! [The one for Objective-C](https://github.com/wordpress-mobile/WordPress-iOS/blob/93d5bf4cf8f1853cae63a859ba0bf7159fab0070/WordPress/Classes/Models/WPAccount%2BLookup.swift#L121) does exactly this. I'm inclined to make this change, what do you think @jkmassel?

crazytonyli:
Thanks for the review, @mokagio .

@jkmassel can probably talk about the grand plan. But in my understanding, yes, DI would be one of the end goals.

I wasn't thinking about self-hosted sites, but good point, it's definitely worth checking! I was mainly thinking about WP.com features, like posts you see in Reader shouldn't show an "Edit" button if the logged in user doesn't have permission to do so, etc.

mokagio (approved):
@crazytonyli sorry for not following up on this one, so much so that it now has merge conflicts 😞 

The changes are sound as far as I can see. Let's get it into `trunk` so we can add a bit more confidence via the few people who will use installable builds that will include these changes.

crazytonyli:
@mokagio That's okay. I also forgot to push up the commits that address your comments. 🥲

WordPress/Classes/Services/AccountSettingsService.swift
jkmassel: Agreed, though we'd need to decide on a particular type for it – `Int64` is semantically correct, though it's an alias for `Int` in almost every case we'd run this code. 

FWIW I'd vote `Int64`

WordPress/Classes/Services/AccountService.m
jkmassel: The tricky thing is that there can be WP.com and WP.org accounts – `defaultAccount` doesn't really specify which we're looking up 😔

WordPress/Classes/Services/WordPressComSyncService.swift
jkmassel: No strong opinion on it, though the main improvement here is IMHO the fact that we don't need an `AccountService` object to tell us whether a given account is the default WP.com one. 

I'd be fine to rework this – tbh I tend to try to avoid `!thing` as well, but usually via restructuring the surrounding code using `guard` or something.

For now, I'd leave this since we'll be coming back to it anyway in the future, but that's just me

WordPress/Classes/Utility/AccountHelper.swift
jkmassel: So IMHO this is a symptom of poor error handling throughout the app, but is also a step toward the right direction though it feels weird because it's half-done.

My thought – anything that does IO (or depends on something that does IO, but isn't UI) should `throw`. That's so that we don't do things like "return 0 if there's an error" – that swallows the error, so a user might wind up with an empty list instead of an error message like "Network Error – Unable to Connect" because the error was thrown away before it reached the UI layer. 

The reason the `try? [...] ?? [default]` pattern is here is because the parent method isn't marked `throws` – this can be for a few reasons, including Obj-C interop, and a desire to keep small changesets.

IMHO though, this is a step in the right direction, because a search for `try?` in the app will reveal cases where we need to refactor for error handling. 

Anyway, happy to hear y'alls thoughts!

WordPress/Classes/ViewRelated/Me/App Settings/PrivacySettingsViewController.swift
jkmassel: I think just to keep the linter happy?

WordPress/Classes/ViewRelated/NUX/SignupEpilogueViewController.swift
jkmassel: IMHO this comes back to the stuff above about error handling – if the APIs we rely on can't emit errors, we end up with bad UX like this

jkmassel:
@crazytonyli I'm ok to merge this, though we'll need to sort out a merge conflict again (sorry!)

Because my name is on the PR I don't think I can approve it, but feel free to merge when it's clean.

PR #16170, opened 2021-03-26
Apply Default Values to the CoreData items added in 117

chipsnyder:
Fixes a migration issue that was a result of changes made when adding some values to the 117 Data Model.

## To test:
1. Install a previous build such as one from `trunk`
2. Navigate to the page creation screen where you can pick starter content
    - This will populate entities that will be migrated in the next step
3. Install a build from this branch
4. **Expect** the Migration to be successful
5. Navigate to the page creation screen where you can pick starter content
6. **Expect** to not have any issues loading that page

## PR submission checklist:

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -3,6 +3,24 @@
 This file documents changes in the data model. Please explain any changes to the
 data model as well as any custom migrations.
 
+## WordPress 118
+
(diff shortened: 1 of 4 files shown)


mkevins (approved):
I tested this by the steps described, and it's working without a migration failure now. Thank you for tackling this with the version bump Chip!

PR #16177, opened 2021-03-29
Add Notifications screen for unified prologue carousel

frosty:
Ref #15056. This PR adds the Notifications screen to the prologue carousel.

| iPhone | iPad |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-03-29 at 10 37 37 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-03-29 at 10 41 30 |

**To test**

* Build, run and log out if necessary
* Verify that the notifications (third) page of the login prologue is consistent with the designs (see issue #15056; see also example screenshot below)

PR submission checklist:


WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -8733,6 +8738,7 @@
 			isa = PBXGroup;
 			children = (
 				3F851414260D0A3300A4B938 /* UnifiedPrologueEditorContentView.swift */,
+				1788106E260E488B00A98BD8 /* UnifiedPrologueNotificationsContentView.swift */,
```
Gio2018, line 8741: we should probably move this file in a separate group or, if we feel that we have too many sub-groups, just remove the topic level (e.g. `Editor`).

Gio2018 (changes requested):
Hey @frosty , as discussed on Slack, there's a crash happening, likely due to `NSAttributedString` not being called from the main thread. Other than that, I left a code comment.

frosty:
@Gio2018 Thanks for catching this! It turns out that the NSAttributedString construction from HTML needs to be run on the main thread, but there's no guarantee that SwiftUI will call it there. I've gone for a simpler approach to avoid having to use this at all. Let me know what you think.

WordPress/WordPress.xcodeproj/project.pbxproj
frosty: Good catch – I'll wait until this is merged to address this as I don't want to distrupt any other PRs.

Gio2018 (approved):
Works great now!

A few minor comments comparing to the designs:

- the "reply" icon should be bigger and moved a bit to the right
- the "star" icon maybe moved just a bit lower and a bit to the right
- the "comments" icon maybe moved a bit to the left

All these are of course super minor and can be safely ignored.

PR #16187, opened 2021-03-30
Unified prologue - Stats screen

Gio2018:
Ref #15056

NOTE: this PR calculates the distance between the prologue title and the content view relative to the actual view height, and removes the `Spacer` above and below the SwiftUI view. As a consequence, if we follow this approach, other prologue pages need to be updated

To test:

- build/run and log out if needed
- Make sure the stats prologue screen look consistent with the designs (#15056). See also the example screenshots below

iPhone | iPad
-- | --
 | 

PR submission checklist:


WordPress/Classes/ViewRelated/NUX/UnifiedProloguePages.swift
```diff
@@ -75,9 +75,14 @@ class UnifiedProloguePageViewController: UIViewController {
     private func configureContentView() {
         contentView.translatesAutoresizingMaskIntoConstraints = false
         view.addSubview(contentView)
-
+        let contentViewHeight: CGFloat
+        if let windowHeight = UIApplication.shared.mainWindow?.frame.height {
```
frosty: I'm not 100% sure about using the main window, as this isn't very future proof if we support multiple scenes / windows in the future? Also if the window size changes (for example rotating an iPad)?

frosty (commented):
This is looking good! If I was going to nitpick against the original designs I would say: 

- The purple face icon could be a little lower and perhaps a little bigger
- The people icon could be a little bigger too
- The chart element could be a little taller (the map may want to be a little less tall to account for this)

But these are all very minor.

- Also, on iPhone 12 Pro this content view appears a little lower than the editor content view – I think because the title runs over 3 lines instead of 2. Again, this might be something we address separately.

Gio2018:
Thanks for reviewing @frosty 

> * The purple face icon could be a little lower and perhaps a little bigger
> * The people icon could be a little bigger too
> * The chart element could be a little taller (the map may want to be a little less tall to account for this)
> 

sizes and positions changed.

> * Also, on iPhone 12 Pro this content view appears a little lower than the editor content view – I think because the title runs over 3 lines instead of 2. Again, this might be something we address separately.

Agreed to fix these in a separate PR.

frosty (approved):
Looks good!

PR #16192, opened 2021-03-30
Add WPAccount Lookup

jkmassel:
This PR works further toward #16168, which does the same sort of work as #15887, but for Accounts.

Because #16168 got so big, I've split this change out of it. This PR contains only the new methods that'll be called once #16168 is merged.

To test:
- Ensure unit tests pass – they cover the full functionality of the changes.

## Regression Notes
- None possible, given that this is entirely new code that doesn't touch anything else

- Not applicable to this PR

- Added automated tests to cover all cases I could think of for the new code.
 
PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/WPAccount+Lookup.swift
+++ b/WordPress/Classes/Models/WPAccount+Lookup.swift
@@ -126,6 +126,7 @@ public extension WPAccount {
     /// An Objective-C wrapper around the `lookupDefaultWordPressComAccount` method.
     ///
     /// Prefer using `lookupDefaultWordPressComAccount` directly
+    @available(swift, obsoleted: 1.0)
     @objc(lookupDefaultWordPressComAccountInContext:)
     static func objc_lookupDefaultWordPressComAccount(in context: NSManagedObjectContext) -> WPAccount? {
         return try? lookupDefaultWordPressComAccount(in: context)
@@ -134,6 +135,7 @@ public extension WPAccount {
     /// An Objective-C wrapper around the `lookupHasDefaultWordPressComAccount` method.
     ///
     /// Prefer using `lookupHasDefaultWordPressComAccount` directly
+    @available(swift, obsoleted: 1.0)
     @objc(lookupHasDefaultWordPressComAccountInContext:)
     static func objc_lookupHasDefaultWordPressComAccount(in context: NSManagedObjectContext) -> Bool {
         return (try? lookupHasDefaultWordPressComAccount(in: context)) ?? false
@@ -142,6 +144,7 @@ public extension WPAccount {
     /// An Objective-C wrapper around the `lookupDefaultWordPressComAccount` method.
     ///
     /// Prefer using `lookupDefaultWordPressComAccount` directly
+    @available(swift, obsoleted: 1.0)
     @objc(lookupNumberOfAccountsInContext:)
     static func objc_lookupNumberOfAccounts(in context: NSManagedObjectContext) -> Int {
         return (try? lookupNumberOfAccounts(in: context)) ?? 0
@@ -150,6 +153,7 @@ public extension WPAccount {
     /// An Objective-C wrapper around the `lookup(withUsername:context:)` method.
     ///
     /// Prefer using `lookup(withUsername:context:)` directly
+    @available(swift, obsoleted: 1.0)
     @objc(lookupWithUsername:context:)
     static func objc_lookupWithUsername(username: String, context: NSManagedObjectContext) -> WPAccount? {
         return try? lookup(withUsername: username, in: context)
--- a/WordPress/Classes/Models/WPAccount+Lookup.swift
+++ b/WordPress/Classes/Models/WPAccount+Lookup.swift
@@ -29,7 +29,7 @@ public extension WPAccount {
     /// The default WordPress.com account is the one used for Reader and Notifications.
     ///
     static func lookupDefaultWordPressComAccount(in context: NSManagedObjectContext) throws -> WPAccount? {
-        guard let uuid = UserSettings.defaultDotComUUID, uuid.count > 0 else {
+        guard let uuid = UserSettings.defaultDotComUUID, !uuid.isEmpty else {
             // No account, or no default account set. Clear the defaults key.
             UserSettings.defaultDotComUUID = nil
             return nil
--- a/WordPress/Classes/Models/WPAccount+Lookup.swift
+++ b/WordPress/Classes/Models/WPAccount+Lookup.swift
@@ -8,7 +8,7 @@ public extension WPAccount {
     ///
     @objc
     var isDefaultWordPressComAccount: Bool {
-        guard let uuid = UserSettings.defaultDotComUUID, uuid.count > 0 else {
+        guard let uuid = UserSettings.defaultDotComUUID else {
             return false
         }
 
--- a/WordPress/Classes/Models/WPAccount+Lookup.swift
+++ b/WordPress/Classes/Models/WPAccount+Lookup.swift
@@ -71,7 +71,6 @@ public extension WPAccount {
     static func lookup(withUsername username: String, in context: NSManagedObjectContext) throws -> WPAccount? {
         let fetchRequest = NSFetchRequest<Self>(entityName: WPAccount.entityName())
         fetchRequest.predicate = NSPredicate(format: "username = [c] %@ || email = [c] %@", username, username)
-        fetchRequest.includesPendingChanges = true // TODO: this shouldn't be necessary, it was copied from the `AccountService`
 
         guard let account = try context.fetch(fetchRequest).first else {
             return nil
@@ -90,7 +89,6 @@ public extension WPAccount {
     static func lookup(withUserID userID: Int64, in context: NSManagedObjectContext) throws -> WPAccount? {
         let fetchRequest = NSFetchRequest<Self>(entityName: WPAccount.entityName())
         fetchRequest.predicate = NSPredicate(format: "userID = %ld", userID)
-        fetchRequest.includesPendingChanges = true // TODO: this shouldn't be necessary, it was copied from the `AccountService`
 
         guard let account = try context.fetch(fetchRequest).first else {
             return nil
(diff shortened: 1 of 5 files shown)


WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        return try lookup(withUUIDString: uuid, in: context)
+    }
+
+    /// Does a default WordPress.com account exist?
+    ///
+    /// The default WordPress.com account is the one used for Reader and Notifications.
+    ///
+    static func lookupHasDefaultWordPressComAccount(in context: NSManagedObjectContext) throws -> Bool {
```
aerych: Could maybe just be `hasDefaultWordPressComAccount`?   It might just be me but when prefixed with `lookup` I could see this mistaken for the actual look up method at a glance. `lookup` also makes me thing an account instance could be returned where `has` implies a boolean.  Thoughts?

WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    static func lookup(withUUIDString uuidString: String, in context: NSManagedObjectContext) throws -> WPAccount? {
+        let fetchRequest = NSFetchRequest<Self>(entityName: WPAccount.entityName())
+        fetchRequest.predicate = NSPredicate(format: "uuid = %@", uuidString)
+
+        guard let defaultAccount = try context.fetch(fetchRequest).first else {
+            return nil
+        }
+
+        /// This was brought over from the `AccountService`, but can (and probably should) be moved to an accessor for the property
```
aerych, line 56: Looks like it was introduced in https://github.com/wordpress-mobile/WordPress-iOS/commit/2110d392e4eee99e9f2608e7e892b8e510d1bee0.  Not sure if a property accessor or handling when consuming API results would be better, but I'd lean to addressing it in the service or remote service layer. 🤔

WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// - Parameters:
+    ///   - username: The username associated with the account
+    ///   - context: An NSManagedObjectContext containing the `WPAccount` object with the given `username`.
+    /// - Returns: The `WPAccount` object associated with the given `username`, if it exists.
+    ///
+    static func lookup(withUsername username: String, in context: NSManagedObjectContext) throws -> WPAccount? {
+        let fetchRequest = NSFetchRequest<Self>(entityName: WPAccount.entityName())
+        fetchRequest.predicate = NSPredicate(format: "username = [c] %@ || email = [c] %@", username, username)
+        fetchRequest.includesPendingChanges = true // TODO: this shouldn't be necessary, it was copied from the `AccountService`
```
aerych: Per the [docs](https://developer.apple.com/documentation/coredata/nsfetchrequest/1506724-includespendingchanges) the default is `true`.  Maybe just nuke it rather than add a comment to remind us to nuke it later? :)

WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+    ///
+    /// - Parameters:
+    ///   - userID: The userID associated with the account
+    ///   - context: An NSManagedObjectContext containing the `WPAccount` object with the given `userID`.
+    /// - Returns: The `WPAccount` object associated with the given `userID`, if it exists.
+    ///
+    static func lookup(withUserID userID: Int64, in context: NSManagedObjectContext) throws -> WPAccount? {
+        let fetchRequest = NSFetchRequest<Self>(entityName: WPAccount.entityName())
+        fetchRequest.predicate = NSPredicate(format: "userID = %ld", userID)
+        fetchRequest.includesPendingChanges = true // TODO: this shouldn't be necessary, it was copied from the `AccountService`
```
aerych: Same here :)

WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+        let fetchRequest = NSFetchRequest<Self>(entityName: WPAccount.entityName())
+        fetchRequest.includesSubentities = false
+        return try context.count(for: fetchRequest)
+    }
+
+    // MARK: - Objective-C Compatibility Wrappers
+
+    /// An Objective-C wrapper around the `lookupDefaultWordPressComAccount` method.
+    ///
+    /// Prefer using `lookupDefaultWordPressComAccount` directly
```
aerych, line 118: I haven't tried this, but I wonder if `@available` could be leveraged her to mark the objc wrappers as deprecated in swift but available to objc. I'm not sure of the behavior when called from objc code. 🤔

aerych (approved):
Hola @jkmassel 👋 
Shiny :) Everything looked good to me, I just had a few nitpicky suggestions. Up to you sir.
Unit tests pass.
:shipit: from me.

WordPress/Classes/Models/WPAccount+Lookup.swift
frosty: We do this kind of thing in `UIColor+MurielColorsObjC.swift`:

```
@available(swift, obsoleted: 1.0)
```

WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+
+public extension WPAccount {
+
+    // MARK: - Relationship Lookups
+
+    /// Is this `WPAccount` object the default WordPress.com account?
+    ///
+    @objc
+    var isDefaultWordPressComAccount: Bool {
+        guard let uuid = UserSettings.defaultDotComUUID, uuid.count > 0 else {
```
frosty: I wonder if `!uuid.isEmpty` (or `uuid.isEmpty == false`) would be slightly clearer here than checking the count? It makes it more obvious that we're dealing with a string, perhaps (same for the method below too).

WordPress/Classes/Models/WPAccount+Lookup.swift
frosty: This makes sense to me, and helps distinguish the two methods. Alternatively, `defaultWordPressComAccountExists(in:)`?

WordPress/Classes/Models/WPAccount+Lookup.swift
```diff
@@ -0,0 +1,157 @@
...
+    // MARK: - Object Lookups
+
+    /// Returns the default WordPress.com account, if one exists
+    ///
+    /// The default WordPress.com account is the one used for Reader and Notifications.
+    ///
+    static func lookupDefaultWordPressComAccount(in context: NSManagedObjectContext) throws -> WPAccount? {
+        guard let uuid = UserSettings.defaultDotComUUID, uuid.count > 0 else {
+            // No account, or no default account set. Clear the defaults key.
+            UserSettings.defaultDotComUUID = nil
```
frosty, line 34: If the guard fails, presumably this will already be nil? (Or I guess, it could be an empty string)

frosty (approved):
Looking good! I think I agree with all of @aerych's comments, and left one or two of my own.

WordPress/Classes/Models/WPAccount+Lookup.swift
aerych: Ya know... when I was reviewing I did a full stop to check that `count` was valid syntax there.  I'm still so used to strings having a length and arrays having a count 😆.

frosty: Right, I stopped to double check what type `uuid` was 🙂

WordPress/Classes/Models/WPAccount+Lookup.swift
jkmassel: Oh this is a lovely solution! Added in 9b2fbf564046d6829ef26c720b2217b468e53341, thanks for the tip!

WordPress/Classes/Models/WPAccount+Lookup.swift
jkmassel: Yeah, this is an awkward one for sure. It only exists as syntactic sugar to reduce `lookupDefaultWordPressComAccount(in: context) == nil` usage (which only needed to be used 3 times).

Given that you both agree on this (and the fact that's it's kinda awkward anyway, because IMHO Core Data isn't the source of truth for this – it's the `AccountDefaultDotcomUUID` key in `NSUserDefaults`) I'll just remove it to prevent any more confusion or additional ways to do the same thing.

Thanks for flagging this one :) 

Removed in 694b5f36fe301ce129d84a3a7bc7d4e7bc7e9a9c

WordPress/Classes/Models/WPAccount+Lookup.swift
jkmassel: Nice ok! Yeah, I wasn't sure if there was any particular reason why its as there which was why I left the note, but definitely safe to remove here!

Removed in 561ccce352bdeeea3a6865b71955231f299f61ea

WordPress/Classes/Models/WPAccount+Lookup.swift
jkmassel: This doesn't actually need to exist anymore – I mentioned the Objective-C bridging bit in the next comment, but in this case `self.uuid == ""` would return `false` anyway, so no reason to have a special-case check for it.

Removed in 71b8d2b4

WordPress/Classes/Models/WPAccount+Lookup.swift
jkmassel: Ah yeah, we can clean that up now. It was a holdover from Obj-C bridging because of:

```objc
NSString *string = nil;
XCTAssertFalse([string isEmpty]);
XCTAssertTrue([string length] == 0);
```

The Swift-based property wrapper will deal with that case for us now, so we just need to handle the empty case, and !isEmpty is much more clear there for sure 🙂

Adjusted in f6ce941

PR #16201, opened 2021-03-31
Jetpack: include unified prologue files to Jetpack target

leandroalonso:
This PR simples add Unified Prologue files to the Jetpack target.

### To test

1. Make sure the Jetpack target builds

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/ContentViews/Editor/UnifiedPrologueNotificationsContentView.swift
+++ b/WordPress/Classes/ViewRelated/NUX/ContentViews/Editor/UnifiedPrologueNotificationsContentView.swift
@@ -19,7 +19,7 @@ struct UnifiedPrologueNotificationsContentView: View {
                 RoundRectangleView {
                     HStack {
                         NotificationIcon(image: Appearance.topImage, size: notificationIconSize)
-                        Text(string: Appearance.topElementTitle)
+                        Text(Appearance.topElementTitle)
                             .font(notificationFont)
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/NUX/ContentViews/Editor/UnifiedPrologueNotificationsContentView.swift
```diff
@@ -19,7 +19,7 @@ struct UnifiedPrologueNotificationsContentView: View {
                 RoundRectangleView {
                     HStack {
                         NotificationIcon(image: Appearance.topImage, size: notificationIconSize)
-                        Text(string: Appearance.topElementTitle)
+                        Text(Appearance.topElementTitle)
```
leandroalonso: I was having build errors when trying to build that from the Jetpack target.

I removed the `string: ` and it seems to work from both WordPress and Jetpack. But perhaps I missed something, so it would be good if you can double-check @Gio2018 

Thanks in advance!

frosty: Ah, I think we need to add Text+BoldSubString.swift to the Jetpack target (and then re-add this init parameter!) 🙂

leandroalonso: Done!

Gio2018 (approved):
:shipit: !

PR #16215, opened 2021-04-01
Feature/16183 unreplied comments filter

### LINKED ISSUE #16183: Comments: Add an option to filter unreplied comments
Add a new filter to the My Sites > Comments feature that shows recent comments that the user has not replied to.

aerych:
Closes #16183 

This PR wires up the unreplied comments filter, which shows the most recent top level comments from other users that the user has not yet replied to.

When selected, the most recent batch of published and pending comments are fetched.  The results are then filtered to remove comment threads that the user has replied to, remove child comments from other users, and remove top level comments from the user.  The remaining comments are top level comments from other users. 

Simulator Screen Shot - iPhone 11 Pro - 2021-04-01 at 12 20 19

To test:
- Smoke test the other filters and confirm they continue to work correctly.  Confirm you can still load more comments.
- Select the unreplied comments filter.  Ensure you do not see any comments from yourself.
- Confirm top level pending comments show up in the unreplied comments filter.
- Confirm top level comments from other users show up in the comments filter.  Confirm that any comment where you have commented in its thread do not appear.

## Regression Notes
Confirm that other comment filters continue to work as expected.

PR submission checklist:


RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 * [*] Reordered categories in page layout picker [#16156]
 * [*] Added preview device mode selector in the page layout previews [#16141]
 * [**] Block Editor: Added Contact Info block to sites on WPcom or with Jetpack version >= 8.5.
+* [*] Comments can be filtered to show the most recent unreplied comments from other users.
```
ScoutHarris:
```suggestion
* [*] Comments can be filtered to show the most recent unreplied comments from other users. [#16215]
```

ScoutHarris:
Hey @aerych . Replying to a Comment doesn't behave as expected, and causes a crash.

- Go to any filter (I tried from Unreplied and Pending).
- Select a comment and reply.
- The comment appears in `All` and `Approved` correctly.
- However, the reply appears in _all_ filters incorrectly.
  -  It's at the bottom of every list.
  - It has `Anonymous` as the user.
  - When tapped, the app crashes in `CommentAnalytics`.

| All (gooood) | Unreplied (baaaaaad) |
|--------|-------|
| Screen Shot 2021-04-01 at 12 02 58 PM | Screen Shot 2021-04-01 at 12 03 42 PM |

Screen Shot 2021-04-01 at 12 08 21 PM


I restarted the app, but the broken comment still appears. Pull to refresh doesn't remove it either. I reinstalled the app and it finally went away.

jkmassel:
👋 We're cutting the `17.1` release today, so this PR's milestone will be updated to `17.2`!

aerych:
Howdy @ScoutHarris 👋 

I was never able to reproduce the crash but I've added a nil check in analytics to catch it just in case. 🤞 I've tried on a device and simulator.  Curious if you can still reproduce and what other steps / debug info I might try to reproduce.

Regarding that anonymous reply that was showing up in all filters:
This seems to be an existing issue, one that's been around for a long time.  When replying the app inserts a "draft" comment into core data before the reply is synced to the server.  Normally this gets cleaned up but if there's a network hiccup it can be orphaned.  I'm guessing this is what you saw.  I could artificially create this state by short circuiting the upload of a comment reply.  
I'm not sure what's best here.  On the one hand, I want to filter these out so they are not visible to the user.  On the other, if there is an error the user could salvage their comment and try again tho it's not a great experience.  As a compromise I've left the local reply visible under the ALL filter but hidden from the other filters. I'd not be opposed to hide the local reply from the All filter as well but wanted to get your thoughts. 

Finally I've added updates to address self-hosted sites, which I'd overlooked before. Props to Klym for mentioning this.  To do this I needed to make an [update to WordPressKit-iOS](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/387) which can be tested with this PR.

I think we're (finally) ready for another round.

WordPress/Classes/Services/CommentService.m
```diff
@@ -207,6 +229,56 @@ - (void)syncCommentsForBlog:(Blog *)blog
     }];
 }
 
+- (NSArray *)filterUnrepliedComments:(NSArray *)comments forAuthor:(NSString *)author {
+    NSMutableArray *marr = [comments mutableCopy];
+
+    NSMutableArray *foundIDs = [NSMutableArray array];
+    NSMutableArray *discardables = [NSMutableArray array];
+
+    // get ids of comments that user has replied to.
```
ScoutHarris, line 238: Thank you for commenting each step. 👍

ScoutHarris (approved):
> I've added a nil check in analytics to catch it just in case.

Awesome! Thank you!

> Normally this gets cleaned up but if there's a network hiccup it can be orphaned. I'm guessing this is what you saw. 

Wicked. I repro'd it consistently. I guess that was a bad day for my internet. And of course I can't repro it now; my replies appear in `All` and `Approved` correctly. Thanks for the explanation!

> As a compromise I've left the local reply visible under the ALL filter but hidden from the other filters.

I think that's reasonable. Before filtering, it seems these would have shown in the flat Comments list anyway, which would now equate to `All`. Makes sense.

I've retested and all works as expected.

:shipit:

aerych:
Thank you @ScoutHarris !

PR #16217, opened 2021-04-01
Unified prologue - add background circles

Gio2018:
REF #15056 

This PR adds circles in the background of the unified prologue screens. In this implementation we are using a single background for all the prologue screens, as opposed to the initial designs that had different backgrounds for each screen. Thus there is no exact match between the circles in the designs (#15056) and the actual circles in the background, both in terms of colors (which here use our color palette) and position, but it should be close enough.

Here are some examples

light appearance | dark appearance
-- | --
| 

To test:

- build, run and logout if needed
- test on a variety of devices, and make sure the circles look consistent with the examples above

## Regression Notes
This new view with circles only applies to the prologue. 

It's still a good idea to test on as many devices as possible to make sure the view looks good. I tested on a variety of simulators on both iOS 13 and iOS 14

NA

PR submission checklist:


frosty (approved):
While there's not an exact match on the designs, I tested on a range of devices and it all seems to be working well for me!

PR #16235, opened 2021-04-05
Changes

### LINKED ISSUE #15925: Accessibility: can't create a new site/choose a domain if iOS is set to a larger text size
When you create a new site from the app, we give the option to start with WordPress.com. You first choose a layout and then you move to the next screen, called "Choose a Domain". Users with larger text sizes will be stuck there.

Similar to #15655 but this one is apparently fixed.

### Expected behavior

Users should be able to:

- See the field **Search Domains** in order to type it
- See the results of the domain they search for

### Actual behavior

Depending on the text size, you'll see the **Search Domains** field and can type a domain. However, you can't see the results | On even larger text sizes, you can't even see the **Search Domains** text field.
------------ | -------------
 | 

In both cases, it's impossible to scroll down.

1. On your phone, go to System Settings > Accessibility > Display & Text Size
2. Make sure that **Larger Accessibility Sizes** is checked
3. Increase the font size 
4. Try to scroll down to either see the search results or to simply find the **Search Domains** field

##### Tested on iPhone 7, iOS 14.4, WPiOS 16.6.1

Report: 3746854-zen

kduncanwelke:
Fixes #15925

To test: In old code, set device text side under accessibility to use large text (use a high setting). Open the app and try to add a new site - with large text, the search field for a url is covered by the keyboard.

In this edited code, the table has had an offset applied, which pushes the search field up above the keyboard, making it visible. This code will apply if the table exceeds the height of the device, like on small devices or if the text in the site creation header is very large. It will be ignored on iPads/bigger devices.

## Regression Notes
The code could have applied erroneously when not needed, or caused an increasing offset.

Tested on different devices sizes to make sure this did not happen, and tested going back and forth between views and hiding and showing the keyboard.

None, as the success of the fix depended very much on the visual, as confirmed by manual testing and eyeballing.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Site Creation/Shared/TableViewOffsetCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Site Creation/Shared/TableViewOffsetCoordinator.swift
@@ -27,7 +27,7 @@ final class TableViewOffsetCoordinator {
 
     /// Tracks the content offset introduced by the keyboard being presented
     private var keyboardContentOffset = CGFloat(0)
-    
+
     /// Track scroll position of tableview for reloads
     private var scrollPosition = CGPoint(x: 0, y: 0)
 
@@ -61,17 +61,17 @@ final class TableViewOffsetCoordinator {
         guard let tableView = tableView, keyboardContentOffset > 0, tableViewHasBeenAdjusted == false else {
             return
         }
-        
+
         // if table is taller than screen, apply an inset to make the search field at the bottom visible
         if UIScreen.main.bounds.height < tableView.bounds.height {
             // no change
         } else {
             tableView.contentInset.bottom = keyboardContentOffset*1.5
         }
-        
+
         tableViewHasBeenAdjusted = true
         scrollPosition = tableView.contentOffset
-        
+
         // removed insets that had been here, as they made content disappear with large text sizes
         // replaced with contentoffset and scroll of tableview
     }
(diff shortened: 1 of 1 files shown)


chipsnyder:
Hi @kduncanwelke, 

I just came across this PR for https://github.com/wordpress-mobile/WordPress-iOS/issues/15925. Thank you so much for proposing a fix. I'll run this fix today and leave some code review notes. 

I see that our automated bot called out a few style issues. You should be able to correct all of these with `rake lint:autocorrect`

chipsnyder:
Hey @kduncanwelke 👋 

Thanks again for taking a look at this issue. I was wondering if you could help me connect the pieces. I wasn't able to locate where this class was being used on the screen we're targeting. It seems as if it might not be referenced but perhaps there is something I'm missing?

kduncanwelke:
Heyo @chipsnyder thank you for looking into this! It has been some months since I touched this issue, but from what I recall it took some time to figure out that this class was responsible. I am cloning the project (new machine) and will fix the style issues and get back to you as soon as possible on how this fix connects with the issue.

kduncanwelke:
Hi again @chipsnyder - I have fixed the whitespacing issue in the repo, so hound should like it now. 

I am unfortunately unable to get to the bottom of where TableViewOffsetCoordinator is implemented, as I am on an M1 computer, which is not compatible with the project. What I suspect, however, is that it is implemented in one of the included WordPress frameworks, as this was the case with a previous accessibility site creation issue I fixed (the problem was fixed in WordPress Authenticator).

Changes to TableViewOffsetCoordinator do have the desired effect in fixing the problem, so it is responsible, but I'm afraid with the M1 not having support for this project yet, I cannot point out directly where the implementation is. 😓

chipsnyder:
Hi @kduncanwelke Thanks for cleaning up those formatting issues. 

> as I am on an M1 computer, which is not compatible with the project.

We did finally add support for M1 devices so you may just need to update your fork (I'm on a new M1 iMac and I was able to get your branch up and running after merging in the latest `develop`).

I'd be happy to help with any errors you might be having. 

----- 

Unfortunately, these changes don't appear to be working for me 😞 this is what I see when I run your branch updated with `develop`

https://user-images.githubusercontent.com/3384451/126842204-305b02d1-8e5c-4e0a-b430-c5e12908a6ac.mp4

We have been doing a lot of work in this area though so potentially your change was more closely tied to an area that was refactored.

I was focused on some bug fixes this week so I took a stab at this one as well. If you'd like to compare notes here is the path I took https://github.com/wordpress-mobile/WordPress-iOS/pull/16923

kduncanwelke:
@chipsnyder Ah, thank you for the info about the M1 now working with the project! Mine must be rather out of date. 😅

I am going to ahead and close this pr, as the changes aren't performing as expected any more, per the video, and it sounds like you have it in hand already. Thank you for taking the time to check this pr out - it's just not a good fit with where the code has evolved to now. 👍🏻

PR #16240, opened 2021-04-06
Fix Notifications no results view offset (and other issues)

frosty:
This PR fixes a number of issues in the Notifications section of the app, some of which were highlighted or introduced as part of the large titles / white headers changes:

- The 'no results' view in Notifications would be displayed too far down the screen, with a large gap visible between the no results view and the nav bar header
- Tapping the Notifications button when the Notifications view was already visible would scroll to an incorrect offset in the tableview, cutting off the content of some cells. Now it should scroll to the top, keeping the nav bar title small.
- On initial login to a new account with no notifications, the detail pane on iPad was gray. Now it should be white.

**Before**

**After**

|   |   |   |   |
|---|---|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 01 51 |  Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 03 33 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 10 29 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 10 27 |
|   | Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 32 05 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 32 03 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-06 at 16 33 34 |

**To test**

- Build and run
- Create a new account and ensure that the Notifications screen shows a seamless white background with the no results view centered
- If you like, trigger the display of the app rating header prompt – the easiest way is to modify line 184 of NotificationsViewController to say `} else if true {`. When you run, you should see the prompt appear in the header of the table view. When you dismiss it, the gap should be closed.
- Trigger a notification to be added to this account. You could either create a site for the account, create a post, and like it from another account. Or use this account to comment on a test site and then reply to that comment from another account.
- When the notification is added, you should see the filter bar appear in this view. You should see content in some tabs, and a no results view in other tabs (for example, if you replied to this user's comment you may see a notification under Comments, but nothing under Likes).
- Also test with the `newNavAppearance` feature flag disabled and check that everything still looks okay (although this feature has already shipped, so hopefully we won't be rolling back!)

## Regression Notes

There are quite a few cases here with the various headers available (filter tabs and prompts), and the header has changed with the new nav bar appearance.

I tried to ensure that I tested each case – the app rating prompt, the filter tab bar, both, neither, a new account with no notifications, the transition to having a notification, and an existing account with a lot of notifications.

This was really a purely visual change.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -1266,18 +1266,25 @@ private extension NotificationsViewController {
         addChild(noResultsViewController)
         tableView.insertSubview(noResultsViewController.view, belowSubview: tableHeaderView)
         noResultsViewController.view.frame = tableView.frame
-        adjustNoResultsViewSize()
+        setupNoResultsViewConstraints()
         noResultsViewController.didMove(toParent: self)
     }
 
-    func adjustNoResultsViewSize() {
-        noResultsViewController.view.frame.origin.y = tableHeaderView.frame.size.height
-
-        if inlinePromptView.alpha == WPAlphaFull {
-            noResultsViewController.view.frame.size.height -= tableHeaderView.frame.size.height
-        } else {
-            noResultsViewController.view.frame.size.height = tableView.frame.size.height - tableHeaderView.frame.size.height
+    func setupNoResultsViewConstraints() {
+        guard let nrv = noResultsViewController.view else {
+            return
         }
+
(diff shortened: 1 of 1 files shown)


emilylaguna (approved):
- Reproduced the issue on develop
- ✅ Confirmed it was fixed using this PR. 
- Tested with a new WP.com account
- Tested with a Jetpack site via site address login

LGTM. 🚀

PR #16243, opened 2021-04-06
Merge 17.1 code freeze into develop

jkmassel:
Merges the 17.1 code freeze into `develop`


mokagio:
🤔 the tests passed in the release notes commit, but failed after. I can see the was a – seemingly harmless – pods update. Maybe that broke something?

Here's the unit test failure, in case it helps. Looks like a simple copy change.

```
Test Suite WordPressTest.xctest started
PostCompactCellTests
    ✗ testShowDate, XCTAssertEqual failed: ("Optional("now")") is not equal to ("Optional("just now")")


WordPressTest.PostCompactCellTests
  testShowDate, XCTAssertEqual failed: ("Optional("now")") is not equal to ("Optional("just now")")
  /Users/distiller/project/WordPress/WordPressTest/PostCompactCellTests.swift:44

        XCTAssertEqual(postCell.timestampLabel.text, "just now")
```

jkmassel:
Ah thanks for pointing this out @mokagio! Should be resolved now!

WordPress/WordPressTest/PostCompactCellTests.swift
```diff
@@ -41,7 +41,7 @@ class PostCompactCellTests: XCTestCase {
 
         postCell.configure(with: post)
 
-        XCTAssertEqual(postCell.timestampLabel.text, "just now")
+        XCTAssertEqual(postCell.timestampLabel.text, "now")
```
mokagio, line 44: Too easy! 😄 

This brings up an interesting question, one to which I don't have an answer.

- Should this test have used something more down to the implementation level of the library, e.g. asserting it's equal to `valueSetInLibraryForNow`? Or
- It's actually great that we test at a higher level here so if we drop the library we don't need to change the tests (_test behavior not implementation_). Or
- Maybe this is at the wrong abstraction level, only the library should concern itself with testing the type of formatting

Again, I don't have an answer, I don't think there's a _right_ answer anyways, just a set of tradeoffs.

mokagio (approved):
:shipit: 

I saw you enabled auto-merge. This is the first time I use it, keen to see how it behaves.

mokagio:
> I saw you enabled auto-merge. This is the first time I use it, keen to see how it behaves.

It behaved like this:

Screen Shot 2021-04-09 at 9 45 25 am

I actually find it confusing... I can see how the author should be @jkmassel as you authorized the auto-merge, but this doesn't tell me whether Jeremy the person did it or it was the GitHub automation acting on Jeremy's behalf.

🤷‍♂️ I'm probably overthinking this...

WordPress/WordPressTest/PostCompactCellTests.swift
jkmassel: > Maybe this is at the wrong abstraction level, only the library should concern itself with testing the type of formatting

Yeah, this is more my preference – option 1 IMHO just test that the code does what the code does. Option 2 is basically what this change is, which is an ok balance of ensuring that we're still testing the text behaviour 🤷. Seems the most in the middle

jkmassel:
> Jeremy the person did it or it was the GitHub automation acting on Jeremy's behalf

Finally, I can merge PRs in the middle of the night and still get a good night's sleep 😂

PR #16258, opened 2021-04-08
Jetpack App: Update the login flow

emilylaguna:
### Summary
This PR does the following:
- If the user enters a Jetpack site address and it is linked to a WordPress.com account, we redirect them to login to their WordPress.com account
- If the user enters a Jetpack site that is not linked, we display an error
- If the user logs into a WordPress.com account that does not contain any Jetpack sites we display an error
- Disables the 'enableSignUp' flags in the WPAuthenticator options

**Demo videos are at the bottom of this description.** 

### Notes / Known Issues

- The error view and model are placeholders until we have designs for this view. 
- I decided to include the new files in both schemes to prevent having to use a bunch of `#if JETPACK` checks, and to use the `AppConfiguration`

**WPAuthenticator `enableSignUp` issues:**
- For some reason the user can still sign up by entering an email
- The log button still reads "Login and sign up using WordPress.com"
- I'll fix these in a later PR. 

### To test:

There's a lot of testing scenarios. 

#### WordPress.com account with Jetpack sites
1. Tap the 'Login with WordPress' button
2. Enter WordPress.com credentials for an account that has at least 1 Jetpack site linked
3. After logging in you should be brought to the login epilogue (the view that displays your sites)

#### WordPress.com account with NO Jetpack sites
1. Tap the 'Login with WordPress' button
2. Enter WordPress.com credentials for an account that is not linked to any Jetpack sites
3. After logging in you should be brought to the error view with the title being 'No Jetpack Sites'

#### WordPress.com account with NO Jetpack sites Part 2
1. After seeing the error from the step above, force quit the application, and restart it
2. You should be brought to the login prologue view again

#### Enter Site Address: Jetpack site that's connected to WordPress.com
1. Tap the 'Enter Site Address' button
2. Enter Jetpack site URL that is linked to a WordPress.com account, look here for examples: http://do.wpmt.co/
3. You should be brought to the WordPress.com email login view

#### Enter Site Address: Jetpack site that's **NOT** connected to WordPress.com
1. Tap the 'Enter Site Address' button
2. Enter `http://do.wpmt.co/jp-disconnected/` or another site that has Jetpack installed but not connected
3. You should be brought to an error view with the message "Jetpack Not Found"

#### Enter Site Address: Site that has Jetpack installed but not activated
1. Tap the 'Enter Site Address' button
2. Enter `https://emily.tools/jetpack-installed-but-not-active`
3. You should be brought to an error view with the message "Jetpack Not Found"

### Regression Notes
The WordPress login flow.

I tested using the same accounts on the WordPress scheme to verify the login flow still works correctly.

None.

### PR submission checklist:

#### Enter Site Address with a site that is not connected to a WordPress.com account
https://user-images.githubusercontent.com/793774/114060431-5dee1380-9863-11eb-801f-4b107a4fe04c.mov

#### Enter Site Address with a site that is connected to a WordPress.com account
https://user-images.githubusercontent.com/793774/114060469-66dee500-9863-11eb-860e-fa339ac318de.mov|

#### WordPress.com login that has no Jetpack sites
https://user-images.githubusercontent.com/793774/114060741-af969e00-9863-11eb-9dfb-8501d0a27699.mov

#### WordPress.com login that has Jetpack sites
https://user-images.githubusercontent.com/793774/114060726-aad1ea00-9863-11eb-9618-1ae30b0891bd.mov


WordPress/Jetpack/Classes/JetpackAuthenticationManager.swift
```diff
@@ -0,0 +1,72 @@
...
+        ///
+        DDLogWarn("⚠️ Present password controller for site: \(site.url)")
+        let authenticationResult: WordPressAuthenticatorResult = .presentPasswordController(value: false)
+        onCompletion(authenticationResult)
+    }
+
+    static func presentLoginEpilogue(in navigationController: UINavigationController, for credentials: AuthenticatorCredentials, onDismiss: @escaping () -> Void) -> Bool {
+        guard Self.hasJetpackSites() == false else {
+            return false
+        }
```
momo-ozawa: [nits] What do you think of rewriting it like this?
```suggestion
        if Self.hasJetpackSites() {
            return false
        }
```

momo-ozawa (commented):
Works as described, tested all the flows:
- WordPress.com account with Jetpack sites
- WordPress.com account with NO Jetpack sites
- WordPress.com account with NO Jetpack sites Part 2
- Enter Site Address: Jetpack site that's connected to WordPress.com
- Enter Site Address: Jetpack site that's NOT connected to WordPress.com
- Enter Site Address: Site that has Jetpack installed but not activated

WordPress/Jetpack/AppConfiguration.swift
```diff
@@ -1,6 +1,11 @@
 import Foundation
 
+/**
+ * Jetpack Configuration
+ */
 @objc class AppConfiguration: NSObject {
     @objc static let isJetpack: Bool = true
     @objc static let allowSiteCreation: Bool = false
+    @objc static let jetpackLogin: Bool = true
```
momo-ozawa: I'm wondering if there's a better name for this, but I can't come up with a better solution... 🤔 
I understand it means "login flow for jetpack", but I think it can be kinda confusing if you just see `AppConfiguration.jetpackLogin`

emilylaguna: Same here! I originally had "jetpackLoginFlow" but that was like meh then I just used "isJetpack" and that didn't feel right either, so I just used "jetpackLogin". It's not great, but I can't think of a better name. 

@leandroalonso any thoughts? 🤔

WordPress/Jetpack/Classes/JetpackAuthenticationManager.swift
emilylaguna: Yeah that makes more sense, I've been coding PHP for too long now. haha

leandroalonso:
@emilylaguna it all works as described!

As for the `AppConfiguration.jetpackLogin` I tried to come up with a POC to:

* Avoid the static class
* Remove the `AppConfiguration.jetpackLogin`
* Remove the need to add JP classess into WP

You can see the result here:

https://github.com/wordpress-mobile/WordPress-iOS/commit/22e64bccc650332ecc64a6d4200d4be904086396

The main idea is to use dependency injection. We can inject (or not) the dependency depending on the app. Doing that, you don't need to refer to the new VCs and classes directly and don't need to include them in the WP target.

I also thought about inheriting from `WordPressAuthenticationManager` but that felt a little bit weird and I think composition can be a better friend in this case.

Naming can still be better though. :P 

Let me know what do you both think (cc @momo-ozawa).

momo-ozawa:
@leandroalonso @emilylaguna 

> The main idea is to use dependency injection. We can inject (or not) the dependency depending on the app. Doing that, you don't need to refer to the new VCs and classes directly and don't need to include them in the WP target.

That sounds like a good idea to me. I like that by injecting `AuthenticationHandler` into `WordPressAuthenicationManager`, we won't have to add the new Jetpack VCs/classes to WP target

emilylaguna:
@leandroalonso @momo-ozawa Nice, I like it! Updated the code to use it. May I please get another 👀 ?

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
```diff
@@ -240,13 +244,24 @@ extension WordPressAuthenticationManager: WordPressAuthenticatorDelegate {
...
     ///
     func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (WordPressAuthenticatorResult) -> Void) {
+        if let authenticationHandler = authenticationHandler {
+            authenticationHandler.shouldPresentUsernamePasswordController(for: siteInfo, onCompletion: onCompletion)
+            return
+        }
+
         let result: WordPressAuthenticatorResult = .presentPasswordController(value: true)
         onCompletion(result)
+        return
```
leandroalonso: [nit] No need for this `return`

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
```diff
@@ -10,8 +10,11 @@ class WordPressAuthenticationManager: NSObject {
     static var isPresentingSignIn = false
     private let windowManager: WindowManager
 
-    init(windowManager: WindowManager) {
+    private let authenticationHandler: AuthenticationHandler?
```
leandroalonso, line 15: Do you think a comment about what `AuthenticationHandler` is might worth having here?

WordPress/Jetpack/Classes/AuthenticationHandler.swift
```diff
@@ -0,0 +1,8 @@
+import WordPressAuthenticator
+
+protocol AuthenticationHandler {
```
leandroalonso: This is discussable: I think this protocol should be on WP.

If we delete everything under the `Jetpack` group, WP should still build without any issues.

leandroalonso (changes requested):

WordPress/Jetpack/Classes/AuthenticationHandler.swift
emilylaguna: I agree. I debated with myself on where to put it initially.

leandroalonso (approved):

WordPress/Jetpack/Classes/AuthenticationHandler.swift
emilylaguna: Done in: 1a5d8da6e215859b3a5948fbba94bf5ceca957ed

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
emilylaguna: Done in: f5d2ed0cf7a6eac165fed3eecf570e6201ff2769

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
emilylaguna: Done in: 2044272a20c9bff44abb638c7074c488fe957259

PR #16261, opened 2021-04-08
Update custom icons

frosty:
This PR updates the custom icons provided by the app.

| Before | After |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-04-08 at 19 11 37 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-08 at 19 07 48 |

Most of the old icons have now been removed, so they are mapped over to the equivalent new icon when you launch the app:

Screen Shot 2021-04-06 at 12 09 51 PM

The only exception to this is Pride – the existing icons are still bundled with the app (but hidden in the UI) so if you were using that 

**To test**

- Build and run develop, and choose a custom icon from Me > App Settings > App Icon
- Build and run this branch and check that the icon automatically switches to the equivalent, shown in the table above. Check in App Settings and ensure the correct icon is ticked (try this for some other icons too).
- Ensure you can switch between the new icons in the app.
- Try upgrading from the pride icon and ensure you still see that icon on the homescreen after the upgrade (Spectrum will be selected in the UI as it's the closest equivalent, but I can remove this mapping if it's confusing).

## Regression Notes

Nothing functional, just icon changes.

I tested the migration from each old icon type

PR submission checklist:


Gio2018 (approved):
Woks great! Tried several mappings and they worked as expected.

> Spectrum will be selected in the UI as it's the closest equivalent, but I can remove this mapping if it's confusing).

Not sure, I think that also not having an icon selected in the UI could be confusing, maybe.

mattmiklic (changes requested):
This is looking really good so far! I'm waiting on the build to finish to test IRL, but I wanted to suggest two tweaks:

- The icon size is now 57pt I believe, which is good (the larger visual helps with the new gradients). We should increase the corner curve to 10pt at that size, to match the radius of app icons on the home screen.
- This can be considered a bonus, but it looks like on iOS 13 and above, there are ways to use Apple's nicer continuous curves ([ref 1](https://stackoverflow.com/a/56442423), [ref 2](https://developer.apple.com/documentation/apple_news/cornermask)) rather than the basic cornerRadius. This would eliminate those pesky [unresolved tangency breaks](https://www.newyorker.com/magazine/2015/02/23/shape-things-come). :)

mattmiklic:
Looking at the code, actually I see the icons are 60pt, so the radius on those would be 13pt.

frosty:
@mattmiklic I've updated the corner radius and switched to continuous curves (thanks for the tip, I didn't know that was present in the API!). I've also re-introduced the legacy icons and displayed them in the UI. I know we discussed showing a prompt to the user – I'll add that in a separate PR as this one's already a reasonable size.

|   |   |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-04-09 at 15 46 27 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-09 at 15 46 30 |

@Gio2018 Apologies, would you mind re-reviewing (not super urgent)? You should no longer be automatically updated to a new icon – instead, the old 'legacy' icons are now reflected at the bottom of the icon selection screen.

mattmiklic (approved):
This looks great now. I appreciate the adjustments!

Gio2018 (approved):
:shipit: !

frosty:
@mattmiklic FYI I'm going to merge this one as-is, and we can look at different presentations for the icons in a separate PR.

PR #16275, opened 2021-04-12
Fix a translation issue

jkmassel:
Fixes an issue with translations


mokagio (approved):
👍 

How did it happen? Just curious in case there's something I can do to avoid this on my end.

PR #16269, opened 2021-04-09
Set author display name on Post / Page creation, send author changes to API.

twstokes:
This is one PR in a series towards implementing #9015.

This PR is branched from https://github.com/wordpress-mobile/WordPress-iOS/pull/16298, so it should be merged first.

### Affected areas:

This PR aims to accomplish two things:

1. Sending an updated authorID when an author is changed for a Post or Page.
**Why:** To support a future PR (https://github.com/wordpress-mobile/WordPress-iOS/pull/16281) we want the value of the author user ID to be sent to the API. Currently it's not sent. This PR updates the `RemotePost`'s `authorID` value so that it will be picked up by WordPressKit (see PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/381) and ultimately sent to the API. REST API [reference 1](https://developer.wordpress.com/docs/api/1.2/post/sites/%24site/posts/new/), [reference 2](https://developer.wordpress.com/docs/api/1.2/post/sites/%24site/posts/%24post_ID/), XML-RPC API [reference 1](https://codex.wordpress.org/XML-RPC_WordPress_API/Posts) (see `wp.newPost` and `wp.editPost`).

2. Setting an author user ID and display name on a newly created Post or Page.
**Why:** To support a future PR (https://github.com/wordpress-mobile/WordPress-iOS/pull/16281) we want the author display name and author user ID to be known on Post or Page creation to allow changing an author for a multi-user blog. We must know both so that the user interface can show the author's name (by default, theirs), and showing the currently selected author (which is determined by the author ID) in the author list. Currently, neither of these are set.

The ability to change an author in the UI will be provided in an upcoming PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16281

### To test:

**Local changes to the author**

- Run the unit tests `PostsTests`. Manipulating this from the app interface isn't currently possible, but will be in https://github.com/wordpress-mobile/WordPress-iOS/pull/16281.

**Setting the authorID and author name on Post and Page creation**

1. Add breakpoints on `EditPageViewController:56` and `EditPostViewController:127`.
2. Create a new Post or Page using the app.
3. When execution is paused for debugging, observe the value of the Post or Page (`newPost` or `newPage`).
4. Both the `newPost` and `newPage` should have a populated `author` and `authorID` field.

**The authorID is now sent on Post / Page creation and modification**

1. Add a breakpoint on `PostService.m:266`.
2. Create or update a Post or Page.
[...]

Issue #9015, opened 2018-03-31
Post/Page Options: Allow users to change the post/page author


lizkarkoski:
Feature request brought via Twitter to add the "change author" functionality to the iOS app. 

Tweet: https://twitter.com/WegrynENT/status/979936474090082304

version 9.6

rachelmcr:
See also past requests in this issue (from the old editor): https://github.com/wordpress-mobile/WordPress-Editor-iOS/issues/632

It also came up in an app review for version 9.8.

SiobhyB:
This came up in a recent review:

> While this app is a lifesaver and easy to use for the most part, I'm continually frustrated by how many features from the main site are missing. For instance, there doesn't seem to be any way to assign authorship to a blog post in the app.

KokkieH:
Another in https://twitter.com/ChristianBlogR/status/1157245753196236800

thehenrybyrd:
Another request for this in #2871252-zen

thehenrybyrd:
Another in #3825927-zen

twstokes:
I'm game to take it! 👋

twstokes:
Hey @guarani 👋 , I didn't automatically close this with the final PR. Would you mind closing it if it's the appropriate time to do so? Thank you!

guarani:
Thanks for catching this, closing now (noting that https://github.com/wordpress-mobile/WordPress-iOS/pull/16281 is the PR that shipped this feature).

PR #16281, opened 2021-04-12
Feature: Allow the user to change the author for multi-author sites for Posts and Pages

twstokes:
This is one PR in a series towards implementing #9015.

This PR was branched from https://github.com/wordpress-mobile/WordPress-iOS/pull/16349, so it should be merged first.

## Description:

For blogs with more than one author:

- Allows the user to select a new author on Post or Page creation.
- Allows the user to select a different author for an existing Post or Page.

## To test:

**Note:** If the app's logged in user's role is downgraded, leftover state (e.g. previously downloaded users) may remain, resulting in failures of the tests below. It's recommended to test user role changes by going from lowest up, or by wiping the app state clean each time. This should be considered a bug.

### Self-hosted
There is a discrepancy for the Editor role. On web, an Editor can choose new authors. On WPiOS, the client cannot query for the list of users, and thus can't populate the authors, so the user will not have the Author option. In other words, for self-hosted sites the user role must be Administrator to use this feature.

- [ ] A site that has multiple users (all Contributor role or higher) gives Administrator roles the option to change the author.
- [ ] Roles lower than Administrator do not see the option to change an author, even if there are multiple author candidates.
- [ ] Contributor roles do not show as an author option and do not add to the author count.
- [ ] Creating a new Post and setting the author. (see Creating steps)
- [ ] Updating an existing Post and changing the author. (see Updating steps)
- [ ] Deleting authors. (see Deleting steps)

### WordPress.com
Administrator and Editor roles should be able to use this feature.

- [ ] A site that has multiple users gives Administrator and Editor roles the options to change the author.
- [ ] Creating a new Post and setting the author. (see Creating steps)
- [ ] Updating an existing Post and changing the author. (see Updating steps)
- [ ] Deleting authors. (see Deleting steps)

---

**Creating steps:**

1. Create a new Post or Page on a site that has multiple users.
2. Access the Post or Page settings by tapping the horizontal ellipses (More Options) icon at the top right and choosing Settings.
3. Observe that "Author" appears as the top row under the "Publish" section.
4. Tap on the row and observe that authors are shown by their Display Name sorted alphabetically. 
5. Select a new author and publish or save the Post or Page.
6. Observe in the app and on the web that the new Post or Page has the chosen author.

**Updating steps:**

1. Select an existing Post or Page on a site that has multiple users.
2. Access the Post or Page settings by tapping the horizontal ellipses (More Options) icon at the top right and choosing Settings.
3. Observe that "Author" appears as the top row under the "Publish" section.
4. Tap on the row and observe that authors are shown by their Display Name sorted alphabetically.
5. Select a new author and update the Post or Page.
6. Observe in the app and on the web that the new Post or Page has the chosen author.

**Deleting steps:**

1. Add three users to a site.
2. Select an existing Post or Page.
3. Access the Post or Page settings by tapping the horizontal ellipses (More Options) icon at the top right and choosing Settings.
4. Observe that "Author" appears as the top row under the "Publish" section.
5. Tap this row and observe that all three users are shown as a selectable option.
6. Using the WordPress Dashboard, remove one of the users from the site.
7. Restart WPiOS or return to the "My Site" view to refresh the blog settings.
8. Repeat steps 2-5. Observe that only the two remaining users are shown.
9. Repeat steps 6 and 7.
10. Repeat steps 2-3. Observe that there is no "Author" top row in the "Publish" section.

## Regression Notes

- Other items in the "Publish" section of the Post / Page Settings view.

- I relied on existing automated tests and I completed all of the manual tests under the section above entitled "To Test".

- Automated tests that were added in #16349, #16269 handle syncing author state and testing that our UI shows the "Update" button when an author is changed. Moving to a view model pattern in the future could further help test UI expectations for a given state.

PR submission checklist:


WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -448,6 +450,9 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
 
     } else if (sec == PostSettingsSectionStickyPost) {
         return 1;
+
+    } else if (sec == PostSettingsSectionAuthor && self.apost.author) {
```
twstokes, line 454: Just realized there's a consequence of checking for an author here. On new Pages / Posts one isn't set. If this is removed the user can successfully set an author on creation, but the UI needs a little more polish. It'll probably be best to set the post's author beforehand.

twstokes:
~I need to add this one to `RELEASE-NOTES`, putting a reminder here.~ Done!

jkmassel:
This PR is being moved to the `17.3` milestone, as we're cutting the `17.2` release branch today 🙂

RELEASE-NOTES.txt
```diff
@@ -5,6 +5,7 @@
 * [*] Comments can be filtered to show the most recent unreplied comments from other users. [#16215]
 
 * [*] Fixed the navigation bar color in dark mode. [#16348]
+* [**] A new author can be chosen for Posts and Pages on multi-author sites. [#16269, #16281, #16349]
```
guarani:
```suggestion
* [**] A new author can be chosen for Posts and Pages on multi-author sites. [#16281]
```
The PR link will mainly be used by beta testers testing the feature, so just referencing this PR should provide the needed testing steps for them.

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
```diff
@@ -218,6 +219,8 @@ import Foundation
             return "editor_post_featured_image_changed"
         case .editorPostStickyChanged:
             return "editor_post_sticky_changed"
+        case .editorPostAuthorChanged:
+            return "editor_post_author_changed"
```
guarani, line 223: Thanks for adding this!

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        return activeAuthors.compactMap {
+            /// Require a display name to be available.
+            guard let displayName = $0.displayName else {
+                return nil
+            }
+
+            return (displayName, $0.userID, $0.avatarURL)
+        }.sorted(by: { $0.displayName.lowercased() < $1.displayName.lowercased() })
```
guarani: If there are three authors: Sérgio, Seth, Sam, ideally they should be sorted Sam, Sérgio, Seth.
Does this code handle this?

twstokes: Great catch! I didn't consider that! 

The accented name Sérgio will currently come after "Seth" because 'e' precedes 'é'. I'll get it fixed. Thanks!

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override init!(style: UITableView.Style, andDictionary dictionary: [AnyHashable: Any]!) {
+        super.init(style: style, andDictionary: dictionary)
+    }
+
+    override init(style: UITableView.Style) {
+        super.init(style: style)
+    }
```
guarani, line 62: Are these overrides needed?

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
```diff
@@ -0,0 +1,82 @@
+import UIKit
+
+@objc class PostAuthorSelectorViewController: SettingsSelectionViewController {
+    /// The post to change the author.
+    private var post: AbstractPost!
```
guarani, line 5:
```suggestion
    private let post: AbstractPost
```
If this is always populated by the initializer, I think it can be a constant. Wdyt?

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    /// A completion block that is called after the user selects an option.
+    @objc var completion: (() -> Void)?
```
guarani, line 8: This might be able to go in the initializer along with `post`, but that might make things more complicated at the call site where `PostAuthorSelectorViewController` is initialized, so I'm happy to leave as-is.

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
twstokes: I've introduced a [private extension](https://github.com/wordpress-mobile/WordPress-iOS/pull/16281/files#diff-8ffdadf2fa39482fe8dc6c477d6c084c2e1aa486556fa102eac058fc2275cb66R85) to resolve this in the latest commit. Let me know thoughts!

Tested in a Playground:

```
let names = ["Sérgio", "Seth", "Sam"]

let newNames = names.sorted(by: { $0.diacriticsAndCaseInsensitive() < $1.diacriticsAndCaseInsensitive() } )

extension String {

    func diacriticsAndCaseInsensitive() -> String {
        return self.folding(options: [.diacriticInsensitive, .caseInsensitive], locale: .current)
    }
    
}

print(newNames)
```

guarani: Thanks 🙇

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+            }
+
+            return (displayName, $0.userID, $0.avatarURL)
+        }.sorted(by: { $0.displayName.diacriticsAndCaseInsensitive() < $1.displayName.diacriticsAndCaseInsensitive() })
+    }
+}
+
+private extension String {
+    func diacriticsAndCaseInsensitive() -> String {
+        return self.folding(options: [.diacriticInsensitive, .caseInsensitive], locale: .current)
```
guarani: Curious if there are tradeoffs between this approach vs. using `localizedCaseInsensitiveCompare`?

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -532,7 +547,10 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
     }
 
     if (sectionId == PostSettingsSectionMeta) {
-        if (indexPath.row == RowIndexForPassword) {
+        if (indexPath.row == PostMetaRowAuthor && ![self canShowAuthorSelection]) {
+            return 0;
+        }
+        if (indexPath.row == PostMetaRowPassword) {
```
guarani: When a post shouldn't show authors, is the author cell still present in the table view data source but is rendered with a height of zero?

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -46,10 +47,17 @@ typedef NS_ENUM(NSInteger, PostSettingsRow) {
     PostSettingsRowExcerpt
 };
 
+typedef NS_ENUM(NSInteger, PostMetaRow) {
+    PostMetaRowAuthor = 0,
+    PostMetaRowPublishDate,
+    PostMetaRowPublishStatus,
+    PostMetaRowVisibility,
+    PostMetaRowPassword
+};
```
guarani: Nice to see these indices in an enum ❤️

guarani (commented):
Just left a few comments in the code for now, I haven't tested yet, will do that in the next round of review.

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
twstokes: Nope, thanks for pointing that out. I much prefer that and overlooked that it was available.

twstokes: This should be resolved with the latest commit.

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
twstokes: My Swift -> ObjC -> Swift is weak here, but we're subclassing `SettingsSelectionViewController` and on `init` of our subclass calling `super.init(dictionary: NSDictionary)`. That method inside of `SettingsSelectionViewController` expects to return an `instancetype` of the return of the call to:

```initWithStyle:(UITableViewStyle)style andDictionary:(NSDictionary *)dictionary;```
                
It appears if that's not available in the subclass, we crash. This is also the reason why the `post` property of `PostAuthorSelectorViewController` needs to be an implicitly unwrapped var, otherwise its initialization would be required in the `super.init` call to these methods.

I'm open to ideas - this was mostly boilerplate I took from `PostVisibilitySelectorViewController`.

File: WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift (hunk 6 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift:5]: I touched on this a little here regarding the override inits: https://github.com/wordpress-mobile/WordPress-iOS/pull/16281#discussion_r619296509

File: WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift (hunk 8 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift]: Nope! `localizedCaseInsensitiveCompare` is the way to go, I've pushed the change. Thanks!

Comment (guarani) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift]: No worries!

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m (hunk 9 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: Correct, it's also marked as `hidden` [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/16281/files#diff-37b18ea4939f28d52d0e63f7a5fa45de61fe0ad85c3d402e98b6e5e10a1dc45cR668).

I don't find this pleasant or graceful at all, but it's what came up in my research. 😄  Will be glad to research more. My biggest concern would be a11y (we don't want it to exist if hidden), and I can double-check that.

File: WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift (hunk 7 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift:8]: I generally like to inject the dependency at the call site when possible as well, but I'm good either way. I'm fine with it as well.

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m (hunk 9 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: VoiceOver did not appear to pick up the Author button when it wasn't shown.

File: WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift (hunk 7 above)
Comment (guarani) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift:8]: Sounds good!

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m (hunk 9 above)
Comment (guarani) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: I hadn't noticed that it was hidden as well, so that helps. I was wondering though, would actually not including the cell at all be possible? If a cell is not needed I would have expected it to not be in the data source at all, but that doesn't appear to be the case here. What made you prefer hiding the cell instead here?

Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: Originally all rows were shown, but the Password was optional. This was easy to control because we could just +1 the rows in the meta section.

Since the Author option was at the top I took a different approach. I agree, the cell should never be created if it isn't needed. I'll come up with a better solution.

**Comment by guarani** [comment]
Regarding the test cases, I think we could expand them to be more specific to cover the feature more completely. The test cases serve a variety of purposes:
- Demonstrate what the PR author has done to verify the changes work as expected
- Provide a baseline for what testing PR reviewers should do
- Provide beta testers with an overview of the feature (a sentence or two describing the feature, a link to the main issue provides more details) and a baseline for what testing should be done

> ### To test:
> 
> **Creating:**
> 
>     1. Create a new Post or Page on a self-hosted or WordPress.com site that has multiple users.

Since the above step doesn't specify which site type was tested, it's hard to know what was tested and it's also not clear what's the expected baseline for testing (e.g. should reviewers and beta testers test either self-hosted or WordPress.com sites, or both?).

>     6. Observe in the app and on the web that the new Post or Page has the chosen author.

It's great that these tests cover checking the end result of the (change author) operation 👍 

I think the test steps adequately cover the different roles that can be used to manage the new author picker. Would it be possible to also specify which roles should be available in the picker itself? For example, I wonder if a [contributor](https://wordpress.org/support/article/roles-and-capabilities/#summary-of-roles) on the site should appear in the list of available authors.

> ## Regression Notes
> 
>     1. Potential unintended areas of impact

It feels like other items in the "meta" section could be affected since we're modifying table view indices here.

>     1. What I did to test those areas of impact (or what existing automated tests I relied on)

If you agree with the above, specifying which manual tests you did here (or which automated test were used) would be good.
 
>     1. What automated tests I added (or what prevented me from doing so)
> 
> 
>     * None for this PR.

It would be good to mention the reasons for not adding tests.

**Comment by twstokes** [comment]
> Since the above step doesn't specify which site type was tested, it's hard to know what was tested and it's also not clear what's the expected baseline for testing (e.g. should reviewers and beta testers test either self-hosted or WordPress.com sites, or both?).

This was written a way to express that all permutations (four) of that sentence should be tested. I feel that reviewers and testers should not base their testing (nor necessarily trust 😄) what the author has already claimed (though I did test these paths). I understand that SDLCs can differ, especially if a dedicated QA role doesn't exist.

I'll update the text to better fit this model.

> For example, I wonder if a contributor on the site should appear in the list of available authors.

This is a good point and something I should've considered.

> It feels like other items in the "meta" section could be affected since we're modifying table view indices here.

Will update this and the rest. I've put this back in a Draft mode.

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m (hunk 9 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: A proposed fix has been implemented.

**Comment by twstokes** [comment]
There's a discrepancy in this issue's label and title. Should this be a bug or feature?

File: WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift (hunk 5 above)
Comment (twstokes) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift:62]: I believe [this is our answer](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html#ID221).

Since we have inits calling other inits via subclass, Swift expects the more "specialized" subclass to have the appropriate initializers. I guess the compiler isn't smart enough to catch this? (it compiles without the init overrides)

**Comment by twstokes** [comment]
> Would it be possible to also specify which roles should be available in the picker itself? For example, I wonder if a contributor on the site should appear in the list of available authors.

The way web works for both WordPress.com and self-hosted sites, users who have the Contributor role can be set as authors for any Post or Page. This logic makes sense to me and matches the functionality in the app.

On self-hosted sites there's an additional role - Subscriber. They do not come over when the app queries the API for users, so they are not populated as an author, maintaining parity with web. They will not show in the list of authors so they cannot be selected as one.

**Comment by guarani** [comment]
> There's a discrepancy in this issue's label and title. Should this be a bug or feature?

Thanks for flagging! I shouldn't have put that bug label – fixed now.

File: WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift (hunk 5 above)
Comment (guarani) [reply WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift:62]: Thanks for the link! That looks like it.

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m (hunk 9 above)
Comment (guarani) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: Thanks! I really like the new solution. I think the approach used for the password row was not scalable and was bound to break and now just happened to be that time.

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m (hunk 10 above)
Comment (guarani) [reply WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m]: Now that we've gone with the new approach of dynamically generating the list of rows, I wonder if this new enum is really needed, wdyt? I think we could just as easily use the existing enum which has a case for each row.

File: WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m [hunk 11]
```diff
@@ -435,10 +443,8 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
         return 2;
 
     } else if (sec == PostSettingsSectionMeta) {
-        if (self.apost.password) {
-            return 4;
-        }
-        return 3;
+        [self configureMetaSectionRows];
```
guarani: Is there an advantage to putting this here in `numberOfRowsInSection` instead of in lifecycle method of the view controller itself? If (and I might be wrong) the number of cells is not changed after the view controller is initialized, it seems like this would be better of in an initializer or `viewDidAppear` or similar.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController_Internal.h
```diff
@@ -18,6 +18,8 @@ typedef enum {
 
 @property (nonnull, nonatomic, strong) NSArray *sections;
 
+@property (nonnull, nonatomic, strong) NSArray *postMetaSectionRows;
```
guarani: Would this be better off in the `.m` file? That seems like the default location since this looks to be a private property.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -635,10 +643,38 @@ - (UITableViewCell *)configureTaxonomyCellForIndexPath:(NSIndexPath *)indexPath
...
+    [metaRows addObjectsFromArray:@[ @(PostMetaRowPublishDate),
+                                      @(PostMetaRowPublishStatus),
+                                      @(PostMetaRowVisibility) ]];
+
+    if (self.apost.password) {
+        [metaRows addObject:@(PostMetaRowPassword)];
+    }
+
+    self.postMetaSectionRows = [metaRows copy];
+}
```
guarani: This approach looks great! The app would probably benefit from this approach in other places as well :D

guarani (commented):
Thanks again for the updates, @twstokes! I've left some comments. I think it's looking good but will leave the testing for the next round of review.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
twstokes: Totally agree! Deleted code is best code. Thanks for spotting this.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
twstokes: Good thoughts, and I totally agree I'd love for this area to be as declarative as possible - it wasn't my first choice, but I ultimately decided to follow the pattern we did with `configureSections`.

The number of cells can change if the Visibility setting for the Post / Page changes (e.g. it will show or hide the Password row), so there is some dynamic mutating going on.

- I don't believe `viewDidAppear` has an effect because `viewWillAppear` already called `reloadData`, so we're too late.
- Putting it before `reloadData` in `viewWillAppear` semi-works, but the cell height isn't right.
- Putting it in the init probably won't work due to it needing to be dynamic.

I'll look into this some more and see if something better is available.

twstokes: > Putting it before reloadData in viewWillAppear semi-works, but the cell height isn't right.

Turns out this was a red herring that led me to the correct solution. I failed to do a proper lookup on the Password row in `heightForRowAtIndexPath`.

The proposed and implemented solution is:

- Calling `configureSections` in `viewWillAppear` right before we refresh the table with `reloadData`
- Properly detecting when we're calculating the height for the Password row

WordPress/Classes/ViewRelated/Post/PostSettingsViewController_Internal.h
twstokes: Good call - I was following what we did with `sections` as a model but this makes total sense. Done.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -532,7 +531,8 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
     }
 
     if (sectionId == PostSettingsSectionMeta) {
-        if (indexPath.row == RowIndexForPassword) {
+        NSInteger row = [[self.postMetaSectionRows objectAtIndex:indexPath.row] integerValue];
```
twstokes, line 534: This isn't the prettiest, so if you think we would benefit stylistically by breaking it out into a function that this and `configureMetaPostMetaCellForIndexPath` would call, I'll be glad to.

twstokes:
> Thanks again for the updates, @twstokes! I've left some comments. I think it's looking good but will leave the testing for the next round of review.

Thanks @guarani for all the great feedback. I believe I've addressed all your comments and made the appropriate changes.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
```diff
@@ -69,6 +69,7 @@ @interface PostSettingsViewController () <UITextFieldDelegate,
 @property (nonatomic, strong) AbstractPost *apost;
 @property (nonatomic, strong) UITextField *passwordTextField;
 @property (nonatomic, strong) UIButton *passwordVisibilityButton;
+@property (nonatomic, strong) NSArray *postMetaSectionRows;
```
twstokes, line 72: I sort of wedged it in here with the other arrays.

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
guarani: Resolving this comment thread now – I think your spot on with the reason why these overrides are needed.

WordPress/Classes/ViewRelated/Post/PostAuthorSelectorViewController.swift
guarani: Thanks again, looks like these are needed. Resolving this thread.

WordPress/Classes/ViewRelated/Post/PostSettingsViewController.m
guarani: I think it's ok, I don't have any strong opinions about this one.

guarani:
> > Since the above step doesn't specify which site type was tested, it's hard to know what was tested and it's also not clear what's the expected baseline for testing (e.g. should reviewers and beta testers test either self-hosted or WordPress.com sites, or both?).
> 
> This was written a way to express that all permutations (four) of that sentence should be tested. I feel that reviewers and testers should not base their testing (nor necessarily trust 😄) what the author has already claimed (though I did test these paths). I understand that SDLCs can differ, especially if a dedicated QA role doesn't exist.

I think I understand where you're coming from. Without a template to follow there's lots of leeway to write the testing section of the PR out how the author sees fit. Using previous PRs as a guide is useful, but PRs vary a lot (bugs, features, infrastructure updates) and a good understanding of "the norm" here comes with time.
The only concerns with the previous testing instructions were that it felt a bit too open to interpretation. Some might interpret the OR statement as "do all permutations of these tests" as you suggest while others might understand it as "either of these test scenarios accomplish the same thing, so take your pick". I feel that being a little explicit and not assuming knowledge on the PR reviewer's part can take a lot of the guess-work out of testing.

twstokes:
> > > Since the above step doesn't specify which site type was tested, it's hard to know what was tested and it's also not clear what's the expected baseline for testing (e.g. should reviewers and beta testers test either self-hosted or WordPress.com sites, or both?).
> > 
> > 
> > This was written a way to express that all permutations (four) of that sentence should be tested. I feel that reviewers and testers should not base their testing (nor necessarily trust 😄) what the author has already claimed (though I did test these paths). I understand that SDLCs can differ, especially if a dedicated QA role doesn't exist.
> 
> I think I understand where you're coming from. Without a template to follow there's lots of leeway to write the testing section of the PR out how the author sees fit. Using previous PRs as a guide is useful, but PRs vary a lot (bugs, features, infrastructure updates) and a good understanding of "the norm" here comes with time.
> The only concerns with the previous testing instructions were that it felt a bit too open to interpretation. Some might interpret the OR statement as "do all permutations of these tests" as you suggest while others might understand it as "either of these test scenarios accomplish the same thing, so take your pick". I feel that being a little explicit and not assuming knowledge on the PR reviewer's part can take a lot of the guess-work out of testing.

Totally agree! And I think 

> "the norm" here comes with time 

nails it. 🔨

twstokes:
Looks like `PostAuthorSelectorViewController` wasn't compiled for the tests. 🤔 

The target membership for Jetpack wasn't set on the file, so I'll set that now.

twstokes:
This may need to get bumped to 17.4.

jkmassel:
👋 We're freezing 17.3 today, so this PR is being bumped to 17.4

guarani (approved):
Testing the latest commit and it looks good to me! :shipit: 

## Self-hosted

- [x] A site that has multiple users (all Contributor role or higher) gives Administrator roles the option to change the author.
- [x] Roles lower than Administrator do not see the option to change an author, even if there are multiple author candidates.
- [x] Contributor roles do not show as an author option and do not add to the author count.
- [x] Creating a new Post and setting the author. (see Creating steps)
- [x] Updating an existing Post and changing the author. (see Updating steps)
- [x] Deleting authors. (see Deleting steps)

## WordPress.com

- [x] A site that has multiple users gives Administrator and Editor roles the options to change the author.
- [x] Creating a new Post and setting the author. (see Creating steps)
- [x] Updating an existing Post and changing the author. (see Updating steps)
- [x] Deleting authors. (see Deleting steps)

guarani:
Thank you very much @twstokes for the work here bringing author selection to WPiOS! ❤️

PR #16298, opened 2021-04-14
Query for authors and hide separator in post list UI for empty author names.

### LINKED ISSUE #16297: Bug: Author names don't appear in Blog Posts list
This bug only applies to self-hosted WordPress sites.

### Expected behavior

In the Blog Posts view either:

1. An author name should be shown following the dot separator after the published date.
2. The dot separator should not be shown after the date if an author name isn't available or if an author shouldn't be shown.

(please see the attached screenshot)

### Actual behavior
In some cases no name is shown and the dot separator is still present.

1. Publish a sample blog post to a self-hosted WordPress site.
2. From the My Site view, tap "Blog Posts".
3. Observe a dot separator with no author name following. **Note:** For Administrator roles, view posts for "Everyone".

twstokes:
Fixes #16297

This PR aims to:

1. Show author names for posts (when possible) for self-hosted WordPress sites.
2. Hide the published date and author name separator when no author name is available.

To test:

1. Publish a sample blog post to a self-hosted WordPress site.
2. From the My Site view, tap "Blog Posts".
3. Observe a dot separator with no author name following. **Note:** For Administrator roles, view posts for "Everyone".

## Regression Notes
None identified.

For the UI portion, existing tests.

A test has been added for the scenario when the author name should be shown, but no name is available.

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 17.2
 -----
 
+* [*] Fixed an issue where post author display names weren't visible. [#16297]
 * [***] Updated custom app icons. [#16261]
 
 17.1
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,7 @@
 17.2
 -----
 
-* [*] Fixed an issue where post author display names weren't visible. [#16297]
+* [*] Fixed an issue where some author display names weren't visible for self-hosted sites. [#16297]
 * [***] Updated custom app icons. [#16261]
 
 17.1
--- a/WordPress/Classes/Models/Blog+BlogAuthors.swift
+++ b/WordPress/Classes/Models/Blog+BlogAuthors.swift
@@ -3,7 +3,7 @@ import CoreData
 
 
 extension Blog {
-    @NSManaged public var authors: NSSet?
+    @NSManaged public var authors: Set<BlogAuthor>?
 
 
     @objc(addAuthorsObject:)
@@ -17,4 +17,9 @@ extension Blog {
 
     @objc(removeAuthors:)
     @NSManaged public func removeFromAuthors(_ values: NSSet)
+
+    @objc
+    func getAuthorWith(id: NSNumber) -> BlogAuthor? {
+        return authors?.first(where: { $0.userID == id })
+    }
 }
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -804,7 +804,10 @@ typedef void (^AutosaveSuccessBlock)(RemotePost *post, NSString *previewURL);
 - (void)updatePost:(AbstractPost *)post withRemotePost:(RemotePost *)remotePost {
     NSNumber *previousPostID = post.postID;
     post.postID = remotePost.postID;
-    post.author = remotePost.authorDisplayName;
+    // Used to populate author information for self-hosted sites.
+    BlogAuthor *author = [post.blog getAuthorWithId:remotePost.authorID];
+
+    post.author = remotePost.authorDisplayName ?: author.displayName;
     post.authorID = remotePost.authorID;
     post.date_created_gmt = remotePost.date;
     post.dateModified = remotePost.dateModified;
@@ -824,7 +827,7 @@ typedef void (^AutosaveSuccessBlock)(RemotePost *post, NSString *previewURL);
     if (post.pathForDisplayImage.length == 0) {
         [post updatePathForDisplayImageBasedOnContent];
     }
-    post.authorAvatarURL = remotePost.authorAvatarURL;
+    post.authorAvatarURL = remotePost.authorAvatarURL ?: author.avatarURL;
     post.mt_excerpt = remotePost.excerpt;
     post.wp_slug = remotePost.slug;
     post.suggested_slug = remotePost.suggestedSlug;
--- a/WordPress/Classes/ViewRelated/Post/Revisions/RevisionsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Post/Revisions/RevisionsTableViewController.swift
@@ -96,8 +96,11 @@ private extension RevisionsTableViewController {
     }
 
     private func getAuthor(for id: NSNumber?) -> BlogAuthor? {
-        let authors: [BlogAuthor]? = post?.blog.authors?.allObjects as? [BlogAuthor]
-        return authors?.first { $0.userID == id }
+        guard let authorId = id else {
+            return nil
+        }
+
+        return post?.blog.getAuthorWith(id: authorId)
     }
 
     private func getRevisionState(at indexPath: IndexPath) -> RevisionBrowserState {
@@ -193,12 +196,13 @@ extension RevisionsTableViewController: WPTableViewHandlerDelegate {
         }
 
         let revision = getRevision(at: indexPath)
-        let authors = getAuthor(for: revision.postAuthorId)
+        let author = getAuthor(for: revision.postAuthorId)
+
         cell.title = revision.revisionDate.shortTimeString()
-        cell.subtitle = authors?.username ?? revision.revisionDate.mediumString()
+        cell.subtitle = author?.username ?? revision.revisionDate.mediumString()
         cell.totalAdd = revision.diff?.totalAdditions.intValue
         cell.totalDel = revision.diff?.totalDeletions.intValue
-        cell.avatarURL = authors?.avatarURL
+        cell.avatarURL = author?.avatarURL
     }
 
 
--- a/WordPress/Classes/ViewRelated/Post/PostCardCell.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostCardCell.swift
@@ -40,10 +40,12 @@ class PostCardCell: UITableViewCell, ConfigurablePostView {
     private var currentLoadedFeaturedImage: String?
     private weak var interactivePostViewDelegate: InteractivePostViewDelegate?
     private weak var actionSheetDelegate: PostActionSheetDelegate?
-    var isAuthorHidden: Bool = false {
+    var shouldHideAuthor: Bool = false {
         didSet {
-            authorLabel.isHidden = isAuthorHidden
-            separatorLabel.isHidden = isAuthorHidden
+            let emptyAuthor = viewModel?.author.isEmpty ?? true
+
+            authorLabel.isHidden = shouldHideAuthor || emptyAuthor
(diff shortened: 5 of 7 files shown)


WordPress/Classes/ViewRelated/Post/Revisions/RevisionsTableViewController.swift
```diff
@@ -96,8 +96,11 @@ private extension RevisionsTableViewController {
     }
 
     private func getAuthor(for id: NSNumber?) -> BlogAuthor? {
```
twstokes, line 98: I realize this isn't a best practice per the style guideline, but due to the double-optional nature of a post and revision.postAuthorId, I struggled to find a cleaner solution. I decided to keep the private function so that configureCell would stay remain succinct.

Any tips for improving this are certainly welcome!

WordPress/Classes/ViewRelated/Post/Revisions/RevisionsTableViewController.swift
```diff
@@ -96,8 +96,11 @@ private extension RevisionsTableViewController {
...
-        let authors: [BlogAuthor]? = post?.blog.authors?.allObjects as? [BlogAuthor]
-        return authors?.first { $0.userID == id }
+        guard let authorId = id else {
```
twstokes, line 99: I realize this isn't a best practice per the style guidelines, but due to the double-optional nature of a post and revision.postAuthorId, I struggled to find a cleaner solution. I decided to keep the private function so that configureCell would remain succinct.

Any tips for improving this are certainly welcome!

guarani:
Since this is a fork, I've made a temporary PR in https://github.com/wordpress-mobile/WordPress-iOS/pull/16312 to trigger the CI. Once CI passes on the other PR, GitHub should automatically put green CI checks here because it will detect that the same commit found in this PR passed the checks in the other PR.

twstokes:
> Since this is a fork, I've made a temporary PR in #16312 to trigger the CI.

Thank you @guarani!

WordPress/Classes/Models/Blog+BlogAuthors.swift
```diff
@@ -17,4 +17,9 @@ extension Blog {
 
     @objc(removeAuthors:)
     @NSManaged public func removeFromAuthors(_ values: NSSet)
+
+    @objc
+    func getAuthorWith(id: NSNumber) -> BlogAuthor? {
+        return authors?.first(where: { $0.userID == id })
```
guarani, line 23: Does the `==` here compare the user ID values or the memory address of the `NSNumber` objects?

twstokes: This should be a comparison between two `NSNumber`s now that `authors` is a `Set<BlogAuthors>` and `id` is the passed in parameter.

For better clarity I could go with `getAuthorWith(userID: NSNumber)`. Possibly that'd be better?

guarani: Sorry, I meant to ask if
```
return authors?.first(where: { $0.userID == id })
```
is the correct way to compare `NSNumber` objects in Swift. 

Is there any difference between using `==` (as you do) and this?
```
return authors?.first(where: { $0.userID.isEqual(to: id) })
```

> For better clarity I could go with getAuthorWith(userID: NSNumber). Possibly that'd be better?

I think it's fine as-is (just my personal opinion).

RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 17.2
 -----
 
+* [*] Fixed an issue where post author display names weren't visible. [#16297]
```
guarani: It might be worth specifying that this was a bug specific to self-hosted sites (which I understand from the issue was the case).
(The convention of adding `[#16297]` to the release notes is an aid to internal beta testers. When editorial make the App Store release notes from these bullet points, they won't be looking at the PR so won't know which site types were affected by this bug.)

WordPress/Classes/Models/Blog+BlogAuthors.swift
twstokes: Ah, I see! Sorry about that! According to [Apple's docs](https://developer.apple.com/documentation/foundation/nsnumber/1411315-isequal), `.isEqual` on a NSNumber:

> Returns a Boolean value that indicates whether the number object’s value and a given number are equal.

So I believe it only cares about the value inside and not the object itself, which would be equatable to `==`.

WordPress/Classes/Services/PostService.m
```diff
@@ -824,7 +827,7 @@ - (void)updatePost:(AbstractPost *)post withRemotePost:(RemotePost *)remotePost
     if (post.pathForDisplayImage.length == 0) {
         [post updatePathForDisplayImageBasedOnContent];
     }
-    post.authorAvatarURL = remotePost.authorAvatarURL;
+    post.authorAvatarURL = remotePost.authorAvatarURL ?: author.avatarURL;
```
guarani, line 830: Could you please expand on this change?
It looks like what this does is use the `avatarURL` from the `Blog.authors` instead of the remote post. Had you observed an issue with avatars, similar to the issue with author names, that led you to make this change?

WordPress/Classes/Services/PostService.m
```diff
@@ -804,7 +804,10 @@ - (NSDictionary *)remoteSyncParametersDictionaryForRemote:(nonnull id <PostServi
 - (void)updatePost:(AbstractPost *)post withRemotePost:(RemotePost *)remotePost {
     NSNumber *previousPostID = post.postID;
     post.postID = remotePost.postID;
-    post.author = remotePost.authorDisplayName;
+    // Used to populate author information for self-hosted sites.
+    BlogAuthor *author = [post.blog getAuthorWithId:remotePost.authorID];
+
+    post.author = remotePost.authorDisplayName ?: author.displayName;
```
guarani, line 810: My understanding here is that `remotePost.authorDisplayName` is `nil` for self-hosted sites, so we fall back to using the `Blog.authors` set to retrieve the author display name, is that correct?

If so, do you know if the API is supposed to return the `authorDisplayName` but not doing so?

WordPress/Classes/Models/Blog+BlogAuthors.swift
twstokes: I understand now, thanks!

This is what Apple's docs are showing for the `==` between two NSObjects like these, so I think it's equivalent.

```
extension NSObject : Equatable, Hashable {

    /// Returns a Boolean value indicating whether two values are
    /// equal. `NSObject` implements this by calling `lhs.isEqual(rhs)`.
    ///
    /// Subclasses of `NSObject` can customize Equatable conformance by overriding
    /// `isEqual(_:)`. If two objects are equal, they must have the same hash
    /// value, so if you override `isEqual(_:)`, make sure you also override the
    /// `hash` property.
    ///
    /// - Parameters:
    ///   - lhs: A value to compare.
    ///   - rhs: Another value to compare.
    public static func == (lhs: NSObject, rhs: NSObject) -> Bool
```

guarani: That makes sense! Thanks for confirming and answering this question I had in the back of my mind.

twstokes: Very welcome! Thanks for bringing it up!

RELEASE-NOTES.txt
twstokes: Good call, and thanks for the extra information ! I'll get this fixed.

WordPress/Classes/ViewRelated/Post/Revisions/RevisionsTableViewController.swift
guarani: If you're referring to the additional guard statement, I think its is fine here 👍

WordPress/Classes/Services/PostService.m
twstokes: Glad to!

> It looks like what this does is use the avatarURL from the Blog.authors instead of the remote post.

Correct - it will try to use a queried author's avatarURL in the case that the remotePost didn't include it. 

In my testing of this, WordPress.com sites included an author on the remote post ([see the author object response parameter here](https://developer.wordpress.com/docs/api/1.1/get/sites/%24site/posts/%24post_ID/)), so `remotePost.authorAvatarURL` was always available.

Self-hosted sites, however, did not include an author object so this is why the author query was introduced.

Though `author.avatarURL` is not used for this bug fix, I set the property because I felt it made the solution more complete (e.g. if we have an author already, let's set the avatarURL instead of `nil`).

Please let me know if I should clarify any of this! Thanks.

WordPress/Classes/Services/PostService.m
twstokes: > My understanding here is that remotePost.authorDisplayName is nil for self-hosted sites, so we fall back to using the Blog.authors set to retrieve the author display name, is that correct?

Correct.

> If so, do you know if the API is supposed to return the authorDisplayName but not doing so?

From my understanding, only an author ID is returned from the XML-RPC call ([see the Return Values section for `wp.getPost` here](https://codex.wordpress.org/XML-RPC_WordPress_API/Posts)).

RELEASE-NOTES.txt
guarani: Sure, that's not documented anywhere (I'd like to add a "how to maintain release notes" section to our release notes) 🙇

guarani: > I'd like to add a "how to maintain release notes" section to our release notes

That's something I'll follow up on myself.

guarani (commented):
Smoke testing self-hosted and WP.com sites and this is looking good! Just left a few comments to clarify a few things.

WordPress/Classes/ViewRelated/Post/Revisions/RevisionsTableViewController.swift
twstokes: Yep, good deal!

RELEASE-NOTES.txt
twstokes: This is updated. 👍

WordPress/Classes/Services/PostService.m
guarani: I see, thank you for the link 🙇

WordPress/Classes/Services/PostService.m
guarani: That makes sense. Looks like the XML-RPC API only provides the post's author ID so it makes sense to use the author object from the Blog to fetch these extra attributes like avatar URL.

guarani (approved):
LGTM! Thank you @twstokes! Nice work ❤️

PR #16301, opened 2021-04-14
Sync authors with the Blog Service

twstokes:
This is one PR in a series towards implementing #9015.

This PR is branched from #16269, so it should be merged first.

**Why this change is needed:** Once we provide the ability for the user to select a new author for a Post or Page, we don't want removed authors to stay in that list.

**What changed:**

1. The `addToAuthors` and `removeFromAuthors` methods that took `NSSet` were updated for a parameter of `Set<BlogAuthor>`. This should now match their single value counterpart.
2. If our local set of authors includes a userID that doesn't exist in our latest `RemoteUser` array, we remove that author from the local store because they've been removed from the blog.

### To test:

The only area I've identified as being impacted is the History view for Posts:

1. Make changes to a WordPress.com Post as multiple users (e.g. User A and User B have both made updates to the same Post).
2. With the post presented in the app, tap the horizontal ellipses (More Options) icon at the top right.
3. Select "History".
4. Observe the user names and avatars next to each revision.
5. Remove an author from the blog.
6. Restart the app.
7. View the history for the same post.
8. Observe that the removed author no longer has a display name nor avatar icon.

## Regression Notes

- Any areas that expect a deleted author's information to be available. 
- Since the `authors` property is NSManaged, I would appreciate confirmation that using `Set<BlogAuthor>` over the original `NSSet` parameter won't cause any negative effects.

Used the test sequence above.

Added `BlogServiceAuthorTests`.

PR submission checklist:


twstokes:
Requesting a review from @guarani. Thank you!

guarani (commented):
> **Why this change is needed:** Once we provide the ability for the user to select a new author for a Post or Page, we don't want removed authors to stay in that list.

This appears to be a change in behavior and I just want to confirm that this is actually what we want. Would this change bring the app closer to the web or make it behave differently?

>     2. If our local set of authors includes a userID that doesn't exist in our latest `RemoteUser` array, we remove that author from the local store because they've been removed from the blog.

Again here, the test steps suggest that this PR changes how history works by no longer showing names or avatar icons of users who are no longer associated with the site. Is this a change in behavior we want? 

Rather than answering the question for ourselves, I wonder if it's simpler to avoid this question entirely by not changing History or other parts of the app while adding the functionality to the app to edit post authors.

I appreciate you making this PR, I just want to make sure we're not changing other parts of the app (e.g. post history) if we don't need to here, or providing a different experience to WordPress on the web.

twstokes:
Hey @guarani, great points. My thought process was:

"Without adding more complexity (e.g. a way to retain author information locally, but mark them as "removed"), what would impact the user more - The ability to select an invalid author for a Post or Page and cause an error, or losing the ability to see the name and avatar of a removed author for a Post or Page's history." In hindsight this is perhaps a false dichotomy.

I went with dropping the latter because it felt less disruptive, but you're right - it's straying away from how the web works (WordPress.com, at least - it seems self-hosted may remove the revision?).

Rethinking this, I would lean towards adding a property to `BlogAuthor` that tracks whether they have been removed (in this case, didn't appear in the `RemoteUsers` array) that gets set in `updateBlogAuthors`, and filter on it when generating the list of authors to choose from in `PostSettingsViewController`.

Thanks for your thoughts and review!

guarani:
Thanks for clarifying. I think I'd overlooked the importance of this for the change author functionality and focused instead on its side-effects. It's great to see there's a solution that doesn't affect the post history screen at all, while achieving the goal of not displaying stale users in the change author picker.

I'm looking forward to re-reviewing once the changes are ready 🙇‍♂️

twstokes:
Closing this and replacing with https://github.com/wordpress-mobile/WordPress-iOS/pull/16349. Thanks!

Issue #16245, opened 2021-04-06
Like Notification > User Profile: show site when selected


ScoutHarris:
When the user's site is selected in the [User Profile](https://github.com/wordpress-mobile/WordPress-iOS/issues/16244), show the site in the Reader's site view in a fullscreen modal.

References:
- Likes Enhancements i2 design - pbArwn-1CQ-p2

![site_display](https://user-images.githubusercontent.com/1816888/113787278-f49dc180-96f7-11eb-9fd0-d4365a6c2ebc.png)

PR #16302, opened 2021-04-14
User Profile: show site modal when site row tapped

ScoutHarris:
Ref #16245 

When the site is tapped on the User Profile, the site is displayed using the Reader site view it in a fullscreen modal.

The Reader site view is what is presented when tapping the site name on a Reader post card. The differences here are:
- The fullscreen modal.
- `X` left nav bar button to dismiss the view.

Otherwise they should look the same.
 
| Reader Site | User Site |
|--------|-------|
| reader_site | user_site |

---
To test:

Since we don't have real user info yet, I've stubbed in a bogus siteID. You'll need valid siteIDs for the modal to show. The [siteID](https://github.com/wordpress-mobile/WordPress-iOS/blob/e56b67b9d3d85419685980c57affc8f8fc8d6d13/WordPress/Classes/ViewRelated/User%20Profile%20Sheet/UserProfileSheetViewController.swift#L134) is set in  `UserProfileSheetViewController:showSite`.

---
With an invalid siteID (as-is):
- Enable the `newLikeNotifications` feature.
- Go to Notifications > Likes and tap a notification.
- Tap a user, then the site row.
- Nothing will happen visually. This error will appear in the logs: `Error: could not find ReaderSiteTopic for siteID [xxx]`

---
With a valid siteID:
- Set the siteID value as mentioned above.
- Go to Notifications > Likes and tap a notification.
- Tap a user, then the site row.
- Verify the site is displayed:
  - In a fullscreen modal.
  - With the `X` left nav bar button.

| iPhone | iPad |
|--------|-------|
| iphone | ipad |

---
## Regression Notes
N/A. Feature is incomplete.

N/A. Feature is incomplete.

N/A. Feature is incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/ReaderTopicService.h
+++ b/WordPress/Classes/Services/ReaderTopicService.h
@@ -238,16 +238,6 @@ extern NSString * const ReaderTopicFreshlyPressedPathCommponent;
  */
 - (ReaderSiteTopic *)findSiteTopicWithFeedID:(NSNumber *)feedID;
 
-/**
- Find a site topic by siteID.
- Combines `findSiteTopicWithSiteID` and `findSiteTopicWithFeedID` to get a topic by ID.
- Useful when it's not known if the siteID is internal or external.
- 
- @param siteID The site id of the topic
- @return A matched site topic
- */
-- (ReaderSiteTopic *)siteTopicForSiteID:(NSNumber *)siteID;
-
 @end
 
 @interface ReaderTopicService (Tests)
--- a/WordPress/Classes/Services/ReaderTopicService.h
+++ b/WordPress/Classes/Services/ReaderTopicService.h
@@ -238,6 +238,16 @@ extern NSString * const ReaderTopicFreshlyPressedPathCommponent;
  */
 - (ReaderSiteTopic *)findSiteTopicWithFeedID:(NSNumber *)feedID;
 
+/**
+ Find a site topic by siteID.
+ Combines `findSiteTopicWithSiteID` and `findSiteTopicWithFeedID` to get a topic by ID.
+ Useful when it's not known if the siteID is internal or external.
+ 
+ @param siteID The site id of the topic
+ @return A matched site topic
+ */
+- (ReaderSiteTopic *)siteTopicForSiteID:(NSNumber *)siteID;
+
 @end
 
 @interface ReaderTopicService (Tests)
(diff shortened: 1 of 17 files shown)


WordPress/Classes/Services/ReaderTopicService.h
```diff
@@ -238,6 +238,16 @@ extern NSString * const ReaderTopicFreshlyPressedPathCommponent;
...
 
+/**
+ Find a site topic by siteID.
+ Combines `findSiteTopicWithSiteID` and `findSiteTopicWithFeedID` to get a topic by ID.
+ Useful when it's not known if the siteID is internal or external.
+ 
+ @param siteID The site id of the topic
+ @return A matched site topic
+ */
+- (ReaderSiteTopic *)siteTopicForSiteID:(NSNumber *)siteID;
```
aerych: Being so similar to `findSiteTopicWithSiteID` I wonder if we could call it something like `siteTopicForSiteOrFeedID` to better distinguish it and make usage more obvious at a glance? Or even something like `siteTopicForUnknownID` or `siteTopicForGenericID`.

WordPress/Classes/Services/ReaderTopicService.m
```diff
@@ -1176,6 +1176,11 @@ - (ReaderSiteTopic *)findSiteTopicWithFeedID:(NSNumber *)feedID
     return (ReaderSiteTopic *)[results firstObject];
 }
 
+- (ReaderSiteTopic *)siteTopicForSiteID:(NSNumber *)siteID
+{
+    return [self findSiteTopicWithSiteID:siteID] ?: [self findSiteTopicWithFeedID:siteID];
```
aerych: These methods will return a topic that is already synced to the device, but it seems likely that the user won't already be following many likers and we'd need to fetch site or feed info in those cases. 
`siteTopicForSiteWithID:isFeed:success:failure:` is set up to do this. Maybe the existing method would be a better use case?  
If it helps, the siteIDs we get from the likes endpoints should never be feedIDs, just siteIDs. (I think. Pretty sure Jetpack sites are siteIDs.  FeedIDs would only be external sites crawled and stored in the feedbag).

aerych (changes requested):
Howdy @ScoutHarris 👋 
Works as advertised if I happen to be following the topics I'm testing with, but I don't think we can count on that. I left some comments to that effect. Lemme know what you think.
Back to you!

ScoutHarris:
Hey @aerych . 

Thank you for your help! I've made these changes:
- Per our Slack discussion, I've removed the logic to display external sites (and the poorly thought out `siteTopicForSiteID` method). I'll address external sites in a follow up. 
- The User Profile VC now uses `readerTopicService.siteTopicForSite` to fetch a site by ID. 

aerych (approved):
Shiny! :shipit:  :)

Issue #15870, opened 2021-02-11
Alamofire.AFError.responseValidationFailed(reason: Alamofire.AFError.ResponseValidationFailureRea...


jkmassel:
**Sentry Url**: https://sentry.io/share/issue/eb331d147e8749af9d475641b46704bf/
**User Count**: 6
**Count**: 39
**First Release**: 16.7.0.20210208
**First Seen**: 2021-02-09T07:51:43.339000Z
**Last Seen**: 2021-02-11T20:19:55Z
**24 Hours**: 8
**30 Days**: 40

```
Alamofire.AFError.responseValidationFailed(reason: Alamofire.AFError.ResponseValidationFailureRea...

```

startuptester:
@guarani as the CTOW, can you please take a look?

ceyhun:
Assigning low priority as currently it affects ~500 users and since it doesn't seem to be a crash but a logged error.

bjtitus:
I made a change (https://github.com/wordpress-mobile/WordPress-iOS/pull/16314) to stop logging 404 errors as they would be expected when no site icon exists. This should reduce the volume of these errors by roughly half. Next steps should be to investigate the 400s (the next largest error group). I have a theory that they may be caused when Photon requests are made with bad parameters or when the URL passed to Photon has some kind of error.

jkmassel:
This issue is no longer being reported by Sentry

PR #14091, opened 2020-05-08
Don't log sentry error for cancelled Site Icon downloads

bjtitus:
Fixes #14051

To test:
- Set a breakpoint on UIImageView+SiteIcon lines 93 and 101.
- Search for a site using the Reader search page .
- When rapidly switching between searches (quick enough so that some site icons do not finish loading), you should NOT see the error handler (line 101) called when the `AFIError.requestCancelled` error occurs.

Throttling the network connection with Charles or Link Conditioner should help make it easier to switch between searches.

PR submission checklist:

- [x] I have considered adding unit tests where possible.
- [x] I have considered adding accessibility improvements for my changes.
- [x] ~I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.~ Not user facing.

PR #16314, opened 2021-04-15
Don't log errors on site icon 404s

bjtitus:
Removes error logging when receiving a 404 from the site icon request. It's expected that these icons may not exist, so we don't need to log these. The intent is to bring down the noise on some of these long running errors which may by inhibiting our ability to actually track the problem cases.

Looking at the results for #15870 over the last 14 days we can see that 404s show up > 50% of the time:

status | count
-- | --
404 | 1895
400 | 1565
403 | 201
429 | 2
500 | 5

### Testing

* Switch to a site with no site icon.
* Ensure that the 404 is not logged by checking for the console message.
* Use a proxy to rewrite the response of the site icon call to return a 400 error code.
* Ensure that this error code is shown as logged in the console.

## Regression Notes

This only impacts logging so there is no impact on users.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
+++ b/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
@@ -1,4 +1,5 @@
 import AlamofireImage
+import Alamofire
 import AutomatticTracks
 import Foundation
 import Gridicons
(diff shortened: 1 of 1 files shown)


bjtitus:
@frosty I'm assigning this to you only because you reviewed this slightly similar PR for me last year: https://github.com/wordpress-mobile/WordPress-iOS/pull/14091

This is just a further reduction of this error logging to Sentry for "expected" cases to reduce the noise on #15870. Let me know if you may not have time to review and I'm happy to reach out to others!

frosty (approved):
Tested following your instructions – a 404 doesn't log out errors, and other error codes I tested with (e.g. 400) do trigger the `logError` call.

PR #16317, opened 2021-04-15
Create Screenshots for 17.1

jkmassel:
Fixes screenshots as part of the `17.1 `release

To test:
1. Checkout this PR
2. Run `rake mocks`
3. In Xcode, run the `WordPressScreenshotGeneration` test on a blank simulator.

## Regression Notes
n/a

n/a

n/a

**PR submission checklist:**


Podfile
```diff
@@ -358,9 +358,9 @@ end
 ## ===================
 ##
 def wordpress_mocks
-  pod 'WordPressMocks', '~> 0.0.9'
+  # pod 'WordPressMocks', '~> 0.0.9'
   # pod 'WordPressMocks', :git => 'https://github.com/wordpress-mobile/WordPressMocks.git', :commit => ''
-  # pod 'WordPressMocks', :git => 'https://github.com/wordpress-mobile/WordPressMocks.git', :branch => ''
+  pod 'WordPressMocks', :git => 'https://github.com/wordpress-mobile/WordPressMocks.git', :branch => 'fix/reader'
```
mokagio: I assume before merging this PR we should have a new version of the mocks library out, yeah?

fastlane/ScreenshotFastfile
```diff
@@ -24,21 +24,17 @@ platform :ios do
...
     scan(
-      workspace: File.join(fastlane_directory, "../../WordPress.xcworkspace"),
+      workspace: WORKSPACE_PATH,
       scheme: "WordPressScreenshotGeneration",
       build_for_testing: true,
       derived_data_path: derived_data_path,
     )
 
-    languages = "da de-DE en-AU en-CA en-GB es-ES fr-FR id it ja ko no nl-NL pt-BR pt-PT ru sv th tr zh-Hans zh-Hant en-US".split(" ")
+    languages = "ru".split(" ")
```
mokagio: I believe this is debug code that sneaked into the PR?

fastlane/Snapfile
```diff
@@ -1,59 +0,0 @@
...
-# end
-
-# teardown_language do |lang, device|
-#   puts "Finished with #{lang} on #{device}"
-# end
-
-# teardown_device do |device|
-#   puts "Cleaning device #{device}"
-#   system("./cleanup.sh")
-# end
```
mokagio, line 59: +1 for moving the settings in this file to the dedicated `Fastfile` for screenshots. We don't need to spread information between files, in particular in when using a tool that does _a lot_ behind the scenes like Fastlane. Thanks.

mokagio (changes requested):
The tests passed from Xcode:

image

They also worked fine (as far as I can see) via Fastlane (with the hardcoded `"ru"` locale only still) once I [updated the helper](https://github.com/wordpress-mobile/WordPress-iOS/commit/9e4ad9f68b680fcb5c53a0d93ce755fbf3ec696b) as prompted by Fastlane. Feel free to cherry-pick.

screencapture-file-Users-gio-Developer-a8c-wpios-fastlane-screenshots-screenshots-html-2021-05-11-16_15_44

Requesting changes because of the hardcoded `"ru"` locale.

fastlane/ScreenshotFastfile
jkmassel: Ah yep, thanks!!

Podfile
jkmassel: Yes indeed – this is updated!

fastlane/ScreenshotFastfile
```diff
@@ -34,7 +34,7 @@ platform :ios do
       derived_data_path: derived_data_path,
     )
 
-    languages = "ru".split(" ")
+    languages = "da de-DE en-AU en-CA en-GB es-ES fr-FR id it ja ko no nl-NL pt-BR pt-PT ru sv th tr zh-Hans zh-Hant en-US".split(" ")
```
mokagio: Random question: have you considered making this an array? Would it help maintaining the setup? 🤔

mokagio (approved):
I did not run the screenshots automation locally with the new changes, but by looking at the commits I'm 99.9% confident the result would be the same as my previous run.

I didn't see the [snapshot helper update commit](https://github.com/wordpress-mobile/WordPress-iOS/commit/9e4ad9f68b680fcb5c53a0d93ce755fbf3ec696b). I don't think that's a blocker, but I wonder if we should add it? On my machine, I wasn't able to run the `screenshots` lane before applying those changes 🤔

fastlane/ScreenshotFastfile
jkmassel: Yeah, good question – ideally we'd pull this from a global "list of locales we support" that's aligned between screenshots, metadata, glotpress, etc. For now I'm inclined to leave it, but I look forwarding to having it be a lot nicer, as you say!

PR #16319, opened 2021-04-15
Show posts in the app via their wordpress.com subdomain link

Gio2018:
This PR allows to show posts in the reader by providing a link in the following format:

`site_name.wordpress.com/post_year/post_month/post_day/post_title`

Fixes #NA

To test:

1. Find the method `handleWebActivity` in `WordPRessAppDelegate
2. Change the following call, hardcoding a valid URL of the format specified above:
    - `UniversalLinkRouter.shared.handle(url: url)`
3. Build and run in a simulator
4. in the terminal launch the following command:
    - `xcrun simctl openurl booted [currently_recognized_url]`
    where `currently_recognized_url` is an url that would be currently recognized by the app as a valid url, for example a valid url in the following form:
    - `https://wordpress.com/read/feeds/feed_id/posts/post_id` (this is needed to trigger the handling of the web activity)
5. Make sure that the post of the hardcoded link is displayed in the app, in the Reader Detail page.

## Regression Notes
low/none, this only adds a new route for the reader, as well a new (set) of associated domains. It also refactors an existing method used to show posts

Tested that the existing routes still work as expected, especially the existing one for a reader post.

None

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Universal Links/Routes+Reader.swift
+++ b/WordPress/Classes/Utility/Universal Links/Routes+Reader.swift
@@ -14,6 +14,7 @@ enum ReaderRoute {
     case blog
     case feedsPost
     case blogsPost
+    case wpcomPost
 }
 
 extension ReaderRoute: Route {
@@ -45,6 +46,8 @@ extension ReaderRoute: Route {
             return "/read/feeds/:feed_id/posts/:post_id"
         case .blogsPost:
             return "/read/blogs/:blog_id/posts/:post_id"
+        case .wpcomPost:
+            return "/:post_year/:post_month/:post_day/:post_name"
         }
     }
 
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Utility/Universal Links/Routes+Reader.swift
```diff
@@ -106,6 +109,10 @@ extension ReaderRoute: NavigationAction {
             if let (blogID, postID) = blogAndPostID(from: values) {
                 coordinator.showPost(with: postID, for: blogID, isFeed: false)
             }
+        case .wpcomPost:
+            if let urlString = values["matched-route-url"], let url = URL(string: urlString) {
```
frosty: You can use `MatchedRouteURLComponentKey.url` here, instead of the literal value.

frosty: I wonder if we should also double check that post_year, post_month, post_day are all numbers? 🤔 Otherwise, we have no guarantee that this is definitely the type of URL we want, other than it has 4 path components.

frosty (commented):
This works as expected, just a couple of suggestions in the code :)

frosty (approved):
Looks good! Might be nice to add a unit test for the `isValidWpcomUrl`, but not a deal breaker :)

PR #16327, opened 2021-04-16
User Profile: track events

ScoutHarris:
Ref: #15662

### New Events

Two new events are added to track:
- When the User Profile is shown.
- When a site is viewed from the User Profile.

User profile shown: 
- `Tracked: user_profile_sheet_shown <source: like_notification_list>`. 
- The `source` property is there so that in the future if the User Profile is shown from somewhere else it can be specified.

Site viewed:
- `Tracked: user_profile_sheet_site_shown <>`
- I didn't think any properties where necessary here, but I can add something if we like.

### Changed Event

I added a `source` property to the `reader_blog_previewed` event so that we know it was shown from the User Profile. Examples:
- From user profile: `Tracked: reader_blog_previewed <site: Site Name, source: user_profile, subscription_count: 99>`
- From a reader card: `Tracked: reader_blog_previewed <site: Site Name, source: reader, subscription_count: 99>`

cc @develric 

---
To test:
- Enable the `newLikeNotifications` feature.
- Go to Notifications > Likes and tap a notification.
- Tap a user.
  - Verify `Tracked: user_profile_sheet_shown <source: like_notification_list>` is logged.
- Tap the site.
  -  Verify `Tracked: user_profile_sheet_site_shown <>` is logged.
  - Verify `Tracked: reader_blog_previewed <site: Site Name, source: user_profile, subscription_count: 99>` is logged if the site is internal. (You'll have to change the siteID [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/4a310da51fc2e5fa6b52c2fb12043242c15f6c36/WordPress/Classes/ViewRelated/User%20Profile%20Sheet/UserProfileSheetViewController.swift#L146).)

---
## Regression Notes
N/A. Just adding some tracks.

N/A. Just adding some tracks.

N/A. Just adding some tracks.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -163,8 +163,8 @@ import Foundation
     case categoryFilterDeselected
 
     // User Profile Sheet
-    case userProfileShown
-    case userProfileSiteShown
+    case userProfileSheetShown
+    case userProfileSheetSiteShown
 
     /// A String that represents the event
     var value: String {
@@ -450,10 +450,10 @@ import Foundation
             return "category_filter_deselected"
 
         // User Profile Sheet
-        case .userProfileShown:
-            return "user_profile_shown"
-        case .userProfileSiteShown:
-            return "user_profile_site_shown"
+        case .userProfileSheetShown:
+            return "user_profile_sheet_shown"
+        case .userProfileSheetSiteShown:
+            return "user_profile_sheet_site_shown"
         }
     }
 
--- a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
+++ b/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
@@ -162,6 +162,10 @@ import Foundation
     case categoryFilterSelected
     case categoryFilterDeselected
 
+    // User Profile Sheet
+    case userProfileShown
+    case userProfileSiteShown
+
     /// A String that represents the event
     var value: String {
         switch self {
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
```diff
@@ -444,6 +448,12 @@ import Foundation
             return "category_filter_selected"
         case .categoryFilterDeselected:
             return "category_filter_deselected"
+
+        // User Profile Sheet
+        case .userProfileShown:
+            return "user_profile_shown"
+        case .userProfileSiteShown:
+            return "user_profile_site_shown"
```
aerych: Would it make sense to call these `liker_profile_*` rather than `user_profile_*`?   I'm thinking that someone looking at tracks might at a glance assume that user_profile referred to the logged in user's own profile.

aerych (approved):
Hola @ScoutHarris 👋 
Changes look good. Tested and confirmed the events were dispatched. 
I had one question about the name. (cc @develric) 
Other than that :shipit:

WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift
ScoutHarris: Well, per the User Profile designs (pbArwn-K3-p2), this sheet could be displayed from somewhere other than the liker. So I made it specific to the sheet, not the presenter. That's why I added a source to `user_profile_shown`. I could add a `source` to `user_profile_site_shown` if we like. 

Or, I could change them to `user_profile_sheet_*` if that's better. Which it may be...

aerych: > So I made it specific to the sheet, not the presenter

Oh I see.  Makes sense.

>  I could change them to user_profile_sheet_*

I think I'm a fan :)

ScoutHarris: Can do. Thanks!

ScoutHarris:
Hey @aerych @develric . I updated the event names and the PR description. Thanks!

PR #16328, opened 2021-04-17
Merge 17.1 final into develop

jkmassel:
Merges remaining changes from the `17.1` release into `develop`


fastlane/Deliverfile
```diff
@@ -1,7 +1,7 @@
 require 'dotenv'
 Dotenv.load('~/.wpios-env.default')
 
-screenshots_path "./screenshots/"
+screenshots_path "./fastlane/promo-screenshots/"
```
mokagio, line 4: Promo screenshots 🎉

fastlane/Fastfile
```diff
@@ -235,7 +235,7 @@ import "./ScreenshotFastfile"
     ios_finalize_prechecks(options)
     unless ios_current_branch_is_hotfix
       ios_update_metadata(options)
-      ios_lint_localizations(input_dir: 'WordPress/Resources', allow_retry: true)
+      #ios_lint_localizations(input_dir: 'WordPress/Resources', allow_retry: true)
```
mokagio, line 238: From e7a2760a7ad2162e31d87d0bfb1536f5154db94a, it's clear that this was not an accidental leftover.

What's the plan to fix it? I'm fine to merge with it if it's a "long" plan.

mokagio (approved):
👍 

Approving but holding merge just in case there is a quick fix for the commented `ios_lint_localizations` lane.

fastlane/Fastfile
jkmassel: We need to break a string into multiple in GlotPress, but then we need to wait for it to be translated, so can't be done within the scope of this PR :(

mokagio: Gotcha 👍

PR #16330, opened 2021-04-17
Merge 17.0 final into trunk

jkmassel:
It looks like I missed this one during the `17.0` finalization – this PR merges the changes from the `17.0` series into `trunk`.


mokagio (approved):
I only checked the version `.xcconfig`s given the size. They look good.

mokagio:
This version has already been released to the App Store. Merging it in.

PR #16335, opened 2021-04-19
AB testing: remove stories AB testing

leandroalonso:
This PR removes the stories' AB testing.

### To test

1. Add breakpoints on `ABTest.swift` lines 20 and 28
2. Run the app
3. No breakpoint should be hit
4. Tap the FAB, it should display the post options (including Story as the last item, if applicable)

## Regression Notes
N/a.

N/a.

N/a.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/AB Testing/ABTest.swift
+++ b/WordPress/Classes/Utility/AB Testing/ABTest.swift
@@ -2,7 +2,6 @@ import AutomatticTracks
 
 enum ABTest: String, CaseIterable {
     case unknown = "unknown"
-    case storyFirst = "wpios_create_menu_story_first"
 
     /// Returns a variation for the given experiment
     var variation: Variation {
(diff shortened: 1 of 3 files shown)


momo-ozawa (approved):
LGTM 👍 
Works as described!

yaelirub:
>Tap the FAB, it should display the post options (including Story as the last item, if applicable)

@leandroalonso , if the test shows better results for Stories is at the top why are we reverting ?

leandroalonso:
@yaelirub I'm probably biased. :)

The AB testing implementation's main goal was to test the platform itself, that's why I just removed it (thus "reverting" the order). I think the same was done in Android (cc @renanferrari).

If we want to keep "Stories" as the first option that should be an easy change. (ccing @aforcier)

aforcier:
I don't have a strong opinion on this; it feels like a WP mobile general product question than a stories specific one and probably warrants a bit more consideration, since the results showed more story posts published but at the expense of fewer regular posts.

(Some discussion in pbxNRc-Ku-p2)

PR #16342, opened 2021-04-20
Fix: Undo action restores Gutenberg template content as Classic editor

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/3342
**Related PR:** https://github.com/wordpress-mobile/WordPressKit-iOS/pull/385

## Description
This addresses https://github.com/wordpress-mobile/gutenberg-mobile/issues/3342 by pulling in the changes in the related PR. ~It also addresses an issue I noticed while testing. After restoring a deleted post, the post would be converted to a draft but it remained in the published list. This adds a refresh similar to the way the delete works to make sure the list is refreshed.~

Refreshing the list was separated to https://github.com/wordpress-mobile/WordPress-iOS/pull/16358

## To test:

1. Create a site that generates posts such as Cassel under the Blog category. 
    * This is an important step because for some reason that I wasn't able to track down logging in on a different device restores the content as a Gutenberg Post without then added parameter.
2. Navigate to a Post and ensure it's a Gutenberg Post.
3. Delete it. 
4. Restore it. 
5. Validate that the restored post ~(now in drafts)~ uses the Gutenberg Editor. 

## Regression Notes
This should be the only affected flow.

N/A

Refactoring to establish tests would dramatically increase the size of this effort.

## PR submission checklist:


chipsnyder:
@guarani Would you mind taking a look at this for me? I saw you recently made changes to WPKit but if you have a lot on your plate let me know.

RELEASE-NOTES.txt
```diff
@@ -21,7 +21,7 @@
 * [**] Block Editor: Added Contact Info block to sites on WPcom or with Jetpack version >= 8.5.
 * [**] We updated the app's color scheme with a brighter new blue used throughout. [#16213, #16207]
 * [**] We updated the login prologue with brand new content and graphics. [#16159, #16177, #16185, #16187, #16200, #16217, #16219, #16221, #16222]
-* [**] We updated the app's color scheme with a brighter new blue used throughout. [#16213, #16207] 
+* [**] We updated the app's color scheme with a brighter new blue used throughout. [#16213, #16207]
```
guarani: There's a couple of unexpected changes here to the release notes, are these meant to be here?

RELEASE-NOTES.txt
```diff
@@ -1,6 +1,6 @@
 17.3
 -----
-
+* [**] Fix issue where Undo would sometimes convert content to the the classic editor. [#16342]
```
guarani:
```suggestion
* [**] Fix issue where Undo would sometimes convert content to the classic editor. [#16342]
```

guarani: I wonder if this needs to be more specific since Undo could refer to the block editor undo button. Also, might be worth clarifying that it would "sometimes make posts/pages lose their block formatting", if that makes sense – up to you. Not sure if there's a way of being more specific without getting too technical.
When I viewed the HTML of a post where this happened, it looks like blocks lose their block delimiters, but that's too technical for the release notes.

enejb:
I tested this PR and it worked well for me. 
Thanks for the fixes Chip!

RELEASE-NOTES.txt
chipsnyder: Ah good call this is a result of the text editor I use cleaning up trailing whitespace characters. Not intentional I just missed excluding this line.

RELEASE-NOTES.txt
chipsnyder: >Not sure if there's a way of being more specific without getting too technical.

This was my fear. If I elaborated on the steps it would probably be a bit too technical. I agree that Undo isn't clear enough though. I'll re-word this line.

> When I viewed the HTML of a post where this happened, it looks like blocks lose their block delimiters, but that's too technical for the release notes.

Yeah, the app uses the presence of one formatted block to decide if we should use the classic editor or not.

RELEASE-NOTES.txt
guarani: Gotcha, I hadn't noticed that it was just a whitespace change.

RELEASE-NOTES.txt
guarani: Ok, sounds good!

guarani (approved):
Tested and works for me! Thanks @chipsnyder ❤️

PR #16358, opened 2021-04-22
Fix: Refresh the post list after restoring a post

chipsnyder:
**Fixes** https://github.com/wordpress-mobile/gutenberg-mobile/issues/3342
**Related PR:** https://github.com/wordpress-mobile/WordPressKit-iOS/pull/385

## Description
This addresses an issue I noticed while testing https://github.com/wordpress-mobile/WordPress-iOS/pull/16342. After restoring a deleted post, the post would be converted to a draft but it remained in the published list. This adds a refresh similar to the way the delete works to make sure the list is refreshed.

## To test:

1. Navigate to a Post select a published post
3. Delete it. 
4. Restore it. 
5. Validate that the restored post is now in drafts

## Regression Notes
This should be the only affected flow.

N/A

Refactoring to establish tests would dramatically increase the size of this effort.

## PR submission checklist:


guarani (commented):
I gave this a quick test and it works but seems to result in a misleading UI state. 
1. Delete the post
2. Restore the post
3. Notice that the Undo button to restore the post is still there

Before this change, the restored post was put back in the Published list and the Undo button was no longer visible. Now, the Undo button is still visible even after the user has restored a post.

Not sure if WPAndroid handles this, if so we might be able to copy its behavior.

chipsnyder:
Thanks for testing @guarani we seem to be getting a lot of different results. I'm not experiencing the Undo button remaining after a restore:


https://user-images.githubusercontent.com/3384451/115765886-4b043480-a375-11eb-8415-3e9e4388034c.mov

chipsnyder:
I did notice some weird UI on drafts that I added a patch to handle.

guarani:
I'll give this another test with these changes and report back.

Just want to note that I ran into a crash when trashing and restoring posts and made a separate issue for that since it's not related to the changes here: https://github.com/wordpress-mobile/WordPress-iOS/issues/16359

guarani:
> Just want to note that I ran into a crash when trashing and restoring posts and made a separate issue for that since it's not related to the changes here: #16359

To clarify, that issue is also present on this branch, since I'd first noticed it on the latest commit here 29b65f0604c6987b2eb5a018b258e99e7189cac2 but determined that it's also present on 17.2 beta.

Podfile.lock
```diff
@@ -401,7 +401,7 @@ PODS:
     - WordPressKit (~> 4.18-beta)
     - WordPressShared (~> 1.12-beta)
     - WordPressUI (~> 1.7-beta)
-  - WordPressKit (4.32.0-beta.3):
+  - WordPressKit (4.32.0-beta.4):
```
guarani: I think this change can be reverted since https://github.com/wordpress-mobile/WordPress-iOS/pull/16342 will be updating this.

RELEASE-NOTES.txt
```diff
@@ -1,5 +1,7 @@
 17.3
 -----
+* [**] Fix issue where deleting a post and selecting undo would sometimes convert the content to the classic editor. [#16342]
```
guarani, line 3: This also looks like it doesn't belong

guarani (approved):
Change looks good, just left a few minor comments. Thanks @chipsnyder for fixing this!

Podfile.lock
chipsnyder: Since this was branched off of that branch this change should be resolved as soon as that other one merges.

RELEASE-NOTES.txt
chipsnyder: Same issue as mentioned here https://github.com/wordpress-mobile/WordPress-iOS/pull/16358#discussion_r618729975

Podfile.lock
guarani: Makes sense, thanks!

PR #16360, opened 2021-04-22
Fix UBE tracks custom event format

### LINKED ISSUE #3228: [Analytics] Sessions reporting having unsupported blocks differ a lot between Android and iOS
I took a look at the number of editor sessions that report having unsupported blocks and looks like there's a huge difference between Android and iOS, to the degree that doesn't make sense to me on first look, hinting to a possible bug at reporting.

Here's a short capture of the data on a weekly basis:

Week (2021) | Session counts Android | With unsupported Android | | Session count iOS | With unsupported iOS
-|-|-|-|-|-
March 01 | 833,175 | 27,387||606,674 | 79
February 22 | 795,269 | 26,509||522,920 | 108
February 15 | 768,914 | 27,349||496,394 | 121
February 08 | 762,708 | 27,480||471,946 | 101

The "with unsupported block" sessions on iOS are 2 orders of magnitude lower than Androids, while Android's sessions seem to hover around 3% of sessions having unsupported blocks. Without looking further into the data, the 3% feels quite natural to me, so I wonder if the iOS reports are undercounted.

guarani:
Fixes https://github.com/wordpress-mobile/gutenberg-mobile/issues/3228

The issue here was that events were being rejected by Tracks because they were malformed. The editor_session_start has both a `has_unsupported_blocks` and a `unsupported_blocks` custom event properties. `unsupported_blocks`, which is a list of unsupported blocks, was not being stringified on WPiOS before being sent to the backend, causing these events to be discarded.

## To test

Ensure that `wpios_editor_session_start` are no longer being rejected by the tracks system when a post is opened that contains unsupported blocks.

1. Open a post in the block editor that contains multiple unsupported blocks
2. Check in the Tracks backend that `wpios_editor_session_start` events are being logged in the live view (they should not appear in searches for rejected events)

## Regression Notes

This should only affect the `wpios_editor_session_start` event. If there was a crash in this code, it would prevent users from opening the editor.

I'm relying on the Gutenberg UI tests to make sure the editor still works fine.

I think the Gutenberg UI tests that CircleCI runs are enough to cover this scenario.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
@@ -33,10 +33,14 @@ struct PostEditorAnalyticsSession {
         // Let's make sure to round the value and send an integer for consistency
         let startupTimeNanoseconds = DispatchTime.now().uptimeNanoseconds - startTime
         let startupTimeMilliseconds = Int(Double(startupTimeNanoseconds) / 1_000_000)
-        return [
-            Property.startupTime: startupTimeMilliseconds,
-            Property.unsupportedBlocks: unsupportedBlocks
-        ].merging(commonProperties, uniquingKeysWith: { $1 })
+        var properties: [String: Any] = [ Property.startupTime: startupTimeMilliseconds ]
+
+        // Tracks custom event types can't be arrays so we need to convert this to JSON
+        if let data = try? JSONSerialization.data(withJSONObject: unsupportedBlocks, options: .fragmentsAllowed) {
+            let blocksJSON = String(data: data, encoding: .utf8)
+            properties[Property.unsupportedBlocks] = blocksJSON
+        }
+        return properties.merging(commonProperties, uniquingKeysWith: { $1 })
     }
 
     mutating func `switch`(editor: Editor) {
--- a/WordPress/WordPressTest/Analytics/EditorAnalytics/PostEditorAnalyticsSessionTests.swift
+++ b/WordPress/WordPressTest/Analytics/EditorAnalytics/PostEditorAnalyticsSessionTests.swift
@@ -74,7 +74,8 @@ class PostEditorAnalyticsSessionTests: XCTestCase {
         let tracked = TestAnalyticsTracker.tracked.first
 
         XCTAssertEqual(tracked?.stat, WPAnalyticsStat.editorSessionStart)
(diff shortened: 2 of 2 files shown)


guarani:
There's room for adding functionality to the Tracks library to do this sort of "stringifying" of custom event properties automatically (something that `Automattic-Tracks-Android` does, I think.

fluiddot:
Looks like we should to update the unit tests because one of them is failing, here is some info:
```
[00:04:19]: ▸ PostEditorAnalyticsSessionTests
[00:04:19]: ▸     ✗ testTrackUnsupportedBlocksOnStart, XCTAssertEqual failed: ("nil") is not equal to ("Optional(["unsupported"])")
[00:04:19]: ▸     ✗ testTrackUnsupportedBlocksOnStartWithEmptyList, XCTAssertEqual failed: ("nil") is not equal to ("Optional([])")
[00:04:19]: ▸ WordPressTest.PostEditorAnalyticsSessionTests
[00:04:19]: ▸   testTrackUnsupportedBlocksOnStart, XCTAssertEqual failed: ("nil") is not equal to ("Optional(["unsupported"])")
[00:04:19]: ▸   /Users/distiller/project/WordPress/WordPressTest/Analytics/EditorAnalytics/PostEditorAnalyticsSessionTests.swift:77
```

fluiddot (commented):
I've done some tests and all events were accepted 🎊 , here are the results:

**Post with multiple unsupported blocks:**
```
has_unsupported_blocks: 1
unsupported_blocks: ["jetpack\/gif","syntaxhighlighter\/code","core\/table"]
```

**Post with only one unsupported block:**
```
has_unsupported_blocks: 1
unsupported_blocks: ["jetpack\/gif"]
```

**Post without unsupported blocks:**
```
has_unsupported_blocks: 0
unsupported_blocks: []
```

~~I was wondering whether we should add the `unsupportedBlocks` property in case the post doesn't have unsupported blocks instead of an empty array, wdty?~~

**EDIT:** I noticed that on Android we're sending an empty array when there are no unsupported blocks so for consistency I think we could leave it this way 👍 .

guarani:
Thanks for the review, @fluiddot! I've fixed the failing tests and verified them locally. Re-requesting a review now.

fluiddot (approved):
LGTM 🎊 !

guarani:
Thanks @fluiddot!

PR #16361, opened 2021-04-23
Likes Notifications:  add core data entities for user details.

ScoutHarris:
Ref #15662 

This adds two new entities to store Likes data, and initial classes for each.
- `LikeUser` - for the user info.
- `LikeUserPreferredBlog` - for the user's preferred blog.

Nothing in the app is actually using them yet.

To test:
- Verify the app builds and runs. 
- Verify it is using version 122 of the data model and the new entities are present.
- Check `LikeUser` attributes to be sure they match up with WPKit `RemoteLikeUser` [properties](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/RemoteUser%2BLikes.swift#L3). 
- Check `LikeUserPreferredBlog` attributes to be sure they match up with WPKit `RemoteLikeUserPreferredBlog` [properties](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/develop/WordPressKit/RemoteUser%2BLikes.swift#L29). 
- Verify the relationships between `LikeUser` and `LikeUserPreferredBlog` is 1-to-1, optional from `LikeUser` to `LikeUserPreferredBlog`.

## Regression Notes

PR submission checklist:

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -3,6 +3,15 @@
 This file documents changes in the data model. Please explain any changes to the
 data model as well as any custom migrations.
 
+## WordPress 122
+
+@scoutharris 2021-04-23
+
+- Added new entities:
+- `LikeUser`
+- `LikeUserPreferredBlog`
+- Created one-to-one relationship between `LikeUser` and `LikeUserPreferredBlog`
+
 ## WordPress 121
 
 @twstokes 2021-04-21
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -3,6 +3,15 @@
 This file documents changes in the data model. Please explain any changes to the
 data model as well as any custom migrations.
 
+## WordPress 121
+
+@scoutharris 2021-04-22
+
+- Added new entities:
+  - `LikeUser`
+  - `LikeUserPreferredBlog`
+- Created one-to-one relationship between `LikeUser` and `LikeUserPreferredBlog`
+
 ## WordPress 120
 
 @chipsnyder 2021-04-12
--- a/WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
@@ -1,0 +1,6 @@
+import CoreData
+
+@objc(LikeUser)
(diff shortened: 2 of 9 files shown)


aerych (changes requested):
Hola @ScoutHarris 👋 
Looking good.  :)  I had a couple of questions/suggestions:

Screen Shot 2021-04-23 at 10 23 20 AM
It seems like the `preferredBlog` relationship should be optional since it can be `null` from the endpoint and including a model that was empty wouldn't be too useful. 
The deletion rule for `preferredBlog` should probably be cascade rather than nullify so it can be automatically cleaned up when a `LikeUser` is deleted.

Screen Shot 2021-04-23 at 10 30 48 AM
I wonder if `dateLiked` would be better as a Date in order to order results according to date via a NSFetchRequest's NSPredicate.  Or, since we want to also pass the date as a query param, maybe it would be helpful to have two fields, one that's a String and one that's a Date?

Thoughts?

ScoutHarris:
Hey @aerych .

> It seems like the `preferredBlog` relationship should be optional since it can be `null` from the endpoint

Woops. Thank you!

> The deletion rule for `preferredBlog` should probably be cascade rather than nullify so it can be automatically cleaned up when a `LikeUser` is deleted.

Done.

> I wonder if `dateLiked` would be better as a Date in order to order results according to date via a NSFetchRequest's NSPredicate. Or, since we want to also pass the date as a query param, maybe it would be helpful to have two fields, one that's a String and one that's a Date? 

Per our Slack discussion, I changed `dateLiked` to a Date. If we need the String I'll add it later.

I had to add a new data model version (122), so if you could re-review everything I'd really appreciate it!

ScoutHarris:
> maybe it would be helpful to have two fields, one that's a String and one that's a Date?

@aerych I saw the light. 😄 There is now a `dateLikedString` field.

aerych (approved):
Looks good @ScoutHarris! Thanks for the change :)  
One last thing we might consider is whether dateLiked should be an optional or not.  Seems like a date is always required so unchecking might be fine? What do ya think? 
Otherwise :shipit: 
Screen Shot 2021-04-23 at 3 52 34 PM

ScoutHarris:
> Seems like a date is always required so unchecking might be fine? What do ya think?

Done. Thanks!

PR #16365, opened 2021-04-23
Improves the coloring of the search fields throughout the App.

diegoreymendez:
This change was spawned by a chat with @osullivanchris.

This PR improves the coloring of the search fields in dark mode throughout the App.  It changes the background to match the navigation bar color, and it changes the search-query-tex-color to be the default (black for light mode, and white for dark mode).

## Search Field Placeholder Text Coloring

| Before | After |
| --- | --- |
|  |  |
|  |  |

 |

## Search Field Input Text Coloring

| Before | After |
| --- | --- |
|  |  |
|  |  |

## To test:

Test in light and dark mode.

1. Go to any place where you expect a search field.
2. Make sure the color of the search field matches the color of the navigation bar.

## Regression Notes

Any search field that doesn't have a navigation bar.  Below is a list of places where the search bar is configured using the changes in this PR.

I tested each of those fields in-app.

✅ AztecPostViewController.swift: this isn't easy to access because it's a search bar for the media picker in the classic editor.  You need to install the classic editor plugin in your site, create a post, and then edit it using WPiOS and add an image from the  WordPress media library.
✅ BlogListViewController.m: the behavior here seems to be what I'd expect, but I do think it's worth mentioning when you give focus to the search field, the background changes color in a slightly unexpected way (code wise).
✅ BlogSelectorViewController.m: this is the site picker.  There's a divider between the VC title and the search bar that I wasn't able to remove without risking introducing regressions.  This was discussed with both @frosty and @osullivanchris and we agreed it could be left like this for now.
✅ SiteTagsViewController.swift: My Site > Site Settings > Tags.  You need to have at least 1 tag created to see the search field.  There's some divider weirdness worth mentioning here, although it wasn't introduced by my changes.
✅ LanguageSelectorViewController.swift: you need a wp.com site to find this one.  My Site > Site Settings > Language.  Same background changing effect as reported for the blog list.
✅ SiteIconPickerPresenter.swift: My Site > Tap Site Icon > Change Site Icon > "WordPress Media" album
✅ TenorPicker.swift: New post > Add media > Add image > Free GIF Library
✅ StockPhotosPicker.swift: New post > Add media > Add image > Free Photo Library
✅ MediaLibraryViewController.swift: My Site > Media
✅ ParentPageSettingsViewController.swift: My Site > Site Pages > Tap on "..." for an existing page > Set Parent
✅ PluginDirectoryViewController.swift: My Site > Plugins (you'll need a site with plugin support)
✅ AbstractPostListViewController.swift: posts and pages list for a site.
✅ PostSettingsViewController.m: Edit a post > Tap "..." on the top right corner > Post Settings > Set Featured Image > "WordPress Media" album.
❌  PostGeolocationViewController.swift: The search field and the coloring here looks quite broken, but this wasn't introduced by this PR, so it's not actually a regression.
✅ ReaderFollowedSitesViewController.swift: Reader > Tap the gear in the top right corner  > "Followed Sites" tab
✅ ReaderSearchViewController.swift: Reader > Tap the magnifying lens on the top right corner
✅ ThemeBrowserViewController.swift: My Site > Themes
✅ TimeZoneSelectorViewController.swift: My Site > Site Settings > Timezone

None because this change is not easy to automatically test against.

## PR submission checklist:


osullivanchris (changes requested):
👋  Hey @diegoreymendez thanks for setting up the PR!

In the after, the text doesn't look white. Didn't we say we'd fix it to the higher contrast colour? Maybe that conversation was after this PR was made.

to do a good review, it would also be good to either get my hands on a build, or see more screenshots of light mode, and other locations where the search bar will change.

diegoreymendez:
@osullivanchris - I've updated the screenshots to better reflect the changes that were done.

- The placeholder text color is now the OS default.  There seems to be a higher contrast in dark mode, but in light mode I'm not sure it's changed much.
- The input text color is now the OS default.  The contrast is higher in both modes since the input color is now black and white.
- The magnifying lens icon is custom, so we have to apply the tint manually to try and match the text color.  In light mode the lens looks slightly darked than the text because of a coloring issue with our default colors vs Apple's.  If it's not too big of an issue I'd suggest we talk about this separately via DM.

diegoreymendez:
Ready for review.  The failing CircleCI test is unrelated and should be fixed soon.

osullivanchris:
Looks great so far, thanks for all the extra effort and discussions. The only details I'm noticing is the text looks a little small. 

I believe it should be Body 17pt https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/

Not sure if we are overwriting the default type to customise and make it smaller, or if we are not using the default type at all.

WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+Search.swift
```diff
@@ -24,25 +24,25 @@ extension WPStyleGuide {
 
         // We have to manually tint these images, as we want them
         // a different color from the search bar's cursor (which uses `tintColor`)
-        let cancelImage = UIImage(named: "icon-clear-searchfield")?.imageWithTintColor(.neutral(.shade30))
-        let searchImage = UIImage(named: "icon-post-list-search")?.imageWithTintColor(.neutral(.shade30))
+        let cancelImage = UIImage(named: "icon-clear-searchfield")?.withTintColor(.secondaryLabel)
```
frosty: I wonder if we should introduce a new named color reference in UIColor+MurielColors called something like `searchBarItem`?

frosty (approved):
Changes look good! I ran through the various use cases you mentioned and didn't spot anything other than the existing issues you already highlighted. Just one minor code query, but otherwise :shipit:

diegoreymendez:
@osullivanchris - Adjusted the text size for the placeholder, and the input query in search fields as per your request.  It's now set to font style `body`.

New screenshots below.

| Light | Dark |
| --- | --- |
|  |  |
|  |  |

WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+Search.swift
diegoreymendez: I've addressed this and created some new colors for search field customization.

frosty (approved):

diegoreymendez:
As reported by @frosty and reviewed by @osullivanchris, I've also fixed some styling issues in a search bar in the reader section of the app.

Before:

Screenshot 2021-04-27 at 16 40 13

After:

| Placeholder | Input Query |
| --- | --- |
|  |  |

osullivanchris:
@diegoreymendez its mostly looking great but I have one more change sorry 🙈 

The text size is now correct. But the height of the input field seems to have increased as well. Maybe it is related. If you look at the screenshots at the top of this PR, the input field is approx 32pt in height. That matches up with the components I have, and with other Apple apps. In the latest screenshots and build I just looked at, the input field looks approx 44pt height.

I wonder if the text has made that happen, or there's some other issue. The original height, with the larger text, would have be perfect.

osullivanchris (changes requested):
forgot to use the 'add review' feature. Only change requested is the input field height as described above.

diegoreymendez:
@osullivanchris - Can I ask you to try again with build 46702?

osullivanchris (approved):
🙌  Yesss! All issues resolved. Looking great. Thanks for all the extra effort to get it tuned up!!

diegoreymendez:
Thanks folks!  Much appreciated!

PR #16383, opened 2021-04-26
Post likes: add to Core Data when fetched

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/issues/15662
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/391

This adds a new `PostService+Likes` class that:
- Fetches likes for a Post and returns an array of `LikeUser`s.
  - The new method is basically a Swift version of the [original](https://github.com/wordpress-mobile/WordPress-iOS/blob/8de804c16e55f31cc03da52083bbb7e5b8110983/WordPress/Classes/Services/PostService.m#L688) `PostService` ObjC method, which will be removed later.
- Purges any existing users before creating new ones.

Notes:
- The likes list UI still displays the old `RemoteUser` list.
- The user purging (`deleteExistingUsersFor:postID`) is not filtered. I need to add `postID` to the WPKit response parsing.
- This change is only for Post likes. Comment likes coming soon.

To test:

`LikesListController` still uses the old fetching method to display the likers. I've added a temporary call to the new fetch method that logs out the `LikeUser` array. 

- Tap on a Post like notification.
- Verify the user array appears on the console (prefaced by `🔴 likeUsers: `. The users should match the displayed users.

## Regression Notes
N/A. Feature incomplete.

N/A. Feature incomplete.

N/A. Feature incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
@@ -3,33 +3,4 @@ import CoreData
 @objc(LikeUser)
 public class LikeUser: NSManagedObject {
 
-    static func createUserFrom(remoteUser: RemoteLikeUser, context: NSManagedObjectContext) -> LikeUser? {
-
-        guard let likeUser = NSEntityDescription.insertNewObject(forEntityName: "LikeUser", into: context) as? LikeUser else {
-            return nil
-        }
-
-        likeUser.userID = remoteUser.userID.int64Value
-        likeUser.username = remoteUser.username
-        likeUser.displayName = remoteUser.displayName
-        likeUser.primaryBlogID = remoteUser.primaryBlogID.int64Value
-        likeUser.avatarUrl = remoteUser.avatarURL
-        likeUser.bio = remoteUser.bio ?? ""
-        likeUser.dateLikesString = remoteUser.dateLiked ?? ""
-        likeUser.dateLiked = DateUtils.date(fromISOString: likeUser.dateLikesString)
-        likeUser.preferredBlog = createPreferredBlogFrom(remotePreferredBlog: remoteUser.preferredBlog, forUser: likeUser, context: context)
-
-        return likeUser
-    }
-
-    static func createPreferredBlogFrom(remotePreferredBlog: RemoteLikeUserPreferredBlog?,
-                                        forUser user: LikeUser,
-                                        context: NSManagedObjectContext) -> LikeUserPreferredBlog? {
-
-        guard let remotePreferredBlog = remotePreferredBlog else {
-            return nil
-        }
-
-        return LikeUserPreferredBlog.createBlogFrom(remoteBlog: remotePreferredBlog, forUser: user, context: context)
-    }
 }
--- a/WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
@@ -3,4 +3,33 @@ import CoreData
 @objc(LikeUser)
 public class LikeUser: NSManagedObject {
 
+    static func createUserFrom(remoteUser: RemoteLikeUser, context: NSManagedObjectContext) -> LikeUser? {
+
+        guard let likeUser = NSEntityDescription.insertNewObject(forEntityName: "LikeUser", into: context) as? LikeUser else {
+            return nil
+        }
+
+        likeUser.userID = remoteUser.userID.int64Value
+        likeUser.username = remoteUser.username
+        likeUser.displayName = remoteUser.displayName
+        likeUser.primaryBlogID = remoteUser.primaryBlogID.int64Value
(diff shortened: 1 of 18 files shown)


WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
```diff
@@ -3,4 +3,33 @@ import CoreData
 @objc(LikeUser)
 public class LikeUser: NSManagedObject {
 
+    static func createUserFrom(remoteUser: RemoteLikeUser, context: NSManagedObjectContext) -> LikeUser? {
```
aerych: I'm not sure how much of a concern this should be. By adding these factory methods to the managed object we're coupling our model layer to our network remotes  via a type dependency, rather than letting the service layer act as an intermediary.  
If we want to keep with our service pattern we might consider moving the model factory methods into a service helper that could be reused between the post and comments services. 
On the other hand, I'm not sure how close we're sticking to that pattern/architecture these days. 
Thoughts?

WordPress/Classes/Services/PostService+Likes.swift
```diff
@@ -0,0 +1,68 @@
...
+
+            self.deleteExistingUsersFor(postID: postID)
+
+            guard let remoteLikeUsers = remoteLikeUsers,
+                  !remoteLikeUsers.isEmpty else {
+                success(nil)
+                return
+            }
+
+            success(self.createNewUsers(from: remoteLikeUsers))
```
aerych: A core data best practice would be to handle the deletion and creation of new assets in the background rather than on the main thread. `ContextManager.shared.newDerivedContext`, `context.performBlock` and `ContextManager.shared.saveContext:withCompletionBlock:` would be things to look at.  Also `DispatchQueue.main.async` for calling the success and failure blocks back on the main thread.
The service itself could be created with a derived context (I think it's using the mainContext now), which might simplify things, but it still might make sense to call the success/failure blocks on the main thread from within the service.

WordPress/Classes/Services/PostService+Likes.swift
```diff
@@ -0,0 +1,68 @@
...
+     
+     @param postID  The ID of the post to fetch likes for
+     @param siteID  The ID of the site that contains the post
+     @param success A success block
+     @param failure A failure block
+     */
+    func getLikesFor(postID: NSNumber,
+                     siteID: NSNumber,
+                     success: @escaping (([LikeUser]?) -> Void),
+                     failure: ((Error?) -> Void)? = nil) {
```
aerych: Curious why a default nil for the failure block.  Seems like we'd always want to handle failure?  Maybe this is related to future plans?
Also, should this be flagged `@escaping` as well?

aerych (changes requested):
Howdy @ScoutHarris 👋 
I had a few questions and suggestion from a quick review.  Lemme know what you think! Happy to go into more detail. :)

WordPress/Classes/Services/PostService+Likes.swift
ScoutHarris: > Curious why a default nil for the failure block. Seems like we'd always want to handle failure? Maybe this is related to future plans?

Yep. I haven't handled failures yet, so I'm not sure if this should be required or not. 

> should this be flagged @escaping as well?

Ah, I did try, but Swifty gets all complainy.

Screen Shot 2021-04-26 at 4 58 04 PM

WordPress/Classes/Models/Notifications/Likes/LikeUser+CoreDataClass.swift
ScoutHarris: I put these methods in the model classes so they could be accessed from both services. However, I didn't realize that was an issue. I have no problem moving them to a service helper.

WordPress/Classes/Services/PostService+Likes.swift
ScoutHarris: I'll look into it and give it another go. Thank you!!

WordPress/Classes/Services/PostService+Likes.swift
ScoutHarris: Nevermind. I went ahead and changed it. 😄

ScoutHarris:
Hey @aerych .  Thanks for the feedback/suggestions!

I moved the create methods into a new `LikeUserHelper` class. And I updated `PostService+Likes` to use a derived context when adding & removing users.

> The service itself could be created with a derived context (I think it's using the mainContext now), which might simplify things, but it still might make sense to call the success/failure blocks on the main thread from within the service.

I went ahead and left the service context as-is since that seems to be the "standard" (and expected?). I explicitly created a derived context when needed. 

See what you think now.

aerych (approved):
Shiny! Thanks for the change. :)  
:shipit: with great 🚢 ing! 

Oh! I just saw CI was complaining about `Build Jetpack (without testing) `.  We need to manually include the new file in the Jetpack build target.  I'd forgotten about this (and am very glad to see we have some tooling for it).
After that ship away!

PR #16426, opened 2021-05-03
[RNMobile] Add "Set as Featured" Button to Image Block

SiobhyB:
Paritial fix for: wordpress-mobile/gutenberg-mobile#1011

`gutenberg`: https://github.com/WordPress/gutenberg/pull/31415
`gutenberg-mobile`: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3450

The main PR for the branch this individual PR will be merged into, `add/featured-functionality-to-image-block-ios`, can be found here: https://github.com/wordpress-mobile/WordPress-iOS/pull/16427

## Description

This PR will add a **Set as Featured** button to the image block's settings on iOS devices, with the purpose being to make it simpler for users to set a featured image within the post's editor. Users will also be able to **Remove as Featured** directly from the block's setting.

### Main Test Flows

[The Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31415) can be referred to as the central, most up-to-date repository for the main test flows. In addition, we should verify that the Tracks events edited/introduced with this PR work as expected.

### Tracks Verification

1. After following the steps to set and remove an image as featured directly from the image block, navigate to the internal Tracks tool and choose the **Live View** option from the menu.
2. Enter `wpios_editor_post_featured_image_changed` into the tool's **Events** field and then the username associated with the account you tested with in the **Username** field. Click **Search**.
3. Verify that the event was logged correctly. Note, there is usually a delay of ~5 minutes before an event is logged in this tool.
4. Click **Props** and scroll down to the **eventprops** section. The following props should be logged, depending on your actions:
-- **action** should either log `removed` or `added`, depending on whether you have removed or set a featured image.
-- **via** should log `gutenberg` when you set/remove a featured image via the image block or `settings` when this is done via Post Settings. 

## Regression Notes

As this PR adds another flow for users to set/remove featured images, there is some potential for it to unintentionally have a negative impact on the existing flow (found via Post Settings).

Another unintended area of impact is the `wpios_editor_post_featured_image_changed` Track event. This event currently only fires when a featured image is changed via Post Settings. With this PR, it should also be fired when a featured image is changed via the image block, with a `gutenberg` property to differentiate it.

The steps taken to test whether the intended flow works as expected can be found in [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31415) with additional steps for verifying the Tracks event fires as expected listed above.

N/A

PR submission checklist:


SiobhyB:
👋 @guarani, I was wondering whether you may have any availability to review this PR this coming week? 

The code is ready for review, but reliant on [the Android PR](https://github.com/WordPress/gutenberg/pull/31705) being merged first, which is why the chain of PRs are still set as drafts. For that reason, [the Gutenberg PR](https://github.com/WordPress/gutenberg/pull/31415) still includes some files that will eventually be merged with that Android one. I've listed the ones that are specific to iOS (and will therefore need to be reviewed as part of this iOS chain) in the PR's description.

As the Android PR is still pending a final review (and I'm not sure when that will happen), this isn't urgent and there's no worries if you're swamped. I just thought I'd check in to see if you might have availability for reviewing some time this week, as you helped me a lot with the previous iOS one. :)

guarani:
👋 @SiobhyB, sure, feel free to add me as a reviewer!

SiobhyB:
Thank you so much, I've assigned you now. 🙇‍♀️

WordPress/Classes/ViewRelated/Gutenberg/GutenbergFeaturedImageHelper.swift
```diff
@@ -0,0 +1,37 @@
...
+        self.gutenberg = gutenberg
+        super.init()
+    }
+
+    func setFeaturedImage(mediaID: Int32) {
+        let media = Media.existingMediaWith(mediaID: NSNumber(value: mediaID), inBlog: post.blog)
+        post.featuredImage = media
+
+        if mediaID == mediaIdNoFeaturedImageSet {
+            gutenberg.showNotice(NSLocalizedString("Removed as Featured Image", comment: "Notice confirming that an image has been remove as the post's featured image."))
```
guarani:
```suggestion
            gutenberg.showNotice(NSLocalizedString("Removed as Featured Image", comment: "Notice confirming that an image has been removed as the post's featured image."))
```

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -648,6 +653,49 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
+        let alertController = UIAlertController(title: NSLocalizedString("Replace?", comment: "Title message on dialog that prompts user to confirm or cancel the replacement of a featured image."),
+                                                message: NSLocalizedString("You already have a featured image set. Do you want to replace it?", comment: "Main message on dialog that prompts user to confirm or cancel the replacement of a featured image."),
+                                                preferredStyle: .actionSheet)
+
+        let replaceAction = UIAlertAction(title: NSLocalizedString("Replace", comment: "Button to confirm the replacement of a featured image."), style: .default) { (action) in
+            self.featuredImageHelper.setFeaturedImage(mediaID: mediaID)
+        }
+
+        alertController.addAction(replaceAction)
+        alertController.addCancelActionWithTitle(NSLocalizedString("Replace", comment: "Button to cancel the replacement of a featured image."))
```
guarani:
```suggestion
        alertController.addCancelActionWithTitle(NSLocalizedString("Cancel", comment: "Button to cancel the replacement of a featured image."))
```

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -648,6 +653,49 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
         mediaInserterHelper.cancelUploadOf(media: media)
     }
 
+    func gutenbergDidRequestToSetFeaturedImage(for mediaID: Int32) {
+        // dismiss current controller in order to allow the new alert controller to be presented
+        presentedViewController?.dismiss(animated: false, completion: nil)
```
guarani: Is this to remove any view controller that might be currently presented, a sort of catch-all? If so, the only thing that I'm concerned about is that its dismissing without an animation which could be jarring to users. 

Would putting all the action sheet code inside the completion handler above work? I don't recall if it's called even if there's nothing to dismiss though, but it might be worth checking if you haven't already.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -648,6 +653,49 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
+
+        let featuredImageId = post.featuredImage?.mediaID
+        let mediaIdNoFeaturedImageSet = 0
```
guarani: This seems to be declared in two places ([this](https://github.com/wordpress-mobile/WordPress-iOS/pull/16426/files#diff-1f88390d90a5eb0598cd4101f287ac4468bfe3229fa20eb41e78d386efc6fa28R8) being the other).
Would it make sense to put this in a static constant inside `GutenbergFeaturedImageHelper`, like [this](https://github.com/wordpress-mobile/WordPress-iOS/blob/6bd1fbfeae283cc43e04acbf40386b87051fe4aa/WordPress/Classes/ViewRelated/User%20Profile%20Sheet/UserProfileSiteCell.swift#L11)?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -648,6 +653,49 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
+
+        if featuredImageId != nil && mediaID == featuredImageId as! Int32 {
+            // nothing special to do, trying to set the image that's already set as featured
+            return
+        }
```
guarani: If the goal here is to exit the function if `featuredImageId` is equal to `mediaID`, then you could consider using Swift's guard-statement. It's often used to abort a function if certain conditions are not met:
```
        guard featuredImageId as? Int32 != mediaID else {
            return
        }
```

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -648,6 +653,49 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
+
+        if mediaID == mediaIdNoFeaturedImageSet {
+            // user tries to clear the featured image setting
+            featuredImageHelper.setFeaturedImage(mediaID: mediaID)
+            return
+        }
```
guarani: Again you could consider using the guard-statement. Note that the condition is negated when using it (i.e. only continue if the media ID _is not_ zero, otherwise do some clean up and exit):
```
        guard mediaID != mediaIdNoFeaturedImageSet else {
            // user tries to clear the featured image setting
            featuredImageHelper.setFeaturedImage(mediaID: mediaID)
            return
        }
```

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -648,6 +653,49 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
...
+
+        if featuredImageId == nil {
+            // current featured image is not set so, go ahead and set it to the provided one
+            featuredImageHelper.setFeaturedImage(mediaID: mediaID)
+            return
+        }
```
guarani: Here a guard-statement could be used instead, since we're again aborting if conditions are not met.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergFeaturedImageHelper.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import Gutenberg
+
+class GutenbergFeaturedImageHelper: NSObject {
+    fileprivate let post: AbstractPost
+    fileprivate let gutenberg: Gutenberg
+
+    let mediaIdNoFeaturedImageSet = 0
```
guarani: As noted in other comments below, this could be a static let instead, allowing you to define it once here and easily use it in other files.

guarani (commented):
Looking great! Your Swift is coming along nicely :D I left a few comments, but for the most part it's working well. Feel free to re-request a review when this is ready again.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
SiobhyB: Thanks for spotting this! I've gone ahead to apply the suggestion.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
SiobhyB: @guarani, the bottom sheet modal (containing the image block's settings) is currently dismissed in the JS [here](https://github.com/WordPress/gutenberg/blob/53013060c4bbda4b79bff116f80a226322de47fd/packages/block-library/src/image/edit.native.js#L690) as soon as the user taps the `Set/Remove as Featured` button. But that doesn't seem to be being automatically "communicated" back to the iOS side, which is what led me adding this line of code. If I remove that line, the replace functionality fails with the following error:

`Attempt to present <UIAlertController: 0x7ff44aa16400> on <WordPress.AztecNavigationController: 0x7ff44a068200> (from <WordPress.GutenbergViewController: 0x7ff44a808e00>) which is already presenting <RCTModalHostViewController: 0x7ff437eeb340>.`

I found the open issue you created at https://github.com/WordPress/gutenberg/issues/29672 that seems to explain the reason for this behaviour but wasn't certain of the correct way to go about addressing this in my case, where the a new dialog is generated from the iOS side. Does the approach I've taken make more sense to you now with that extra context or is there something else that should be done here?

> Would putting all the action sheet code inside the completion handler above work? I don't recall if it's called even if there's nothing to dismiss though, but it might be worth checking if you haven't already.

Just to make sure I understand: Are you imagining something like this?

`presentedViewController?.dismiss(animated: false, completion: {self.showAlertForReplacingFeaturedImage(mediaID: mediaID)})
`

If so, I tested adding that at the bottom of the `gutenbergDidRequestToSetFeaturedImage` function and it works as expected. :) Would this be a better approach to take, as there will always be the bottom sheet that needs to be dismissed? 

Is there a chance that the way bottom sheets are dismissed may change in the future and negate the need for this extra dismissal in iOS (therefore break the functionality if we're relying on the completion handler)?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
SiobhyB: Ah, that makes a lot of sense, I can see how that would be better practice! I've gone ahead to make that change in https://github.com/wordpress-mobile/WordPress-iOS/pull/16426/commits/9d1aaf599e76370edcb0fedcc3d1077e9e8ee590. 🙇‍♀️

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
SiobhyB: Thank you for this! I've gone ahead to refactor the if statements to use guard in https://github.com/wordpress-mobile/WordPress-iOS/pull/16426/commits/25da33ea7dd83766704ec42ef6d067e554a86be0 and will know to consider using guard if I do any further iOS work in the future. :D

SiobhyB:
Thank you so much for this, Paul. 🙇‍♀️ I've gone ahead to implement the changes you've suggested and had a question on one of the suggestions [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/16426#discussion_r644000484).

guarani:
Just a heads-up that the unit tests seem to be failing and I'm seeing an error when the block editor opens. Once this is done I can test the Tracks events which I'd forgotten to do in the first review.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
guarani: > But that doesn't seem to be being automatically "communicated" back to the iOS side, which is what led me adding this line of code. If I remove that line, the replace functionality fails with the following error:

True, this sounds familiar and thanks for highlighting that existing issue. I was thinking something along the lines of this:

```
    func gutenbergDidRequestToSetFeaturedImage(for mediaID: Int32) {

        let presentAlert = { [weak self] in
            guard let `self` = self else { return }

            // ... the various guard-statements go here ...

            // ask the user to confirm changing the featured image since there's already one set
            self.showAlertForReplacingFeaturedImage(mediaID: mediaID)
        }

        if presentedViewController != nil {
            dismiss(animated: true, completion: presentAlert)
        } else {
            presentAlert()
        }
    }
```

I think this may work now and also if the bottom sheet behavior is fixed in the future, wdyt?

SiobhyB: Thank you so much for sharing that with me, I can see how it's a more eloquent solution and have gone ahead to implement it in 89494c318e69c07418b2f69781e2686ce2fc3e72.

Just a note: I kept `animated: false` as setting this value to `true` caused a crash and the animation is still handled on the JS side when the BottomSheet's dismissed.

Let me know if you spot any further issues here!

SiobhyB:
@guarani, sorry for missing that error. 🤦‍♀️   

I've double-checked and everything should work without error with the latest changes on this branch now.

guarani:
> ### Tracks Verification

Tested 287307b08c54036db7bf2c0babe4ee2beeac74cf:

- [x] Tracks events logged correctly when adding and removing a feature image via the Image block
- [x] Tracks events logged correctly when adding and removing a feature image via Post Settings

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
guarani: Looks good! Testing now and things seem to work well 👍

SiobhyB:
@guarani, this is ready for the next review when you have a chance. :) I've updated all the PRs in this chain so that they're up-to-date following HACK week and my recent AFK. The most recent changes also include [a fix for the iPad-specific crash you surfaced](https://github.com/wordpress-mobile/WordPress-iOS/pull/16426/commits/998e411936f10df080e335a5b3d7e9fd19f58b52).

SiobhyB:
Highlighting that I found an issue where [the button breaks the editor's undo/redo functionality](https://github.com/WordPress/gutenberg/issues/32937), which I'll work on in a separate issue. I'd like to propose merging this issue into the main feature branch (if no other issues are found in this PR chain) irrespective of that particular known issue, as it'll be worked on separately to this.

guarani (commented):
I tested this on an iPad and the replace action sheet is being shown right at the top of the screen which looks incorrect.


https://user-images.githubusercontent.com/1898325/123307750-dd998f00-d4f0-11eb-83c3-f44f328db66b.MP4

SiobhyB:
Thanks @guarani, I've gone ahead to make some changes in https://github.com/wordpress-mobile/WordPress-iOS/pull/16426/commits/8f473649cfb790a736257fb742e8bfea7c31adc5 so that the dialog is always centred in the screen. I also removed the arrow so that the message is displayed (rather than only the title, which seems to happen when an arrow is specified), which I think looks a bit better in this case:

Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-06-24 at 21 48 22

Let me know what you think!

guarani (approved):
Code looks great and works as expected! 💯 
Approved via https://github.com/WordPress/gutenberg/pull/31415#pullrequestreview-692255299
Thanks @SiobhyB!

Issue #16385, opened 2021-04-26
Larger Text: Posts and Pages screen don't look great 


startuptester:
There's overlap between UI elements and it's crowded

![img_2099-1](https://user-images.githubusercontent.com/66104034/116154352-cc293780-a6a5-11eb-8643-adf8a5f039b5.png)
![img_2109-1](https://user-images.githubusercontent.com/66104034/116154358-cdf2fb00-a6a5-11eb-9193-09e84f574464.png)

### Steps to reproduce the behavior
In settings, enable Larger Text and choose a large setting
View Posts and Pages

startuptester:
Adding @osullivanchris' sketch: 

![image](https://user-images.githubusercontent.com/66104034/116157240-de0cd980-a6a9-11eb-9a85-9a76b12ac42d.png)

momo-ozawa:
Closed via #16437

PR #16437, opened 2021-05-05
Improved large text support in the pages list.

diegoreymendez:
Addresses the issue reported in #16385 for the pages list.

## To test:

Test in both iPhone and iPad.

1. Go to the pages list.
2. Rotate to portrait or landscape, make sure the list of pages looks good.
3. Go to iOS > Settings > Accessibility > Larger Text > Select a really large font size
4. Go back to the App and check that the pages list still looks good.

## Regression Notes

None that I can think of.

Nada.

UI changes for large text support.  This isn't testable automatically.

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
+++ b/WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
@@ -1,5 +1,12 @@
 import Foundation
 
+/// `UIImageView`s without height and width constraints are automatically resized to their intrinsic content size when
+/// an image is set.  This subclass exists to allow the creation of image views that don't have size constraints and that
+/// are NOT resized when their image is set.
+///
+/// This is useful to let the dimensions of the `UIImageView` be defined by neighbor views.  A good example of this is when you want
+/// the image view to have the same height as a neighbor text field.
+///
 class FixedSizeImageView: UIImageView {
     override var intrinsicContentSize: CGSize {
         return .zero
--- a/WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
+++ b/WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+class FixedSizeImageView: UIImageView {
+    override var intrinsicContentSize: CGSize {
+        return .zero
+    }
+}
--- a/WordPress/Classes/ViewRelated/Pages/PageListTableViewCell.xib
+++ b/WordPress/Classes/ViewRelated/Pages/PageListTableViewCell.xib
@@ -22,30 +22,30 @@
                 <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
                 <subviews>
                     <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="6l3-mv-Tbu" userLabel="Labels">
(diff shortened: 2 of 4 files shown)


WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+class FixedSizeImageView: UIImageView {
```
frosty, line 10: Could we add a doc comment to explain what this is for and why it exists?

frosty (commented):
Changes look good, and this is a big improvement over what we had before!

The one oddity I noticed, which we discussed on Slack, is how the Homepage label doesn't change size in portrait on an iPhone:



I wonder if we could either just hide the type label if the content size is an accessibility size – something like this:

```
    [[NSNotificationCenter defaultCenter]
      addObserver:self
         selector:@selector(contentSizeCategoryDidChange)
             name:UIContentSizeCategoryDidChangeNotification
           object:nil];
}

- (void)contentSizeCategoryDidChange
{
    NSArray *accessibilitySizes = @[
                              UIContentSizeCategoryAccessibilityMedium,
                              UIContentSizeCategoryAccessibilityLarge,
                              UIContentSizeCategoryAccessibilityExtraLarge,
                              UIContentSizeCategoryAccessibilityExtraExtraLarge,
                              UIContentSizeCategoryAccessibilityExtraExtraExtraLarge
    ];

    BOOL shouldHide = [accessibilitySizes containsObject:self.traitCollection.preferredContentSizeCategory];
    [self.typeLabel setHidden:shouldHide];
}
```

or put the labels in a stack view and flip to vertical layout if we're using an accessibility size, so the type label and its icon get moved to the row below. What do you think?

WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
diegoreymendez: Yeah, I actually had that in mind when I created the class, but somehow forgot to do it.  Thanks!

diegoreymendez:
Hey @frosty , I tried adding a stack view as suggested but it's clipping the text and I'm not sure why.

Regarding the suggestion to hide the label if the text size is an accessibility size, I feel like that logic would hide the label in a lot more scenarios than the one reported in this case.  For example on an iPad it may not be necessary to hide it, even if you're seeing large text.

Additionally I'm not sure that using the notification in the sample code would be enough to properly hide and show the label.  I believe we'll need to also check when the VC is shown, and possibly when the app comes back to foreground.

Since this PR already provides an overall improvement that tackles (IMHO) the most important issues, would you mind if we moved ahead to merge?

If you agree, I will open another issue for the label so that we can explore it separately.

frosty:
Sure, let's go ahead and merge :) Could you still add a comment to the `FixedSizeImageView` class?

WordPress/Classes/ViewRelated/Views/FixedSizeImageView.swift
diegoreymendez: Added

diegoreymendez:
Just added the documentation, ready for another look.

diegoreymendez:
Also opened a GH issue for the broken label: https://github.com/wordpress-mobile/WordPress-iOS/issues/16473

frosty (approved):

PR #16438, opened 2021-05-06
Reader deep links: Automatically scroll to comment if an anchor is present in URL

frosty:
In the previous release, we added support for deep linking URLs such as `mysite.wordpress.com/2021/05/06/some-post` to the Reader. However, if these links included a comment anchor (`#comment-50547`), we didn't attempt to display the selected comment, which could cause problems if you were trying to view a comment link!

This PR fixes that by attempting to automatically present the comments view for a post and navigate to the specified comment if an anchor is present in the URL.

I know that the code to perform the final navigation is a bit on the hacky side, but I couldn't currently find a nicer way to do this. It seemed to work just fine in all my testing, though.

**To test**

- Build and run
- Find a comment on the web (that the app will be able to access), and copy its link. You can then launch it as a deeplink using the `xcrun` command. Here's the link I was using, which you can use too:

```
xcrun simctl openurl booted https://mycoastaladventures.wordpress.com/2014/07/24/dreaming-of-elsewhere/\#comment-10
```
- You should be launched into the comment view for that post and the "Great photographs" post from "markjohnson" should be scrolled to the top.
- Check a couple of different posts to ensure you're taken to the right comment.
- Also test deep links _without_ a comment fragment to ensure you're just taken to the post itself and not the comments view.
- Finally, check navigating to posts normally (manually) in the Reader and their comment views and ensure nothing automated happens.

## Regression Notes

I think these changes should be limited to only Reader views instantiated with a post URL that contains a comment anchor.

I tested normal Reader navigation, both manually and via deep links.

I added a test to check the code that extracts comment IDs from post URLs.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
@@ -22,6 +22,19 @@ class ReaderDetailCoordinator {
     /// A post URL to be loaded and be displayed
     var postURL: URL?
 
+    /// A comment ID used to navigate to a comment
+    var commentID: Int? {
+        // Comment fragments have the form #comment-50484
+        // If one is present, we'll extract the ID and return it.
+        if let fragment = postURL?.fragment,
+           fragment.hasPrefix("comment-"),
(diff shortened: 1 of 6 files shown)


diegoreymendez (approved):
Works as expected. :shipit:

PR #16443, opened 2021-05-07
Prevent freezing when the app becomes active

koke:
Every time the app becomes active, it calls `GuttenbergSettings.performGutenbergPhase2MigrationIfNeeded()`. One of the things this method does is go through all sites and check if there's any with Aztec as its editor.

I've been noticing the app randomly freezing when becoming active and I think this is the cause:
- `GutenbergSettings` will instantiate a new `UserDefaults` every time it initializes. This doesn't take a lot of time, but when you do it once for each of my 100+ blogs, it adds up.
- Each `Blog` will instantiate its own `GutenbergSettings` to check, even though it's all being called from a `GutenbergSettings` method.

To improve this I suggest a couple of things in this PR:

1. Make `GutenbergSettings` use `UserDefaults.standard` instead of initializing a new one each time.
2. Use a fetch request to find if there are any blogs with Aztec instead of iterating through each one. This differs from the existing implementation as the current one will autoenable Gutenberg for each blog that didn't have a set value. I'm not sure what the impact of this would be.

I'm not following the code completely, but it looks like the rollout is at 100%, so maybe the simplest alternative is to remove this logic completely and enable Gutenberg for everyone?

To test:

I don't fully understand the current desired logic for the rollout, so I'm not sure exactly what to test. Take this as a suggestion (with code) to improve performance of the existing implementation.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/BlogService.h
+++ b/WordPress/Classes/Services/BlogService.h
@@ -176,6 +176,8 @@ extern NSString *const WPBlogUpdatedNotification;
 
 - (NSArray *)blogsWithPredicate:(NSPredicate *)predicate;
 
+- (NSInteger)blogCountWithPredicate:(NSPredicate *)predicate;
+
 /**
  Returns every stored blog, arranged in a Dictionary by blogId.
  */
--- a/WordPress/Classes/Utility/Editor/GutenbergRollout.swift
+++ b/WordPress/Classes/Utility/Editor/GutenbergRollout.swift
@@ -29,7 +29,7 @@ struct GutenbergRollout {
     }
 
     private func atLeastOneSiteHasAztecEnabled() -> Bool {
-        let allBlogs = BlogService(managedObjectContext: context).blogsForAllAccounts()
-        return allBlogs.contains { $0.editor == .aztec }
+        let predicate = NSPredicate(format: "mobileEditor == %@", "aztec")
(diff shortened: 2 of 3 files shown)


mchowning:
Thanks for this @koke ! 🙇 

> I'm not following the code completely, but it looks like the rollout is at 100%, so maybe the simplest alternative is to remove this logic completely and enable Gutenberg for everyone?

I think this is right. and we should be able to remove most, if not all, of this rollout code.

> I've been noticing the app randomly freezing when becoming active and I think this is the cause

How confident are you that this is causing a slowdown (did you notice performance improvements with your changes)? No worries if you're not sure. Certainly if we are causing a noticeable slowdown with the app, that makes removing this basically-dead code a much more urgent priority.

koke:
> How confident are you that this is causing a slowdown (did you notice performance improvements with your changes)? No worries if you're not sure. Certainly if we are causing a noticeable slowdown with the app, that makes removing this basically-dead code a much more urgent priority.

I can't say I'm 100% sure, but this seems like the most likely cause for it. The problem doesn't always happen, and so far I've been using the TestFlight version, so I haven't been able to debug it precisely. However, I managed to attach an Instruments session and it was clearly stuck for 4-5 seconds on `UserDefaults.init`, and this was the only code path I could see that does that when the app becomes active.

PR #16448, opened 2021-05-07
Integrate gutenberg-mobile release 1.52.1

fluiddot:
## Description
This PR incorporates the 1.52.1 release of gutenberg-mobile.
For more information about this release and testing instructions, please see the related Gutenberg-Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3467

Release Submission Checklist


Podfile.lock
```diff
@@ -763,6 +763,6 @@ SPEC CHECKSUMS:
   ZendeskSupportSDK: e100a7a0a1bb5d7d43abbde3338727d985a4986d
   ZIPFoundation: e27423c004a5a1410c15933407747374e7c6cb6e
 
-PODFILE CHECKSUM: 969b56afb3d1417d365819ff405243ba0c0fc29a
+PODFILE CHECKSUM: 494c4001e0a9752267871abec5b4e90fa1a6f1db
 
-COCOAPODS: 1.10.0
+COCOAPODS: 1.10.1
```
fluiddot: @dcalhoun I've just noticed the version change of cocoapods here, not sure if we should use 1.10.0 or keep 1.10.1 when updating this file with the tag v1.52.1 🤔 .

dcalhoun: @fluiddot this may be the result of running `pod install` rather than `bundle exec pod install`. The latter should use the `cocopods` version declared in the project. 

https://github.com/wordpress-mobile/WordPress-iOS/blob/36434b7f834508a5faea96df0515a366a3cb302b/Gemfile.lock#L56

We should probably try rerunning the pod installation. I will do that shortly and push a commit unless I hear otherwise from you.

fluiddot: Yeah probably, although I ran `rake dependencies`, not sure how runs the pod install underneath 🤔 . I'll use `bundle exec pod install` next time, thanks for the info!

dcalhoun: Ah, interesting. Yeah, I would think `rake dependencies` could be the most appropriate way to manage this. Curious that it leverages `pod install`. 

https://github.com/wordpress-mobile/WordPress-iOS/blob/36434b7f834508a5faea96df0515a366a3cb302b/Rakefile#L73-L77

dcalhoun (approved):
LGTM. Smoke tested on a iPhone SE (build `47372`).

Podfile.lock
fluiddot: I think the commands ran through the `Rakefile` are using the versions installed via bundler. In my case, I realized that I updated the gems by mistake so that's why it was using a newer version of cocoapods 😄 . I reset the `Gemfile.lock`, run `bundle clean && bundle install` and now it's using the correct version 🎊 .

Issue #16068, opened 2021-03-11
Media: Unable to upload audio file from Voice Memos app


mdrockwell:
### Expected behavior
Audio files recorded in the Voice Memos app and saved to Files should upload to media library.

### Actual behavior
Error occurs instead:

```
The operation couldn't be completed.
(WordPress.MediaURLExporter.URLExportError
error 2.)
```

### Steps to reproduce the behavior
1. Record an audio clip in the Voice Memos app.
2. Tap on the … icon and select "Save to Files".
3. Save the file in any location.
4. In the WordPress app, navigate to the Media Library
5. Tap on the + button.
6. Select "Other Apps"
7. Select your audio recording from Voice Memos
8. An error appears.

![Markup on 2021-03-11 at 11:54:18](https://user-images.githubusercontent.com/42454301/110824067-b3350600-8260-11eb-98dd-feed27bba2fd.png)

##### Tested on iPhone 11 Pro, iOS 14.4, WPiOS 16.9.0.0

jd-alexander:
@ceyhun could this in any way impact Audio block flow? meaning someone may try to do an upload this way to be able to access it from the Audio block?

ceyhun:
Yes indeed this seems to not work on Audio block as well. I was able to reproduce this on WPiOS 16.8 and 16.9 both from media library and Audio block on a wpcom site with premium plan.

eduardozulian:
Reported by a user in 3866804-zen with WPiOS 16.9

mdrockwell:
Another report here: 3901679-zen

ceyhun:
I came across a similar issue in 3926774-zen and I think this is in general about .m4a files rather than specifically Voice Memos (which also happen to use .m4a format). 

When I try uploading an .m4a from Safari and Firefox using an Audio block, I get this error:

<img width="600" src="https://user-images.githubusercontent.com/1845482/115589134-c90e0000-a2cf-11eb-97de-731f14ac08c1.png">

But when I tried uploading an .m4a from Safari using web's media library from https://wordpress.com/media/ **and it was successful**! I then used that media in an Audio block and it worked! Although I had some playback issue when I previewed the page:

<img width="600" src="https://user-images.githubusercontent.com/1845482/115590039-cfe94280-a2d0-11eb-94f9-5362eb03983d.png">

[...]

PR #16450, opened 2021-05-07
Fix audio file upload issues for audio block and from media library

ceyhun:
Fixes audio file upload issues for audio block and from media library
Could "potentially fix" https://github.com/wordpress-mobile/WordPress-iOS/issues/16068 as well.

<sup>internal reference: p5T066-2fh-p2#comment-8147</sup>

To test:

Try to upload an .mp3 file:
1. Using Audio block in the editor
2. Through the media library

Try on different sites:
1. Personal plan
2. Business plan (atomic)
3. Self-hosted without Jetpack
4. Self-hosted with Jetpack

@bjtitus This change [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/15980/files#r583359092) in that PR seems to have caused some of the issues uploading audio files in the block editor using Audio block and also through media library. Do you think Stories could be affected by this change? 

## Regression Notes

- Media upload for Stories

- Tested uploading a video from Stories

- I didn't add any. I saw that there are automated tests for adding an Image block, but it seems like it's relying on the example images that are already on the device and using the basic album/photo picker. Doing the same thing for Audio block would require first figuring out how to put an audio file on the simulator/device and then figuring out how to pick that file using the complicated iOS file browser interface during the automated test, which might require substantial amount of effort.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Media/MediaURLExporter.swift
+++ b/WordPress/Classes/Utility/Media/MediaURLExporter.swift
@@ -77,9 +77,10 @@ class MediaURLExporter: MediaExporter {
     ///
     func exportURL(fileURL: URL, onCompletion: @escaping OnMediaExport, onError: @escaping OnExportError) -> Progress {
         // Verify the export is permissible
-        if let fileExtension = fileURL.typeIdentifierFileExtension {
+        if let urlExportOptions = urlOptions, let fileExtension = fileURL.typeIdentifierFileExtension {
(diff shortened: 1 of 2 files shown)


bjtitus (approved):
:shipit: Tested with video from Stories + audio directly in the Editor and in the Media Picker and all seem to work properly.

The addition of the check for `MediaImportService.defaultAllowableFileExtensions` resolves the video upload issue from what I can tell.

PR #16454, opened 2021-05-08
Notification Likes: purge likes older than 7 days

ScoutHarris:
Fixes #n/a

This adds a new `dateFetched` property to `LikeUser`. After new post or comment likes are fetched, any like that was last fetched at least 7 days ago is deleted.

To test:

To see that "old" likes are deleted, you can set the `dateFetched` to some time ago when `LikeUser` is created [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/c5a4a79b9012984084c3ca8fce2cedbde564f266/WordPress/Classes/Services/LikeUserHelpers.swift#L26). Example: 
`likeUser.dateFetched = Calendar.current.date(byAdding: .day, value: -14, to: Date()) ?? Date()`

In `purgeStaleLikes`, you can set a breakpoint to inspect the contents of `users` [fetched](https://github.com/wordpress-mobile/WordPress-iOS/blob/c5a4a79b9012984084c3ca8fce2cedbde564f266/WordPress/Classes/Services/LikeUserHelpers.swift#L68) (or log out `users`) .  

- Select both Comment and Post notifications.
- Check the contents of `users` (as noted above) to ensure old likes are deleted.

## Regression Notes
N/A. Feature incomplete.

N/A. Feature incomplete.

N/A. Feature incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/PostService+Likes.swift
+++ b/WordPress/Classes/Services/PostService+Likes.swift
@@ -28,6 +28,8 @@ extension PostService {
             DDLogError(String(describing: error))
             failure(error)
         }
+
+        LikeUserHelper.purgeStaleLikes()
     }
 
 }
--- a/WordPress/Classes/Services/PostService+Likes.swift
+++ b/WordPress/Classes/Services/PostService+Likes.swift
@@ -23,13 +23,12 @@ extension PostService {
             self.createNewUsers(from: remoteLikeUsers, postID: postID, siteID: siteID) {
                 let users = self.likeUsersFor(postID: postID, siteID: siteID)
                 success(users)
+                LikeUserHelper.purgeStaleLikes()
             }
         } failure: { error in
             DDLogError(String(describing: error))
             failure(error)
         }
-
-        LikeUserHelper.purgeStaleLikes()
     }
 
 }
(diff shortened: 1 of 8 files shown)


WordPress/Classes/Services/PostService+Likes.swift
```diff
@@ -28,6 +28,8 @@ extension PostService {
             DDLogError(String(describing: error))
             failure(error)
         }
+
+        LikeUserHelper.purgeStaleLikes()
```
aerych: I wonder if this should happen from inside the `getLikesForPostID` success block to avoid a scenario where the list might start populated, then quickly empty when this spot is reached, then be filled again when the api request completes.  I'm wondering what a user might think of that (hopefully) momentary flicker.  I guess it might also end up emptying a stale list while the user was offline. Not sure if that is the best experience 🤔  
Thoughts?

aerych (approved):
Hola @ScoutHarris 👋 
Patch looks good.  Tested and confirmed that the clean up logic runs and works.  I had one question about when it might be best to call `purgeStaleLikes`. 
Otherwise :shipit:  :)

WordPress/Classes/Services/PostService+Likes.swift
ScoutHarris: Ah, good point! I shall move it. Thanks!

PR #16458, opened 2021-05-10
Changed the quick action button coloring.

diegoreymendez:
Addresses the issue reported here: https://github.com/wordpress-mobile/WordPress-iOS/issues/15750#issuecomment-832564747

## To test:

You should test this PR with the `newNavBarAppearance` flag ON, which is the default.  That said, setting that flag OFF should also maintain the previous appearance.

1. Go the "My Site" tab.
2. Check that the quick action buttons have the same background color as the view they're in.
3. Keep your finger on any of the buttons and make sure the "activated" color is different and looks good.

## Regression Notes

This PR could potentially break the quick action buttons coloring when the `newNavBarAppearance` flag is OFF.

That said, it would not be a big deal as this flag is ON for all our users, and I don't know of any plans to go back to the previous UI.

I tested setting the flag OFF and the buttons look good.

None, since it's really hard to test for this, and since the flag is staying ON as far as I'm aware.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift
@@ -264,6 +264,32 @@ extension UIColor {
         return UIColor(light: .white, dark: .black)
     }
 
+    // MARK: - Quick Action Buttons
+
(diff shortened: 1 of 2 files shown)


frosty (approved):
:shipit:

PR #16464, opened 2021-05-10
Notification Likes: show cached before refetching

ScoutHarris:
Fixes #n/a

When the Likes list is accessed, the cached Likes is now displayed before refetching.

Note: Failures aren't being handled yet, so a message is not yet displayed when fetching fails and there are no cached Likes (i.e. when there is no network connection).

To test:

You might want to start with a fresh install to clear any previously cached Likes.

- Disable your connection.
- Select a notification.
  - Verify the likes list is empty (only the header showing).
- Enable your connection.
- Reselect the notification.
  - Verify Likes are displayed.
- Disable your connection.
- Reselect the notification.
  - Verify the Likes are displayed.

| No cached | Cached |
|--------|-------|
| no_cached | cached |

## Regression Notes
N/A. Feature incomplete.

N/A. Feature incomplete.

N/A. Feature incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Likes/LikesListController.swift
+++ b/WordPress/Classes/ViewRelated/Likes/LikesListController.swift
@@ -107,15 +107,14 @@ class LikesListController: NSObject {
             self?.isLoadingContent = false
         }
 
-        fetchStoredLikes(success: successBlock, failure: failureBlock)
+        fetchStoredLikes(completion: successBlock)
         fetchLikes(success: successBlock, failure: failureBlock)
     }
 
     /// Fetch Likes from Core Data depending on the notification's content type.
     /// - Parameters:
-    ///   - success: Closure to be called when the fetch is successful.
-    ///   - failure: Closure to be called when the fetch failed.
-    private func fetchStoredLikes(success: @escaping ([LikeUser]) -> Void, failure: @escaping (Error?) -> Void) {
+    ///   - completion: Closure to be called when the fetch is complete.
+    private func fetchStoredLikes(completion: @escaping ([LikeUser]) -> Void) {
 
         let users: [LikeUser]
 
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -97,16 +97,38 @@ class LikesListController: NSObject {
...
+        }
+
+        let failureBlock: (Error?) -> Void = { [weak self] _ in
             // TODO: Handle error state
             self?.isLoadingContent = false
-        })
+        }
+
+        fetchStoredLikes(completion: successBlock)
+        fetchLikes(success: successBlock, failure: failureBlock)
```
aerych: I wonder if we should pass the same success block to both calls. The block will set isLoadingContent = false, which would happen when `fetchStsoredLikes` completes but while `fetchLikes` was still working.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -97,16 +97,38 @@ class LikesListController: NSObject {
...
+
+        switch content {
+        case .post(let postID):
+            users = postService.likeUsersFor(postID: postID, siteID: siteID)
+        case .comment(let commentID):
+            users = commentService.likeUsersFor(commentID: commentID, siteID: siteID)
+        }
+
+        completion(users)
     }
```
aerych: I'm curious about the use of the completion block rather than just returning the fetched users.  There's not really an async element to what this method and the services are doing.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
aerych: Maybe fetchedStoredLikes could just be an assignment, and the success block just be used by fetch likes?

aerych (changes requested):
Howdy @ScoutHarris 👋 
I had a question or two.  Lemme know what ya think.

ScoutHarris:
Hey @aerych . Block tomfoolery removed. 🤦 Ready for another go please.

aerych (approved):
:shipit: ! :)

ScoutHarris:
Thanks @aerych !

PR #16467, opened 2021-05-10
Fixes truncated suggested username in the change username screen

### LINKED ISSUE #8961: Signup: Long usernames are cut off in "Change Username" list
### Expected behavior

I expect on the Change Username to have a full preview of each username I could select.

### Actual behavior

With a very long username, the end of the username is cut off with an ellipsis and there's no way to view the full username. This is especially visible when the Display Text is set to a very large font.

1. Go to the Settings app > General > Accessibility > Larger Text and set a very large text size.
2. Sign up for a new account in the WordPress app.
3. On the epilogue screen, select the username to open the "Change Username" screen.
4. Search for a long username.
5. Note that long usernames are cut off; try rotating the screen or selecting usernames to view the full name. There's no way to view them.

##### Tested on iPhone 6, iOS 11.2.6, WPiOS 9.7.0.20180316

Gio2018:
Fixes #8961

To test:
- On the device or simulator, open the "Settings" app, then go to "Accessibility -> Display & Text Size -> Large Text" and turn on the "Larger Accessibility Sizes" option
- Build/run and go to "Me -> Account Settings -> Username"
- if you don't have longer usernames in the suggestions, type a keyword to trigger them
- Make sure that the longer usernames suggestions are not truncated. You should see them wrapped at the last visible character of the first line (see example screenshot below)

<p align=center>

</p>

## Regression Notes
None: the changes are made so that it will only impact the change username view

test the change username screen

4. NA

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Me/My Profile/Change Username/ChangeUsernameViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/My Profile/Change Username/ChangeUsernameViewController.swift
@@ -15,11 +15,7 @@ class ChangeUsernameViewController: SignupUsernameTableViewController {
     init(service: AccountSettingsService, settings: AccountSettings?, completionBlock: @escaping CompletionBlock) {
         self.viewModel = ChangeUsernameViewModel(service: service, settings: settings)
         self.completionBlock = completionBlock
-
         super.init(nibName: nil, bundle: nil)
-
-        self.suggestionCellNumberOfLines = 0
-        self.suggestionCellLineBreakMode = .byCharWrapping
     }
 
     required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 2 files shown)


frosty (commented):
Hey, changes are working as expected 🙂

Is there a reason we're not using the same style for cells in both screens? That could simplify the code.

Otherwise, I wonder if having an overriddable configuration method instead of properties might be a little cleaner? Something like this in SignupUsernameTableViewController:

```
private func suggestionCell(username: String, checked: Bool) -> UITableViewCell {
        let cell = UITableViewCell()

        cell.textLabel?.text = username

        if checked {
            cell.accessoryType = .checkmark
        }

        configureSuggestionCell(cell)

        return cell
    }

    func configureSuggestionCell(_ cell: UITableViewCell) {
        cell.textLabel?.textColor = .neutral(.shade70)

        cell.textLabel?.numberOfLines = suggestionCellNumberOfLines
        cell.textLabel?.lineBreakMode = suggestionCellLineBreakMode

        cell.indentationWidth = SuggestionStyles.indentationWidth
        cell.indentationLevel = SuggestionStyles.indentationLevel
    }
```

and then in ChangeUsernameViewController, something like:

```
override func configureSuggestionCell(_ cell: UITableViewCell) {
    super.configureSuggestionCell(cell)

    self.suggestionCellNumberOfLines = 0
    self.suggestionCellLineBreakMode = .byCharWrapping
}
```

Gio2018:
> Is there a reason we're not using the same style for cells in both screens? 

Actually there isn't. I updated the code to use the same style on both screens.

diegoreymendez (approved):
Works as described.

PR #16493, opened 2021-05-13
SharingService: Ensure a completion block is called during sync

frosty:
I spotted an issue where `syncBlogAndAllMetadata` may never complete. One of the sync calls in this method is to the sharing service. If this method is called on a site for which sharing isn't possible, the SharingService sync returns but never calls a completion block. This means that the `dispatch_group_notify` in `syncBlogAndAllMetadata` is never hit.

**To test**

* Put a breakpoint in the `dispatch_group_notify` at the end of `syncBlogAndAllMetadata` in `BlogService.m`.
* Add a self hosted site to the app.
* Navigate to Blog Details for that site. This will call the blog sync method and your breakpoint should be hit.
* You can also test on develop following the same steps, and you'll see that your breakpoint is never hit.

## Regression Notes

I can't really see any here. Whether we fail or succeed the SharingService sync, both outcomes trigger a `dispatch_group_leave`. And exiting the dispatch group at the end is more correct than our current behaviour.
I've also checked all the places where we use the sharing service sync, and they also don't differentiate the outcome based on success / failure, so the fact we call the failure block shouldn't matter.

I tested syncing publicize connections, and reviewed the code where this method is called.

I added a new sharing service test to ensure that a completion block is always called for a non-dotcom site.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/SharingService.swift
+++ b/WordPress/Classes/Services/SharingService.swift
@@ -60,16 +60,17 @@ open class SharingService: LocalCoreDataService {
     ///
     @objc open func syncPublicizeConnectionsForBlog(_ blog: Blog, success: (() -> Void)?, failure: ((NSError?) -> Void)?) {
         let blogObjectID = blog.objectID
-        guard let remote = remoteForBlog(blog) else {
-            failure?(nil)
+//        guard let remote = remoteForBlog(blog) else {
+            failure?(SharingServiceError.siteWithNoRemote as NSError)
             return
-        }
-        remote.getPublicizeConnections(blog.dotComID!, success: { remoteConnections in
-
-            // Process the results
-            self.mergePublicizeConnectionsForBlog(blogObjectID, remoteConnections: remoteConnections, onComplete: success)
-        },
-        failure: failure)
+//        }
+//                
+//        remote.getPublicizeConnections(blog.dotComID!, success: { remoteConnections in
+//
+//            // Process the results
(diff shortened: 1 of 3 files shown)


WordPress/Classes/Services/SharingService.swift
```diff
@@ -61,6 +61,7 @@ open class SharingService: LocalCoreDataService {
     @objc open func syncPublicizeConnectionsForBlog(_ blog: Blog, success: (() -> Void)?, failure: ((NSError?) -> Void)?) {
         let blogObjectID = blog.objectID
         guard let remote = remoteForBlog(blog) else {
+            failure?(nil)
```
diegoreymendez: IMHO `success()` and `failure(error)` make sense, whereas `failure(nil)` feels a bit unecessarily ambiguous.

Considering we know exactly what happened it may be a good idea to return an error object here.

diegoreymendez (approved):
Works as described.  I added a suggestion that I believe is a good practice for failure situations.

WordPress/Classes/Services/SharingService.swift
frosty: Thanks for the note! I've added an error, which is now logged out when syncing blog details:

```
2021-05-17 11:44:39:942 WordPress[52465:2163375] Failed syncing publicize connections for blog https://example.wordpress.com: WordPress.SharingService.SharingServiceError.siteWithNoRemote
```

PR #16496, opened 2021-05-14
Run pod install update when pod install fails

cameronvoell:
This changes makes it so `rake dependencies` command does not fail when you you are missing pod repo updates. 

To test:

## Regression Notes

PR submission checklist:

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -72,7 +72,7 @@ namespace :dependencies do
 
     task :install do
       fold("install.cocoapds") do
-        pod %w[install]
+        sh "bundle exec pod install || bundle exec pod install --repo-update"
       end
     end
 


Rakefile
```diff
@@ -72,7 +72,7 @@ namespace :dependencies do
 
     task :install do
       fold("install.cocoapds") do
-        pod %w[install]
+        sh "bundle exec pod install || bundle exec pod install --repo-update"
```
AliSoftware, line 75: 🤔  The `Rakefile` is already run in the context of bundler (you're supposed to invoke it via `bundle exec rake dependencies` so not sure the call to `bundle exec` is necessary here… tho I see that the `pod` function that was used prior to this change does use `bundle exec` too… so meh, maybe we can keep it 🤷 

What do you think about still using that `pod` function helper that we've been using so far, and use ruby's `rescue` to catch when the first call failed and threw an exception, to try the repo-update then?

```suggestion
        pod %w[install] rescue pod %w[install --repo-update]
```

Alternate indentation (since we're already in a `do…end` block which provides a scope for catching the exception):
```suggestion
        pod %w[install]
      rescue
        pod %w[install --repo-update]
```


I tried both solutions, temporarily changing `install` to `xinstall` to make the first one fail on purpose, and can confirm that in those cases if falls back to the second one

```ruby
    task :install do
      fold("install.cocoapds") do
        pod %w[install]
      rescue
        pod %w[install --repo-update]
      end
    end
```

twstokes:

PR #16499, opened 2021-05-14
Reader: avoid duplicate requests

### LINKED ISSUE #16466: Reader: duplicated posts under lists
Few users reported duplicated posts under lists. This normally is fixed after a pull to refresh:

### Steps to reproduce the issue

1. Follow any list
2. The first time the issue will probably not happening
3. Leave the app in the background for a few hours
4. When you get back posts will be duplicated

leandroalonso:
Fixes #16466

As explained in the issue, requests to the Reader API are being made twice. This results in: more data consumption and duplicated posts for lists.

The reason is that `didTapResetFilterButton()` is called everytime, which triggers the `setContent`. I removed the call to this method, this fixes the issue.

### To test:

Make sure you're following any list. Use Charles to intercept requests or add a `print` here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/WPContentSyncHelper.swift#L51

2. Open Reader
3. Go to your list
4. Make sure no duplicated requests are made to retrieve posts for this list
5. Play around and make sure the functionality is still the same

Test also a change in the Reader behavior after this change:

1. Go to Following
2. Apply any filter
3. Go to another tab
4. Go back to following
5. The filter should not be reset*

*I'm not sure if that was the intended behavior when implementing this functionality. Tbh I like more to keep the filter applied, open to suggestions anyway. :) 

## Regression Notes
n/a

Manual testing.

I haven't. The overall code in Reader needs a lot of refactoring in order to be able to include automated tests. :(

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Tab Navigation/ReaderTabView.swift
@@ -197,7 +197,6 @@ private extension ReaderTabView {
             resetFilterButton.isHidden = false
             setFilterButtonTitle(existingFilter.topic.title)
         } else {
-            didTapResetFilterButton()
             addContentToContainerView()
         }
 


emilylaguna (changes requested):
While this does fix the duplicate requests, I encountered a few new bugs:

- I can't reproduce this reliably but a few times the view displays a failed message after loading

- When relaunches the app the discover loads then displays the empty view

https://user-images.githubusercontent.com/793774/118319209-71c31000-b4c8-11eb-9ed1-6d5e9d05482d.mov


I'm not sure what the repercussions are of switching to a reusable view controller (which I agree is the right move), I think we'll need to test this extensively before releasing.

jkmassel:
We're cutting `17.5` today, so this PR is being bumped to `17.6`

leandroalonso:
Hey @Gio2018, can I get your review on this one since touches some code you wrote? Thanks!

Gio2018 (commented):
Hey @leandroalonso , this seems to solve the duplicate calls. when testing, I noticed the same issue mentioned by @emilylaguna: when the app restarts (specifically restoring the state from Reader/Discover) I see the content for a split second and then the no result VC is shown. BUT..
I then re-instated the `didTapResetFilterButton()` call and still saw the same issue. Can you please confirm that you see the issue in both cases? If so, I think it's safe to merge and open another issue for the no results. I also agree to have a reusable VC (which, if I remember correctly, was the way this was setup initially?)

leandroalonso:
@Gio2018 I see a somewhat different issue: no result VC is shown; then the content appears:



This is probably another issue, due to:

1. Discover posts are clean when the app starts
2. This causes this behavior if the app restores showing Discover

I created the issue here: https://github.com/wordpress-mobile/WordPress-iOS/issues/16663

Gio2018 (approved):

emilylaguna (approved):

PR #16508, opened 2021-05-17
Merge 17.3 final into develop

jkmassel:
Merges the last-second 17.3 fix and finalization into `develop`.

The reason for the huge drops in translatable strings is that we're no longer including certain Gutenberg strings.


loremattei (commented):
Is the removal of 1500+ strings expected? 🤔 
Looking at the diff against `trunk`, I guess they were added by mistake, but I just want to double check that this is actually the case.

fluiddot:
> Is the removal of 1500+ strings expected? 🤔
> Looking at the diff against `trunk`, I guess they were added by mistake, but I just want to double check that this is actually the case.

Yeah, it's expected.

As a reference, [here is the fix](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3467/files?file-filters%5B%5D=.swift#diff-d5d2ed0bfbc4117c67ed9c3822eda75a06c4cd100b1d51961a0df2bbe9718db5) we did for removing the extra strings added by mistake (we removed 504 strings). For each string, we introduce three lines in the `Localizable.strings` files, so for removing 504, we're actually removing around 1512 lines from those files.

Screenshot 2021-05-17 at 13 43 14

jkmassel:
> Is the removal of 1500+ strings expected? 🤔
> Looking at the diff against `trunk`, I guess they were added by mistake, but I just want to double check that this is actually the case.

Oh – yep – that's why I put "The reason for the huge drops in translatable strings is that we're no longer including certain Gutenberg strings." in the PR description, but perhaps I could've been more clear 😅

oguzkocer (approved):

PR #16520, opened 2021-05-18
Fixes the font scaling logic by updating WPShared.

diegoreymendez:
Fixes an issue that was causing some labels to show enormous font sizes.

| Before | After |
| --- | --- |
| Screen Shot 2021-05-18 at 03 28 15 | Screen Shot 2021-05-18 at 03 23 09 |

## Related PRs:

WordPressShared: https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/294

This PR is a dependency to resolve the large-text layout issues in `NewBlogDetailsHeaderView`: https://github.com/wordpress-mobile/WordPress-iOS/pull/16521

## To reproduce the original issue:

Test in `develop`.

1. Before launching WordPress, make sure the device is using the largest possible accessibility font size.
2. Launch the app.
3. Go the "My Site" and check out the font size for the site title.

## To test the fix:

Repeat the test above, but using this branch.

## Regression Notes

1. Potential unintended areas of impact

The logic for requesting scalable fonts was off, and this should only have been noticeable when using large fonts.  That said, since this code is called from several places, it would be a good idea to compare some of the App screens against `develop`.

2. What I did to test those areas of impact (or what existing automated tests I relied on)

I've compared some of the App screens.  That said, I believe the logic change is not only correct, but also necessary to resolve other issues.

3. What automated tests I added (or what prevented me from doing so)

None, because this is impossible to automate AFAIK.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16521, opened 2021-05-18
Improved large-text support in NewBlogDetailsHeaderView.

diegoreymendez:
Implements proper large text support in `NewBlogDetailHeaderView`.

The guidelines I followed for implementing these changes are the ones listed here: pbArwn-2d9-p2

Changes:
- The site title and subtitle now shrink to fit in the space they have available, without truncating. (I'm not sure if this is desirable)
- The site title and subtitle now move below the site icon when the user selects an accessibility font size.
- The action buttons now stack vertically when the user selects an accessibility font size.

## Screenshots:

| Regular Size | Accessibility Size |
| --- | --- |
| Screen Shot 2021-05-18 at 03 48 25 |  |

## Related PRs:

Depends on https://github.com/wordpress-mobile/WordPress-iOS/pull/16520

## To test:

When testing large-text support make sure to both test changing the accessibility font size before launching the App, and while it's running.

Also test this PR in iPhone and iPad.

## Regression Notes

None

Nothing

None

PR submission checklist:


osullivanchris:
I did a sketch before on this. 
- Like you, I moved the title to the next line when it gets larger. But I would keep the image left aligned
- I agree with you that we can reduce the font size for longer strings. But when it moves to the next line, it might also be worth allowing it to wrap too, if the user has made the type larger
- Quick actions could go to a 2x2 layout before they go 1 per row
- when the quick actions text gets larger, we could also consider making the circle and icon bigger too (to a reasonable level) so that they look more in keeping

thanks for looking into this
quick-sketch

Also my image was for similar issues that were raised on iPad, where the left section is tight. So we should make something that works for iPhone and iPad here.

diegoreymendez:
@osullivanchris - I've looked into all of your feedback, including the scaling factors we discussed via DM.

What was fixed:

- I've configured a minimum scale factor of 0.75 for the title and subtitle for regular font sizes.
- I've configured a minimum scale factor of 0.5 for the title and subtitle for regular font sizes.

Because I'm needing to wrap this PR up and move on to another task, I wasn't able to fix the following:

- I spent a few hours today pairing up the action buttons, and there are layout issues that will require additional work to take care of.
- The same applies to the left-alignment of the site icon in vertical mode.
- Since I tried pairing up the action buttons, I didn't try making them bigger.

At this point I really need to move on to something else, and since IMHO this PR provides a noticeable readability improvement as it is I was wondering if you'd be ok with moving forward.

If you agree with this I'll spawn issues for the things I could not fix.

osullivanchris:
@diegoreymendez thanks for the update. Yeah its good to get some improvements in there but I still think we can make further improvements, so let's capture the outstanding issues as separate tasks if we are parking this one for now.

osullivanchris:
@frosty had a good idea to hide Quick Actions when the display is very large. We are struggling to find a nice way to display their layout. And, the list items will be large at that point too, so you don't need large Quick Actions and large list items of the same thing.

frosty (commented):
This looks like a big improvement over what we already have! I agree with Chris that I think it might be best to simply hide the quick actions when we have very large text, as the corresponding items in the table itself will be equally large.

I also think there might be an issue with the scaling of the icon on the site link. It's looking much smaller to me at default size than on develop:

| develop | this branch |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-05-20 at 11 22 38 | Simulator Screen Shot - iPhone 12 Pro - 2021-05-20 at 10 00 08 |

I've not tried it myself, but I wonder if we can use something like [scaledValue](https://developer.apple.com/documentation/uikit/uifontmetrics/2877387-scaledvalue) ([see also](https://littlebitesofcocoa.com/309-uifontmetrics)) to help here, or perhaps detect the height of the label itself?

frosty:
Just tested `scaledValue` and this sort of thing seems to work for me (adjusting the size when the traits change):

```
let size: CGFloat = 13.0
let metrics = UIFontMetrics(forTextStyle: .footnote)
let scaledSize = metrics.scaledValue(for: size)
subtitleButton.setImage(UIImage.gridicon(.external, size: CGSize(width: scaledSize, height: scaledSize)), for: .normal)
```

I chose 13 there because that's the size it seems to currently be on `develop` at default font size. I guess we'd want to move that to a constant somewhere.

Gio2018 (approved):
Great work @diegoreymendez , this worked as described.

About the other comments in the PR I think they can safely be addressed at another time, since this is a big improvement already and it follows the guidelines

PR #16548, opened 2021-05-20
Fixes video format upload problem

diegoreymendez:
Fixes an issue first reported by @elibud with video uploads that weren't viewable from Firefox or Android devices.

The actual issue was twofold:

- We weren't converting our videos to an open video compression standard.  We were passing the video through in Apple's default compression format.  I've addressed this issue by forcing videos to be converted to H264, which is broadly supported.
- We were allowing videos to retain their original storage format (.mov).  This is less of an issue than the previous point since `.mov` support has grown considerably over time, but I've taken the opportunity to force the storage format to be forced to `.mp4` since it's even more broadly supported, and less likely to cause issues.

Upsides:
- Better compatibility of uploaded videos

Downsides:
- The compression time is now longer than it was before.

## To reproduce the original issue:

Test in `release/17.4`.

1. Create a post.
2. Upload a video.
3. Publish the post
4. View the post in Firefox or using an Android device.  You'll notice the video doesn't  display properly.

## To test:

Repeat the reproduction steps in this branch.  The video should work just fine.

## Regression Notes

Slower video uploads.

Nothing much to do, other than to recognize the tradeoff.

None.

## PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,6 +8,7 @@
 * [*] Fixed the Follow Sites Quick Start Tour so that Reader Search is highlighted. [#16391]
 * [*] Enabled approving login authentication requests via push notification while the app is in the foreground. [#16075] 
 * [**] Added pull-to-refresh to the My Site screen when a user has no sites. [#16241]
+* [***] Fixed a bug that was causing uploaded videos to not be viewable in other platforms. [#16548]
 
 17.3
 -----
--- a/WordPress/Classes/Services/MediaSettings.swift
+++ b/WordPress/Classes/Services/MediaSettings.swift
@@ -29,7 +29,7 @@ class MediaSettings: NSObject {
             case .size3840x2160:
                 return AVAssetExportPreset3840x2160
             case .sizeOriginal:
-                return AVAssetExportPresetPassthrough
+                return AVAssetExportPresetHighestQuality
             }
         }
 
--- a/WordPress/Classes/Utility/Media/MediaAssetExporter.swift
+++ b/WordPress/Classes/Utility/Media/MediaAssetExporter.swift
@@ -167,8 +167,7 @@ class MediaAssetExporter: MediaExporter {
         }
 
         // Configure a video exporter to handle an export session.
-        var exporterVideoOptions = videoOptions ?? MediaVideoExporter.Options()
-
+        let exporterVideoOptions = videoOptions ?? MediaVideoExporter.Options()
         let originalFilename = videoResource.originalFilename
 
         // Request an export session, which may take time to download the complete video data.
--- a/WordPress/Classes/Utility/Media/MediaAssetExporter.swift
+++ b/WordPress/Classes/Utility/Media/MediaAssetExporter.swift
@@ -169,10 +169,6 @@ class MediaAssetExporter: MediaExporter {
         // Configure a video exporter to handle an export session.
         var exporterVideoOptions = videoOptions ?? MediaVideoExporter.Options()
 
-        if exporterVideoOptions.preferredExportVideoType == nil {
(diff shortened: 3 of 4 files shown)


diegoreymendez:
Fyi @elibud, @yaelirub, @hypest and @Gio2018.

SergioEstevao:
@diegoreymendez was this issue happening on all WP sites? I thought that with sites that have VideoPress this shouldn't be a problem? VideoPress should convert the video in the server.

For non VideoPress sites your fix looks correct, but how are you sure that that Preset is always saving to the H264 format?

diegoreymendez:
> @diegoreymendez was this issue happening on all WP sites? I thought that with sites that have VideoPress this shouldn't be a problem? VideoPress should convert the video in the server.
> 
> For non VideoPress sites your fix looks correct, but how are you sure that that Preset is always saving to the H264 format?

Thanks for the insights Sergio.  I didn't know VideoPress took care of that.  I still think it probably makes sense to convert the video locally, since it should work well in most scenarios, but let me know if you think otherwise.

About H264 - I found some information about it here:

https://developer.apple.com/documentation/avfoundation/media_assets_and_metadata/exporting_video_to_alternative_formats

> For example, to convert a movie file to H.264/MPEG-4 AVC, use an existing preset that encodes to H.264, such as AVAssetExportPresetHighestQuality, and set the output file type to mov.

And also if you go to the declaration of  `AVAssetExportPresetHighestQuality` in `AVAssetExportSession.h`, you will see the following:

```objc
/* These export options can be used to produce movie files with video size appropriate to the device.
   The export will not scale the video up from a smaller size. The video will be compressed using
   H.264 and the audio will be compressed using AAC.  */
AVF_EXPORT NSString *const AVAssetExportPresetLowQuality         API_AVAILABLE(macos(10.11), ios(4.0), tvos(9.0)) API_UNAVAILABLE(watchos);
AVF_EXPORT NSString *const AVAssetExportPresetMediumQuality      API_AVAILABLE(macos(10.11), ios(4.0), tvos(9.0)) API_UNAVAILABLE(watchos);
AVF_EXPORT NSString *const AVAssetExportPresetHighestQuality     API_AVAILABLE(macos(10.11), ios(4.0), tvos(9.0)) API_UNAVAILABLE(watchos);
```

Basically these presets are meant to recompress a video to H264.

frosty (approved):
I've been able to verify that the fix works for me, uploading to a free wpcom site. I tested on Firefox on my Mac as well as Chrome on Android.

I'd be keen to get @SergioEstevao's 👍 before we merge though!

diegoreymendez:
Sounds good to me.

@SergioEstevao as additional info, VideoPress seems to be doing the transcoding just fine, although it seems the video is not being replaced - I'm not sure if that's a VideoPress or a Gutenberg bug.  Feel free to DM me if you want the details.

The solution in this PR can be tested in both a free or VideoPress - enabled site, and it should work in both scenarios.

SergioEstevao (approved):
:shipit:

hypest:
> as additional info, VideoPress seems to be doing the transcoding just fine, although it seems the video is not being replaced - I'm not sure if that's a VideoPress or a Gutenberg bug. Feel free to DM me if you want the details.

👋 @diegoreymendez , just to be sure, with the solution in this PR where we are now locally converting the video before uploading, would we still need the markup to update after VideoPress does its own pass too?

diegoreymendez:
For playback to work no.  It will work just fine with my changes.

But you'll need to update the markup to avoid "breaking" VideoPress' play controls on the web.  When the markup is updated, in a post on a VideoPress-enabled site, the video has a special set of play controls on the web that are part of the VideoPress feature.

But I'd imagine that's a different priority than the videos not working.

Hope that helps!

diegoreymendez:
You can see how the VideoPress player looks like in the video at this link:

https://wordpress.com/support/videopress/

PR #16547, opened 2021-05-20
Likes List: show activity indicator when content is loading.

ScoutHarris:
When Likes are loading, an activity indicator is displayed in the list table footer.

To test:
- On a slow network, go to Notifications > Likes.
- Select a notification.
  - Verify the activity indicator is displayed at the bottom until the content is loaded.
- Scroll to the bottom.
  - Verify the activity indicator is displayed at the bottom until the next page's content is loaded.

Hack Tip:
You can modify the comment/post fetching to better test pagination (thus loading) if you don't have a lot of Likes. In [`LikesListController.fetchLikes`](https://github.com/wordpress-mobile/WordPress-iOS/blob/dcce2f8856681c3ea1a1025340a99f25fa8aae66/WordPress/Classes/ViewRelated/Likes/LikesListController.swift#L150):
- Set the postID/commentID and siteID of a post/comment you know to have a lot of Likes.
- And/or add a `count` param to the `getLikesFor` methods to fetch less Likes at a time. Example:
```
postService.getLikesFor(postID: 999,
                        siteID: 999,
                        count: 12,
                        before: lastFetchedDate,
                        purgeExisting: isFirstLoad,
                        success: success,
                        failure: failure)
```

## Regression Notes
1. Potential unintended areas of impact
N/A. Feature incomplete.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N/A. Feature incomplete.

3. What automated tests I added (or what prevented me from doing so)
N/A. Feature incomplete.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16553, opened 2021-05-20
Likes List: filter out users already fetched

ScoutHarris:
WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/402

 There is a scenario where multiple users might like a post/comment at the same time, and then end up split between pages of results. So we'll track which users we've already fetched for the `lastFetchedDate`, and send those to the Post and Comment services to filter out of the response so we don't get duplicates.

Also, since this is the last part of Like Notifications, the feature flag is enabled for all.

To test:
- Select a notification with:
  - A lot of Likes (See #16547 for a hack tip if needed.)
    - If you manually set the IDs, you'll need to do the same [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/004747e5ccdd5482a7c2e9717b892854382f19e8/WordPress/Classes/ViewRelated/Likes/LikesListController.swift#L184).
  - Likes with the same date/time. See hack tip below to muck with the dates if needed.
- Scroll to the bottom of the list so the next page loads.
- Verify there are no duplicates at the end of one page and the beginning of the next, and no users are missing.

Hack tip for dates:
In order to have multiple Likes with the same date/time, this should work. 
- In the `createNewUsers` methods in [`PostService+Likes`](https://github.com/wordpress-mobile/WordPress-iOS/blob/d943090f9236b9405ca8c6116a05c5f2892bd5e5/WordPress/Classes/Services/PostService%2BLikes.swift#L103) / [`CommentService+Likes`](https://github.com/wordpress-mobile/WordPress-iOS/blob/d943090f9236b9405ca8c6116a05c5f2892bd5e5/WordPress/Classes/Services/CommentService%2BLikes.swift#L103) when they call `createUserFrom`, manually change the `dateLiked` to the same date. For example:

```
//            remoteLikeUsers.forEach {
//                LikeUserHelper.createUserFrom(remoteUser: $0, context: derivedContext)
//            }
            
            for (index, user) in remoteLikeUsers.enumerated() {
                let endDate = remoteLikeUsers.last?.dateLiked
                
                if index == remoteLikeUsers.endIndex - 3 {
                    user.dateLiked = endDate
                }
                
                LikeUserHelper.createUserFrom(remoteUser: user, context: derivedContext)
            }
``` 

## Regression Notes
N/A. Feature incomplete.

N/A. Feature incomplete.

N/A. Feature incomplete.

PR submission checklist:


ScoutHarris:
Hey @aerych . I made the changes we discussed via Slack. I believe it's ready for review now. Please note there is now an accompanying WPKit PR. Thank you kindly!

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -148,22 +150,47 @@ class LikesListController: NSObject {
...
+
+        var fetchedUsers = [LikeUser]()
+        switch content {
+        case .post(let postID):
+            fetchedUsers = postService.likeUsersFor(postID: postID, siteID: siteID, after: modifiedDate)
+        case .comment(let commentID):
+            fetchedUsers = commentService.likeUsersFor(commentID: commentID, siteID: siteID, after: modifiedDate)
+        }
+
+        excludeUserIDs = fetchedUsers.filter { $0.dateLiked > lastDate }.map { NSNumber(value: $0.userID) }
```
aerych: Do we need the filter here? It seems like `fetchedUsers` would already be the ones we want to exclude.  Am I missing something?

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -148,22 +150,47 @@ class LikesListController: NSObject {
...
         }
     }
+
+    // There is a scenario where multiple users might like a post/comment at the same time,
+    // and then end up split between pages of results. So we'll track which users we've already
+    // fetched for the lastFetchedDate, and send those to the endpoints to filter out of the response
+    // so we don't get duplicates or gaps.
+    private func trackUsersToExclude() {
+        guard let lastDate = likingUsers.last?.dateLiked,
+              let modifiedDate = Calendar.current.date(byAdding: .second, value: -1, to: lastDate) else {
```
aerych: It might be handy to factor the logic to get a modified date into a reusable method so it can also be used to construct a `before` date.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -148,22 +150,47 @@ class LikesListController: NSObject {
     ///   - success: Closure to be called when the fetch is successful.
     ///   - failure: Closure to be called when the fetch failed.
     private func fetchLikes(success: @escaping ([LikeUser], Int) -> Void, failure: @escaping (Error?) -> Void) {
+
         switch content {
         case .post(let postID):
             postService.getLikesFor(postID: postID,
                                     siteID: siteID,
                                     before: lastFetchedDate,
```
aerych: We'll also need to modify this date string rather than pass `lastFetchedDate`, otherwise any likers in the next page that have this as their date_liked would be excluded. 
Something like this might do the trick:
```
      if let lastDate = likingUsers.last?.dateLiked,
           let _ = lastFetchedDate,
           let modifiedDate = modifiedBeforeDate(date: lastDate) {
            let beforeStr = ISO8601DateFormatter().string(from: modifiedDate).replacingMatches(of: "T", with: " ").replacingMatches(of: "Z", with: "")
        }
```
In the example above, the `ISO8601DateFormatter` call will return a string formatted like `YYYY-MM-DDTHH:MM:SSZ`.  The endpoint expects a format like `YYYY-MM-DD HH:MM:SS`. It isn't expecting the `T` or `Z`, hence the `replacingMatches` calls.

aerych (changes requested):
Hola @ScoutHarris 👋 
Looking good! I think we're super close, we might need just a couple of tweaks.  Lemme know what you think.

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
ScoutHarris: Oops. I failed to remove that from the first iteration. Thanks!

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
ScoutHarris: Thank you!!

ScoutHarris:
OK @aerych one more time 🤞 ! Thank you!

WordPress/Classes/ViewRelated/Likes/LikesListController.swift
```diff
@@ -148,22 +150,62 @@ class LikesListController: NSObject {
...
+
+        excludeUserIDs = fetchedUsers.map { NSNumber(value: $0.userID) }
+    }
+
+    private func modifiedBeforeDate() -> Date? {
+        guard let lastDate = likingUsers.last?.dateLiked else {
+            return nil
+        }
+
+        return Calendar.current.date(byAdding: .second, value: -1, to: lastDate)
```
aerych: `value: -1,` 
I think we'd want this to be `value: 1,`. Dates are a little confusing because we're fetching in descending order. If 2:00:01 PM is the time of the last liker's 'dateLiked', and we want to fetch another page that includes any other likers who liked at 2:00:01 or before, we'd need to add a second to fetch all likers before 2:00:02.

WordPress/Classes/Services/PostService+Likes.swift
```diff
@@ -71,6 +74,25 @@ extension PostService {
...
+    /**
+     Fetches a list of users from Core Data that liked the post with the given IDs after the given Date.
+     
+     @param postID  The ID of the post to fetch likes for.
+     @param siteID  The ID of the site that contains the post.
+     @param after   Filter results to likes after this Date.
+     */
+    func likeUsersFor(postID: NSNumber, siteID: NSNumber, after: Date) -> [LikeUser] {
+        let request = LikeUser.fetchRequest() as NSFetchRequest<LikeUser>
+        request.predicate = NSPredicate(format: "likedSiteID = %@ AND likedPostID = %@ AND dateLiked > %@", siteID, postID, after as CVarArg)
```
aerych: `AND dateLiked > %@",`
I think this would need to be `AND dateLiked < %@",` to get likers who liked later the value of `after`.

WordPress/Classes/Services/CommentService+Likes.swift
```diff
@@ -71,6 +74,25 @@ extension CommentService {
...
+    /**
+     Fetches a list of users from Core Data that liked the post with the given IDs after the given Date.
+     
+     @param commentID   The ID of the comment to fetch likes for.
+     @param siteID      The ID of the site that contains the post.
+     @param after       Filter results to likes after this Date.
+     */
+    func likeUsersFor(commentID: NSNumber, siteID: NSNumber, after: Date) -> [LikeUser] {
+        let request = LikeUser.fetchRequest() as NSFetchRequest<LikeUser>
+        request.predicate = NSPredicate(format: "likedSiteID = %@ AND likedCommentID = %@ AND dateLiked > %@", siteID, commentID, after as CVarArg)
```
aerych: `AND dateLiked > %@",`
I think this would need to be `AND dateLiked < %@",` to get likers who liked later the value of `after`.

aerych (changes requested):
Did another round of testing and I noticed we were passing all the previous liker's user_ids to the exclude parameter when paginating. Looks like we just had a few signs flipped. Easy fix.  Dates are hard.

ScoutHarris:
> Did another round of testing and I noticed we were passing all the previous liker's user_ids to the exclude parameter when paginating.

Thanks for catching that! And one more time!

aerych (approved):
Took a bit on this last round because I spotted something odd and I wanted to investigate whether it was an API glitch.  Happily the API is fine and WPKit is fine, the odd thing was something more subtle and not related to the changes here.  Something we want to fix tho. 

In PostService+Likes (and CommentService+Likes probably as well) when `createNewUsers` is called, a call is made to `let users = self.likeUsersFor(postID: postID, siteID: siteID, before: before)` to get all the users that were just fetched. However, because of the modified `before` date the users include the last user from the previous request. So if 20 were returned from the endpoint, 21 would be fetched out of core data.  These are passed to the `success` block which is handled in LikesListController.  In the users are handled around line 124/125 `self?.likingUsers.append(contentsOf: users)` where they are appened to the previous likers.  This ends up duplicating a row. 
Would it make sense to just fetch all liking users here rather than appending the ones returned?
Everything else looks good to go, so :shipit: from me.  And we can resolve the glitch in another pr or as an add on to this one. However you like. :)

ScoutHarris:
> And we can resolve the glitch in another pr 

Yes please. 😄 Thank you soooo much for your help!

PR #16570, opened 2021-05-24
Implements BloggingReminders and BloggingRemindersStore.

### LINKED ISSUE #16536: Blogging Reminders: Local persistance
Tasks:

- Define what data needs to be stored locally for the blogging reminders.
- Implement code to persist the data, read it, modify it and delete it.

Considering the nature of this task it should be possible to write 100% unit tested code.

This is a possible data structure we can use for starters:

```swift
enum ReminderScheduleWeekday {
    case monday(requestIdentifier: String)
    case tuesday(requestIdentifier: String)
    case wednesday(requestIdentifier: String)
    case thursday(requestIdentifier: String)
    case friday(requestIdentifier: String)
    case saturday(requestIdentifier: String)
    case sunday(requestIdentifier: String)
}

enum ReminderSchedule {
    case weekdays(days: [ReminderScheduleWeekday])
    // other cases could be necessary in the future
}

struct BlogReminderSchedule {
    let blogId: String
    let schedule: ReminderSchedule
}

let persistedData: [BlogReminderSchedule]
```

### LINKED ISSUE #16573: Blogging Reminders: Implement scheduling logic and notification
Title has it.

diegoreymendez:
Fixes #16573. Implements `BloggingReminders` as the main interface for scheduling blogging reminder notifications. 
Fixes #16536. Implements `BloggingRemindersStore` as the underlying mechanism for persisting the user-defined reminder schedule.

The entry points for interacting with these notifications are:

```swift
// Instantiate a scheduler with the default parameters.
let scheduler = BloggingRemindersScheduler()

// Schedule notifications
scheduler.schedule(.weekly([days]))

// Query the active notifications schedule.
let schedule = scheduler.schedule
```

Since this code is not really being used yet, and since blogging reminders will sit behind a feature flag for now, I propose merging this so we can start integrating it with @frosty 's PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16587

## Things we still need to do.

- Add unit tests for the scheduling logic.
- Add error handling to `BloggingRemindersStore` through exceptions.  Remove all logging from these classes as the logging should be done by whoever instantiates the classes.
- Add calendar locale support.  Or otherwise make sure the weekday logic doesn't break with a locale change.

## To test:

This cant be tested yet since the code is not connected to the UI.
We'll make sure to test this thoroughly before enabling blogging reminders.

## Regression Notes

None.

PR submission checklist:


diegoreymendez:
@frosty, @Gio2018 - There's no need to review this PR as I still need to wrap it up.  I've published it as a draft in case you want to see what I've done for scheduling notifications.

Remember it's a work in progress but feel free to ask questions or make suggestions.

EDIT: check the PR description for details on the main entry points for scheduling blogging reminders.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    enum Weekday: Int {
+        case sunday = 1 // Keep this at 1 to match Apple's `DateComponents`' weekday number.
+        case monday
+        case tuesday
+        case wednesday
+        case thursday
+        case friday
+        case saturday
```
Gio2018: Are we planning to handle locales (like it: IT) where the week does not start on Sunday?

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
```diff
@@ -0,0 +1,157 @@
...
+
+        /// The default reminder hour.  In the future we may want to replace this constant with a more customizable approach.
+        ///
+        static let defaultHour = 10
+    }
+
+    /// Singleton for use in the App.
+    ///
+    static let `shared` = BloggingReminders()
```
Gio2018: I admit that I should read the whole thing before commenting on this 😀, but do we really need a singleton here? Since we are creating a new feature that's not interconnected with existing code, I'd expect we can inject instances of objects without singletons

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
```diff
@@ -0,0 +1,157 @@
+import Foundation
+
+/// Main interface for scheduling blogging reminders
+///
+class BloggingReminders {
```
Gio2018: Since this is the type that schedules, maybe we can call it `BloggingReminderScheduler`?

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
```diff
@@ -0,0 +1,157 @@
...
+
```
Gio2018: Maybe we can even do a protocol and then a concrete type for the weekday schedule? Or something like that?

Gio2018 (commented):
Hey @diegoreymendez , this is a very good start!
I ran out of time so I will continue tomorrow, but left a few comments in the meantime.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: I think you're right on this.  I'll try playing a bit with the code to remove the singletons.

As a side note this made me wonder if we want `BloggingRemindersStore` to be thread safe, since it's writing and reading from a file.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: I considered it but I was wondering if we could ever want to have more in here than just the scheduling.  Any thoughts on this?

If we feel this is unlikely, I'm good with the name `BloggingRemindersScheduler`.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: I didn't understand this suggestion.  Can I ask you to expand, or to ping me via DM to chat about it a bit more?

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Well, I think nothing in the current code is really locale specific.  The only reason behind sunday being the first day in the enum is that, as far as I'm aware, `DateComponents().weekday` uses a numbering that has sunday as its first day - I don't think this changes in `DateComponents` if the locale is changed.

diegoreymendez: Actually,  you may be right on this! :)

Thanks for the comment.  I'll explore and test this a bit more to make sure it works in different locales.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
Gio2018: It really depends. If it works (and aside from the name), maybe we can start considering a smaller set of responsibilities for this class and expand it (and rename it) if necessary?

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
Gio2018: Sorry the comment was totally unclear. I had in mind that this is a scheduler, so I wondered if we could do a `Scheduler` protocol for more general purposes and then a concrete implementation of `BloggingReminderScheduler`. On the other hand, I then noticed that we already have a concrete type called `Scheduler` so, reconsidering it all, it might be just me overthinking. In short, I think this comment can be ignored for now 😀

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -0,0 +1,306 @@
+import Combine
```
Gio2018: 👍🎉

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -0,0 +1,306 @@
...
+
+/// Store for the blogging reminders.  This class should not be used directly.  Please refer to `BloggingReminders` instead.
+///
+class BloggingRemindersStore {
```
Gio2018, line 6: Totally optional, but this could also be a protocol and we could have different concrete implementations if, for example, in the future we decide to use a different type of persistence.

Gio2018 (commented):
Left a couple more comments

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Sounds good to me.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Thanks, that makes it clearer.  If you're ok with it I'll move ahead as it is for now, and then we can iterate on these classes as needed.

Gio2018: yes 100%. Thank you!!

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
diegoreymendez: Hah... I had the intention of using it but really didn't find it necessary.  Any suggestions on this front are werlcome but otherwise I'll probably remove this.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
frosty: I've looked into this, and on the UI side we should be able to use a combination of `calendar.firstWeekday` and `calendar.shortWeekdaySymbols`. I guess it doesn't really matter the order we store them behind the scenes, just as long as we schedule the correct reminder for the correct day.

(the code will be something like this):

```
let days: [String] = {
    var calendar = Calendar.current
    calendar.locale = Locale.autoupdatingCurrent
    
    let firstWeekday = calendar.firstWeekday
    let symbols = calendar.shortWeekdaySymbols

    let firstPart = symbols[firstWeekday - 1 ..< symbols.count]
    let secondPart = symbols[0 ..< firstWeekday - 1]
    return Array(firstPart + secondPart)
}()
```

frosty: That said, we'll need some kind of property to map from the our enum here to the values we display in the UI, and to get them in the right order.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
```diff
@@ -0,0 +1,168 @@
...
+    ///
+    /// - Parameters:
+    ///     - weekday: the weekday to schedule a notification for.
+    ///
+    /// - Returns: the weekday with the associated notification ID.
+    ///
+    private func scheduled(_ weekday: Weekday) -> BloggingRemindersStore.Weekday {
+        let notificationID = scheduleNotification(for: weekday)
+        switch weekday {
+        case .monday:
```
frosty: How about we move this to an initializer on `BloggingRemindersStore.Weekday`? Then you could just do `return BloggingRemindersStore.Weekday(weekday, notificationID)` or similar. It feels like it might fit better there.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -0,0 +1,306 @@
...
+                return .thursday
+            case .friday:
+                return .friday
+            case .saturday:
+                return .saturday
+            }
+        }
+    }
+
+    private let dataFileURL: URL?
```
frosty: Shall we move this down with the other dataFile properties below?

frosty (commented):
Looking pretty good! I just left a few small comments! My only other concern would be confusion around duplication of having two weekday enums and two schedule enums... do you think it's worth trying to avoid that? 🤔

diegoreymendez:
> My only other concern would be confusion around duplication of having two weekday enums and two schedule enums... do you think it's worth trying to avoid that? 🤔

I've tried addressing this in a recent commit, not by removing both data structures, but by trying to bring clarity to their names and definitions.  In the process I managed to simplify  the `Coding` logic considerably.

Within `BloggingRemindersStore` I've renamed:

- `Schedule` to `ScheduledReminders`; and
- `Weekdays` to `ScheduledWeekdays`

I've also changed the latter to make it a class instead of another enum.  I think this helps bring clarity, but please let me know what you think.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: This is now outdated due to some recent changes, but the solution is not exactly the one proposed here.

Can you check if it looks good now?

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
diegoreymendez: Ended up removing this, since I'm not really using Combine (at least just yet).

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Thanks for the snippets James!

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Changed the name of the class to `BloggingRemindersScheduler`.

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Marking as resolved for now.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
diegoreymendez: The properties for the default store have been removed from this class as part of removing the singletons from these classes.

Can you let me know if those are the properties you were referring to?

WordPress/Classes/Utility/Blogging Reminders/BloggingReminders.swift
diegoreymendez: Removed the singletons.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
diegoreymendez: Would you mind if I leave it as-is for now?

I don't mind if we iterate right after to make it more flexible - I'm just not sure it's a good idea to try to go this route for the initial implementation of the solution.  From a time perspective I believe it may make sense to "separate concerns".

diegoreymendez: Do you think we could leave this as something to tackle iteratively?

I think this is something we can discuss and implement iteratively to keep the PR simpler / more focused for v1.

frosty (approved):
These changes look great! You managed to take out so much code and the weekday vs scheduled weekday distinction makes much more sense to me now 👍

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
Gio2018: sounds good to me

Gio2018 (approved):

PR #16576, opened 2021-05-25
Add default BuildKite pipeline

jkmassel:
Adds a default BuildKite build pipeline

To test:
- Ensure CI passes

## Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,41 @@
+# By default, run on macOS agents
+agents:
+  os: "macOS"
+
+# This is the default pipeline – it will build and test the app
+steps:
+
+  #################
+  # Build the app
+  #################
+  - label: ":pipeline: Build"
+    key: "build"
+    command: |
+      echo "--- 🔧 Setting up Gems"
+      gem install bundler
+      bundle install
+      echo "--- 🔨 Setting up Pods"
+      bundle exec pod install --repo-update
+      echo "--- ✍️ Copy Files"
+      cp fastlane/env/project.env-example .configure-files/project.env
+      echo "--- 🛠 Building"
+      bundle exec fastlane build_for_testing
+      echo "--- 🗜 Zip Build Products"
+      tar -cf build-products.tar DerivedData/Build/Products/
+    agents:
+      os: "macOS"
+    artifact_paths:
+      - build-products.tar
+
+  #################
+  # Run Unit Tests
+  #################
+  - label: "🧪 Unit Tests"
+    command: |
+      buildkite-agent artifact download build-products.tar .
+      tar -xf build-products.tar
+      bundle install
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,41 @@
...
+# This is the default pipeline – it will build and test the app
+steps:
+
+  #################
+  # Build the app
+  #################
+  - label: ":pipeline: Build"
+    key: "build"
+    command: |
+      echo "--- 🔧 Setting up Gems"
```
AliSoftware, line 14: https://buildkite.com/docs/pipelines/managing-log-output#collapsing-output 🤯

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,41 @@
...
+      gem install bundler
+      bundle install
+      echo "--- 🔨 Setting up Pods"
+      bundle exec pod install --repo-update
+      echo "--- ✍️ Copy Files"
+      cp fastlane/env/project.env-example .configure-files/project.env
+      echo "--- 🛠 Building"
+      bundle exec fastlane build_for_testing
+      echo "--- 🗜 Zip Build Products"
+      tar -cf build-products.tar DerivedData/Build/Products/
```
AliSoftware, line 24: Wondering if it wouldn't be more readable to extract those two commands in their own step? So that instead of having one step with collapsible output markers, we have 3 steps ("setup" step = commands up to line 20, "build" step = line 21-22, and "artefacts" step = line 23-24)?

(Except if the concept of "steps" in BuildKite is different from the concept of steps in CircleCI?)

AliSoftware: [EDIT] nvm, just read [this in the BuildKite docs](https://buildkite.com/docs/pipelines/defining-steps#adding-steps), which means that "steps" in BuildKite parlance might not be the same as "steps" in CircleCI parlance after all (but more like CircleCI's "jobs" I think?).

So this way of configuring things seems to be the expected solution after all (even if imho it feels a little bit less readable, and will lead to log truncation based on full log not individual commands output, so less useful when reading logs inline without downloading… 😢 ; will have to dig into the BuildKite docs at some point to see if there's a better way, but not urgent and this should do for now)

> When you run a pipeline, a build is created. Each of the steps in the pipeline end up as jobs in the build, which then get distributed to available agents.

_Source: https://buildkite.com/docs/pipelines#what-is-a-pipeline_

> However you add steps to your pipeline, keep in mind that steps may run on different agents. It is good practice to install your dependencies in the same step that you run them.

_Source: https://buildkite.com/docs/pipelines/defining-steps#adding-steps_

AliSoftware (approved):
Build is green [here](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/93) so this seems to work as expected, :shipit:

PR #16579, opened 2021-05-26
Fix: Notice overlap with SiteIconActionSheet

### LINKED ISSUE #16465: Quick Start "Site title updated" toast overlaps next step actions
### Expected behavior
Toasts should not interfere with actions 

### Actual behavior

https://user-images.githubusercontent.com/66104034/117725856-f9e0a700-b1a2-11eb-8205-446bf3ca13cb.MP4

1. Create a new site and accept guidance
2. Change the site title and select "Yes, Show me" on the Upload Site Icon step
3. Notice site title toast appears now and covers next steps

enejb:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/16465

To test:
1. On Update the site title. 
2. Add the site icon. 
Notice that the site title update notice doesn't over lap and interfere with the action that you wanted todo.
3. Do the same thing but now that you have the site icon. Notice that that the site title update notice doesn't interfere with the action that you wanted to do. 

Screenshots

https://user-images.githubusercontent.com/115071/119584709-00177b80-bd7e-11eb-9f97-4c3d61a246a4.mp4

IPad
Screen Shot 2021-05-25 at 5 23 39 PM

## Regression Notes
N/A

None

PR submission checklist:


frosty (commented):
Thanks for looking at this issue, @enejb! However, I'm afraid I don't feel that this is an appropriate fix.

On iPhone, displaying the alert as a bottom sheet is better for usability, as it's generally easier to the user's thumbs as they're holding the device. Also, I understand removing the sheet if the user doesn't already have a site icon added – however, from what I remember the reason we still show the sheet in this case is as a prompt to the user that that's what tapping on the icon does. Otherwise, if they tap it and see a media picker, they might not understand what they're choosing.

I haven't spent a lot of time thinking about this yet, but I _think_ a better solution might be to clear out the notice queue if the user taps "Yes, show me" on a Quick Start prompt. That way, no queued up notices will be displayed, so they won't be able to disrupt the tour. We could do this by adding a call to `ActionDispatcher.dispatch(NoticeAction.empty)` in the `if accepted` block here: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift#L107

I hope you don't mind me proposing an alternative, and thanks for taking a look at the issue!

enejb:
Thanks for your feedback @frosty. It is very much welcome and appreciated. :)
 
In order to not show the notice that could potentially overlap the Action sheet. I implemented a lock for the notices. 
Which prevents any notices from showing up on top of the Action sheet until the user decides on an action and the Actionsheet is dismissed. 

You can see it in action here. 
https://user-images.githubusercontent.com/115071/119848731-e92d7200-bec0-11eb-805c-cfc2c111252a.mp4

would love some input on this approach.

WordPress/Classes/Stores/NoticeStore.swift
```diff
@@ -150,6 +167,7 @@ struct NoticeStoreState {
 /// - SeeAlso: `NoticeAction`
 class NoticeStore: StatefulStore<NoticeStoreState> {
     private var pending = Queue<Notice>()
+    private var noticeLock = Bool()
```
frosty: You can just do `private var noticeLock = false`

WordPress/Classes/Stores/NoticeStore.swift
```diff
@@ -183,15 +205,27 @@ class NoticeStore: StatefulStore<NoticeStoreState> {
...
+        state.notice = nil
+        noticeLock = true
+    }
+
+    private func unlock() {
+        noticeLock = false
+        dequeueNotice()
+    }
+
     private func dequeueNotice() {
```
frosty: A bit nickpicky, but I'd have the enqueue and dequeue methods together, and then lock and unlock together below that.

WordPress/Classes/Stores/NoticeStore.swift
```diff
@@ -183,15 +205,27 @@ class NoticeStore: StatefulStore<NoticeStoreState> {
...
-        if state.notice == nil {
+        if state.notice == nil && !noticeLock {
             state.notice = notice
         } else {
             pending.push(notice)
         }
     }
 
+    private func lock() {
+        state.notice = nil
```
frosty, line 224: We perhaps want to check that we're not already locked, and the same in unlock check we're not already unlocked.

WordPress/Classes/Stores/NoticeStore.swift
frosty: In fact, maybe we should just name this `locked` or `storeLocked` or something? Then our checks read a little nicer – `if locked`, `if !locked`

frosty (commented):
Thanks for sharing a different approach! Locking the queue seems interesting 🤔 I added a few comments to the code. I think my main concern here is that this is added just for this one specific situation, and I wonder if we have the same happening elsewhere in Quick Start.

I'm going to ping @shiki for his thoughts as he's worked on the Notice system a fair amount and might have some other ideas!

shiki:
Will try to take a look today. If not, then early Monday morning. 🙂

WordPress/Classes/Stores/NoticeStore.swift
```diff
@@ -130,6 +143,10 @@ enum NoticeAction: Action {
     case clearWithTag(Notice.Tag)
     /// Removes all Notices except the current one.
     case empty
+    // Prevents the notices from showing up untill an unlock action.
+    case lock
+    // Show the missed notices.
+    case unlock
```
shiki, line 148: If we're going ahead with this strategy, we should consider adding unit tests for these in [`NoticeStoreTests`](https://github.com/wordpress-mobile/WordPress-iOS/blob/fix/notice-overlap-for-site-icon/WordPress/WordPressTest/Classes/Stores/NoticeStoreTests.swift). 🙂

shiki (commented):
I've spent a couple of hours thinking about this and I could not find a better alternative. 😄 

I thought it would be possible to show the `UIAlertController` above the notice. But the `Notice` is unfortunately shown inside `UntouchableWindow`. But the `UIAlertController` is presented inside the main `UIWindow`. And the `UntouchableWindow` is always **above** the main `UIWindow`. 🙈 So I don't think there's anything we can do here. 

I also looked if there's a way to _detect_ if a `UIAlertController` is presented. If we could, then we could make `NoticeStore` hide the current notice when that happens. But it doesn't look like there's any `NSNotification` or a globally accessible callback whenever `UIAlertController` is presented. 🤷 

So, I think what we're doing here, dismissing the notice imperatively whenever a `UIAlertController`, is our best bet. 

## Concerns About Locking

The locking works but I'm concerned about the requirement to call `unlock()`. That seems quite easy to miss and could be a potential bug in the future if someone accidentally removes the `unlock()` call.

I'm curious, why didn't we consider clearing the notices as Frosty mentioned in https://github.com/wordpress-mobile/WordPress-iOS/pull/16579#pullrequestreview-668961708?

## My Suggestion

My suggestion would be to keep things a bit simpler and not rely on locking. Clearing the active notices would most probably be enough. 

Most of the notices are dismissed automatically after a few seconds. IIRC, only Quick Start has permanent notices. So, dismissing only Quick Start notices would probably also be better. In that way, we don't unnecessarily prevent hiding important error notices like [failed uploads](https://github.com/wordpress-mobile/WordPress-iOS/blob/23f5bff82f8ea9dc118509d8db2b3309ad155430/WordPress/Classes/ViewRelated/Post/PostNoticeViewModel.swift#L47-L63). 

We can do this by exposing a `QuickStartTourGuide` function:

https://github.com/wordpress-mobile/WordPress-iOS/blob/23f5bff82f8ea9dc118509d8db2b3309ad155430/WordPress/Classes/ViewRelated/Blog/QuickStartTourGuide.swift#L198-L203

And then calling that before presenting the alert:

https://github.com/wordpress-mobile/WordPress-iOS/blob/23f5bff82f8ea9dc118509d8db2b3309ad155430/WordPress/Classes/ViewRelated/Blog/Blog%20Details/BlogDetailsViewController.m#L1186-L1191

A possible advantage of this is we're hiding the implementation of the “notices” since we're not directly accessing the `NoticeStore`. And also we won't be dismissing non-Quick Start notices. 

The resulting behavior seems to be similar unless I missed something. 😅 

https://user-images.githubusercontent.com/198826/120044876-f7b48000-bfcb-11eb-95e2-d34026f604a3.mp4

Here's my test branch if you'd like to dig in more: [`issue/16579-shiki-test-fix`](https://github.com/wordpress-mobile/WordPress-iOS/compare/fix/notice-overlap-for-site-icon...issue/16579-shiki-test-fix?expand=1).

## Another Issue

While testing this PR, I also noticed another dialog that will have to be fixed. 

https://user-images.githubusercontent.com/198826/120044921-10bd3100-bfcc-11eb-9364-31dec8e61c1b.mp4

---

## Final Notes

Ultimately, that's just my recommendation. I leave the decision to you and @frosty. 🙂

enejb:
Thanks for the suggestions @shiki and @frosty. They are very much appreciated. 

It's true with the locking and unlocking there is more code that needs to get added. Into the different place but I think overall it works better. Since the notices never overlap the ActionSheet.

We will need to update all the different action sheets in order to fix this for all of them.

> why didn't we consider clearing the notices as Frosty mentioned in #16579 (review)?

Dismissing the notice only works if the notice is already present. If you go and update the site tile and quickly click on the Site icon to change it. (Before the API response comes back) This will not remove the notice and you are stuck with the case where the notice overlaps the action sheet again. 
We have some user research and this has happened a couple of times to our users while they are going through the QuickStart flow and they get to this point and then they give up. 

I think the notices are nice to have but they can also hinder things. Maybe we should even make them dismissable with a swipe?

I think we should use the locking system for now. Until we have a better solution. :)

jkmassel:
We're cutting `17.5` today, so this PR is being bumped to `17.6`

WordPress/WordPressTest/Classes/Stores/NoticeStoreTests.swift
```diff
@@ -144,6 +144,36 @@ class NoticeStoreTests: XCTestCase {
...
+        dispatch(.unlock)
+        XCTAssertEqual(alpha, store.currentNotice)
+    }
+
+    func testCanNotDismissNoticesOnLockedStore() {
+        dispatch(.lock)
+        let alpha = Notice(title: "Alpha")
+        let bravo = Notice(title: "Bravo")
+        [alpha, bravo].forEach { dispatch(.post($0)) }
+        dispatch(.dismiss) // You can't dismiss notices in the locked store. Since that would send
```
frosty: Is this comment complete?

frosty (approved):
Thanks so much for the in-depth review, @shiki! @enejb, let's go ahead with this then – I just left one tiny comment, but I think otherwise we're good to go 👍 Thanks for your patience!

Issue #16567, opened 2021-05-24
Blogging Reminders: epilogue screen


Gio2018:
Add a screen that notifies users that their blogging reminders have been updated (see screenshot below)

<p align=center>
<img width="800" src="https://user-images.githubusercontent.com/34376330/119391614-2fea5480-bc94-11eb-8ff7-efa53eca1f90.png">
</p>

Issue #16565, opened 2021-05-24
Blogging Reminders: post-publishing bottom sheet


Gio2018:
After a blogger publishes their first post, we show a bottom sheet that prompts them to set their blogging goals (see screenshot below).
<p align=center>
<img height=400 src="https://user-images.githubusercontent.com/34376330/119381902-f52def80-bc86-11eb-9884-322329d25c89.png">
</p>

PR #16587, opened 2021-05-27
Blogging Reminders Settings Flow i1

frosty:
Refs #16565 #16566 #16567. This PR adds a first rough draft at the settings screens for blogging reminders. The main aim of this PR is to get a framework in place and allow us to share out the work of the individual screens and integrating them with the data. As such, the content of each view is not final so I'm not looking for too much feedback there at this stage. Also, I know that the positioning of elements on the initial screens isn't perfect – I'm going to be looking into that next.

blog-reminders-1

- Enable the blogging reminders feature flag
- Add some code to display it. I swapped out the action of the Stats quick action button around line 43 of `BlogDetailsViewController+Header.swift` with `self?.presentBloggingRemindersSettingsFlow()`

It's currently not hooked up anywhere so you'll need to do the following to test it:

**To test**

- Follow the steps above to add this code
- Build and run
- Get the UI to appear and step through it

## Regression Notes

None, this is all new code.

N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
@@ -1,0 +1,14 @@
+import UIKit
+import SwiftUI
+
+extension BlogDetailsViewController {
+    func presentBloggingRemindersSettingsFlow() {
+        // TODO: Check whether we've already presented this flow to the user. @frosty
+        let navigationController = BloggingRemindersNavigationController(rootViewController: BloggingRemindersSettingsContainerViewController(),
+                                                                         viewControllerDrawerPositions: [.collapsed, .expanded, .collapsed])
+
+        let bottomSheet = BottomSheetViewController(childViewController: navigationController,
+                                                    customHeaderSpacing: 0)
+        bottomSheet.show(from: self)
+    }
+}
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersNavigationController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersNavigationController.swift
@@ -1,0 +1,86 @@
+import UIKit
+
+class BloggingRemindersNavigationController: LightNavigationController {
+
+    private let viewControllerDrawerPositions: [DrawerPosition]
+
+    init(rootViewController: UIViewController, viewControllerDrawerPositions: [DrawerPosition]) {
+        self.viewControllerDrawerPositions = viewControllerDrawerPositions
+
+        super.init(rootViewController: rootViewController)
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override public var preferredContentSize: CGSize {
+        set {
+            viewControllers.last?.preferredContentSize = newValue
+            super.preferredContentSize = newValue
+        }
+        get {
+            guard let visibleViewController = viewControllers.last else {
+                return .zero
+            }
+
+            return visibleViewController.preferredContentSize
+        }
+    }
+
+    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        super.pushViewController(viewController, animated: animated)
+
+        updateDrawerPosition()
+    }
+
+    override func popViewController(animated: Bool) -> UIViewController? {
+        let viewController = super.popViewController(animated: animated)
+
+        updateDrawerPosition()
+
+        return viewController
+    }
+
+    private func updateDrawerPosition() {
+        let index = max(min(viewControllers.count-1, viewControllerDrawerPositions.count-1), 0)
+        let newPosition = viewControllerDrawerPositions[index]
+
+        if let viewController = viewControllers.last {
+            preferredContentSize = viewController.preferredContentSize
+        }
+        
+        if let bottomSheet = self.parent as? BottomSheetViewController, let presentedVC = bottomSheet.presentedVC {
+            presentedVC.transition(to: newPosition)
+        }
+    }
+}
+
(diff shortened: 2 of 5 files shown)


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersSettingsFlow.swift
```diff
@@ -0,0 +1,215 @@
...
+                Text(TextContent.settingsPrompt)
+                    .font(TextContent.settingsPromptFont)
+                    .fontWeight(.semibold)
+                    .lineLimit(nil)
+                    .multilineTextAlignment(.center)
+                Text(TextContent.settingsUpdatePrompt)
+                    .foregroundColor(.gray)
+                    .lineLimit(2)
+            }
+            HStack {
```
diegoreymendez, line 104: I'd consider making a component (custom view) out of this, since the day picker could eventually come handy for other purposes, and since it's going to help modularize the logic a bit.

That said - if we want to keep the ball rolling for now, I'm good with leaving it as it is in this PR.

diegoreymendez (approved):
Hey @frosty, nice work!

I added one comment about a possible improvement, but I think we should move ahead and merge this PR, and keep working on top of this.  The feature flag makes it a great option to make incremental improvements to this PR.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
```diff
@@ -0,0 +1,14 @@
...
+extension BlogDetailsViewController {
+    func presentBloggingRemindersSettingsFlow() {
+        // TODO: Check whether we've already presented this flow to the user. @frosty
+        let navigationController = BloggingRemindersNavigationController(rootViewController: BloggingRemindersSettingsContainerViewController(),
+                                                                         viewControllerDrawerPositions: [.collapsed, .expanded, .collapsed])
+
+        let bottomSheet = BottomSheetViewController(childViewController: navigationController,
+                                                    customHeaderSpacing: 0)
+        bottomSheet.show(from: self)
+    }
```
Gio2018, line 13: we might want to build a more general presenter logic in case we need to initiate the flow from more than on entry point. Does not need to be addressed in this PR, just remind to refactor this code when we do.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersSettingsFlow.swift
```diff
@@ -0,0 +1,215 @@
...
+                ForEach(days.indices, id: \.self) { index in
+                    Toggle(days[index].uppercased(), isOn: $toggles[index])
+                        .toggleStyle(CheckboxToggleStyle())
+                        .frame(minWidth: 0, maxWidth: .infinity)
+                }
+            }
+            Spacer()
+                .frame(maxHeight: .infinity)
```
Gio2018: Curious why this is necessary: doesn't `Spacer()` expand automatically as needed?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersNavigationController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    private func updateDrawerPosition() {
+        let index = max(min(viewControllers.count-1, viewControllerDrawerPositions.count-1), 0)
+        let newPosition = viewControllerDrawerPositions[index]
+
+        if let viewController = viewControllers.last {
+            preferredContentSize = viewController.preferredContentSize
+        }
+        
+        if let bottomSheet = self.parent as? BottomSheetViewController, let presentedVC = bottomSheet.presentedVC {
```
Gio2018, line 53: I wonder if we could do something different instead of relying on the responder chain? Just in case it changes at a later time, for any reason.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersSettingsFlow.swift
frosty: Yeah, I definitely think pulling it out makes sense – however this is likely to chance considerably in a follow-up PR (the days need to wrap lines and use real data), so I think it'll be fine to do it then. Thanks for the prompt!

Gio2018 (approved):
Hey @frosty , nice start!! Left a few code comments. Other than that, the animation between screens looks a bit odd right now, but I guess this is expected since you mentioned that you are not looking for detailed UI feedback for now. If so, I agree with @diegoreymendez: we can merge and keep the feature flag for now.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
frosty: 👍

frosty: We will need to for sure, as it will need to be presented from Site Settings too.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersNavigationController.swift
frosty: Possibly, I took this from the way the prepublishing flow currently works. I'll leave as-is for now, but we can always rework this!

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersSettingsFlow.swift
frosty: You're right, this is unecessary – removed!

PR #16601, opened 2021-06-01
Dismiss suggestions on space keystroke

guarani:
Addresses https://github.com/WordPress/gutenberg/issues/24425 (for iOS)

This PR allows users to easily dismiss the suggestion UI by typing a space character after triggering the UI. This means it will now be easier to add a `@` character to a post without triggering the mentions suggestions UI, or add a `+` character without triggering the cross-post suggestions UI.

I added this change to suggestions in the block editor, but didn't apply it to other parts of the app (e.g. comments, notification replies, etc) because that felt too far out-of-scope here. It would also make the testing steps much longer.

## To test:

1. Verify that in the block editor, typing a space after a `@` character dismisses the mentions suggestions list
2. Verify that in the block editor, typing a space after a `+` character dismisses the cross-positing suggestions list
3. Verify that mentions still works, ensuring that the intended recipient gets notified
4. Verify that cross-posting still works, ensuring that the intended recipient gets notified

## Regression Notes

This change could conceivably break suggestions (both mentions and cross-posting).

The above manual tests.

This change affects the UI so a UI test would be best, but we only add UI tests to core flows.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
@@ -114,6 +114,13 @@ extension GutenbergSuggestionsViewController: UITextFieldDelegate {
             return true
         }
         let searchWord = nsString.replacingCharacters(in: range, with: string)
+
+        let dismissSequence = suggestionType.trigger + " "
+        guard searchWord != dismissSequence else {
+            onCompletion?(.success(""))
+            return true
+        }
+
         if searchWord.hasPrefix(suggestionType.trigger) {
             suggestionsView.showSuggestions(forWord: searchWord)
         } else {
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
@@ -115,6 +115,16 @@ extension GutenbergSuggestionsViewController: UITextFieldDelegate {
         }
         let searchWord = nsString.replacingCharacters(in: range, with: string)
 
+        // This feels a bit hacky, so I'll explain what's being done here:
+        //     1. If the user types "@ ", the so-called "dismiss sequence",
+        //        the user probably typed the space to dismiss the suggestions UI.
+        //     2. So when this happens, we call the success handler and
+        //        pass in an empty string.
+        // We pass in an empty string because on the RN side here, the success
+        // handler inserts the @, followed by the username, followed by a space.
+        // Since we're essentially passing in an empty username, the result is
+        // that a "@ " is inserted into the post, which is the desired result.
+        // The same applies for cross-posts, with "+ " instead of "@ ".
         let dismissSequence = suggestionType.trigger + " "
         guard searchWord != dismissSequence else {
             onCompletion?(.success(""))
--- a/WordPress/Classes/ViewRelated/Suggestions/SuggestionsTableView.m
+++ b/WordPress/Classes/ViewRelated/Suggestions/SuggestionsTableView.m
@@ -246,8 +246,15 @@ CGFloat const STVSeparatorHeight = 1.f;
         self.searchText = word;
         if (self.searchText.length > 1) {
             NSString *searchQuery = [word substringFromIndex:1];
-            NSPredicate *predicate = [self predicateFor: searchQuery];
-            self.searchResults = [[self.suggestions filteredArrayUsingPredicate:predicate] mutableCopy];
+
+            // Users type a space after a trigger character to dismiss the suggestions UI
+            if ([searchQuery isEqualToString:@" "]) {
+                self.searchText = @"";
+                [self.searchResults removeAllObjects];
+            } else {
+                NSPredicate *predicate = [self predicateFor: searchQuery];
+                self.searchResults = [[self.suggestions filteredArrayUsingPredicate:predicate] mutableCopy];
+            }
         } else {
             self.searchResults = [self.suggestions mutableCopy];
         }
--- a/WordPress/Classes/ViewRelated/Suggestions/SuggestionsTableView.m
+++ b/WordPress/Classes/ViewRelated/Suggestions/SuggestionsTableView.m
@@ -246,15 +246,8 @@ CGFloat const STVSeparatorHeight = 1.f;
         self.searchText = word;
         if (self.searchText.length > 1) {
             NSString *searchQuery = [word substringFromIndex:1];
-
-            // Users type a space after a trigger character to dismiss the suggestions UI
-            if ([searchQuery isEqualToString:@" "]) {
-                self.searchText = @"";
-                [self.searchResults removeAllObjects];
-            } else {
-                NSPredicate *predicate = [self predicateFor: searchQuery];
-                self.searchResults = [[self.suggestions filteredArrayUsingPredicate:predicate] mutableCopy];
-            }
+            NSPredicate *predicate = [self predicateFor: searchQuery];
+            self.searchResults = [[self.suggestions filteredArrayUsingPredicate:predicate] mutableCopy];
         } else {
             self.searchResults = [self.suggestions mutableCopy];
         }


WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
```diff
@@ -114,6 +114,13 @@ extension GutenbergSuggestionsViewController: UITextFieldDelegate {
             return true
         }
         let searchWord = nsString.replacingCharacters(in: range, with: string)
+
+        let dismissSequence = suggestionType.trigger + " "
+        guard searchWord != dismissSequence else {
+            onCompletion?(.success(""))
```
guarani, line 130: This feels a bit hacky, so I'll explain what's being done here:
1. If the user types `"@ "`, the so-called "dismiss sequence", the user probably typed the space to dismiss the suggestions UI
2. So when this happens, we call the success handler and pass in an empty string. 

We pass in an empty string because on the RN side [here](https://github.com/WordPress/gutenberg/blob/c02903a4c7f964e96e27e0fae49df975f97e39ac/packages/rich-text/src/component/index.native.js#L502), the success handler inserts the `@`, followed by the username, followed by a space. Since we're essentially passing in an empty username, the result is that a `"@ "` is inserted into the post, which is the desired result.

guarani: The same applies for cross-posts, with `"+ "` instead of `"@ "`.

enejb: Thanks for the great explanation. Could we also add it to the code as a comment?
This will make it easier to understand in the future.

enejb (approved):
I tested this and it worked really well for me. Nicely done. 
Lets 🚢 it

enejb:
I think this PR is ready to merge. 

One thing that I found in my testing is that it is kind of hard to add a `@word` string that doesn't match any username. 
I think we should still make it easier somehow to add the `@word` to the post. Is that behaviour intentional?

WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
guarani: Good idea! Added in 8f511be4520ff52aa8c5f1c690838a379265d7f5

guarani:
> I think this PR is ready to merge.

Thanks!

> One thing that I found in my testing is that it is kind of hard to add a `@word` string that doesn't match any username.
> I think we should still make it easier somehow to add the `@word` to the post. Is that behaviour intentional?

One technique here is to type `@`, dismiss the suggestions UI by tapping outside it, and continue typing the username. I've recorded a video of this:

https://user-images.githubusercontent.com/1898325/120667675-b9034780-c45b-11eb-9617-4307c04fb08a.mov

Do you think this is a good user experience or could it be improved?

enejb:
Thanks for pointing out that tapping on the editor will get you back into this mode. 

This is what I saw. 

https://user-images.githubusercontent.com/115071/120691519-6e81cb00-c45b-11eb-8347-8823bc1af9ba.MP4

I think one improvement we could make is ( how I expected this would work is) if you type a word and then press space you would get back to the editor. Since that how I would have types it sequentially. `@word[space]` 

Not sure if this has been considered before?

guarani:
> I think one improvement we could make is ( how I expected this would work is) if you type a word and then press space you would get back to the editor. Since that how I would have types it sequentially. `@word[space]`
> 
> Not sure if this has been considered before?

Thanks for the video and explanation, I think I understand your concern better now. What prevents us from dismissing the suggestion UI when the user presses space after typing a `@` **followed by some characters** is that we also match display names, not just usernames.

So if a user suggestion is `john` (display name `John Smith`), we can't dismiss the UI when the user types `@john[space]` because they could be filtering by display name, which includes a space.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergSuggestionsViewController.swift
```diff
@@ -114,6 +114,23 @@ extension GutenbergSuggestionsViewController: UITextFieldDelegate {
...
+        //     2. So when this happens, we call the success handler and
+        //        pass in an empty string.
+        // We pass in an empty string because on the RN side here, the success
+        // handler inserts the @, followed by the username, followed by a space.
+        // Since we're essentially passing in an empty username, the result is
+        // that a "@ " is inserted into the post, which is the desired result.
+        // The same applies for cross-posts, with "+ " instead of "@ ".
+        let dismissSequence = suggestionType.trigger + " "
+        guard searchWord != dismissSequence else {
+            onCompletion?(.success(""))
```
mchowning, line 130: :wave: @guarani ! If I'm understanding this correctly, passing this as a "success" means that we'll [send off analytics indicating that the user successfully selected a suggestion](https://github.com/wordpress-mobile/WordPress-iOS/blob/bad64c922becf79ff52aa3126ef949ea9a6a40d6/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift#L914-L917), right? If that's what will happen, that doesn't feel right to me. Seems that we should be treating this as a failure case where the user did _not_ select a suggestion (at least for the purposes of the analytics).

I imagine that the reason you're using the success case is because we get the insertion of the "space" [for free](https://github.com/WordPress/gutenberg/blob/f1f63ab22da1994ab2a71056e092c8900f12467c/packages/rich-text/src/component/index.native.js#L495) when we do that, but maybe there are other reasons I'm not thinking of), whereas if we send a "failure" back to JS, then nothing gets inserted (so the space gets dropped). It feels like we have two options here:

1. Send back a "success" to javascript (so the space gets inserted), but send a "failure" to the analytics (that's what @SiobhyB [has proposed doing on Android](https://github.com/wordpress-mobile/WordPress-Android/pull/14793/files)); or
2. Add a way to send text-to-be-inserted (in this case a single space) back to javascript with a "failure".

Option 2 feels cleaner to me, but it may be over-engineering for the single case where we just want to send a space back since we already get that for free with an "empty" success, so I'm sympathetic to just going with Option 1's simpler approach. If we do that though, I think we may need some comments explaining why we're doing this. Curious to hear what you (and @SiobhyB ) think though.

guarani: Thanks for catching this, @mchowning! I completely overlooked the Tracks aspect. I'm going to revert this PR for the time-being while a fix is made. 

I think option 1 might work on iOS as well, but will take a look at option 2 as well.

guarani: I created a revert PR [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/16641).

SiobhyB: Thanks for looping me into the discussion here @mchowning. :)

If there's a possibility of more cases where certain keystrokes serve to both exit the UI and update the editor (like the one proposed in https://github.com/wordpress-mobile/gutenberg-mobile/issues/2943) then I'd definitely prefer to do this in the cleanest, most scalable way possible.

Are there key differences to how the JavaScript side would handle a "success" vs a "failure" message? I'd made the assumption that entering a space _was_ a success as far as the JavaScript/editor was concerned and am not yet clear what sending a failure would look like.

mchowning: > Are there key differences to how the JavaScript side would handle a "success" vs a "failure" message?

Good question. The only difference that I can imagine at this point is the fact that we want to add a space at the end of a successfully inserted mention, whereas we might not want to do that when a mention was not inserted. For example, _if_ we wanted to have a space exit the mention at any point in the mentions UI (which as Paul points out, [we probably do not want to do](https://github.com/wordpress-mobile/WordPress-iOS/pull/16601#issuecomment-854121540)), then we wouldn't want to add an "extra" space on after the "mention text" which already ends in a space. 

This is a bit of a made up example though since we don't actually want this behavior, and this is my concern with Option 2: we don't really know what other use case there might be that would need failure-with-text, so it's hard to know how to best design it. 

Not that I'm opposed to Option 2--I bias pretty strongly to finding the cleanest approach most often--but in this case I'm not sure whether it's worth much extra effort: your implementation of Option 1 on Android is quite simple and works perfectly. Doing option 1 on iOS probably won't be as simple though, so I'll be curious to see what Paul thinks after he digs in a bit.

SiobhyB: Ah, I see, thank you for that extra context!

guarani: I left a comment, https://github.com/wordpress-mobile/gutenberg-mobile/issues/2943. Once that's resolved I think we'll have a better picture of how to handle this.

guarani: > I left a comment, [wordpress-mobile/gutenberg-mobile#2943](https://github.com/wordpress-mobile/gutenberg-mobile/issues/2943). Once that's resolved I think we'll have a better picture of how to handle this.

Now that https://github.com/wordpress-mobile/gutenberg-mobile/issues/2943#issuecomment-869702797 is closed, we don't have another use case for designing a way to pass text back to the React Native world in the suggestions insertion "failure" case. I think [Option 1](https://github.com/wordpress-mobile/WordPress-iOS/pull/16601#discussion_r646606449), where we send a success event to React Native and a "failure" to analytics, is a reasonable solution here.

I'll go ahead with this on the iOS side and ping you @SiobhyB once this is ready just to sync up on merging these changes.

SiobhyB: Perfect, thank you @guarani!

guarani: FIY the iOS PR is up: https://github.com/wordpress-mobile/WordPress-iOS/pull/16770

PR #16603, opened 2021-06-01
Adds unit tests and improves BloggingRemindersScheduler.

diegoreymendez:
Adds unit tests to `BloggingRemindersScheduler`.

The `Schedule` now implements `Comparable` and `Equatable`.  This effectively means that the schedule is now can be compared and doesn't care about the ordering of the elements to know when two schedules are the same.

## To test:

Since the scheduler isn't wired and testable from within the App yet, just evaluate the changes visually and make sure the unit tests succeed in running.

## Regression Notes

None

PR submission checklist:


WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
```diff
@@ -1,5 +1,20 @@
...
+protocol NotificationScheduler {
+    func add(_ request: UNNotificationRequest, withCompletionHandler completionHandler: ((Error?) -> Void)?)
+    func removePendingNotificationRequests(withIdentifiers identifiers: [String])
+}
+
+extension UNUserNotificationCenter: NotificationScheduler {
+}
+
+/*
+class Test: UNUserNotificationCenter {
```
frosty: Can this be removed?

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
```diff
@@ -54,7 +84,7 @@ class BloggingRemindersScheduler {
 
     /// The notification center
```
frosty: We should update this comment to match

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
```diff
@@ -40,6 +66,10 @@ class BloggingRemindersScheduler {
         /// The default reminder hour.  In the future we may want to replace this constant with a more customizable approach.
         ///
         static let defaultHour = 10
+
+        static func < (lhs: BloggingRemindersScheduler.Weekday, rhs: BloggingRemindersScheduler.Weekday) -> Bool {
```
frosty, line 63: Do we need to implement `==` here as well?

frosty (approved):
This looks great, and tests pass. Just left a few small code comments, otherwise :shipit:

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
diegoreymendez: Yes

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
diegoreymendez: Swift default implementation of `Equatable` for enums covers what we need.

You can try this in a playground:

```swift
enum Test {
    case zero
    case uno
}

let first: Test = .zero
let second: Test = .uno
let third: Test = .uno

print(first == second) // false
print(first == third) // false
print(second == third) // true
```

diegoreymendez:
Hey @frosty, this PR is now ready for another look!

frosty (approved):

PR #16605, opened 2021-06-01
Merge Jetpack code freeze into develop

jkmassel:
This PR merges changes required for the Jetpack code freeze into `develop` (just version number stuff).

**To Test**
- If it passes CI, it's good to go

### DIFF
--- a/WordPress/WordPress.xcodeproj/project.pbxproj
+++ b/WordPress/WordPress.xcodeproj/project.pbxproj
@@ -23336,7 +23336,7 @@
 					"@executable_path/Frameworks",
 				);
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
-				MARKETING_VERSION = 0.2;
+				MARKETING_VERSION = "${VERSION_SHORT}";
 				OTHER_CFLAGS = (
 					"$(inherited)",
 					"-Wno-format-security",
@@ -23406,7 +23406,7 @@
 					"@executable_path/Frameworks",
 				);
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
-				MARKETING_VERSION = 0.2;
+				MARKETING_VERSION = "${VERSION_SHORT}";
 				OTHER_CFLAGS = (
 					"$(inherited)",
 					"-Wno-format-security",
@@ -23475,7 +23475,7 @@
 					"@executable_path/Frameworks",
 				);
 				LIBRARY_SEARCH_PATHS = "$(inherited)";
-				MARKETING_VERSION = 0.2;
+				MARKETING_VERSION = "${VERSION_SHORT}";
 				OTHER_CFLAGS = (
 					"$(inherited)",
 					"-Wno-format-security",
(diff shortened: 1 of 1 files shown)


WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -23305,7 +23286,7 @@
 				CODE_SIGN_ENTITLEMENTS = Jetpack/JetpackDebug.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
-				CURRENT_PROJECT_VERSION = "Set by Build Phase";
+				CURRENT_PROJECT_VERSION = "${VERSION_LONG}";
```
AliSoftware, line 23289: @jkmassel Not sure what to expect here, but is this intentional?

Why was the build phase removed above? did we recently change the way we handle version numbers in Jetpack very recently?
And if so, where are `VERSION_SHORT` and `VERSION_LONG` defined? Usually we define them via `Version.public.xcconfig` files and such, but I don't see such a file being modified in the diff of this PR to bump the version number here like I'd then expect?

mokagio: @jkmassel: @AliSoftware's onto something here:

The project does have `.xcconfig` files for Jetpack (e.g. [`Jetckap.release.xcconfig`](https://github.com/wordpress-mobile/WordPress-iOS/blob/d6a4e9ca9cf191e6636b534ff385fb00df89babb/config/Jetpack.release.xcconfig) and the Jetpack target makes use of them:

Screen Shot 2021-06-04 at 11 35 56 am

But, there's no custom version defined for Jetpack and as such, it falls back to the WordPress version number:

Screen Shot 2021-06-04 at 11 32 57 am

AliSoftware: I'm starting to wonder if this isn't intentional after all, given this mention in p5T066-2mX-p2:

> As for version numbers, Jetpack and WordPress will use the same version number. This means that Jetpack 17.5 is powered by pretty much the same code as WordPress 17.5.

But I'd still like confirmation as the state of where we are with Jetpack release process is still not clear to me nor yet P2'd black on white anywhere…

jkmassel: Ah yeah sorry – the Jetpack app will use exactly the same versioning as the WP one for now

AliSoftware (approved):
Ok :shipit:  if Jetpack and WP are gonna share the same version number,  makes sense then.

_Short explanation: `Jetpack.*.xcconfig` – which the Jetpack target is based on – doesn't import `Version.*.xcconfig` nor declare any `VERSION_SHORT` nor `VERSION_LONG`, but the  configuration set on the whole Xcode project (from which all targets depend on) as the `#include` on `Version.*.xcconfig` so that's where the Jetpack target will pick its version numbers from_ 👍

WordPress/WordPress.xcodeproj/project.pbxproj
mokagio: >  the Jetpack app will use exactly the same versioning as the WP one for now

"For now." Apple doesn't let us go backwards with version numbers, understandably. So I guess we'll stick with this "forever" 😄

mokagio: I mean, obviously the two versions could diverge at some point. I'm just saying there's no going back to a more traditional versioning starting from 1.0 🤷‍♂️ It's just a number...

PR #16615, opened 2021-06-02
Try using Xcode 12.5

jkmassel:
Updates the project to use Xcode 12.5

To test:
- Ensure CI passes

## Regression Notes
n/a

n/a

n/a
PR submission checklist:


.xcversion
```diff
@@ -0,0 +1 @@
+12.5
```
jkmassel, line 1: This file is basically `.ruby-version` for Xcode – some details on this can be found over at https://github.com/xcpretty/xcode-install#xcode-version

.circleci/config.yml
```diff
@@ -21,15 +21,15 @@ parameters:
     default: false
 
 xcode_version: &xcode_version
-  xcode-version: "12.1.0"
+  xcode-version: "12.5.0"
```
mokagio, line 24: Wow, that's quite the jump. TIL.

.xcversion
mokagio: 👍 There's also tools like [`xcenv`](https://github.com/xcenv/xcenv) that read it and perform the same shimming that an `rbenv` or `nvm` would.

Wild idea. When I saw this I though "it's 100% sure to diverge from the CI config at some point." 🤣 "Wouldn't it be good if CI read it instead?".

In Buildkite, you can "[pipe build pipelines to the command allowing you to create scripts that generate dynamic pipelines](https://buildkite.com/docs/agent/v3/cli-pipeline#usage)". That is we could have a templated YML file in which we inject the value read from `.xcversion` so there's a single source of truth.

That obviously assumes that we want CI and local environment to be using the same toolchain. That sounds like a good idea, but might not be feasible at times, e.g. when migrating between major versions, something that is easy locally but hard on CI because the provider needs to support the new macOS version etc.

mokagio (approved):
All green in both CircleCI and Buildkite 🎉

.xcversion
jkmassel: Yeah! Once we've fully out of CircleCI, I'd like this file to be the source of truth for sure! @AliSoftware mentioned a good solution to this as well that I don't fully recall 😅 (it might actually be the same one!!)

AliSoftware: The solutions I mentioned probably were:
 - Using `.xcversion` file indeed. That file alone doesn't really do much if not read by other scripts though, but it's a nice convention for naming that file.
 - Using [the `xcversion` action](https://docs.fastlane.tools/actions/xcversion/) in our Fastfile to make fastlane enforce the use of a particular version of Xcode when invoking xcodebuild, provided that version of Xcode is installed on the machine (and errors if not).
    - So we could use `xcversion(version: File.read('.xcversion').chomp)` in our Fastfile for example to read the version requirement from that file.
    - Under the hood, this action uses `mdfind` and similar tricks to find, amongst the Xcode versions installed in your machine, the path of the `Xcode.app` that matches that version requirement. Then once it found the path to the right `Xcode.app` instance, it sets the `DEVELOPER_DIR` env var with that path. This is an env var that is recognized by `xcodebuild` and `xcrun` and makes `xcrun` uses that specific version of Xcode and its build tools (overriding the one that you might have using `xcode-select`). Because it uses this env var trick, that means that this doesn't require `sudo` and only affect this run of `xcodebuild` without changing your system-wide settings.

mokagio:
~~Merging this in for 17.6. I hope I won't kick myself later today, after discovering that for some reason the Xcode 12.5 toolchain can't build the app for TestFlight or App Center on CircleCI 😅 ~~

Actually... It just occurred to me we might want to:

1. Announce support to the devs
2. Coordinate this with WooCommerce

Which I guess is why @jkmassel didn't merge this already. So maybe it's best to wait to merge this one for 17.7.

PR #16621, opened 2021-06-03
Run Fastfile and Matchfile past Rubocop

mokagio:
Between my work to migrate secrets outside of the repo and my upcoming rotation as release manager, I expect to be touching the `Fastfile` a fair bit.

My Vim setup bombarded me with warnings on Ruby style used in the file, so I thought I'd take the time to address them.

Now, you could argue that I should have disabled the warnings on my end, but I'd say that, because the `Fastfile` _is_ a Ruby file, using a standard style (standard according to Rubocop, at least) is appropriate.

To test:

I struggle to come up with an efficient way to test these changes. Given they're all coding style related, and that the majority of them was automated via `rubocop -a` (see the first commit, 9ef77b4) reviewing this commit-by-commit should suffice.

## Regression Notes
N.A. See testing details above.

N.A.

N.A

PR submission checklist:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,12 @@
+Metrics/BlockLength:
+  Exclude:
+    - fastlane/Fastfile
+
+Style/AsciiComments:
+  Exclude:
+    - fastlane/Fastfile
+
+Layout/LineLength:
+  Exclude:
+    - fastlane/Fastfile
+
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,3 +10,6 @@ Layout/LineLength:
   Exclude:
     - fastlane/Fastfile
 
+Naming/FileName:
+  Exclude:
+    - fastlane/Matchfile
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,7 +14,8 @@ DERIVED_DATA_PATH = File.join(PROJECT_ROOT_FOLDER, 'DerivedData')
 BUILD_PRODUCTS_PATH = File.join(PROJECT_ROOT_FOLDER, 'Artifacts')
 
 # Env file paths to load
-USER_ENV_FILE_PATH = File.join(Dir.home, '.wpios-env.default')
+ENV_FILE_NAME = '.wpios-env.default'
+USER_ENV_FILE_PATH = File.join(Dir.home, ENV_FILE_NAME)
 PROJECT_ENV_FILE_PATH = File.join(PROJECT_ROOT_FOLDER, '.configure-files', 'project.env')
 
 # Other defines used across multiple lanes
@@ -49,7 +50,9 @@ before_all do |lane|
   unless lane == :test_without_building
     # Check that the env files exist
     unless is_ci || File.file?(USER_ENV_FILE_PATH)
-      UI.user_error!("~/.wpios-env.default not found: Please copy fastlane/env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values")
+      example_path = 'fastlane/env/user.env-example '
+      msg = "#{ENV_FILE_NAME} not found: Please copy #{example_path} to #{USER_ENV_FILE_PATH} and fill in the values"
+      UI.user_error!(msg)
     end
     unless File.file?(PROJECT_ENV_FILE_PATH)
       UI.user_error!('project.env not found: Make sure your configuration is up to date with `rake dependencies`')
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -551,8 +551,7 @@ platform :ios do
 
     ios_merge_translators_strings(strings_folder: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources'))
 
-    extract_framework_translations_script_path = File.join(File.join(PROJECT_ROOT_FOLDER, 'Scripts',
-                                                                     'extract-framework-translations.swift'))
+    extract_framework_translations_script_path = File.join(PROJECT_ROOT_FOLDER, 'Scripts', 'extract-framework-translations.swift')
     sh(extract_framework_translations_script_path)
 
     gym(
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -778,7 +778,7 @@ platform :ios do
       res = file.find { |line| line =~ /^(?!\s*#)(?=.*\bgutenberg\b).*(\bcommit|tag\b){1}.+/ }
     end
 
-    UI.user_error!("Can't find any reference to Gutenberg!") unless res.length != 0
+    UI.user_error!("Can't find any reference to Gutenberg!") if res.empty?
     if res.include?('commit')
       UI.user_error!("Gutenberg referenced by commit!\n#{res}") unless UI.interactive?
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -735,14 +735,14 @@ platform :ios do
   desc 'Run tests without building'
   lane :test_without_building do |options|
     # Find the referenced .xctestrun file based on its name
-    buildProductsPath = File.join(DERIVED_DATA_PATH, 'Build', 'Products')
+    build_products_path = File.join(DERIVED_DATA_PATH, 'Build', 'Products')
 
-    testPlanPath = Dir.glob(File.join(buildProductsPath, '*.xctestrun')).select do |e|
+    test_plan_path = Dir.glob(File.join(build_products_path, '*.xctestrun')).select do |e|
       e.include?(options[:name])
     end.first
 
-    unless !testPlanPath.nil? && File.exist?((testPlanPath))
-      UI.user_error!("Unable to find .xctestrun file at #{buildProductsPath}")
+    unless !test_plan_path.nil? && File.exist?((test_plan_path))
+      UI.user_error!("Unable to find .xctestrun file at #{build_products_path}")
     end
 
     multi_scan(
@@ -751,7 +751,7 @@ platform :ios do
       device: options[:device],
       deployment_target_version: options[:ios_version],
       test_without_building: true,
-      xctestrun: testPlanPath,
+      xctestrun: test_plan_path,
       try_count: options[:try_count],
       output_directory: File.join(PROJECT_ROOT_FOLDER, 'build', 'results'),
       result_bundle: true
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -76,7 +76,7 @@ platform :ios do
   ENV['PUBLIC_CONFIG_FILE'] = File.join(PROJECT_ROOT_FOLDER, 'config', 'Version.Public.xcconfig')
   ENV['INTERNAL_CONFIG_FILE'] = File.join(PROJECT_ROOT_FOLDER, 'config', 'Version.internal.xcconfig')
   ENV['DOWNLOAD_METADATA'] = './fastlane/download_metadata.swift'
-  ENV['PROJECT_ROOT_FOLDER'] = PROJECT_ROOT_FOLDER
+  ENV['PROJECT_ROOT_FOLDER'] = PROJECT_ROOT_FOLDER + '/'
   ENV['APP_STORE_STRINGS_FILE_NAME'] = 'AppStoreStrings.po'
 
   ########################################################################
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -3,37 +3,35 @@ fastlane_require 'xcodeproj'
 fastlane_require 'dotenv'
 fastlane_require 'open-uri'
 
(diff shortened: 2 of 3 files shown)


fastlane/Fastfile
```diff
@@ -49,7 +50,9 @@ before_all do |lane|
   unless lane == :test_without_building
     # Check that the env files exist
     unless is_ci || File.file?(USER_ENV_FILE_PATH)
-      UI.user_error!("~/.wpios-env.default not found: Please copy fastlane/env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values")
+      example_path = 'fastlane/env/user.env-example '
+      msg = "#{ENV_FILE_NAME} not found: Please copy #{example_path} to #{USER_ENV_FILE_PATH} and fill in the values"
+      UI.user_error!(msg)
```
mokagio, line 55: I did this before disabling the line length violation. Still, it DRYes a bit (see `ENV_FILE_NAME`) and I don't think it hurst the code, so I decide to leave it.

fastlane/Fastfile
```diff
@@ -778,7 +778,7 @@ platform :ios do
       res = file.find { |line| line =~ /^(?!\s*#)(?=.*\bgutenberg\b).*(\bcommit|tag\b){1}.+/ }
     end
 
-    UI.user_error!("Can't find any reference to Gutenberg!") unless res.length != 0
+    UI.user_error!("Can't find any reference to Gutenberg!") if res.empty?
```
mokagio: First Rubocop suggested to use `res.empty? == false` instead of `res.length != 0`, then to flip `unless res.empty? == false` to `if res.empty?`.

I'm spelling this out to make sure it holds logically, as I'm painfully aware that meddling with if statements can backfire 😅

.rubocop.yml
```diff
@@ -0,0 +1,15 @@
...
+  Exclude:
+    - fastlane/Fastfile
+
+Layout/LineLength:
+  Exclude:
+    - fastlane/Fastfile
+
+Naming/FileName:
+  Exclude:
+    - fastlane/Matchfile
```
mokagio, line 15: I can only guess that `Fastfile` is already excluded by the rule, while `Matchfile` isn't. 

I'd love to look into it on the Rubocop end and see if I can address it.

jkmassel: IMHO we should just get rid of the `Matchfile` anyway, so that would also be a solution 😁

fastlane/Fastfile
```diff
@@ -63,27 +66,27 @@ before_all do |lane|
...
+  # Environment
+  ########################################################################
+  Dotenv.load(USER_ENV_FILE_PATH)
+  Dotenv.load(PROJECT_ENV_FILE_PATH)
+  ENV[GHHELPER_REPO = 'wordpress-mobile/wordpress-iOS']
+  ENV['PROJECT_NAME'] = 'WordPress'
+  ENV['PUBLIC_CONFIG_FILE'] = File.join(PROJECT_ROOT_FOLDER, 'config', 'Version.Public.xcconfig')
+  ENV['INTERNAL_CONFIG_FILE'] = File.join(PROJECT_ROOT_FOLDER, 'config', 'Version.internal.xcconfig')
+  ENV['DOWNLOAD_METADATA'] = './fastlane/download_metadata.swift'
+  ENV['PROJECT_ROOT_FOLDER'] = PROJECT_ROOT_FOLDER
```
jkmassel: I think this change might break things (internally IIRC there are assumptions that the `PROJECT_ROOT_FOLDER` env var ends in a trailing slash).

fastlane/Fastfile
```diff
@@ -330,15 +335,16 @@ import "./ScreenshotFastfile"
...
-  desc "Builds and uploads for distribution"
-  lane :build_and_upload_release do | options |
+  desc 'Builds and uploads for distribution'
+  lane :build_and_upload_release do |options|
     ios_build_prechecks(skip_confirm: options[:skip_confirm],
-      internal: options[:beta_release],
-      external: true)
-    ios_build_preflight()
+                        internal: options[:beta_release],
+                        external: true)
```
jkmassel, line 342: Nit: I find these multi-line method calls without the closing brace on its own line easy to break (and messy in diffs if we need to add/remove arguments).

Is putting the closing branch on its own line rubocop-compliant?

fastlane/Fastfile
```diff
@@ -533,58 +538,59 @@ import "./ScreenshotFastfile"
...
+    update_translations_script_path = File.join(PROJECT_ROOT_FOLDER, 'Scripts', 'update-translations.rb')
+    sh(update_translations_script_path, 'review', 'current')
+    sh(update_translations_script_path, 'review', 'waiting')
+    sh(update_translations_script_path, 'review', 'fuzzy')
 
-    ios_merge_translators_strings(strings_folder: File.join( PROJECT_ROOT_FOLDER, "WordPress", "Resources"))
+    ios_merge_translators_strings(strings_folder: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources'))
 
-    extract_framework_translations_script_path = File.join( File.join( PROJECT_ROOT_FOLDER, "Scripts", "extract-framework-translations.swift"))
+    extract_framework_translations_script_path = File.join(File.join(PROJECT_ROOT_FOLDER, 'Scripts',
```
jkmassel: If we're here anyway – do we need the nested `File.join`?

jkmassel (changes requested):
I saw one potential breaking change, but in general this looks solid, and I like the idea of maintaining according to a standard set of rules!

fastlane/Fastfile
jkmassel: It reads that it's doing the same thing to me!

fastlane/Fastfile
jkmassel: I'm not sure we need to disable the line length violation – splitting method calls onto multiple lines does improve readability anyway 🤔 

WDYT?

fastlane/Fastfile
mokagio: Good point. When I made this change in isolation in 52ee45a, I [grepped for `PROJECT_ROOT_FOLDER`](https://github.com/wordpress-mobile/release-toolkit/search?q=PROJECT_ROOT_FOLDER&type=).

```
➜ g grep PROJECT_ROOT
lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_download_translations_action.rb:        res_dir = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', params[:res_dir])
lib/fastlane/plugin/wpmreleasetoolkit/actions/android/android_update_release_notes.rb:        path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'RELEASE-NOTES.txt')
lib/fastlane/plugin/wpmreleasetoolkit/actions/ios/ios_update_release_notes.rb:        path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'RELEASE-NOTES.txt')
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb:        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb:            files: File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'build.gradle'),
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb:        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_git_helper.rb:          res_dir = File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'src', 'main', 'res')
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb:          gradle_file_path = File.join(ENV['PROJECT_ROOT_FOLDER'] || '.', 'build.gradle')
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb:        # @env PROJECT_ROOT_FOLDER The path to the root of the project (the folder containing the `.git
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb:          UI.user_error!("You need to set the \`PROJECT_ROOT_FOLDER\` environment variable to the path
lib/fastlane/plugin/wpmreleasetoolkit/helper/android/android_version_helper.rb:          File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'build.gradle')
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:          files_list = [File.join(ENV['PROJECT_ROOT_FOLDER'], 'config', '.')]
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:            files_list.append File.join(ENV['PROJECT_ROOT_FOLDER'], ENV['PROJECT_NAME'], 'Resources', ENV['APP_STOR
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/localize.py")
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:        # @env PROJECT_ROOT_FOLDER The path to the git root of the project
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:          Action.sh("cd #{get_from_env!(key: 'PROJECT_ROOT_FOLDER')} && ./Scripts/update-translations.rb")
lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb:          project_root = get_from_env!(key: 'PROJECT_ROOT_FOLDER')
lines 1-19/19 (END)
```

AFAICS, all the instances are used within `File.join` which handles the `/`. That includes the [assignment to `project_root`](https://github.com/wordpress-mobile/release-toolkit/blob/develop/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb#L65-L70) in the last line of the `git grep` output.

Can you see anything that I missed? Otherwise, how do you feel about risking it and see if it breaks anything? (I could do it once I roll over as release manager)

I'm saying it mainly because removing the `/` is the right thing to do™

fastlane/Fastfile
mokagio: Totally!

I'm guessing it's an opt-in kind of setting? Otherwise it would have told me about it I'd say

fastlane/Fastfile
mokagio: 😅 good point

fastlane/Fastfile
mokagio: Yeah true. I mainly disabled it because of long string lines like https://github.com/wordpress-mobile/wordpress-ios/blob/fastlane-rubocop/fastlane/Fastfile#L419.

There's probably a Ruby way to split the string without messing up the horizontal line length 🤔

.rubocop.yml
mokagio: +1.

Maybe something I'll get to do over the next few weeks. Together with folding the `Deliverfile` and standalone `deliver` call into a lane to make it consistent with the rest of the workflow.

mokagio:
@jkmassel I requested a new review as per my comment [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/16621#discussion_r647111063).

fastlane/Fastfile
jkmassel: > how do you feel about risking it and see if it breaks anything

I'd be ok with it, but then we're committing to a major version bump of this library – do we want that?

IMHO, while it is more correct to adjust this, I'd prefer to make a larger change to not require these global variables anymore (even if it's "not right" for a while).

If you're ok with bumping the major version we can do the change though 😃

jkmassel: IMHO it's a cosmetic issue, so while it's not correct, I'd _personally_ prefer to address the structural problem (global variables) in the release toolkit? I'd be concerned about the potential for things silently breaking during a transition.

That said, you're the one who'd have to catch them, not me, so maybe just give it a think and feel free to go either way? 🙂

jkmassel (approved):
Approved to avoid blocking, but left some thoughts.

fastlane/Fastfile
mokagio: I was tempted to say "Yeah no worries. I checked the code and I'm sure that won't be any problems." Then I had visions of broken code freezes and tears of frustration.

Reverted! fb23430 (🙌 past me for small atomic commits)

Thank you for being wise.

PR #16631, opened 2021-06-03
Username picker loader UI change

### LINKED ISSUE #8761: Sign up: Username changer loading shouldn't be a modal
screen shot 2018-02-28 at 14 59 44

When the username list loads, it shouldn't use a modal view with the loader, as the above.

The change could be done at two levels.

### Level 1

We don't show anything while loading: the list is pre-populated with the existing username (thus, it has at least one item, selected) while loads the first time, and for loads while searching the list simply updates once the API is returned.

### Level 2

If we want explicit indicators, here's a design for it:

mobile-username-picker-loading-states-i5

nikola-milicevic:
Fixes [#8761](https://github.com/wordpress-mobile/WordPress-iOS/issues/8761)

### Description:
This PR replaces modal loader for the username picker, with the loading indicator located in the search cell, which is shown while fetching available usernames.

Notes:

- The original issue was related to the username picker in signup flow for the new user, but it's used in 'Account settings' screen for the existing user as well, so that's fixed too.
- The fix for magnifying glass icon is also included. It was almost invisible, now it has proper color (both in light and dark mode).

### To test:
New user:
1. Sign up with email
2. Tap on sign up link from email
3. Account details screen is presented.
4. Tap on 'Username'
5. Screen with username picker is presented with desired behaviour

Existing user:
1. Log in
2. On 'My Site' tab tap on Profile button in the top right corner
3. Tap on 'Account Settings'
4. Tap on 'Username'
5. Screen with username picker is presented with desired behaviour

### Video:
(disregard 'We had trouble loading data' blue bar on the first screen)

https://user-images.githubusercontent.com/12729242/120724878-f7891880-c4d4-11eb-90ca-6a36da5e33f5.mp4

### Regression Notes
n/a

n/a

n/a

### PR submission checklist:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -206,7 +206,7 @@ abstract_target 'Apps' do
 
     pod 'Gridicons', '~> 1.1.0'
 
-     pod 'WordPressAuthenticator', '~> 1.37.0'
+    pod 'WordPressAuthenticator', '~> 1.38.0-beta-3'
     # While in PR
     # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
     # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
--- a/Podfile
+++ b/Podfile
@@ -206,7 +206,7 @@ abstract_target 'Apps' do
 
     pod 'Gridicons', '~> 1.1.0'
 
-    pod 'WordPressAuthenticator', '~> 1.38.0-beta-3'
+    pod 'WordPressAuthenticator', '~> 1.38.0-beta.3'
     # While in PR
     # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
     # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
--- a/WordPress/Classes/ViewRelated/Me/My Profile/Change Username/ChangeUsernameViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/My Profile/Change Username/ChangeUsernameViewController.swift
@@ -57,14 +57,12 @@ private extension ChangeUsernameViewController {
         viewModel.suggestionsListener = { [weak self] state, suggestions, reloadAllSections in
             switch state {
             case .loading:
-                let cell = self?.tableView.cellForRow(at: IndexPath(row: 0, section: 1)) as? SearchTableViewCell
-                cell?.showLoader()
+                self?.showLoader(true)
             case .success:
                 if suggestions.isEmpty {
                     WPAppAnalytics.track(.accountSettingsChangeUsernameSuggestionsFailed)
                 }
-                let cell = self?.tableView.cellForRow(at: IndexPath(row: 0, section: 1)) as? SearchTableViewCell
-                cell?.hideLoader()
+                self?.showLoader(false)
                 self?.suggestions = suggestions
                 self?.reloadSections(includingAllSections: reloadAllSections)
             default:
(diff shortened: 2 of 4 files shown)


WordPress/Classes/ViewRelated/NUX/SignupUsernameTableViewController.swift
```diff
@@ -279,3 +281,23 @@ extension SignupUsernameTableViewController {
         }
     }
 }
+
+extension SearchTableViewCell {
```
leandroalonso: I wonder if we should add these methods to the `WordPressAuthenticator` lib. Wdyt?

leandroalonso (commented):
@nikola-milicevic works as described! Left a minor comment, let me know what your thoughts are.

I also find a minor issue: when toggling between light/dark mode the magnifying glass icon doesn't change it's color. Not a big issue tbh, so up to you if you want to fix it.

WordPress/Classes/ViewRelated/NUX/SignupUsernameTableViewController.swift
nikola-milicevic: Yeah, totally agree. Should I do it?

nikola-milicevic:
> @nikola-milicevic works as described! Left a minor comment, let me know what your thoughts are.
> 
> I also find a minor issue: when toggling between light/dark mode the magnifying glass icon doesn't change it's color. Not a big issue tbh, so up to you if you want to fix it.

Thanks for review! Yes, I used the same image for both light and dark mode, since it looked okay. If you think other color might be better I can tweak it a bit.

WordPress/Classes/ViewRelated/NUX/SignupUsernameTableViewController.swift
leandroalonso: Yes, please! :)

WordPress/Classes/ViewRelated/Me/My Profile/Change Username/ChangeUsernameViewController.swift
```diff
@@ -55,12 +57,14 @@ private extension ChangeUsernameViewController {
         viewModel.suggestionsListener = { [weak self] state, suggestions, reloadAllSections in
             switch state {
             case .loading:
-                SVProgressHUD.show(withStatus: Constants.Alert.loading)
+                let cell = self?.tableView.cellForRow(at: IndexPath(row: 0, section: 1)) as? SearchTableViewCell
```
leandroalonso: We have multiple usages of this code all around, what do you think about extracting it to a method?

leandroalonso (changes requested):
@nikola-milicevic One small last request!

WordPress/Classes/ViewRelated/Me/My Profile/Change Username/ChangeUsernameViewController.swift
nikola-milicevic: Sure!

Podfile
```diff
@@ -206,7 +206,7 @@ abstract_target 'Apps' do
 
     pod 'Gridicons', '~> 1.1.0'
 
-     pod 'WordPressAuthenticator', '~> 1.37.0'
+    pod 'WordPressAuthenticator', '~> 1.38.0-beta-3'
```
emilylaguna: Minor nitpick. 

```suggestion
    pod 'WordPressAuthenticator', '~> 1.38.0-beta.3'
```

emilylaguna (changes requested):
👋  @nikola-milicevic! 

Overall this looks good! I tested on iOS 13 and iOS 14 both on iPad and iPhone with both LTR (left to right) and RTL (right to left) languages, and dark and light modes and didn't experience any issues. 

I left a minor comments below.

nikola-milicevic:
@emilylaguna @leandroalonso thanks for the comments, I updated the code!

emilylaguna (changes requested):
Thanks for making the change!

`showLoader(true)` and `showLoader(false)` looks a bit weird, what do you think?

nikola-milicevic:
> Thanks for making the change!
> 
> `showLoader(true)` and `showLoader(false)` looks a bit weird, what do you think?

Well yes, but `toggleLoaderVisibility(isHidden: Bool)` is a bit too long I think. Do you have some suggestion?

emilylaguna:
> but toggleLoaderVisibility(isHidden: Bool) is a bit too long I think

True. 

> Do you have some suggestion?

What about creating an extension with methods for showing and hiding the loader?

nikola-milicevic:
> What about creating an extension with methods for showing and hiding the loader?

Thanks for the suggestion. I updated the code. Let me know what you think :)

emilylaguna (approved):
Looks good to me! 🚀

PR #16635, opened 2021-06-04
Update release-toolkit to 1.1.0 + add update check

AliSoftware:
### What

* Update toolkit to 1.1.0
* Add a call to its new `check_for_toolkit_updates` action in `before_all`
* Bonus: Also fixed the `get_pullrequests_list` lane while I was at it

### To test

* Run any lane that doesn't risk create side effect.
   * I chose to use the `get_pullrequests_list` for testing this (which is when I realized it was broken and fixed it 😛 )
* Confirm it checks for release toolkit updates, and that it doesn't find any

### Notes

Making this PR targeting the `release` branch because that's where you will be the next time you run the toolkit lanes, so figured it made more sense.

Also, the changes in `Fastfile` include a change in indentation, so it'll be easier to review if you turn "ignore whitespaces" on.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -331,7 +331,7 @@ DEPENDENCIES
   fastlane-plugin-appcenter (~> 1.8)
   fastlane-plugin-sentry
   fastlane-plugin-test_center
-  fastlane-plugin-wpmreleasetoolkit (~> 1.0)
+  fastlane-plugin-wpmreleasetoolkit (~> 1.1)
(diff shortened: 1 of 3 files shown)


jkmassel (approved):
Works for me – I don't even know if we use that lane, but at least now it's fixed!! 😁

AliSoftware:
I don't think we ever use it indeed… but it's nice to have it around for testing dummy stuff like this PR 😄

PR #16640, opened 2021-06-05
Add Buildkite UI Tests

jkmassel:
Adds UI Testing in Buildkite

**To Test:**
- Ensure UI tests pass in CircleCI and Buildkite

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -39,3 +39,30 @@ steps:
     depends_on: "build"
     agents:
       os: "macOS"
+
+  #################
+  # UI Tests
+  #################
+  - label: "🧪 UI Tests (iPhone)"
+    command: |
+      buildkite-agent artifact download build-products.tar .
+      tar -xf build-products.tar
+      bundle install
+      bundle exec pod install
+      rake mocks &
+      bundle exec fastlane test_without_building name:WordPressUITests try_count:3 device:"iPhone 11" ios-version:"14.1"
+    depends_on: "build"
+    agents:
+      os: "macOS"
+
+  - label: "🧪 UI Tests (iPad)"
+    command: |
+      buildkite-agent artifact download build-products.tar .
+      tar -xf build-products.tar
+      bundle install
+      bundle exec pod install
+      rake mocks &
+      bundle exec fastlane test_without_building name:WordPressUITests try_count:3 device:"iPad Air (4th generation)" ios-version:"14.1"
+    depends_on: "build"
+    agents:
+      os: "macOS"


.buildkite/pipeline.yml
```diff
@@ -39,3 +39,30 @@ steps:
     depends_on: "build"
     agents:
       os: "macOS"
+
+  #################
+  # UI Tests
+  #################
```
mokagio, line 45: Annoying nitpick: Is this comment header necessary? 😝 

More serious rationale: I can see there are two UI tests steps and I can see why we might want to group them together. But, I don't think the comment header makes them more discoverable—if I `Cmd F` for "UI Tests", I'll find the step anyways.

.buildkite/pipeline.yml
```diff
@@ -39,3 +39,30 @@ steps:
...
+    command: |
+      buildkite-agent artifact download build-products.tar .
+      tar -xf build-products.tar
+      bundle install
+      bundle exec pod install
+      rake mocks &
+      bundle exec fastlane test_without_building name:WordPressUITests try_count:3 device:"iPhone 11" ios-version:"14.1"
+    depends_on: "build"
+    agents:
+      os: "macOS"
```
mokagio, line 56: Is it necessary to specify the agent config at this level when it's already the default in the pipeline config?

mokagio (approved):
CI is happily running all the tests

Screen Shot 2021-06-07 at 10 57 04 am

And they're actually passing (so no false positive due to us still learning how to use Buildkite 👍 )

Screen Shot 2021-06-07 at 10 57 43 am

---

~~Out of scope but I noticed there is no `wait` step between the first build step and the rest of the steps in the pipeline.~~

~~My Buildkite-foo is a bit rusty, but IIRC and also going by [the example pipeline](https://buildkite.com/docs/pipelines/defining-steps#example-pipeline), tests are executed in parallel, unless otherwise specified.~~

~~The way I'm reading this pipeline, if we had 4 agents, all steps would run at the same time and that would result in the test steps either downloading the artifact to run from a previous build (assuming the `artifact` command stores files outside the scope of a build lifecycle, which I haven't verified) or in the tests steps failing because there is no artifact to download.~~

~~Now, why are the tests passing ATM? I can see three possiblities:~~

- ~~My understanding is wrong 😄~~
- ~~They pass because the tests download the artifact built on a previous build~~
- ~~They pass because we only have one agent (do we?) so even though the steps _would_ run in parallel, they effectively run sequentially because there's only one agent at a time that can run them.~~

Scrap all that 👆 😫

I went digging through the pipeline and agents timeline and saw that:

- The test jobs all waited for the build to finish
- The test jobs run in parallel

And that's because I missed the `depends_on: build` while reading the code 🤦 👍 My Buildkite-foo is rusty indeed, I didn't know about that configuration option.

.buildkite/pipeline.yml
jkmassel: Yeah, this is a good point. I started with really big obvious comment headers, but we could potentially pare them down for sure. I'd say it's safe to assume nothing in this file is permanent at _this_ point anyway! 😁

.buildkite/pipeline.yml
jkmassel: Same as above – possibly not, but I've had a couple different times where this happened.

Not sure if we want to specify something globally, or make each step specify its own runner agent (because not everything may want to run on macOS)

.buildkite/pipeline.yml
```diff
@@ -39,3 +39,30 @@ steps:
...
+    command: |
+      buildkite-agent artifact download build-products.tar .
+      tar -xf build-products.tar
+      bundle install
+      bundle exec pod install
+      rake mocks &
+      bundle exec fastlane test_without_building name:WordPressUITests try_count:3 device:"iPad Air (4th generation)" ios-version:"14.1"
+    depends_on: "build"
+    agents:
+      os: "macOS"
```
AliSoftware, line 68: image
triggered 😆

PR #16643, opened 2021-06-07
Add ratings configuration file

jkmassel:
This file will ensure that the App Store Connect age rating settings are updated for each release

**To test:**
- In App Store Connect, verify that the WordPress iOS age rating is set to 4+
- Run `bundle exec fastlane deliver --skip_screenshots --skip_binary_upload` (you'll need to apply [this `.patch` file in advance](https://github.com/wordpress-mobile/WordPress-iOS/files/6610690/0001-Fix-metadata.txt))
- Note that the change does _not_ show up in the confirmation screen
- Say yes, and allow the changes to be uploaded
- Verify that the age rating is now 17+

## Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/fastlane/Deliverfile
+++ b/fastlane/Deliverfile
@@ -14,3 +14,4 @@ precheck_include_in_app_purchases false
 
 api_key_path ".configure-files/app_store_connect_fastlane_api_key.json"
 
+app_rating_config_path "./fastlane/metadata/ratings_config.json"
--- a/fastlane/metadata/ratings_config.json
+++ b/fastlane/metadata/ratings_config.json
@@ -1,0 +1,17 @@
+{
+  "alcoholTobaccoOrDrugUseOrReferences": "NONE",
+  "contests": "NONE",
+  "gamblingSimulated": "NONE",
+  "horrorOrFearThemes": "NONE",
+  "matureOrSuggestiveThemes": "NONE",
+  "medicalOrTreatmentInformation": "NONE",
+  "profanityOrCrudeHumor": "NONE",
+  "sexualContentGraphicAndNudity": "NONE",
+  "sexualContentOrNudity": "NONE",
+  "violenceCartoonOrFantasy": "NONE",
+  "violenceRealisticProlongedGraphicOrSadistic": "NONE",
+  "violenceRealistic": "NONE",
+  "gambling": false,  
+  "seventeenPlus": false,
+  "unrestrictedWebAccess": true
+}


fastlane/Deliverfile
```diff
@@ -14,3 +14,4 @@ precheck_include_in_app_purchases false
 
 api_key_path ".configure-files/app_store_connect_fastlane_api_key.json"
 
+app_rating_config_path "./fastlane/metadata/ratings_config.json"
```
AliSoftware, line 17: Ah, so there _is_ a way to auto-update those with Deliver! 🎉  Nice!

fastlane/metadata/ratings_config.json
```diff
@@ -0,0 +1,17 @@
...
+  "matureOrSuggestiveThemes": "NONE",
+  "medicalOrTreatmentInformation": "NONE",
+  "profanityOrCrudeHumor": "NONE",
+  "sexualContentGraphicAndNudity": "NONE",
+  "sexualContentOrNudity": "NONE",
+  "violenceCartoonOrFantasy": "NONE",
+  "violenceRealisticProlongedGraphicOrSadistic": "NONE",
+  "violenceRealistic": "NONE",
+  "gambling": false,  
+  "seventeenPlus": false,
```
AliSoftware, line 15: Shouldn't this be set to `"seventeenPlus": true` – and shouldn't the `"matureOrSuggestiveThemes"` be set to "some" or whatever – to address paaHJt-2bZ-p2 ?

AliSoftware (commented):
At my EOD and off my mac so not able to test, but quickly read the diff from GH

fastlane/metadata/ratings_config.json
jkmassel: `matureOrSuggestiveThemes` in this case isn't required – the `unrestrictedWebAccess` is the one that best fits our profile, and causes the app to be bumped to `17+`

AliSoftware: Strange that there's a `seventeenPlus: false` in the JSON if the age rating is determined by the other factors (and I don't think we can select 17+ true/false in ASC's UI anyway, but only a level for the other (non-boolean) categories that we also see listed in that json). Is that `seventeenPlus` entry a legacy field then? (if so, couldn't it be completely omitted from the JSON to avoid confusion?)

JavonDavis (approved):
Changes lgtm! It looks like the rating was already changed to 17+ assuming that's from @AliSoftware test run

AliSoftware:
@JavonDavis I didn't test (only read code from GitHub diff from my iPad) 😅

mokagio:
It went to ASC all right:

Screen Shot 2021-06-10 at 9 28 29 am

PR #16648, opened 2021-06-08
Use ScreenObject Swift package

mokagio:
This adds the new [`ScreenObject`](https://github.com/Automattic/ScreenObject) Swift package to the project and uses it instead of `BaseScreen` in the `MeTabScreen`. I picked that screen because it was part touched in the login tests flow which we are currently working on (so there's fresher knowledge of them).

To test:

Run the `LoginTests` `XCTestCase` from the WordPressUITests scheme with the mocks (`rake mocks`) running on the terminal, it should build and pass.

## Regression Notes
N.A.

N.A.

None.

- [x] I have considered adding unit tests for my changes. N.A.
- [x] I have considered adding accessibility improvements for my changes. N.A.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. N.A.

---

This PR is still a draft, but I would appreciate your [30% feedback](https://www.linkedin.com/pulse/5-30-90-feedback-why-matters-manley-/) on it.

### DIFF
--- a/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,7 +6,7 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": "trunk",
-          "revision": "201ace5a0e5eb7ed2005a34763041b7022625f43",
+          "revision": "8de3def0a24c2cfe9d9f714b3e72a507526ef97a",
           "version": null
         }
       }
--- a/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -5,9 +5,9 @@
         "package": "ScreenObject",
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
-          "branch": "trunk",
-          "revision": "8de3def0a24c2cfe9d9f714b3e72a507526ef97a",
-          "version": null
+          "branch": null,
+          "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
+          "version": "0.1.0"
         }
       }
     ]
--- a/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "ScreenObject",
+        "repositoryURL": "https://github.com/Automattic/ScreenObject",
+        "state": {
+          "branch": "trunk",
+          "revision": "201ace5a0e5eb7ed2005a34763041b7022625f43",
+          "version": null
+        }
+      }
+    ]
+  },
+  "version": 1
+}
--- a/WordPress/WordPressUITests/Flows/LoginFlow.swift
+++ b/WordPress/WordPressUITests/Flows/LoginFlow.swift
@@ -68,7 +68,7 @@ class LoginFlow {
         XCTContext.runActivity(named: "Log out of app if currently logged in") { (activity) in
             if TabNavComponent.isLoaded() {
                 Logger.log(message: "Logging out...", event: .i)
-                let meScreen = TabNavComponent().gotoMeScreen()
+                let meScreen = try! TabNavComponent().gotoMeScreen()
                 if meScreen.isLoggedInToWpcom() {
                     _ = meScreen.logoutToPrologue()
                 } else {
--- a/WordPress/WordPressUITests/Flows/LoginFlow.swift
+++ b/WordPress/WordPressUITests/Flows/LoginFlow.swift
@@ -5,7 +5,7 @@ class LoginFlow {
 
     @discardableResult
     static func login(email: String, password: String) -> MySiteScreen {
-        logoutIfNeeded()
+        try! logoutIfNeeded()
 
         return PrologueScreen().selectContinue()
             .proceedWith(email: email)
@@ -17,7 +17,7 @@ class LoginFlow {
     // Login with self-hosted site via Site Address.
     @discardableResult
     static func login(siteUrl: String, username: String, password: String) -> MySiteScreen {
-        logoutIfNeeded()
+        try! logoutIfNeeded()
 
         return PrologueScreen().selectSiteAddress()
             .proceedWith(siteUrl: siteUrl)
@@ -38,7 +38,7 @@ class LoginFlow {
     // Login with WP site via Site Address.
     @discardableResult
     static func login(siteUrl: String, email: String, password: String) -> MySiteScreen {
-        logoutIfNeeded()
+        try! logoutIfNeeded()
 
         return PrologueScreen().selectSiteAddress()
             .proceedWithWP(siteUrl: siteUrl)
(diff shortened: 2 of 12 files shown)


WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,16 @@
...
+        "state": {
+          "branch": "trunk",
+          "revision": "201ace5a0e5eb7ed2005a34763041b7022625f43",
+          "version": null
+        }
+      }
+    ]
+  },
+  "version": 1
+}
```
mokagio, line 16: My Xcode generates this file, but I seems redundant given the packages info is already in the repo.

I added it just to show it to you and see if you know more. In Simplenote we decided not to add it.

It _might_ be due to the fact that I'm still on Catalina (:flush: – This week I'll update. Promise)

WordPress/WordPressUITests/Flows/LoginFlow.swift
```diff
@@ -4,7 +4,7 @@ class LoginFlow {
 
     @discardableResult
     static func login(email: String, password: String) -> MySiteScreen {
-        logoutIfNeeded()
+        try! logoutIfNeeded()
```
mokagio: I force tried (`try!`) here because I didn't want to keep traversing the call graph. In the "final" implementation, I'd like not to have any `try!`. The reason I'm saying this is because XCTest knows what to do when a `try` `throws` and report it as a test failure instead of a crash.

_Shameless plug: I wrote of such an instance [today](https://mokacoding.com/blog/how-to-test-async-await-code-in-swift/)._

WordPress/WordPressUITests/Screens/MeTabScreen.swift
```diff
@@ -1,7 +1,7 @@
-import Foundation
```
mokagio: This import was unnecessary because:

- I'm pretty sure XCTest imports Foundation
- There was no direct call to a Foundation specific type, like `URL`

WordPress/WordPressUITests/Screens/MeTabScreen.swift
```diff
@@ -10,8 +10,11 @@ class MeTabScreen: BaseScreen {
     let notificationSettingsButton: XCUIElement
     let doneButton: XCUIElement
 
-    init() {
-        let app = XCUIApplication()
+    // TODO: Make this public in ScreenObject
+    let app: XCUIApplication
```
mokagio: I'll try not to forget about this 😄

WordPress/WordPressUITests/Screens/MeTabScreen.swift
```diff
@@ -20,7 +23,7 @@ class MeTabScreen: BaseScreen {
         notificationSettingsButton = app.cells["notificationSettings"]
         doneButton = app.navigationBars.buttons["doneBarButton"]
 
-        super.init(element: appSettingsButton)
+        try super.init(element: appSettingsButton)
```
mokagio: This is it, the changes between `BaseScreen` and `ScreenObject` ([see implementation](https://github.com/Automattic/ScreenObject/blob/trunk/Sources/ScreenObject/ScreenObject.swift)) are minimal at this point.

The main thing I did was removing the forced `try` from it, as it didn't seem appropriate for a standalone framework to brutally crash its consumers 😅

mokagio (commented):
@jkmassel @pachlava @thehenrybyrd I'd love your [30% feedback](https://www.linkedin.com/pulse/5-30-90-feedback-why-matters-manley-/) on this, even though I've left the PR as a draft.

Thanks 🙏 🙇

WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
mokagio: While working on the WooCommerce version of this experiment, I found what I think is [the right answer to the question](https://github.com/woocommerce/woocommerce-ios/pull/4392#discussion_r648001032) of whether tracking this file is appropriate.

Yes it is

> This looks to me like the `Podfile.lock` or `package-lock.json` equivalent for the Xcode managed dependency resolution. The Xcode project is the source of truth for the configuration and this is the snapshot of the dependencies as resolved from that configuration that the project expects.

pachlava (approved):
@mokagio sorry for taking so long to reply, I had troubles setting this up yesterday, and also needed some time to get a bit acquainted. I still have some points that are not clear to me, and I'll ask them in separate comments. The `LoginTests` were executed with no fails on iPhone12.

It also looks like there's the same kind of warning related to `Info.plist` as it was for Simplenote:

Screenshot_2021-06-09 17 20 50_da5UxZ

WordPress/WordPressUITests/Tests/EditorAztecTests.swift
```diff
@@ -14,14 +14,14 @@ class EditorAztecTests: XCTestCase {
...
+    override func tearDownWithError() throws {
         takeScreenshotOfFailedTest()
         if editorScreen != nil && !TabNavComponent.isVisible() {
             EditorFlow.returnToMainEditorScreen()
             editorScreen.closeEditor()
         }
-        LoginFlow.logoutIfNeeded()
-        super.tearDown()
+        try LoginFlow.logoutIfNeeded()
+        try super.tearDownWithError()
```
pachlava, line 24: It's not related to your change, I was just confused by a call of superclass method. After some more advanced googling, I think I found the [explanation](https://stackoverflow.com/a/53455661) for using `super.xxx` in `setUp..` and `tearDown..`, saying that this becomes meaningful if the test hierarchy has more than one level.

It looks like the calls to superclass were generated by Xcode automatically upon files creation, until a certain version, but then it stopped? And these calls remain here from the times when they were generated automatically?

mokagio: Good catch @pachlava!

You inspired me to start [a micro-project to track this kind of changes](https://github.com/mokagio/SwiftUI-Xcode-Template).

Indeed, neither Xcode 12.5.1 nor 13 put the call to `super` in the template implementation for the set-up and tear-down methods. [The docs](https://developer.apple.com/documentation/xctest/xctest/3521150-setupwitherror) don't mention the need to call `super`, either. One would assume that if it was critical for the correct functioning of the method, they'd at least mention it in the docs.

I'll followup with a dedicated PR to remove all this kind of `super` calls. Thanks 🙌

mokagio:
I'm going to cash in on @pachlava's approval, even though it's old, and merge this.

The Info.plist mentioned was addressed [in this commit](https://github.com/Automattic/ScreenObject/commit/3d93c36483884db3fab88aea5c181ce75ceeac1d#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677e).

Also notice that I run the UI tests in CI and they're green 😄

dcalhoun:
For posterity, in case anyone runs into similar issues, I ran into authentication failures fetching the `ScreenObject` package from GitHub after pulling down this change from origin. I needed to perform the following to fix the Xcode build:

1. Remove local `git` config I had in place that defaulted to SSH<sup>1</sup>:
    ```
    [url "git@github.com:"]
      insteadOf = https://github.com/
    ```
2. Generate a GitHub [personal access token](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) and set it for my GitHub account with Xcode preferences. 
3. Quit and restart Xcode to clear a "Missing package product WordPressFlux" error. 

---

[1]: It would appear that Xcode does not support SSH keys encrypted via the `ed25519` algorithm, which is the [recommendation from GitHub](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent). An alternative to the personal accessibility token would likely be generating and using a SSH key via the `rsa` algorithm.

PR #16649, opened 2021-06-08
Adds tracking to the blogging reminders flow.

diegoreymendez:
Adds tracking of almost all events to the blogging reminders flow.

The events that are left out are:

- `blogging_reminders_scheduled`
- `blogging_reminders_cancelled`

These will be taken care of once we wire the notifications scheduling logic.

## Test setup:

Wire `presentBloggingRemindersSettingsFlow(for:source:)` from a location you use to trigger the call.  I wired it in file `BlogDetailsViewController+Header.swift`, line 42 so that the Stats quick action button code looks like this:

```swift
guard let self = self else {
    return
}
                
self.presentBloggingRemindersSettingsFlow(for: self.blog, source: .blogSettings)
```

Use the filter "Tracked: blogging" in your console when testing, as it makes it easier to spot the tracked events.

### Successful flow:

The successful flow, is the user going all the way from the beginning to the end of the flow.  It should look like this:

```
🔵 Tracked: blogging_reminders_flow_start <blog_id: $id, blog_type: wpcom, blog_url: $url, source: blog_settings>
🔵 Tracked: blogging_reminders_screen_shown <blog_id: $id, blog_type: wpcom, blog_url: $url, screen: main>
🔵 Tracked: blogging_reminders_button_pressed <blog_id: $id, blog_type: wpcom, blog_url: $url, button: continue, screen: main>
🔵 Tracked: blogging_reminders_screen_shown <blog_id: $id, blog_type: wpcom, blog_url: $url, screen: day_picker>
🔵 Tracked: blogging_reminders_button_pressed <blog_id: $id, blog_type: wpcom, blog_url: $url, button: continue, screen: day_picker>
🔵 Tracked: blogging_reminders_screen_shown <blog_id: $id, blog_type: wpcom, blog_url: $url, screen: all_set>
🔵 Tracked: blogging_reminders_button_pressed <blog_id: $id, blog_type: wpcom, blog_url: $url, button: continue, screen: all_set>
🔵 Tracked: blogging_reminders_flow_completed <blog_id: $id, blog_type: wpcom, blog_url: $url>
```

### Dismiss by tapping X button:

If on the first or second screens you tap on the X button, these are the last two events you should see:

```
🔵 Tracked: blogging_reminders_button_pressed <blog_id: $id, blog_type: wpcom, blog_url: $url, button: dismiss, screen: main>
🔵 Tracked: blogging_reminders_flow_dismissed <blog_id: $id, blog_type: wpcom, blog_url: $url, source: main>
```

### Dismiss by dragging down:

In this case you should not see the button_pressed event from the previous case, only the flow_dismissed event.

```
🔵 Tracked: blogging_reminders_flow_dismissed <blog_id: $id, blog_type: wpcom, blog_url: $url, source: main>
```

### Dismissing the "All Set" screen

No matter how you dismiss this screen (pressing the button, tapping X, or dragging the sheet down), the last event you should see is:

```
🔵 Tracked: blogging_reminders_flow_completed <blog_id: $id, blog_type: wpcom, blog_url: $url>
```

## Regression Notes

None since I'm working on a part of the App that is still not available to the general public.

PR submission checklist:


diegoreymendez:
@renanferrari - Can I ask you to check that the test cases make sense?  I'm mostly interested in the ordering of the events, and making you aware of the tracking on iOS in case you notice anything that needs synchronization across platforms.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
```diff
@@ -40,8 +40,11 @@ extension BlogDetailsViewController {
 
         let actionItems: [ActionRow.Item] = [
             .init(image: .gridicon(.statsAlt), title: NSLocalizedString("Stats", comment: "Noun. Abbv. of Statistics. Links to a blog's Stats screen.")) { [weak self] in
-                self?.tableView.deselectSelectedRowWithAnimation(false)
-                self?.showStats(from: .button)
+                guard let self = self else {
```
frosty: I don't think you meant to commit this! :)

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
```diff
@@ -1,10 +1,159 @@
 import UIKit
 import SwiftUI
 
+class BloggingRemindersTracker {
```
frosty: I think there's enough here that we could probably move it to its own file.

frosty (commented):
Looking good! Two notes (and I posted two code comments):

* If I dismiss a screen using the X, I see two dismiss events posted. I think we can remove the call to `tracker.flowDismissed(source: .main)` in `dismissTapped`, because `handleDismiss` seems to cover this case.
* I'm not sure that we need to track the blog ID and URL. Do you think we would use that for anything? 🤔

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
diegoreymendez: I actually did!  It's to be able to reference `self.blog` for the call below.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
diegoreymendez: Yeah, this is actually a mistake.  I created the class here for convenience, but forgot to move it out.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Header.swift
diegoreymendez: Actually, ignore me - you're right and I didn't.

diegoreymendez: Rolled back.

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController+Reminders.swift
diegoreymendez: Moved out to a new file of its own.

diegoreymendez:
Ready for another look.

Gio2018 (changes requested):
Hey @diegoreymendez , this looks good! I noticed an issue: when completing the flow and tapping either the 'X' or the "Done" button, the `blogging_reminders_flow_completed` event is fired twice. A similar thing happens when I dismiss the screen with the 'X'  before completing the flow: the event `blogging_reminders_flow_dismissed` is fired twice. I believe this happens because the event is triggered both by the specific action (e.g. button pressed) and the delegate method `handleDismiss` of the bottom sheet. In fact, the event is only fired once if you dismiss by dragging down.

diegoreymendez:
@frosty, @Gio2018 - Ready for another look.

Changes:

- Changed a parameter in the presentation method to make it slightly clearer.
- Removed blog ID and blog URL information from the tracking.
- Removed the duplicated screen-dismiss and flow-completion events.

Gio2018 (changes requested):
Hey @diegoreymendez , juste tested it on iPad Pro 2nd gen (device) iOS 14.5 and I don't see the `flow_dismissed` event anymore, when tapping the X

diegoreymendez:
Hey @Gio2018 - it seems like `BottomSheetViewController` behaves differently in iPhone and iPad, which is causing the behaviour you're seeing.

I see two possible solutions to this problem:

1. Try to change `BottomSheetViewController` so that it behaves the same way for iPhone and iPad.
2. Try to fix this with a quick patch.

Option 1 is the one I prefer, because I think ignoring the underlying issue may be counter productive for the App.  It is riskier though because there's a chance it may bring regressions - but I can analyze the existing code to try to avoid them.

Option 2 would quickly get us out of trouble, but it wouldn't be very future proof (if someone fixed the underlying issue, they'd probably break the patch), and I'd rather avoid.

I'll try to see if it's possible for me to push a PR that addresses option 1, and will submit it separately before continuing with this PR.

mokagio:
I'm going to bump this to 17.7 as part of the open PRs walkthrough before starting the 17.6 code freeze because there are changes requested. See also [my comment in this related PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/16672#issuecomment-860333812).

Ping me here if this can't wait two weeks and needs to go into 17.6 once approved 👍

diegoreymendez:
Hey folks, apologies for the delay but it took me a while to find a suitable solution to avoid duplicated or missing events in all scenarios.

I've re-tested all paths with an iPhone and an iPad and the tracking should be correct now.

Can you give it another look?

Gio2018 (approved):

PR #16651, opened 2021-06-08
Fix date medium string warnings

frosty:
This PR fixes all of the medium string warnings we currently have in the project.

**To test**

- Check over the code changes. I removed one method entirely because it didn't appear to be used any more.
- Build and run, and if you like verify dates still appear correctly in each of the following situations:
- In the Posts list, on post cards
- In the Media section, select a media item and check the uploaded date
- Post revision rows and detail view
- Reader detail date labels and the accessibility label read when selecting the header in Voiceover
- Reader post cards
- I tried to verify the changes to notifications sent by the Share Extension, but when my posts updated they always just said 'now'. The code looks fine to me though.

## Regression Notes

Only areas where these calls are used – however, I think they're all used for non-critical visual elements, so I think it should be fine.

I verified each of the areas changed.

None

PR submission checklist:


jkmassel (changes requested):
There's one failing test, but then this should be good to go

mokagio:
> This PR fixes all of the medium string warnings we currently have in the project.

Fixes + All of + A type of warning. ❤️ Thanks @frosty 🤘 

I'm going to bump this to 17.7 as part of the open PRs walkthrough before starting the 17.6 code freeze because it's doesn't seem time sensitive and there are changes requested.

Ping me here if this can't wait two weeks and needs to go into 17.6 once approved. 👍

frosty:
@jkmassel Test fixed – I opted to make it a little more general, as the existing test was dependent on the exact formatting coming out of the date formatter, which would vary depending on language, etc.

WordPress/WordPressTest/PostCardCellTests.swift
```diff
@@ -82,7 +82,7 @@ class PostCardCellTests: XCTestCase {
 
         postCell.configure(with: post)
 
-        XCTAssertEqual(postCell.dateLabel.text, "just now")
+        XCTAssert(postCell.dateLabel.text?.isEmpty == false)
```
mokagio: Mixed feelings about this. I like how you made it language independent. At the same time, checking that the string is not empty may pass in any number of scenarios, including someone accidentally hardcoding a value in the text. 🤔 

Another option could be to check that the result is the same running the input `Date` through the formatter method. That, on the other hand, would bind this code to the implementation detail that we're using the formatter 😫

This is a tricky problem... 🤔 

I think maybe updating to "now" and bank on the fact they'll always run in English might make it more robust. "_bank on the fact they'll always run in English_" is actually something we could enforce in the Test Plan that runs these tests 🤔 

Also, nitpick:

```suggestion
        XCTAssertFalse(postCell.dateLabel.text?.isEmpty)
```

Same thing, but with one less operation. The result if the test fails is, unfortunately, equally uninformative, but that's an XCTest limitation not an issue with the test code 😞 

Screen Shot 2021-06-16 at 9 52 01 am

mokagio (approved):

jkmassel (approved):

PR #16609, opened 2021-06-01
Fix: Notice over quickstart removal next steps

enejb:
This is a quick follow up on https://github.com/wordpress-mobile/WordPress-iOS/pull/16579
It fixed the found issue with the QuickStart Remove step flow.

See
https://user-images.githubusercontent.com/198826/120044921-10bd3100-bfcc-11eb-9364-31dec8e61c1b.mp4

To test:
1. Create a new site or if you have a site that has quickstart already. 
2. Update the site title. (To create a notice) See that the site title update notice shows up.  
3. Update the site title and clicking click the three dots. 
4. Cancel the flow. 
5. See that the site title update notice shows up. 
6. Update the site title and clicking click the three dots. 
7. Click the Remove Next Steps. 
8. Cancel the alert. 
9. See that the site title update notice shows up. 
10. Follow steps 5 - 7. 
11. Click remove in the alert.
12. See that the site title update notice shows up.  

See video.

https://user-images.githubusercontent.com/115071/120399431-de277700-c2f0-11eb-98f7-4e0f460f2396.mp4

Note that in the video. I had commented out the part of the code that removes the quickstart so that is why on the last step you can still it.

## Regression Notes
N/A 

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16658, opened 2021-06-09
Fix: Remove the notice that over laps with the Editor More ActionSheet

enejb:
Fixes the notice overlap that currently shows up when you save a post and then click on the three dots. 

To test:
1. Create a new post or page in the editor. 
2. Save the post as draft. 
3. Once the post is done saving notice that when you click on the 3 dots again that the notice that shows up doesn't overlap the ActionSheet that. 

See for how the action sheet works on this PR. 

<table>
<tr><td>Before</td><td>After</td></tr>
<tr><td></td><td>

https://user-images.githubusercontent.com/115071/121298625-0b53c680-c8a9-11eb-89d8-2871eb445a6a.mp4

</td></tr>
</table>

The fix is very similar to https://github.com/wordpress-mobile/WordPress-iOS/pull/16609

## Regression Notes
Not sure. 

This was tested manually. 

I don't know how to add automated tests for this. 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
@@ -798,7 +798,6 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
             case .close:
                 cancelEditing()
             case .more:
-                ActionDispatcher.dispatch(NoticeAction.lock)
                 displayMoreSheet()
             case .switchBlog:
                 blogPickerWasPressed()
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
@@ -798,6 +798,7 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
             case .close:
                 cancelEditing()
             case .more:
+                ActionDispatcher.dispatch(NoticeAction.lock)
                 displayMoreSheet()
             case .switchBlog:
                 blogPickerWasPressed()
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -798,6 +798,7 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
             case .close:
                 cancelEditing()
             case .more:
+                ActionDispatcher.dispatch(NoticeAction.lock)
```
chipsnyder: NitPick: Would it be possible to move this call to the implementation of [displayMoreSheet()](https://github.com/wordpress-mobile/WordPress-iOS/blob/1aaa478dbdcb3e26457217f59f5f87c8e48f89ce/WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController%2BMoreActions.swift#L13)?

I think having it in the same file/location will make it easier to see the relationship between locking and unlocking.

chipsnyder (commented):
Thank you for adding the related issue @enejb that really helped gain some context. 

Since we seem to be repeating this in many places, I wonder if it would be worthwhile to subclass the UIAlertController and override `viewDidAppear` and `viewWillDisappear` to lock and unlock. Then the only risk would be to an engineer not knowing to use the subclassed Alert but that could be solvable with lint tools. WDYT?

cc @frosty and @shiki since the two of you have put some thought into this.

enejb:
Thanks for the review @chipsnyder. 
I think this new approach of overriding the UIAlertController would really interesting to implement but I am just not sure how. 
Would you place the code in WordPress-mobile/WordPressUI-iOS ?

shiki:
> Since we seem to be repeating this in many places, I wonder if it would be worthwhile to subclass the UIAlertController and override viewDidAppear and viewWillDisappear to lock and unlock. 

It's not a bad idea. 🙂 Fewer mistakes I guess! 

Though the documentation mentions this: 

> The UIAlertController class is intended to be used as-is and does not support subclassing. The view hierarchy for this class is private and must not be modified.

Not quite sure what that means but the functions do not seem to be overridable. 🤔 

image

If not that, maybe a simple class that has the same set of functions but uses `UIAlertController` internally might work. A decorator design pattern I think?

chipsnyder:
Ah good call out on that @shiki I scrolled right past that note about UIAlertController not being sub-classable. A decorator pattern could work well. I'll spend some time playing with that and take a look but won't hold up this change for that.

chipsnyder (approved):
This looks good to me 🚢 

I think we can tackle the other suggestions I had as a separate task. But probably something to consider if we need more.

PR #16662, opened 2021-06-10
Establish a shared UI automation fondation framework

mokagio:
Moves all of the shared UI automation _utilities_ into a newly created UITestsFoundation framework.

I haven't changed any of the logic that I moved, with the exception of the `WireMock` class for which I'll refer you to the inline comments.

The next step will be to move all of the shared `BaseScreen` subclasses in that framework. This will simplify the UI tests setup because we won't need to keep sharing file across different targets, simply having a screen in the framework will make it available to the UI tests and the WordPress and Jetpack screenshot generation tests.

Screen Shot 2021-06-15 at 8 40 40 pm

You'll see a bunch of `TODO` comments. I don't normally open PRs with `TODO` comments. This one is the first step of a bigger piece of work though, so I left myself those breadcrumbs in the code.

To test: feel free to checkout the branch and to run the UI tests locally. Also notice that I run the full suite in CI and it worked 😄 

image

## Regression Notes
N.A.

N.A.

N.A.

---


mokagio:
FYI @pachlava @thehenrybyrd. Just in case you though of doing something along these lines for the remainder of HACK week 😄

WordPress/JetpackScreenshotGeneration/Info.plist
```diff
@@ -20,9 +20,5 @@
 	<string>????</string>
 	<key>CFBundleVersion</key>
 	<string>1</string>
-	<key>WIREMOCK_HOST</key>
-	<string>$(WIREMOCK_HOST)</string>
-	<key>WIREMOCK_PORT</key>
-	<string>$(WIREMOCK_PORT)</string>
```
mokagio, line 26: I removed the two `WIREMOCK_` settings from the bundle in favor of 1) having them hardcoded and 2) override-able via environment variable. More details in `WireMock.swift` below.

WordPress/UITestsFoundation/BaseScreen.swift
```diff
@@ -0,0 +1,113 @@
+import XCTest
+
+// Access control `open` so types outside of this framework can subclass it, which is exactly how
+// this type is meant to be used.
+open class BaseScreen {
```
mokagio, line 5: This is extracted from the WordPressUITests target. It was actually added to multiple targets. 

One extra benefit of moving logic shared between targets in a dedicated framework is that it gets a clear home and you don't need to _remember_ to add new files to all targets, because the build system manages framework linking for you 😄

WordPress/UITestsFoundation/FancyAlertComponent.swift
```diff
@@ -1,11 +1,10 @@
-import Foundation
 import XCTest
 
-class FancyAlertComponent: BaseScreen {
+public class FancyAlertComponent: BaseScreen {
```
mokagio, line 3: As you can see from the GitHub UI, this is another file move. The addition of `public` access control is because now that this code is in a different target from its original, all the functions that the UI tests call from here need to be explicitly set as `public` in order to be accessed.

WordPress/UITestsFoundation/WireMock.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+public class WireMock {
+
+    public static func URL() -> Foundation.URL {
+        let host = ProcessInfo().environment["WIREMOCK_HOST"] ?? "localhost"
+        let port = ProcessInfo().environment["WIREMOCK_PORT"] ?? "8282"
```
mokagio, line 7: The previous implementation of this class would read the host and port values from the `Bundle`, which picked them up from the `Info.plist`, which in turn read them from user defined build settings.

Once I moved the class to a different target than the one were those user defined build settings were set, the setup broke because the `Info.plist` of the new UITestsFoundation target doesn't have those values.

But, since the were hardcoded in the project file anyways and there was no logic in our automation that passed custom `xcargs` values to override them, did they really need to be read from there anymore?

This solution allows the class to live in this target without any change to its external API. By using the launch environment variables, we can also retain the ability to override the settings.

Unless I'm missing some context on the previous implementation, I think in this new iteration of the architecture this approach is a good fit.

WordPress/WordPressUITests/Flows/EditorFlow.swift
```diff
@@ -1,3 +1,4 @@
+import UITestsFoundation
```
mokagio, line 1: You'll see a few of these import statements in files from the other UI tests targets. The reason they are here is because these tests used one or more of the utils that have been extracted into the framework.

WordPress/WordPress.xcodeproj/xcshareddata/xcschemes/UITestsFoundation.xcscheme
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
```
mokagio, line 2: This is the scheme to use to build the UITestFoundation framework in isolation.

mokagio:
I have enabled auto-merge to keep the work moving. I'll address any non-blocking comment as part of the next PRs in this stream of work (whenever I'll get to it...)

mokagio:
@rachelmcr @momo-ozawa according to GitHub you have worked in this neck of the woods recently, so I've asked your review also as a way to cc you on these changes. @pachlava let me know if you have any questions about the setup. I'll address any suggestions about the coding as part of the followup pass and/or when extracting the WordPress agnostic utils to [XCUITestHelpers](https://github.com/Automattic/XCUITestHelpers/)

rachelmcr:
Thanks for the heads up! I'm not working in this code currently but I'm excited to see a shared framework coming together. :)

momo-ozawa (approved):
Looking awesome! LGTM with 1 question.

❓ Are you planning on moving [SnapshotHelper](https://github.com/wordpress-mobile/WordPress-iOS/blob/b47a299b2764d95d03363ff8ca6035dd4866950c/WordPress/WordPressScreenshotGeneration/SnapshotHelper.swift) and the [thenTakeScreenshot](https://github.com/wordpress-mobile/WordPress-iOS/blob/a3eae1d75066b28c215180cafd12e4947c74e556/WordPress/WordPressScreenshotGeneration/WordPressScreenshotGeneration.swift#L113-L123) BaseScreen extension to the new UITestsFoundation framework?

mokagio:
> ❓ Are you planning on moving SnapshotHelper and the thenTakeScreenshot BaseScreen extension to the new UITestsFoundation framework?

Great question @momo-ozawa, thanks! I haven't given much thought about it, yet. I think your suggestion is the way to go 🙇‍♂️

PR #16670, opened 2021-06-11
Add emoji site icon creator

frosty:
This PR adds an option to create a site icon using an emoji with a background color.

emoji-picker-3

|   |   |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-06-11 at 12 03 52 | Simulator Screen Shot - iPhone SE (2nd generation) - 2021-06-11 at 11 50 51 |

I've updated the 'change site icon' alert sheet to now show an option for "Create from emoji". I've also limited the feature to iOS 14, as there are a few SwiftUI 2 features that make this work or perform better (in particular, `ScrollViewReader` and `LazyHStack`).

**To test**

* Build and run
* Tap the site icon for a site, and check you see the 'Create from Emoji' option in the sheet
* Try building an icon! Choose an emoji (check the emoji group picker bar also works to jump to a particular section of emoji) and a color, then tap Save.
* Try dismissing the screen without tapping Save and ensure nothing gets updated.
* Check the previous 'choose from my device' option for setting a site icon still works.
* Disable the feature flag and check it still works as before (the alert menu is slightly different before the feature flag).
* Test on an iOS 13 simulator and ensure the new option isn't shown.

## Regression Notes

* Breaking the site icon update flow, or the site icon Quick Start tour

I manually tested both of these areas and they still worked as expected

None

PR submission checklist:


diegoreymendez (approved):
This may be a bit subjective, but when turning the device, the visible scroll bar looks a bit funny when it's near the X button.

Screen Shot 2021-06-11 at 14 56 42

Additionally on iPad, in portrait mode, I think the margings and the sizing could eventually use a review.

That said, the feature looks solid to me, and I think it's in a very good point for a developer flagged feature.

Really nice work - I'm excited to start using these for my sites!

osullivanchris:
Looks really cool. Nice job. Already discussed a few ideas with you on DM. 

One thing that strikes me is when you see it on a small screen or on landscape, it starts to feel like the emoji and the colour could be two steps, rather than having to scroll. But when space is available, doing both together is pretty nice.

Two things that I think are great here which we could expand on:
- explaining what a site icon is. Would be cool if we did that for normal images too. Maybe we need a custom sheet initially rather than the action sheet. We could explain the thing there. 
- might be nice to have other options, e.g. pick an image from free photo library

frosty:
Thanks both! I’m going to merge this now to hopefully get some more feedback over the next two weeks before I go AFK, and I can address some of the remaining issues over that time.

PR #16674, opened 2021-06-11
[Tooling] Fix GitHub releases for intermediate betas

AliSoftware:
### Why?

_Similar PR than WCiOS's https://github.com/woocommerce/woocommerce-ios/pull/4404_

Fix the GitHub releases drafts being incorrect for intermediate betas:
 - They used the final version number (`x.y`) for the release name and tag instead of the build number (`x.y.0.N`)
 - They had the incorrect release notes (used the previous ones for any beta — including the first one — created before we manually update the notes with Editorial copy)
 - They didn't have the "Pre-release" checkbox checked on them

### To Test

Not easy to test from start to finish without creating a new beta… we're close to do the `finalize_release` for this Sprint later today, so not worth create a new beta just couple of minutes ago I guess, so not sure if there's a way to dry-run/test this 😒 

So maybe the best way to test this besides just reviewing the code is to test this during the next cycle…

## Regression Notes
-
-
-

### PR submission checklist:

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -346,7 +346,7 @@ platform :ios do
 
     build_and_upload_internal(skip_prechecks: true, skip_confirm: options[:skip_confirm]) if options[:beta_release]
     build_and_upload_itc(skip_prechecks: true, skip_confirm: options[:skip_confirm],
-                         final: !options[:beta_release], create_release: options[:create_gh_release])
+                         beta_release: options[:beta_release], create_release: options[:create_gh_release])
   end
 
   #####################################################################################
@@ -477,13 +477,13 @@ platform :ios do
   # This lane builds the app and upload it for external distribution
   # -----------------------------------------------------------------------------------
   # Usage:
-  # bundle exec fastlane build_and_upload_itc [skip_confirm:<skip confirm>] [create_release:<Create release on GH> ]
+  # bundle exec fastlane build_and_upload_itc [skip_confirm:<skip confirm>] [create_release:<Create release on GH>] [beta_release:<intermediate beta?>]
   #
   # Example:
   # bundle exec fastlane build_and_upload_itc
   # bundle exec fastlane build_and_upload_itc skip_confirm:true
   # bundle exec fastlane build_and_upload_itc create_release:true
-  # bundle exec fastlane build_and_upload_itc create_release:true final:true
+  # bundle exec fastlane build_and_upload_itc create_release:true beta_release:true
   #####################################################################################
   desc 'Builds and uploads for distribution'
   lane :build_and_upload_itc do |options|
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -106,6 +106,9 @@ platform :ios do
 
     ios_bump_version_release
     new_version = ios_get_app_version
+    extract_release_notes_for_version(version: new_version,
+      release_notes_file_path: FIle.join(ENV["PROJECT_ROOT_FOLDER"], 'RELEASE-NOTES.txt'),
```
mokagio:
```suggestion
      release_notes_file_path: File.join(ENV["PROJECT_ROOT_FOLDER"], 'RELEASE-NOTES.txt'),
```

This wouldn't have happened in _insert compiled language_ 😢 I wonder if there's some kind of static analysis that we could put in place to catch this kind of stuff for us.

fastlane/Fastfile
```diff
@@ -106,6 +106,9 @@ platform :ios do
...
+      extracted_notes_file_path: './WooCommerce/Resources/release_notes.txt')
```
mokagio:
```suggestion
      extracted_notes_file_path: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources', 'release_notes.txt'))
```

fastlane/Fastfile
```diff
@@ -244,7 +247,6 @@ platform :ios do
 
     # Wrap up
     version = ios_get_app_version
-    # ios_clear_intermediate_tags(version: version)
```
mokagio, line 247: Apart from the fact this is a line of commented code, so whether it's there or deleted the result is the same, you chose to remove this because we've agreed to not remove the intermediate tags, right? So we're in line with Android etc.

mokagio (approved):
I tried the `extract_release_notes_for_version` step locally (after applying my suggestions) on top of 17.5 and I saw that `release_notes.txt` updated accordingly to the info in `RELEASE-NOTES.txt`.

> So maybe the best way to test this besides just reviewing the code is to test this during the next cycle…

Since I'll be doing the next cycle, it's only apt that I'll approve and test it that way 😄

fastlane/Fastfile
```diff
@@ -106,6 +106,9 @@ platform :ios do
 
     ios_bump_version_release
     new_version = ios_get_app_version
+    extract_release_notes_for_version(version: new_version,
+      release_notes_file_path: File.join(ENV["PROJECT_ROOT_FOLDER"], 'RELEASE-NOTES.txt'),
+      extracted_notes_file_path: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources', 'release_notes.txt'))
```
mokagio, line 111: Notice that one uses `ENV` and the other the `PROJECT_ROOT_FOLDER`. I should have properly used the same way to read the value for consistency but I realized it only after making the commit.

I'll followup with a PR that removes all accesses of `ENV["PROJECT_ROOT_FOLDER"]` in favor of using the constant, which is our desired approach.

mokagio:
Hey @AliSoftware, sorry for brutally taking over this PR. I did it because I'm working on the 17.6 code freeze today and I want this fix to be into `develop` before starting.

👍 

Thank you for going above and beyond and porting this fix over from Woo! 👏

fastlane/Fastfile
AliSoftware: Yep exactly.

I think this line got commented a while ago when there was the bug that deleted all the tags accidentally, and never got un-commented once the fix shipped in toolkit apparently.

But since now we decided to create the GH releases for intermediate betas like for Android, we agreed to keep the tags too so that the GitHub releases don't go back from published state back to draft state if we deleted the tags. Hence no more tag deletion and removing this line once and for all

fastlane/Fastfile
AliSoftware: Took me a while to see what was the difference in your suggestion diff 😅 this `I` vs `i` was so sneaky 😰 good eye!

Wonder if rubocop would have caught this? (We don't yet lint our Fastfile, but we definitively should)

AliSoftware:
> Hey @AliSoftware, sorry for brutally taking over this PR. I did it because I'm working on the 17.6 code freeze today and I want this fix to be into `develop` before starting.

That's perfect, that was what was intended (having the RM merge this before starting today's code freeze to test it even if they had to fix the quirks themselves while they were at it) and your changes make sense, so 👌

PR #16679, opened 2021-06-13
Issue/14055 - Mark referrer as spam

### LINKED ISSUE #14055: Stats: Referrers: Not Possible to Mark a Referrer as Spam
In Calypso, referrers which are not whitelisted can be marked spam by tapping the three dots to the right of the referrer. This is not the case in the WordPress iOS app (or Android).

Expected:
I expected to be able to mark referrers as spam from within the app.

Actual: 
The option to mark referrers as spam doesn’t exist within the app.

Stats referrers in WPiOS and Calypso web (same site) on iPhone SE:
iosreferrers

##### Tested on iPhone SE, iOS 13.1.2, WPiOS 14.6
This came up in #2934822-zen when a user wanted to mark some referrers as spam but couldn't.

nikola-milicevic:
Fixes [#14055](https://github.com/wordpress-mobile/WordPress-iOS/issues/14055)

Depends on [WordPressKit PR](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/407).

### Description:

This PR adds marking / unmarking as spam functionality to the referrer.

### To test:
1. On 'My Site' tab go to 'Stats' screen.
2. Tap on Days / Weeks / Month / Years
3. Tap on a referrer to mark it as spam.
4. Refreshing the data will remove the referrer marked as spam from the list.

Note: if there are many referrers, marking referrer as spam can be done from the referrers list screen as well (accessed by tapping on 'View more' button).

### Video:

https://user-images.githubusercontent.com/12729242/122479467-6c774a80-cfcb-11eb-8dc8-2bd42508984d.mov

### Regression Notes
n/a

n/a

- StatsPeriodStore mark / unmark referrer as spam

PR submission checklist:


WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
```diff
@@ -197,6 +209,24 @@ private extension SiteStatsPeriodTableViewController {
...
+
+        let title = isSpam ? unmarkTitle : markTitle
+        let toggleSpamAction = UIAlertAction(title: title, style: .default) { [weak self] _ in
+            self?.viewModel?.toggleSpamState(for: referrerDomain)
+        }
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+        [toggleSpamAction, cancelAction].forEach {
+            sheet.addAction($0)
+        }
+        present(sheet, animated: true)
```
emilylaguna: It looks like this logic is duplicated elsewhere, is there a way we can remove the duplicates?

WordPress/Classes/ViewRelated/Stats/Shared Views/Stats Detail/SiteStatsDetailTableViewController.swift
```diff
@@ -363,3 +367,24 @@ extension SiteStatsDetailTableViewController: SiteStatsTableHeaderDelegate {
...
+
+        let title = isSpam ? unmarkTitle : markTitle
+        let toggleSpamAction = UIAlertAction(title: title, style: .default) { [weak self] _ in
+            self?.viewModel?.toggleSpamState(for: referrerDomain)
+        }
+        let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+        [toggleSpamAction, cancelAction].forEach {
+            sheet.addAction($0)
+        }
+        present(sheet, animated: true)
```
emilylaguna: It looks like this logic is duplicated elsewhere, is there a way we can remove the duplicates?

WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
```diff
@@ -236,6 +253,23 @@ private extension StatsTotalRow {
...
+                moreButton = UIButton()
+                moreButton?.addTarget(self, action: #selector(didTapMoreButton), for: .touchUpInside)
+                moreButton?.translatesAutoresizingMaskIntoConstraints = false
+                addSubview(moreButton!)
+
+                NSLayoutConstraint.activate([
+                    moreButton!.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -4),
+                    moreButton!.topAnchor.constraint(equalTo: topAnchor, constant: 4),
+                    moreButton!.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -4),
+                    moreButton!.widthAnchor.constraint(equalTo: moreButton!.heightAnchor)
```
emilylaguna: I know this code is temporary but we should avoid using force unwraps.

WordPress/Classes/Stores/StatsPeriodStore.swift
```diff
@@ -1343,6 +1343,45 @@ extension StatsPeriodStore {
...
+                    for (j, _) in r.children.enumerated() {
+                        self.state.topReferrers!.referrers[i].children[j].isSpam.toggle()
+                    }
+                    DDLogInfo("Stats Period: Referrer \(r.title) isSpam set to \(self.state.topReferrers!.referrers[i].isSpam)")
+                }, failure: { error in
+                    DDLogInfo("Stats Period: Couldn't toggle spam state for referrer \(r.title), reason: \(error.localizedDescription)")
+                })
+                break
+            }
+        }
```
emilylaguna: I left some notes below on how we can improve the readability of this code:

- Variable naming should be more explicit
- There is some duplicated code that can be reduced
- We should avoid using force unwraps
- Error Handling: If the request fails for whatever reason the user may not be aware of that. I think using the `Notice` to let the user know what happened could be helpful
- Success Handling: While the table does change color it can be hard to tell that anything happened. I wonder if we can display a `Notice` for the success as well?

Also, it looks like a lot of this code is just to retrieve a `StatsReferrer` obj from the store and check the state, is there a better way to accomplish this? Maybe pass the original `StatsReferrer` from the row to this method?

What do you think?

WordPress/Classes/Stores/StatsPeriodStore.swift
```diff
@@ -1343,6 +1343,45 @@ extension StatsPeriodStore {
...
+
+        for (i, r) in state.topReferrers!.referrers.enumerated() {
+            if r.children.isEmpty {
+                if r.url?.host == referrerDomain {
+                    let currentValue = state.topReferrers!.referrers[i].isSpam
+                    statsServiceRemote?.toggleSpamState(for: referrerDomain, currentValue: currentValue, success: { [weak self] in
+                        guard let self = self else {
+                            return
+                        }
+                        self.state.topReferrers!.referrers[i].isSpam.toggle()
```
emilylaguna: TIL! I was not aware that Swift has a `toggle` method for Bools!

emilylaguna (changes requested):
👋 @nikola-milicevic! Nice work so far, I left some comments below, let me know your thoughts. 

I also encountered a few issues with the logic as well:

When comparing the ability to mark an item as spam on wordpress.com and in the PR, the PR allowed me to mark some items as spam when it shouldn't have. 

Some examples:
- Facebook
- Search Engines
- WordPress.com Reader

It may be worth exploring how this was implemented on .com, luckily this code is open source  and lives here [wp-calypso](https://github.com/Automattic/wp-calypso/). 

I'm also thinking we should add some units tests for the store mark as spam logic, what do you think?

WordPress/Classes/Stores/StatsPeriodStore.swift
nikola-milicevic: I agree with everything :)

nikola-milicevic:
@emilylaguna Thanks for the comments, I updated the code, including additional logic to distinguish between referrers that can and cannot be marked as spam, and added tests for store mark as spam logic.

WordPress/Classes/Stores/StatsPeriodStore.swift
```diff
@@ -172,15 +172,38 @@ struct PeriodStoreState {
     var postStatsFetchingStatuses = [Int: StoreFetchingStatus]()
 }
 
+protocol StatsPeriodStoreDelegate: AnyObject {
+    func didChangeSpamState(for referrerDomain: String, isSpam: Bool)
+    func chagingSpamStateForReferrerDomainFailed(oldValue: Bool)
```
emilylaguna:
```suggestion
    func changingSpamStateForReferrerDomainFailed(oldValue: Bool)
```

WordPress/Classes/Stores/StatsPeriodStore.swift
```diff
@@ -172,15 +172,38 @@ struct PeriodStoreState {
...
+}
+
+extension StatsPeriodStoreDelegate where Self: UIViewController {
+    func didChangeSpamState(for referrerDomain: String, isSpam: Bool) {
+        let markedText = NSLocalizedString("Marked as spam", comment: "Indicating that referrer was marked as spam").lowercased()
```
emilylaguna: Do we need the `.lowercased()` ?

WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
```diff
@@ -41,11 +43,17 @@ struct StatsTotalRowData {
...
         self.statSection = statSection
+        self.isReferrerSpam = isReferrerSpam
     }
 
     var hasIcon: Bool {
         return self.icon != nil || self.socialIconURL != nil || self.userIconURL != nil
     }
+
+    var canMarkReferrerAsSpam: Bool {
+        disclosureURL != nil && disclosureURL!.absoluteString.contains(name) ||
```
emilylaguna: Can we remove the force unwrap?

emilylaguna: I think this makes a good candidate for some unit testing, what do you think?

WordPress/Classes/ViewRelated/Stats/Period Stats/SiteStatsPeriodTableViewController.swift
```diff
@@ -196,7 +208,12 @@ private extension SiteStatsPeriodTableViewController {
     func viewIsVisible() -> Bool {
         return isViewLoaded && view.window != nil
     }
+}
 
+// MARK: - StatsPeriodStoreDelegate
+
+extension SiteStatsPeriodTableViewController: StatsPeriodStoreDelegate {
+    /* using default implementtion in protocol extension */
```
emilylaguna:
```suggestion
    /* using default implementation in protocol extension */
```

WordPress/Classes/ViewRelated/Stats/Shared Views/Stats Detail/SiteStatsDetailTableViewController.swift
```diff
@@ -267,6 +268,12 @@ private extension SiteStatsDetailTableViewController {
 
 }
 
+// MARK: - StatsPeriodStoreDelegate
+
+extension SiteStatsDetailTableViewController: StatsPeriodStoreDelegate {
+    /* using default implementtion in protocol extension */
```
emilylaguna:
```suggestion
    /* using default implementation in protocol extension */
```

WordPress/Classes/Stores/StatsPeriodStore.swift
```diff
@@ -172,15 +172,38 @@ struct PeriodStoreState {
     var postStatsFetchingStatuses = [Int: StoreFetchingStatus]()
 }
 
+protocol StatsPeriodStoreDelegate: AnyObject {
+    func didChangeSpamState(for referrerDomain: String, isSpam: Bool)
+    func chagingSpamStateForReferrerDomainFailed(oldValue: Bool)
+}
+
+extension StatsPeriodStoreDelegate where Self: UIViewController {
+    func didChangeSpamState(for referrerDomain: String, isSpam: Bool) {
```
emilylaguna, line 181: This is a very interesting approach that I haven't seen used before. Nice work.

emilylaguna (changes requested):
Nice work implementing the logic, it looks to be working nicely now 🙌

I left a few comments below, let me know what you think.

WordPress/Classes/Stores/StatsPeriodStore.swift
nikola-milicevic: Oops 😁

WordPress/Classes/Stores/StatsPeriodStore.swift
nikola-milicevic: Well not sure what's the general approach (if we need lowercase to be displayed at all, and not `www.someting.com Marked as spam`). Should localized file contain already lowercase string, or we do the modification at the call site (like i did)?

WordPress/Classes/ViewRelated/Stats/Shared Views/StatsTotalRow.swift
nikola-milicevic: Yes, I agree, I'll add tests for this. Also, I thought force unwrap is okay in this case since with this short circuit evaluation crash can never happen (since first check is `!= nil`). But I can remove the `!` operator.

emilylaguna: >Also, I thought force unwrap is okay in this case since with this short circuit evaluation crash can never happen (since first check is != nil). But I can remove the ! operator

While this is an appropriate use we like to avoid using force unwraps in the codebase.

WordPress/Classes/Stores/StatsPeriodStore.swift
emilylaguna: I'd say to use the lowercase string directly instead of `lowercased` since I trust the translator to use the correct lowercased string more than I trust swifts interpretation.

emilylaguna (approved):
LGTM!

PR #16682, opened 2021-06-14
Adds blogging reminders to site settings.

diegoreymendez:
Adds blogging goals to settings.

Migrates some legacy code to Swift.

## To test:

You should make sure to test all of the options under Site Settings > General.
You should make sure to test these using both an iPad and an iPhone.
When checking the values are updated in the web, I recommend going into WP Admin > Settings > General, which is where all of these settings are stored.

### Site Title:

- Make sure you can edit the site title
- Make sure the edits can be seen on the web
- Make sure when exiting the editing screen, the cell value is properly updated

### Tagline:

- Make sure you can edit the tagline
- Make sure the edits can be seen on the web
- Make sure when exiting the editing screen, the cell value is properly updated

### Address:

- This should not be editable or selectable in any way.

### Privacy

- Make sure you can edit the privacy setting
- Make sure changing privacy settings is reflected on the web
- Make sure when exiting the privacy setting picker screen, the cell value is properly updated

### Language

- Make sure you can edit the language setting
- Make sure changing language settings is reflected on the web
- Make sure when exiting the language setting picker screen, the cell value is properly updated

### Timezone

- Make sure you can edit the timezone setting
- Make sure changing timezone settings is reflected on the web
- Make sure when exiting the timezone setting picker screen, the cell value is properly updated

### Blogging Reminders

- You should just see a cell with value "Undefined" on it.
- Make sure the cell isn't visible when you switch the feature flag off.
- Tapping on this cell should launch the blogging reminders flow.

## Regression Notes

To be able to add a new option under site settings, I needed to make some bigger changes to how we were managing the data for the table view.  I decided the easiest and cleanest solution was to switch some of the legacy code to Swift.

This means there's a change some of the old code could be broken.

I tested all of the changed settings manually, and wrote a step by step to verify them during the review of this PR.

None

## PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Site Settings/SiteSettingsViewController+Swift.swift
```diff
@@ -168,3 +168,233 @@ extension SiteSettingsViewController {
...
+
+        if Feature.enabled(.bloggingReminders) {
+            visibleRows.append(.bloggingReminders)
+        }
+
+        return visibleRows
+    }
+
+    @objc
+    func generalSettingsVisibleRowCount() -> Int {
```
frosty: Nitpick, this could be a var

WordPress/Classes/ViewRelated/Blog/Site Settings/SiteSettingsViewController+Swift.swift
```diff
@@ -168,3 +168,233 @@ extension SiteSettingsViewController {
...
+        case title
+        case tagline
+        case url
+        case privacy
+        case language
+        case timezone
+        case bloggingReminders
+    }
+
+    func generalSettingsVisibleRows() -> [GeneralSettingsRow] {
```
frosty: I'm not 100% sure about the name `visible` here – to me, it's usually used to indicate that something is actually _visible_ on screen, like visible rows in a tableview. Could we just call this `generalSettingsRows`? Some other suggestions could be `activeGeneralSettingsRows`, `currentGeneralSettingsRows`, `availableGeneralSettingsRows`, `enabledGeneralSettingsRows`.

WordPress/Classes/ViewRelated/Blog/Site Settings/SiteSettingsViewController+Swift.swift
```diff
@@ -168,3 +168,233 @@ extension SiteSettingsViewController {
...
+            presentBloggingRemindersFlow(indexPath: indexPath)
+            break
+        default:
+            break
+        }
+    }
+
+    // MARK: - Cell Configuration
+
+    private func configureForTitle(_ cell: SettingTableViewCell) {
```
frosty: How about `configureCellForTitle`, `configureCellForTagline`, etc?

frosty (commented):
Looking great functionally, and great improvements moving this over to Swift! Just a few code level comments.

WordPress/Classes/ViewRelated/Blog/Site Settings/SiteSettingsViewController+Swift.swift
diegoreymendez: I think you're right, and there's no need to have anything more than `generalSettingsRows`.  Good observation!

WordPress/Classes/ViewRelated/Blog/Site Settings/SiteSettingsViewController+Swift.swift
diegoreymendez: Sounds good.

WordPress/Classes/ViewRelated/Blog/Site Settings/SiteSettingsViewController+Swift.swift
diegoreymendez: I wanted to hint that it was better to cache the value, rather than use it multiple times directly, but I believe in most scenarios this value won't be used multiple times in a row anyway.

diegoreymendez: Resolved.  Did the same for `generalSettingsRows()`.

diegoreymendez:
Ready for another look @frosty.

frosty (approved):
Thanks for making the changes!

PR #16688, opened 2021-06-15
Address compiler warnings in UI tests targets

mokagio:
Addresses two minor warnings:

- Using '_' to ignore the result of a Void-returning function is redundant
- Unused `let` assignent

If CI builds the app, we're good.

## Regression Notes
N.A.

N.A.

N.A.

---

### DIFF
--- a/WordPress/WordPressUITests/Screens/MySiteScreen.swift
+++ b/WordPress/WordPressUITests/Screens/MySiteScreen.swift
@@ -43,7 +43,6 @@ class MySiteScreen: BaseScreen {
 
     init() {
         let app = XCUIApplication()
-        let blogTable = app.tables[ElementStringIDs.blogTable]
         tabBar = TabNavComponent()
         removeSiteButton = app.cells[ElementStringIDs.removeSiteButton]
         removeSiteSheet = app.sheets.buttons.element(boundBy: 0)
--- a/WordPress/WordPressUITests/Tests/LoginTests.swift
+++ b/WordPress/WordPressUITests/Tests/LoginTests.swift
@@ -95,7 +95,7 @@ class LoginTests: XCTestCase {
     // Unified self hosted login/out
     // Replaces testSelfHostedUsernamePasswordLoginLogout
     func testSelfHostedLoginLogout() {
-        _ = PrologueScreen().selectSiteAddress()
+        PrologueScreen().selectSiteAddress()
             .proceedWith(siteUrl: WPUITestCredentials.selfHostedSiteAddress)
             .proceedWith(username: WPUITestCredentials.selfHostedUsername, password: WPUITestCredentials.selfHostedPassword)
             .verifyEpilogueDisplays(siteUrl: WPUITestCredentials.selfHostedSiteAddress)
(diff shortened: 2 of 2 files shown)


WordPress/WordPressUITests/Tests/LoginTests.swift
```diff
@@ -95,7 +95,7 @@ class LoginTests: XCTestCase {
     // Unified self hosted login/out
     // Replaces testSelfHostedUsernamePasswordLoginLogout
     func testSelfHostedLoginLogout() {
-        _ = PrologueScreen().selectSiteAddress()
+        PrologueScreen().selectSiteAddress()
```
mokagio, line 98: > Using '_' to ignore the result of a Void-returning function is redundant

This syntax is used to ignore the return value of a method if you want to call the method but don't care about the return value, which can happen for methods that perform a side-effect while also returning a value (which, by the way, is not a good pattern, either perform a side-effect or return a value, avoid doing both).

We ignore the return value with `_` because the compiler warns about unused assignments ("hey, you called `let carrots = pickCarrots(quantity: 3)` but never used your `carrots`").

But ☝️ the method at the end of this chain of calls returned `Void` (nothing) so there was nothing to ignore and the syntax was redundant.

WordPress/WordPressUITests/Screens/MySiteScreen.swift
```diff
@@ -43,7 +43,6 @@ class MySiteScreen: BaseScreen {
 
     init() {
         let app = XCUIApplication()
-        let blogTable = app.tables[ElementStringIDs.blogTable]
```
mokagio, line 46: Unused code

🧹💨 🧹💨 🧹💨

pachlava (approved):
No more warnings about redundant usage of `_` statement for a Void-returning functions, and also no warnings coming from UI Tests target in overall 👍.  All tests in the `WordPressUITests` scheme are passing locally with iPhone 12. 

Side question: is it normal that the total number of warnings that I see in the project is close to 40? I suspect I might have something misconfigured.

mokagio:
> Is it normal that the total number of warnings that I see in the project is close to 40? I suspect I might have something misconfigured.

Were you expecting _less_ or _more_? 😅 

Jokes aside, yeah that's normal. Not good, but a known thing in the project.

I'm of the opinion that the only good number of warnings is _zero_, but it's a sad truth of software development and of projects that move fast with many people working on them that the warnings count tends to grow.

There's also to add that every new version of Xcode usually introduces new warnings, be them deprecations or stricter compiler settings. The bigger the codebase, the harder it gets to stay on top of them.

Finally, there's an argument to be made for the return of investment in addressing warnings vs building valuable "user facing" features. There's also a counter argument 😄 the _broken window theory_ that Jeff Atwood (@coding-horror) describes in this [post](https://blog.codinghorror.com/the-broken-window-theory/).

pachlava:
Thank you for the interesting read @mokagio!
I wondered some time ago why Simplenote doesn't have a 'recovery' sprint from time to time, to fix such things (I think that even the 'behind the curtain' improvements are user-facing in the end, because they indirectly affect the future quality of development). Now, I understand that SN is nowhere near the state of other repos 😄 .

Also, this:
> Were you expecting less or more? 

😅 I'm spoiled by SN

PR #16693, opened 2021-06-16
Wires the blogging reminders scheduler to the UI.

diegoreymendez:
## What to test:

Test that your day selection is persisted for each blog you  configure.
Test that changing the calendar locale between executions of the reminders flow doesn't affect the day selection.

## What to ignore:

Even though the notifications scheduling is wired, let's ignore testing notifications for now.  Since this feature is behind a flag, we can make this call safely, and focus on testing the notifications at a future time.

## To test:

Test the blogging reminders flow a few times on the same blog, then switch to another and come back to the original one.  The day selection should always be persisted across executions of the flow.

Also, try:

1. Set the device region to "United States"
2. Start the reminders flow.
3. Make sure "Sunday" is the first day shown in the day picker.
4. Select the days for your reminders
5. Complete the flow, so that the reminders are stored.
6. Change the region of your device to "Argentina"
7. Re-run the blogging reminders flow
8. Make sure the first day shown in the day picker is "Monday"
9. Make sure the day selection is correct and matches what it was when the region was "United States".

## Regression Notes

None.

PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
+        daysBottomInnerStackView.addArrangedSubviews(bottomRow.map({ createButton($0) }))
+    }
 
     // MARK: - Actions
 
     @objc private func notifyMeButtonTapped() {
         tracker.buttonPressed(button: .continue, screen: .dayPicker)
 
+        do {
+            try scheduler.schedule(.weekdays(weekdays))
```
diegoreymendez: This is where the scheduling happens.  The issue is that I designed this call (`scheduler.schedule(...)`) without a completion block.

My thinking is that I need to add a completion block here, and handle any errors in the notifications scheduling, but I'm not sure what we should do if, for example, a single notification raises an error when we try to schedule it (remember that each day the user picks is one notification).

Do we show the user an error?  Do we silently ignore the error?

diegoreymendez: We decided to leave this as-is for now, and review the error handling at another time.  This should be fine as the feature is not public.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
```diff
@@ -48,7 +48,7 @@ class BloggingRemindersScheduler {
     }
 
     enum Weekday: Int, Codable, Comparable {
-        case sunday = 1 // Keep this at 1 to match Apple's `DateComponents`' weekday number.
+        case sunday = 0 // Keep this at 1 to match Apple's `DateComponents`' weekday number.
```
frosty: If we're changing this to 0 we should update the comment

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlow.swift
```diff
@@ -1,19 +1,25 @@
...
+    static func present(from viewController: UIViewController, for blog: Blog, source: BloggingRemindersTracker.FlowStartSource) {
 
         let blogType: BloggingRemindersTracker.BlogType = blog.isHostedAtWPcom ? .wpcom : .selfHosted
 
         let tracker = BloggingRemindersTracker(blogType: blogType)
         tracker.flowStarted(source: source)
 
         // TODO: Check whether we've already presented this flow to the user. @frosty
-        let flowIntroViewController = BloggingRemindersFlowIntroViewController(tracker: tracker)
+        let flowIntroViewController = BloggingRemindersFlowIntroViewController(blogURIRepresentation: blog.objectID.uriRepresentation(), tracker: tracker)
```
frosty: Why can't we just pass a blog in here, and the view controller take the uriRepresentation itself?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowIntroViewController.swift
```diff
@@ -145,8 +147,14 @@ class BloggingRemindersFlowIntroViewController: UIViewController {
     @objc private func getStartedTapped() {
         tracker.buttonPressed(button: .continue, screen: .main)
 
-        let flowSettingsViewController = BloggingRemindersFlowSettingsViewController(tracker: tracker)
-        navigationController?.pushViewController(flowSettingsViewController, animated: true)
+        do {
+            let flowSettingsViewController = try BloggingRemindersFlowSettingsViewController(blogURIRepresentation: blogURIRepresentation, tracker: tracker)
```
frosty: I feel like it's unusual to have instantiating a view controller throw an error. I think it might be best to continue to show the view controller, but it can display an error if it's been unable to create the scheduler.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
-
-    /// Localized short weekday names, starting at the correct weekday for the current calendar
-    private let days: [String] = {
-        var calendar = Calendar.current
-        calendar.locale = Locale.autoupdatingCurrent
-        let firstWeekday = calendar.firstWeekday
-        var symbols = calendar.shortWeekdaySymbols
+        let createButton: (Int) -> CalendarDayToggleButton = { [unowned self] dayIndex in
+            let isSelected: Bool
+            let localizedDayIndex = (dayIndex + calendar.firstWeekday - 1) % calendar.shortWeekdaySymbols.count
```
frosty: I wonder if this would be good moved to a small calendar extension?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
+
+            let weekday = BloggingRemindersScheduler.Weekday(rawValue: localizedDayIndex) ?? {
+                // The rawValue above should always fall within 0 ..< 7 and this case
+                // should not be possible.  We're still going to handle it by logging
+                // an error and returning .monday so that the app keeps running.
+                //
+                // Is there any other sensible approach to handle this?
+                //
+                DDLogError("Cannot initialize BloggingRemindersScheduler.Weekday with rawValue: \(localizedDayIndex)")
+                return .monday
```
frosty: I wonder if it would be better to just not create a button in this case?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
-        daysBottomInnerStackView.addArrangedSubviews(bottomRow.map({ CalendarDayToggleButton(weekday: $0) }))
-    }
-
-    /// Localized short weekday names, starting at the correct weekday for the current calendar
-    private let days: [String] = {
-        var calendar = Calendar.current
-        calendar.locale = Locale.autoupdatingCurrent
-        let firstWeekday = calendar.firstWeekday
-        var symbols = calendar.shortWeekdaySymbols
+        let createButton: (Int) -> CalendarDayToggleButton = { [unowned self] dayIndex in
```
frosty: Could we just move this to a standard method?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
+                //
+                DDLogError("Cannot initialize BloggingRemindersScheduler.Weekday with rawValue: \(localizedDayIndex)")
+                return .monday
+            }()
+
+            isSelected = weekdays.contains(weekday)
+
+            return CalendarDayToggleButton(
+                weekday: weekday,
+                dayName: self.calendar.shortWeekdaySymbols[localizedDayIndex],
```
frosty: Nitpick, in the designs the day names are uppercased

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
+                        weekday == button.weekday
+                    }
+                }
+            }
+        }
 
-        // Switch around the order of days so that the correct day is at the beginning
-        let firstWeekdayToEnd = symbols[firstWeekday-1 ..< Calendar.current.shortWeekdaySymbols.count]
-        let beginningToFirstWeekday = symbols[0 ..< firstWeekday - 1]
+        let topRow = 0 ..< 4
```
frosty: Thinking about it, we should probably put this 4 as a constant – `Metrics.topRowDayCount` perhaps?

frosty (commented):
Looking good! I tested out with a range of locales and everything persisted as I expected. I've just left some fairly minor code level comments / suggestions.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: Yes... that's probably the best option.  I'll change this.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlow.swift
diegoreymendez: Since the VCs are keeping strong references to this, I wanted to avoid passing a full CoreData object.  But I'm changing this to be a blog, as it may make the logic easier to follow.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowIntroViewController.swift
diegoreymendez: As chatted via DM... the reason this was done this way is because it's an error scenario that's very unlikely to happen, but that would prevent the flow from working fine.  So my thinking is I didn't want to let the user go forward as if nothing happened.

By interrupting the flow I'm hoping we'll get at least reports of an issue.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -124,12 +124,37 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
         return button
     }()
 
+    // MARK: - Properties
+
+    let calendar: Calendar
```
Gio2018: can `calendar` be `private` too?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
Gio2018: +1

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
-        var calendar = Calendar.current
-        calendar.locale = Locale.autoupdatingCurrent
-        let firstWeekday = calendar.firstWeekday
-        var symbols = calendar.shortWeekdaySymbols
+        let createButton: (Int) -> CalendarDayToggleButton = { [unowned self] dayIndex in
+            let isSelected: Bool
+            let localizedDayIndex = (dayIndex + calendar.firstWeekday - 1) % calendar.shortWeekdaySymbols.count
+
+            let weekday = BloggingRemindersScheduler.Weekday(rawValue: localizedDayIndex) ?? {
+                // The rawValue above should always fall within 0 ..< 7 and this case
```
Gio2018: can we be more concise with the comments? I don't think the history of a decision will be that useful to anyone who will work on this code in the future.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -227,31 +252,67 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
+            DDLogError("Error scheduling notifications: \(error)")
+
+            // TODO: Properly handle error situation.
+            //
+            // For now I'm dismissing the flow in this scenario, although I think showing an inline
+            // error message would be best (an error that's informative and non-blocking, that lets
+            // the user decide if to dismiss or continue).
+            //
+            // I've pinged @osullivanchris about this and will update when I get feedback.
+            //
```
Gio2018: same comment about the code comments above

Gio2018 (approved):
This worked well @diegoreymendez ! Left a few code comments

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: This code was removed.

That said what's being documented is not the history of the decision, but the "why" of a very special scenario which may not be clear by just reading the code.  By documenting the "why" we make sure whoever reads the code can understand what is required to change it or remove it.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: As replied above... I disagree.  This is exactly what documentation should look like.

What's being documented is not the history of the decision, but the "why" of a very special scenario which may not be clear by just reading the code.  The question is "why are we doing this?", and the documentation answers that question.

By documenting the "why" we make sure whoever reads the code in the future can understand why it was done this way, and what is required to change it or remove it, and can update the comment if needed.

As I learnt a long time ago, never document the "what", always document the "why".

That said I'll remove the last line of the documentation (where I say I'll ping Chris) as that's already resolved.

Just to complete the idea, here's an example of poor code documentation (documenting the "what", instead of the "why"):

```swift
// Assign the last index
let lastIndex = 4
```

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: Done

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: I'm not sure there's value in doing this as the resulting index aims specifically to match a `rawValue` of `BloggingRemindersScheduler.Weekday`.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: I ended up updating this anyway, but I'll leave this as unresolved because it describes the way in which I document the code.

diegoreymendez:
Ready for another look.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -274,6 +336,20 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
 
         dismiss(animated: true, completion: nil)
     }
+
+    // MARK: - Completion Paths
+
+    private func presentCompletionViewController() {
+        let viewController = BloggingRemindersFlowCompletionViewController(tracker: self.tracker)
```
frosty: Nitpick, we can omit `self` here and in the line below.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -224,47 +249,84 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
+        guard let weekday = BloggingRemindersScheduler.Weekday(rawValue: localizedDayIndex) else {
+            return nil
+        }
+
+        let isSelected = weekdays.contains(weekday)
+
+        return CalendarDayToggleButton(
+            weekday: weekday,
+            dayName: self.calendar.shortWeekdaySymbols[localizedDayIndex].uppercased(),
+            isSelected: isSelected) { button in
```
frosty: We need a weak self here, otherwise we have a retain cycle between the buttons and the VC and neither gets released.

frosty (changes requested):
Just two more small comments and then I think we're good!

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: Good catch!

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
Gio2018: I was not arguing against the why, just the length of the comment. Too much information in a comment can be confusing, nor I think comments should be used as a todo list (the action to ping someone that you just removed, for example).

diegoreymendez:
Ready again @frosty

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -271,7 +271,11 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
             dayName: self.calendar.shortWeekdaySymbols[localizedDayIndex].uppercased(),
-            isSelected: isSelected) { button in
+            isSelected: isSelected) { [weak self] button in
+
+            guard let self = self else {
+                return
+            }
 
             if button.isSelected {
                 self.weekdays.append(button.weekday)
```
frosty, line 281: We could just do `self?.weekdays...` in this if / else instead of the guard, but up to you :)

frosty (approved):
Approved! Thanks for fixing :) One final small comment, up to you if you change it!

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
diegoreymendez: I usually prefer the guard because really none of the code below needs to run if `self == nil`.

PR #16694, opened 2021-06-16
Blogging reminders: Prompt for permission

frosty:
This PR implements prompting the user for permission to send push notifications, and provides a guidance screen if push is disabled.

|   |   |
|---|---|
| reminders-notifications |  |

**To test**

### 1: Fresh install, allow

* Build and run a fresh install of the app and log in.
* Trigger the Blogging Reminders flow from the card, and tap the **Notify me** button on the settings screen.
* You should see a system prompt for notifications. Tap Allow.
* You should be taken to the end of the flow.

### 2: Fresh install, don't allow + analytics

* Repeat the above steps but choose Don't Allow.
* You should be shown the new guidance screen.
* Tap **Turn on notifications**, and ensure you're taken to the system settings for WordPress.
* Don't change any settings and return to the app. You should still be on the same screen.
* Return to settings and enable Notifications.
* Return to the app. You should be taken to the end of the flow.
* Check your console for analytics logs. You should see the following at the appropriate place in the flow:

```
Tracked: blogging_reminders_screen_shown <blog_type: wpcom, screen: enable_notifications>
Tracked: blogging_reminders_button_pressed <blog_type: wpcom, button: notificationSettings, screen: enable_notifications>
```

You can also try dismissing the guidance screen to see:

```
Tracked: blogging_reminders_button_pressed <blog_type: wpcom, button: dismiss, screen: enable_notifications>
Tracked: blogging_reminders_flow_completed <blog_type: wpcom>
```

### 3: Notifications already enabled

* After following the previous step, you should have the app installed and notifications enabled.
* Follow the blogging reminders flow again and see that you're not prompted for permission, and are able to navigate straight through the flow.

### 4: Notifications already disabled

* Go back to system settings and disable Notifications for WordPress.
* Run through the reminders flow again, and you should once again see the new guidance screen.
* Tap Turn on notifications, enable them in Settings, and you should be able to proceed.

### 5: Analytics

* Navigate through 

## Regression Notes

None, this should only affect the reminders flow.

N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
@@ -252,8 +252,21 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
     @objc private func notifyMeButtonTapped() {
         tracker.buttonPressed(button: .continue, screen: .dayPicker)
 
-        let flowCompletionVC = BloggingRemindersFlowCompletionViewController(tracker: tracker)
-        navigationController?.pushViewController(flowCompletionVC, animated: true)
+        InteractiveNotificationsManager.shared.requestAuthorization { [weak self] allowed in
+            guard let self = self else {
+                return
+            }
+
+            DispatchQueue.main.async {
+                if allowed {
+                    let flowCompletionVC = BloggingRemindersFlowCompletionViewController(tracker: self.tracker)
+                    self.navigationController?.pushViewController(flowCompletionVC, animated: true)
+                } else {
+                    let pushPromptVC = BloggingRemindersPushPromptViewController(tracker: self.tracker)
+                    self.navigationController?.pushViewController(pushPromptVC, animated: true)
+                }
+            }
+        }
     }
 
     @objc private func dismissTapped() {
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
@@ -258,13 +258,13 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
             }
 
             DispatchQueue.main.async {
+                let targetVC: UIViewController
                 if allowed {
-                    let flowCompletionVC = BloggingRemindersFlowCompletionViewController(tracker: self.tracker)
-                    self.navigationController?.pushViewController(flowCompletionVC, animated: true)
+                    targetVC = BloggingRemindersFlowCompletionViewController(tracker: self.tracker)
                 } else {
-                    let pushPromptVC = BloggingRemindersPushPromptViewController(tracker: self.tracker)
-                    self.navigationController?.pushViewController(pushPromptVC, animated: true)
+                    targetVC = BloggingRemindersPushPromptViewController(tracker: self.tracker)
                 }
+                self.navigationController?.pushViewController(targetVC, animated: true)
             }
         }
     }
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
@@ -125,7 +125,7 @@ class BloggingRemindersPushPromptViewController: UIViewController {
         // If a parent VC is being dismissed, and this is the last view shown in its navigation controller, we'll assume
         // the flow was completed.
         if isBeingDismissedDirectlyOrByAncestor() && navigationController?.viewControllers.last == self {
-            tracker.flowCompleted()
+            tracker.flowDismissed(source: .enableNotifications)
         }
 
     }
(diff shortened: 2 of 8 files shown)


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -252,8 +252,21 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
...
-        navigationController?.pushViewController(flowCompletionVC, animated: true)
+        InteractiveNotificationsManager.shared.requestAuthorization { [weak self] allowed in
+            guard let self = self else {
+                return
+            }
+
+            DispatchQueue.main.async {
+                if allowed {
+                    let flowCompletionVC = BloggingRemindersFlowCompletionViewController(tracker: self.tracker)
+                    self.navigationController?.pushViewController(flowCompletionVC, animated: true)
```
diegoreymendez: Nitpicky, but maybe we can take the `pushViewController` line out of the `if` block, since it's called in both paths - the conditional blocks seems to only be selecting which one the next VC is going to be.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
```diff
@@ -0,0 +1,248 @@
...
+        super.viewDidAppear(animated)
+    }
+
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+
+        // If a parent VC is being dismissed, and this is the last view shown in its navigation controller, we'll assume
+        // the flow was completed.
+        if isBeingDismissedDirectlyOrByAncestor() && navigationController?.viewControllers.last == self {
+            tracker.flowCompleted()
```
diegoreymendez: Assuming there's no weird race-condition scenario where the user manages to schedule the notifications and then dismiss this VC, this should probably always be `tracker.flowDismissed()`.

diegoreymendez (changes requested):
Looks really nice!

I did find one tracking issue that needs to be addressed though.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
frosty: Thanks, good catch!

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
frosty: Yup, good improvement 👍

frosty:
@diegoreymendez Addressed your feedback – ready for another look!

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
```diff
@@ -0,0 +1,248 @@
...
+    let tracker: BloggingRemindersTracker
+
+    // MARK: - Initializers
+
+    init(tracker: BloggingRemindersTracker) {
+        self.tracker = tracker
+
+        super.init(nibName: nil, bundle: nil)
+
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationBecameActive), name: UIApplication.didBecomeActiveNotification, object: nil)
```
Gio2018, line 93: can we do

```
NotificationCenter.default.addObserver(forName: UIApplication.didBecomeActiveNotification, object: nil, queue: nil) { [weak self] _ in
            self?.refreshPushAuthorizationStatus()
 }
```

This would not require the called method to be exposed to Objective C and we can avoid adding one method `applicationBecameActive()` to wrap another one.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
```diff
@@ -0,0 +1,248 @@
...
+        calculatePreferredContentSize()
+    }
+
+    private func calculatePreferredContentSize() {
+        let size = CGSize(width: view.bounds.width, height: UIView.layoutFittingCompressedSize.height)
+        preferredContentSize = view.systemLayoutSizeFitting(size)
+    }
+
+    @objc
+    private func applicationBecameActive() {
```
Gio2018, line 144: see above comment on line 93

Gio2018 (approved):
This works great @frosty ! Just a couple of code comments, other than that :shipit:

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
frosty: Thanks for the suggestion! I went with the selector-based observer because they don't need to be manually unregistered unlike block-based ones. If I do block-based, we need a variable to keep hold of the token, and to implement `deinit` to unregister the block later. (see [this blog](https://oleb.net/blog/2018/01/notificationcenter-removeobserver/) and the [Apple docs](https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver) for more info).

diegoreymendez (approved):

PR #16695, opened 2021-06-16
Improve navigation transitions in the Blogging Reminders setup flow on iPad

Gio2018:
Fixes #NA

This PR improves the navigation in the Blogging Reminders setup flow on iPad.

To test:
- build/run on iPad
- Tap on "Get Started" in the blogging reminders card
- Navigate through the screens (tapping "Setup Goals" then "Next") and make sure the transition animations look smooth (see example below).
- Build/run on iPhone and make sure the same animations still look good

<p align=center>

</p>

## Regression Notes
None: this change did not change anything existing, just adds a custom animation to a custom navigation controller (`BloggingRemindersNavigationController`)

-

-

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersAnimator.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersAnimator.swift
@@ -1,0 +1,37 @@
+/// A transition animator that moves in the pushed view controller horizontally.
+/// Does not handle the pop animation since the BloggingReminders setup flow does not allow to navigate back.
+class BloggingRemindersAnimator: NSObject, UIViewControllerAnimatedTransitioning {
+
+    func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
+        return 0.2
+    }
+
+    func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
+
+        guard let sourceViewController =
+                transitionContext
+                .viewController(forKey: UITransitionContextViewControllerKey.from),
+              let destinationViewController =
+                transitionContext
+                .viewController(forKey: UITransitionContextViewControllerKey.to) else {
+
+            return
+        }
+        // final position of the destination view
+        let endFrame = transitionContext.finalFrame(for: destinationViewController)
+
+        // initial position of the destination view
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersAnimator.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
+        return 0.2
+    }
+
+    func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
+
+        guard let sourceViewController =
+                transitionContext
+                .viewController(forKey: UITransitionContextViewControllerKey.from),
```
diegoreymendez: Nitpicky: The line breaking between `transitionContext` and its method `.viewController(...)` seems a bit nonstandard.  IMHO it makes readability suffer slightly.

Any thoughts on standardizing this to be:

```swift
guard let sourceViewController =
        transitionContext.viewController(forKey: UITransitionContextViewControllerKey.from)
```

diegoreymendez (approved):
Looking great!

I've added a comment that IMHO would help improve readability, but it's up to you.

Since auto-merge is disabled and you'll get a chance to evaluate my comment I'm approving this PR.

frosty (commented):
Changes look good! However, can we also animate out the old view, rather than instantly setting the alpha to zero? Could we just push its view over to the left as well to match the incoming view? As it is, it looks a little odd that it just disappears:

reminders-anim-1

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersAnimator.swift
Gio2018: sure

Gio2018:
> Changes look good! However, can we also animate out the old view, rather than instantly setting the alpha to zero? Could we just push its view over to the left as well to match the incoming view? As it is, it looks a little odd that it just disappears:
> 
> reminders-anim-1

I tried other combinations of animations but honestly in 0.2 seconds they don't make any difference, so I just went with the simplest.

PR #16700, opened 2021-06-17
[Gutenberg] Add Sentry React native SDK

fluiddot:
`gutenberg-mobile` PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3629
`gutenberg` PR: https://github.com/WordPress/gutenberg/pull/32768
`Automattic-Tracks-iOS` PR: https://github.com/Automattic/Automattic-Tracks-iOS/pull/183

To test:
Follow the testing instructions from `gutenberg` PR: https://github.com/WordPress/gutenberg/pull/32768

## Regression Notes
The lanes (`Fastfile`) related to building and uploading the build as they're using a new lane I've added for uploading the source maps to Sentry.

I manually tested the lane by running the command `bundle exec fastlane upload_gutenberg_sourcemaps internal:true` but I haven't tested the ones related to building and uploading builds.

N/A

PR submission checklist:


WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -930,6 +930,12 @@ extension GutenbergViewController: GutenbergBridgeDelegate {
                 handleMissingBlockAlertButtonPressed()
         }
     }
+
+    func gutenbergDidLogSentryEnvelope(_ envelope: [String: Any]) {
+        DispatchQueue.main.async {
+            WordPressAppDelegate.crashLogging?.logEnvelope(envelope)
+        }
```
fluiddot: Calls to `WordPressAppDelegate` are required to be done in the main thread.

fastlane/Fastfile
```diff
@@ -664,6 +673,41 @@ import "./ScreenshotFastfile"
...
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js.map'), File.join(sourcemaps_folder, 'main.jsbundle.map'))
+
+    sentry_upload_sourcemap(
+      auth_token: get_required_env("SENTRY_AUTH_TOKEN"),
+      org_slug: 'a8c',
+      project_slug: 'wordpress-ios',
+      version: build_version,
+      dist: build_version,
+      rewrite: true,
+      strip_common_prefix: true,
```
fluiddot: When the React native bundle is generated, the source map file references include the local machine path, with these options it automatically strips this part.

fastlane/Fastfile
```diff
@@ -664,6 +673,41 @@ import "./ScreenshotFastfile"
...
+  #####################################################################################
+  lane :upload_gutenberg_sourcemaps do | options |
+    build_version = ios_get_build_version(internal: options[:internal])
+    
+    gutenberg_bundle = File.join(PROJECT_ROOT_FOLDER, "Pods/Gutenberg/bundle/ios")
+    sourcemaps_folder = File.join(gutenberg_bundle, "sourcemaps")
+    FileUtils.mkdir_p(sourcemaps_folder)
+    # Name of bundle file must be set to main.jsbundle
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js'), File.join(sourcemaps_folder, 'main.jsbundle'))
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js.map'), File.join(sourcemaps_folder, 'main.jsbundle.map'))
```
fluiddot: It's important that the source map files have specific names, otherwise, Sentry doesn't un-minify the stack traces.

fastlane/Fastfile
```diff
@@ -664,6 +673,41 @@ import "./ScreenshotFastfile"
...
+  # This lane uploads the Gutenberg source maps
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane upload_gutenberg_sourcemaps [internal:<internal>]
+  #
+  # Example:
+  # bundle exec fastlane upload_gutenberg_sourcemaps internal:true
+  #####################################################################################
+  lane :upload_gutenberg_sourcemaps do | options |
+    build_version = ios_get_build_version(internal: options[:internal])
```
fluiddot, line 811: The build version is different depending on the `internal` option, here is an example:
- `internal: true`: `17.6.0.20210614` ([reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/config/Version.internal.xcconfig))
- `internal: false`: `17.6.0.0` ([reference](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/config/Version.public.xcconfig))

fluiddot:
@mokagio @jkmassel I've added a new lane to the `Fastfile` to upload the React native source maps to Sentry, I'd appreciate it if you could take a look at that part just in case I'm overlooking anything, thanks 🙇 !

antonis (approved):
Tested via https://github.com/WordPress/gutenberg/pull/32768#pullrequestreview-692939310 
The code changes LGTM too 🎉

fastlane/Fastfile
```diff
@@ -418,6 +418,9 @@ platform :ios do
       dsym_path: lane_context[SharedValues::DSYM_OUTPUT_PATH]
     )
 
+    # Upload Gutenberg source maps
+    upload_gutenberg_sourcemaps(internal: true)
```
mokagio: Nitpick: given the well chosen method name, the comment is redundant 😄 

```suggestion
    upload_gutenberg_sourcemaps(internal: true)
```

fastlane/Fastfile
```diff
@@ -474,6 +477,9 @@ platform :ios do
       project_slug: 'wordpress-ios',
       dsym_path: lane_context[SharedValues::DSYM_OUTPUT_PATH]
     )
+
+    # Upload Gutenberg source maps
+    upload_gutenberg_sourcemaps(internal: true)
```
mokagio: Nitpick: given the well chosen method name, the comment is redundant 😄 

```suggestion
    upload_gutenberg_sourcemaps(internal: true)
```

fastlane/Fastfile
```diff
@@ -519,6 +525,9 @@ platform :ios do
       dsym_path: lane_context[SharedValues::DSYM_OUTPUT_PATH]
     )
 
+    # Upload Gutenberg source maps
+    upload_gutenberg_sourcemaps(internal: false)
```
mokagio: Nitpick: given the well chosen method name, the comment is redundant 😄 

```suggestion
    upload_gutenberg_sourcemaps(internal: false)
```

fastlane/Fastfile
```diff
@@ -677,6 +686,41 @@ platform :ios do
...
+  #####################################################################################
+  lane :upload_gutenberg_sourcemaps do | options |
+    build_version = ios_get_build_version(internal: options[:internal])
+    
+    gutenberg_bundle = File.join(PROJECT_ROOT_FOLDER, "Pods/Gutenberg/bundle/ios")
+    sourcemaps_folder = File.join(gutenberg_bundle, "sourcemaps")
+    FileUtils.mkdir_p(sourcemaps_folder)
+    # Name of bundle file must be set to main.jsbundle
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js'), File.join(sourcemaps_folder, 'main.jsbundle'))
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js.map'), File.join(sourcemaps_folder, 'main.jsbundle.map'))
```
mokagio: Thanks for the info. I think this comment would be useful in the lane, too.

```suggestion
    # Name of bundle file must be set to main.jsbundle.
    #
    # It's important that the source map files have specific names, otherwise, Sentry 
    # won't un-minify the stack traces.
    FileUtils.cp(File.join(gutenberg_bundle, 'App.js'), File.join(sourcemaps_folder, 'main.jsbundle'))
    FileUtils.cp(File.join(gutenberg_bundle, 'App.js.map'), File.join(sourcemaps_folder, 'main.jsbundle.map'))
```

fastlane/Fastfile
mokagio: Thank you for the info. I think it would be valuable to add them in the code, too:

```suggestion
      # When the React native bundle is generated, the source map file references
      # include the local machine path, with the `rewrite` and `strip_common_prefix` 
      # options Sentry automatically strips this part.
      rewrite: true,
      strip_common_prefix: true,      
```

fastlane/Fastfile
```diff
@@ -677,6 +686,41 @@ platform :ios do
...
+      org_slug: 'a8c',
+      project_slug: 'wordpress-ios',
+      version: build_version,
+      dist: build_version,
+      rewrite: true,
+      strip_common_prefix: true,
+      sourcemap: sourcemaps_folder
+    )
+
+    FileUtils.rm_rf(sourcemaps_folder)
```
mokagio: Ruby suggestion: there's a way to run this code that depends on a temporary directory in a block, so that even if something fails, the runtime will ensure the directory is removed.

It should be something like:

```ruby
require 'tmpdir'

Dir.mktmpdir do |sourcemaps_folder|
  # all the code should "just work" the way it's written right now
end
```

fastlane/Fastfile
```diff
@@ -677,6 +686,41 @@ platform :ios do
...
+    # Name of bundle file must be set to main.jsbundle
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js'), File.join(sourcemaps_folder, 'main.jsbundle'))
+    FileUtils.cp(File.join(gutenberg_bundle, 'App.js.map'), File.join(sourcemaps_folder, 'main.jsbundle.map'))
+
+    sentry_upload_sourcemap(
+      auth_token: get_required_env("SENTRY_AUTH_TOKEN"),
+      org_slug: 'a8c',
+      project_slug: 'wordpress-ios',
+      version: build_version,
+      dist: build_version,
```
mokagio: Just to check, what's the difference between these two parameters and why is it okay for them to have the same value?

mokagio (commented):
Thank you for jumping into adding logic to the Fastlane automation @fluiddot 👏 🙇‍♂️ 

I only left nitpick / minor improvement comments

fastlane/Fastfile
fluiddot: Thanks for the suggestion 🙇 ! I'll update the code to use the temporary directory block.

fastlane/Fastfile
fluiddot: They are documented [here](https://docs.sentry.io/platforms/react-native/manual-setup/sourcemaps/#configure-cli), the `version` value should match the one set [here](https://github.com/Automattic/Automattic-Tracks-iOS/blob/aa7040e504377a0322dd91ff366f115e9873a369/Automattic-Tracks-iOS/Crash%20Logging/CrashLogging.swift#L39). However, I'm not sure about what to use for `dist`, when I checked Sentry events, I found that they had the same value (see attached examples) so I went ahead and use the same version value.

Screenshot 2021-07-01 at 12 48 22

I've just checked from where the Sentry iOS SDK takes the `dist` value ([code reference](https://github.com/getsentry/sentry-cocoa/blob/b750f8305717eeb46a11de89b15e2adddc8acc8e/Sources/Sentry/SentryClient.m#L367)) and it's `CFBundleVersion`, so looks like that in the end it's actually the same value 😄 . One thing I could try is to remove the `dist` parameter from the action and check if it affects somehow 👍.

fastlane/Fastfile
fluiddot: I applied your suggestion in [this commit](https://github.com/wordpress-mobile/WordPress-iOS/pull/16700/commits/20f1892b50477fb1a415e0991e7135c5e99a1aaf) and run some tests, it worked like charm 🎊 .

fastlane/Fastfile
fluiddot: I've done a test by removing the `dist` parameter and so far it works, however, this change also implies initializing the SDK in the RN side without that parameter ([code reference](https://github.com/wordpress-mobile/gutenberg-mobile/blob/ca6d40ffcd02a18bec7aab258d76c4147ff6e2a3/src/sentry/index.js#L36)).

To be honest, I don't have a strong opinion on whether include it or not, wdyt?

fluiddot:
As posted in [this comment](https://github.com/WordPress/gutenberg/pull/32768#issuecomment-874051619), the Sentry native module has been moved to the WordPress-iOS project.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergViewController.swift
```diff
@@ -265,8 +265,12 @@ class GutenbergViewController: UIViewController, PostEditor {
         return GutenbergImageLoader(post: post)
     }()
 
+    private lazy var gutenbergSentry: GutenbergSentry = {
+        return GutenbergSentry()
+    }()
```
mokagio: This condensed syntax should work, too

```suggestion
    private lazy var gutenbergSentry = GutenbergSentry()
```

The result is the same, but I though it worth sharing because it's fun to see how the compiler unpacks the condensed syntax I suggested into the full one you used 😄

fastlane/Fastfile
mokagio: @fluiddot I dug a bit into the documents and found a definition of what the `dist` value is meant to be [here](https://docs.sentry.io/platforms/react-native/configuration/releases/) which matches your finding:

> You will also need to pass a distribution to `dist` for events to be attributed correctly to a release. Usually this value is the build number, for example 52. This value needs to be unique to only each release.

Given that, your original approach of passing the build number was spot on.

The snippet in that pages shows the release as something that looks more like a node module name that the build number

```js
Sentry.init({
  release: "my-project-name@2.3.12",
  dist: "52",
});
```

I guess they're making a bit of confusion on their end between `release`, `dist`, `version`, and `build` in the attempt of supporting different platforms? 🤷‍♂️ 

From what I see in your explanation and screenshots, this setup seems fine. 👍

twstokes:

PR #16708, opened 2021-06-18
Blogging reminders: format summary text at end of flow

frosty:
This PR formats the summary text at the end of the reminders flow. It shows a count of the days added along with a formatted list of the days.

| No days selected | One day selected | Multiple days (US) | Multiple days (UK) |
| --- | --- | --- | --- |
| Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 26 37 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 26 27 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 26 47 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 28 01 |

**To test**

### No days selected

- Run through the flow and select no days. Check that the prompt tells you you selected no days.

### One day selected

- Run through the flow and select no days. Check that the prompt tells you you will be reminded **once**, and lists the correct day.

### Multiple days selected (sunday first)

- Set your simulator's region to US
- Run through the flow and select multiple days. Check that the prompt tells you the correct count, and lists the days properly formatted and in the correct order starting with Sunday.

### Multiple days selected (monday first, the correct way 😅 )

- Set your simulator's region to UK
- Run through the flow and select multiple days. Check that the prompt tells you the correct count, and lists the days properly formatted and in the correct order starting with Monday.

## Regression Notes

None

N/A

PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
```diff
@@ -154,6 +163,42 @@ class BloggingRemindersFlowCompletionViewController: UIViewController {
...
+        if selectedDays.count == 1 {
+            promptText = String(format: TextContent.completionPromptSingular, markedUpDays.first ?? "")
+        } else {
+            let formatter = ListFormatter()
+            let formattedDays = formatter.string(from: markedUpDays) ?? ""
+            promptText = String(format: TextContent.completionPromptPlural, "<strong>\(selectedDays.count)</strong>", formattedDays)
+        }
+
+        let style = NSMutableParagraphStyle()
+        style.lineSpacing = 1.5
```
Gio2018: can this number go in the `Metrics` enum?

Gio2018 (changes requested):
Hey @frosty , I had one small code comment, but other than that, the summary font color needs to be fixed in dark mode (see screenshot). Note that this only happens when you have at least one day selected. The "You have no reminders set" text is shown correctly.
<p align=center>

</p>

osullivanchris:
Looks great @frosty . Just one thing. You have mentioned defining S-S or M-S based on UK/US region. But in site settings the user can actually customise this format. Would the logic here respect the customised setting? Maybe US/UK is just the name for that setting I'm not sure

frosty:
Thanks @Gio2018! I've addressed both of these items now, ready for another look.

@osullivanchris Thanks for raising this, I didn't realise that option was there! I think sticking with the device settings is probably the best option for now though, as that's where the user will be receiving notifications. It could be that their site is in a different country or timezone, so the settings may not be right to apply to the user's reminders.

frosty:
Thanks @Gio2018! I've addressed both of these items now, ready for another look.

@osullivanchris Thanks for raising this, I didn't realise that option was there! I think sticking with the device settings is probably the best option for now though, as that's where the user will be receiving notifications. It could be that their site is in a different country or timezone, so the settings may not be right to apply to the user's reminders.

osullivanchris:
> @osullivanchris Thanks for raising this, I didn't realise that option was there! I think sticking with the device settings is probably the best option for now though, as that's where the user will be receiving notifications. It could be that their site is in a different country or timezone, so the settings may not be right to apply to the user's reminders.

Oh cool. Good point too. I hadn't considered that in Weekly Roundup, where I was thinking we'd lean on site settings to find out their week too. But if we can get this info from the device that seems strong, you're right.

WordPress/Classes/Extensions/NSCalendar+Helpers.swift
```diff
@@ -9,4 +9,8 @@ extension Calendar {
         let components = dateComponents([.day], from: fromDate, to: toDate)
         return components.day!
     }
+
+    public func localizedDayIndex(_ index: Int) -> Int {
```
diegoreymendez, line 13: I still think this isn't very reusable - but since this was moved to an extension I think we should document what it does.  It would suffice to have a description so that the reader can quickly figure out if it does what they need.

I tried documenting it to make a suggestion here for you:

```
/// Maps a zero-based index number that represents the ordered day of the week,
/// to a localized index that can be used to access day information from calendar properties
/// such as `shortWeekdaySymbols`.
```

But honestly I'm not sure that's clear / easy to follow.

Gio2018 (approved):

diegoreymendez (commented):
Works perfectly.

PR #16709, opened 2021-06-18
Blogging reminders: Configure Notify Me button based on day selection

frosty:
This PR configures the Notify Me button on the settings screen of the blogging reminders flow, dependent on the user's selection of days.

| No reminders, no selection | No reminders, day selected | Existing reminders |
|---|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 44 23 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 44 22 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-18 at 19 44 06 |

**To test**

### No existing reminders

- Enter the flow with no existing reminders set
- The button should have the title 'notify me' and be disabled
- Try selecting and deselecting days – the button should become enabled if days are selected, and disabled if none are selected
- Complete the flow once you have days selected

### Existing reminders

- Re-enter the flow with days already selected for reminders
- The button should have the title 'update' and be enabled
- Try selecting and deselecting days. Even if you deselect all days, the button should remain enabled, allowing you to clear out your reminders.

## Regression Notes

None

N/A

PR submission checklist:


Gio2018 (commented):
This worked as expected @frosty !

A couple of comments:

1. Not sure if we should change the background of the disabled button in dark mode? (see screenshot). I think it's fine, but if we want to reflect what's in light mode, it should be gray

2. If the user has existing reminders, should the "Update" button be greyed out until they actually make changes?

Pinging @osullivanchris for both these questions

<p align=center>

</p>

osullivanchris:
> Not sure if we should change the background of the disabled button in dark mode? (see screenshot). I think it's fine, but if we want to reflect what's in light mode, it should be gray

This is a standard button that we're using in multiple places so I don't think we should make any change here. If we think the button is not communicating clearly we can raise it elsewhere.

> If the user has existing reminders, should the "Update" button be greyed out until they actually make changes?

If we keep finding more logic to put in this button, eventually I'd be considering just changing it to 'Done' or 'Continue' as I mentioned on slack 😄 I'm not certain on this one. We could. But then if the user changes it back to the original, we'd have to change the label again right? E.g.
- reminder is set to Monday
- I come to change it. Button says e.g. "Keep"
- I add Tuesday. Button changes to "Update".
- I deselect Tuesday. It would return to "Keep".

WDYT @frosty ?

frosty:
@osullivanchris Yeah, it would mean keeping track of whether they'd changed from the current stored value although that shouldn't be too difficult. I'm not 100% sure about the 'keep' terminology – we could just go with Done or Close or something? But at that point, I'm not sure it's worth making the change really... there are other ways to dismiss the view too so it's not entirely necessary.

We could merge as-is for now and always iterate if we want to tweak it?

osullivanchris:
@frosty  yeah I don't really have anything there I think is a brilliant solution, was just thinking out loud/sharing options. How does it work in the current PR? Does it say "Update" before and after they make changes? 

If so I think its fine. No blocker anyway.

...
For the record I just thought of one more option.
- user sets Monday
- closes
- comes back
- button says 'Notify me'
- user adds Tuesday 
- button says 'Update'

still. No obvious standout improvement I can see

Gio2018:
> We could merge as-is for now and always iterate if we want to tweak it?

Sold!

Gio2018 (approved):

PR #16715, opened 2021-06-21
Issue/16710 - Add referrer details scene

nikola-milicevic:
Part of [#16710](https://github.com/wordpress-mobile/WordPress-iOS/issues/16710)
Depends on [#16714](https://github.com/wordpress-mobile/WordPress-iOS/pull/16714)

This PR includes addition of referrer details scene. Dummy data is currently displayed. Real data passing will be implemented in the next PR.

### To test:

1. Tap on 'Stats' button located in 'My Site' tab.
2. Scroll down to referrers section.
3. Tapping on a referrer displays referrer details scene.
4. Tapping on 'View more', and then on a referrer also displays referrer details scene.

### Outcome:

122753674-e1c87100-d292-11eb-906d-18cbe4858b33

## Regression Notes
1. Potential unintended areas of impact
n/a

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #16710, opened 2021-06-18
Stats - Referrers card / details view redesign


nikola-milicevic:
### Goal

Under stats screen, referrers card view and list view (when tapped on 'View more') should be redesigned to use flat lists, instead of collapse / expand functionality. Also, a referrer details screen needs to be added which will display a list of child referrers (if any) and 'Mark / Unmark as spam' button (if applicable).

### Proposed design

![Referrers - List copy](https://user-images.githubusercontent.com/12729242/122752995-e6405a00-d291-11eb-9b73-59fcc93d699e.png)

### Tasks

- [x] Flatten referrers card view and list view: [#16714](https://github.com/wordpress-mobile/WordPress-iOS/pull/16714)
- [x] Add referrer details scene: [#16715](https://github.com/wordpress-mobile/WordPress-iOS/pull/16715)
- [x] Move referrer logic from stats scene to details scene: [#16716](https://github.com/wordpress-mobile/WordPress-iOS/pull/16716)
- [x] Add unit tests: [#16730](https://github.com/wordpress-mobile/WordPress-iOS/pull/16730)

### Fine tuning
- [x] Check iPhone landscape
- [x] Check iPad
- [x] Check dark mode
- [x] Check small devices
- [x] Check RTL languages
- [x] Check Accessibility: [#16727](https://github.com/wordpress-mobile/WordPress-iOS/pull/16727)

PR #16716, opened 2021-06-21
Issue/16710 - Implement logic for referrer details scene

nikola-milicevic:
Part of [#16710](https://github.com/wordpress-mobile/WordPress-iOS/issues/16710)
Depends on [#16715](https://github.com/wordpress-mobile/WordPress-iOS/pull/16715)

This PR adds logic for marking / unmarking referrer as spam.

### To test:

1. Tap on 'Stats' button located in 'My Site' tab.
2. Scroll down to referrers section.
3. Tapping on a referrer displays referrer details scene.
4. Tapping on 'View more', and then on a referrer also displays referrer details scene.
5. Tap on 'Mark as spam' or 'Mark as not spam'

### Outcome:

https://user-images.githubusercontent.com/12729242/122833573-7e1f6180-d2ed-11eb-912a-962b50525507.mp4

### Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
@@ -28,6 +28,7 @@ extension ReferrerDetailsCell {
 // MARK: - Private methods
 private extension ReferrerDetailsCell {
     func setupViews() {
+        selectionStyle = .none
         setupReferrerLabel()
         setupViewsLabel()
         setupSeparatorView()
--- a/WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
@@ -18,9 +18,9 @@ final class ReferrerDetailsCell: UITableViewCell {
 
 // MARK: - Public Methods
 extension ReferrerDetailsCell {
-    func configure(isLast: Bool) {
-        referrerLabel.text = "Test"
-        viewsLabel.text = "123"
+    func configure(data: ReferrerDetailsRow.DetailsData, isLast: Bool) {
+        referrerLabel.text = data.name
+        viewsLabel.text = data.views
         separatorView.isHidden = !isLast
     }
 }
--- a/WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
@@ -29,7 +29,7 @@ extension ReferrerDetailsCell {
 private extension ReferrerDetailsCell {
     func setupViews() {
         selectionStyle = .none
-        backgroundColor = .listForeground
+        backgroundColor = Style.cellBackgroundColor
         setupReferrerLabel()
         setupViewsLabel()
         setupSeparatorView()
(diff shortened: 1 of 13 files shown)


emilylaguna (changes requested):
Heya, nice work. The new design looks good! I noticed a few things when testing:

- When tapping on the <kbd>Mark as Spam</kbd> button I found it a bit odd to then have to tap another <kbd>Mark as Spam</kbd> button the action sheet to trigger the action. What do you think about removing the extra step?
- When testing with slower internet it can feel as though nothing is happening after I choose to mark an item as spam. I wonder if we should display an indicator?
- What do you think about showing the WebView instead of the details view if a user taps on an item that can not be marked as spam and has 1 row to display? Example being: `WordPress.com Reader`
- The app crashes when tapping <kbd>Mark as Spam</kbd> table cell on iPad

nikola-milicevic:
> * When tapping on the Mark as Spam button I found it a bit odd to then have to tap another Mark as Spam button the action sheet to trigger the action. What do you think about removing the extra step?

Yeah, totally agree.

> * When testing with slower internet it can feel as though nothing is happening after I choose to mark an item as spam. I wonder if we should display an indicator?

I was thinking about that, but didn't implement it since I wasn't sure about the design. As discussed over slack, now the action text is replaced with the loader while loading is in progress. I updated the video in the description.

> * What do you think about showing the WebView instead of the details view if a user taps on an item that can not be marked as spam and has 1 row to display? Example being: `WordPress.com Reader`

Yes, definitely feels better.

> * The app crashes when tapping Mark as Spam table cell on iPad

Although these changes are not tested on iPad yet (will be done in future PR), the crash shouldn't happen now since we got rid of action sheet 😊

Thanks for the comments, I updated the code!

WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsViewModel.swift
```diff
@@ -20,6 +21,10 @@ extension ReferrerDetailsViewModel {
     func update(with data: StatsTotalRowData) {
         self.data = data
     }
+
+    func setLoadingState(_ value: Bool) {
```
emilylaguna, line 25: Question: Is there a benefit to exposing just the setter vs making the entire var public?

WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsCell.swift
```diff
@@ -29,6 +29,7 @@ extension ReferrerDetailsCell {
 private extension ReferrerDetailsCell {
     func setupViews() {
         selectionStyle = .none
+        backgroundColor = .listForeground
```
emilylaguna: What do you think about setting this in WPStyleGuide?

emilylaguna (changes requested):
Nice, I think the flow makes more sense now and the loading spinner looks good! 👍 

I left a comment and question for you.

WordPress/Classes/ViewRelated/Stats/Referrer Details/ReferrerDetailsViewModel.swift
nikola-milicevic: I think it's easier to track what's going on that way, it's more explicit. But really I don't have any strong argument against the other approach. Should I leave it as is, or make the property non-private?

emilylaguna: Ah makes sense. I think we can leave it the way it is right now. I don't have a strong opinion either way.

emilylaguna (approved):
LGTM!

PR #16717, opened 2021-06-21
Blogging reminders: Unschedule all on logout

frosty:
This PR unschedules and removes scheduled blogging reminders in a couple of situations:

- If you log out of the default wpcom account, all reminders for all blogs belonging to that account will be removed.
- If you delete a self hosted site from the app, the reminders for that site will be removed.

**To test**

* Build and run, and ensure tests pass
* Log into a wpcom site and add reminders to a site or two. Check that removing reminders works too.
* Add a self hosted site and add some reminders to it
* Log out of the wpcom account and ensure that the self hosted site still has reminders
* Log back into the wpcom account and ensure the reminders didn't persist and were removed

## Regression Notes

None

N/A

PR submission checklist:


WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -78,8 +79,17 @@ class BloggingRemindersStore {
         return configuration[blogIdentifier] ?? .none
     }
 
+    func allScheduledReminders() -> [ScheduledReminders] {
+        return Array(configuration.values)
```
diegoreymendez: In case we need to keep this method:

1. There's no need for the `return` keyword for one-liner methods returning a value.
2. Instead of `Array(Configuration.values)` we could use the less verbose `configuration.values`
3. This could be a calculated var

The method would become:

```swift
var allScheduledReminders: [ScheduledReminders] {
   configuration.values
}
```

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -78,8 +79,17 @@ class BloggingRemindersStore {
         return configuration[blogIdentifier] ?? .none
     }
 
+    func allScheduledReminders() -> [ScheduledReminders] {
```
diegoreymendez: This method seems not to be in use, so we should probably remove it.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -96,6 +106,17 @@ class BloggingRemindersStore {
             throw Error.configurationFileCreationFailed(url: dataFileURL, data: data)
         }
     }
+
+    /// Deletes all current reminders permanently from the system.
+    ///
+    func deleteAllReminders() throws {
```
diegoreymendez: Since we're not using this method we should probably remove it.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersStore.swift
```diff
@@ -14,6 +14,7 @@ class BloggingRemindersStore {
         case configurationDecodingFailed(error: Swift.Error)
         case configurationEncodingFailed(error: Swift.Error)
         case configurationFileCreationFailed(url: URL, data: Data)
+        case removingConfigurationFileFailed(url: URL)
```
diegoreymendez: This can probably be removed, see comments below.

diegoreymendez (commented):
Added some comments on the code in case you want to start looking at it.  I'll be adding feedback from my testing session separately.

WordPress/Classes/Services/SiteManagementService.swift
```diff
@@ -31,6 +31,9 @@ open class SiteManagementService: LocalCoreDataService {
                     let blogService = BlogService(managedObjectContext: self.managedObjectContext)
                     blogService.remove(blog)
 
+                    let scheduler = try? BloggingRemindersScheduler()
```
diegoreymendez: Since the call above to `blogService.remove(blog)` is already taking care of the unscheduling, should we remove this call?

diegoreymendez (changes requested):
Hey @frosty.  This is working mostly fine.

The only issue I've run across is that when disconnecting a Jetpack site from the web, the site is removed from WPiOS when I refresh the list of sites, but the reminders are not removed.

I tested connecting a jurassic.ninja site using Jetpack, and then disconnecting it.

diegoreymendez (approved):

PR #16729, opened 2021-06-23
Issues/16599 follow conversation from post

### LINKED ISSUE #16599: Posts: Add Follow Conversation option to post detail menu.
More background at p1622053441004000-slack-p2-mobile

Now that more posts can be deep linked into the apps, let's make it easier to follow the conversation on those posts, and add this as a menu option on the reader post detail.

aerych:
Closes #16599 

This PR adds a "Follow conversation by email" option to the Reader post menu and is available from post cards and the post detail.  The menu action leverages the existing `FollowCommentsService`, whose code is modified to update the `isSubscribedComments` property on the relevant post.  Use the menu option to follow or unfollow by email the conversation on a post. 

Example:
Simulator Screen Shot - iPhone 11 - 2021-06-23 at 14 46 40
Simulator Screen Shot - iPhone 11 - 2021-06-23 at 14 46 46

To test:
- Follow a conversation from the Reader post card menu, and the Reader post detail menu (they are powered by the same code). 
- Confirm you can follow and unfollow successfully. 
- Confirm the correct notice is shown after following/unfollowing
- Confirm that the correct status is shown on the web. The easiest way was to see if the "Follow/Following" button on a P2 post shows the correct status.
- Change status on the web. Refresh the reader post list in the app.  Confirm the menu shows the correct status. 

Note: I've spotted a glitch with some external sites (e.g. msnbc) incorrectly being flagged as subscribe-able.  This is being addressed API side in a separate PR and should not otherwise impact these changes.  The silver lining is they can be used to test the failure prompts ;).  

@ScoutHarris would you be game for a review?

## Regression Notes
The existing follow conversation feature available on the reader comments screen.

Tested to confirm that there were no regressions.

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderHelpers.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderHelpers.swift
@@ -350,23 +350,15 @@ struct ReaderPostMenuButtonTitles {
     class func dispatchToggleSubscribeCommentMessage(subscribing: Bool, success: Bool) {
         let title: String
         if success {
-            title = subscribing ?
-                NSLocalizedString("Successfully followed conversation", comment: "The app successfully subscribed to the comments for the post") :
-                NSLocalizedString("Successfully unfollowed conversation", comment: "The app successfully unsubscribed from the comments for the post")
-
+            title = subscribing ? NoticeMessages.commentFollowSuccess : NoticeMessages.commentUnfollowSuccess
         } else {
-            title = subscribing ?
-                NSLocalizedString("Unable to follow conversation", comment: "The app failed to subscribe to the comments for the post") :
-                NSLocalizedString("Failed to unfollow conversation", comment: "The app failed to unsubscribe from the comments for the post")
+            title = subscribing ? NoticeMessages.commentFollowFail : NoticeMessages.commentUnfollowFail
         }
         dispatchNotice(Notice(title: title))
     }
 
     class func dispatchToggleSubscribeCommentErrorMessage(subscribing: Bool) {
-        let title = subscribing ?
-            NSLocalizedString("Could not subscribe to comments", comment: "The app failed to subscribe to the comments for the post") :
-            NSLocalizedString("Could not unsubscribe from comments", comment: "The app failed to unsubscribe from the comments for the post")
-
+        let title = subscribing ? NoticeMessages.commentFollowError : NoticeMessages.commentUnfollowError
         dispatchNotice(Notice(title: title))
     }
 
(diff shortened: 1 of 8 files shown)


ScoutHarris:
One issue I noticed. If you toggle Follow from the web, then in-app tap the post card menu, the followed status is not updated. If you tap the post, the post details menu isn't updated either. However, if you view the post's comments, it is updated (the follow option in the header). The new options aren't updated until the post is updated some other way, like pull to refresh.

Example, for a post I followed on the web:

| Post Card | Post Details | Post Comments |
|--------|-------|-------|
| Simulator Screen Shot - iPhone 12 Pro Max - 2021-06-23 at 15 17 36 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-06-23 at 15 17 43 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-06-23 at 15 17 49 |

aerych:
> One issue I noticed. If you toggle Follow from the web, then in-app tap the post card menu, the followed status is not updated. 

The same would be true for liking/unliking a post, unfollowing a followed site, seen/unseen status, etc.  I'm not sure we'd gain much by aggressively re-fetching posts on the off chance that something changed on the web.  At least, it doesn't seem like it's been an issue users have commented about.

> However, if you view the post's comments, it is updated (the follow option in the header). 

Yep. I've left the previous logic used by the comments list in place but I think we could probably just rely on what's in the ReaderPost model rather than having two mechanisms.  I'd rather address that in a separate PR (and after the API patch has landed). 

Thoughts?

ScoutHarris:
>  I'm not sure we'd gain much by aggressively re-fetching posts on the off chance that something changed on the web. At least, it doesn't seem like it's been an issue users have commented about.

Yea, that's what I figured. It seemed edge-casey.

> I think we could probably just rely on what's in the ReaderPost model

Uhhh 🤔 does this mean the comments header would be out of sync as well? I can understand the menus being out of sync, but I don't think the comments header should be.

> I'd rather address that in a separate PR (and after the API patch has landed).

Certainly. We'll be mucking with following via the comments header anyway in Unified Comments. Iff you'd rather leave it til later, we can centralize this logic then.

aerych:
> Uhhh 🤔 does this mean the comments header would be out of sync as well? 

Not necessarily. I don't think it would be an issue but if there was a concern could make the comment list's refresh mechanism more comprehensive (refresh the whole post rather than fetching an individual property) and still use the post as canonical data so things are more consistent over all.  Bonus: it would update any other changes in seen status, likes, etc.

WordPress/Classes/ViewRelated/Reader/ReaderShowMenuAction.swift
```diff
@@ -132,6 +132,24 @@ final class ReaderShowMenuAction {
...
+                                               style: .default,
+                                               handler: { (action: UIAlertAction) in
+                                                if let post: ReaderPost = ReaderActionHelpers.existingObject(for: post.objectID, in: context) {
+                                                    ReaderSubscribeCommentsAction().execute(with: post,
+                                                                                 context: context,
+                                                                                 completion: {
+                                                                                    // No op
+                                                                                 }, failure: { _ in
+                                                                                    // No op.
+                                                                                 })
```
ScoutHarris: Since these blocks are optional, can they just be removed?

WordPress/Classes/ViewRelated/Reader/ReaderHelpers.swift
```diff
@@ -345,6 +347,29 @@ struct ReaderPostMenuButtonTitles {
         dispatchToggleFollowSiteMessage(siteTitle: topic.title, siteID: topic.siteID, following: topic.following, success: success)
     }
 
+    class func dispatchToggleSubscribeCommentMessage(subscribing: Bool, success: Bool) {
+        let title: String
+        if success {
+            title = subscribing ?
+                NSLocalizedString("Successfully followed conversation", comment: "The app successfully subscribed to the comments for the post") :
```
ScoutHarris: Nitpick: There's a `NoticeMessages` struct that contains the other notice messages. Maybe these can be moved into that?

ScoutHarris (approved):
Just a couple nitpicks. Otherwise, LGTM!

:shipit:

WordPress/Classes/ViewRelated/Reader/ReaderShowMenuAction.swift
aerych: Oh good catch. Yeah they totally can.

WordPress/Classes/ViewRelated/Reader/ReaderHelpers.swift
aerych: Done!

ScoutHarris: Thank you!

aerych:
Changes made. One final peek for paranoia sake?

ScoutHarris:
Thanks for the changes! Code looks good, un/following still works. It's a go!

aerych:
Yay! Thank you!!

PR #16733, opened 2021-06-24
Add localization download and upload automation for Jetpack

mokagio:
Adds the tooling required to download the localized metadata for Jetpack and to upload it to App Store Connect.

#### Metadata download

1. Create a new branch from this one
2. Push it to `origin` so that when the release toolkit logic tries to push it won't fail
3. Run `git revert f93a7cfb7b` to ensure you'll have localization changes to fetch
3. Run `bundle exec fastlane download_localized_strings_and_metadata`
4. The lane should succeed and you should see three commits in your branch updating the different localized files

```
* fd94ac7d55 - (HEAD -> test/mokagio-pr-16733) Update Jetpack metadata translations (70 seconds ago)  Gio Lodi
* 99a4f67586 - (origin/test/mokagio-pr-16733) Update metadata translations (85 seconds ago)  Gio Lodi
* 0ee83d8851 - Update translations (2 minutes ago)  Gio Lodi
* 5c21a48c39 - Revert "Update metadata translations" (4 minutes ago)  Gio Lodi
```

[Here](https://github.com/wordpress-mobile/WordPress-iOS/commits/test/mokagio-pr-16733)'s the commit history for my test branch if you want to compare.

#### Metadata upload

Unfortunately, the new lane that uploads metadata to ASC for both WordPress and Jetpack can't be tested end-to-end unless we actually upload the data. If we cancel the `deliver` submission for WordPress, the lane interrupts without giving us a chance to test the Jetpack call. As such, Jetpack needs to be tested in isolation.

1. Checkout this branch
2. Run `bundle exec fastlane update_metadata_on_app_store_connect`
3. Verify the `Preview.html` contains valid data. In particular, that it's not attempting to upload screenshots

Screen Shot 2021-06-24 at 3 27 12 pm

4. **Cancel the upload** 🙏 `Does the Preview on path './fastlane/Preview.html' look okay for you? (y/n) n`
5. Repeat for Jetpack `bundle exec fastlane update_jetpack_metadata_on_app_store_connect`

Screen Shot 2021-06-24 at 3 30 00 pm

_Notice how the description says Jetpack, not WordPress 😉 

6. **Cancel the upload** 🙏 `Does the Preview on path './fastlane/Preview.html' look okay for you? (y/n) n`

Optionally, call one or both lanes with `with_screenshots:true` to see that it will attempt to load the screenshots

Screen Shot 2021-06-24 at 3 32 24 pm

_Obviously, one should only call this lane if they have the screenshots locally_

## Regression Notes
N.A.

N.A.

N.A.

---


fastlane/Fastfile
```diff
@@ -217,7 +217,7 @@ platform :ios do
   desc 'Updates a release branch for a new beta release'
   lane :new_beta_release do |options|
     ios_betabuild_prechecks(options)
-    ios_update_metadata(options)
+    download_localized_strings_and_metadata(options)
```
mokagio, line 220: This is a new lane that does the `.strings` download for WordPress (which Jetpack shares) and the metadata download for WordPress and Jetpack.

I decide to name it like that because I found the release toolkit action name misleading. This name seems more accurate to me. What do you think?

In particular, "download" clearly tells me where the data is coming from, where as "update" could be going both way (e.g. it could be the app updating/uploading metadata to ASC)

fastlane/Fastfile
```diff
@@ -298,6 +298,52 @@ platform :ios do
...
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane update_metadata_on_app_store_connect [with_screenshots:<with screenshots>]
+  #
+  # Example:
+  # bundle exec fastlane update_metadata_on_app_store_connect
+  # bundle exec fastlane update_metadata_on_app_store_connect with_screenshots:true
+  #####################################################################################
+  desc 'Updates the App Store Connect localized metadata'
+  lane :update_metadata_on_app_store_connect do |options|
```
mokagio, line 334: We hinted at having a lane replacing the call to `deliver` in this PR.

The reason I think it's valuable to use a lane is that we could remove the `Deliverfile` and have the `Fastfile` as the single source of truth. What do you think?

I went down that path for the Jetpack logic, but kept the current setup for WordPress – there's enough going on in here already 😄 

I'm happy to revert to use `deliver` if you don't think this is a good approach, maybe passing a dedicated `Jetpack-Deliverfile` to the call. But I'd ask to do it as a followup PR against the next release since this one is getting close to the release finalization date.

fastlane/Jetpack-Fastfile
```diff
@@ -1,3 +1,7 @@
+# frozen_string_literal: true
```
mokagio, line 1: This is just a nice to have, but a no-op for the changes in this PR.

fastlane/Jetpack-Fastfile
```diff
@@ -140,6 +144,67 @@ lane :update_jetpack_appstore_strings do |options|
...
+  # No need to `cd` into `fastlane` because of how Fastlane manages its paths
+  # internally.
+  sh './download_metadata.swift jetpack'
+
+  jetpack_metadata_glob = File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'jetpack_metadata/**/*.txt')
+  sh "git add #{jetpack_metadata_glob}"
+  git_commit(
+    path: jetpack_metadata_glob,
+    message: 'Update Jetpack metadata translations'
+  )
```
mokagio, line 166: I decided to hand-roll the automation to download metadata using the Swift script in its first iteration instead of extracting directly into the release toolkit.

I'm not actually sure what would be the best way to extract it... Maybe move the script to the release toolkit and pass it a JSON/YML config file where to read all the params such as the GlotPress URL that are currently hardcoded?

fastlane/Jetpack-Fastfile
```diff
@@ -140,6 +144,67 @@ lane :update_jetpack_appstore_strings do |options|
...
+lane :update_jetpack_metadata_on_app_store_connect do |options|
+  # Skip screenshots by default. The naming is "with" to make it clear that
+  # callers need to opt-in to adding screenshots. The naming of the deliver
+  # parameter, on the other hand, uses the skip verb.
+  with_screenshots = options.fetch(:with_screenshots, false)
+  skip_screenshots = with_screenshots == false
+
+  deliver(
+    app_identifier: APP_IDENTIFIER,
+    app_version: read_version_from_config,
```
mokagio, line 194: I think we should to this (reading the version from the `.xcconfig`) for all the iOS apps. It would remove the need to modify the `Deliverfile` during the code freeze.

fastlane/Jetpack-Fastfile
```diff
@@ -140,6 +144,67 @@ lane :update_jetpack_appstore_strings do |options|
...
+#####################################################################################
+desc 'Uploads the Jetpack metadata to App Store Connect, localized, and optionally including screenshots.'
+lane :update_jetpack_metadata_on_app_store_connect do |options|
+  # Skip screenshots by default. The naming is "with" to make it clear that
+  # callers need to opt-in to adding screenshots. The naming of the deliver
+  # parameter, on the other hand, uses the skip verb.
+  with_screenshots = options.fetch(:with_screenshots, false)
+  skip_screenshots = with_screenshots == false
+
+  deliver(
```
mokagio, line 192: All the settings below are ports of what's already defined in `Deliverfile`.

fastlane/Jetpack-Fastfile
```diff
@@ -140,6 +144,67 @@ lane :update_jetpack_appstore_strings do |options|
...
+    api_key_path: File.join(PROJECT_ROOT_FOLDER, '.configure-files', 'app_store_connect_fastlane_api_key.json'),
+    metadata_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'jetpack_metadata'),
+    skip_screenshots: skip_screenshots,
+    screenshots_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'jetpack_promo_screenshots'),
+    overwrite_screenshots: true, # if skip_screenshots true, this is ignored
+    skip_binary_upload: true,
+    phased_release: true,
+    precheck_include_in_app_purchases: false,
+    # Reusing the WordPress one here, the values are the same after all
+    app_rating_config_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'metadata', 'ratings_config.json')
```
mokagio, line 204: Makes sense?

fastlane/download_metadata.swift
```diff
@@ -6,7 +6,26 @@ let glotPressSubtitleKey = "app_store_subtitle"
...
+    static let jetpack = Config(
+        baseFolder: "./jetpack_metadata",
+        baseURLString: "https://translate.wordpress.com/projects/jetpack/apps/ios/release-notes/"
+    )
+
+    static func config(for argument: String?) -> Config {
+        guard let argument = argument else { return .wordPress }
+        return argument == "jetpack" ?  .jetpack : .wordPress
+    }
+}
```
mokagio, line 28: Had a bit of Swift fun here. But I worry I ended up with something over-engineered...

fastlane/download_metadata.swift
```diff
@@ -37,9 +56,13 @@ let languages = [
     "zh-Hant": "zh-tw",
 ]
 
-func downloadTranslation(languageCode: String, folderName: String) {
+func downloadTranslation(
+    config: Config = .config(for: CommandLine.arguments.second),
```
mokagio, line 61: There's actually a more idiomatic way of doing this arguments management in the form of [`ArgumentParser`](https://swift.org/blog/argument-parser/) but it seemed out of scope for the context of this PR.

fastlane/download_metadata.swift
```diff
@@ -73,7 +96,7 @@ func downloadTranslation(languageCode: String, folderName: String) {
 
         jsonDict.forEach({ (key: String, value: Any) in
 
-            guard let index = key.index(of: Character(UnicodeScalar(0004))) else {
+            guard let index = key.firstIndex(of: Character(UnicodeScalar(0004))) else {
```
mokagio, line 100: The warning on this had been nagging me for ages!

fastlane/Fastfile
AliSoftware: 💯  + 1000000 on misleading toolkit action names, that has been bothering me for a while and has been a big barrier to entry that made me almost lose my mind the first times I had to work with the toolkit 😅 

In the long term I'm hoping we could rename them all (probably deprecating the existing ones and creating new ones in their place, taking the occasion to modernize and refactor their code), but that's for another day.

fastlane/Fastfile
```diff
@@ -298,6 +298,52 @@ platform :ios do
...
+  #####################################################################################
+  desc 'Downloads localized metadata for App Store Connect from GlotPress'
+  lane :download_localized_strings_and_metadata do |options|
+    # WordPress: leverage the default release toolkit setup.
+    # It's not clear in the action name, but this downloads both `.strings`
+    # translations for the app and App Store Connect metadata.
+    ios_update_metadata(options)
+    # Jetpack: use the custom setup. Notice that we don't need extra steps to
+    # download `.strings` because Jetpack uses the same codebase as WordPress.
+    download_jetpack_localized_app_store_metadata
```
AliSoftware, line 317: For info: this lane is defined in the dedicated and separate `Jetpack-Fastfile`.

fastlane/Fastfile
AliSoftware: I agree it can make sense to have a single source of truth. But:

 - Having that PR ending up doing it one way for Jetpack and another for WordPress actually moves away from that goal, because now we have 3 potential source of truth (`Deliver` + a bit in `Fastfile` too for WP, and `Jetpack-Fastfile` for Jetpack. So personally I would revert the change in that PR to make it only call `deliver` for now – aligning this with the current setup used for WordPress – and work on this on a separate PR
 - That separate PR thwt would work on having a unique source of truth might lead to a debate (e.g. given `Jetpack-Fastfile` is already a separate file, should we put the source of truth in `Fastfile`, making it not obvious when reading the `Jetpack-Fastfile` that some constants come from the main `Fastfile` but are not obivous and visible in the `Jetpack-Fastfile`? Or given we have 2 fastfiles, should we actually consider the `Deliverfile` as a good thing to be used as a **common** source of truth for both?

If we only had one Fastfile I'd agree to get rid of the `Deliverfile` and use the `Fastfile` as the single source of truth, but given we already split it the fastfile there… I think we should brainstorm what's best and keep that debate for a separate PR

fastlane/Jetpack-Fastfile
AliSoftware: See previous comment, it make sense if we agree to split indeed, but it might be too soon to decide how to split and I'd prefer to migrate to the new way for both WordPress and Jetpack all at once, instead of having a mix of new way for Jetpack, old way for WP.

fastlane/Jetpack-Fastfile
AliSoftware: Good point and good idea! 👍

fastlane/Jetpack-Fastfile
```diff
@@ -166,7 +231,18 @@ private_lane :jetpack_appstore_code_signing do |options|
...
+def read_version_from_config
+  fastlane_require 'Xcodeproj'
+  version = ''
+  # If the file is not available, the method will raise so we should be fine
+  # not handling that case. We'll never return an empty string.
+  File.open(File.join(PROJECT_ROOT_FOLDER, 'Config', 'Version.public.xcconfig')) do |config|
+    configuration = Xcodeproj::Config.new(config)
+    version = configuration.attributes['VERSION_SHORT']
+  end
+  return version
```
AliSoftware, line 247: I _think_ that `File.open` is able to return whatever the block we pass to returns?

Which means we could get rid of the intermediate variable and use this instead:
```suggestion
  # If the file is not available, the method will raise so we should be fine
  # not handling that case. We'll never return an empty string.
  File.open(File.join(PROJECT_ROOT_FOLDER, 'Config', 'Version.public.xcconfig')) do |config|
    configuration = Xcodeproj::Config.new(config)
    configuration.attributes['VERSION_SHORT']
  end
```
With the return being implicit in ruby (returning the return value of the last statement, here the `File.open`, which should itself implicitly return the return value of the last statement of its block, here `configuration.attributes['VERSION_SHORT')`)

Worth double-checking it works tho, since I haven't tested this and only mention this from memory.

fastlane/download_metadata.swift
AliSoftware: I like it (but it might be because I miss Swift too and playing with it 😅 )

fastlane/download_metadata.swift
AliSoftware: That would require us to use a `Package.swift` to pull the ArgumentParser package first via SwiftPM to be then able to use it, tho 😓

fastlane/download_metadata.swift
```diff
@@ -133,3 +156,11 @@ func downloadTranslation(languageCode: String, folderName: String) {
...
     downloadTranslation(languageCode: value, folderName: key)
 }
+
+extension Array where Element == String {
+
+    var second: String? {
+        guard count >= 2 else { return .none }
+        return self[1]
+    }
+}
```
AliSoftware, line 167: nit (I too want to have some Swift fun! 😄 )

```suggestion
extension Array {
    var second: Element? { dropFirst().first }
}
```

AliSoftware (commented):
Quick initial review after skimming at code from high level, so you can already start replying/addressing initial feedback; but will require a more deeper review later anyway

mokagio:
@AliSoftware raised some good points regarding the difference in approaches etc. I'm going to close this one in favor of #16764 for the download part and I'll followup with a different PR or RFC issue regarding the `Deliverfile` approach. 👍

mokagio:
_Please don't delete the branch, I want to keep it around for the other PR and for reference till the work is done._

fastlane/Jetpack-Fastfile
mokagio: Thank you for the suggestion @AliSoftware 🙇 

I just tried that for an upcoming PR and it works 👌

mokagio:
Followup: the `Deliverfile` related part of this PR has finally been extracted into https://github.com/wordpress-mobile/WordPress-iOS/pull/16805

PR #16734, opened 2021-06-24
Improves large text support throughout Blogging Reminders.

diegoreymendez:
This PR adds basic accessibility large-text support for Blogging Reminders.  This PR aims to make it possible to interact with the feature, but it's far from perfect and still presents quite a few visual issues we can work on iteratively.

What this PR does:

- It makes sure all of the text in the view controllers that make up the Blogging Reminders flow support dynamic text.  You can try this out by changing it on the fly, and making sure the text in the VCs adapt to the iOS accessibility setting.
- It makes it possible for the user to interact with all of the UI elements, even at the largest font setting.

What this PR doesn't do:

- It does not offer a perfect experience under the largest font size.  Some of the UI still looks ugly.
- It does not address presentation issues with `BottomSheetViewController`, which still needs to be manually dragged up and down when dynamically changing the iOS accessibility font size.

## Screenshots:

Before and after comparisons, to make it easier to understand the changes.

### Intro Screen

I'm aware that the height seems off.  I couldn't figure out a way to fix that yet.

| Before | After |
| --- | --- |
|  |  |
|  |  |

### Day Picker Screen

I'm hiding the tip when showing this with accessibility font sizes.

| Before | After |
| --- | --- |
|   |  |
|  |  |

### Enable Push Notifications Screen

On iPad there's a sizing issue here that I believe it caused by `BottomSheetViewController` refreshing issues.
I'm hiding the hint when showing this with accessibility font sizes.

| Before | After |
| --- | --- |
|  |  |
|  |  |

### All Set! Screen

On iPad there's a sizing issue here that I believe it caused by `BottomSheetViewController` refreshing issues.

| Before | After |
| --- | --- |
|  |  |
|  |  |

## To test:

Try out the changes with iPhone and iPad.
Enable and disable large text to see that the font size and UI adapt.

## Regression Notes

I could've broken UI elements in the regular navigation, but I've manually tested all scenarios and couldn't find any difference with what we had before.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
@@ -229,7 +229,7 @@ class BloggingRemindersFlowCompletionViewController: UIViewController {
             let formattedDays = formatter.string(from: markedUpDays) ?? ""
             text = String(format: TextContent.completionPromptPlural, "<strong>\(selectedDays.count)</strong>", formattedDays)
         }
-        
+
         let htmlData = NSString(string: text).data(using: String.Encoding.unicode.rawValue) ?? Data()
         let options: [NSAttributedString.DocumentReadingOptionKey: Any] = [.documentType: NSAttributedString.DocumentType.html]
 
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
@@ -1,6 +1,5 @@
 import UIKit
 
-
 class BloggingRemindersFlowCompletionViewController: UIViewController {
 
     // MARK: - Subviews
@@ -24,8 +23,10 @@ class BloggingRemindersFlowCompletionViewController: UIViewController {
 
     private let titleLabel: UILabel = {
         let label = UILabel()
+        label.adjustsFontForContentSizeCategory = true
+        label.adjustsFontSizeToFitWidth = true
         label.font = WPStyleGuide.serifFontForTextStyle(.title1, fontWeight: .semibold)
-        label.numberOfLines = 0
+        label.numberOfLines = 2
         label.textAlignment = .center
         label.text = TextContent.completionTitle
         return label
@@ -33,8 +34,10 @@ class BloggingRemindersFlowCompletionViewController: UIViewController {
 
     private let promptLabel: UILabel = {
         let label = UILabel()
+        label.adjustsFontForContentSizeCategory = true
+        label.adjustsFontSizeToFitWidth = true
         label.font = .preferredFont(forTextStyle: .body)
-        label.numberOfLines = 0
+        label.numberOfLines = 4
         label.textAlignment = .center
         label.textColor = .text
         return label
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
```diff
@@ -129,6 +134,16 @@ class BloggingRemindersFlowCompletionViewController: UIViewController {
         calculatePreferredContentSize()
     }
 
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        if traitCollection.preferredContentSizeCategory.isAccessibilityCategory {
```
frosty: Nitpick, we could do `hintLabel.isHidden = traitCollection.preferredContentSizeCategory.isAccessibilityCategory`

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -209,6 +216,16 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
         calculatePreferredContentSize()
     }
 
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        if traitCollection.preferredContentSizeCategory.isAccessibilityCategory {
```
frosty: Same as nitpick above, we could assign directly to `bottomTipPanel.isHidden`

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersPushPromptViewController.swift
```diff
@@ -143,6 +150,16 @@ class BloggingRemindersPushPromptViewController: UIViewController {
         calculatePreferredContentSize()
     }
 
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        if traitCollection.preferredContentSizeCategory.isAccessibilityCategory {
```
frosty: Same as above

frosty (changes requested):
Changes look good, I just spotted a few small issues!

At some font sizes, the icons get compressed:

Screenshot 2021-06-25 at 11 08 01

For singular day selections, we still see the html in the final prompt text:

Screenshot 2021-06-25 at 11 05 42

diegoreymendez:
@frosty - Good catches.  This also made me find another issue where the transition from large to regular in the day picker screen wasn't handled properly for all elements.

This PR is ready for another look.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
```diff
@@ -24,27 +23,33 @@ class BloggingRemindersFlowCompletionViewController: UIViewController {
 
     private let titleLabel: UILabel = {
         let label = UILabel()
+        label.adjustsFontForContentSizeCategory = true
```
frosty, line 26: What do you think about having some convenience factory method on UILabel that sets both of these properties to true?

frosty (approved):
Looks good, just one more comment for your consideration! Up to you though :)

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowCompletionViewController.swift
diegoreymendez: It may make sense, but if we decided to do that, I'd consider doing it separately and also take into consideration standard `minimumScaleFactor` values for our labels.

Right now I'm content with having these values enabled on a label by label basis - less changes to take care of today.

PR #16736, opened 2021-06-24
Editor Onboarding: Enable tooltip for 50% of users

AmandaRiu:
Related WPAndroid PR: https://github.com/wordpress-mobile/WordPress-Android/pull/14897

Enables the editor onboarding tooltip for 50% of users. The process to determine if the user can view the editor onboarding tooltip has been updated to do the following:
1. Get the ID of the user. If the user logged into the app via the site address option for a self-hosted site (i.e - not a WPcom account), then we use an anonymous `userId` generated specifically for this rollout. This `userId` is saved to the `KeyValueDatabase` used by `GutenbergSettings` to guarantee the same value across app launches. 
2. Append "can_view_editor_onboarding" to the `userId` as a seed to keep the calculated hashed result unique to this feature.
3. Hash the `userId+seed` using [djb2hash](http://www.cse.yorku.ca/~oz/hash.html) which conveniently returns an int.
5. Divide the hashed int by 100 to get the _remainder_, which will be a value between 0 and 100. 
6. If the result from step 5 is greater than or equal to 50, then the user can view the onboarding tooltip.

### To test:
The test steps in this section only verify that the onboarding tooltipis displayed properly when the editor is opened for the very first time after a fresh install to ensure nothing was broken in the process of adding the changes included in this PR. 

To actually verify the changes in this PR, add some logging to `GutenbergOnboardingRollout.convertUserIdToRank()` to verify both the `userId` being passed to the method is valid, and the calculated rank.

Repeat the following test steps using a **WPcom account**, and then logging into a **self-hosted site** (non-Jetpacked) **using the site address option**. 

1. Uninstall WordPress from the test device and then install using the build from this PR or branch.
2. Log into the app.
3. Create a new blog post. Verify the **Tap to add content** tooltip appears directly above the inserter button (`+`).
4. Close the editor to return to My Site page. 
5. Tap the button to create another new blog post. Verify the tooltip does not appear. 

## Regression Notes
None that I can think of. This is a temporary measure to rollout this feature to only a percentage of users. Once this feature is fully released, all this code will be removed.

n/a

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
+++ b/WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
@@ -1,5 +1,3 @@
-import CryptoKit
-
 /// This structs helps encapsulate logic related to Gutenberg editor onboarding rollout phases.
 ///
 struct GutenbergOnboardingRollout {
@@ -10,10 +8,19 @@ struct GutenbergOnboardingRollout {
     }
 
     func convertUserIdToRank(userId: String) -> Int {
-        let inputString = userId + "can_view_editor_onboarding"
-        let inputData = Data(inputString.utf8)
-        let hashed = SHA256.hash(data: inputData)
-        let hashRank = abs(hashed.hashValue) % 100
-        return hashRank
+        let key = userId + "can_view_editor_onboarding"
+        let inputString = key.replacingOccurrences(of: "-", with: "")
+        return abs(inputString.djb2hash) % 100
+    }
+}
+
+extension String {
+    // Ref: http://www.cse.yorku.ca/~oz/hash.html
+    // hash(0) = 5381
+    // hash(i) = hash(i - 1) * 33 ^ str[i];
+    var djb2hash: Int {
+        unicodeScalars.map { $0.value }.reduce(5381) {
+            ($0 << 5) &+ $0 &+ Int($1)
+        }
     }
 }
--- a/WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
+++ b/WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
@@ -1,9 +1,19 @@
+import CryptoKit
+
 /// This structs helps encapsulate logic related to Gutenberg editor onboarding rollout phases.
 ///
 struct GutenbergOnboardingRollout {
-    private let phasePercentage = 0
+    private let phasePercentage = 50
 
     func isUserIdInPhaseRolloutPercentage(_ userId: Int) -> Bool {
-        return userId % 100 >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+        return convertUserIdToRank(userId: String(userId)) >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+    }
+
+    func convertUserIdToRank(userId: String) -> Int {
+        let inputString = userId + "can_view_editor_onboarding"
+        let inputData = Data(inputString.utf8)
+        let hashed = SHA256.hash(data: inputData)
+        let hashRank = abs(hashed.hashValue) % 100
+        return hashRank
     }
 }
--- a/WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
+++ b/WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
@@ -3,12 +3,12 @@
 struct GutenbergOnboardingRollout {
     private let phasePercentage = 50
 
-    func isUserIdInPhaseRolloutPercentage(_ userId: Int) -> Bool {
-        return convertUserIdToRank(userId: String(userId)) >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+    func isRolloutIdInPhaseRolloutPercentage(_ uniqueRolloutId: Int) -> Bool {
+        return convertRolloutIdToRank(uniqueRolloutId: String(uniqueRolloutId)) >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
     }
 
-    func convertUserIdToRank(userId: String) -> Int {
-        let key = userId + "can_view_editor_onboarding"
+    func convertRolloutIdToRank(uniqueRolloutId: String) -> Int {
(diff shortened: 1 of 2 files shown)


AmandaRiu:
@dcalhoun This one is ready for  review! 🥳

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
```diff
@@ -171,14 +174,22 @@ class GutenbergSettings {
 
     func canViewEditorOnboarding() -> Bool {
         guard
-            ReachabilityUtils.isInternetReachable(),
-            let account = AccountService(managedObjectContext: context).defaultWordPressComAccount()
+            ReachabilityUtils.isInternetReachable()
         else {
             return false
```
dcalhoun: Does it make sense to leverage the anonymous ID here rather than returning false if there is no internet connection? WDYT?

WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
```diff
@@ -1,9 +1,26 @@
...
 
     func isUserIdInPhaseRolloutPercentage(_ userId: Int) -> Bool {
-        return userId % 100 >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+        return convertUserIdToRank(userId: String(userId)) >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+    }
+
+    func convertUserIdToRank(userId: String) -> Int {
+        let key = userId + "can_view_editor_onboarding"
+        let inputString = key.replacingOccurrences(of: "-", with: "")
+        return abs(inputString.djb2hash) % 100
```
dcalhoun, line 13: I notice the `abs` call here. Is it expected that user IDs themselves may be negative (which is what I encountered and captured in image below) or that the hashed value may be negative? It seemed odd to me that the user ID itself was negative, but I don't know if it matters. 

negative-user-id

dcalhoun (commented):
LGTM. I left a few comments based on the logs I saw while testing. Additionally, I think it is likely wise to request review from someone who has more iOS programming experience, possibly Paul as he has reviewed other PRs for this project. WDYT?

WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
AmandaRiu: I don't think it matters since we only use the absolute value to calculate the rank either way. I saw a lot of negative userIds during testing which is why I use `abs()`

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
AmandaRiu: I understand what you mean, but if I were to do it that way then the next time this block of code runs and the internet is available, the calculated rank will change for this user if they are a WPcom login and that would skew our analytics.

AmandaRiu:
> I think it is likely wise to request review from someone who has more iOS programming experience, possibly Paul as he has reviewed other PRs for this project. WDYT?

@dcalhoun I completely agree. @guarani can you review this PR as well? Many thanks 👍

WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
```diff
@@ -1,9 +1,26 @@
 /// This structs helps encapsulate logic related to Gutenberg editor onboarding rollout phases.
 ///
 struct GutenbergOnboardingRollout {
-    private let phasePercentage = 0
+    private let phasePercentage = 50
 
     func isUserIdInPhaseRolloutPercentage(_ userId: Int) -> Bool {
-        return userId % 100 >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
+        return convertUserIdToRank(userId: String(userId)) >= (100 - phasePercentage) || BuildConfiguration.current ~= [.localDeveloper, .a8cBranchTest]
```
guarani, line 7: nitpick: It might be good to name this something other than `userId` because it could be easily misinterpreted as always being a WP.com user ID, when in fact for non-WP.com users it's a `UUID().hashValue` value.
Maybe, `uniqueUserIdentifier` could also work, just to help distinguish it from WP.com user IDs.

guarani: nitpick: It might be good to name this something other than userId because it could be easily misinterpreted as always being a WP.com user ID, when in fact for non-WP.com users it's a UUID().hashValue value.
Maybe, uniqueUserIdentifier could also work, just to help distinguish it from WP.com user IDs.

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
guarani: Looking at this code, I wonder if it would be advisable to a) to completely remove the internet connection check and b) treat all users, online/offline/WP.com/non-WP.com, the same by just using `UUID().hashValue` instead of their WP.com user ID (in the case of WP.com users).

guarani (commented):
Just left a few comments, @AmandaRiu! Great work with the Swift there! 💯

WordPress/Classes/Utility/Editor/GutenbergSettings.swift
AmandaRiu: Hrm, that makes perfect sense to me! I'll update it 👍

AmandaRiu: @guarani updated in 4f623aa87923b37b99df26300b19064988ceb4d1

WordPress/Classes/Utility/Editor/GutenbergOnboardingRollout.swift
AmandaRiu: Updated in 4f623aa87923b37b99df26300b19064988ceb4d1

AmandaRiu:
@guarani I've removed the check for internet and updated the logic so all users use the generated rollout Id. I also updated the naming conventions used for clarity as well.  Ready for another review! 👍

dcalhoun (commented):
LGTM. I retested both WPCom and direct self-hosted login. I observed expected values for the ID and percentage-based computation. Additionally, the tooltip functioned as expected only showing up for the first editor session in the alpha build.

guarani (approved):
Tested a WP.com site and a non-Jetpack self-hosted site and everything seems to work great!

PR #16743, opened 2021-06-24
Blogging Reminders: Notification copy and action

frosty:
This PR updates the copy of Blogging Reminder notifications and implements the action to be performed when a user taps on them.

reminders-notes-1

### To Test

* To make this easier to test, I updated the lines in BloggingRemindersScheduler.swift that control the timing of the swift (around line 256) to make the next notification appear at the start of the next minute after creating the notifications:

```
        let date = Date()
        let comps = date.dateAndTimeComponents()
        dateComponents.hour = Weekday.defaultHour
        dateComponents.hour = comps.hour 
        dateComponents.minute = comps.minute! + 1
```

* Build and run

#### 1: In-app, same site visible

* Using the reminders card and the code above, schedule a notification for a site on the current day.
* Remain on the My Site screen, and you should see the notification pop in at the correct time.
* Tap it, and you should see the bottom sheet. Tap Blog post, and check the correct site is selected in the editor.

#### 2: In-app, different site visible

* Follow the instructions above, but after scheduling the notification, quickly switch to a different site in My Site, and perhaps even to a different tab
* When the notification comes in, interact with it and ensure the correct site is displayed in My Site and the editor is shown for the correct site.

#### 3: Out of app, same site

* Perform the same test as test 1, but leave the app after scheduling the post.

#### 4: Out of app, different site

* Perform the same test as test 2, but leave the app after scheduling the post

## Regression Notes

This is mostly confined to blogging reminders, however I noticed a bug in `currentlyVisibleBlog` of `WPTabBarController`, where it wasn't working with the new My Site hierarchy.

I checked the code where this was used, and as it's only in a handful of places I tested it manually.

None

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
+++ b/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
@@ -19,8 +19,6 @@ extension InteractiveNotificationsManager: PushNotificationAuthorizer {
 ///
 class BloggingRemindersScheduler {
 
-    static let notificationCategoryIdentifier = "blogging-reminder-weekly"
-
     // MARK: - Convenience Typealiases
 
     typealias BlogIdentifier = BloggingRemindersStore.BlogIdentifier
@@ -243,7 +241,7 @@ class BloggingRemindersScheduler {
             content.title = TextContent.noTitleNotificationTitle
         }
         content.body = TextContent.notificationBody
-        content.categoryIdentifier = BloggingRemindersScheduler.notificationCategoryIdentifier
+        content.categoryIdentifier = InteractiveNotificationsManager.NoteCategoryDefinition.bloggingReminderWeekly.rawValue
         if let blogID = blog.dotComID?.stringValue {
             content.threadIdentifier = blogID
         }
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
```diff
@@ -19,6 +19,8 @@ extension InteractiveNotificationsManager: PushNotificationAuthorizer {
 ///
 class BloggingRemindersScheduler {
 
+    static let notificationCategoryIdentifier = "blogging-reminder-weekly"
```
frosty: I don't love that this is duplicated here and in InteractiveNotificationsManager, but I couldn't easily see a nicer way of doing it. I think we'd have to make all the notification categories there public, which might be okay?

diegoreymendez: Yes, IMHO they should be public, since notification requests can be initiated outside of `InteractiveNotificationsManager`.

diegoreymendez (approved):
Works really well.  Just one comment for your consideration.

Gio2018 (approved):

PR #16746, opened 2021-06-24
Show blogging reminders setup flow after publishing for the first time

### LINKED ISSUE #16745: Show prompt after publishing a post

Gio2018:
Fixes #16745

This PR allows to show the blogging reminders setup flow after publishing a post or a story post for the first time

To test:
Prerequisite: it is recommended to have an account with at least three sites.
Test on iPhone and iPad
- Build/run, and select a site that has never set blogging reminders before
- Create a new post and publish it
- verify that after the editor has been dismissed, the blogging reminders setup flow starts
- Go to another site that had never set blogging reminders (or clear `UserDefaults` or delete and reinstall)
- Create a new story post and publish it
- verify that after the story editor has been dismissed, the blogging reminders setup flow starts
- Go to another site that had never set blogging reminders (or clear `UserDefaults` or delete and reinstall)
- Go to blog posts, drafts
- Create and save a post as draft if you don't already have one
- on that post, tap more, then publish now then publish (alternatively, open the post and then tap on the three dots and Publish then Publish Now)
- Verify that  the blogging reminders setup flow starts after the post has published
- In all the above cases, verify that the publishing notice appears after you dismiss any of the blogging reminders setup screen

## Regression Notes
This PR has some changes to the post publishing actions, but mainly adds optional closures that present the blogging reminders flow, so there should not be any impact on the existing functionalities.

smoke test the publishing flow

none
PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlow.swift
```diff
@@ -1,7 +1,15 @@
...
 
 class BloggingRemindersFlow {
-    static func present(from viewController: UIViewController, for blog: Blog, source: BloggingRemindersTracker.FlowStartSource) {
+    static let weeklyRemindersKeyPrefix = "blogging-reminder-weekly-"
+    static func present(from viewController: UIViewController,
+                        for blog: Blog,
+                        source: BloggingRemindersTracker.FlowStartSource,
+                        alwaysShow: Bool = true) {
+
+        guard alwaysShow || !UserDefaults.standard.bool(forKey: Self.weeklyRemindersKeyPrefix + blog.objectID.uriRepresentation().absoluteString) else {
```
frosty: It might be slightly clearer to split the key declaration onto its own line?

frosty: It would also be good to wrap all the references to UserDefaults and the creation of the key into some helper methods to abstract away the details?

e.g. `hasShownWeeklyRemindersFlow(for: blog)`, `setHasShownWeeklyRemindersFlow(for: blog)`

WordPress/Classes/ViewRelated/Post/EditPostViewController.swift
```diff
@@ -250,9 +250,25 @@ class EditPostViewController: UIViewController {
     }
 
     @objc func closePostPost(animated: Bool) {
+        // this reference is needed in the completion
+        let presentingController = self.presentingViewController
```
frosty, line 254: Is there a reason we can't just do `self.presentingViewController` in the block?

WordPress/Classes/ViewRelated/Post/PostEditor+Publish.swift
```diff
@@ -534,17 +534,38 @@ extension PublishingEditor where Self: UIViewController {
...
 
         if let onClose = onClose {
+            // if this closure exists, the presentation of the Blogging Reminders flow (if needed)
+            // needs to happen in the closure.
             onClose(didSave, false)
         } else if isModal() {
-            presentingViewController?.dismiss(animated: true, completion: nil)
+            if let controller = presentingViewController {
+                controller.dismiss(animated: true) {
+                    if presentBloggingReminders {
```
frosty: Can we extract anything here to avoid the duplication between these two if / else clauses?

frosty (commented):
This is working for me as expected – I left a few code comments :) 

I guess my only concern at this point is that we have four different places that all now know about blogging reminders and presenting the flow, so four places to update any time we want to change something. Do you have any thoughts if there are any ways to decouple this more? 🤔

Gio2018:
> This is working for me as expected – I left a few code comments :)
> 
> I guess my only concern at this point is that we have four different places that all now know about blogging reminders and presenting the flow, so four places to update any time we want to change something. Do you have any thoughts if there are any ways to decouple this more? 🤔

I had the same concern, but I could not find a unique place to handle the post publishing UI actions. There is a good candidate, which is `dismissOrPopView(_:)` in `PublishingEditor`but that method is overridden by an optional closure that, for example, is not nil when the publishing happens from the editor. There is yet another call when the publishing  happens from a post list. There is certainly room for refactoring, but that would be out of the scope of this PR or Project.

WordPress/Classes/ViewRelated/Post/EditPostViewController.swift
Gio2018: Yes. At completion time, that reference would not exist anymore and the presentation would not happen. Another approach would be to use the main window, but I think this is preferable.

diegoreymendez (approved):
Looks good.  Tested on iPhone, iPad.

WordPress/Classes/ViewRelated/Post/EditPostViewController.swift
frosty: Thanks for clarifying :)

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlow.swift
```diff
@@ -25,7 +26,17 @@ class BloggingRemindersFlow {
...
+        setHasShownWeeklyRemindersFlow(for: blog)
+    }
+
+    private static func hasShownWeeklyRemindersFlow(for blog: Blog) -> Bool {
+        let key = Self.weeklyRemindersKeyPrefix + blog.objectID.uriRepresentation().absoluteString
+        return UserDefaults.standard.bool(forKey: key)
+    }
+
+    private static func setHasShownWeeklyRemindersFlow(for blog: Blog) {
+        let key = Self.weeklyRemindersKeyPrefix + blog.objectID.uriRepresentation().absoluteString
```
frosty, line 38: Nitpick, but we could have this key defined in a single static func for `userDefaultsKey(for:)`, but fine to leave it as is for now :)

frosty (approved):
Looks good, thanks for making those changes!

PR #16759, opened 2021-06-25
Removed the newNavigationBarAppearance feature flag.

diegoreymendez:
Removes the new navigation bar feature flag.

## Suggestion for the code review:

For reviewing the code I suggest starting by reviewing the files with the least changes, and marking them as viewed (GitHub review feature).

Once you're done reviewing a single file, you can tap on the "Viewed" checkbox in the top-right corner:

Screenshot 2021-06-25 at 23 01 11

Once you tap it the changes for that file are hidden:

Screenshot 2021-06-25 at 23 02 06

## To test:

For testing I recommend that you run the latest release in your device, and run this PR in your simulator to compare the colors.  Also remember to test in both dark and light modes.

- Make sure the quick action buttons coloring looks right.
   - Tap on them and check the background color change when tapped.
   - Make sure the borders look the same.
- Make sure the application navigation bars have the same background and text colors.
- Make sure the navigation bar fonts look the same.
- Make sure the status bar color is the same.
- Go to My Site > Sharing > Facebook > Connect.  Once the web view loads, make sure the title colors look the same.
- Go to My Site > View Site and make sure the title colors look the same.
- Make sure the Me button border and color looks the same.
- Create a new site and accept the quick start help.  When the list of tasks come up, tap on "View your site".  Make sure the popup is dismissed appropiately and the "View Site" option is highlighted under My Site.
- Make sure that the blog detail header view under My Site (the view containing the blog title, url, icon and quick action buttons) looks the same.
- Go to My Site > Switch Site and make sure it looks the same.  Cell rows should not show a disclosure indicator, and the overall look and coloring should be the same as in our previous release.
- Go through the login flow and make sure the navigation bar coloring looks the same.
- Go to the notifications section and make sure it's still showing the large title.
- Make sure the post editor title looks the same.
- Make sure the reader tab still shows a large title and the coloring is right.
- Make sure the Jetpack app main navigation bar color looks right.

## Regression Notes

The coloring and styling of navigation bars in the App could be affected if any of the cleanups I did was done incorrectly.

I took several steps to minimize the risks:

   - I tried to avoid making improvements to the code and opted for a very minimalistic cleanup in most cases.  We can always make improvements to the code later on.
   - I self-reviewed my PR, ticking each file as "Viewed" once I was sure the changes were correct.  I encourage the reviewers to do the same, and to take all the time needed to make sure the changes make sense.  There's no rush.
   - I'm pushing this change at the beginning of the sprint, so that we have 2 weeks before the code freeze, and another 2 weeks in testing before this is released.
   - I went through all the changes and compiled a list of things to test.  I encourage both PR reviewers and beta testers to use the list I compiled above to ensure there are no regressions.
   - I'm pinging @startuptester to keep you in the loop.

This is a cleanup PR so we're just removing code that is not being used.  The reason why this cleanup is important is because keeping our codebase simple ensures we have something that's easy for us to maintain.

## PR submission checklist:


WordPress/Classes/ViewRelated/System/WPTabBarController.m
```diff
@@ -449,10 +449,6 @@ - (BOOL)tabBarController:(UITabBarController *)tabBarController shouldSelectView
     if (selectedIndex != tabBarController.selectedIndex) {
         switch (selectedIndex) {
             case WPTabMySites: {
-                if (![Feature enabled:FeatureFlagNewNavBarAppearance]) {
```
diegoreymendez, line 456: For context: bypassing the blog list view controller made sense when it was the first VC shown under the "My Site" tab.  With the `newNavBarAppearance` feature flag, the first VC shown under "My Site" is the blog details VC.  The bypassing code thus becomes obsolete.

This code is currently equivalent to:

```swift
if false {
    // We only need to bypass the blog list if we're using the old presentation style
    [self bypassBlogListViewControllerIfNecessary];
}
```

WordPress/Classes/Utility/WebKitViewController.swift
```diff
@@ -88,16 +88,12 @@ class WebKitViewController: UIViewController, WebKitAuthenticatable {
     private var widthConstraint: NSLayoutConstraint?
     private var onClose: (() -> Void)?
 
-    private var useLightStyle: Bool {
-        navigationController is LightNavigationController || FeatureFlag.newNavBarAppearance.enabled
```
diegoreymendez, line 92: This was replaced as being always true because:

`any || true == true`

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -429,10 +429,6 @@ - (void)viewDidAppear:(BOOL)animated
         [self scrollToElement:QuickStartTourElementViewSite];
         self.shouldScrollToViewSite = NO;
     }
-    if (![Feature enabled:FeatureFlagNewNavBarAppearance] &&
```
diegoreymendez, line 432: @Gio2018 - This modification looks odd, but it seems to be logically correct.  Do you know if this is right, or there's a bug in here?

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
```diff
@@ -669,7 +665,7 @@ - (CGFloat)tableView:(UITableView *)tableView heightForFooterInSection:(NSIntege
 - (CGFloat)tableView:(UITableView *)tableView heightForHeaderInSection:(NSInteger)sectionNum {
     BlogDetailsSection *section = self.tableSections[sectionNum];
 
-    if (section.showQuickStartMenu == true || (sectionNum == 0 && [Feature enabled:FeatureFlagNewNavBarAppearance])) {
+    if (section.showQuickStartMenu == true || sectionNum == 0) {
```
diegoreymendez, line 668: This is another change that needs some extra attention IMHO.

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -214,16 +205,7 @@ - (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIVi
...
-        [self updateBarButtons];
-        return;
-    }
-
-    if ([self shouldShowEditButton]){
-        self.navigationItem.leftBarButtonItem = self.editButtonItem;
-    } else {
-        self.navigationItem.leftBarButtonItem = nil;
-    }
+    [self updateBarButtons];
```
aerych: With these changes, would it make sense to replace the two calls to `updateEditButton` with calls to `updateBarButtons` and get rid of this pass-thru method?

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
Gio2018: Yeah I think this call is replaced by[ this one](https://github.com/wordpress-mobile/WordPress-iOS/blob/3185df95cb2d837ca4c0898c04f9eddcb7695f8c/WordPress/Classes/ViewRelated/Blog/My%20Site/MySiteViewController.swift#L76) when `newNavBarAppearance` is enabled

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -961,17 +901,7 @@ - (void)toggleAddSiteButton:(BOOL)enabled
...
-        return;
-    }
-
-    if (![self shouldShowAddSiteButton]) {
-        [self addMeButtonToNavigationBarWithEmail:[[self defaultWordPressComAccount] email] meScenePresenter:self.meScenePresenter];
-        return;
-    }
-
-    self.navigationItem.rightBarButtonItem = self.addSiteButton;
+    [self updateBarButtons];
```
aerych: Looks like `setAddSiteBarButtonItem` is only called from `viewWillAppear` which has a call to `updateEditButton`. Both this and that method are now pass-thru's to call `updateBarButtons`, so the work would be done twice.  Do we need this method after the changes?

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
```diff
@@ -481,24 +463,6 @@ - (void)presentInterfaceForAddingNewSiteFrom:(UIView *)sourceView
...
-{
-    if ([Feature enabled:FeatureFlagNewNavBarAppearance]) {
-        // We should never bypass the list when we're using the new navigation bar
-        return false;
-    }
-
-    return self.dataSource.displayedBlogsCount == 1;
-}
-
-- (void)bypassBlogListViewController
```
aerych, line 494: Shall we remove the method declaration from BlogListViewController.h, and the remaining instance it's called on line 276 in `showNoResultsViewForSiteCount`?

WordPress/Classes/ViewRelated/Blog/Blog Details/BlogDetailsViewController.m
aerych: I'll defer to Gio here but removing it lgtm. 🤷  Afaict we're alreday presenting "What is New" from elsewhere.

aerych (approved):
Hola Diego 👋 
I tested as described and everything looked good compared to the app store version.  I didn't spot any regressions while smoke testing. 
I left a couple of small minor suggestions in the code. :shipit: from me with those changes if you agree. :)

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
diegoreymendez: Thanks, I missed this!

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
```diff
@@ -56,14 +44,9 @@ class MySitesCoordinator: NSObject {
         splitViewController.restorationIdentifier = MySitesCoordinator.splitViewControllerRestorationID
         splitViewController.presentsWithGesture = false
         splitViewController.setInitialPrimaryViewController(navigationController)
-        splitViewController.dimsDetailViewControllerAutomatically = !FeatureFlag.newNavBarAppearance.enabled
+        splitViewController.dimsDetailViewControllerAutomatically = false
```
Gio2018: isn't this property `false` by default? If so, can we remove this line?

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
```diff
@@ -56,14 +44,9 @@ class MySitesCoordinator: NSObject {
...
         splitViewController.tabBarItem = navigationController.tabBarItem
-
-        if Feature.enabled(.newNavBarAppearance) {
-            splitViewController.wpPrimaryColumnWidth = .default
-        } else {
-            splitViewController.wpPrimaryColumnWidth = .narrow
-        }
+        splitViewController.wpPrimaryColumnWidth = .default
```
Gio2018: same here, isn't `.default` the default value? can we remove this line?

Gio2018 (approved):
Only a few code comments, other than that :shipit:!

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: I'll check and remove if it is.

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: Will check, thanks!

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
diegoreymendez: Removed `setAddSiteBarButtonItem()`.

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
diegoreymendez: Removed `updateEditButton()`.

WordPress/Classes/ViewRelated/Blog/Blog List/BlogListViewController.m
diegoreymendez: Removed the declaration of `bypassBlogListViewController()` and the instance where it was being called.

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: Done

WordPress/Classes/ViewRelated/System/Coordinators/MySitesCoordinator.swift
diegoreymendez: Done

PR #16779, opened 2021-06-30
Test PR for Fastlane secrets our of repository

mokagio:
_TODO_

PR #16764, opened 2021-06-28
Add lane to download localized metadata for Jetpack

mokagio:
Adds the tooling required to download the localized metadata.

This is a spin-off of #16733, because in the original PR there are a few open question on code that is unrelated to this feature (which is a nice reminder that even when you think your PR is small, it's always best to keep it focused on doing one thing).

## How to test

1. Checkout this branch
3. Run `bundle exec fastlane download_localized_strings_and_metadata`
4. The lane should succeed and you should see a new commit that deleted all of the `a/fastlane/jetpack_metadata/**/release_notes.txt` file because `download_metadata.swift` it looking for the 17.7 translations, which are not available yet.

You can also change the `glotPressWhatsNewKey` to use 17.6 and run the lane again, you should see some of the `release_notes.txt` changing.

## Regression Notes
1. Potential unintended areas of impact
N.A.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N.A.

3. What automated tests I added (or what prevented me from doing so)
N.A.

---

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes. **N.A.**
- [x] I have considered adding accessibility improvements for my changes. **N.A.**
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. **N.A.**

PR #16780, opened 2021-06-30
Move Fastlane secrets outside of repository

mokagio:
This moves all the secrets used by our Fastlane automation outside of the repo.

I'm targeting `release/17.7` because I have some ASC automation work (see #16764) that I'd like to get in here, and it will need to use the key in the new location. Targeting the release branch and then merging release into `develop` to propagate the changes to the feature devs seems better than opening the PR against `develop` and _remembering_ to update the new ASC-related code from the release branch to use the key at a different location during or after the finalized release merge into `develop`.

## To test

- Checkout this branch
- Run `bundle exec fastlane run configure_apply`
- Run `ls ~/.configure/wordpress-ios/secrets`, you should see

```
app_store_connect_fastlane_api_key.json
google_cloud_keys.json
project.env
```

I opened a [test PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/16779) which runs the lanes that exercise the code touched by this changed. It passed, or rather, [it failed](https://github.com/wordpress-mobile/WordPress-iOS/pull/16779#issuecomment-871131350) as expected (follow link for more details).

If you have time to spare, you can test each locally:

- Run any of the `_code_signing` lanes for WordPress and Jetpack, e.g. `alpha_code_signing` and `jetpack_alpha_code_signing`.
- Run the lanes that uploads WordPress and Jetpack to TestFlight: `build_and_upload_itc` and `build_and_upload_jetpack_for_app_store`.

## Regression Notes
The Developer Portal and ASC automation

See the tests described above

N.A.

---

### DIFF
--- a/.configure
+++ b/.configure
@@ -15,7 +15,7 @@
     },
     {
       "file": "iOS/app_store_connect_fastlane_api_key.json",
-      "destination": ".configure-files/app_store_connect_fastlane_api_key.json",
+      "destination": "~/.configure/wordpress-ios/secrets/app_store_connect_fastlane_api_key.json",
       "encrypt": true
     },
     {
--- a/.configure
+++ b/.configure
@@ -10,7 +10,7 @@
     },
     {
       "file": "iOS/WPiOS/project.env",
-      "destination": ".configure-files/project.env",
+      "destination": "~/.configure/wordpress-ios/secrets/project.env",
       "encrypt": true
     },
     {
--- a/.configure
+++ b/.configure
@@ -5,7 +5,7 @@
   "files_to_copy": [
     {
       "file": "shared/google_cloud_keys.json",
-      "destination": ".configure-files/google_cloud_keys.json",
+      "destination": "~/.configure/wordpress-ios/secrets/google_cloud_keys.json",
       "encrypt": true
     },
     {
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -16,7 +16,7 @@ BUILD_PRODUCTS_PATH = File.join(PROJECT_ROOT_FOLDER, 'Artifacts')
 # Env file paths to load
 ENV_FILE_NAME = '.wpios-env.default'
 USER_ENV_FILE_PATH = File.join(Dir.home, ENV_FILE_NAME)
-PROJECT_ENV_FILE_PATH = File.join(PROJECT_ROOT_FOLDER, '.configure-files', 'project.env')
+PROJECT_ENV_FILE_PATH = File.join(Dir.home, '.configure', 'wordpress-ios', 'secrets', 'project.env')
 
 # Other defines used across multiple lanes
 REPOSITORY_NAME = 'WordPress-iOS'
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -54,10 +54,12 @@ before_all do |lane|
 
   # Check that the env files exist
   unless is_ci || File.file?(USER_ENV_FILE_PATH)
-    UI.user_error!("~/.wpios-env.default not found: Please copy fastlane/env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values")
+    example_path = 'fastlane/env/user.env-example '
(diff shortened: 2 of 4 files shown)


fastlane/Fastfile
```diff
@@ -54,10 +55,12 @@ before_all do |lane|
 
   # Check that the env files exist
   unless is_ci || File.file?(USER_ENV_FILE_PATH)
-    UI.user_error!("~/.wpios-env.default not found: Please copy fastlane/env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values")
+    example_path = 'fastlane/env/user.env-example '
+    msg = "#{ENV_FILE_NAME} not found: Please copy #{example_path} to #{USER_ENV_FILE_PATH} and fill in the values"
+    UI.user_error!(msg)
```
AliSoftware: super-nitpicky-me: not sure it's worth having `msg = …` compared to directly calling `UI.user_error!` with the string literal 😛

.configure
```diff
@@ -5,17 +5,17 @@
...
       "file": "iOS/WPiOS/project.env",
-      "destination": ".configure-files/project.env",
+      "destination": "~/.configure/wordpress-ios/secrets/project.env",
       "encrypt": true
     },
     {
       "file": "iOS/app_store_connect_fastlane_api_key.json",
-      "destination": ".configure-files/app_store_connect_fastlane_api_key.json",
+      "destination": "~/.configure/wordpress-ios/secrets/app_store_connect_fastlane_api_key.json",
       "encrypt": true
```
AliSoftware, line 19: I wonder how we could nudge people into cleaning up the old locations – i.e. running `git clean` or `rm .configure-files/*` – after this migration 🤔 
(Or does `configure_apply` takes care of cleaning up this dir before re-applying the config? Though that means the action would have to know about this path being a legacy default path being used…)

AliSoftware (approved):
Nice 👌  :shipit: 

PS: Do you plan to handle similar migrations to other repos too as part of your work on moving secrets outside the repos? Or were you planning to let each of us handle it (not sure we each have that in our radar if so though)

.configure
mokagio: Good point 🤔 

There are a few files still to move (see #16806) once that's done, `.configure-files` would become a legacy location.

I wonder if we should do something a bit dirty and make `configure` aware of whether it's running inside one of the Apps division projects and take care of running `git clean` or suggesting to do so to the user if it detects that all the files in `.configure` are pointing to the new `~/.configure` home.

I say that's dirty because it adds Automattic/Apps specific information to the tool and that's undesirable from an OSS point of view. Pragmatically speaking, though, that might be an acceptable tradeoff. The tool is open-source but it's so backed into Automattic specific workflows (even just the fact that it's part of the release toolkit instead of being standalone) that I highly doubt anyone other than us is using it right now or would consider doing so. Another factor in favor of doing this is that we're talking about removing secrets for security reasons.

fastlane/Fastfile
mokagio: I think I originally split it to address a line length violation. But now that I checked, I don't get that violation anymore...

Addressed in 2b76a1a04b.

mokagio:
> Do you plan to handle similar migrations to other repos too as part of your work on moving secrets outside the repos? Or were you planning to let each of us handle it (not sure we each have that in our radar if so though)

I plan to take care of all the migrations. It makes sense from a context point of view, I think. Also, the secrets leaking was originally my mess so I feel responsible for cleaning it up 😅 

The release toolkit version of `configure` is already forcing users to decrypt files in locations that are ignored by Git, so there's no pressing urgency in getting this done. That is, we wouldn't get much by running this job in concurrently with multiple developers.

.configure
AliSoftware: What about adding support in the tool for an entry in the `.configure` file specifying paths to clean? That way we wouldn't hardcode in the tool the `.configure-files` path to be cleaned, but make it configurable, and thus avoid the issue in making it too a8c-specific/aware. Best of both worlds, right?

PR #16782, opened 2021-06-30
Try/fix xmlrpc url crash

diegoreymendez:
This PR offers a workaround for a crash we're sometimes seeing when logging out.

The issue is caused by the App being in a bad state, so this is just a patch to prevent this specific crash from happening, but there's a possibility we'll see the crash appear elsewhere (as the underlying issue is truly the bad App state).

## To test:

Unfortunately this can't be tested easily, but the logic change is quite straightforward and should be easy to review.

## Regression Notes

None, as we're adding a simple `nil` check.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Blog.m
+++ b/WordPress/Classes/Models/Blog.m
@@ -210,7 +210,15 @@ NSString * const OptionsKeyIsWPForTeams = @"is_wpforteams_site";
 {
     NSError *error = nil;
     NSRegularExpression *xmlrpc = [NSRegularExpression regularExpressionWithPattern:@"xmlrpc.php$" options:NSRegularExpressionCaseInsensitive error:&error];
-    return [xmlrpc stringByReplacingMatchesInString:self.xmlrpc options:0 range:NSMakeRange(0, [self.xmlrpc length]) withTemplate:path];
+    
+    NSString *xmlrpcURL = self.xmlrpc;
+    
+    // Quick fix for a crash that was being caused by self.xmlrpc being nil.
+    if (xmlrpcURL == nil) {
+        return nil;
+    }
+    
+    return [xmlrpc stringByReplacingMatchesInString:xmlrpcURL options:0 range:NSMakeRange(0, [xmlrpcURL length]) withTemplate:path];
 }
 
 - (NSString *)adminUrlWithPath:(NSString *)path


diegoreymendez:
Fyi @startuptester

WordPress/Classes/Models/Blog.m
```diff
@@ -210,7 +210,15 @@ - (NSString *)urlWithPath:(NSString *)path
...
     NSError *error = nil;
     NSRegularExpression *xmlrpc = [NSRegularExpression regularExpressionWithPattern:@"xmlrpc.php$" options:NSRegularExpressionCaseInsensitive error:&error];
-    return [xmlrpc stringByReplacingMatchesInString:self.xmlrpc options:0 range:NSMakeRange(0, [self.xmlrpc length]) withTemplate:path];
+    
+    NSString *xmlrpcURL = self.xmlrpc;
+    
+    // Quick fix for a crash that was being caused by self.xmlrpc being nil.
+    if (xmlrpcURL == nil) {
+        return nil;
+    }
```
mokagio, line 219: @diegoreymendez nicely done. This will definitely prevent the crash `nil argument` crash 👍 

I wondered what's the implication of returning `nil` from this method. The "Callers" feature in Xcode says this method is used by:

- `Blog adminUrlWithPath:`
- `Blog loginUrl`
- WordPressOrgRestApi.init?(blog:)`

image

I wrote some tests to verify what happens if this method returns `nil` because of a `nil` `xmlrpc` in #16785. The results:

- The admin URL string becomes `(null)path`, where `path` is the given `withPath` parameter
- `loginUrl` is an empty string
- The `WordPressOrgRestApi` instance is `nil`

I don't have enough context to know if the app will deal correctly with those values or not 😞 and I fear if would be hard to follow all the threads to get a complete picture. 

You have much more experience here. What do you think, are the values generated like this safe to propagate through the app?

My uninformed two cents are: Yes, let's ship this crash fix to begin with and see if it has an impact. We know for a fact this is an edge case anyway. We can then followup and little by little introduce better type management and checks to ensure that these edge cases are handled gracefully or, even better, prevented.

diegoreymendez: Honestly, if this crash happens it probably means the App is in a bad state, and we may get another crash elsewhere.

I also tried to check the same thing but it's a bit hard to know.

My 2 cents: if we have a big number of cases with this crash, let's merge it.  Otherwise let's not.

mokagio: Thanks @diegoreymendez! 

Okay... 🤔 I don't feel I have enough context on the code to make a call on whether to merge or not.

So far, I've seen only 2 instances of this crash and other 5 that may or may not be related:

image

You say "if we have a big number of cases with this crash, let's merge it. Otherwise let's not." Going by those numbers, maybe we shouldn't merge then and defer to a "proper" fix of the root cause of having the app in an inconsistent state?

diegoreymendez: Yup, I agree.  I'm closing this for now.

PR #16784, opened 2021-06-30
Fix Blogging Reminders flow when called from Site Settings

Gio2018:
Fixes #NA

When tapping on "Blogging Reminders" in "Site Settings", the prologue screen will only appear the first time the user sees the Blogging Reminders flow. After that, the flow will start with the "Blogging Reminders Settings" screen.

The behavior of the blogging reminders flow after post publishing is unchanged.

To test:
Make sure you have more than one site with no blogging reminders set (or delete and re-install the app)
Test on iPhone and iPad
- Build/run and select a site where you had not previously seen the Blogging Reminders flow
- Go to "Site Settings" then tap "Blogging Reminders" and make sure the prologue screens shows up
- Dismiss the flow (by completing the flow or dismissing any of the screens)
- Tap again on "Blogging Reminders" and verify that this time the prologue is not presented and the flow starts from the settings screen
- Publish a post (from the editor or from the draft list) and verify that the Blogging Reminders flow is not presented after publishing
- Select another site where you had not seen the Blogging Reminders yet
- Publish a post and make sure the Blogging Reminders flow starts, and the post publishing prologue is the first screen

Note: this is targeting `develop` since this is not a user blocker

## Regression Notes
None. This change only configures the starting point of the Blogging Reminders flow

Test that the blogging reminders flow works as expected in all the use cases.

none

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
@@ -198,6 +198,8 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
         configureConstraints()
         populateCalendarDays()
         refreshNextButton()
+
+        navigationController?.setNavigationBarHidden(true, animated: false)
     }
 
     override func viewDidAppear(_ animated: Bool) {
--- a/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
@@ -198,8 +198,6 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
         configureConstraints()
         populateCalendarDays()
         refreshNextButton()
-
-        navigationController?.setNavigationBarHidden(true, animated: false)
     }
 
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
```diff
@@ -198,6 +198,8 @@ class BloggingRemindersFlowSettingsViewController: UIViewController {
         configureConstraints()
         populateCalendarDays()
         refreshNextButton()
+
+        navigationController?.setNavigationBarHidden(true, animated: false)
```
diegoreymendez: Nitpicky: IMHO this is an intrinsic property of our `BloggingRemindersNavigationController` and setting it here isn't great.  My reasoning is that this VC could be shown in the future in another navigation controller that doesn't hide its bar.

My suggestion would be to move this call, and the same call in `BloggingRemindersFlowIntroViewController:106` to somewhere inside `BloggingRemindersNavigationController` (either the `init` method or `viewWillAppear` / `viewDidAppear`).

diegoreymendez (approved):
Just 1 comment, but this works fine and is approved for merging.

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/BloggingRemindersFlowSettingsViewController.swift
Gio2018: Good catch!!

PR #16790, opened 2021-07-01
Fix blogging reminders for the Jetpack App

Gio2018:
Fixes #NA

To test:
test on iPhone and iPad 
- Build/run the Jetpack target on a real device
- Select any site then go to "Site Settings"
- Tap on "Blogging Reminders" and make sure you can go through the flow and set some dates in the date picker
- Build/run the WordPress target on a real device, and repeat the steps above to make sure it still works as expected in the WordPress app

cc @mokagio this PR targets `17.7`

## Regression Notes
There should be none: this only changes the location of a file, using a standard directory

Test the feature on real devices

none
PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
+++ b/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
@@ -112,11 +112,11 @@ class BloggingRemindersScheduler {
     private static var defaultDataFileName = "BloggingReminders.plist"
 
     private static func defaultDataFileURL() throws -> URL {
-        guard let url = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: WPAppGroupName) else {
-            throw Error.cantRetrieveContainerForAppGroup(appGroupName: WPAppGroupName)
-        }
-
-        return url.appendingPathComponent(defaultDataFileName)
+        try FileManager.default.url(for: .documentDirectory,
+                                    in: .userDomainMask,
+                                    appropriateFor: nil,
+                                    create: true)
+            .appendingPathComponent(defaultDataFileName)
     }
 
     // MARK: - Initializers
--- a/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
+++ b/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
@@ -112,7 +112,7 @@ class BloggingRemindersScheduler {
     private static var defaultDataFileName = "BloggingReminders.plist"
 
     private static func defaultDataFileURL() throws -> URL {
-        try FileManager.default.url(for: .documentDirectory,
+        try FileManager.default.url(for: .applicationSupportDirectory,
                                     in: .userDomainMask,
                                     appropriateFor: nil,
                                     create: true)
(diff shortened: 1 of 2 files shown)


WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -23709,7 +23709,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.automattic.jetpack;
 				PRODUCT_MODULE_NAME = WordPress;
 				PRODUCT_NAME = "$(TARGET_NAME)";
-				PROVISIONING_PROFILE_SPECIFIER = "WPiOS Development Profile";
+				PROVISIONING_PROFILE_SPECIFIER = "Jetpack iOS Development";
```
leandroalonso, line 23712: @Gio2018 is this an intended change? Because it seems to be affecting the WordPress `pbxproj`.

Gio2018: Yes it is, it only affects the Jetpack target

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
```diff
@@ -112,11 +112,11 @@ class BloggingRemindersScheduler {
     private static var defaultDataFileName = "BloggingReminders.plist"
 
     private static func defaultDataFileURL() throws -> URL {
-        guard let url = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: WPAppGroupName) else {
-            throw Error.cantRetrieveContainerForAppGroup(appGroupName: WPAppGroupName)
-        }
-
-        return url.appendingPathComponent(defaultDataFileName)
+        try FileManager.default.url(for: .documentDirectory,
```
diegoreymendez: I think we should use the `applicationSupportDirectory` directory instead.

From https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW13:

#### About the "Documents" directory:

> Use this directory to store user-generated content. The contents of this directory can be made available to the user through file sharing; therefore, this directory should only contain files that you may wish to expose to the user.

#### About the `ApplicationSupport` directory:

> Use this directory to store all app data files except those associated with the user’s documents. For example, you might use this directory to store app-created data files, configuration files, templates, or other fixed or modifiable resources that are managed by the app. An app might use this directory to store a modifiable copy of resources contained initially in the app’s bundle. A game might use this directory to store new levels purchased by the user and downloaded from a server.

Edit: I think this requires `Application supports iTunes file sharing` to be ON, but IMHO considering the guidance above the Application Support directory seems like the best fit for additional configuration files, IMHO.

diegoreymendez (approved):
The PR does what it's supposed to do, so I'm approving it.

Added one comment that I think would be good to tackle, but up to you.

WordPress/WordPress.xcodeproj/project.pbxproj
Gio2018: To better clarify this answer, the change here sets the Jetpack iOS Development profile on the Jetpack target, needed to build on a device.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduler.swift
Gio2018: Good point @diegoreymendez thanks for finding this! I updated to `applicationSupportDirectory`. It did not seem to need the iTunes Sharing when I built locally, but let's make sure we test it on the next beta cut, cc @mokagio

mokagio:
@Gio2018 this has been bundled as part of 17.7 beta 3 (17.0.02).

Thanks for your work 🙌

PR #16791, opened 2021-07-01
Issues/fixes likes glitch

aerych:
Refs: p1625015997228100-wpmobile-slack

This patch addresses a reported issue with the likes list and likes train.   To reproduce the issue: 

1. Tap on a like notification for a post
2. Tap the header to view the post
3. Note that the the train of likers is missing from the post.
4. Tap to view the post's comments, then tap the back button.
5. Note that the train of likers is now visible.
6. Tap Back again to return to the list of likers.
7. Note that the rows now only show an "@" character.

### The missing likers train
At first I thought this was going to be an issue with layout not refreshing as it should, but the actual issue was `fetchLikes` was called before the detail had a post.  Moving the call to the `render` method ensures that a post has been assigned to the reader detail and resolves the issue. 

### The @ symbol
This was happening because the original set of likes was being deleted by the purge logic when the second likes list was opened.  We've been deleting all likes on a first load so data was always fresh.  This approach has worked well up til we discovered the particular set of steps leading the the reported glitch. 

There were a couple of ways to address this. For example, the likes list could have checked for another likers list in the navigation stack and only fetched local posts. I didn't like this option as it's better for controllers to not worry about what else is in the nav stack.  The list could have skipped refreshing if there were already recently fetched likes, but this risks stale data so I rejected this approach as well.  I settled on a slight refactor of the service to update, rather than delete any existing likers on the first fetch. I like this approach as it keeps data concerns contained within the service layer, letting the view layer query for what it needs without a care. 

To test:
Attempt to reproduce the error. 
Confirm the likers train appears when expected.
Confirm that the original likes list looks as expected after navigating back from a secondary likes list.
Smoke test the likes list in the reader and other notifications. Confirm you find no regressions.

## Regression Notes
Other parts of the Likes feature

Smoke tested.

N/a
PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/LikeUserHelpers.swift
+++ b/WordPress/Classes/Services/LikeUserHelpers.swift
@@ -32,7 +32,7 @@ import Foundation
 
         let request = LikeUser.fetchRequest() as NSFetchRequest<LikeUser>
         request.predicate = NSPredicate(format: "userID = %@ AND likedSiteID = %@ AND likedPostID = %@ AND likedCommentID = %@",
-                                        argumentArray: [userID, siteID, postID, commentID])
+                                        userID, siteID, postID, commentID)
         return try? context.fetch(request).first
     }
 
--- a/WordPress/Classes/Services/LikeUserHelpers.swift
+++ b/WordPress/Classes/Services/LikeUserHelpers.swift
@@ -5,11 +5,8 @@ import Foundation
 ///
 @objc class LikeUserHelper: NSObject {
 
-    @objc class func createUserFrom(remoteUser: RemoteLikeUser, context: NSManagedObjectContext) {
-
-        guard let likeUser = NSEntityDescription.insertNewObject(forEntityName: "LikeUser", into: context) as? LikeUser else {
-            return
-        }
+    @objc class func createOrUpdateFrom(remoteUser: RemoteLikeUser, context: NSManagedObjectContext) -> LikeUser {
+        let likeUser = likeUser(for: remoteUser, context: context) ?? LikeUser(context: context)
 
         likeUser.userID = remoteUser.userID.int64Value
         likeUser.username = remoteUser.username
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Services/LikeUserHelpers.swift
```diff
@@ -5,33 +5,43 @@ import Foundation
...
-        likeUser.dateFetched = Date()
+    class func likeUser(for remoteUser: RemoteLikeUser, context: NSManagedObjectContext) -> LikeUser? {
+        let userID = remoteUser.userID ?? 0
+        let siteID = remoteUser.likedSiteID ?? 0
+        let postID = remoteUser.likedPostID ?? 0
+        let commentID = remoteUser.likedCommentID ?? 0
+
+        let request = LikeUser.fetchRequest() as NSFetchRequest<LikeUser>
+        request.predicate = NSPredicate(format: "userID = %@ AND likedSiteID = %@ AND likedPostID = %@ AND likedCommentID = %@",
+                                        argumentArray: [userID, siteID, postID, commentID])
```
ScoutHarris: Is `argumentArray` necessary? Can it just be `NSPredicate(format: "userID = %@ AND likedSiteID = %@ AND likedPostID = %@ AND likedCommentID = %@", userID, siteID, postID, commentID)`?

ScoutHarris (approved):
I have one nitpick/question. Otherwise 🥇 !

:shipit:

WordPress/Classes/Services/LikeUserHelpers.swift
aerych: Hmm., seems so.  I went with the `argumentArray` syntax because the other was yielding a warning.  I guess Xcode was just being cranky because it seems fine now. I'll update!

aerych:
Thanks @ScoutHarris !!

mokagio:
@aerych  this has been bundled as part of 17.7 beta 3 (17.0.02).

Thanks for your work 🙌

PR #16801, opened 2021-07-02
Jetpack: fix tapping on notifications

leandroalonso:
The Jetpack app has been presenting some issues regarding notifications:

* It doesn't change to the Notifications tab
* Sometimes notifications appear blank

### To test:

**After a TestFlight is release**

1. Check that when tapping on Notifications they open the Notifications tab with the correctly content

**While we don't have a TestFlight release**

I couldn't find a reliable way to test this using notification due to certificates and etc. The best way to test it is to add this to `WordPressAppDelegate.application(didFinishLaunchingWithOptions:)`:

```swift
        DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
            WPTabBarController.sharedInstance().showNotificationsTab()
        }
```

Run both WordPress and Jetpack and make sure that after 5 seconds the app displays the Notifications tab.

## Regression Notes
WordPress

Explained above

None, this would require UI tests.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/App Configuration/AppConstants.swift
+++ b/WordPress/Classes/Utility/App Configuration/AppConstants.swift
@@ -24,8 +24,8 @@ import WordPressAuthenticator
 // MARK: - Tab bar order
 @objc enum WPTab: Int {
     case mySites
-    case notifications
     case reader
+    case notifications
 }
 
 // MARK: - Localized Strings
--- a/WordPress/Classes/Utility/App Configuration/AppConstants.swift
+++ b/WordPress/Classes/Utility/App Configuration/AppConstants.swift
@@ -21,6 +21,13 @@ import WordPressAuthenticator
     #endif
 }
 
+// MARK: - Tab bar order
+@objc enum WPTab: Int {
+    case mySites
+    case notifications
+    case reader
+}
+
 // MARK: - Localized Strings
 extension AppConstants {
 
--- a/Podfile
+++ b/Podfile
@@ -195,7 +195,7 @@ abstract_target 'Apps' do
 
     # Production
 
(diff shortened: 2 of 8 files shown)


WordPress/Classes/Utility/App Configuration/AppConstants.swift
```diff
@@ -21,6 +21,13 @@ import WordPressAuthenticator
     #endif
 }
 
+// MARK: - Tab bar order
+@objc enum WPTab: Int {
+    case mySites
+    case notifications
+    case reader
```
momo-ozawa: @leandroalonso I think this needs to be
```suggestion
    case reader
    case notifications
```

momo-ozawa (commented):
Works as described for JP! 🎉 

On WP, this PR opens up the Reader tab instead of the Notifications tab. Added a comment below.

WordPress/Classes/Utility/App Configuration/AppConstants.swift
leandroalonso: Oops, someone ended up committing some test code.

momo-ozawa (approved):

leandroalonso:
@startuptester @thehenrybyrd  JFYI: we fixed a bug involving tapping notifications on the Jetpack app — they were not displaying the correct tab.

I had to refactor some WP code, so I'll just update test instructions for this PR in a concise way:

**For both apps**

1. Make sure they are closed
2. When you receive a notification tap on it
3. Make sure the app opens with the notifications tab being displayed

startuptester:
Thanks @leandroalonso, I see the milestone but I want to double check, this refactoring will **not affect** WP 17.7, correct?

leandroalonso:
@startuptester only 17.8.

PR #16805, opened 2021-07-05
Deprecate `Deliverfile` in favor of `Fastfile`

mokagio:
_I decided to go ahead and implement all the code for this proposal in order to discuss it with a clear example available._

_Also notice this PR is against the `move-fastlane-secrets-outside-repo`, which itself is against `release/17.7`. That's because this PR use the App Store Connect API key path and it made sense to me to use it in the "new and proper" mode introduced in `move-fastlane-secrets-outside-repo`. As for the release branch, this automation is for the release process, so merging it here will allow me to test it IRL sooner rather than later._

I'd like to suggest we remove `Deliverfile` as the way to configure the parameters for `deliver` in favor of calling it with a dedicated lane holding all the settings.

Our current approach to updated the metadata on ASC is to call `deliver`, i.e.: `bundle exec fastlane deliver --skip_screenshots`. I don't know whether that was an intentional choice or it's a byproduct of following the Fastlane documentation, but I think that we would be better served by dropping `Deliverfile` and using Fastlane as the single source of truth.

Alongside having a single source of truth, which is always useful in terms of clarity, running the metadata step through a lane makes it consistent with the rest of the steps in the release management process. It also allows us to add additional checks as part of the process in the future, which we cannot do right now by calling `deliver` directly.

The downsides of this approach might be that using `Deliverfile` is, as far as I understand, the standard approach (if only because setting up `deliver` generates a `Deliverfile` for you). Also worth pointing out that, while it's handy to have a single source of truth, our `Fastfile` is pretty long already, so we're liable of information overload and difficulties in navigating the file.

Keen to hear what you think of this approach.

## To test

Unfortunately, the new lane that uploads metadata to ASC for both WordPress and Jetpack can't be tested end-to-end unless we actually upload the data. If we cancel the `deliver` submission for WordPress, the lane interrupts without giving us a chance to test the Jetpack call. As such, Jetpack needs to be tested in isolation.

1. Checkout this branch
2. Run `bundle exec fastlane update_metadata_on_app_store_connect`
3. Verify the `Preview.html` contains valid data. In particular, that it's not attempting to upload screenshots
4. **Cancel the upload** 🙏 `Does the Preview on path './fastlane/Preview.html' look okay for you? (y/n) n`
5. Repeat for Jetpack `bundle exec fastlane update_jetpack_metadata_on_app_store_connect`

_Notice how the description says Jetpack, not WordPress 😉_

6. **Cancel the upload** 🙏 `Does the Preview on path './fastlane/Preview.html' look okay for you? (y/n) n`

Optionally, call one or both lanes with `with_screenshots:true` to see that it will attempt to load the screenshots. Notice that doing so works only if you have the screenshots locally.

## Regression Notes
The metadata upload process, once the user approves and they reach App Store Connect.

See the steps above for checking as much as possible locally. If anything else is broken I, as the release train conductor, should be able to address it as part of the release finalization.

N.A.

---


fastlane/Fastfile
```diff
@@ -18,23 +18,26 @@ ENV_FILE_NAME = '.wpios-env.default'
...
+  APP_STORE_VERSION_BUNDLE_IDENTIFIER,
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressShare",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressDraftAction",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressTodayWidget",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressStatsWidgets",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressNotificationServiceExtension",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressNotificationContentExtension",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressAllTimeWidget",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressThisWeekWidget",
+  "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.WordPressIntents"
```
mokagio: This might be considered excessive DRY, but I figured that, once extracted in a constant for later reuse to upload the metadata to ASC, it made sense to take inspiration from the [Zero one infinity rule](https://en.wikipedia.org/wiki/Zero_one_infinity_rule).

fastlane/Fastfile
```diff
@@ -839,3 +878,26 @@ platform :ios do
...
+  File.open(File.join(PROJECT_ROOT_FOLDER, 'Config', 'Version.public.xcconfig')) do |config|
+    configuration = Xcodeproj::Config.new(config)
+    configuration.attributes['VERSION_SHORT']
+  end
+end
+
+# Application-agnostic settings for the `upload_to_app_store` (also known as
+# `deliver`) action.
+UPLOAD_TO_APP_STORE_SETTINGS = {
+  app_version: read_version_from_config,
```
mokagio: Notice that this has to be defined _after_ `read_version_from_config` otherwise it won't be able to access it.

fastlane/Jetpack-Fastfile
```diff
@@ -94,7 +94,7 @@ APP_IDENTIFIER = 'com.automattic.jetpack'
     testflight(
       skip_waiting_for_build_processing: true,
       team_id: "299112",
-      api_key_path: File.join(SECRETS_DIR, 'app_store_connect_fastlane_api_key.json')
+      api_key_path: APP_STORE_CONNECT_KEY_PATH
```
mokagio, line 97: Extra benefit of extracting `APP_STORE_CONNECT_KEY_PATH` is that we don't need to duplicate the logic between application `Fastfile`s.

fastlane/Jetpack-Fastfile
```diff
@@ -166,6 +166,41 @@ lane :download_jetpack_localized_app_store_metadata do
...
+  with_screenshots = options.fetch(:with_screenshots, false)
+  skip_screenshots = with_screenshots == false
+
+  upload_to_app_store(
+    # The reason we have these constant settings is to be able to able to
+    # share them with the Jetpack implementation (see below).
+    UPLOAD_TO_APP_STORE_SETTINGS.merge(
+      app_identifier: APP_IDENTIFIER,
+      metadata_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'jetpack_metadata'),
+      screenshots_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'jetpack_promo_screenshots'),
```
mokagio: Notice that these parameters are different from the WordPress one 👍

AliSoftware:
Haven't got a chance to review the code yet (plan to look at it later today, but wanted to give you a chance to see my reply before your EOD) but given that:
 - I totally agree with using a dedicated lane rather than `fastlane deliver` in the command line to update metadata, to be consistent with other steps in release process and allow us to add checks and all
 - But the drawback of this is to have a `Fastfile` that grows too much in size, and also that we already have a `JetpackFastfile` and `ScreenshotFastfile` split in some repos already, which makes this "single source of truth and all things in a single place" idea kinda fall apart

then what do you think about an alternative split of:
 - Dropping the `Deliverfile` like you suggest
 - But splitting the `Fastfile` into subfiles that are a bit better organized that what we currently have? Since we already split screenshot-related lanes in `ScreenshotFastfile` in some repos, and Jetpack-related into `JetpackFastfile`, I'm thinking of having wordpress-specific lanes in a `WordpressFastfile` for example, and then have the main `Fastfile` be reduced to only contain (1) constants common to all products, (2) helper methods useful for all products, (3) then finally `fastlane_require` the other fastfiles.

At least I don't see why the main `Fastfile` should contain WordPress-specific stuff if we're gonna have a `JetpackFastfile` for Jetpack, as opposed to extract both WP and JP in their separate, dedicated `*Fastfile`. Otherwise it feels a bit unbalanced to me 🤷 😅

fastlane/Fastfile
AliSoftware: nit (my OCD and internal-excessive-DRYer speaking): what do you think about using a `map` to apply the same prefix to each extension and DRYing even more?

```ruby
ALL_BUNDLE_IDENTIFIERS = [APP_STORE_VERSION_BUNDLE_IDENTIFIER] + %w[
  WordPressShare
  WordPressDraftAction
  WordPressTodayWidget
  WordPressStatsWidgets
  WordPressNotificationServiceExtension
  WordPressNotificationContentExtension
  WordPressAllTimeWidget
  WordPressThisWeekWidget
  WordPressIntents
].map { |ext| "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.#{ext}" }
```

fastlane/Fastfile
```diff
@@ -839,3 +878,26 @@ platform :ios do
...
+def read_version_from_config
+  fastlane_require 'Xcodeproj'
+
+  # If the file is not available, the method will raise so we should be fine
+  # not handling that case. We'll never return an empty string.
+  File.open(File.join(PROJECT_ROOT_FOLDER, 'Config', 'Version.public.xcconfig')) do |config|
+    configuration = Xcodeproj::Config.new(config)
+    configuration.attributes['VERSION_SHORT']
+  end
+end
```
AliSoftware, line 910: 🤔 Don't we already have an action in our release toolkit to read this? (even if I'm not sure we're using `Xcodeproj::Config` to read it, while we definitively should)

If not (I see we have actions to read the `VERSION_LONG` but indeed maybe not for `VERSION_SHORT` …) we should definitively migrate this code to the toolkit anyway (†)

_(†) side note that my take is we should avoid having a single action for every time of version read we want to expose, like one action for version_short and one for version_long, one for public and one for internal, one for release and one for beta, etc… but instead have a unified `ios_get_version(…)` action taking parameters to tell which of public vs internal, short vs long, etc we want. But that's probably another debate for another day?_

fastlane/Jetpack-Fastfile
```diff
@@ -166,6 +166,41 @@ lane :download_jetpack_localized_app_store_metadata do
...
+lane :update_jetpack_metadata_on_app_store_connect do |options|
+  # Skip screenshots by default. The naming is "with" to make it clear that
+  # callers need to opt-in to adding screenshots. The naming of the deliver
+  # parameter, on the other hand, uses the skip verb.
+  with_screenshots = options.fetch(:with_screenshots, false)
+  skip_screenshots = with_screenshots == false
+
+  upload_to_app_store(
+    # The reason we have these constant settings is to be able to able to
+    # share them with the Jetpack implementation (see below).
```
AliSoftware: Unsure I understand this comment given where it is made: aren't we _already_ in the Jetpack implementation here? Was this a comment you wrote back when these lines were still inside the main `Fastfile` while you were working on this, and forgot to update the comment when you moved this comment here?

Also:
```suggestion
    # The reason we have these constant settings is to be able to
```

fastlane/Fastfile
```diff
@@ -283,6 +286,42 @@ platform :ios do
...
+    # Skip screenshots by default. The naming is "with" to make it clear that
+    # callers need to opt-in to adding screenshots. The naming of the deliver
+    # parameter, on the other hand, uses the skip verb.
+    with_screenshots = options.fetch(:with_screenshots, false)
+    skip_screenshots = with_screenshots == false
+
+    # WordPress
+    upload_to_app_store(
+      # The reason we have these constant settings is to be able to able to
+      # share them with the Jetpack implementation (see below).
```
AliSoftware: Ok I get where the copy/pasted comment in `JetpackFastfile` was coming from now 😛 

The comment and its location was confusing to be tbh. I'd suggest instead to:
 - name the constant `COMMON_UPLOAD_TO_APPSTORE_PARAMS`
 - move the comment next to that constant's declaration site, rather than here, and making it something like "Common settings for upload_to_app_store calls shared between WordPress and Jetpack"

fastlane/Fastfile
```diff
@@ -283,6 +286,42 @@ platform :ios do
...
+      UPLOAD_TO_APP_STORE_SETTINGS.merge(
+        app_identifier: APP_STORE_VERSION_BUNDLE_IDENTIFIER,
+        screenshots_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'promo-screenshots'),
+        skip_screenshots: skip_screenshots
+      )
```
AliSoftware: A more idiomatic ruby way to transform a `Hash` into a list of arguments is to use ruby's splat operator (i.e. `*a` to transform an Array into a list of unnamed params and vice-versa, `**h` to do transform a Hash into a list of named params and vice-versa).

Which should look like this:
```ruby
upload_to_app_store(
  **COMMON_UPLOAD_TO_APPSTORE_PARAMS,
  app_identifier: APP_STORE_VERSION_BUNDLE_IDENTIFIER,
  screenshots_path: File.join(PROJECT_ROOT_FOLDER, 'fastlane', 'promo-screenshots'),
  skip_screenshots: skip_screenshots
)
```

fastlane/Fastfile
```diff
@@ -283,6 +286,42 @@ platform :ios do
...
+    )
+
+    # Jetpack: Delegate to dedicated lane from Jetpack-Fastfile
+    update_jetpack_metadata_on_app_store_connect(options)
```
AliSoftware: This is triggering my OCD a bit, because we have a common lane (which acts on both WP and JP) in the common `Fastfile`, but JP-specific things separate. Why should Jetpack be separate (good idea)… but not WP (which lives in the main Fastfile which is used both for WP-only and for common lanes)…

My suggestion/idea:
 - Move the main body of this lane except those 2 jetpack lines into a `WordPressFastfile` file, into a lane called `update_wordpress_metadata_on_app_store_connect` – to mirror how you did it to JP for WP 
 - Then make this lane in the main `Fastfile` just call `update_wordpress_metadata_on_app_store_connect` then `update_jetpack_metadata_on_app_store_connect(options)`

Effectively making the `Fastfile` be the entry point that only contains lanes that act on _both_ apps at once, and keeping the code specific to WP in `WordPressFastfile` similar to how you kept the code specific to JP in `JetpackFastfile`

Or am I being too nit-picky and OCD'd here?

fastlane/Fastfile
AliSoftware: One more reason to migrate that method to the toolkit 🙃

fastlane/Fastfile
mokagio: 🤩 I like this! It didn't occur to me to do something like that.

fastlane/Fastfile
mokagio: Fair. Thanks for bringing this up.

fastlane/Fastfile
mokagio: :TIL: 🤩 😮

fastlane/Fastfile
mokagio: I think your suggestion makes total sense @AliSoftware, it's not over the top at all.

I, too, felt it was a bit dissonant to have a mix of direct `deliver` call and delegation to an app specific lane.

The point I was coming from was that the Jetpack logic should be extracted out of the repo ASAP, so it made sense to me to leave `Fastfile` focused on WordPress and calling out to `JetpackFastfile` for the Jetpack-specific logic.

I'm probably "deluding" myself on the fact that we'll move Jetpack to its own repo soon, so it does make sense to spend a bit of time in separating the automation code. That could also be a good occasion to identify further opportunities for abstracting logic to the release toolkit level.

fastlane/Jetpack-Fastfile
mokagio: Ah! Good catch.

mokagio:
@AliSoftware thank you so much for your thoughtful suggestions. 

I explained where I was coming from [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/16805#discussion_r664235038) but overall I agree with your point of view and am excited to get to work on tidying things up like you propose (even just as a way to fix the indentations of the lanes and the comments 🤣 )

I should have addressed all the low hanging fruits from your review, including making the lane consistent by having a main lane and two app specific ones (79e9142).

I think the suggestion to split the Fastfile and to move the code to read the version into the release toolkit deserve their own PR, though. I'd like to tackle each in isolation.

WDYT?

fastlane/Fastfile
AliSoftware: Over-achiever idea:
 - Agree on a convention that every WP/JP-specific lane, which would go in their respective `*Fastfile`, follow the naming convention of starting with `wordpress_`/`_jetpack` respectively
- Then in the main `Fastfile`, use Ruby magic to generate the top level lanes which would call both WP & JP specific corresponding lane in turn) in a DRY way

I'm thinking something like:

```ruby
# not sure about the list here, didn't check the exact cases, it's just to give the gist of it
lanes_acting_on_both_wp_jp = %i[
  upload_to_playstore
  download_metadata
  build_and_upload_to_app_store
  …
]
# auto-declare lanes dynamically, making them call both wp and jp lanes in turn
lanes_acting_on_both_wp_jp.each do |lane_name|
  desc "Call wordpress_#{lane_name} then jetpack_#{lane_name}"
  lane lane_name do |options|
    # I don't actually know the syntax to call another name given its name as a string/symbol, so strawman syntax here,
    # we'll need to search the fastlane doc on how to do this but I'm sure there's a way
  
    Shared::Lanes["wordpress_#{lane_name}"].call(options)
    Shared::Lanes["jetpack_#{lane_name}"].call(options)
  end
end
```
(Concept code, not tested)

Again, this is me being over-the-top and all, not necessary to go that far, but just thought it was an interesting idea/implementation possibility to think about 🙃

AliSoftware:
> I think the suggestion to split the Fastfile and to move the code to read the version into the release toolkit deserve their own PR, though. I'd like to tackle each in isolation.

Fair enough 👍

fastlane/Fastfile
AliSoftware: Actually just `send` (which is a ruby core method to call other methods using a Symbol, a bit like good ol' `[NSObject performSelector:]` 😄 ) seems to be enough and work:

Tested with some dummy lanes:
```ruby
  # The kind of thing we'd put in Fastfile
  %i[foo bar].each do |lane_name|
    lane lane_name do |options|
      send("wp_#{lane_name}".to_sym, options)
      send("jp_#{lane_name}".to_sym, options)
    end
  end

  # The kind of lane we'd put in WordpressFastfile
  lane :wp_foo do |options|
    UI.message("wp_foo #{options.inspect}")
  end

  lane :wp_bar do |options|
    UI.message("wp_bar #{options.inspect}")
  end

  # The kind of lane we'd put in JetpackFastfile
  lane :jp_foo do |options|
    UI.message("jp_foo #{options.inspect}")
  end

  lane :jp_bar do |options|
    UI.message("jp_bar #{options.inspect}")
  end
```
Then testing with `fl foo x:1 y:hey`:
```
[11:51:38]: -----------------------------------
[11:51:38]: --- Step: Switch to wp_foo lane ---
[11:51:38]: -----------------------------------
[11:51:38]: Cruising over to lane 'wp_foo' 🚖
[11:51:38]: wp_foo {:x=>"1", :y=>"hey"}
[11:51:38]: Cruising back to lane 'foo' 🚘
[11:51:38]: -----------------------------------
[11:51:38]: --- Step: Switch to jp_foo lane ---
[11:51:38]: -----------------------------------
[11:51:38]: Cruising over to lane 'jp_foo' 🚖
[11:51:38]: jp_foo {:x=>"1", :y=>"hey"}
[11:51:38]: Cruising back to lane 'foo' 🚘
```

fastlane/Fastfile
```diff
@@ -18,24 +18,27 @@ ENV_FILE_NAME = '.wpios-env.default'
...
+  WordPressTodayWidget
+  WordPressStatsWidgets
+  WordPressNotificationServiceExtension
+  WordPressNotificationContentExtension
+  WordPressAllTimeWidget
+  WordPressThisWeekWidget
+  WordPressIntents
+].map { |suffix| "#{APP_STORE_VERSION_BUNDLE_IDENTIFIER}.#{suffix}" }
+
+ALL_BUNDLE_IDENTIFIERS = [APP_STORE_VERSION_BUNDLE_IDENTIFIER] + EXTENSIONS_BUNDLE_IDENTIFIERS
```
AliSoftware: btw, just FYI and to improve your ruby knowledge about the splat operator that I talked about in the other comment, know that you could have used the splat operator in a similar way here to transform the `EXTENSIONS_BUNDLE_IDENTIFIERS` array into a list of values… and put that at the end of the array like if you inserted that list literally. i.e.:

`ALL_BUNDLE_IDENTIFIERS = [APP_STORE_VERSION_BUNDLE_IDENTIFIER, *EXTENSIONS_BUNDLE_IDENTIFIERS]`

Not saying that you should do that – I think both ways of assembling the final array are totally fine 👍  – but just thought you'd be interested and curious about this to improve your ruby knowledge 🙂

fastlane/Fastfile
```diff
@@ -283,6 +286,56 @@ platform :ios do
...
+  # bundle exec fastlane update_wordpress_metadata_on_app_store_connect
+  # bundle exec fastlane update_wordpress_metadata_on_app_store_connect with_screenshots:true
+  #####################################################################################
+  desc 'Uploads the WordPress metadata to App Store Connect, localized, and optionally including screenshots.'
+  lane :update_wordpress_metadata_on_app_store_connect do |options|
+    # Skip screenshots by default. The naming is "with" to make it clear that
+    # callers need to opt-in to adding screenshots. The naming of the deliver
+    # parameter, on the other hand, uses the skip verb.
+    with_screenshots = options.fetch(:with_screenshots, false)
+    skip_screenshots = with_screenshots == false
```
AliSoftware, line 329: 👍  I had a doubt while reading this that the option would be interpreted as a String and that passing `with_screenshots:false` for example would make `with_screenshots` variable be `"false"` String and that the comparison would end up with `skip_screenshots = false`…
But, turns out, fastlane is smart and interprets the option as a boolean… and even better, it accepts other spelling of the truthy/falsy values, e.g calling the. lane with `with_screenshots:yes` also results in `skip_screenshots = false` and `with_screenshots:no` results in `skip_screenshots = true`, so 👌 🎉 

(Only nit I have is that I'd have written it `skip_screenshots = !with_screenshots`, but 🤷 )

fastlane/Fastfile
```diff
@@ -283,6 +286,56 @@ platform :ios do
...
+
+    upload_to_app_store(
+      **COMMON_UPLOAD_TO_APPSTORE_PARAMS,
```
AliSoftware: 💭   I know that I'm the one who suggested this constant name 😅  but now that I read it I'm thinking maybe it should be spelled `UPLOAD_TO_APPSTORE_COMMON_PARAMS`… 🤷

fastlane/Fastfile
```diff
@@ -839,3 +892,26 @@ platform :ios do
     UI.message("Gutenberg version: #{(res.scan(/'([^']*)'/))[0][0]}")
   end
 end
+
+def read_version_from_config
```
AliSoftware, line 901: 💡  If you find it useful, it could be nice to add a fixme comment until you tackle this in separate PR (except if you plan to do it like super soon)
```suggestion
# FIXME: Move to release-toolkit
def read_version_from_config
```

fastlane/Fastfile
```diff
@@ -839,3 +892,26 @@ platform :ios do
...
+  # If the file is not available, the method will raise so we should be fine
+  # not handling that case. We'll never return an empty string.
+  File.open(File.join(PROJECT_ROOT_FOLDER, 'Config', 'Version.public.xcconfig')) do |config|
+    configuration = Xcodeproj::Config.new(config)
+    configuration.attributes['VERSION_SHORT']
+  end
+end
+
+# Application-agnostic settings for the `upload_to_app_store` (also known as
+# `deliver`) action.
```
AliSoftware: I should really stop being nitpicky 😅  but I know you like it and have similar OCD as me so I know you'll indulged me 😸 
```suggestion
# Application-agnostic settings for the `upload_to_app_store`
# (also known as `deliver`) action.
```

fastlane/Deliverfile
```diff
@@ -1,17 +0,0 @@
...
-app_version "17.7"
-
-skip_binary_upload true
-overwrite_screenshots true
-phased_release true
-precheck_include_in_app_purchases false
-
-api_key_path ".configure-files/app_store_connect_fastlane_api_key.json"
-
-app_rating_config_path "./fastlane/metadata/ratings_config.json"
```
AliSoftware, line 17: 👍  Just to say I checked that each setting that was in this deleted `Deliverfile` was now mirrored back in the `Fastfile`, either in the [common settings constant](https://github.com/wordpress-mobile/WordPress-iOS/pull/16805/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R909-R917) or for each app-specific call, and can confirm that all of them were there, so I think we're indeed safe to delete this now 🎉

AliSoftware (approved):

fastlane/Fastfile
mokagio: Thank you for that. I wasn't happy with the look of it or with the unnecessary creation of an Array just for `APP_STORE_VERSION_BUNDLE_IDENTIFIER`.

I'm disappointed it didn't occur to me to try and use the splat which you already mentioned 🙃

I'll definitely adopt this as I keep working in this area

fastlane/Fastfile
mokagio: Ahahah! Yeah, I can see how that might be visually bothering.

I format long lines using my text editor and it would restore it to that form next time I asked it to format the line, so I decided to changed the wording to remove the parenthesis in 8ed1baa0cd 🤷‍♂️

fastlane/Fastfile
```diff
@@ -748,7 +801,7 @@ platform :ios do
       type: 'enterprise',
       team_id: get_required_env('INT_EXPORT_TEAM_ID'),
       readonly: true,
-      app_identifier: ALL_BUNDLE_IDENTIFIERS.map { |id| id.sub('org.wordpress', 'org.wordpress.alpha') }
+      app_identifier: ALL_BUNDLE_IDENTIFIERS.map { |id| id.sub(APP_STORE_VERSION_BUNDLE_IDENTIFIER, 'org.wordpress.alpha') }
```
AliSoftware, line 804: @mokagio After a good night sleep I woke up this morning with the realization that it would have been a better idea to maybe have `ALL_BUNDLE_IDENTIFIERS` instead be a function dynamically generating the list of identifiers from a base one, and having that base one be the `APP_STORE_VERSION_BUNDLE_IDENTIFIER` by default.

e.g change [those lines 29-41 above](https://github.com/wordpress-mobile/WordPress-iOS/pull/16805/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R29-R41) to something like
```ruby
def all_bundle_identifiers(root: APP_STORE_VERSION_BUNDLE_IDENTIFIER)
  extensions_bundle_ids = %w[
    WordPressShare
    WordPressDraftAction
    WordPressTodayWidget
    WordPressStatsWidgets
    WordPressNotificationServiceExtension
    WordPressNotificationContentExtension
    WordPressAllTimeWidget
    WordPressThisWeekWidget
    WordPressIntents
  ].map { |suffix| "#{root}.#{suffix}" }

  [root, *extensions_bundle_ids]
end
```

And then in most places use `all_bundle_identifiers` (and let the default value for `root` be used), but in places like here for the alpha, use `all_bundle_identifiers(root: 'org.wordpress.alpha')` instead (which we could consider abstracting away into a `ALPHA_VERSION_BUNDLE_IDENTIFIER` constant at the top of the Fastfile, next to the AppStore one)

Would feel neater to me than to use `id.sub` 🙂  wdyt?

mokagio: Nice! Will open a PR to adopt this ✨

PR #16806, opened 2021-07-05
Extract swift secret files outside of repo

mokagio:
Follow up on _PR_ to extract the remaining secrets outside of the repository. These secrets are `.swift` file containing values that the app needs at build time.

I built on top of the already existing "Run Script" Build Phase, with the following modifications:

- Moved all the code to a dedicated script to make it easier to modify and review
- The automation now looks for the secrets in the new `~/.configure/wordpress-ios/secrets/` folder
- If secrets cannot be found in a Debug build, fall back to the example ones. In a Release build, fail the build early.

~~_Note that this PR is based on top of `move-fastlane-secrets-outside-repository` for ease of review. Ideally, we should land this into 17.7 (see milestone) to test it end-to-end in CI, but it's okay if we don't get there._~~

## How to test

1. Checkout this branch and run a build in one, some, or all of the "WordPress", "WordPress-Alpha", "WordPress-Internal", and "Jetpack" schemes.
2. Verify it shows the example secrets warning:

Screen Shot 2021-07-05 at 12 10 52 pm

3. Try to archive the project in the same scheme
4. Verify the build fails with the error from the script:

Screen Shot 2021-07-05 at 12 15 29 pm

5. Run `bundle exec fastlane run configure_apply`
6. Repeat the tests above and verify they succeed

## Regression Notes

All the steps above represent a build from Xcode. Interestingly, I found that an extra step was required for the automated build via Fastlane (see inline comments). It's possible there are other steps that might not work in an automated build which I'm not aware of.

I tested the build locally and ensured it worked. I also run the installable build step for both WordPress and Jetpack and saw them both [succeed](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-iOS/23471/workflows/a4c75067-a5c4-41e0-8186-4fb27c8ff37b).

N.A.

---

### DIFF
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -3,7 +3,7 @@
 set -e
 
 # The Secrets File Sources
-SECRETS_ROOT="${PROJECT_ROOT}/.configure-files"
+SECRETS_ROOT="${HOME}/.configure/wordpress-ios/secrets"
 
 PRODUCTION_SECRETS_FILE="${SECRETS_ROOT}/WordPress-Secrets.swift"
 INTERNAL_SECRETS_FILE="${SECRETS_ROOT}/WordPress-Secrets-Internal.swift"
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -65,35 +65,35 @@ mkdir -p $(dirname "$SECRETS_DESTINATION_FILE")
 # If the WordPress Production Secrets are available for WordPress, use them
 if [ -f "$PRODUCTION_SECRETS_FILE" ] && [ "$BUILD_SCHEME" == "WordPress" ]; then
     echo "Applying Production Secrets"
-    cp -v $PRODUCTION_SECRETS_FILE "${SECRETS_DESTINATION_FILE}"
+    cp -v "$PRODUCTION_SECRETS_FILE" "${SECRETS_DESTINATION_FILE}"
     exit 0
 fi
 
 # If the WordPress Internal Secrets are available, use them
 if [ -f "$INTERNAL_SECRETS_FILE" ] && [ "${BUILD_SCHEME}" == "WordPress Internal" ]; then
     echo "Applying Internal Secrets"
-    cp -v $INTERNAL_SECRETS_FILE "${SECRETS_DESTINATION_FILE}"
+    cp -v "$INTERNAL_SECRETS_FILE" "${SECRETS_DESTINATION_FILE}"
     exit 0
 fi
 
 # If the WordPress Alpha Secrets are available, use them
 if [ -f "$ALPHA_SECRETS_FILE" ] && [ "${BUILD_SCHEME}" == "WordPress Alpha" ]; then
     echo "Applying Alpha Secrets"
-    cp -v $ALPHA_SECRETS_FILE "${SECRETS_DESTINATION_FILE}"
+    cp -v "$ALPHA_SECRETS_FILE" "${SECRETS_DESTINATION_FILE}"
     exit 0
 fi
 
 # If the Jetpack Secrets are available (and if we're building Jetpack) use them
 if [ -f "$JETPACK_SECRETS_FILE" ] && [ "${BUILD_SCHEME}" == "Jetpack" ]; then
     echo "Applying Jetpack Secrets"
-    cp -v $JETPACK_SECRETS_FILE "${SECRETS_DESTINATION_FILE}"
+    cp -v "$JETPACK_SECRETS_FILE" "${SECRETS_DESTINATION_FILE}"
     exit 0
 fi
 
 # If the developer has a local secrets file, use it
 if [ -f "$LOCAL_SECRETS_FILE" ]; then
     echo "Applying Local Secrets"
-    cp -v $LOCAL_SECRETS_FILE "${SECRETS_DESTINATION_FILE}"
+    cp -v "$LOCAL_SECRETS_FILE" "${SECRETS_DESTINATION_FILE}"
     exit 0
 fi
 
@@ -114,6 +114,6 @@ case $CONFIGURATION in
   *)
     echo "warning: $COULD_NOT_FIND_SECRET_MSG. Falling back to $EXAMPLE_SECRETS_FILE. In a Release build, this would be an error. $INTERNAL_CONTRIBUTOR_MSG and try again. If you are an external contributor, you can ignore this warning."
     echo "Applying Example Secrets"
-    cp -v $EXAMPLE_SECRETS_FILE $SECRETS_DESTINATION_FILE
+    cp -v "$EXAMPLE_SECRETS_FILE" "$SECRETS_DESTINATION_FILE"
     ;;
 esac
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -1,6 +1,4 @@
-#!/bin/sh
-
-set -e
+#!/bin/sh -euo pipefail
 
 # The Secrets File Sources
 SECRETS_ROOT="${HOME}/.configure/wordpress-ios/secrets"
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -79,7 +79,7 @@ if [ -f "$INTERNAL_SECRETS_FILE" ] && [ "${BUILD_SCHEME}" == "WordPress Internal
 fi
 
 # If the WordPress Alpha Secrets are available, use them
-if [ -f "$INTERNAL_SECRETS_FILE" ] && [ "${BUILD_SCHEME}" == "WordPress Alpha" ]; then
+if [ -f "$ALPHA_SECRETS_FILE" ] && [ "${BUILD_SCHEME}" == "WordPress Alpha" ]; then
     echo "Applying Alpha Secrets"
     cp -v $ALPHA_SECRETS_FILE "${SECRETS_DESTINATION_FILE}"
     exit 0
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -52,6 +52,19 @@ if [ -f "$LOCAL_SECRETS_FILE" ]; then
     exit 0
 fi
 
-# Use the example secrets file as a last resort
-echo "Applying Example Secrets"
-cp -v $EXAMPLE_SECRETS_FILE $SECRETS_DESTINATION_FILE
+# None of the above secrets was found. Use the example secrets file as a last
+# resort, unless building for Release.
+case $CONFIGURATION in
+  # There are three release configurations: Release, Release-Alpha, and
+  # Release-Internal. Since they all start with "Release" we can use a pattern
+  # to check for them.
+  Release*)
+    echo "error: Could not find secrets at $SECRETS_ROOT. Cannot continue release build."
+    exit 1
+    ;;
+  *)
+    echo "Applying Example Secrets"
+    echo "warning: Could not find secrets at $SECRETS_ROOT, falling back to $EXAMPLE_SECRETS_FILE. In a release build, this would be an error."
+    cp -v $EXAMPLE_SECRETS_FILE $SECRETS_DESTINATION_FILE
+    ;;
+esac
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -13,7 +13,7 @@ EXAMPLE_SECRETS_FILE="${SRCROOT}/Credentials/Secrets-example.swift"
 
 # The Secrets file destination
 SECRETS_DESTINATION_FILE="${BUILD_DIR}/Secrets/Secrets.swift"
-mkdir -p "basename $SECRETS_DESTINATION_FILE"
+mkdir -p $(dirname $SECRETS_DESTINATION_FILE)
 
 # If the WordPress Production Secrets are available for WordPress, use them
 if [ -f "$PRODUCTION_SECRETS_FILE" ] && [ "$BUILD_SCHEME" == "WordPress" ]; then
--- a/Scripts/BuildPhases/ApplyConfiguration.sh
+++ b/Scripts/BuildPhases/ApplyConfiguration.sh
@@ -5,13 +5,60 @@ set -e
 # The Secrets File Sources
 SECRETS_ROOT="${HOME}/.configure/wordpress-ios/secrets"
 
+# To help the Xcode build system optimize the build, we want to ensure each of
+# the secrets we want to copy is defined as an input file for the run script
+# build phase.
+#
+# > The Xcode Build System will use [these files] to determine if your run
+# > scripts should actually run or not. So this should include any file that
+# > your run script phase, the script content, is actually going to read or
+# > look at during its process.
+#
+# > If you have no input files declared, the Xcode build system will need to
+# > run your run script phase on every single build.
+#
+# https://developer.apple.com/videos/play/wwdc2018/408/
+function ensure_is_in_input_files_list() {
+  # Loop through the file input lists looking for $1. If not found, fail the
+  # build.
+  if [ -z "$1" ]; then
+    echo "error: Input file list verification needs a path to verify!"
+    exit 1
+  fi
+
+  i=0
+  found=false
+  while [[ $i -lt $SCRIPT_INPUT_FILE_LIST_COUNT && "$found" = false ]]
+  do
+    # Need this two step process to access the input at index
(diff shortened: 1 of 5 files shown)


Scripts/BuildPhases/ApplyConfiguration.sh
```diff
@@ -1,16 +1,21 @@
...
+ALPHA_SECRETS_FILE="${SECRETS_ROOT}/WordPress-Secrets-Alpha.swift"
+JETPACK_SECRETS_FILE="${SECRETS_ROOT}/Jetpack-Secrets.swift"
+
 LOCAL_SECRETS_FILE="${SRCROOT}/Credentials/Secrets.swift"
 EXAMPLE_SECRETS_FILE="${SRCROOT}/Credentials/Secrets-example.swift"
 
 # The Secrets file destination
 SECRETS_DESTINATION_FILE="${BUILD_DIR}/Secrets/Secrets.swift"
-mkdir -p "basename $SECRETS_DESTINATION_FILE"
+mkdir -p $(dirname $SECRETS_DESTINATION_FILE)
```
mokagio: This code never worked but we never noticed because:

1. There was no `set -e` in the script
2. The folder was created anyway because of the "Output Files" parameter in the Run Script configuration

Scripts/BuildPhases/ApplyConfiguration.sh
```diff
@@ -47,6 +52,19 @@ if [ -f "$LOCAL_SECRETS_FILE" ]; then
...
+  # There are three release configurations: Release, Release-Alpha, and
+  # Release-Internal. Since they all start with "Release" we can use a pattern
+  # to check for them.
+  Release*)
+    echo "error: Could not find secrets at $SECRETS_ROOT. Cannot continue release build."
+    exit 1
+    ;;
+  *)
+    echo "Applying Example Secrets"
+    echo "warning: Could not find secrets at $SECRETS_ROOT, falling back to $EXAMPLE_SECRETS_FILE. In a release build, this would be an error."
```
mokagio: Keen on suggestions on the wording of these two errors. The PR description shows how they appear in Xcode. I'd like them to be actionable for new A12s and clear for external contributors. 

When we'll get down to zero warnings, this one will be visible to everyone running the project without secrets. Currently, it gets lost in the noise. I wonder if an emoji could help make it more noticeable.

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -16293,10 +16291,8 @@
 			files = (
 			);
 			inputFileListPaths = (
-				"$(SRCROOT)/Credentials/secrets-manifest.xcfilelist",
 			);
 			inputPaths = (
-				"$(SRCROOT)/../Scripts/BuildPhases/ApplyConfiguration.sh",
```
mokagio: Interestingly, I had to keep the output paths parameter, otherwise the build would fail, but only when run through `xcodebuild`, not from Xcode.

image

```
❌  error: Build input file cannot be found: '/Users/gio/Developer/a8c/wpios/DerivedData/Build/Intermediates.noindex/ArchiveIntermediates/Jetpack/BuildProductsPath/Secrets/Secrets.swift' (in target 'Jetpack' from project 'WordPress')
```

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -15658,10 +15658,8 @@
...
```
jkmassel: I think we still want to re-run the script if it fails, no?

WordPress/Credentials/secrets-manifest.xcfilelist
```diff
@@ -1,6 +0,0 @@
-#  secrets-manifest.xcfilelist
-
-$(SRCROOT)/../.configure-files/Secrets.swift
```
jkmassel, line 3: Should we keep this file and reference the files using `${HOME}/.configure/...` in order to ensure we rebuild if they change?

jkmassel (commented):
Left some suggestions for your consideration

mokagio:
@jkmassel thank you for your review.

It turns out my understanding of what the input and output parameters are for was flawed. I did some reading:

- https://developer.apple.com/videos/play/wwdc2018/408/
- https://developer.apple.com/documentation/xcode-release-notes/build-system-release-notes-for-xcode-10
- https://nathanwong.co.uk/post/xcode-buildphases/
- https://indiestack.com/2014/12/speeding-up-custom-script-phases/

I have a better picture now and I'll update accordingly. In short, I'll add back and see if I can optimize the way we specify both inputs and outputs.

**Thanks!**

WordPress/Credentials/secrets-manifest.xcfilelist
mokagio: > Should we keep this file

Following up to my comment [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/16806#issuecomment-876000701) I think the answer is yes. Thanks noticing this.

> reference the files using `${HOME}/.configure/...`

That would be great. Xcode has [some issues](https://github.com/mokagio/HomeFolderTest) expanding `$HOME` when used in the project file, but maybe it can manage in a build phase. I'll try 🤞

Scripts/BuildPhases/ApplyConfiguration.sh
```diff
@@ -1,16 +1,21 @@
 #!/bin/sh
 
+set -e
+
```
AliSoftware: What about the more common way of setting this on the shebang directly, and add other common options for super-safe mode?

```suggestion
#!/bin/sh -euo pipefail
```

Scripts/BuildPhases/ApplyConfiguration.sh
```diff
@@ -1,16 +1,21 @@
...
 LOCAL_SECRETS_FILE="${SRCROOT}/Credentials/Secrets.swift"
 EXAMPLE_SECRETS_FILE="${SRCROOT}/Credentials/Secrets-example.swift"
 
 # The Secrets file destination
 SECRETS_DESTINATION_FILE="${BUILD_DIR}/Secrets/Secrets.swift"
-mkdir -p "basename $SECRETS_DESTINATION_FILE"
+mkdir -p $(dirname "$SECRETS_DESTINATION_FILE")
 
 # If the WordPress Production Secrets are available for WordPress, use them
 if [ -f "$PRODUCTION_SECRETS_FILE" ] && [ "$BUILD_SCHEME" == "WordPress" ]; then
```
AliSoftware, line 68: I noticed that in the following lines where we invoke `cp` down below, the input paths to `cp` are not quoted, could be worth adding quotes for safety and future-proofing this.

For example:
```
cp -v "${PRODUCTION_SECRETS_FILE}" "${SECRETS_DESTINATION_FILE}"
```

Scripts/BuildPhases/ApplyConfiguration.sh
AliSoftware: What about:
```
echo "error: Unable to find secrets for file ${SECRETS_DESTINATION_FILE} in ${SECRETS_ROOT}. Cannot continue release build. If you are an Automattician, run fastlane configure_apply to update your secrets."
```
and
```
echo "warning: Unable to find secrets for file ${SECRETS_DESTINATION_FILE} in ${SECRETS_ROOT}. Falling back to ${EXAMPLE_SECRETS_FILE} example file. If you are an Automattician, you will need to fix this if you want to be able to do Release builds."
```

A bit verbose maybe, but better be actionable and clear than succinct and unclear 😉

WordPress/WordPress.xcodeproj/project.pbxproj
AliSoftware: Yep, or if its source changed.

I agree we should restore this, and same for line 16299 below.

WordPress/WordPress.xcodeproj/project.pbxproj
AliSoftware: Yep, that's how outputPaths works and what it's for.

I think the fact that it didn't fail from Xcode GUI is a bug/glitch, maybe because the project was already build from your `xcodebuild` test and didn't have to incremental-build when you tried in Xcode?

But the goal is to tell Xcode which files this script phase generates, so that it can resolve the dependency graph when figuring out which targets, dependencies and phases need to be run and which can be skipped in incremental builds if their output are already up-to-date.

It also allows Xcode to know that this file – which is one of the input of the "Compile Sources" build phase, so which needs to be present before that phase runs – will be generated by this script. Since Xcode does the full dependency graph analysis of the various build phases first, and only runs them once it has resolved the whole graph (both in order to avoid starting to build stuff only to fail later if it can predict earlier on that the failure will happen, and to optimize build time in incremental builds by determining what can be skipped), you need to fill in that Output Files here to let it resolve the dependency graph and optimize incremental builds properly.

Scripts/BuildPhases/ApplyConfiguration.sh
```diff
@@ -1,16 +1,21 @@
...
+PRODUCTION_SECRETS_FILE="${SECRETS_ROOT}/WordPress-Secrets.swift"
+INTERNAL_SECRETS_FILE="${SECRETS_ROOT}/WordPress-Secrets-Internal.swift"
+ALPHA_SECRETS_FILE="${SECRETS_ROOT}/WordPress-Secrets-Alpha.swift"
+JETPACK_SECRETS_FILE="${SECRETS_ROOT}/Jetpack-Secrets.swift"
+
 LOCAL_SECRETS_FILE="${SRCROOT}/Credentials/Secrets.swift"
 EXAMPLE_SECRETS_FILE="${SRCROOT}/Credentials/Secrets-example.swift"
 
 # The Secrets file destination
 SECRETS_DESTINATION_FILE="${BUILD_DIR}/Secrets/Secrets.swift"
```
AliSoftware, line 64: I wonder if it would not be better and more idiomatic to put that file in `$(DERIVED_FILE_DIR)/Secrets.swift` instead (and adjust the project references accordingly), since `DERIVED_FILE_DIR` is the location in which Xcode puts files generated during the build (e.g. CoreData generated classes, Intents generated class files, …) and where it expects us to put these kind of generated files (which it then automatically picks to be compiled)…

Maybe not a big deal as long as we are super-sure that whatever `SECRETS_DESTINATION` we use is `.gitignore`'d and does not risk to be commited, just figured I'd mention what the usual conventions are there 🙂

AliSoftware (commented):
I agree with Jeremy's feedback and suggestion to restore the input files.

I also spotted another issue in the script (which I couldn't inline-comment on since it was not part of this diff and already present before this PR) which is worth fixing: on line 35 of the script where you're about to copy the file for Alpha, the `-f` test is on `INTERNAL_SECRETS_FILE` instead of testing the `ALPHA_SECRETS_FILE` that you're about to copy.

---

Finally, maybe not for this PR and an optimization only for later, but it could be nice to **configure this script as a dedicated "Aggregate Target"** that the WordPress target would depend on, rather than a build phase in the target.

The advantages of doing this are:
 * You can DRY the logic and reuse it between WP and JP (and other targets if needed) without having to repeat the phase in each target
 * But more importantly, Xcode will resolve the dependencies first and will thus execute this Aggregate Target… before trying to build any dependency of the WordPress or Jetpack target, which means that e.g. when you try to Archive, it will error right from the beginning – instead of waiting your time building the dependencies / pods first and only then running the app target afterwards and error-ing during the script build phase only then.

FYI, I gave it a quick try to compare:
<details><summary>12s for clean Archive to fail with the "Aggregate Target as a dependency" technique</summary>

Since Xcode only prepares the directories for other dependant targets but doesn't start building those targets before it runs the first dependency that is my "Generate Secrets" aggregate target, allowing it to fail early.

image
</details>
<details><summary>Compared to 2'47" (!) for clean Archive to fail with the current "Build Phase" technique</summary>

Here Xcode first builds all the dependant targets (i.e. Pods) first, and only then gets to running the WordPress target and thus the build phase, taking way longer to reach the failure point.

image

</summary>

Scripts/BuildPhases/ApplyConfiguration.sh
mokagio: 👍

Scripts/BuildPhases/ApplyConfiguration.sh
mokagio: I had never seen this, but I like it.

I'm curious to know what makes you say that it is more common? That's not to challenge your statement, but since I'd never seen that syntax, I'm curious to know which authoritative resources I can look into next time 😄 Thanks.

AliSoftware: Ahah i don't have any authoritative source on that, it's just that my experience is opposite to you, as I saw the flags set in the shebang way more often that seeing them set using `set` via a separate command, so I assumed the former was more common 😛

I've often seen, in articles suggesting to set those flags in eg Script Build Phases — to ensure the Xcode build fails as soon as one command in the script fails instead of only looking at exit code of the last command of the script — suggestions to set it on the "Shell" field (eg `/bin/sh -e`) too, so that's what I got used to do… but YMMV 🙃

WordPress/Credentials/secrets-manifest.xcfilelist
mokagio: BTW, `$HOME` in the file list file worked 🎉 See f8531a8b809d64cd294111071dd3aa863041820f

Scripts/BuildPhases/ApplyConfiguration.sh
mokagio: Thanks for the extra context. I like this setup and I'm keen to use it. 

I tried researching whether it was a appropriate but didn't get far, too much conflicting information.

I say let's adopt it (it's neat to have them all in a oneliner) and see how it plays out for us.

Scripts/BuildPhases/ApplyConfiguration.sh
mokagio: Resolved in a43b65b

mokagio:
> Finally, maybe not for this PR and an optimization only for later, but it could be nice to configure this script as a dedicated "Aggregate Target" that the WordPress target would depend on, rather than a build phase in the target.

💯 – This is how Simplenote and WooCommerce iOS do it, so it would also make the setup consistent, on top all the advantages you listed which I agree with. It didn't occur to me to list it as a possible improvement, but we should definitely do it as a followup to this secrets work.

Scripts/BuildPhases/ApplyConfiguration.sh
mokagio: `DERIVED_FILE_DIR` is indeed the location where Xcode expects / suggests to put output files, as one can verify by adding a new file to the list:

image

I didn't do it originally because I didn't know about it 😅 . But, after spending some time trying to make it work, I think it wasn't done like that because of how the project is setup. Namely because developers can read `Secrets.swift` via the file navigator to edit it, which I think is a capability we want to keep for external contributors?

By putting the file in the `BUILD_PRODUCTS` we can reference it in the project using the "Relative to Build Products" location

image

Unfortunately, there is not "Relative to Derived Sources" option:

image

(I'm calling it "Derived Sources" because that's one of the [aliases for `DERIVED_FILE_DIR`](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW43))

It is possible to reference the file in the project by adding a custom location in Preferences > Locations > Custom Paths then update the location for the file:

image

With that, the app compiles but the file is still not accessible, it seems like the path used at build time and the one Xcode computes are different 🤔 

Screen Shot 2021-08-03 at 10 26 06 am

Even if we could make it work, the problem I see with that option is that it would require every developer to make that change in their Xcode setup, and repeat it on every new machine. That sounds like it would generate a lot of support requests. Granted, I'm sure we could find a way to automate it as part of the project bootstrap process, but I wonder if it's worth it.

I also tried bypassing the problem by removing the file reference, but that obviously didn't work because then Xcode didn't know where to find the values defined in `Secrets.swift`.

---

**TL;DR** I wasn't able to make it work in a portable way using `DERIVED_FILE_DIR`. I'm keen on suggestions and improvements because the flow is a bit awkward –even the fact of having the example file tracked in the project is a bit confusing– but I think it's work that might deserve dedicated focus later on.

AliSoftware: Agreed with all that, good points. Let's keep it the way it is then

mokagio:
@AliSoftware @jkmassel I think I addressed all your comments and concerns. Thank you so much for the guidance and patience.

This PR took a long time to finish (assuming you'll approve it) but at least I learned a lot and it made working on https://github.com/Automattic/simplenote-macos/pull/1012 and https://github.com/woocommerce/woocommerce-ios/pull/4713 a breeze.

Could you have another look? Thanks!

Scripts/BuildPhases/ApplyConfiguration.sh
jkmassel: Well this thread answers _that_ question for me 😜

jkmassel (approved):
This LGTM – I was a little confused because I couldn't get the warning to trigger – then I realized it's because I've reviewed this PR previously 🤦 

:shipit:

PR #16803, opened 2021-07-02
Fix Blogging Reminders flow when called from Site Settings

Gio2018:
Fixes #NA
This is the same as #16784 except it targets `17.7`

To test:
Make sure you have more than one site with no blogging reminders set (or delete and re-install the app)
Test on iPhone and iPad

- Build/run and select a site where you had not previously seen the Blogging Reminders flow
- Go to "Site Settings" then tap "Blogging Reminders" and make sure the prologue screens shows up
- Dismiss the flow (by completing the flow or dismissing any of the screens)
- Tap again on "Blogging Reminders" and verify that this time the prologue is not presented and the flow starts from the settings screen
- Publish a post (from the editor or from the draft list) and verify that the Blogging Reminders flow is not presented after publishing
- Select another site where you had not seen the Blogging Reminders yet
- Publish a post and make sure the Blogging Reminders flow starts, and the post publishing prologue is the first screen

this targets the frozen branch so that we could apply all the fixes to Blogging Reminders in `17.7`; cc @mokagio @startuptester 
## Regression Notes
1. Potential unintended areas of impact
None. This change only configures the starting point of the Blogging Reminders flow

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Test that the blogging reminders flow works as expected in all the use cases.

3. What automated tests I added (or what prevented me from doing so)
none
PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16809, opened 2021-07-05
Improve logging for site icon download errors

diegoreymendez:
We're having a really large number of errors caused by issues in downloading site icons.  These errors are not crashes though, so I decided to log them to the local log file instead of directly to Sentry.

The reason I'm aiming this at 17.7 is because there's a huge number logs for this error (No 1 in the last 14 days), and if we assume it to be a crash, it looks much more important than it really is.

I'm also adding some logs to try and narrow down what the actual issue is.

## To test:

Simply build and make sure everything works as before.  I just  changed how we're logging some errors.

## Regression Notes

None.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
+++ b/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
@@ -258,13 +258,12 @@ private extension UIImageView {
     }
 
     private func logURLOptimization(from original: String, to optimized: URL, for blog: Blog) {
-        let blogInfo: String = {
-            guard blog.isAccessibleThroughWPCom() else {
-                return "self-hosted with url: \(blog.url ?? "unknown")"
-            }
-
-            return "dot-com-accessible: \(blog.url ?? "unknown"), id: \(blog.dotComID ?? 0)"
-        }()
+        let blogInfo: String
+        if blog.isAccessibleThroughWPCom() 
+            blogInfo = "dot-com-accessible: \(blog.url ?? "unknown"), id: \(blog.dotComID ?? 0)"
+        } else {
+            blogInfo = "self-hosted with url: \(blog.url ?? "unknown")"
+        }
 
         DDLogInfo("URL optimized from \(original) to \(optimized.absoluteString) for blog \(blogInfo)")
     }
--- a/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
+++ b/WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
@@ -51,7 +51,7 @@ extension UIImageView {
             return
         }
 
-        trackURLOptimization(from: path, to: siteIconURL)
+        logURLOptimization(from: path, to: siteIconURL)
 
         let request = URLRequest(url: siteIconURL)
         downloadSiteIcon(with: request, placeholderImage: placeholderImage)
@@ -94,12 +94,9 @@ extension UIImageView {
             case .failure(let error):
                 if case .requestCancelled = (error as? AFIError) {
                     // Do not log intentionally cancelled requests as errors.
-                } else if case let Alamofire.AFError.responseValidationFailed(reason) = error,
-                          case let Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code) = reason,
-                          code == 404 {
-                    // Do not log 404 errors since they are expected for site icons
                 } else {
                     WordPressAppDelegate.crashLogging?.logError(error)
+                    DDLogError(error.localizedDescription)
                 }
             }
         })
@@ -127,11 +124,11 @@ extension UIImageView {
             return
         }
 
-        trackURLOptimization(from: siteIconPath, to: siteIconURL)
+        logURLOptimization(from: siteIconPath, to: siteIconURL, for: blog)
 
         let host = MediaHost(with: blog) { error in
             // We'll log the error, so we know it's there, but we won't halt execution.
-            WordPressAppDelegate.crashLogging?.logError(error)
+            DDLogError(error.localizedDescription)
         }
 
         let mediaRequestAuthenticator = MediaRequestAuthenticator()
@@ -141,7 +138,7 @@ extension UIImageView {
             onComplete: { [weak self] request in
                 self?.downloadSiteIcon(with: request, placeholderImage: placeholderImage)
         }) { error in
-            WordPressAppDelegate.crashLogging?.logError(error)
+            DDLogError(error.localizedDescription)
         }
     }
 }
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Fyi @startuptester - this issue turned out not user facing, but it's currently listed as the No. 1 issue in Sentry.  By logging this differently I'm making sure it's not interpreted as a crash, thus reducing some noise from Sentry.

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
```diff
@@ -247,3 +246,26 @@ extension UIImageView {
...
+    }
+
+    private func logURLOptimization(from original: String, to optimized: URL, for blog: Blog) {
+        let blogInfo: String = {
+            guard blog.isAccessibleThroughWPCom() else {
+                return "self-hosted with url: \(blog.url ?? "unknown")"
+            }
+
+            return "dot-com-accessible: \(blog.url ?? "unknown"), id: \(blog.dotComID ?? 0)"
+        }()
```
mokagio: Just for fun, here's a different way to write this code that _might_ be more concise. Totally up to the author's preference. No even a nitpick, just something neat in Swift that sometimes makes it unnecessary to define an inlined closure for a conditional assignment.

```suggestion
        let blogInfo: String
        if blog.isAccessibleThroughWPCom() 
            blogInfo = "dot-com-accessible: \(blog.url ?? "unknown"), id: \(blog.dotComID ?? 0)"
        } else {
            blogInfo = "self-hosted with url: \(blog.url ?? "unknown")"
        }
```

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
```diff
@@ -92,13 +94,9 @@ extension UIImageView {
...
                     // Do not log intentionally cancelled requests as errors.
-                } else if case let Alamofire.AFError.responseValidationFailed(reason) = error,
-                          case let Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code) = reason,
-                          code == 404 {
-                    // Do not log 404 errors since they are expected for site icons
-                }
-                else {
+                } else {
                     WordPressAppDelegate.crashLogging?.logError(error)
+                    DDLogError(error.localizedDescription)
```
mokagio: In the rest of the code, the call to `logError` has been replaced with `DDLogError()`. Here, it runs alongside it. Is this intentional?

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
```diff
@@ -92,13 +94,9 @@ extension UIImageView {
             case .failure(let error):
                 if case .requestCancelled = (error as? AFIError) {
                     // Do not log intentionally cancelled requests as errors.
-                } else if case let Alamofire.AFError.responseValidationFailed(reason) = error,
-                          case let Alamofire.AFError.ResponseValidationFailureReason.unacceptableStatusCode(code) = reason,
-                          code == 404 {
-                    // Do not log 404 errors since they are expected for site icons
-                }
-                else {
```
mokagio, line 100: Removing this conditional branch means if we get a 404, it will be logged.

Did you do this intentionally to help diagnosing the root cause of the issue?

mokagio (commented):
Code LGTM, only left a minor question.

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
diegoreymendez: I may be wrong about my interpretation.  Ping @bjtitus since I think you know the context of this.

The reason why I think the error was not being logged at all is that initially we were logging it as an error to Sentry, which is quite more "aggressive" than regular logging.  In other words I think someone found themselves in the same situation I'm in right now - we're logging this too aggressively.

In terms of Groundskeeping this can come across as a crash and raise alarms.

With my change we're logging both 404 and 400 errors locally so we can figure out eventually the root cause behind the downloads failing - if the site icon is fine, and the URL is correct, we should get none of these errors AFAIK.  My theory is we're breaking some icons with the URL optimization we're doing for photon, so logging cases in which we're receiving a 404 is, IMHO, desired.

In practical terms, when I look at the next log file, I would like to know of any errors here that are not caused by the user cancelling the download.

Does that make sense?

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
diegoreymendez: This is not intentional... I forgot to clean up the original line, good catch!

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
diegoreymendez: Yeah that was intentional.  I prefer the inline approach in general because to me it looks much cleaner and contained.

That said I don't feel strongly about either solution so I'm good with accepting your suggestion.

diegoreymendez: The only downside about code suggestions is it's hard to know beforehand if there's any issue with it. :P

But it's in.

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
diegoreymendez: Removed

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
mokagio: Makes total sense to me 👍 I think logging everything makes total sense. I mainly asked because that error was originally ignored and I wondered why 👍

diegoreymendez:
@mokagio - The PR is ready for re-review.  I haven't added anything new, but I need 1 approval before being able to merge.

WordPress/Classes/Extensions/UIImageView+SiteIcon.swift
bjtitus: That interpretation is correct. The original intent is maintained if the errors are just logged locally.

mokagio: Thanks @bjtitus 👍

mokagio (approved):
:shipit:

mokagio:
@diegoreymendez ✅ :shipit: 

I've held on shipping a new beta just for this one in light of having #16803 close to be merged and @startuptester telling me there's another bugfix on its way.

Also, given this is something we see in production but haven't been able to effectively reproduce ourselves, it seems to me like there's not much extra data we could gather from a beta. 

But ☝️ Happy to ship a dedicated beta if you think it'd be valuable 😄 Just let me know.

diegoreymendez:
> But ☝️ Happy to ship a dedicated beta if you think it'd be valuable 😄 Just let me know.

I think there's no need to.  I was hoping to have it out in production mostly, because the noise in Sentry isn't good IMHO.  But there's no rush for a dedicated beta release.

PR #16819, opened 2021-07-07
Fix follow site button refresh issues

diegoreymendez:
As reported by @elibud the "follow site" button in the notifications tab is sometimes misbehaving and showing errors, not allowing the user to follow or unfollow a site, or misreporting the action that was performed by the user.

This PR addresses the existing issues in two ways:

1. If the user tries to follow a site they're already following (this can happen when the follow-state is not synchronized / up to date), and the backend returns an "already following" error (or, in the case of unfollowing a "not subscribed" error), then the UI will no longer report an error, and will tell the user all is good - since what they wanted to do is ultimately what they'll get.
2. By changing the follow state we use for the dispatch notices.  We want to make sure the message the user sees corresponds with the follow or unfollow attempt (regardless of what the current follow state for the site is).  This change is important because the follow state of a site in CoreData can be changed while the service is executed (for example if a user taps the follow button repeatedly).
3. By disabling user interaction in the follow buttons until the service calls complete.  This is necessary because when tapping the buttons multiple times in a row I've found it's actually possible for the follow and unfollow services to return out-of-order (this is most likely because the follow service calls requires two calls to remote, whereas the unfollow service call only one).

### Notifications Tab:

1. Go going into a comment notification, and tapping on top until you get to the post.
2. Then tap the follow site button, on or many times.
3. The button should not work until the result of the previous tap is processed.

### Reader Tab:

1. Go to Reader > Top Right Gear Icon > Followed Sites > Select a site to go into the detail for it.
2. Then tap the follow site button, on or many times.
3. The button should not work until the result of the previous tap is processed.

## Regression Notes

Any location where there's a follow button.

I've tested manually extensively.

None, as I don't believe this is something that can be tested automatically.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/ReaderPostService.m
+++ b/WordPress/Classes/Services/ReaderPostService.m
@@ -324,15 +324,15 @@ static NSString * const ReaderPostGlobalIDKey = @"globalID";
 }
 
 - (void)toggleFollowingForPost:(ReaderPost *)post
-                       success:(void (^)(void))success
-                       failure:(void (^)(NSError *error))failure
+                       success:(void (^)(BOOL follow))success
+                       failure:(void (^)(BOOL follow, NSError *error))failure
 {
     // Get a the post in our own context
     NSError *error;
     ReaderPost *readerPost = (ReaderPost *)[self.managedObjectContext existingObjectWithID:post.objectID error:&error];
     if (error) {
         if (failure) {
-            failure(error);
+            failure(true, error);
         }
         return;
     }
@@ -379,7 +379,7 @@ static NSString * const ReaderPostGlobalIDKey = @"globalID";
             [self refreshPostsForFollowedTopic];
         }
         if (success) {
-            success();
+            success(follow);
         }
         
         dispatch_async(dispatch_get_main_queue(), ^{
@@ -395,7 +395,7 @@ static NSString * const ReaderPostGlobalIDKey = @"globalID";
         [self setFollowing:oldValue forPostsFromSiteWithID:post.siteID andURL:post.blogURL];
 
         if (failure) {
-            failure(error);
+            failure(follow, error);
         }
     };
 
--- a/WordPress/Classes/Services/ReaderTopicService.m
+++ b/WordPress/Classes/Services/ReaderTopicService.m
@@ -473,14 +473,16 @@ static NSString * const ReaderTopicCurrentTopicPathKey = @"ReaderTopicCurrentTop
     }];
 }
 
-- (void)toggleFollowingForSite:(ReaderSiteTopic *)siteTopic success:(void (^)(void))success failure:(void (^)(NSError *error))failure
+- (void)toggleFollowingForSite:(ReaderSiteTopic *)siteTopic
+                       success:(void (^)(BOOL follow))success
+                       failure:(void (^)(BOOL follow, NSError *error))failure
 {
     NSError *error;
     ReaderSiteTopic *topic = (ReaderSiteTopic *)[self.managedObjectContext existingObjectWithID:siteTopic.objectID error:&error];
     if (error) {
         DDLogError(error.localizedDescription);
         if (failure) {
-            failure(error);
+            failure(true, error);
         }
         return;
     }
@@ -509,7 +511,7 @@ static NSString * const ReaderTopicCurrentTopicPathKey = @"ReaderTopicCurrentTop
         [self refreshPostsForFollowedTopic];
         
         if (success) {
-            success();
+            success(newFollowValue);
         }
     };
 
@@ -531,7 +533,7 @@ static NSString * const ReaderTopicCurrentTopicPathKey = @"ReaderTopicCurrentTop
         [[ContextManager sharedInstance] saveContext:self.managedObjectContext];
 
         if (failure) {
-            failure(error);
+            failure(newFollowValue, error);
         }
     };
 
--- a/WordPress/Classes/Services/ReaderTopicService.m
+++ b/WordPress/Classes/Services/ReaderTopicService.m
@@ -515,7 +515,17 @@ static NSString * const ReaderTopicCurrentTopicPathKey = @"ReaderTopicCurrentTop
 
     // Define failure block
     void (^failureBlock)(NSError *error) = ^void(NSError *error) {
-        // Revert changes on failure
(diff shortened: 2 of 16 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderStreamHeader.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 public protocol ReaderStreamHeaderDelegate: NSObjectProtocol {
-    func handleFollowActionForHeader(_ header: ReaderStreamHeader)
+    func handleFollowActionForHeader(_ header: ReaderStreamHeader, completion: @escaping () -> ())
```
aerych: `completion: @escaping () -> ())`
Nitpicky of me but it seems like we use the `() -> Void` syntax more often. Could be `completion: @escaping () -> Void)` for extra consistency.

WordPress/Classes/ViewRelated/Reader/ReaderStreamViewController.swift
```diff
@@ -1296,13 +1296,14 @@ import WordPressFlux
 
 extension ReaderStreamViewController: ReaderStreamHeaderDelegate {
 
-    func handleFollowActionForHeader(_ header: ReaderStreamHeader) {
+    func handleFollowActionForHeader(_ header: ReaderStreamHeader, completion: @escaping () -> ()) {
```
aerych: Also here.

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -324,15 +324,15 @@ - (void)setFollowing:(BOOL)following
...
+                       success:(void (^)(BOOL follow))success
+                       failure:(void (^)(BOOL follow, NSError *error))failure
 {
     // Get a the post in our own context
     NSError *error;
     ReaderPost *readerPost = (ReaderPost *)[self.managedObjectContext existingObjectWithID:post.objectID error:&error];
     if (error) {
         if (failure) {
-            failure(error);
+            failure(true, error);
```
aerych, line 335: Should this be `follow`?  If I'm following the logic correctly, by setting it to `true` it looks like `dispatchToggleFollowSiteMessage` would only ever dispatch `NoticeMessages.followFail`.  Am I missing something?

WordPress/Classes/Services/ReaderTopicService.m
```diff
@@ -473,14 +473,16 @@ - (void)tagTopicForTagWithSlug:(NSString *)slug success:(void(^)(NSManagedObject
...
+                       success:(void (^)(BOOL follow))success
+                       failure:(void (^)(BOOL follow, NSError *error))failure
 {
     NSError *error;
     ReaderSiteTopic *topic = (ReaderSiteTopic *)[self.managedObjectContext existingObjectWithID:siteTopic.objectID error:&error];
     if (error) {
         DDLogError(error.localizedDescription);
         if (failure) {
-            failure(error);
+            failure(true, error);
```
aerych, line 485: Should this be `follow` instead of `true`?

aerych (approved):
Hola @diegoreymendez 👋 
Looking good.  I tested via the simulator with the link conditioner set to Edge. This made requests slow enough for me to happy-click the follow buttons and confirm the UI was disabled until the current request completed.
Code wise I had just a couple of questions. Lemme know what you think.  Both were minor so :shipit: from me with or with out changes as you feel is best. :)

WordPress/Classes/ViewRelated/Reader/ReaderStreamHeader.swift
diegoreymendez: I actually did a search and missed this for some reason.  Thanks for  spotting!

WordPress/Classes/Services/ReaderPostService.m
diegoreymendez: This is something that I had particular difficulty in "resolving", if we can even consider it resolved.  Please be critical about what I say below, since I may be mistaken.

### First some context about what I did here, and why:

My understanding is that I can't know the follow value I should be returning here, because the `ReaderPost` object no longer exists.

With this in mind, and considering this is an odd scenario and (I believe) such an edge case, and in lack of a simple and good solution, I opted for returning `true` here - with the risk of showing the wrong error message to the user.  If we decide to move forward with this, I can add a comment to explain.

I think the ideal solution would be to ditch the `toggle...` methods in favour of more deterministic `follow...` and `unfollow...` methods, since I think that'd offer us a much easier mechanism to know the user's intention - which is clear in the UI (the user taps a button that reads "follow" or "unfollow") but becomes shady because of the `toggle...` logic.

The only problem with this approach is that it would not be a minor change.

### A question for you:

> Should this be `follow`?

Could I ask you to expand on what you mean by this precisely?  Maybe even show it to me in terms of code?

I agree with what you're saying, but the follow value isn't known here (unless maybe we pass it as an input parameter, kind of defeating the purpose of the `toggle...` method name).

aerych: > Could I ask you to expand on what you mean by this precisely? 

Ugh. This was me mis-reading the method signature and interpreting `follow` as a method argumement, rather than a block argument. 🤦   Mea Culpa. Sorry for the confusion there. 

> First some context about what I did here

Thanks for that :).  It makes perfect sense to me now.  I agree that getting away from the toggle approach would be best, and I agree that it's a larger change that's out of scope. :)  
Let's ship this as it does resolve the issue and, as you pointed out, it's unlikely for a user to encounter a missing model object. (And if they do we probably have a bigger problem.)

diegoreymendez: Thanks Eric!  I appreciate the time and guidance!

WordPress/Classes/Services/ReaderTopicService.m
diegoreymendez: For future reference: see the conversation above.

startuptester:
@diegoreymendez and @aerych, since this was a new regression in the beta, shouldn't this be merged to the frozen 17.7? 

cc: @mokagio

diegoreymendez:
> @diegoreymendez and @aerych, since this was a new regression in the beta, shouldn't this be merged to the frozen 17.7?
> 
> cc: @mokagio

For history: this was most likely long standing (due to the nature of my changes).  I didn't push for 17.8 because I think this wasn't a regression in 17.7, and was afraid to break 17.7.

PR #16823, opened 2021-07-07
Update Global Styles endpoint to support non-experimental paths

chipsnyder:
This updates the Global styles endpoint to use the non-experimental paths for fetching the block editor settings. 

**Related PR** https://github.com/wordpress-mobile/WordPressKit-iOS/pull/416

## To test:
For each of these cases, it'll be enough to open the editor and validate that the correct color options are available. It would also be useful to test using a network monitoring tool to make sure the correct paths are called. This will help ensure you aren't seeing cached data if you're reusing the same site for testing.

- [ ] Self-Hosted Site with WP 5.8 and Gutenberg 11.1
    - Expect the request to be made with `/wp-block-editor/v1`
- [ ] Self-Hosted Site with Gutenberg version <= 11.0
    - Expect the request to be made with `/wp-block-editor/v1` which will fail, so the fallback of `wp/v2/themes` should be used
- [ ] WPCom or Jetpack connected site with WP 5.8 and Gutenberg 11.0
    - Expect the request to be made with `/wp-block-editor/v1`
    - This will need to be tested with D63704 until that's deployed
- [ ] WPCom or Jetpack connected site with WP 5.8 and Gutenberg version >= 11.1
    - Expect the request to be made with `/wp-block-editor/v1`=
    - This will need to be tested with D63704 until that's deployed

## Regression Notes
N/A

N/A

Update unit tests to validate each flow

## PR submission checklist:


WordPress/Classes/Services/BlockEditorSettingsService.swift
```diff
@@ -106,25 +119,33 @@ private extension BlockEditorSettingsService {
...
                     let originalChecksum = self.blog.blockEditorSettings?.checksum ?? ""
                     self.updateBlockEditorSettingsCache(originalChecksum: originalChecksum, remoteSettings: remoteSettings, completion: completion)
                 }
-            case .failure(let error):
-                DDLogError("Error loading Block Editor Settings: \(error)")
+            case .failure(let err):
+                DDLogError("Error fetching editor settings: \(err)")
+                // The user may not have the gutenberg plugin installed so try /wp/v2/themes to maintain feature support.
+                // In WP 5.9 we may be able to skip this attempt.
+                self.fetchTheme(completion)
```
antonis, line 148: Retrying on error should be fine. I wonder if we can add an extra check for the specific error type (`404`).

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
```diff
@@ -191,9 +231,89 @@ class BlockEditorSettingsServiceTests: XCTestCase {
...
+        service = BlockEditorSettingsService(blog: blog, remoteAPI: mockOrgRemoteApi, context: context)
+
+        let waitExpectation = expectation(description: "Theme should be successfully fetched")
+        service.fetchSettings { _ in
+            waitExpectation.fulfill()
+        }
+
+        // The app will call the none-experimental path first but fail because of compatibility reasons
+        mockOrgRemoteApi.completionPassedIn!(.failure(NSError(domain: "test", code: 404, userInfo: nil)), HTTPURLResponse())
+        // The app will then retry the experimental path.
```
antonis: The test looks great. We just need to update the comment to reflect the removal of the experimental path.

antonis (approved):
Great work @chipsnyder 👍 
Tested this on an iPhone SE 2020 (iOS 14.4.1) and everything works as expected. The code LGTM 🎉 
I just left two minor comments.

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
chipsnyder: Good callout I missed this one.

WordPress/Classes/Services/BlockEditorSettingsService.swift
chipsnyder: Nothing specific here but I wasn't sure of about all of the ways this could fail so I decided to play it safe.

Issue #16740, opened 2021-06-24
Comment/Notification lists: update table section headers


ScoutHarris:
Create new date section header, and use in both Comments and Notifications lists.

- [x] Create new view matching the design. PR: #16826
- [x] #16852
- [x] #16851
- [x] Update Comment `sectionIdentifier` to match Notification `sectionIdentifier` so the lists are grouped the same.

Research notes:
- Should be able to copy and modify `NoteTableHeaderView` for both purposes.

Design from Zeplin:

<kbd><img width="450" alt="Screen Shot 2021-06-24 at 1 59 47 PM" src="https://user-images.githubusercontent.com/1816888/123324851-8d243080-d4f4-11eb-8d97-c073a96b82d1.png"></kbd>

PR #16826, opened 2021-07-08
Create section header view for Comments and Notifications

dvdchr:
Refs #16740

## Summary
This PR introduces a new header view for Comments Unification. Currently, I'm planning to put all related components in ViewRelated > Views > Snippet List.

A note on the name: previously I've considered `ListItem`, `TextSnippet`, `ContentSnippet`, `Notifiable`... But ended up on `Snippet` as the simplest one that kinda fits in Notifications and Comments. Naming sure is hard, and suggestions are welcome if you have a better name! 😅  

Here's how the new header looks on the Notifications screen:

Light 🌝  | Dark 🌚 
-- | --
16740_header_light | 16740_header_dark

## To Test
The new table header is still detached from the application flow. Only way to test this is by manually changing the header view in `NotificationsViewController`:
1. Register `SnippetTableHeaderView` as reusable header footer view.
2. Always return `SnippetTableHeaderView` in `tableView(_ tableView: UITableView, viewForHeaderInSection section: Int)` method.

## Regression Notes
None.

The new component is detached from application flow.

None.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
+++ b/WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
@@ -3,7 +3,7 @@
 extension WPStyleGuide {
     public enum List {
         // MARK: Section Headers
-        public static let sectionHeaderFont = WPStyleGuide.fontForTextStyle(.footnote, fontWeight: .semibold)
+        public static let sectionHeaderFont = WPStyleGuide.fontForTextStyle(.caption1, fontWeight: .medium)
         public static let sectionHeaderTitleColor = UIColor.textSubtle
         public static let sectionHeaderBackgroundColor = UIColor.basicBackground
 
--- a/WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
+++ b/WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
@@ -1,4 +1,4 @@
-/// Convenience constants catalog related for styling in snippet list.
+/// Convenience constants catalog related for styling List components.
 ///
 extension WPStyleGuide {
     public enum List {
--- a/WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
+++ b/WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
@@ -1,7 +1,7 @@
 /// Convenience constants catalog related for styling in snippet list.
 ///
 extension WPStyleGuide {
-    public enum Snippet {
+    public enum List {
         // MARK: Section Headers
         public static let sectionHeaderFont = WPStyleGuide.fontForTextStyle(.footnote, fontWeight: .semibold)
         public static let sectionHeaderTitleColor = UIColor.textSubtle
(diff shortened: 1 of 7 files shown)


aerych (changes requested):
Heya @dvdchr 👋 

Naming is hard. :)  Since this is a pretty generic control, and something that might eventually be used in other parts of the app as designs are refreshed over time, maybe a more generic name would work?  Maybe something like `ListHeaderFooterView`, or something else that's descriptive without being too closely paired with a particular feature?  What do ya think?

The layout is pretty close. :). There was an issue with Zeplin (now fixed) that might have made it hard to see all the design details. I think we need just a couple of tweaks:

The view seems a little tall. Zeplin shows a 28pt/px height.  It currently looks to be about 40pt/px. 
Screen Shot 2021-07-08 at 2 30 44 PM
Screen Shot 2021-07-08 at 2 52 15 PM
Screen Shot 2021-07-08 at 2 52 34 PM

The font specified in code is `.footnote` which seems a little large (13pt vs 12pt). It looks like we want to use `.caption1`.  I _think_ that should sort things out, but we can ping @mattmiklic for a design check to be sure.
Screen Shot 2021-07-08 at 2 33 43 PM

One other thing we should look at is how the control works with accessibility. It looks like the internal `textLabel` is still being used for accessibility labels.  In this example you can see where the value `5` is used for the accessibility label. 
Screen Shot 2021-07-08 at 3 03 25 PM

Code looks sharp. Nice job!

Back to you sir!

mattmiklic:
> The view seems a little tall. Zeplin shows a 28pt/px height. It currently looks to be about 40pt/px.
> The font specified in code is .footnote which seems a little large (13pt vs 12pt).

Good eye @aerych, agreed on both points 😃

dvdchr:
Thanks @aerych and @mattmiklic for the review! 

> Maybe something like ListHeaderFooterView, or something else that's descriptive without being too closely paired with a particular feature? What do ya think?

Yep, sounds good! I initially thought about using `List`, but felt like it's _too_ generic – hence I ended up with `Snippet`, since each list item essentially shows a snippet of _some content_. However, `ListTableHeaderView` and `ListTableViewCell` looks nicer in the eye, so I'll go ahead with your suggestion!

> The view seems a little tall. Zeplin shows a 28pt/px height. It currently looks to be about 40pt/px.
The font specified in code is .footnote which seems a little large (13pt vs 12pt).

Noted. I've confirmed that the ruler is working well now 🙂 , so I'll do the adjustments soon.

>  It looks like the internal textLabel is still being used for accessibility labels.

Good catch! I completely missed this 😬 . Will fix this soon! 💨

dvdchr:
### Updates

- Headers are fixed to match the designs.
  Light | Dark
  -- | --
  16740_light_fix | 16740_dark_fix

  - Text style adjusted to `caption1`, with `medium` weight. Turns out the whole time I've been testing with my simulator set to Medium type size instead of Large (the default). This brings `footnote` size to 12pt, and `caption1` to 11pt. 😅 
  - Header size adjusted to 26pt.

- Accessibility labels are added!

  16740_a11y_fixed

WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
```diff
@@ -0,0 +1,13 @@
+/// Convenience constants catalog related for styling in snippet list.
```
aerych: Should we change "snippet" in this comment?

aerych (approved):
Looks great @dvdchr!  I had one tiny question.  Other than that :shipit: ! :)

WordPress/Classes/ViewRelated/Views/List/WPStyleGuide+List.swift
dvdchr: Yes, thank you. 😄 Fixed on the latest commit.

PR #16834, opened 2021-07-09
Cherry-pick "Fix follow site button refresh issue (#16819)"

mokagio:
Like the titles says, it cherry-picks @diegoreymendez excellent work from #16819, which has already been reviewed and approved by @aerych, into the 17.7 release branch.

**Why?** The issue with the refresh button has been identified as a regression in 17.7 and it seemed appropriate to address it ASAP. It would have been different if it had been a long standing issue. The bug itself seemed important enough to grant this last minute addition to the release branch. Finally, it was possible to cherry-pick the changes cleanly (`git cherry-pick -n -m1 e8fe22c`, with `-n` to not commit immediately and give me a chance to discard the release notes change). cc @startuptester.

If you look at this diff side by side with #16819, you'll see that, other than for the `RELEASE-NOTES.txt` change which would be inappropriate in this branch, they are exactly the same. That should come as no surprise given the clean cherry-pick mentioned above, but is worth pointing out because it means there's no need to code review as it's been already done in the parent PR, and it's only necessary to manually verify the fix apply here too (I've done that, of course, but can always do with an extra set of eyes).

Following are the testing instructions and regression notes copied from #16819. P.S. You'll need the Network Link Conditioner. If, like me, you forgot how to set it up, [these instructions](https://medium.com/macoclock/how-to-simulate-poor-network-conditions-on-ios-simulator-and-iphone-faf35f0da1b5) will be helpful.

### Notifications Tab:

1. Go going into a comment notification, and tapping on top until you get to the post.
2. Then tap the follow site button, on or many times.
3. The button should not work until the result of the previous tap is processed.

### Reader Tab:

1. Go to Reader > Top Right Gear Icon > Followed Sites > Select a site to go into the detail for it.
2. Then tap the follow site button, on or many times.
3. The button should not work until the result of the previous tap is processed.

## Regression Notes

Any location where there's a follow button.

I've tested manually extensively. [Me too, I even ended up unfollowing a site directly from the reader screen and now I don't remember which one it is and I'm very sad about it –@mokagio]

None, as I don't believe this is something that can be tested automatically.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/ReaderPostService.h
+++ b/WordPress/Classes/Services/ReaderPostService.h
@@ -117,8 +117,8 @@ extern NSString * const ReaderPostServiceToggleSiteFollowingState;
  @param failure block called if there is any error. `error` can be any underlying network error.
  */
 - (void)toggleFollowingForPost:(ReaderPost *)post
-                       success:(void (^)(void))success
-                       failure:(void (^)(NSError *error))failure;
+                       success:(void (^)(BOOL follow))success
+                       failure:(void (^)(BOOL follow, NSError *error))failure;
 
 /**
  Toggle the saved for later status of the specified post.
--- a/WordPress/Classes/Services/ReaderPostService.m
+++ b/WordPress/Classes/Services/ReaderPostService.m
@@ -324,15 +324,15 @@ static NSString * const ReaderPostGlobalIDKey = @"globalID";
 }
 
 - (void)toggleFollowingForPost:(ReaderPost *)post
-                       success:(void (^)(void))success
-                       failure:(void (^)(NSError *error))failure
+                       success:(void (^)(BOOL follow))success
+                       failure:(void (^)(BOOL follow, NSError *error))failure
 {
     // Get a the post in our own context
     NSError *error;
     ReaderPost *readerPost = (ReaderPost *)[self.managedObjectContext existingObjectWithID:post.objectID error:&error];
     if (error) {
         if (failure) {
-            failure(error);
+            failure(true, error);
         }
         return;
     }
@@ -379,7 +379,7 @@ static NSString * const ReaderPostGlobalIDKey = @"globalID";
             [self refreshPostsForFollowedTopic];
         }
         if (success) {
-            success();
+            success(follow);
         }
         
         dispatch_async(dispatch_get_main_queue(), ^{
@@ -395,7 +395,7 @@ static NSString * const ReaderPostGlobalIDKey = @"globalID";
         [self setFollowing:oldValue forPostsFromSiteWithID:post.siteID andURL:post.blogURL];
 
         if (failure) {
-            failure(error);
+            failure(follow, error);
         }
     };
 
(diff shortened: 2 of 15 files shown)


mokagio:
Update: I also left the Network Link Conditioner on after finishing testing and got very annoyed at my internet for being _so damned slow!_

diegoreymendez:
For history:  don't think this issue is a regression in 17.7 - at least a big part of it is due to following has been working for a while.  That was part of why the PR was submitted against 17.8, instead of 17.7 - another part was to avoid causing trouble with the release.

mokagio:
Hey @diegoreymendez! Thanks for adding the extra context here 🙇‍♂️ 

I knew you had a reason for it, but with you being offline (the world is too big 😆) @startuptester and I just went ahead and made the call to add this.

AliSoftware:
@mokagio @diegoreymendez  I've tested the Installable Build (Alpha build 52615 – as seen in the about page on the video – that I triggered minutes ago on this PR), and if I stress-tap the follow button I still get the toast with "unable to follow" ("Impossible de s'abonner au site" in 🇫🇷 ) error popping up multiple times – which I think is what this cherry-pick was supposed to fix?

https://user-images.githubusercontent.com/216089/125068662-05beeb80-e0b6-11eb-9ee9-74b9a21ae8fb.mp4

mokagio:
Thanks for testing @AliSoftware. That's real odd.

It works on my end... Both with and without network link conditioner:

#### Simulator



#### iPhone 12 Pro


https://user-images.githubusercontent.com/1218433/125070634-91a73680-e0fb-11eb-977a-6628dcdd8553.MP4

_I wish the iOS video showed it, but I_ was _"happy tapping" the button._

AliSoftware:
Update: If I switch my device to English, then try to stress-test the button, the bug is fixed and I don't see the error toast. If I switch it back to French, and redo the stress-test, the error toast is back 😱

mokagio:
> Update: If I switch my device to English, then try to stress-test the button, the bug is fixed and I don't see the error toast. If I switch it back to French, and redo the stress-test, the error toast is back 😱

Confirmed. I tried with English and Italian, no issue; French, can reproduce. 😱 _That's all on device_.

mokagio:
I’m of two minds about what to do…

a) Merge because “it works for me, which is something”
b) Don’t merge, because if it doesn’t work 100% then it’s better not to add extra code to the build: code is a liability

Overall, I'm leaning towards merging.

Rationale: it doesn’t work in French, but it does work for English (and Italian) so it’s an improvement. @diegoreymendez  and @aerych  tested it, too, albeit against `develop` and considered it good.

Code _is_ a liability, but I think it this case the fix value justifies the risk.

I had a chat about this with @AliSoftware via Slack and we came to the conclusion above. Enabling auto-merge, so if Olivier approves, it'll get merged immediately and I'll move on with the code freeze.

AliSoftware (approved):
Approving to unblock the release, as explained by @mokagio 's comment above, but we'll still need to look into that case in French and what was happening here 🤔

diegoreymendez:
@AliSoftware @mokagio - For what it's worth, this PR doesn't prevent the follow and unfollow services from failing.  It merely fixes some instances in which we were registering a failure when there wasn't one.  Most specifically, we fix an issue where if you tap follow and you're already following, the App will no longer report a failure and roll the change back (and the same for unfollowing) - it will instead tell the user all is good since they attempted to follow and the end result is they're following.

I still can sometimes get 500 Internal errors from our backend when following and unfollowing repeatedly through Paw.app (which executes service calls).  In such scenarios, it's perfectly reasonable for the error message to pop up.

PR #16581, opened 2021-05-26
Fix missing support options.

diegoreymendez:
As reported in p5T066-2hL-p2#comment-8405

We were missing some support options because the ZenDesk credentials weren't being stored anymore.

The original issue was introduced by the removal of these lines: https://github.com/wordpress-mobile/WordPress-iOS/commit/a68b7f51722e1fe5afca7946c47bbb0d2aa2041b#diff-a8966393c02a2ab855ff7a8c50e5af636646d1c75292f797c1593e7458ebc340L351-L353

## To test:

In the log in flow, tap "Help" in the top-right  corner.
Make sure you can contact support and have a "My Tickets" option as well.

## Regression Notes

1. Potential unintended areas of impact

None that I can think of.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
3. What automated tests I added (or what prevented me from doing so)

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16837, opened 2021-07-09
Add UI test to verify Help section loads

thehenrybyrd:
This PR is part of updating our login tests to account for the Unified Login & Signup flows. Context & status: paaHJt-1Kx-p2
This PR adds a new UI test,  `testSupportScreenLoads`,  to check that the Support page loads and expected items (contact support, etc) appear.

This test should catch issues like https://github.com/wordpress-mobile/WordPress-iOS/pull/16581 in the future.
So, it's important that the Support page loads, but also that the key items appear.

To test:
Run `testSupportScreenLoads` from SignupTests.swift in Xcode.

One structure question I still have: is SignupTests a good place for this test? 
I put it there because:
1. LoginTests is getting pretty busy
2. It doesn't require logging in (it tests Help from a non-logged-in state)
But I'm not sure either reason justifies putting it in SignupTests. Should it have it's own file or be elsewhere?

Please comment on implementation as well as code quality! ❤️


WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
```diff
@@ -123,15 +124,15 @@ private extension SupportTableViewController {
...
-            helpSectionRows.append(HelpRow(title: LocalizedText.contactUs, action: contactUsSelected()))
-            helpSectionRows.append(HelpRow(title: LocalizedText.myTickets, action: myTicketsSelected(), showIndicator: ZendeskUtils.showSupportNotificationIndicator))
+            helpSectionRows.append(HelpRow(title: LocalizedText.contactUs, action: contactUsSelected(), accessibilityIdentifier: "contact-support-button"))
+            helpSectionRows.append(HelpRow(title: LocalizedText.myTickets, action: myTicketsSelected(), showIndicator: ZendeskUtils.showSupportNotificationIndicator, accessibilityIdentifier: "my-tickets-button"))
             helpSectionRows.append(SupportEmailRow(title: LocalizedText.contactEmail,
                                                    value: ZendeskUtils.userSupportEmail() ?? LocalizedText.emailNotSet,
                                                    accessibilityHint: LocalizedText.contactEmailAccessibilityHint,
-                                                   action: supportEmailSelected()))
+                                                   action: supportEmailSelected(), accessibilityIdentifier: "set-contact-email-button"))
```
mokagio: Nitpick. The rest of the arguments are defined on new lines, this new code should be consistent with that style.

```suggestion
                                                   action: supportEmailSelected(),
                                                   accessibilityIdentifier: "set-contact-email-button"))
```

As an aside, I find it useful to break long parameters list this way, because it removes the need for scrolling back and forth horizontally.

For example, to see the rest of the code at line 131, I need to scroll horizontally

image

But once I've done that, I can't see the code in the other lines, so I need to scroll back horizontally

image

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
```diff
@@ -278,11 +279,13 @@ private extension SupportTableViewController {
         let title: String
         let showIndicator: Bool
         let action: ImmuTableAction?
+        let accessibilityIdentifier: String?
 
-        init(title: String, action: @escaping ImmuTableAction, showIndicator: Bool = false) {
+        init(title: String, action: @escaping ImmuTableAction, showIndicator: Bool = false, accessibilityIdentifier: String? = nil) {
```
mokagio, line 285: Nice use of the default parameter there 👍

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
```diff
@@ -0,0 +1,45 @@
+import UITestsFoundation
+import XCTest
+
+// This screen object is for the Support section. In the app, it's a modal we can get to from Me > Help & Support, or, when logged out, from Prologue > tap either continue button > Help.
```
mokagio: Nitpick / Tip. You could make this a documentation comment with `///` instead of `//` and move it above the `SupportScreen` definition. Doing so would allow Xcode to generate an helper tooltip with that information.

This format:

image

Then becomes this in the autocompletion window (see bottom of the screenshot)

image

And this if you `Option Click` on the type name:

image

It's a minor thing, but in the optic of enabling developer to write these tests for themselves as part of the standard feature work, having documentation like this handy is an advantage.

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
```diff
@@ -0,0 +1,45 @@
...
+class SupportScreen: BaseScreen {
+    let closeButton: XCUIElement
+    let helpCenter: XCUIElement
+    let contact: XCUIElement
+    let myTickets: XCUIElement
+    let contactEmail: XCUIElement
+    let activityLogs: XCUIElement
+
+    init() {
+        let app = XCUIApplication()
```
mokagio, line 25: Nitpick.

Instantiating `XCUIApplication` this way, in the guts of the type, binds the resulting instance to the value returned by `XCUIApplication`. 

When called with no arguments `XCUIApplication()` generates a proxy class for the app the test target is configured to test.

Screen Shot 2021-07-12 at 4 07 39 pm

But, nothing stops us from instantiating a different app using `XCUIApplication(bundleIdentifier:)`. If by any chance we wanted to create this screen object around an app other than the one configured in the target, we wouldn't be able to do it.

Granted, it's unlikely we'd want to do that, because, by their very nature, screen objects are application specific (hence the nitpick). Still, there's a small tweak that we can do to this code that leaves the door open for that use case and makes the type follow the Dependency Injection best practice.

```suggestion
    init(app: XCUIApplication = XCUIApplication()) {
```

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
```diff
@@ -0,0 +1,45 @@
...
+        //Check that each item is loaded:
+        XCTAssert(helpCenter.exists)
+        XCTAssert(contact.exists)
+        XCTAssert(myTickets.exists)
+        XCTAssert(contactEmail.exists)
+        XCTAssert(activityLogs.exists)
+    }
+
+    static func isLoaded() -> Bool {
+        return XCUIApplication().buttons[ElementStringIDs.activityLogs].exists
```
mokagio, line 48: Same nitpick as above:

```suggestion
    static func isLoaded(in app: XCUIApplication = XCUIApplication()) -> Bool {
        return app.buttons[ElementStringIDs.activityLogs].exists
```

WordPress/WordPressUITests/Tests/SignupTests.swift
```diff
@@ -26,4 +26,15 @@ class SignupTests: XCTestCase {
...
+
+    // Test Support Section Loads
+    // From Prologue > continue, tap "help" and make sure Support Screen loads
+    func testSupportScreenLoads() {
+        let supportScreen = PrologueScreen().selectContinue().selectHelp()
+
+        XCTAssert(supportScreen.isLoaded())
+
+        //close SupportScreen modal before teartown
+        SupportScreen().closeButton.tap()
```
mokagio: Have you considered using the screen object instance already available?

```suggestion
        supportScreen.closeButton.tap()
```

WordPress/WordPressUITests/Tests/SignupTests.swift
```diff
@@ -26,4 +26,15 @@ class SignupTests: XCTestCase {
...
+    }
```
mokagio: In the PR description you ask:

> One structure question I still have: is SignupTests a good place for this test?
> I put it there because:
> 
> - LoginTests is getting pretty busy
> - It doesn't require logging in (it tests Help from a non-logged-in state)
>   But I'm not sure either reason justifies putting it in SignupTests. Should it have it's own file or be elsewhere?

I like the idea of having it in its own file, because this is sort of part of both the logic and signup flow, but also part of neither. I'm not sure how I would call the file itself, but it doesn't matter _too_ much. Maybe `SupportScreenTests` or `UnauthenticatedSupportScreenTests`? 🤷‍♂️

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -926,6 +926,9 @@
 		638E2F3D4DFE9E93AF660CED /* Pods_WordPressAllTimeWidget.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 0075A07B739C4EDB5EC613B8 /* Pods_WordPressAllTimeWidget.framework */; };
 		660A036E6EE9D353F2712081 /* Pods_WordPressNotificationServiceExtension.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E2A297CCB53FCF6851D79331 /* Pods_WordPressNotificationServiceExtension.framework */; };
 		6867C0633E597372235CB5E0 /* Pods_WordPressDraftActionExtension.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F7E3CC306AECBBCB71D2E19C /* Pods_WordPressDraftActionExtension.framework */; };
+		6EC71EC32689A67400ACC0A0 /* SupportScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6EC71EC22689A67400ACC0A0 /* SupportScreen.swift */; };
+		6EC71F8C2689A7A900ACC0A0 /* SupportScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6EC71EC22689A67400ACC0A0 /* SupportScreen.swift */; };
+		6EC71FA02689A7AD00ACC0A0 /* SupportScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6EC71EC22689A67400ACC0A0 /* SupportScreen.swift */; };
```
mokagio: This apparent duplication (triplication) is because the screen object is, _correctly_, part of the UI tests target and the two screenshot generation ones.

image

FYI: The next step for the UITestsFoundation setup is to move all the shared screen objects in there. That will make it easier to share screen objects across targets, because once a test target adds UITestsFoundation as a library dependency (which all three already do) it automatically gets access to the screen objects in the library. _Stay tuned_ 🤓

Podfile
```diff
@@ -211,7 +211,7 @@ abstract_target 'Apps' do
 
     pod 'Gridicons', '~> 1.1.0'
 
-    pod 'WordPressAuthenticator', '~> 1.38.0'
+    pod 'WordPressAuthenticator', '~> 1.40.0-beta'
```
mokagio: This `Podfile` change is missing its counterpart `Podfile.lock` update.

Every time we change the `Podfile`, we should also run `bundle exec pod install`. 

This is also what [the failure in CI is about](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-iOS/22840/workflows/51587738-2c04-4ef6-aa81-2ca6a52b0fd2/jobs/52609).

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
```diff
@@ -0,0 +1,45 @@
...
+        activityLogs = app.cells[ElementStringIDs.activityLogs]
+
+        super.init(element: activityLogs)
+
+        //Check that each item is loaded:
+        XCTAssert(helpCenter.exists)
+        XCTAssert(contact.exists)
+        XCTAssert(myTickets.exists)
+        XCTAssert(contactEmail.exists)
+        XCTAssert(activityLogs.exists)
```
mokagio, line 40: Not really a suggestion, more a "did you know that", so no need to update this code.

Did you know that you can create a list of objects and run the same code on each, to avoid writing repetitive code like this?

```swift
[helpCenter, contact, myTickets, contactEmail, activityLogs].forEach { XCTAssert($0.exists) }
```

mokagio: Actually! After having said that I realized that, in this case, it's a bad suggestion 😳 🤦‍♂️ .

Explicitly writing each assertion means that if one fails, Xcode will highlight it. With the `forEach`, we'd get the failure in `XCTAssert($0.exist)` but that doesn't tell us which of the element didn't exist!

mokagio (commented):
Nice job @thehenrybyrd 🙌 

Left mainly nitpicks. I run the tests locally and it passed.

You'll need to address [the missing `Podfile.lock` update failure](https://github.com/wordpress-mobile/WordPress-iOS/pull/16837#discussion_r667656818) before being able to merge. Ping me when you do and I'll have a second/final look.

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
thehenrybyrd: More readable, thanks!

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
thehenrybyrd: Thanks! Credit goes to @leandroalonso who showed me how to do this:
p1625472413270600-slack-C011BKNU1V5

Looking back over this, I realize that `HelpRow` has an `init()` with those default parameters, while `SupportEmailRow` below (line 302) only has the `let`s and `configureCell()`.
Do you see a clear reason `SupportEmailRow` doesn't use an `init()`?

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
thehenrybyrd: Oooh I love this, great tip! Implemented, thanks!

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
thehenrybyrd: Oh yeah, I remember learning .forEach in JS! I think you're right that it's best to leave them separated in this case, but this is a good reminder to me.

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
mokagio: > Do you see a clear reason `SupportEmailRow` doesn't use an `init()`?

I think your question was more about why `SupportEmailRow` doesn't have an `init` method in which it defines some default values, like `HelpRow` does. But, just in case, I'd like to point out that `SupportEmailRow` _has_ an `init` method (although, it's not `init()` which technically speaking would mean an init without arguments, but would be written as `init(title:,value:,accessibilityHint:,action:)`)

The reason it's not explicitly defined is that the Swift compiler synthesizes initializer for `struct`s if no initializer is defined. This is called [memberwise initializer](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html)

> Structure types automatically receive a memberwise initializer if they don’t define any of their own custom initializers

Screen Shot 2021-07-19 at 4 11 23 pm

That's a super useful feature that makes writing and using straightforward code seamless. 🤓 

I don't have an answer to why `SupportEmailRow` doesn't define some defaults, while `HelpRow` does. But maybe the the code history has something say about it ☝️ 

All our code is stored in Git, a version control system (VCS). Developer track discrete changes to the code in the form of commits, and add a title to each commit. There are known best practices on how to write good commit messages, that can help future readers learn about the rational behind a change. 

Git has an ill named command called `git blame` that shows the commit that last changed each line of code. Xcode expose this functionality through the "Authors" UI:

Screen Shot 2021-07-19 at 4 24 44 pm

I find, though, that the tool that offers the best ergonomics to go spelunking through the code's history is GitHub. We can open `SupportTableViewController.swift` in GitHub and select "Blame":

Screen Shot 2021-07-19 at 4 26 53 pm

The blame view shows the commit that last changed each line, as well as a button to scroll back through the previous versions:

Screen Shot 2021-07-19 at 4 29 48 pm

Using that, we can learn that `HelpRow` was added [3+ years ago](https://github.com/wordpress-mobile/WordPress-iOS/commit/6f23bc34ef52ebe1064c1a1a0edcf38c2a8b7c5f#diff-49dc6dec4a731852be5a9f38c831ff0ed74c648e5a412aafd19404fcbcb0bbcaR133-R149) and already had that explicit init. There's no explanation in the commit message as of why it was written like that, but we could ask the author, @ScoutHarris if she remembers 😄 

The only thing I notice is that the `action` property is defined as wrapped in an `Optional`, [as required by the `ImmuTableRow` protocol](https://github.com/wordpress-mobile/WordPress-iOS/blob/17488be698b505545d953c02c20476fb5555c92d/WordPress/Classes/Utility/ImmuTable.swift#L143), but it is not in the `init`. My guess without any extra context would be that @ScoutHarris didn't want to allow callers of `HelpRow` to app a `nil` (i.e. `Optional<ImmuTableAction>.none`) value.

I decided to look into `HelpRow` instead of `SupportEmailRow` because, since the memberwise initializer was introduced in an early version of Swift, it's usually always used instead of explicitly defining one (why doing more work than necessary, right?) so I was more curious about why `HelpRow` had a custom one.

There's a lot more to say about Swift initialization and searching information in Git, but I think I've pontificated long enough for as single comment 😅 

---

Further reading on how to write good commit messages and why it matters:

- https://thoughtbot.com/blog/5-useful-tips-for-a-better-commit-message
- https://mislav.net/2014/02/hidden-documentation
- https://mokacoding.com/blog/your-git-log-should-tell-a-story/
- https://mokacoding.com/blog/better-merging-for-github-pull-requests/

WordPress/WordPress.xcodeproj/project.pbxproj
thehenrybyrd: Gotcha, that makes sense!
Oooh, that should be exciting!

WordPress/WordPressUITests/Tests/SignupTests.swift
thehenrybyrd: That makes sense to me, and leave room for creating further UI tests of various support screens if desired. I think `SupportScreenTests` makes most sense, so I'll go with that. :)

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
thehenrybyrd: > I think your question was more about why SupportEmailRow doesn't have an init method in which it defines some default values, like HelpRow does.

Yep, completely right. I wasn't aware of memberwise initializers, so thanks for pointing that out!

TIL about that "Authors" view as well, though I've used the "history" in the inspectors sidebar to help me find that info in the past. I totally agree that GitHub's _blame_ is way easier to use.

> I decided to look into HelpRow instead of SupportEmailRow because, since the memberwise initializer was introduced in an early version of Swift, it's usually always used instead of explicitly defining one (why doing more work than necessary, right?) so I was more curious about why HelpRow had a custom one.

Ok, that make sense. If I understand, the memberwise initializer will use all the properties defined in the struct (so in this case, `title`, `value`, `accessibilityHint`, `action`, and now `accessibilityIdentifier`). So, unless you have a reason to define custom initializer, you can generally rely on memberwise initialization without any trouble.
Am I getting that right?

Thanks!

Podfile
thehenrybyrd: > Every time we change the `Podfile`, we should also run `bundle exec pod install`.

Ah, good to know! So since I changed the `Podfile`, when I run `bundle exec pod install` it checks to see what versions of things are in the `Podfile`, and updates `podfile.lock` according to that. Is that right?

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
ScoutHarris: > HelpRow was added 3+ years ago and already had that explicit init. 

I _think_ I added an explicit `init` so `showIndicator` could have a default value. Only one row has the indicator, so having it passed for the others was superfluous. 

> The only thing I notice is that the action property is defined as wrapped in an Optional, as required by the ImmuTableRow protocol, but it is not in the init.

This looks like an oversight. Although it does work without `action` being explicitly included in the `init` params, which is probably why I didn't notice.

mokagio: Thank you for adding more context @ScoutHarris 😄 🙏 

> I _think_ I added an explicit `init` so `showIndicator` could have a default value. Only one row has the indicator, so having it passed for the others was superfluous.

Oh I see. I didn't notice that because `showIndicator` was added later on.

mokagio: @thehenrybyrd 

> So, unless you have a reason to define custom initializer, you can generally rely on memberwise initialization without any trouble.
> Am I getting that right?

Yes.

Podfile
mokagio: Yes.

In the `Podfile`, you can define dependency resolution rules, such as `> 1.2`, which will install any version greater than 1.2.

In the `Podfile.lock`, CocoaPods writes the _resolved_ version, the one it actually ended up installing.

When we run `pod install`, CocoaPods does something along the lines of:

1. Checks to see if the checksum is the same between `Podfile` and `Podfile.lock`
2. If it is the same (no changes have been made to the `Podfile`), downloads and installs all the versions as specified by `Podfile.lock`
3. If it is not the same (something changed in the `Podfile`), it resolves the dependencies based on what's defined in the `Podfile`, writes the result in the `Podfile.lock`, then downloads and installs them

Dependency resolution is useful in cases such the following. A project uses three libraries: A, B, and C. Library A uses library B at version 1.1; library C also uses library B, but at version 1.2. Depending on the dependency resolution rules configured in the `Podfile` and in each library, the project might end up installing library B version 1.2 or fail the setup and tell the user they need to update library A to use library B at version 1.2, too, to ensure compatibility.

_Dependency management and resolution is a complex topic._

thehenrybyrd: > Dependency management and resolution is a complex topic. 

💯 So it seems. Thanks!

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
thehenrybyrd: Thanks for bringing this up! Given that there doesn't seem to be a use case for instantiating a different app, I'll leave it as-is for now. However, if we decide later that we'd be able to use this in shared UI test infrastructure and need to change it, I'll be happy to make the changes.

WordPress/WordPressUITests/Screens/Me/SupportScreen.swift
thehenrybyrd: Leaving as-is for now per prior comment, but happy to change if we decide it's necessary later.

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
```diff
@@ -0,0 +1,27 @@
...
+    override func tearDown() {
+        takeScreenshotOfFailedTest()
+        LoginFlow.logoutIfNeeded()
+        super.tearDown()
+    }
+
+    // Test Support Section Loads
+    // From Prologue > continue, tap "help" and make sure Support Screen loads
+    func testSupportScreenLoads() {
+        let supportScreen = PrologueScreen().selectContinue().selectHelp()
```
mokagio, line 20: A key advantage of the screen object pattern is that we can hide all the implementation details  behind the public API of the screen object. In this case, we could move the "continue and tap help" into the screen object and simplify this into:

```suggestion
        let supportScreen = PrologueScreen().loadSupportScreen()
```

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
```diff
@@ -0,0 +1,27 @@
...
+
+        XCTAssert(supportScreen.isLoaded())
+
+        //close SupportScreen modal before teartown
+        supportScreen.closeButton.tap()
```
mokagio: Similarly as above, we don't need the UI test itself to know how to dismiss the support screen, we can move that knowledge in the screen object. Doing so makes the test closer to natural language, and the logic to dismiss the screen reusable.

```suggestion
        supportScreen.dismiss()
```

Moving the logic into the screen object also helps removing comments. It's tempting to use comment to explain what the code does, but it's best to make the code itself clearer, either with better design or naming.

From ["Clean Code" by Uncle Bob](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882):

Screen Shot 2019-10-31 at 9 23 28 am

That's not to say that one should never leave comments in the code. There are many instances in which comments are necessary and add lots of value. Those comment, though, are usually more about the context in which the code sits, the author's rationale for choosing that code, or suggestions for future readers to make improvements.

IMHO, this comment doesn't add much information. It could be removed without impacting a future reader's ability to understand and modify the test.

mokagio (approved):
Run the newly extracted test locally and it worked 👍 

I also triggered the optional tests in CI. We'll need to add some automation around it, but I think it's necessary to run the UI tests in CI if the UI tests suite is modified.

You don't need to address my suggestion right now. We can save them for when we'll get to migrate the `PrologueScreen` to use `ScreenObject`.

I'm going to approve the PR under the assumption the tests will work in CI, too.

mokagio:
Pushing this PR to the next milestone, 18.0, because I'm running the code-freeze today and, even though this is approved, it has merge conflicts that need to be resolved.

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
thehenrybyrd: Makes sense to me! I moved closeButton.tap() into a dismiss() function on SupportScreen.

As for the comment, I wanted to give context that we need this dismiss function in the test itself because tearDown() can't deal with modals currently. Most of the UI tests I'm familiar with don't end with a modal, so it's unusual to have another function after the XCTAssert. I've edited the comment to make that more clear.

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
thehenrybyrd: I think I disagree in this particular case it makes it less readable to me. If I'm following along on my device, `loadSupportScreen()`doesn't tell me anything. We have to tap "continue" before we have access to the "help" button, so it seems better as two separate steps. That way, reading the test makes clear what steps are being followed. 
Happy to discuss more when we migrate though!

mokagio: I'm glad you disagree. Thank you 😄 

> That way, reading the test makes clear what steps are being followed.

That's a totally legit point of view. It occurred to me, too, while making my comment. I guess I'm really excited about Screen Objects right now and so I'm leaning towards encapsulating as much as possible into them. I can't wait to have "in the trenches" experience with the pattern in the context of this and our other apps, so we can see whether "as much encapsulation as possible" is a good rule of thumb, or there is space for this syntax that is more sequential and exposes the high level action a user would have to take in the app 👀

PR #16839, opened 2021-07-09
Merge release 17.7 into trunk

mokagio:
Merges the 17.7 changes into `trunk`.

Leaving this as a draft till we actually release it. Still, I'd appreciate a review anyway.

Thanks 🙇‍♂️ .

PR #16840, opened 2021-07-12
Merge release 17.7 into trunk

mokagio:
This supersedes #16839, which is "out-of-date with the base branch" because its a PR of `release/17.7` into `trunk`, but `release/17.7` came off `develop`.

I wanted to be neat and avoid an intermediate branch, and this is what I get in return...

~~I'm going to admin-merge this because it's the same as #16839, which has already been approved.~~ I wanted to admin-merge this, but I cannot because all branch protection rules are enforced for admin, too, which is fair:

image


oguzkocer (approved):
It looks like this matches `17.7` tag*, so we should be good 👍 

*: `git diff` shows 2 changes to the metadata files, but I am guessing that's a localization issue with git (I've had that happen a couple times) and metadata is not particularly important after it's already released, so 🤷

PR #16841, opened 2021-07-12
Update AppCenter SDK from version 4.1.1 to 4.2.0

mokagio:
This is just a routine update to stay up to date.

The release notes don't seem to point to anything that would affect us,
but they do mention this version should not produce warnings in Xcode 12.5

See:
https://github.com/microsoft/appcenter-sdk-apple/releases/tag/4.2.0

**To test:** I don't think there's anything to test here. I launched an installable build just to verify the compilation in CI. I also installed both resulting builds for WordPress and Jetpack and they worked.

## Regression Notes

There is one thing that comes to mind: the incremental update for Alpha/Internal builds.
Unfortunately, the only way to test this would be to ship one build and then another one on top of it to verify the SDK picks up the new version and suggests the update.
I think that's all responsibility of the SDK, though, even if we discovered an issue, we could only report it to Microsoft.
Moreover, this version has been out for almost a month, an issue of that magnitude would have been noticed and addressed by now.

N.A.

N.A.

---

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -173,8 +173,12 @@ abstract_target 'Apps' do
     ##
     pod 'Charts', '~> 3.2.2'
     pod 'Gifu', '3.2.0'
-    pod 'AppCenter', '4.1.1', :configurations => ['Release-Internal', 'Release-Alpha']
-    pod 'AppCenter/Distribute', '4.1.1', :configurations => ['Release-Internal', 'Release-Alpha']
+
+    app_center_version = '~> 4.1'
+    app_center_configurations = %w[Release-Internal Release-Alpha]
+    pod 'AppCenter', app_center_version, configurations: app_center_configurations
+    pod 'AppCenter/Distribute', app_center_version, configurations: app_center_configurations
+
     pod 'MRProgress', '0.8.3'
     pod 'Starscream', '3.0.6'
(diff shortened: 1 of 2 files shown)


Podfile
```diff
@@ -173,8 +173,12 @@ abstract_target 'Apps' do
     ##
     pod 'Charts', '~> 3.2.2'
     pod 'Gifu', '3.2.0'
-    pod 'AppCenter', '4.1.1', :configurations => ['Release-Internal', 'Release-Alpha']
-    pod 'AppCenter/Distribute', '4.1.1', :configurations => ['Release-Internal', 'Release-Alpha']
+
+    app_center_version = '~> 4.1'
```
mokagio, line 177: I think it makes sense to use this because we want to be up-to-date with new features and fixes, but want to be intentional about updating if there's going to be a breaking change.

Feel free to push back though. For example, one could argue this dependency is one we should update only when a new feature we need demands it.

AliSoftware (approved):
I agree with the standard use of `~> x.y` here, semver ftw!

:shipit:

mokagio:
FYI @wordpress-mobile/penguin-team. I'm not sure how much you use the builds from AppCenter and, as per PR description, this should not affect the integration. Still, go to know about it 😄

PR #16842, opened 2021-07-12
Merge release 17.8 code freeze into develop

mokagio:
Code freeze steps merge, including a tweak to the `Fastfile` to skip updating `Deliverfile` after the changes in d3fe103af8.


Podfile
```diff
@@ -47,7 +47,7 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.37.0-beta'
+    pod 'WordPressKit', '~> 4.37.0'
```
mokagio, line 50: See also https://github.com/wordpress-mobile/WordPressKit-iOS/pull/418

mokagio (commented):

Podfile
mokagio: Something I forgot to mention: There is a new version of WordPressAuthenticator, [1.40.0-beta.1](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/releases/tag/1.40.0-beta.1), but it hasn't been merged into the project's main branch, so I haven't updated.

@thehenrybyrd I think you published that, right? Was not merging into `develop` intentional or is it leftover work? Let me know if you need help with it.

We don't really need to bump in the release branch, though, as [the change only adds an accessibility ID](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/compare/1.39.0...1.40.0-beta.1) that's useful for automation but not in production

AliSoftware: I think this should have been a 5.0.0 not a 4.37.0 for WordPressKit, because this one is a breaking change (API change by renaming some types `State` -> `WPState` and `Country` -> `WPCountry`). See discussion in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/418

oguzkocer (approved):
I see that there are 2 open discussions about version numbers, so I am not merging the PR. @jkmassel you may want to take a quick look since @mokagio is afk tomorrow.

Podfile
jkmassel: Technically, this should've been done, but I'd rather not block this merge on that – if needed we could re-cut the same release as `5.0.0`.

jkmassel (approved):
Approving so we can land this now.

We can deal with the WPKit stuff separately

Podfile
AliSoftware: Yep, agreed.

If we do release `5.0.0` on the same commit as `4.37.0` (+ version bump in podspec) to avoid client projects from breaking, we should then yank `4.37.0`, otherwise client project will still get it with `~> 4.x` and will still break.

mokagio: Just a note that I've seen this message and plan to look into it shortly.

PR #16843, opened 2021-07-12
Add List table view cell component

dvdchr:
Refs #16741

## Summary
This introduces the table view cell for the List component. Some notes regarding this change:

- The cells are still detached from the application flow, and individual adjustments to Comments list and Notifications list will be made in separate PRs.
- Indicator views (the little colored dot on the left) has a default color of `.clear`. This is because both Notifications list assigns blue color for unread state, while Comments list assigns yellow color for unmoderated state. Therefore, it's best for the users of the List component to manage their own indicator colors (will be added in later PRs).
- The number of lines for `titleLabel` is increased to 3 when snippet text is empty or nil.

Here's how the cells look in Notifications screen (note that `ListTableHeaderView` is also used here):

Light 🌝  | Dark 🌚 
-- | --
16741_light_noSnippet | 16741_dark_noSnippet

## To Test
Since the table view cell is still detached from application flow, you'd need to manually register the table view cells in either Comments list (`CommentsViewController`) or Notifications list (`NotificationsViewController`).

## Regression Notes
1. Potential unintended areas of impact
None. The view is detached from application flow.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Manually tested the component in `NotificationsViewController`.

3. What automated tests I added (or what prevented me from doing so)
None.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #16741, opened 2021-06-24
Comment/Notification lists: update table cell


ScoutHarris:
Create new common table cell, and use in both Comments and Notifications lists.

- [x] Create new view matching the design. (#16843)
- [x] Customizable to show pending indicator or unread indicator. (#16843)
- [x] Customizable to show API's Notification title or custom Comment title. (#16843)
- [x] #16852
- [x] #16851

Research notes:
- Should be able to copy and modify `CommentsTableViewCell` for both purposes.
- Remove `WPTableViewCell` inheritance.
- Consider using WKWebView to render content (HTML).

aerych:
> Consider using WKWebView to render content (HTML).

This actually refers to the cells used in the threaded list and the view in the comment detail screen, not the "summary" cells used by the comment lists.

PR #16856, opened 2021-07-13
Implement List component for Comments

### LINKED ISSUE #16851: Use List component in Comments
Update Comments list to use List table components (`ListTableHeaderView` and `ListTableViewCell`) behind a feature flag.

dvdchr:
Resolves #16851
Refs #16740, #16741 
Depends on #16843

## Summary
This PR introduces the new List component to Comments list, with a feature flag. Here's a summary of the changes:
- A new feature flag, `unifiedCommentsAndNotificationsList` is added. The default behavior is `true` for **local developer and integration builds**. This is to ease testing for developers and designers, without having to make any manual changes.
- The relative date section identifier for `Comment` is copied from `Notifications+Interface`. Originally, I wanted to create a protocol (`RelativeDateSectionIdentifiable`) with default implementation, but since `Comment` and `CommentsViewController` are still in Objective-C, protocol extensions don't work. Hence the copy paste solution for now. 😢 
- Since `Comment` can pull avatar image either from URL or from gravatar email, I changed the image download approach in `ListTableViewCell` to be similar to `CommentsTableViewCell` – The `imageURL` prop is removed, and two methods are introduced instead: `configureImage(with url:)` that accepts a URL and `configureImageWithGravatarEmail`, that accepts a String.
- The table view footer in `CommentsViewController` is also updated from `CGRectZero` to a UIView with 1pt height. This is to get rid of the separator for the last row _on the last section_. Previously, the last row in the last section was still showing a separator line.

#### Scenario 1: With disabled feature flag, Comments list should be displayed with the old design
- Go to My Site > Comments.
- Verify that the comments list is displayed with the old design.
- Verify that the comments are sectioned by individual date, instead of relative date range.

#### Scenario 2: With enabled feature flag, Comments list should be displayed with the new design
- Go back to My Site, click your Avatar, and go to App Settings.
- Select the Debug menu, and scroll down until you find `Unified Comments and Notifications List` feature flag.
- Make sure that the feature flag is **enabled**.
- Go back to My Site > Comments.
- Verify that the comments list is displayed with the new design.
- Verify that the comments are sectioned correctly.

## Regression Notes
Comments list not showing correctly when feature flag is turned off.

Ensured that the new and old list are displayed correctly when the feature flag is toggled.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
@@ -178,10 +178,16 @@ static NSString *RestorableFilterIndexKey = @"restorableFilterIndexKey";
         return nil;
     }
 
+    // fetch the section information
+    id<NSFetchedResultsSectionInfo> sectionInfo = [self.tableViewHandler.resultsController.sections objectAtIndex:section];
+    if (!sectionInfo) {
+        return nil;
+    }
+
     ListTableHeaderView *headerView = (ListTableHeaderView *)[self.tableView dequeueReusableHeaderFooterViewWithIdentifier:ListTableHeaderView.reuseIdentifier];
+
     if (headerView) {
-        // TODO: Enable relative date section identifier for Comment objects.
-        headerView.title = @"Section Header";
+        headerView.title = [Comment descriptionForSectionIdentifier:sectionInfo.name];
     }
 
     return headerView;
@@ -397,7 +403,7 @@ static NSString *RestorableFilterIndexKey = @"restorableFilterIndexKey";
 
 - (NSString *)sectionNameKeyPath
 {
-    return @"sectionIdentifier";
+    return [self usesUnifiedList] ? @"relativeDateSectionIdentifier" : @"sectionIdentifier";
 }
 
 #pragma mark - Predicate Wrangling
--- a/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
@@ -131,18 +131,20 @@ static NSString *RestorableFilterIndexKey = @"restorableFilterIndexKey";
 {
     self.tableView.accessibilityIdentifier  = @"Comments Table";
     [WPStyleGuide configureColorsForView:self.view andTableView:self.tableView];
-    
-    // Register the cells
-    NSString *nibName = [CommentsTableViewCell classNameWithoutNamespaces];
-    UINib *nibInstance = [UINib nibWithNibName:nibName bundle:[NSBundle mainBundle]];
-    [self.tableView registerNib:nibInstance forCellReuseIdentifier:CommentsTableViewCell.reuseIdentifier];
 
-    // Register unified list components
-    UINib *listCellNibInstance = [UINib nibWithNibName:[ListTableViewCell classNameWithoutNamespaces] bundle:[NSBundle mainBundle]];
-    [self.tableView registerNib:listCellNibInstance forCellReuseIdentifier:ListTableViewCell.reuseIdentifier];
+    if ([self usesUnifiedList]) {
+        // Register unified list components
+        UINib *listCellNibInstance = [UINib nibWithNibName:[ListTableViewCell classNameWithoutNamespaces] bundle:[NSBundle mainBundle]];
+        [self.tableView registerNib:listCellNibInstance forCellReuseIdentifier:ListTableViewCell.reuseIdentifier];
 
-    UINib *listHeaderNibInstance = [UINib nibWithNibName:[ListTableHeaderView classNameWithoutNamespaces] bundle:[NSBundle mainBundle]];
-    [self.tableView registerNib:listHeaderNibInstance forHeaderFooterViewReuseIdentifier:ListTableHeaderView.reuseIdentifier];
+        UINib *listHeaderNibInstance = [UINib nibWithNibName:[ListTableHeaderView classNameWithoutNamespaces] bundle:[NSBundle mainBundle]];
+        [self.tableView registerNib:listHeaderNibInstance forHeaderFooterViewReuseIdentifier:ListTableHeaderView.reuseIdentifier];
+    } else {
+        // Register the cells
+        NSString *nibName = [CommentsTableViewCell classNameWithoutNamespaces];
+        UINib *nibInstance = [UINib nibWithNibName:nibName bundle:[NSBundle mainBundle]];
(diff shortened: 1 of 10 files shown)


WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -131,12 +136,19 @@ - (void)configureTableView
     NSString *nibName = [CommentsTableViewCell classNameWithoutNamespaces];
     UINib *nibInstance = [UINib nibWithNibName:nibName bundle:[NSBundle mainBundle]];
     [self.tableView registerNib:nibInstance forCellReuseIdentifier:CommentsTableViewCell.reuseIdentifier];
+
+    // Register unified list components
```
ScoutHarris: Can the cell registrations be based on `usesUnifiedList`? At the very least, it will be clear later when the feature flag is removed what needs to be removed/kept.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -131,12 +136,19 @@ - (void)configureTableView
...
+
+    UINib *listHeaderNibInstance = [UINib nibWithNibName:[ListTableHeaderView classNameWithoutNamespaces] bundle:[NSBundle mainBundle]];
+    [self.tableView registerNib:listHeaderNibInstance forHeaderFooterViewReuseIdentifier:ListTableHeaderView.reuseIdentifier];
 }
 
 - (void)configureTableViewFooter
 {
     // Hide the cellSeparators when the table is empty
-    self.tableView.tableFooterView = [[UIView alloc] initWithFrame:CGRectZero];
+    self.tableView.tableFooterView = [[UIView alloc] initWithFrame:CGRectMake(0, 0, self.tableView.frame.size.width, 1)];
```
ScoutHarris, line 153: 👍

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
```diff
@@ -165,8 +206,15 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
 
 - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
 {
+    if ([self usesUnifiedList]) {
+        ListTableViewCell *listCell = (ListTableViewCell *)[tableView dequeueReusableCellWithIdentifier:ListTableViewCell.reuseIdentifier
```
ScoutHarris, line 212: There should probably be a check here to ensure `listCell` exists (similar to the existing `if (!cell)` logic).

WordPress/Classes/ViewRelated/Views/List/ListTableHeaderView.swift
```diff
@@ -12,6 +12,9 @@ class ListTableHeaderView: UITableViewHeaderFooterView, NibReusable {
 
     // MARK: Properties
 
+    @objc static let reuseIdentifier = defaultReuseID
```
ScoutHarris, line 17: Nitpick/request - can a comment be added here indicating `reuseIdentifier` can be removed when the VCs are converted to Swift? It'd be nice to _eventually_ use `defaultReuseID` directly.

WordPress/Classes/ViewRelated/Views/List/ListTableViewCell.swift
```diff
@@ -14,6 +14,9 @@ class ListTableViewCell: UITableViewCell, NibReusable {
 
     // MARK: Properties
 
+    @objc static let reuseIdentifier = defaultReuseID
```
ScoutHarris, line 19: Same nitpick/request as above - can a comment be added here indicating `reuseIdentifier` can be removed when the VCs are converted to Swift? It'd be nice to _eventually_ use `defaultReuseID` directly.

ScoutHarris (changes requested):

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
dvdchr: That makes sense! Will adjust this soon.

WordPress/Classes/ViewRelated/Views/List/ListTableHeaderView.swift
dvdchr: Noted! Will do.

WordPress/Classes/ViewRelated/Views/List/ListTableViewCell.swift
dvdchr: Yup! Agreed. 😄  Will add docs here too.

WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
dvdchr: Since I'm using `dequeueReusableCellWithIdentifier:forIndexPath:`, the method guarantees that it returns a non-null instance of the cell, as long as it's properly registered:

```objc
// newer dequeue method guarantees a cell is returned and resized properly, assuming identifier is registered
- (__kindof UITableViewCell *)dequeueReusableCellWithIdentifier:(NSString *)identifier forIndexPath:(NSIndexPath *)indexPath API_AVAILABLE(ios(6.0));
```

I think it's unnecessary to add additional `if (!cell)` checks, but I'd like to double check this again with you 😄 . If you think we still need the check, I'll add it! 🙂

ScoutHarris: > as long as it's properly registered

Ah, this part I did not know. Reread the documentation for that method, and now I do. 😄 I was distracted by the old code. Sorry for the bother!

ScoutHarris (approved):
Thanks for the changes! The list looks great!

I just spotted one thing - when showing the new UI, these appear in the logs:
```
2021-07-14 11:39:50.309944-0600 WordPress[17754:193539] [TableView] Changing the background color of UITableViewHeaderFooterView is not supported. Use the background view configuration instead.
2021-07-14 11:39:50.315172-0600 WordPress[17754:193539] [TableView] Changing the background color of UITableViewHeaderFooterView is not supported. Use the background view configuration instead.
```

Since that is unrelated to this PR, it can be fixed separately. So GTG here.

:shipit:

dvdchr:
On second thought, I think I'll set the feature flag to `false` right now, so folks don't suddenly experience the change. I'll re-enable to local developer + integration builds again in the PR for Notifications.

> I just spotted one thing - when showing the new UI, these appear in the logs:

Good catch, and what a pair of 🦅  👀  ! ~I'll try to sneak in the fix here.~ Done! 

Plus, I added checks for table header view instantiation, since the method is nullable.

PR #16877, opened 2021-07-16
Fixes a crash when the value returned by Blog.version is not a string

emilylaguna:
Sentry crash: https://sentry.io/organizations/a8c/issues/2511522132/events/?project=1438083

#### Verify the crash
1. Create a self hosted site that is not connected to Jetpack
2. Go to the My Site view
3. Tap the dropdown to view all your sites
4. Tap the + button to add a site
5. Tap Add Self Hosted
6. Add your new site
7. If the app doesn't crash: 

#### Force the crash
1. Open Blog.m
2. Go to line: 449
3. Change it to `return @5.5` 
4. This will force return an NSNumber
5. Relaunch the app
6. It should crash

#### Verify the fix
1. Remove the forced crash
2. Go to Blog.m
3. Line 902
4. Add a line above it for:

```
if([name isEqualToString:@"software_version"]){
    return @5.5;
}
```
This force returns an NSNumber for the software version
5. Relaunch the app
6. It should **not** crash

### Regression Notes
1. Potential unintended areas of impact
- The BlogService calls `[[blog version] floatValue];` since NSString implements this method this will not cause an issue. 
- `hasRequiredWordPressVersion` in Blog.m expects a string anyways

2. What I did to test those areas of impact (or what existing automated tests I relied on)
I did manual testing of that section.

3. What automated tests I added (or what prevented me from doing so)
None. 

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16882, opened 2021-07-19
Merge hotfix 17.7.1 into trunk

mokagio:
Includes #16877 by @emilylaguna.

Notice that I switched to a branch in the release toolkit and then reverted just so I could get the fix submitted quickly. Normally, I would have wanted to ship a new version of the toolkit and switch to it, but this is an hotfix so I prioritized speed.

The build has been approved for release and I started the rollout earlier today. This is ready to review and merge.


mokagio (commented):
- [x] Patch version bump in `.xcconfig`
- [x] No strings or release notes localization changes – Hotfixes bypass localization to ship as soon as possible

mokagio:
Ouch! I did it again, forgot the intermediate branch because the GitHub UI said it could merge without conflict. But there's a difference between being able to merge without conflict and the branch we want to merge being up to date with the base.

oguzkocer:
> Ouch! I did it again, forgot the intermediate branch because the GitHub UI said it could merge without conflict. But there's a difference between being able to merge without conflict and the branch we want to merge being up to date with the base.

@mokagio What is the difference? As in, why does it matter for this PR to be up to date with the base for it to be reviewed and merged?

mokagio:
It doesn't matter from the reviewing point of view, but it blocks the possibility to merge the PR, at least as far as I understand. This is the same thing that happened 

The PR branch needs to be up to date with trunk, but all these branches come from `develop`. To put them in sync with `trunk`, we'd need to merge `trunk` into them. Merging `trunk` into `release/17.7.1` would be inappropriate. Or maybe it isn't... we have the tag as the source of truth and the release branch gets deleted afterwards. Still, right now, while the PR into `release/17.8` hasn't been merged yet, that's definitely not appropriate.

🤔 maybe, instead of using the intermediate branch, I should just wait till the matching PR into `develop` has been merged next time.

Thanks for the food for thought @oguzkocer.

oguzkocer:
@mokagio As I understand it, you're looking to guard against the potential merge conflict if the base branch moves. In this case - in theory - that should never happen because `trunk` should only be updated during a new release which should also update current `develop` and any other release branches with the same commits.

Furthermore, I understand creating an intermediate branch just for the peace of mind and not having to deal with potential merge conflicts, but I don't think the base branch should be merged until it's necessary. That's to say - in my opinion - "up to date with the base" shouldn't be a concern until there is a merge conflict.

For me that comes down to communication through commits. Having that merge commit tells me that there was a conflict to be resolved and I should look into whether that was done correctly while I review it. In this case, I like to re-create the conflict in a temporary branch and resolve it and check that it matches what the author did. Hope that clarifies my point of view!

mokagio:
> "up to date with the base" shouldn't be a concern until there is a merge conflict.

That resonates. At the moment, we have a setting that enforces that requirement, for admin users, too.

Screen Shot 2021-07-20 at 3 20 51 pm

We'd need to disable that first. The rule is not active in `develop`. I also checked WordPress Android and WooCommerce iOS, just for a sample, and they don't have that requirement. I guess this was set as an experiment or even accidentally? What does the rest of @wordpress-mobile/owl-team think of the idea of removing the "Require branches to be up to date before merging" rule from this repo in `trunk`?

mokagio:
> For me that comes down to communication through commits. Having that merge commit tells me that there was a conflict to be resolved and I should look into whether that was done correctly while I review it.

I had never though of that in those terms, but I really like the "communication through commits" idea. I try to write useful and informative commit messages, for example, in a merge I usually leave the list of files that had conflicts and add notes on the rationale behind the resolution I applied.

That a merge commit declares intent is totally in line with that. In that case, not having an intermediate branch is definitely clearer. 

Thank you for sharing 😄

mokagio:
> What does the rest of @wordpress-mobile/owl-team think of the idea of removing the "Require branches to be up to date before merging" rule from this repo in `trunk`?

I haven't gotten a reply on this. I went ahead and disabled it to keep it consistent with other projects. Easy enough to revert this decision.

image

PR #16886, opened 2021-07-19
Fixed pluralization

thehenrybyrd:
Changed "is" to "are" since there are two style guideline documents.


pachlava (approved):
Sorry for the late review @thehenrybyrd !
I fully approve the change 🔥 

:shipit:

PR #16888, opened 2021-07-19
Fix BlockEditorSettingService tests

chipsnyder:
## Description 

Fixes the test errors in https://github.com/wordpress-mobile/WordPress-iOS/pull/16883 

The error was a result of the code going down an unexpected path because the software version wasn't being properly picked up by the blog anymore. 

## Regression Notes
None

N/A

This PR fixes some tests

## PR submission checklist:

### DIFF
--- a/WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
+++ b/WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
@@ -215,7 +215,7 @@ class BlockEditorSettingsServiceTests: XCTestCase {
         XCTAssertGreaterThan(colors.count, 0)
 
         guard let gradients = blockEditorSettings.gradients else {
-            XCTFail("Block editor colors should exist at this point")
+            XCTFail("Block editor gradients should exist at this point")
             return
         }
         XCTAssertGreaterThan(gradients.count, 0)
--- a/WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
+++ b/WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
@@ -20,11 +20,10 @@ class BlockEditorSettingsServiceTests: XCTestCase {
         contextManager = TestContextManager()
         context = contextManager.mainContext
         mockRemoteApi = MockWordPressComRestApi()
-        blog = ModelTestHelper.insertDotComBlog(context: context)
-        blog.dotComID = NSNumber(value: 1)
-        blog.account?.authToken = "auth"
-        blog.setValue("5.8", forOption: "software_version")
-
+        blog = BlogBuilder(context)
+            .with(wordPressVersion: "5.8")
+            .withAnAccount()
+            .build()
         service = BlockEditorSettingsService(blog: blog, remoteAPI: mockRemoteApi, context: context)
 
         try! FeatureFlagOverrideStore().override(FeatureFlag.globalStyleSettings, withValue: false)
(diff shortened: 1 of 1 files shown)


WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
```diff
@@ -85,8 +84,9 @@ class BlockEditorSettingsServiceTests: XCTestCase {
     }
 
     private func validateThemeResponse() {
+        let siteID = blog.dotComID ?? 0
```
mokagio, line 87: I assume it's okay to fallback to `0` for the id if none is found, otherwise you wouldn't have gone through the trouble of doing it. 

I'm curious about why that's the case 🙂

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
```diff
@@ -202,8 +202,23 @@ class BlockEditorSettingsServiceTests: XCTestCase {
...
+        }
+
+        guard let colors = blockEditorSettings.colors else {
+            XCTFail("Block editor colors should exist at this point")
+            return
+        }
+        XCTAssertGreaterThan(colors.count, 0)
+
+        guard let gradients = blockEditorSettings.gradients else {
+            XCTFail("Block editor colors should exist at this point")
```
mokagio:
```suggestion
            XCTFail("Block editor gradients should exist at this point")
```

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
```diff
@@ -20,11 +20,10 @@ class BlockEditorSettingsServiceTests: XCTestCase {
...
         mockRemoteApi = MockWordPressComRestApi()
-        blog = ModelTestHelper.insertDotComBlog(context: context)
-        blog.dotComID = NSNumber(value: 1)
-        blog.account?.authToken = "auth"
-        blog.setValue("5.8", forOption: "software_version")
-
+        blog = BlogBuilder(context)
+            .with(wordPressVersion: "5.8")
+            .withAnAccount()
+            .build()
```
mokagio, line 26: This is the code that addresses

> The error was a result of the code going down an unexpected path because the software version wasn't being properly picked up by the blog anymore.

How does this code addresses the issue? Is there more knowledge within `BlogBuilder` than in the previous setup approach? What part of the change in the hotfix surfaced this issue? (cc @emilylaguna)

mokagio (approved):
Looking good and the tests pass in CI and for me locally.

Left a minor suggestion and a couple of questions.

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
mokagio: BTW I'm asking this to:

- Learn about this overwhelmingly vast codebase 😅 
- Check whether there are other parts of the test suite that should be updated as a followup to avoid these issues in the future (e.g. is `BlogBuilder` the way to go for this? Then maybe it should be adopted across the board?)

mokagio:
I thought about it, and I don't see a reason for holding the merge while waiting for your answers. Even if there might be more work to do, we're talking about the tests not the production code, so nothing that would affect the users. We can keep working on this on `develop`.

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
chipsnyder: These test doesn't really care what the siteID is we just care that the correct path is used. That path includes the siteID. Originally (before using the BlogBuilder I hardcoded it to 1).

WordPress/WordPressTest/BlockEditorSettingsServiceTests.swift
chipsnyder: > How does this code addresses the issue? Is there more knowledge within BlogBuilder than in the previous setup approach? 

I'm still looking into this to try and answer that question and will let you know what I find. But it's a curious case to me too because conceptually the old version of this code is doing the same thing as the new version.

PR #16902, opened 2021-07-21
Add UI test to verify self-hosted site can be added after logging into WordPress.com

thehenrybyrd:
This PR is part of updating our login tests to account for the Unified Login & Signup flows. Context & status: paaHJt-1Kx-p2
This PR adds a new UI test, `testAddSelfHostedSiteAfterWPcomLogin`, to check that it's possible to add a self-hosted site after being logged into WordPress.com.

Since both the WordPress.com login flow and self-hosted login flow are covered in other tests, this test is mostly checking the transition between the two: tap the down arrow next to your site's title to open the site switcher, tap the "+" button, and choose "add self-hosted site".

To test:
Run `testAddSelfHostedSiteAfterWPcomLogin` from SignupTests.swift in Xcode.

Please comment on implementation as well as code quality! ❤️

### DIFF
--- a/WordPress/WordPressUITests/Tests/LoginTests.swift
+++ b/WordPress/WordPressUITests/Tests/LoginTests.swift
@@ -146,18 +146,24 @@ class LoginTests: XCTestCase {
             .proceedWith(password: WPUITestCredentials.testWPcomPassword)
             .verifyEpilogueDisplays(username: WPUITestCredentials.testWPcomUsername, siteUrl: WPUITestCredentials.testWPcomSitePrimaryAddress)
             .continueWithSelectedSite() //returns MySite screen
+
+            //From here, bring up the sites list and choose to add a new self-hosted site.
             .showSiteSwitcher()
-//            .tapPlusButton() // .addSelfHostedSite now includes plusButton.tap, so this should be redundant
             .addSelfHostedSite()
-            // and then we'll direct to the self-hosted login flow:
-            //.selectSiteAddress() this is reduntant with .addSelfHostedSite which returns the same screen
+
+            //Then, go through the self-hosted login flow:
             .proceedWith(siteUrl: WPUITestCredentials.selfHostedSiteAddress)
             .proceedWith(username: WPUITestCredentials.selfHostedUsername, password: WPUITestCredentials.selfHostedPassword)
             .verifyEpilogueDisplays(siteUrl: WPUITestCredentials.selfHostedSiteAddress)
             .continueWithSelfHostedSiteAddedFromSitesList()
+
+            //Login flow returns MySites modal, which needs to be closed.
             .closeModal()
+
+            //TODO: rewrite logoutIfNeeded() to handle logging out of a self-hosted site and then WordPress.com
+            //Currently, logoutIfNeeded() cannot handle logging out of both self-hosted and WordPress.com during tearDown(). So, we remove the self-hosted site before tearDown() starts.
             .removeSelfHostedSite()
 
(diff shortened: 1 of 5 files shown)


WordPress/WordPressUITests/Tests/LoginTests.swift
```diff
@@ -137,4 +137,33 @@ class LoginTests: XCTestCase {
             .tryProceed(password: "invalidPswd")
             .verifyLoginError()
     }
+
+    //Self-Hosted after WordPress.com login
+    //Login to a WordPress.com account, then add a self-hosted site.
+    func testAddSelfHostedSiteAfterWPcomLogin() {
+        _ = PrologueScreen().selectContinue()
```
ealeksandrov: I have a warning here:
```
Using '_' to ignore the result of a Void-returning function is redundant
```

Usually `_ =` used to silence another warning, when function returns something and it's ignored. Try removing it in `testUnsuccessfulLogin()` above, for example. Here we don't need it here because last function in the chain (`removeSelfHostedSite()`) doesn't return anything (= returns Void).

ealeksandrov (commented):
Great work!
I've tested it locally and test performs well ✅ 
Code is well-structured and I appreciate the comments 👏

I would be nitpicky about code (comments) style 🙂

1. Try to limit lines length, so we don't have to scroll it horizontally. I have "page guide" set at 160 columns in Xcode:
image

2. Keep a single space after all comment slashes:
```
// This line is easier to read
//Compared to this
```
3. There is some special treatment for keywords like TODO, FIXME, #warning and difference between `///` and `//`.
Here are some links for research (no need to learn all types of documentation, it's rarely used):
- https://www.vadimbulavin.com/the-art-of-commenting-in-swift/
- https://useyourloaf.com/blog/swift-documentation-quick-guide/
- https://nshipster.com/swift-documentation/
I mostly use `//` for inline comments about code logic, `///` (generated with `⌥⌘/`) for public functions/objects, TODO and MARK.

4. Code in many cases will be self-explanatory, so keep comments that only add value for a reader. For example:
```
/// This screen object represents the My Site screen. This is the home-base screen for...
class MySiteScreen: BaseScreen {
```
Object's name speaks for itself, we don't need to clarify that it's about "My Site screen" if we're looking for it in code. But comment description can describe the screen for someone who is not familiar with screen names - "home-base screen for an individual site".


Let me know what you think 🙂

WordPress/WordPressUITests/Tests/LoginTests.swift
thehenrybyrd: Ah, I see! When I remove this from `testUnsuccessfulLogin()`, I get a warning that `Result of call to 'verifyLoginError()' is unused`. Because `verifyLoginError()` does return something (the LoginPasswordScreen), we expect to use the return.
And `_ =` indicates we're not going to use the return, so there's no need for the warning.

thehenrybyrd:
@ealeksandrov Thanks for the nit-picks around commenting! Very helpful for me 😄 
I've updated the comments accordingly, but do have two questions.

1. Doing a quick search in Xcode, `//` is more common in WPiOS for MARK and TODO. Does this change how TODO works?

2. How do you decide when a function is 'important' enough for `///` ? It doesn't seem necessary for functions that are only called in one or two cases.

ealeksandrov:
> Doing a quick search in Xcode, // is more common in WPiOS for MARK and TODO. Does this change how TODO works?

No, I use TODO and MARK with //, that's just my confusing wording above 😄

> How do you decide when a function is 'important' enough for /// ? It doesn't seem necessary for functions that are only called in one or two cases.

I would say any **public** API should have such documentation - such as methods of networking layer, testing helper libs, etc. Even if they are used once in a project, while they are distributed as public library - they better to have proper full description.

For an internal project code - just use own judgement.
- If a helper function used once, but have potential for wider adoption - good description can encourage other developers to start using it.
- If a function have many non-obvious parameters - Xcode doc format can help lay out proper description for developers that will came across it in the future (and on review).

ealeksandrov (approved):
Thanks for changes, LGTM! :shipit:

PR #16898, opened 2021-07-20
Comment: convert `PostContentProvider` protocol methods to Swift

ScoutHarris:
Ref: #16876

This converts `Comment`'s `PostContentProvider` protocol methods to Swift.

Notes: 
- This is the first step in converting Comment to Swift, so I've started with Swift extensions. In the end, `Comment.swift` will be _the_ Comment.
- I didn't see a need for `WPCommentContentViewProvider` as it was only used by `Comment`. So I removed that and used `PostContentProvider` directly.
- `blogNameForDisplay` and `statusForDisplay` don't appear to be used (or I can't find where). I've simply converted the code for now and will follow up on this later.

To test:

Go to these views and verify these components are displayed as expected. That is, there should be no visible change.

Site > Comments list:
- Author name
- Author avatar
- Comment title
- Comment content

Site > Comments > Comment details:
- Author name
- Author avatar
- Comment title
- Comment content

Reader > post comments:
- Author name
- Author avatar
- Author URL (the author's name is linked to this)
- Comment creation date

## Regression Notes
1. Potential unintended areas of impact
The listed Comment components could be displayed incorrectly.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Compared each component with how it was before.

3. What automated tests I added (or what prevented me from doing so)
N/A

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #16876, opened 2021-07-15
Comment model: convert to Swift


ScoutHarris:
Convert the `Comment` model to Swift.

Ref: pctCYC-6n-p2

- [x] Convert to Swift
       - #16905
       - #16908
       - #16913
       - #16969
       - #16980
       - #16982
- [x] Uses WPCommentContentViewProvider.h. Should be able to include in Swift model. 
       - #16898
- [x] Add a `rawContent` field to the `Comment` model.
        - Ref: #16906
        - #16933
        - #16949

PR #16913, opened 2021-07-22
Comment: address ObjC public methods

ScoutHarris:
Ref: #16876 

This converts the ObjC `Comment` public methods to Swift. Plus these goodies:
- `statusForDisplay` and `blogNameForDisplay` were not used for Comments. Made them optional in the `PostContentProvider` protocol, and removed from `Comment`.
- Reader > post comments > comments Like button label: fixed an issue I introduced with [this](https://github.com/wordpress-mobile/WordPress-iOS/pull/16898/files#diff-2bee401a37e2812bcb276e9ea5a5e5c415ccd20e4b988dd0ecf621b52eb97c98) change. If there are no likes, the number should not be displayed (i.e. just `Like` instead of `0 Likes`).
- `authorIsPostAuthor`: This didn't actually work before. I've removed it for now as more investigation is required.

To test:

Reader > post card:
- Verify the displayed blog name is correct.

Reader > post comments:
- Verify the Like button label for each comment is correct. 
  - No likes = `Like`
  - One like = `1 Like`
  - More than one = `xxx Likes`
- Verify comment author names are blue and tappable if the author has a URL.

Comment Notifications > Comment details:
- Verify the author's URL after the comment date is correct (ex: `yesterday . my.blog`).

My Site > Comments list:
- Swipe left on a comment. verify the correct option is presented (Approve or Unapprove).
- With the `unifiedCommentsAndNotificationsList` feature disabled, verify comments are grouped by day, and the dates are correct.

My Site > Comments > comment details:
- On a site where you cannot moderate comments (your role is Author or Contributor), verify you can only Like and Reply on Approved comments.

## Regression Notes
Any of the above could be incorrect. 😄 

Compared with previous version of the app, verified all look the same.

N/A

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,7 @@
 17.9
 -----
 * [internal] Redirect Terms and service to open the page in an external web view [#16907]
-* [internal] Converted parts of Comment model to Swift. Should be no functional changes, but could cause regressions. [#16898, #16905, #16908]
+* [internal] Converted Comment model methods to Swift. Should be no functional changes, but could cause regressions. [#16898, #16905, #16908, #16913]
 * [*] Enables Support for Global Style Colors with Full Site Editing Themes [#16823]
 
 17.8
--- a/WordPress/Classes/Models/Comment.h
+++ b/WordPress/Classes/Models/Comment.h
@@ -39,7 +39,6 @@
 
 /// Helper methods
 ///
-- (NSString *)sectionIdentifier;
 - (BOOL)authorIsPostAuthor;
 
 @end
--- a/WordPress/Classes/Models/Comment.h
+++ b/WordPress/Classes/Models/Comment.h
@@ -39,7 +39,6 @@
 
 /// Helper methods
 ///
-- (BOOL)hasAuthorUrl;
 - (BOOL)isApproved;
 - (NSString *)sectionIdentifier;
 - (BOOL)isReadOnly;
--- a/WordPress/Classes/Models/Comment.h
+++ b/WordPress/Classes/Models/Comment.h
@@ -40,7 +40,6 @@
 /// Helper methods
 ///
 - (NSString *)sectionIdentifier;
-- (BOOL)isReadOnly;
 - (BOOL)authorIsPostAuthor;
 
 @end
--- a/WordPress/Classes/Models/Comment.h
+++ b/WordPress/Classes/Models/Comment.h
@@ -39,7 +39,6 @@
 
 /// Helper methods
 ///
-- (NSString *)authorUrlForDisplay;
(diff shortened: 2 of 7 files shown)


aerych (changes requested):
Howdy @ScoutHarris 👋 

```
Reader > post card:
      Verify the displayed blog name is correct.
```
Looks good!

```
Reader > post comments:
    Verify the Like button label for each comment is correct.
        No likes = Like
        One like = 1 Like
        More than one = xxx Likes
```
Yep yep.

```
    Verify comment author names are blue and tappable if the author has a URL.
    Verify comment author names are pink and tappable if:
        The comment author is the post author.
        The comment author URL is the same as the post author URL.
```
This seems a bit wonky. The cells are tappable but the color is not consistent.  For example, this Discover post:
author color
I'm sure the wonkiness already existed, we just couldn't see it because the colors never updated.  However, since the color is working again with these changes we should probably sort it out. 
```
Comment Notifications > Comment details:
    Verify the author's URL after the comment date is correct (ex: yesterday . my.blog).
```
Yep yep!
```
My Site > Comments list:
    Swipe left on a comment. verify the correct option is presented (Approve or Unapprove).
```
Yep yep!
```
    With the unifiedCommentsAndNotificationsList feature disabled, verify comments are grouped by day, and the dates are correct.
```
Yep yep!
```
My Site > Comments > comment details:
    On a site where you cannot moderate comments (your role is Author or Contributor), verify you can only Like and Reply on Approved comments.
```
Yep!  However, I happened to notice that the moderation options are available when swiping a cell.  Sneaky sneaky.  Existing issue. We can sort it out later.
Simulator Screen Shot - iPhone 11 - 2021-07-23 at 10 00 07


Back to you!

ScoutHarris:
> I'm sure the wonkiness already existed

It did. Personally I found it strange that it was using URLs to determine if the author was the same. I was going to change it to use `author`, but I noticed that in sometimes the comment.author and the post.author are different, even though they're the same person. So I simply corrected the old URL logic.

I'll just remove the color change. It will take more investigation to figure how to determine if the author is the same (or fix what _should_ work). And clearly it's not urgent since it's been broken for some time.

> I happened to notice that the moderation options are available when swiping a cell. Sneaky sneaky. Existing issue. We can sort it out later.

🤦  I didn't even notice that! Thank you! I will certainly look at it.

ScoutHarris:
Hey @aerych .

I removed the pink highlighting. All author links should follow the `Verify comment author names are blue and tappable if the author has a URL.` behavior once again.

And I've created #16918 for the swipe moderation. 

Back to you!

aerych (approved):
Shiny! :shipit:

ScoutHarris:
Thanks @aerych !

PR #16914, opened 2021-07-22
Themes: Fix theme url to include the correct image resized url

enejb:
This PR fixes the URL that we genetate to display the theme screenshots. 

Before: https://i2.wp.com/s2.wp.com/wp-content/themes/pub/blank-canvas/screenshot.png?ssl=1?w=409
After: https://i2.wp.com/s2.wp.com/wp-content/themes/pub/blank-canvas/screenshot.png?ssl=1&w=409

The fix should download smaller images for each of the themes saving bandwidth and mobile data.
 
This issue was discoverd while trying to clear the screenshot image cache for the blank-canvas theme. 

To test:
Navigate to the theme selection screen for a .com site. Notice that the screenshots show up as expected.
.org sites do not have the theme selection as an option.
Navigate to an atomic and jetpack site notice that the themes show up as expected. 

## Regression Notes
None that I can think of

N/A

None

PR submission checklist:


WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
```diff
@@ -193,10 +193,20 @@ open class ThemeBrowserCell: UICollectionViewCell {
...
+        }
+
+        if false == imageUrl.contains("?") {
+            return imageUrl + "?w=\(presenter!.screenshotWidth)"
+        }
+        return imageUrl + "&w=\(presenter!.screenshotWidth)"
+    }
+
+    fileprivate func refreshScreenshotImage(_ imageUrl: String) {
+        let imageUrlForWidth = self.getImageUrlForWidth( imageUrl: imageUrl )
```
chipsnyder: NitPick: `self` shouldn't be needed here.
```suggestion
        let imageUrlForWidth = imageUrlForWidth( imageUrl: imageUrl )
```

WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
```diff
@@ -193,10 +193,20 @@ open class ThemeBrowserCell: UICollectionViewCell {
         activityView.stopAnimating()
     }
 
-    fileprivate func refreshScreenshotImage(_ imageUrl: String) {
+    private func getImageUrlForWidth(imageUrl: String) -> String {
```
chipsnyder: NitPick: Style nitpick but in Swift, we usually leave off the `get` keyword for getters. 
```suggestion
    private func imageUrlForWidth(imageUrl: String) -> String {
```

WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
```diff
@@ -193,10 +193,20 @@ open class ThemeBrowserCell: UICollectionViewCell {
...
-        let imageUrlForWidth = imageUrl.hasPrefix("http") ? imageUrl + "?w=\(presenter!.screenshotWidth)" :
-            String(format: "http:%@", imageUrl)
+        if false == imageUrl.hasPrefix("http") {
+            return String(format: "http:%@", imageUrl)
+        }
+
+        if false == imageUrl.contains("?") {
+            return imageUrl + "?w=\(presenter!.screenshotWidth)"
+        }
+        return imageUrl + "&w=\(presenter!.screenshotWidth)"
```
chipsnyder: For me this method is a bit hard to read and understand what's happening. I wonder if we can clean it up with a guard statement: 
```suggestion
        guard imageUrl.hasPrefix("http") else {
            return String(format: "http:%@", imageUrl)
        }

        if !imageUrl.contains("?") {
            return imageUrl + "?w=\(presenter!.screenshotWidth)"
        }
        return imageUrl + "&w=\(presenter!.screenshotWidth)"
```


It seems like the goal here is to produce a URL where we append a new query parameter. I wonder if we'll ever want to include a set height or include the devices resolution as well to ensure consistent images. If we wanted this to return back the full URL we could enhance this to be:

```
var components = URLComponents(string: imageUrl)
components?.queryItems?.append(URLQueryItem(name: "w", value: "\(presenter!.screenshotWidth)"))
components?.url
```

As a follow-up note, I do have some issues with the need to force unwrap the presenter with `presenter!`. I realize this was in place beforehand and that a crash here is an edge case but we can probably conditionally unwrap this and not add the parameter if the size isn't available.

chipsnyder (changes requested):
Hey @enejb, great job discovering this issue and proposing a solution. I made a couple of NitPick suggestions and offered up some readability improvements in addition it looks like there is a force unwrap case from the original code that we can probably clean up while we're making this change.

WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
enejb: Remove self and update method name.

WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
enejb: update method name to not have get

WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
enejb: Thanks Chip
I update the method to use the URLComponents.

enejb:
Thanks for all your help @chipsnyder 
Can you take another look when you have a moment?

WordPress/Classes/ViewRelated/Themes/ThemeBrowserCell.swift
```diff
@@ -193,11 +193,34 @@ open class ThemeBrowserCell: UICollectionViewCell {
...
+            // Since not all sites support https
+            screenshotUrl = String(format: "http:%@", imageUrl)
+        }
+
+        guard var components = URLComponents(string: screenshotUrl) else {
+            return screenshotUrl
+        }
+
+        var queryItems: [URLQueryItem] = components.queryItems ??  []
+        queryItems.append(URLQueryItem(name: "w", value: "\(presenter!.screenshotWidth)"))
```
chipsnyder: I [think](https://github.com/wordpress-mobile/WordPress-iOS/pull/16914#discussion_r675602910) we should handle this `presenter!` by conditionally unwrapping. 

I can push a change here though since I know you're away this week.

chipsnyder (approved):
I pushed the change for the small refactor I wanted to see the rest looks good to me. Thanks for handling this Enej!

PR #16917, opened 2021-07-23
Reader Comments: Add styles for table to improve readability

### LINKED ISSUE #16821: Tables are not rendered correctly in reader comments
### Expected behavior
Table columns should maintain a minimum readable width, regardless of the large difference in content length on the other columns.

### Actual behavior
WordPress App | Mobile Web
-- | --
wp-app | mobile-web

1. Create or view a comment that has a large difference of content length between columns (see example ☝️ ). 
2. View the comments from the WordPress app.

💡 Tip: If you have a direct URL to the comment, you can paste it to the Simulator's Safari. If you pull down the page, it will show a "Show in app" banner which will take you to the Reader comments (provided you have sufficient access to the site).

### Findings
- This bug happens in Reader comments. Incidentally, only this section renders comments with rich content. In other parts, such as My Site > Comments, or Notifications, do not render the tables at all.
- Rich text rendering is handled in `WPRichContentView`. Since tables are rendered via the WebKit engine through `NSAttributedString`, we have limited control over how column width is distributed. 
  - One potential workaround is to assign a `min-width` CSS property to all `td` elements. I've tested that only concrete values work (em, px, pt); percentage units, `min-content`, `max-content`, and `fit-content` does not seem to show any effect.

dvdchr:
Fixes #16821

## Summary
This PR adds basic styling to improve readability of tables in Reader Comments section. Here's a summary of style changes:

- Text are vertically aligned to top.
- Font styles for `body` does not apply to the `table` element somehow. This change ensures that the font style for body is applied properly for table elements.
- Added some padding to allow for some reading space.
- Added `min-width` property to ensure that the columns doesn't get too cramped.

Before | After
--|--
before | after

Note that table borders are not added yet, due to the strange behavior of attributed string with HTML tables. 😅 .  We'll look into using a web view instead to display comment details in the later part of the iteration.

### How does this look with many columns?

Image | Description
--|--
Simulator Screen Shot - iPhone 8 - 2021-07-23 at 22 42 27 | In the browser, this table would be horizontally scrollable. TextKit, however, will force the table to fit its contents within the viewport width. Therefore, the `min-width` property will be ignored in this case.

## To Test
- Open a Reader Post that has table(s) in one of the comments.
- Verify that the tables are displayed according to the changes described in the summary above.

## Regression Notes
None.

Only Reader Comments render content with attributed string.

n/a

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
+++ b/WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
@@ -173,12 +173,16 @@ class WPRichContentView: UITextView {
 private extension WPRichContentView {
     class func formattedAttributedString(for string: String, style: AttributedStringStyle) -> NSAttributedString {
         let styleString = "<style>" +
-            "body { font:-apple-system-body; font-family: 'Noto Serif'; font-weight: normal; line-height:1.6; color: #\(style.textColorHex); }" +
+            "body, table { font:-apple-system-body; font-family: 'Noto Serif'; font-weight: normal; line-height:1.6; color: #\(style.textColorHex); }" +
             "blockquote { color:#\(style.blockQuoteColorHex); } " +
             "em, i { font:-apple-system-body; font-family: 'Noto Serif'; font-weight: normal; font-style: italic; line-height:1.6; } " +
             "a { color: #\(style.linkColorHex); text-decoration: none; } " +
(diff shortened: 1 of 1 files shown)


aerych (approved):
Hola @dvdchr 👋 
Tested with some tables in comments and the updated formatting looks good. 
:shipit: from me :)

PR #16927, opened 2021-07-26
Let http handle cookies when downloading images.

### LINKED ISSUE #16926: Notifications: Images are not loading for some comment notifications
Reported in p1626981490045500-slack-apps
> When you land on a comment from a notification on P2, the images don’t render. But if you go back to the main post and scroll down to the comment, they appear 

I was able to reproduce the issue (h/t @ScoutHarris for the assist) on a test p2. pcl4cA-wl-p2.  I tracked the issue to server returning a web page describing a 403 error. ```403: Access Denied &#8212;<br/><br/>This file requires authorization:<br/><br/>You must be logged in<br />and a member of this blog. ```

aerych:
Fixes #16926 

This PR resolves the issue described in #16926 by removing `httpShouldHandleCookies = false`. This allows the image loader's NSURLRequests to use any auth cookies that exist in the cookie store.  The result is that image on private sites (like the p2 in the report) to load successfully.

To test:
Under Notifications:
- Identify a comment notification that contains images. I can assist in creating one if needed.
- View the comment notification via this branch.
- Confirm images load correctly as expected.

Under My Sites 
- Create a new post on a private p2 site.
- Add an image from the site's media library.
- After publishing the post, open it in the editor again.
- Confirm the image loads correctly.

@leandroalonso could I trouble you for a review of this one given your familiarity with the GutenbergMediaEditorImage? I'm not sure I've found the best test case to check for regressions and you might know of a better way. 

## Regression Notes
The image downloader is also used by Gutenberg.

Smoke tested loading images from a private site in the editor.

n/a
PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/AccountService+Cookies.swift
+++ b/WordPress/Classes/Services/AccountService+Cookies.swift
@@ -9,7 +9,7 @@ extension AccountService {
         guard
             let account = service.defaultWordPressComAccount(),
             let auth = RequestAuthenticator(account: account),
-            let url = URL(string:WPComDomain)
+            let url = URL(string: WPComDomain)
         else {
             return
         }
--- a/WordPress/Classes/Services/AccountService+Cookies.swift
+++ b/WordPress/Classes/Services/AccountService+Cookies.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+extension AccountService {
+
+    /// Loads the default WordPress account's cookies into shared cookie storage.
+    ///
+    static func loadDefaultAccountCookies() {
+        let service = AccountService(managedObjectContext: ContextManager.shared.mainContext)
+        guard
+            let account = service.defaultWordPressComAccount(),
+            let auth = RequestAuthenticator(account: account),
+            let url = URL(string:WPComDomain)
+        else {
+            return
+        }
+        auth.request(url: url, cookieJar: HTTPCookieStorage.shared) { _ in
+            // no op
+        }
+    }
+
+}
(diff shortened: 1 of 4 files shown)


leandroalonso:
@aerych `GutenbergMediaEditorImage` is just a helper for the Media Editor, it uses the method you changed to download thumbnails in the case there are multiple images. I don't think there's any user case where this is used on the app right now so we're good to go.

I tested both cases you added on the PR's description. The editor works just fine (tested with a .com private site — I had issues with an Atomic one even on the web).

For a comment it didn't worked, here's what I did:

* Commented on a private .com site and added an image. Here's the content of the comment: `Hello `
* Checked on the app
* The app tries to download the image, however it fails to generate the image:

Screen Shot 2021-07-27 at 17 20 36

(server response is a: `403: Access Denied`)

I don't know if the goal of this task is to address that for *any* private site, but for P2s it looks like is working just fine.

aerych:
Nice catch @leandroalonso  I was able to reproduce this with my own comment on a private site.  

The intention of this PR was to fix the issue everywhere. In this case it looks like auth cookies weren't guaranteed to be loaded into the cookie store.  I've created a helper for loading the default account's auth cookies during app start up and when the account changes.  This _should_ correct the issue for private sites.

leandroalonso:
@aerych I tested and can guarantee that `loadDefaultAccountCookies` is indeed being called but I'm still seeing the same issue. :(

I also requested the same image through the editor to compare the cookies and the values are different. I can DM you the curl request if you want.

leandroalonso:
Oh, a confusing thing.

After loading the image from the editor, it worked. But you can see the issue is still there if you:

1. Delete WordPress app
2. Go to Settings > Safari > Clear History and Website Data
3. Run WordPress
4. Login
5. Go directly to the comment
6. Check that image don't load

leandroalonso (approved):
@aerych I retested without being proxied and it worked as described.

aerych:
Thanks @leandroalonso !!

PR #16906, opened 2021-07-21
Reader: Fixes an issue with reader comments being incorrectly rendered.

aerych:
This PR is part of a bug fix for WPiOS `release/17.8`. 
More in p1626841091202200-slack-hummingbird

WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/419

In https://github.com/wordpress-mobile/WordPressKit-iOS/pull/413 we made a change to consume `raw_content` rather than `content` field from the Comments endpoints.  This [resolved](https://github.com/wordpress-mobile/WordPress-iOS/pull/16768) an [issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/16685) with Authors and Contributors not being able to view Comment content.  However, we missed that it created a new issue with comments in the Reader being incorrectly rendered in many cases as the `raw_comment` field does not include html mark up for non-P2 WordPress sites.  P2s that use the block editor for comments were rendered correctly.  We were unaware of this difference between P2 and non-P2 comments.  The reader's caching behavior likely contributed to missing the regression. 

Example of the badness. Note the cropped lines of text:
Simulator Screen Shot - iPhone 11 - 2021-07-20 at 23 12 22

This PR adopts changes in WordPressKit to consume both `content` and `raw_content` fields returned by the comments endpoint.  Blog comments that are wrangled by the My Site and Notifications feature continue to use the `raw_content` value.  Reader comments are assigned the `content` value to preserve the expected HTML mark up and restore the correct rendering behavior.   The changes are a temporary measure that we'd like to improve in a subsequent update.

### Testing Details
Switch to this branch and test the following: 
- Confirm that reader comments are correctly rendered.  
- Confirm that comments in My Sites are correctly rendered.
- Confirm that authors and contributors can still view comments under My Sites.
- Confirm that comments under Notifications are correctly rendered.
- Confirm that replying to a reader comment works as expected and there are no layout issues. 
- Confirm that editing a my sites or notifications comment works as expected.

## Regression Notes
1. Potential unintended areas of impact
Non-reader comments.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Smoke tested Notifications and My Site comments.

3. What automated tests I added (or what prevented me from doing so)
Requires a visual evaluation.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
[...]

PR #16933, opened 2021-07-27
Comment: store raw content

ScoutHarris:
Ref: #16906, #16876

This stores the `rawContent` property in the `Comment` model, and uses it when displaying:
- Comment snippet in Comments list.
- Comment content in Comment details.
- Comment content when editing a Comment.

`rawContent` contains markup for Gutenberg comments, so that is stripped when the Comment snippets are displayed via My Site > Comments, but is still displayed when editing Comments.

To test:
There should be no visible change where Comment content is displayed. So verify Comments content is displayed as it was before in:

- My Site > Comments.
- My Site > Comments > Comment details.
- My Site > Comments > Comment details > edit.
- Reader > post comments.
- Notifications > Comments > notification details.

Please check P2 and non-P2 sites.

## Regression Notes
Comments could be displayed incorrectly.

Compared Comment displays with previous version of the app to verify they are the same.

I'll address in a future PR.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/CommentService.m
+++ b/WordPress/Classes/Services/CommentService.m
@@ -1008,7 +1008,7 @@ static NSTimeInterval const CommentsRefreshTimeoutInSeconds = 60 * 5; // 5 minut
 
 - (void)updateCommentAndSave:(Comment *)comment withRemoteComment:(RemoteComment *)remoteComment
 {
-    [self updateReaderComment:comment withRemoteComment:remoteComment];
+    [self updateComment:comment withRemoteComment:remoteComment];
     // Find its parent comment (if it exists)
     Comment *parentComment;
     if (comment.parentID) {
@@ -1039,7 +1039,7 @@ static NSTimeInterval const CommentsRefreshTimeoutInSeconds = 60 * 5; // 5 minut
             comment = [NSEntityDescription insertNewObjectForEntityForName:entityName inManagedObjectContext:self.managedObjectContext];
         }
 
-        [self updateReaderComment:comment withRemoteComment:remoteComment];
+        [self updateComment:comment withRemoteComment:remoteComment];
 
         // Calculate hierarchy and depth.
         ancestors = [self ancestorsForCommentWithParentID:comment.parentID andCurrentAncestors:ancestors];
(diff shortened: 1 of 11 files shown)


WordPress/Classes/Services/CommentService.m
```diff
@@ -1136,46 +1136,15 @@ - (Comment *)findCommentWithID:(NSNumber *)commentID fromPost:(ReaderPost *)post
...
-- (void)updateReaderComment:(Comment *)comment withRemoteComment:(RemoteComment *)remoteComment
 {
     comment.commentID = remoteComment.commentID;
     comment.author = remoteComment.author;
     comment.author_email = remoteComment.authorEmail;
     comment.author_url = remoteComment.authorUrl;
     comment.authorAvatarURL = remoteComment.authorAvatarURL;
     comment.content = remoteComment.content;
+    // rawContent contains HTML for P2 comments. Remove it to get true raw content.
+    comment.rawContent = [[[[remoteComment.rawContent stringByDecodingXMLCharacters] trim] stringByStrippingHTML] stringByNormalizingWhitespace];
```
aerych: I've questions about this. When editing a comment it seems like we'd (eventually) want the raw content as it comes from the server, rather than a filtered version.  It might not matter until we get to the block editor, just thinking ahead.
I notice that with this patch I'm seeing html when editing a comment, which I'm not seeing in production or the beta. I think it might be because we're still referencing .content for the editor rather than .rawContent.  
Thoughts?

aerych (changes requested):
Howdy @ScoutHarris 👋 
Looking good. :) I spotted what I think is an unintentional change with the comment editor and left a comment about a concern for the future.  Lemme know what you think!

WordPress/Classes/Services/CommentService.m
ScoutHarris: > It might not matter until we get to the block editor, just thinking ahead.

Ah, I didn't think about that. I'll move the filtering back to `Comment`.

> I notice that with this patch I'm seeing html when editing a comment

Ugh. I thought I checked that. Thank you!

ScoutHarris:
Hey @aerych .

I moved filtering `rawContent` to `Comment` when it's used for display. Editing a Comment now uses `rawContent`. Please note that Gutenberg comments will show markup, but that's as it was before. "Plain" comments won't show any markup.

Back to you please.

aerych (approved):
Shiny! :shipit:

ScoutHarris:
Thanks @aerych !

PR #16935, opened 2021-07-28
Improve Cocoapods caching

jkmassel:
Improves CocoaPods caching by:
- Using the more specialized CocoaPods-specific caching functions
- Setting up the infrastructure for automated cache rebuilds 

To test:
- Ensure CI passes
- Ensure that [the test scheduled task](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/1681) passes as well

## Regression Notes
n/a

n/a

n/a
PR submission checklist:


.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,39 @@
...
+  # Build the CI cache
+  #################
+  - label: ":git: Rebuild git cache"
+    command: |
+      cache_repo "a8c-repo-mirrors"
+    agents:
+      queue: "default"
+    env:
+      IMAGE_ID: xcode-12.5.1
+    <<: *use_plugins
```
mokagio: For cross reference, I left a question regarding having this pipeline tracked in this repo as opposed to a dedicated one [here](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/1#discussion_r679642403).

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,39 @@
...
+  #################
+  # Build the CI cache
+  #################
+  - label: ":git: Rebuild git cache"
+    command: |
+      cache_repo "a8c-repo-mirrors"
+    agents:
+      queue: "default"
+    env:
+      IMAGE_ID: xcode-12.5.1
```
mokagio: The step to cache the repo shouldn't need Xcode, right?

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,39 @@
...
+  #################
+  - label: ":cocoapods: Rebuild CocoaPods cache"
+    command: |
+      echo "--- :rubygems: Setting up Gems"
+      install_gems
+
+      echo "--- :cocoapods: Rebuilding Pod Cache"
+      cache_cocoapods
+    env:
+      IMAGE_ID: xcode-12.5.1
```
mokagio: If the step beneath needs access to the Xcode toolchain, then I'd suggest using the `common_env` anchor defined above here:

```suggestion
    env: *common_env
```

Otherwise, I'd suggest removing the `common_env` definition, since this would be the only usage.

.buildkite/commands/run-ui-tests.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -e
+
+TEST_NAME=$1
+DEVICE=$2
+IOS_VERSION=$3
+
+echo "Running $TEST_NAME on $DEVICE for iOS $IOS_VERSION"
```
mokagio, line 7: Worth adding `set -u` or combining with the already present `set -e` using `set -eu` so that the script will fail if these three values are not set.

I learned about `set -u` just recently, BTW. `u` stands for unbound, as in "fail on unbound variables".

I checked this locally with a dedicated script:

```sh
#!/bin/bash

set -eu

TEST_NAME=$1
DEVICE=$2
IOS_VERSION=$3

echo "Running $TEST_NAME on $DEVICE for iOS $IOS_VERSION"
```

```
$ sh ./t.sh
./t.sh: line 5: $1: unbound variable

$ sh ./t.sh a b c
Running a on b for iOS c
```

.buildkite/pipeline.yml
```diff
@@ -2,8 +2,8 @@
 common_params:
   # Common plugin settings to use with the `plugins` key.
   - &common_plugins
-    - automattic/bash-cache#v1.0.0: ~
-    - automattic/git-s3-cache#add/vm-host-support:
+    - automattic/bash-cache#add/cocoapods-support: ~
+    - automattic/git-s3-cache#v1.1.0:
```
mokagio: Lazy question: How come did this change? What's the role of `~` after the `:`?

mokagio (commented):
None of my comments is a blocker, but I wonder if this should wait to be merged till we publish a version of our plugin? Or at least merge https://github.com/Automattic/bash-cache-buildkite-plugin/pull/1 into the main branch so the pipeline doesn't point to a feature branch that might get deleted.

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,39 @@
...
+    <<: *use_plugins
```
AliSoftware: Is this working? I don't see an anchor named `&use_plugins` defined anywhere in this YAML, so I believe that would produce a YAML error during parsing…

You probably meant:
```suggestion
    plugins: *common_plugins
```
Since `&common_plugins` is defined as an array of the plugins you want to use here…

mokagio: @AliSoftware is right.

I launched this pipeline manually by passing `PIPELINE=cache-builder.yml` and [it failed](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/1775#93a58005-21d0-4653-b963-c3348ef7ff95):

```
2021-08-02 06:04:44 FATAL  Pipeline parsing of "cache-builder.yml" failed (Failed to parse cache-builder.yml: unknown anchor 'use_plugins' referenced)
```

The reason we didn't notice immediately is that this is not the default pipeline.

In the future, we might benefit from some kind of pipeline validation. So far, I've been validating my changes with 

```
ruby -e "require 'yaml'; puts YAML.load_file('./.buildkite/cache-builder.yml')"
```

.buildkite/cache-builder.yml
jkmassel: Ah yep – I originally had this running on macOS, then realized that wasn't necessary but forgot the key 😅

.buildkite/cache-builder.yml
jkmassel: Addressed in b53bd9e962452335c1d9a9b0101b2e1913ff6090, thanks!

.buildkite/cache-builder.yml
jkmassel: Yep good call – addressed this in b53bd9e962452335c1d9a9b0101b2e1913ff6090 – IMHO it's nice to have the header the same in both pipeline files, so I like the reference approach 🙂

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,36 @@
...
+  #################
+  - label: ":cocoapods: Rebuild CocoaPods cache"
+    command: |
+      echo "--- :rubygems: Setting up Gems"
+      install_gems
+
+      echo "--- :cocoapods: Rebuilding Pod Cache"
+      cache_cocoapods
+    <<: *common_env
+    <<: *common_plugins
```
AliSoftware: The `common_env` anchor points to a dictionary that is `IMAGE_ID: xcode-12.5.1`. That dictionary doesn't contain the `env:` key itself, so you need to use the `env:` key at call site here, and use `&common_env` as the content of that `env:` key.

Otherwise, if you use `<<:` like you did here, the `IMAGE_ID: xcode-12.5.1` dictionary will be **merged** with the current dictionary in that context (the one that has `label` and `command` keys and you will end up with a dictionary containing `label`, `command`, and `IMAGE_ID` at the same level.

Similarly for `common_plugins`, I'd actually be surprised is the YAML parser would be happy with that one given that the `&common_plugins` anchor contains not a dictionary, but an array (of 2 items, one being the string `automattic.bash-cache#v1.1.0` and one being a sub-dictionary with a single key `automattic/git-s3-cache#v1.1.0` and a `{bucket:…, repo:…}` as its value). So here with `<<:` you're trying to merge into the current dictionary (the one containing `label` and `command` keys)… an array. You can't merge a dictionary with an array.

```suggestion
    env: *common_env
    plugins: *common_plugins
```
Same thing for other places where you used that syntax.

---

PS: And if YAML ends up considering this valid and not erroring on that, it's probably because it does something other than what you want, like maybe in those occasions it falls back to interpreting this `<<:` as an actual key of name `<<` rather than the merge operator since the merge operator wouldn't make sense between a dict and an array) – EDIT: that's exactly what happens, confirmed by reading the YAML in ruby and printing it out, with the trick that Gio explained above.

AliSoftware (changes requested):

mokagio:
Bumped this to 18.1 because I'm starting the 18.0 code freeze now.

If we think that the Buildkite work could be done during this sprint, I'd be happy to merge it in the release branch for 18.0 and live test it to enjoy the Buildkite's stack better build time.

.buildkite/commands/run-ui-tests.sh
jkmassel: I ended up force-pushing a bunch to clean this up, but this is addressed in 31495d7242f7be8cde610a89b611b65eb070a257

.buildkite/pipeline.yml
jkmassel: It's from the Buildkite docs – tbh I have no idea what it does 😜

.buildkite/cache-builder.yml
jkmassel: Fixed in 91a05f423af0693ccf1a03f6b8447038eb3bbfc0

jkmassel:
@mokagio and @AliSoftware – this should be ready for another review 🙂

WordPress/WordPressUITests/WordPressUITests.xctestplan
```diff
@@ -26,7 +26,8 @@
         "LoginTests\/testSelfHostedUsernamePasswordLoginLogout()",
         "LoginTests\/testUnsuccessfulLogin()",
         "LoginTests\/testWpcomUsernamePasswordLogin()",
-        "SignupTests\/testEmailSignup()"
+        "SignupTests\/testEmailSignup()",
+        "SupportScreenTests"
```
AliSoftware: Is it tracked somewhere that this test was (temporarily, I assume) disabled? (I'm assuming that was done because this test didn't pass but fixing it was a separate task for a separate PR… that needs a tracking issue to not forget about it)

fastlane/Fastfile
```diff
@@ -874,6 +874,7 @@ platform :ios do
       scheme: 'WordPress',
       device: options[:device],
       deployment_target_version: options[:ios_version],
+      ensure_devices_found: true,
```
AliSoftware, line 903: TIL about this option of `multi_scan`

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,35 @@
...
+    plugins: *common_plugins
+
+  #################
+  # Build the CI cache
+  #################
+  - label: ":git: Rebuild git cache"
+    command: "cache_repo a8c-repo-mirrors"
+    plugins: *common_plugins
+    agents:
+      queue: "default"
```
AliSoftware: Can you explain to a Buildkite noob like me when this YML / pipeline would be triggered?

I don't see it referenced in the main pipeline, nor being run in https://buildkite.com/wordpress-mobile/wordpress-ios/builds/1977… so I'm wondering when this is used and how it works, and also why this is necessary.

I assumed that, as for most cases with CI and how we did it in CircleCI, the cache would instead be saved after install (and only if the lockfile differs), as typically we have steps like this in typical CI pipelines:
```
 - restore_cache key:{checksum Podfile.lock} # restore cache if one exists for that key, nothing if cache miss
 - pod install
 - save_cache key:{checksum Podfile.lock} # does nothing if cache for that key already exists, creates if it not
```

So I'm unsure why this is done differently, with the building of the cache handled as an entire separate pipeline that which don't see when it's triggered, rather than doing it the usual way and bundling that restore/build/save dance directly as a buundled custom command as part of the bash-cache plugin?

.buildkite/pipeline.yml
AliSoftware: @mokagio  @jkmassel: `~` is one of the possible ways in YAML to represents a `null` value.
 - Another option being just to not specify the value at all (remove the `~` and keep it empty after the `:`) but is generally not favored as it could lead to ambiguity (especially making you wonder if the next line is a totally separate key, or the nested value of the current key; indentation of the next line is supposed to tell the difference, but still)
 - A more common option is the English spelling `null`. Some prefer `null` for its explicitness, other argue that `~` help avoid the risk of some people thinking `null` could mean the `"null"` literal string… but `~` being less clear (proof: you had to ask the question) kinda defeats that argument imho
 - The last option is to force the type using a tag (tags in YAML start with `!!` and provide metadata info to the YAML parser, like the type in this case): YAML supports type tags like `!!null` for null value, but also others like `!!str`, `!!map`, `!!seq` and `!!bool` for example; `!!bool` could especially be useful in some contexts to disambiguate between the `!!str "yes"` string and the `!!bool yes` true value.

Here, since the `bash-cache` plugin doesn't need additional parameters (unlike other plugins like the `docker` or `git-s3-cache` plugins typically do), the value associated with that key is null (as opposed to a nested dictionary providing the additional parameters). So here with `~` it's a bit like you're basically passing an empty dictionary for the parameters of `bash-cache` (but not sure we can create empty dictionaries in YAML, but we can pass `null` which is what `~` is here)

AliSoftware: As an additional note, the way CircleCI handles those cases (of things sometimes having extra parameters, sometimes not) is to allow you to use single strings instead of a key with null value in those contexts. For example in CircleCI we'd have:
```yml
Screenshots:
    when: << pipeline.parameters.generate_screenshots >>
    jobs:
      - raw-screenshots
      - promo-screenshots:
          requires: [raw-screenshots]
```
Where CircleCI expects `jobs` to be an array for which each item can either be a single String (`raw-screenshots`) with just the name of a job… or a single-item dictionary (`promo-screenshots: <extra options here>` with the key being the job's name and the value being the options), depending if you need to specify extra options or not.

<details><summary>Bonus note</summary>
Note how here in the extra options passed to `promo-screenshots`, the `requires:` option uses `[raw-screenshots]` which is an array, but declared with `[]` brackets (JSON-style) and not with the `-` one-item-per-line-dash-list syntax (more common YAML-style) for conciseness. Both are valid, because YAML is a superset of JSON.
</details>

While in BuildKite syntax, it seems that in those kind of lists, BuildKite expects every item to be a single-item dictionary (rather than allowing sometimes a String, sometimes a single-item dict like CircleCI). And then for items that don't need extra options, expect you to use null (`~`) as the value for those non-existing options.

I have no idea if BuildKite syntax is lenient enough to also allow you to use a single String instead of a "single-item dictionary with null value" in those cases and if both are valid in BuildKite or not, but if the BuildKite docs suggests to use `~`, I think it's best to use what they suggest :wink:

WordPress/WordPressUITests/WordPressUITests.xctestplan
jkmassel: @thehenrybyrd beat me to merging this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17015 – she'll take care of sorting it out! 🙂

.buildkite/cache-builder.yml
jkmassel: For sure – it would be triggered via Buildkite's [Scheduled Jobs](https://buildkite.com/wordpress-mobile/wordpress-ios/settings/schedules) feature to run nightly (or on whatever frequency makes sense)

In terms of the triggering – the git cache has to be managed separately since it's based on a git mirror, which is a different kind of checkout – for that reason we can't just save the result of a regular build.

For Pods, I want to regenerate those from scratch from time to time in order to prevent the cache from containing every version we've ever used of the Pods. I think there's a middle ground here where we maintain some recent-ish versions, but the returns there are relatively small. Of course we could do this as part of the regular job, but the upload and packaging time for these caches is non-zero, so doing it async speeds up every job.

WordPress/WordPressUITests/WordPressUITests.xctestplan
jkmassel: Also I've ~rebased on~ merged with `develop` to fix the conflict associated with that :)

.buildkite/cache-builder.yml
jkmassel: Also as an aside, implementing this in this way was a good way to get set up for the "build via API call" – it'll use a similarly dedicated pipeline specifically for that task.

AliSoftware: > For sure – it would be triggered via Buildkite's Scheduled Jobs feature to run nightly (or on whatever frequency makes sense)

Ah! I thought about that, but then went to look at the Scheduled Jobs settings and didn't see it listed there so thought that I was mistaken and that wasn't where it was being used. But now I realize it's just not there _yet_ – because the pipeline needs to land in `develop` before you can add it there 😅 

> In terms of the triggering – the git cache has to be managed separately since it's based on a git mirror, which is a different kind of checkout – for that reason we can't just save the result of a regular build.

Oh I thought about that at first but from our 1:1 the other day I thought only Tumblr/Orangina used a git mirror while other, less "massive" repos would directly pull from GitHub without an intermediate mirror. It seems you decided to apply the mirror solution to every repo after all, because why not, right? 👍 

> For Pods, I want to regenerate those from scratch from time to time in order to prevent the cache from containing every version we've ever used of the Pods.

I see, and that's indeed a good idea. Could be worth maybe add some comments in that yml file to quickly mention/explain this whole rational?

jkmassel: > apply the mirror solution to every repo

Probably just for the root projects, but not the Pods (WPiOS is > 2GB, so it's worth it, while the pods are all tiny).

But it's easy enough to add to any project that once it hits a point where it's a little slow we can add it to that project without much fuss!

AliSoftware: > Also as an aside, implementing this in this way was a good way to get set up for the "build via API call" – it'll use a similarly dedicated pipeline specifically for that task.

Yep I also thought that this pipeline would only be triggered by API call but then I saw no change in the `Fastfile` to provide a way to use the BuildKite API to do that like the ones we have to trigger CircleCI builds, so thought "maybe not, otherwise he'd have added a way to call that pipeline at least in order to test it".

---

Which I actually think could be nice before merging that PR (I mean, maybe not implement the whole BuildKite API call lane, because that logic will likely be implemented in the release-toolkit instead, but triggering that pipeline at least once before merging the PR just to validate that it works and passes too)

jkmassel: > add some comments in that yml file

Ya good call – added in fbaefc6106673abbc3704fa5313f607dd8ad8c09

.buildkite/cache-builder.yml
```diff
@@ -14,6 +19,9 @@ steps:
 
   #################
   # Build the CocoaPods Base Cache
+  # 
+  # This prevents the base cache from infinite growth caused by storing every
+  # version of every pod we've ever used.
```
AliSoftware, line 24: ❤️

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,46 @@
...
+      echo "--- :cocoapods: Rebuilding Pod Cache"
+      cache_cocoapods
+    env: *common_env
+    plugins: *common_plugins
+
+  #################
+  # Build the Git Repo cache
+  #
+  # Because this repo is so large, we create a Git Mirror and copy it to S3, where
+  # it can be fetch by agents more quickly than from GitHub.
```
AliSoftware:
```suggestion
  # Because this repo is so large, we periodically create a Git Mirror and copy it to S3,
  # from where it can be fetched by agents more quickly than from GitHub, and so that
  # agents then have less new commits to `git pull` by using that cache as starting point
```
Just thought it could help clarify things to mention that even if we fetch the initial Git Repo via the mirror that is build overnight, we still pull the latest commits on top of that mirror/cache afterwards – otherwise people might wonder when reading this if there isn't a risk that we might be missing commits (from the mirror being too old after a couple of hours and new commits) when we run tests on new PRs 😉

jkmassel: Ya for sure! Can never over-document it, right? 😁

.buildkite/cache-builder.yml
```diff
@@ -0,0 +1,47 @@
...
+  # This prevents the base cache from infinite growth caused by storing every
+  # version of every pod we've ever used.
+  #################
+  - label: ":cocoapods: Rebuild CocoaPods cache"
+    command: |
+      echo "--- :rubygems: Setting up Gems"
+      install_gems
+
+      echo "--- :cocoapods: Rebuilding Pod Cache"
+      cache_cocoapods
```
AliSoftware: Had to look at the implementation and especially [what directories were saved](https://github.com/Automattic/bash-cache-buildkite-plugin/blob/trunk/bin/cache_cocoapods#L7-L8) to realize this doesn't cache the `Pods/` directory from the repo (†) but instead cache the `~/.cocoapods` and `~/Library/Caches/CocoaPods/` cache, which makes way more sense as a scheduled pipeline run overnight.

I wonder if we could better document this to avoid the ambiguity.

(†) And I would have had concerns if it were that case because it'd then only cache the `Pods/` dir from `develop` then, and cleaning the cached `Pods/` from PRs and branches that modify the `Podfile.lock`

jkmassel: Hmm yeah – ideally I'd like this to be self-explanatory, so WDYT about changing the `cache_cocoapods` action to something like `cache_cocoapods_specs_repos` or similar?

AliSoftware: Yep, that was gonna be my next suggestion indeed.

I think the confusing thing is that "Pod" / "Pods" / "Pod Cache" is currently used interchangeably with "cocoapods", for example the section is called "Pod Cache" with is probably what made me think of `Pods/` instead, and also `cache_cocoapods` and `install_cocoapods` both use "cocoapods" in their name while one cache the pods repositories and the `~/Library/Cache/Pods` which are system-wide things, while the other install the actual pods in the `Pods/` directory of the repo.

Changing that to `cache_cocoapods_specs_repos` would solve that indeed 👍  – though… this command also caches `~/Library/Caches/CocoaPods/` not just the spec repo at `~/.cocoapods` (but I still like it better than just `cache_cocoapods`)… damn, naming is hard.

jkmassel: Addressed in 1b12300b22ffca7d63e475206dab545e8907399c

.buildkite/commands/run-ui-tests.sh
```diff
@@ -0,0 +1,26 @@
...
+sudo ln -sfn /usr/local/opt/openjdk@11/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk-11.jdk
+
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+echo "--- :cocoapods: Setting up Pods"
+install_cocoapods
+
+echo "--- 🧪 Testing"
+xcrun simctl list >> /dev/null
```
AliSoftware, line 24: Might be worth adding a comment for this one (I think I already saw a comment elsewhere, maybe in another PR or repo, explaining that this was kind of a hack to ensure the simulator would kick off or something?)

AliSoftware (approved):
Approving since I think all changes and questions have been addressed.

But I'd love if we could kick-off the `cache-builder.yml` pipeline at least once, to confirm it goes green before merging. I'll let you trigger it once the other CI checks finish

.buildkite/commands/run-ui-tests.sh
jkmassel: This is tracked here: https://github.com/Automattic/buildkite-ci/issues/7 and I'm hoping to move this into the shared actions repo once I figure out the shape of it in the other apps, so with any luck it'll be replaced soon 😁

AliSoftware: Ah, yeah, [this is the other repo where I saw that comment](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/425/files#diff-9be69f7058741a227f07c6ddadd577559a213b90273c85498216c9f0e2607360R12-R13), which I can back-track from [this comment](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/425#discussion_r679601649) which you already reference in that buildkite-ci issue 👍  
The circle is complete ✔️ 🎉

jkmassel:
Had to fix a typo, but it works: https://buildkite.com/wordpress-mobile/wordpress-ios/builds/2013

PR #16949, opened 2021-07-29
Comment: make `rawContent` optional

ScoutHarris:
Ref: #16933, #16876

Self-hosted sites don't have a `rawContent` property, so it is now optional in the Core Data `Comment` model. 

The Comment content for display and for edit now check for either `rawContent` or `content`.

To test:

1. Verify UI tests pass on this PR.

2. On a self-hosted site:
- Log in.
  - Verify the app doesn't crash.
- Go to My Site > Comments.
  - Verify the comments have content.
  - Edit a comment. Verify the content is displayed.

3. On a WordPress site:
- Go to My Site > Comments.
  - Verify the comments have content.
  - Edit a comment. Verify the content is displayed.

## Regression Notes
1. Potential unintended areas of impact
Comment content could be displayed incorrectly.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Checked the Comments display as listed above.

3. What automated tests I added (or what prevented me from doing so)
N/A

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16944, opened 2021-07-29
Comments: fix crashes

ScoutHarris:
Ref: #16876

This fixes the following issues:
- My Site > Comments > where a comment has no body: app crashes.
- My Site > Comments > where a user has no avatar: app crashes.
- My Site > Comments > Comment details > Reply: app crashes.
- Self-hosted > Comments: all the comment bodies are empty.

To test:
- Go to My Site > Comments with comments:
  - That have no body.
  - A commenter has not avatar.
- Verify the app doesn't crash, and the comments appear as expected.
- Reply to a comment. 
- Verify the app doesn't crash and the reply is sent successfully.
- On a self-hosted site, verify the comments have content.

## Regression Notes
1. Potential unintended areas of impact
N/A

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N/A 

3. What automated tests I added (or what prevented me from doing so)
N/A

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16932, opened 2021-07-27
Tracks when the blogging reminders notification is tapped.

diegoreymendez:
Adds tracking for taps to the blogging reminders notifications.

### Setup:

Change `BloggingRemindersScheduler.swift:255` to become:

dateComponents.hour = Date().dateAndTimeComponents().hour
dateComponents.minute = (Date().dateAndTimeComponents().minute ?? 0) + 1

This makes the notification come up 1 minute after it was scheduled.

### Steps:

1. Set a breakpoint in `InteractiveNotificationsManager:235`.
2. Schedule blogging reminders on the same day as the day you're testing (today it'd be Tuesday).
3. Background the App.
4. When the notification comes up, tap on it.
5. When the breakpoint hits, you should see the following tracked event in the console:

:large_blue_circle: Tracked: notification_tapped <notification_type: blogging_reminders>

## Regression Notes

1. Potential unintended areas of impact

None, as the code that was added only adds a tracking call in a very specific place we're directly testing.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
3. What automated tests I added (or what prevented me from doing so)

## PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16948, opened 2021-07-29
Comments: merge fixes from 17.9

ScoutHarris:
Ref: #16944

This merges #16944 into develop.

To test:

- Go to My Site > Comments with comments:
  - That have no body.
  - A commenter has not avatar.
- Verify the app doesn't crash, and the comments appear as expected.
- Reply to a comment. 
  - NOTE: If there is not already a comment from today, the app will crash. The reply is created, so when relaunching the app it won't crash again. Silver lining? I'll fix this separately.
- Verify the app doesn't crash and the reply is sent successfully.
- On a self-hosted site, verify the comments have content.

## Regression Notes
1. Potential unintended areas of impact

2. What I did to test those areas of impact (or what existing automated tests I relied on)

3. What automated tests I added (or what prevented me from doing so)

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16952, opened 2021-07-30
Merge 17.9 beta 1 and release notes into develop

mokagio:
Includes:

- Add event for blogging reminders notification tap (#16932) by @diegoreymendez 
- Fix comment crashes on `.org` sites (#16944) by @ScoutHarris 

I had conflicts in:

- `WordPress/Classes/Models/Comment.swift`
- `WordPress/Classes/Services/CommentService.m`

I solved both conflicts using the version from `develop`. The conflicts were due to the changes coming from #16944, a crash _quick_ fix related to comments and the fact that, after that was merged, @ScoutHarris and @aerych have been working on a _proper_ fix to also address other comments fixes (see the quick fix introduced in 17.8.0.4 by #16906).

See PRs at for the source of the code used for the conflict resolutions:

- #16933
- #16949

@ScoutHarris @aerych if one of you could double check my work, that would be great. Thanks!


mokagio (commented):
- [x] Version update in `.xcconfig`
- [x] Updates to the localization files (`.strings` and Fastlane metadata) automatically pulled for the new translations that have already been approved in GlotPress
- [x] Diffs from the PRs that made it into this beta
- [x] `AppStoreStrings.po` changed with editorialized release notes

WordPress/Resources/bg.lproj/Localizable.strings
```diff
@@ -33,17 +33,19 @@
...
 
-/* Plural format string for view title displaying the number of post likes. %1$d is the number of likes. */
+/* Plural button title to Like a comment. %1$d is a placeholder for the number of Likes.
+   Plural format string for view title displaying the number of post likes. %1$d is the number of likes. */
 "%1$d Likes" = "%1$d Likes";
 
-/* Singular format string for displaying the number of post likes. %1$d is the number of likes. */
-"%1$d blogger likes this." = "%1$d blogger likes this.";
+/* Singular format string for displaying the number of post likes. %1$d is the number of likes. The underscore denotes underline and is not displayed. */
+"%1$d blogger_ likes this." = "%1$d blogger_ likes this.";
```
AliSoftware, line 46: Was about to comment on the underscore at the end of `blogger_` feeling odd until I read the comment explaining it. Still seems odd that there's only one `_` (i.e. that there isn't a **pair** of underscores **surrounding** the text to be underlined) so still wanted to highlight this to double-check

WordPress/Resources/de.lproj/Localizable.strings
```diff
@@ -33,17 +33,19 @@
...
 
-/* Plural format string for view title displaying the number of post likes. %1$d is the number of likes. */
+/* Plural button title to Like a comment. %1$d is a placeholder for the number of Likes.
+   Plural format string for view title displaying the number of post likes. %1$d is the number of likes. */
 "%1$d Likes" = "%1$d Gefällt mir";
 
-/* Singular format string for displaying the number of post likes. %1$d is the number of likes. */
-"%1$d blogger likes this." = "%1$d Blogger gefällt das.";
+/* Singular format string for displaying the number of post likes. %1$d is the number of likes. The underscore denotes underline and is not displayed. */
+"%1$d blogger_ likes this." = "%1$d Blogger_ gefällt das.";
```
AliSoftware, line 45: For my own information/curiosity: not sure why `Blogger` is titlecased here, and noticed other instances like that in the `de` locale (e.g. `Website` on line 1004 of this same file. as well as around L2283-2286); is it some rules in German language that tells that nouns should be titlecased? 🤔

AliSoftware (approved):
- `.po` files both for Jetpack and WordPress match the suggested copy from Editorial
 - Version bumps look correct
 - Strings file updates made sense (left a question but not really related to that PR, more a curiosity that if it needs solving, would be solved in GlotPress anyway, not the PR)

Approving to not block the merge, especially since this PR contains release notes that needs to make it into `develop` before the weekend. But not merging it myself as I think it could be useful to still wait for @ScoutHarris and/or @aerych to validate the conflict resolution you made (even if I agree with your resolution rationale), as you asked in the PR description.

ScoutHarris (approved):
I'm so sorry for all the trouble. I thought I took care of merge conflicts with #16948, but I didn't even consider the things I did _after_ that. Thank you for taking the time to look at what I'd done and fix it properly!

I took it for a spin and 👍 !

:shipit:

WordPress/Resources/bg.lproj/Localizable.strings
ScoutHarris: There is only one because the highlighting starts at the beginning of the string.  So splitting on `_` in this case provides the correct components.

AliSoftware: > There is only one because the highlighting starts at the beginning of the string.  So splitting on `_` in this case provides the correct components. 

@ScoutHarris Well… in English maybe it does. But I don't think it's guaranteed to always be the case in all locales. There might be locales that have this "N bloggers" part of the text in the middle bc grammar and word order in that locale is different than in English. And more commonly, that assumption won't hold for RTL locales, would it?

ScoutHarris: Ah, I see. Coincidentally, more changes are coming to these strings in [#16946](https://github.com/wordpress-mobile/WordPress-iOS/pull/16946). We're still discussing how we want them to look, so I don't want to make too many changes here (conflicts and all that). But I'll take a look and see what I can do for 17.9. Thank you for noticing that!

AliSoftware: Oh yeah we should definitely not block this PR on this and not fix that here (especially because this PR needs to land before this weekend to have the release notes translated in time); I was just mentioning this to get addressed later (and 16946 seems a good one for that 👍)

ScoutHarris: Ah, ok, cool. With/after 16946, I'll make sure the `_`s are correct. Thank you sir!

AliSoftware:
Merging this now that we have validation for the conflict resolution, so that release notes make it to the translation weekly kit before Monday.

mokagio:
Thank you for taking over merging this @AliSoftware !



Team work is the best work.

PR #16946, opened 2021-07-29
Reader Post: Update likes summary text to include like from self

dvdchr:
Refs: `pbArwn-2mK-p2#comment-3436`
Depends on #16936 

## Summary

This change introduces the "You and... " part when the post is liked by self. Here's some screenshots:

Scenario | Screenshot
--|--
The only liker of the post | only_self
Liked by self and one other person | singular_self
Liked by self and several people | plural_self
Liked by one person | singular
Liked by several people | plural

## To Test

- Open a post with only one like from you. Verify that the displayed text is "You like this".
- Open a post with likes from you and one other person. Verify that the displayed text is "You and 1 blogger like this".
- Open a post with likes from you and several people. Verify that the displayed text is "You and N bloggers like this".
- Open a post with likes from one person. Verify the displayed text is "1 blogger likes this". 
- Open a post with likes from several people. Verify the displayed text is "N bloggers like this".  

## Regression Notes
1. Potential unintended areas of impact
None.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
The changes are scoped only to `ReaderDetailLikesView`, which is only used by `ReaderDetailViewController`.

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #16965, opened 2021-08-02
Likes summary formats: mark the beginning of the underlined text

ScoutHarris:
Ref: https://github.com/wordpress-mobile/WordPress-iOS/pull/16952#discussion_r679885122
Branched from: https://github.com/wordpress-mobile/WordPress-iOS/pull/16946

This updates the Likes Summary string formats to clearly mark the beginning of the underlined text. As @AliSoftware pointed out in the referenced PR, simply marking the end of the underlined text may not work in all languages and could be confusing to translators.

Since the strings have been modified with https://github.com/wordpress-mobile/WordPress-iOS/pull/16946, they aren't translated yet. So below I've put test strings you can use to verify highlighting different parts of the strings.

To test:
- Go to Reader > any post with likes.
- Scroll to the likes summary.
  - Verify the highlighting is as expected.
- Change the strings so the highlighted text is in the middle.
  - Verify the highlighting is as expected.
- Change the strings so the highlighted text is at the end.
  - Verify the highlighting is as expected.

---
Underline within the string:

```
static let onlySelf = NSLocalizedString("Test before _You_ like this.",
                                        comment: "Describes that the current user is the only one liking the post.")
static let singularWithSelf = NSLocalizedString("Test before _You and %1$d blogger_ like this.",
                                                comment: "Singular format string for displaying the number of post likes, including the like from self."
                                                    + " %1$d is the number of likes, excluding the like by current user."
                                                    + " The underscore denotes underline and is not displayed.")
static let pluralWithSelf = NSLocalizedString("Test before _You and %1$d bloggers_ like this.",
                                              comment: "Plural format string for displaying the number of post likes, including the like from self."
                                                + " %1$d is the number of likes, excluding the like by current user."
                                                + " The underscore denotes underline and is not displayed.")
static let singular = NSLocalizedString("Test before _%1$d blogger_ likes this.",
                                        comment: "Singular format string for displaying the number of post likes."
                                            + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
static let plural = NSLocalizedString("Test before _%1$d bloggers_ like this.",
                                      comment: "Plural format string for displaying the number of post likes."
                                        + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
```

---
Underline at end of string:

```
static let onlySelf = NSLocalizedString("Test at end _You_",
                                        comment: "Describes that the current user is the only one liking the post.")
static let singularWithSelf = NSLocalizedString("Test at end _You and %1$d blogger_",
                                                comment: "Singular format string for displaying the number of post likes, including the like from self."
                                                    + " %1$d is the number of likes, excluding the like by current user."
                                                    + " The underscore denotes underline and is not displayed.")
static let pluralWithSelf = NSLocalizedString("Test at end _You and %1$d bloggers_",
                                              comment: "Plural format string for displaying the number of post likes, including the like from self."
                                                + " %1$d is the number of likes, excluding the like by current user."
                                                + " The underscore denotes underline and is not displayed.")
static let singular = NSLocalizedString("Test at end _%1$d blogger_",
                                        comment: "Singular format string for displaying the number of post likes."
                                            + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
static let plural = NSLocalizedString("Test at end _%1$d bloggers_",
                                      comment: "Plural format string for displaying the number of post likes."
                                        + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
```
---

## Regression Notes
Likes summary highlighting could be incorrect.

Tested various formats.

N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
@@ -98,11 +98,11 @@ private extension ReaderDetailLikesView {
         case (true, 0):
             summaryLabel.attributedText = highlightedText(SummaryLabelFormats.onlySelf)
         case (true, 1):
-            summaryLabel.attributedText = highlightedText(String(format: SummaryLabelFormats.singularWithSelf, totalLikes))
+            summaryLabel.attributedText = highlightedText(String(format: SummaryLabelFormats.singularWithSelf))
         case (true, _) where totalLikes > 1:
             summaryLabel.attributedText = highlightedText(String(format: SummaryLabelFormats.pluralWithSelf, totalLikes))
         case (false, 1):
-            summaryLabel.attributedText = highlightedText(String(format: SummaryLabelFormats.singular, totalLikes))
+            summaryLabel.attributedText = highlightedText(String(format: SummaryLabelFormats.singular))
         default:
             summaryLabel.attributedText = highlightedText(String(format: SummaryLabelFormats.plural, totalLikes))
         }
@@ -158,21 +158,20 @@ private extension ReaderDetailLikesView {
 
     struct SummaryLabelFormats {
         static let onlySelf = NSLocalizedString("_You_ like this.",
-                                                comment: "Describes that the current user is the only one liking the post.")
-        static let singularWithSelf = NSLocalizedString("_You and %1$d blogger_ like this.",
-                                                        comment: "Singular format string for displaying the number of post likes, including the like from self."
-                                                            + " %1$d is the number of likes, excluding the like by current user."
-                                                            + " The underscore denotes underline and is not displayed.")
+                                                comment: "Describes that the current user is the only one liking a post."
+                                                    + " The underscores denote underline and is not displayed.")
+        static let singularWithSelf = NSLocalizedString("_You and 1 blogger_ like this.",
+                                                        comment: "Describes that the current user and one other user like a post."
+                                                            + " The underscores denote underline and is not displayed.")
         static let pluralWithSelf = NSLocalizedString("_You and %1$d bloggers_ like this.",
-                                                      comment: "Plural format string for displaying the number of post likes, including the like from self."
+                                                      comment: "Plural format string for displaying the number of post likes, including the like from the current user."
                                                         + " %1$d is the number of likes, excluding the like by current user."
-                                                        + " The underscore denotes underline and is not displayed.")
-        static let singular = NSLocalizedString("_%1$d blogger_ likes this.",
-                                                comment: "Singular format string for displaying the number of post likes."
-                                                    + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
+                                                        + " The underscores denote underline and is not displayed.")
+        static let singular = NSLocalizedString("_1 blogger_ likes this.",
+                                                comment: "Describes that only one user likes a post. The underscores denote underline and is not displayed.")
         static let plural = NSLocalizedString("_%1$d bloggers_ like this.",
                                               comment: "Plural format string for displaying the number of post likes."
-                                                + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
+                                                + " %1$d is the number of likes. The underscores denote underline and is not displayed.")
     }
 
     func highlightedText(_ text: String) -> NSAttributedString {
--- a/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
+++ b/WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
@@ -157,34 +157,38 @@ private extension ReaderDetailLikesView {
     }
 
     struct SummaryLabelFormats {
-        static let onlySelf = NSLocalizedString("You_ like this.",
+        static let onlySelf = NSLocalizedString("_You_ like this.",
                                                 comment: "Describes that the current user is the only one liking the post.")
-        static let singularWithSelf = NSLocalizedString("You and %1$d blogger_ like this.",
+        static let singularWithSelf = NSLocalizedString("_You and %1$d blogger_ like this.",
                                                         comment: "Singular format string for displaying the number of post likes, including the like from self."
                                                             + " %1$d is the number of likes, excluding the like by current user."
                                                             + " The underscore denotes underline and is not displayed.")
-        static let pluralWithSelf = NSLocalizedString("You and %1$d bloggers_ like this.",
+        static let pluralWithSelf = NSLocalizedString("_You and %1$d bloggers_ like this.",
                                                       comment: "Plural format string for displaying the number of post likes, including the like from self."
                                                         + " %1$d is the number of likes, excluding the like by current user."
                                                         + " The underscore denotes underline and is not displayed.")
-        static let singular = NSLocalizedString("%1$d blogger_ likes this.",
+        static let singular = NSLocalizedString("_%1$d blogger_ likes this.",
                                                 comment: "Singular format string for displaying the number of post likes."
                                                     + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
-        static let plural = NSLocalizedString("%1$d bloggers_ like this.",
+        static let plural = NSLocalizedString("_%1$d bloggers_ like this.",
                                               comment: "Plural format string for displaying the number of post likes."
                                                 + " %1$d is the number of likes. The underscore denotes underline and is not displayed.")
     }
 
     func highlightedText(_ text: String) -> NSAttributedString {
         let labelParts = text.components(separatedBy: "_")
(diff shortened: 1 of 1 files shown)


AliSoftware:
I'm AFK and not in front of Mac so won't be able to test this until Wednesday at the earliest, but I think it's worth adding a test case to that PR's instruction — and actually why not a unit test as well on the `highlightedText` method! — for the case of a RTL locale. Like what does `first` match means in those cases, and will the assumptions still hold in Right-to-Left?

Also, I have to admit I'm surprised that this `highlightedText` method (1) is located specifically in that file, while it feels it could be a utility method for wider use in multiple places in the app that might need underlines, and (2) that there isn't already some method elsewhere in the codebase with similar logic to build a AttributedString from a localized String with tags/underscores or similar markets? (As I'd expect this kind of use case to already have been needed in the past for other features?)

ScoutHarris:
Hey @AliSoftware .

> I'm AFK and not in front of Mac so won't be able to test this until Wednesday at the earliest

No worries. 


> why not a unit test as well on the highlightedText method

I can look at this separately.

> what does first match means in those cases

:thinking: I thought this was fairly clear. Since `labelParts = text.components(separatedBy: "_")`, `firstPart` is... the first part. Or component if you like.

> will the assumptions still hold in Right-to-Left?

It's hard to verify 100% as the strings aren't translated yet. That's why I added test strings to test splitting in various ways. But it should - the order of the text shouldn't matter as long as the underscores are retained.

> I have to admit I'm surprised that this highlightedText method (1) is located specifically in that file, while it feels it could be a utility method for wider use in multiple places in the app that might need underlines

Generally speaking, I'm not a fan of adding global methods assuming it'll work for all cases. They tend to just get in the way. If it's needed globally/elsewhere it can be modified then. 


> that there isn't already some method elsewhere in the codebase with similar logic to build a AttributedString from a localized String with tags/underscores or similar markets? (As I'd expect this kind of use case to already have been needed in the past for other features?)

I didn't scour the entire codebase, but the one similar case I found was in Auth.

dvdchr (approved):
LGTM for me, with minor optional comments. :shipit:

dvdchr:
Adding my 2¢:

>> what does first match means in those cases
>>
> 🤔   I thought this was fairly clear. Since labelParts = text.components(separatedBy: "_"), firstPart is... the first part. Or component if you like.

My guess is that in RTL language, the `firstPart` here may actually contain the last part of the text (e.g. `c _b_ a`) – so the variable name may be semantically incorrect in RTL context. 

However, we can look at it as "parts of the array" instead of parts of the text. This way, the meaning would still be correct regardless of the order, as it's always the middle part that gets underlined (of course, only in the context of `ReaderDetailLikesView`).

>> I have to admit I'm surprised that this highlightedText method (1) is located specifically in that file, while it feels it could be a utility method for wider use in multiple places in the app that might need underlines
>> 
> Generally speaking, I'm not a fan of adding global methods assuming it'll work for all cases. They tend to just get in the way. If it's needed globally/elsewhere it can be modified then.

Extracting this logic out to a common utility class is a good idea, since it improves testability and reusability. But I think that we should do this in separate iteration, since there are two concerns:

- It's not clear yet who would be using this method.
- The implementation needs to be more "airtight". If we're talking about the `ReaderDetailLikesView`, there's only one part of the text being highlighted, and that highlight will _always_ be in the middle. If we want to make this a common utility, at least we need to handle:
  - Multiple highlighted parts (e.g. `foo _bar_ baz _foo_ bar`)
  - When the underscore isn't ended properly (e.g. `foo _bar_ baz_`)
  - When somebody needs the underscore to be there as part of the text. Imagine we want to highlight a text that contains the username `john_doe`.

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
```diff
@@ -157,34 +157,38 @@ private extension ReaderDetailLikesView {
     }
 
     struct SummaryLabelFormats {
-        static let onlySelf = NSLocalizedString("You_ like this.",
+        static let onlySelf = NSLocalizedString("_You_ like this.",
                                                 comment: "Describes that the current user is the only one liking the post.")
-        static let singularWithSelf = NSLocalizedString("You and %1$d blogger_ like this.",
+        static let singularWithSelf = NSLocalizedString("_You and %1$d blogger_ like this.",
```
dvdchr: I was thinking. Since we know that the number for this is guaranteed to be `1`, do we still need to use format specifier here? Could we just use `_You and 1 blogger_ like this.`?

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
```diff
@@ -157,34 +157,38 @@ private extension ReaderDetailLikesView {
...
                                                         comment: "Singular format string for displaying the number of post likes, including the like from self."
                                                             + " %1$d is the number of likes, excluding the like by current user."
                                                             + " The underscore denotes underline and is not displayed.")
-        static let pluralWithSelf = NSLocalizedString("You and %1$d bloggers_ like this.",
+        static let pluralWithSelf = NSLocalizedString("_You and %1$d bloggers_ like this.",
                                                       comment: "Plural format string for displaying the number of post likes, including the like from self."
                                                         + " %1$d is the number of likes, excluding the like by current user."
                                                         + " The underscore denotes underline and is not displayed.")
-        static let singular = NSLocalizedString("%1$d blogger_ likes this.",
+        static let singular = NSLocalizedString("_%1$d blogger_ likes this.",
```
dvdchr: Same as above comment, could we use `_1 blogger_ likes this.`?

ScoutHarris:
Thanks @dvdchr .

> we can look at it as "parts of the array" instead of parts of the text.

Yea, that clarifies how I look at it. They're just parts of the array. The text order doesn't matter. 

> If we want to make this a common utility, at least we need to handle: 

And this is why I shy away from making common utilities right off the bat. I don't have a crystal ball to see possible future use cases. :smile: There are a lot of assumptions and considerations to make for something that might never be used.

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
ScoutHarris: Sure enough. I'll change it.

AliSoftware: According to common L10n guidelines, you should always use the placeholder though when you need numeric value (unless you spell the number out in all letters).

> If you use a numeric in the format string, as in 1 file remaining for English, it may not be localized when the user changes the region (for example, if the number set changes). Instead, use the format specifier, as in %d file remaining; otherwise, spell out the number, as in One file remaining.
_(Source: [Strings File Format reference](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/StringsdictFileFormat/StringsdictFileFormat.html), amongst others)_

One of the reasons is because the way numeric values are rendered may depend on the user's locale and region settings. For example, there are two types of Arabic Numerals: Eastern Arabic Numerals ( ٠ ١ ٢ ٣ ٤ ٥ ٦ ٧ ٨ ٩ – in RTL) and Western Arabic Numerals (0, 1, 2, 3, 4, 5, 6, 7, 8, 9 – the ones we all know in our western culture). So using a `%1$d` placeholder allows iOS to use the right numeral according to the user's locale/region settings, rather than forcing one system over the other on the user by hardcoding the `1` (and thus forcing Arabic translators to hardcode the `1` too). 🤷

dvdchr: oh TIL. thank you for sharing the reference! I'll note this and propose a change to use the format specifier later on. 🙂

mokagio:
@ScoutHarris @AliSoftware 

> > why not a unit test as well on the highlightedText method
>
> I can look at this separately.

I'm biased, but I'd love to see a test there too. Code that takes a `String` as input and inspect its content is susceptible to all sorts of edge cases. Plus, localization is a tricky subject but also something that is deeply important for WordPress.

Let me know if you want to work on this together.

WordPress/Classes/ViewRelated/Reader/Detail/Views/ReaderDetailLikesView.swift
AliSoftware: Side note: another option, as said in that source, is to spell out the text instead of using a numeric value. For example, `_You and another blogger_ like this`.

dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/16984.

ScoutHarris:
Hey @mokagio . 

> I'd love to see a test there too.

👍 It's on my TODO list! 😄

PR #16967, opened 2021-08-02
Bundle local static React Native assets

dcalhoun:
* **Gutenberg Mobile:** https://github.com/wordpress-mobile/gutenberg-mobile/pull/3810

Now that we attempt to load local static image assets in the React Native bundle in https://github.com/WordPress/gutenberg/pull/33790, we must copy those assets for them to properly show up in the WordPress iOS app build. 

To test: 
1. Use an app build that does not rely upon a local React Native Metro server. 
2. Open the block editor. 
3. Tap the three-dot menu button in the top-right corner. 
4. Tap Help. 
5. Tap one of the available articles.
6. ℹ️ **Expected:** Verify the images are displayed and not missing. 

<details><summary>Screenshot of Help Image</summary>

Gutenberg Help image

</details>

## Regression Notes
  Modifications to the build configuration could cause the build process to fail. 
  Verified the Debug build succeeded. 
  I do not believe this could/should be tested with automation, as build failures should be fairly obvious. 

PR submission checklist:

### DIFF
--- a/WordPress/WordPress.xcodeproj/project.pbxproj
+++ b/WordPress/WordPress.xcodeproj/project.pbxproj
@@ -16572,7 +16572,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


ceyhun (commented):
Thanks @dcalhoun! Tested it with the WordPress target with debug config and it's working great! Do you think maybe we should add the same change to Jetpack target as well?
Also maybe we can trigger the installable builds to verify if it builds correctly on CI with release config for both targets.

dcalhoun:
> Do you think maybe we should add the same change to Jetpack target as well?

I had not considered that. Based on my understanding of how the Jetpack app also consumes Gutenberg, I would imagine we should update Jetpack's target as well. I will do that. Thank you! 

> Also maybe we can trigger the installable builds to verify if it builds correctly on CI with release config for both targets.

Great idea. I started a CI task now to create one.

dcalhoun:
@ceyhun I updated the Jetpack target configuration to match, and verified the images loaded via a build with Xcode. Additionally, I created installable builds for both WordPress and Jetpack; those are attached to this PR. Ready for another review when you have time. Thanks!

ceyhun (approved):
Verified that both builds have the images. Thanks David!

PR #16969, opened 2021-08-03
Comment model: convert properties to Swift

ScoutHarris:
Ref: #16876 

This moves the ObjC Comment properties to `Comment+CoreDataProperties.swift`. Since this is the last remnants of the ObjC class:
- `Comment.swift` is renamed to `Comment+CoreDataClass.swift` and is now the Comment model.
- The old `Comment.m` class is removed.

In addition, these properties have been removed:
- `isNew`: is not in Core Data, and is not used.
- `attributedContent`: is not in Core Data, and is not used.
- `author_ip`: is in Core Data, but is not being populated or used. There is no `author_ip` in the endpoint response, but there is an `ip_address`. I'll leave `author_ip` in Core Data for now, and replace it with `ip_address` in a future change since we'll need it for Unified Comments.

For now, the Swift properties honor the optionality set in Core Data where possible. For Ints and Bools, which cannot be optional, default values are set in `CommentService`. I will see about moving that to Core Data in a follow up.

In `Comment+CoreDataProperties.swift`, the properties are organized a bit abnormally. That is, they're separated by how they're treated in Core Data. I'll organize them better in a future change, but right now this makes it clear what each property is. Primarily so I could clearly identify the `Optional in Core Data with no default values` ones since they're most likely to cause issues. In a future change, I'll look at where we can provide default values for these.

Also in a future change, I'll look at where methods can be replaced with calculated properties now that it's a Swift class.

To test:

Go to My Site > Comments.
- Verify all the comment components are displayed correctly.
  - User avatar
  - Comment title
  - Comment body
- Reply to a comment. Verify it is successful.
- Edit a comment. Verify the comment is updated.
- Mark a comment as Spam or Trash. Delete the comment. Verify it is successful.
- Toggle the `unifiedCommentsAndNotificationsList` feature. Verify the Comments list date sections are correct.
- Repeat the steps above on a self-hosted site.

Go to the Reader > a post with comments.
- View the comment thread. Verify the comment hierarchy is correct (the comments are indented correctly).
- Reply to a comment. Verify the hierarchy is correct.
- (Un)like a comment. Verify the count is updated.

## Regression Notes
Comment content could be displayed incorrectly. Comment actions could misbehave.

Tested functionality and compared with previous app version.

I'll address in a future PR.

PR submission checklist:


ScoutHarris:
> For Ints and Bools, which cannot be optional, default values are set in `CommentService`.

To clarify this more, I've provided default values in code for now. I will see about moving that to Core Data in a follow up.

WordPress/Classes/Models/Comment+CoreDataClass.swift
```diff
@@ -140,7 +135,37 @@ extension Comment: PostContentProvider {
...
+        guard let url = author_url,
+              !url.isEmpty else {
+            return nil
+        }
+        return URL(string: url)
     }
 
 }
+
+// When CommentViewController and CommentService are converted to Swift, this can be simplified to a String enum.
```
dvdchr, line 147: 👍

WordPress/Classes/Services/CommentService.m
```diff
@@ -982,23 +984,24 @@ - (Comment *)createHierarchicalCommentWithContent:(NSString *)content withParent
     return comment;
 }
 
-- (void)setHierarchAndDepthOnComment:(Comment *)comment withParentComment:(Comment *)parentComment
+- (void)setHierarchyAndDepthOnComment:(Comment *)comment withParentComment:(Comment *)parentComment
```
dvdchr, line 987: 👍 !!

dvdchr (approved):
Tested these and everything works. LGTM! :shipit: 

```
Go to My Site > Comments.
- Verify all the comment components are displayed correctly.
  - User avatar
  - Comment title
  - Comment body
- Reply to a comment. Verify it is successful.
- Edit a comment. Verify the comment is updated.
- Mark a comment as Spam or Trash. Delete the comment. Verify it is successful.
- Toggle the unifiedCommentsAndNotificationsList feature. Verify the Comments list date sections are correct.
- Repeat the steps above on a self-hosted site.

Go to the Reader > a post with comments.
- View the comment thread. Verify the comment hierarchy is correct (the comments are indented correctly).
- Reply to a comment. Verify the hierarchy is correct.
- (Un)like a comment. Verify the count is updated.
```

ScoutHarris:
Thanks @dvdchr !

PR #16978, opened 2021-08-04
Reader: Remove obsolete code interfering with content insets.

### LINKED ISSUE #16623: Reader: Part of the post is cut off when rotating device to landscape
**NOTE:** I only managed to reproduce this issue on posts that don't have a featured image.

### Expected behavior
The post should be fully visible.

### Actual behavior
Part of the post (the top area) is cut off after rotating the device to landscape.

1. Go to the reader screen.
2. Open a post that doesn't have a featured image.
3. Rotate the device to landscape.
4. Observe that the top area of the content is cut off by the navigation header.
5. Rotate the device to portrait.
6. Observe that the top area of the content is still cut off by the navigation header.

https://user-images.githubusercontent.com/14905380/120619292-1eefcf00-c45c-11eb-9bdb-3cdd360103bb.mp4

aerych:
Fixes #16623

This PR removes some obsolete code from an earlier version of the reader that was modifying the reader detail's `scrollView`'s `contentInsets` in a way that cropped part of the header view.  The code is no longer needed as featured images are handled differently in the current iteration of the reader.  In the issue description it's reported that posts with featured images weren't affected, however it looks like the cropping was just less obvious.

@fluiddot would you be up for a quick review since you're familiar with the issue? 

To test:
Test on both an iPhone and an iPad form factor.  
Test with a short post, a long post, a pos with a featured image, and a post without a featured image.
- View the post in the reader.
- Rotate the device and confirm that content is no longer clipped at the top of the post. 
- Rotate back and confirm that content is still not clipped.

## Regression Notes
Reader detail presentation

Smoke tested many posts in different orientations.

n/a

PR submission checklist:


fluiddot:
> @fluiddot would you be up for a quick review since you're familiar with the issue?

Sure, I could confirm that the issue is no longer present but I'd like to note that I'm not very familiar with the Reader detail view code in case you'd rather have someone else to double-check the code.

aerych:
> Sure, I could confirm that the issue is no longer present but I'd like to note that I'm not very familiar with the Reader detail view code in case you'd rather have someone else to double-check the code.

Thank ya sir!  No worries.  I'm sure it'll be fine. :)

fluiddot (commented):
Thanks @aerych for addressing the issue 🙇 !

I tested the changes on an iPhone 11 (iOS 14.2) and I noticed that for the case of having a post with a featured image, the post's header is not displayed when rotating the device to landscape, you need to scroll down a bit to see it (see attached video record).

**Behavior with changes from this PR:**
<video src=https://user-images.githubusercontent.com/14905380/128486561-2c5a63d7-eb32-48bf-b33a-b1e932c639ec.MP4>

I think this is a regression because I checked in the last version 17.8 and the behavior is different (see attached video record).

**Behavior on 17.8 production version:**
<video src=https://user-images.githubusercontent.com/14905380/128487014-507aa926-bf14-42d6-891d-6b547cf043b2.MP4>

aerych:
Well spotted @fluiddot  :)  While testing I had only been looking at posts with landscape images.  The one you tried was a portrait image and that made a difference.  

I've updated the patch to preserve the previous logic but only in the case the feature image is not landscape.  The measure I've opted to use is `1.1 > (width / height)`  which allows the insets to be applied to portrait and squarish images but not landscape ones.  I've checked it against portrait, square, and landscape featured images and it seems to be working well for me.  Text only posts also are looking good with this change.  

Ready for another look!

fluiddot:
> Well spotted @fluiddot :) While testing I had only been looking at posts with landscape images. The one you tried was a portrait image and that made a difference.
> 
> I've updated the patch to preserve the previous logic but only in the case the feature image is not landscape. The measure I've opted to use is `1.1 > (width / height)` which allows the insets to be applied to portrait and squarish images but not landscape ones. I've checked it against portrait, square, and landscape featured images and it seems to be working well for me. Text only posts also are looking good with this change.

I'm wondering why we use the value `1.1` and not `1.0`, maybe I'm overlooking something but if we want to adjust the insets only for squareish and portrait, the condition could be `return 1.0 >= (imageSize.width / imageSize.height)`, wdyt?

> Ready for another look!

Thanks @aerych for addressing the issue 🙇 , I'll take another look tomorrow.

fluiddot (commented):
I tested the last changes by using a landscape, portrait, squarish featured image, and now looks fine 🎊 . However, I'd like to check with you that the following behavior is right when setting a landscape featured image on a post:

1. Open the post in portrait mode
2. Rotate the device to landscape mode
3. Observe that the featured image cover half of the screen
Screenshot 2021-08-10 at 11 24 25

4. Navigate back to the Posts screen
5. Open the same post
6. Observe that the featured image now only covers around a quarter part of the screen
Screenshot 2021-08-10 at 11 24 13

_Tested on Simulator iPhone 11 Pro Max (iOS 14.4) and Simulator iPad Pro (12.9-inch) - 4th generation (iOS 14.4)._

aerych:
Heya @fluiddot :wave:.  Sorry it's taken a bit to loop back to this one.  I was enjoying some much needed afk. 

I took another look at this issue with a fresh set of eyes, and I've realized I've been off track on what the proper fix would be. I think it's a case where I just "knew" what the issue was and that made me a bit blind to the actual cause of the clipping.

I'm going to close this PR and branch and open a new one with a clean history that better describes the proper fix.

My apologies for all the back and forth on what should have been a simple patch and thank you for being so thorough with your testing and review!


> I'm wondering why we use the value 1.1 and not 1.0,

No longer needed, but this was to address a case were an image wasn't exactly square, but not quite the dimensions we'd call landscape.  It allowed for a margin of error in the image's dimensions where we'd want to call `updateInitialHeight` under the previous logic.

aerych:
Closing in favor of #17041

fluiddot:
No worries, I hope you enjoyed your AFK 😊 !

Thanks @aerych for working on the fix, I'll take a look at the new PR as soon as I can 👍 .

PR #16980, opened 2021-08-04
Comment entity: set default values on attributes

ScoutHarris:
Ref: #16876

This sets default values on the `Comment` model attributes. 
- The attributes are still optional in Core Data.
- For `Int`s and `Bool`s, removed setting default values in `CommentService`.
- For optional `String`s, the properties in `Comment+CoreDataProperties` are now required.

To test:

My Site > Comments:
- Comment statuses are correct (in the correct filters).
- Comment content is displayed correctly.
  - Author info
  - Post title
  - Body
- Edit a comment. Content is displayed correctly.
- Comments correctly reflect your Like status.
- Long press on a Comment date. Verify `Copy Link to Comment` copies the URL.
- Toggle the `unifiedCommentsAndNotificationsList` feature. Verify the Comments list date sections are correct.
- Repeat the steps above on a self-hosted site.

Reader posts:
- Comment content is displayed correctly.
- Comments correctly reflect your Like status.
- Reader post comments hierarchy is correct.

## Regression Notes
Comment content could be displayed incorrectly.

Tested functionality and compared with previous app version.

I'll address in a future PR.

PR submission checklist:


WordPress/Classes/Models/Comment+CoreDataClass.swift
```diff
@@ -43,69 +43,58 @@ public class Comment: NSManagedObject {
...
     }
 
     func authorName() -> String {
-        guard let authorName = author,
-              !authorName.isEmpty else {
-            return NSLocalizedString("Anonymous", comment: "the comment has an anonymous author.")
+        return !author.isEmpty ? author : NSLocalizedString("Anonymous", comment: "the comment has an anonymous author.")
+    }
+
+    func availableContent() -> String {
```
aerych, line 78: It might be nice to document the behavior of available content.  How it's relevant to the XML-RPC API results from self-hosted sites that don't have raw_content vs wpcom/jetpack sites that do.   I want to say self-hosted sites would have a raw_content from the core rest api, and a note about the approach could be helpful if we ever get to switch away from XML-RPC.

WordPress/Classes/Models/Comment+CoreDataClass.swift
```diff
@@ -43,69 +43,58 @@ public class Comment: NSManagedObject {
...
     public func authorForDisplay() -> String {
-        var displayAuthor = authorName().stringByDecodingXMLCharacters().trim()
-
-        if displayAuthor.isEmpty {
-            displayAuthor = author_email?.trim() ?? String()
-        }
-
-        return displayAuthor
+        let displayAuthor = authorName().stringByDecodingXMLCharacters().trim()
+        return !displayAuthor.isEmpty ? displayAuthor : gravatarEmailForDisplay()
```
aerych, line 101: Nice.

WordPress/Classes/Services/CommentService.m
```diff
@@ -1141,7 +1142,7 @@ - (Comment *)findCommentWithID:(NSNumber *)commentID fromPost:(ReaderPost *)post
 
 - (void)updateComment:(Comment *)comment withRemoteComment:(RemoteComment *)remoteComment
 {
-    comment.commentID = [remoteComment.commentID intValue] ?: 0;
```
aerych, line 1144: I'm a little leery about removing the defaults here.  I know core data _should_ use the default that is declared in the model editor, but RemoteContent still lacks nullibility modifiers and can technically still pass nil as a value.  I'm not quite sure how this would be handled when a property is defined as optional but has a default and we might want to keep these just to be safe.  Thoughts?

aerych (changes requested):
Heya @ScoutHarris 👋 
Smoke tested Reader, My Sites, and Notification comments lists.  Tested editing. Tested filters. Tested likes. Tested tapping the date to get the copy url prompt. Tested swipe behaviors. Tested changing comment status. Tested wpcom and self-hosted sites.  No issues found. Nice. :) 
I had a couple of questions from a read of the changes.  Lemme know what you think. 
The only change I think we for sure would want to look at is checking the box to use a default string for `author_ip` for consistency with other properties.
Screen Shot 2021-08-05 at 10 28 47 AM
Back to you!

ScoutHarris:
> The only change I think we for sure would want to look at is checking the box to use a default string for author_ip for consistency with other properties.

Sorry, I didn't think to mention that I was handling that separately as I had to make changes to WPKit. It's done with #16982 (it's based on this PR).

WordPress/Classes/Models/Comment+CoreDataClass.swift
ScoutHarris: Can do.

WordPress/Classes/Services/CommentService.m
ScoutHarris: I did test with `RemoteComment`s not having those values at all, and it works fine. So I felt comfortable removing them. But if you'd still prefer, I can add them back.

aerych: Nah! I'm being paranoid and our tests and conversation in Slack demonstrate this is fine.  Thanks for talking (and testing) through it!

WordPress/Classes/Models/Comment+CoreDataClass.swift
ScoutHarris: And done.

ScoutHarris:
@aerych back to you please!

aerych (approved):
:shipit: ! Thank you!!

PR #16990, opened 2021-08-06
Extract automation to check translations in dedicated lane

mokagio:
This way, I can run it in isolation to check the progress without the release finalization pre-checks overhead.

To test:

1. Checkout this branch
2. Run `bundle exec fastlane check_all_translations` and verify it checks all translation in one go

check_not_interactive

1. Repeat adding the `interactive:true` parameter and verify it asks whether to continue after each check (assuming there are values below the threshold)

check_interactive

## Regression Notes
N.A.

N.A.

N.A.

---

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -269,6 +269,38 @@ platform :ios do
   end
 
   #####################################################################################
+  # check_all_translations
+  # -----------------------------------------------------------------------------------
+  # This lane checks the translation progress for all the projects in GlotPress.
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane check_all_translations
+  #
+  # Example:
+  # bundle exec fastlane check_all_translations
+  #####################################################################################
+  desc 'Check translation progress for all GlotPress projects'
+  lane :check_all_translations do
+    UI.message('Checking app strings translation status...')
+    check_translation_progress(
+      glotpress_url: 'https://translate.wordpress.org/projects/apps/ios/dev/',
+      abort_on_violations: false
+    )
+
+    UI.message('Checking WordPress release notes strings translation status...')
+    check_translation_progress(
+      glotpress_url: 'https://translate.wordpress.org/projects/apps/ios/release-notes/',
+      abort_on_violations: false
+    )
+
+    UI.message('Checking Jetpack release notes strings translation status...')
+    check_translation_progress(
+      glotpress_url: 'https://translate.wordpress.com/projects/jetpack/apps/ios/release-notes/',
+      abort_on_violations: false
+    )
+  end
+
+  #####################################################################################
   # finalize_release
   # -----------------------------------------------------------------------------------
   # This lane finalize a release: updates store metadata, bump final version number,
@@ -286,24 +318,8 @@ platform :ios do
     UI.user_error!('To finalize a hotfix, please use the finalize_hotfix_release lane instead') if ios_current_branch_is_hotfix
 
     ios_finalize_prechecks(options)
-    
-    UI.message('Checking app strings translation status...')
-    check_translation_progress(
-      glotpress_url: 'https://translate.wordpress.org/projects/apps/ios/dev/',
-      abort_on_violations: false
-    )
 
-    UI.message("Checking WordPress release notes strings translation status...")
-    check_translation_progress(
-      glotpress_url: 'https://translate.wordpress.org/projects/apps/ios/release-notes/',
-      abort_on_violations: false
-    )
-
-    UI.message("Checking Jetpack release notes strings translation status...")
-    check_translation_progress(
-      glotpress_url: 'https://translate.wordpress.com/projects/jetpack/apps/ios/release-notes/',
-      abort_on_violations: false
-    )
+    check_all_translations
 
     download_localized_strings_and_metadata(options)
     # FIXME: (2021.06.17) This is disabled because we currently have a >256 chars string which GlotPress truncates when exporting  the `.strings` files,
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -272,31 +272,41 @@ platform :ios do
   # check_all_translations
   # -----------------------------------------------------------------------------------
   # This lane checks the translation progress for all the projects in GlotPress.
+  #
+  # By default, it will print all values. You can make it ask whether to continue when
+  # if found translations that are below the threshold by passing `interactive: true`.
   # -----------------------------------------------------------------------------------
   # Usage:
-  # bundle exec fastlane check_all_translations
+  # bundle exec fastlane check_all_translations [interactive:<interactive>]
   #
   # Example:
   # bundle exec fastlane check_all_translations
+  # bundle exec fastlane check_all_translations interactive:true
   #####################################################################################
   desc 'Check translation progress for all GlotPress projects'
-  lane :check_all_translations do
+  lane :check_all_translations do |options|
+    abort_on_violations = false
+    skip_confirm = options.fetch(:interactive, false) == false
+
     UI.message('Checking app strings translation status...')
     check_translation_progress(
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -269,41 +269,67 @@ platform :ios do
...
-  # bundle exec fastlane finalize_release
-  # bundle exec fastlane finalize_release skip_confirm:true
+  # bundle exec fastlane check_all_translations
+  # bundle exec fastlane check_all_translations interactive:true
   #####################################################################################
-  desc 'Trigger the final release build on CI'
-  lane :finalize_release do |options|
-    UI.user_error!('To finalize a hotfix, please use the finalize_hotfix_release lane instead') if ios_current_branch_is_hotfix
+  desc 'Check translation progress for all GlotPress projects'
+  lane :check_all_translations do |options|
```
mokagio, line 287: The thought occurred to me to make this a bit more organized so that one could check the individual progress.

At this time, I don't see the need for checking individual projects. So it would be a "nice to have" that would bring unnecessary overhead. Besides, at that point, it wouldn't be much different than calling `check_translation_progress` manually, just with less typing.

fastlane/Fastfile
```diff
@@ -269,41 +269,67 @@ platform :ios do
...
+    abort_on_violations = false
+    skip_confirm = options.fetch(:interactive, false) == false
```
mokagio, line 289: `interactive` made more sense to me as a parameter name, but I can revert to the standard (?) `skip_confirm` if you feel it would be more appropriate.

What I like about `interactive` is that is a positive assertion, whereas `skip_confirm` is negative.

- "Should I show the confirm message?" "Yes" "Okay I'll show it"
- "Should I show the confirm message?" "No" "Okay I won't show it"

versus:

- "Should I _not_ show the confirm message?" "Yes" "Okay I won't show it"
- "Should I _not_ show the confirm message?" "No" "Okay I'll show it"

fastlane/Fastfile
```diff
@@ -269,41 +269,67 @@ platform :ios do
...
-  # bundle exec fastlane finalize_release skip_confirm:true
+  # bundle exec fastlane check_all_translations
+  # bundle exec fastlane check_all_translations interactive:true
   #####################################################################################
-  desc 'Trigger the final release build on CI'
-  lane :finalize_release do |options|
-    UI.user_error!('To finalize a hotfix, please use the finalize_hotfix_release lane instead') if ios_current_branch_is_hotfix
+  desc 'Check translation progress for all GlotPress projects'
+  lane :check_all_translations do |options|
+    abort_on_violations = false
```
AliSoftware, line 288: 💄  Any reason to create a variable with that hardcoded value as opposed to just using `false` inline at call site on line 294? 🤷  (reminds me of [this tweet thread by Nick](https://twitter.com/nicklockwood/status/1421047660006494209?s=20))

fastlane/Fastfile
```diff
@@ -269,41 +269,67 @@ platform :ios do
...
+  desc 'Trigger the final release build on CI'
+  lane :finalize_release do |options|
+    UI.user_error!('To finalize a hotfix, please use the finalize_hotfix_release lane instead') if ios_current_branch_is_hotfix
+
+    ios_finalize_prechecks(options)
+
+    check_all_translations(interactive: true)
 
     download_localized_strings_and_metadata(options)
     # FIXME: (2021.06.17) This is disabled because we currently have a >256 chars string which GlotPress truncates when exporting  the `.strings` files,
```
AliSoftware, line 335: 📝 Unrelated to this, but we should really address that FIXME at some point, in order to re-enable translation linting, especially since this is an easy fix (just split the `NSLocalizedString` that is too long in the codebase into 2 shorter ones). That would allow us to go back to ensure security at runtime (avoiding risk of crashes due to incorrect placeholders in translations)

AliSoftware (approved):
- ✅  Run `bundle exec fastlane check_all_translations` and verify it checks all translation in one go
- ✅ Repeat adding the `interactive:true` parameter and verify it asks whether to continue after each check (assuming there are values below the threshold)

Works as advertized. :shipit: 


Two things unrelated to that PR but while I'm at it:
 - I'm surprised that in the GIFs you posted in your PR description, when the release toolkit detects that you're on 1.3.1 while there's a 1.4.0 version, **it doesn't stop with a prompt to ask you if you want to update or not**, like it's supposed to do (and like it did during my tests of this PR on my machine). Is it something that you disabled explicitly on you end, is it something that is not working on your machine like expected?!
 - We should definitively address (in a separate PR) the FIXME about the localization linting being disabled in WPiOS, as it's not reassuring that localizations haven't been linted for a while in WPiOS just because of one string being too long

mokagio:
Thanks @AliSoftware! Regarding the missing update prompt, I exported `CHECK_FOR_TOOLKIT_UPDATES_SKIP_UPDATE_SUGGESTION=1` to make my feedback loop faster while testing.

Maybe it might be less confusing, albeit less informative, to skip the entire update check flow if the parameter is `true` as opposed to showing there's a new version but [not prompting the user to update it for them](https://github.com/wordpress-mobile/release-toolkit/blob/e02e57d58986967bda8e59a5ca681e629cf7fbd3/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/check_for_toolkit_updates_action.rb#L29)?

fastlane/Fastfile
mokagio: Good point. I was considering passing that parameter too, so I put it in a var in the meantime. Then, I decided not to, but left it there.

Nick's got a point in favoring inline literals. Although, I'd say in this case I like having the value extracted in a single place because, if we want to toggle it to `true`, we'll need to change only one line. I'm also aware that "if we want to..." is a smell for code that YAGNI. 💄 😅

AliSoftware:
> Maybe it might be less confusing, albeit less informative, to skip the entire update check flow if the parameter is true as opposed to showing there's a new version but not prompting the user to update it for them?

The rationale for that env var read by the _fastlane action_ only skipping the suggestion (but still doing the check) was that, if you don't want to do the check _at all_, you can implement the opt-out in your `Fastfile` yourself by just not calling the action. While if you want to do the check but skip the prompt, you couldn't modify the action from the `Fastfile` if that skip parameter didn't exist.

(Also, my thinking was that we don't want to skip the check completely, at least not every time, and it wouldn't make sense to have a `ConfigItem(env_name: 'SKIP_CHECK_FOR_TOOLKIT_UPDATES', …)` in the action to skip the _whole_ update check, as it would end up doing nothing and ultimately for that behavior it'd be simpler to not call the action at all)

But if you feel you sometimes need to skip the check, feel free to modify your `Fastfile` to something like:
`check_for_toolkit_updates unless ENV['SKIP_CHECK_FOR_TOOLKIT_UPDATES'] == '1'`. (I just wouldn't recommend you to set that env var in your `.zshrc` but instead only use `SKIP_CHECK_FOR_TOOLKIT_UPDATES=1 bf <lane>` on demand, otherwise that kinda defeats the point of the update check being introduced in the first place 😄 )

fastlane/Fastfile
mokagio: > I'd say in this case I like having the value extracted in a single place because, if we want to toggle it to `true`, we'll need to change only one line

I was reading through the Twitter thread, and saw [this tweet](https://twitter.com/nicklockwood/status/1421047662590234627?s=20) in the thread that says something similar:

> Constants serve two functions:
>
> 1) giving context to something that is otherwise unclear (e.g. a magic number)
>
> 2) **allowing a value to be re-used in multiple places without repetition**
>
> If a constant serves neither of those purposes, it's redundant. It makes the code worse.

AliSoftware: Yeah I only realized _after_ I wrote that comment that it was used in 3 call sites (the 3 translation checks) below, not just the one on line 294, and realized then why it could make sense to extract it after all, in order to toggle it more easily for all 3 at once, so all good 👍

PR #16991, opened 2021-08-06
Comments: Improved deletion failure.

### LINKED ISSUE #525: Error message is not shown if something went wrong during the deletion of a comment
Reported by @daniloercoli 
- Go to the comments list
- Tap on a comment
- Disable the XML-RPC services on your blog, or shutdown the server (, or reply with an invalid response)
- Delete the comment

Error message is not shown to the user, and the comment you tried to delete is not in the list. You need to refresh the comments list to see it again.

Original trac thread [here](http://ios.trac.wordpress.org/ticket/1049)

aerych:
Closes #525
This PR attempts to improve the experience a user has when deleting a comment fails. As described in #525, the experience has been rather poor for self-hosted sites.  Recent changes to comments has improved the experience as an error is now shown but the comment detail remains in a broken state.  This PR dismisses the comment detail after the deletion failure and preserves the error message shown. 

I'd tried a few other approaches before settling on this. I explored waiting to delete the comment until success, but this interfered with swipe to delete. I tried rebuilding the deleting comment on failure but detail remained in a broken state as the rebuilt comment was a different instance than the deleted/faulted one. I tried immediately dismissing the detail screen but this resulted in a race condition between the delete call, and the call to refresh the list. (The list _must_ be refreshed to reflect status changes.)  Where I've ended up seems like the best compromise under the circumstances.

Example:
Screen Recording 2021-08-06 at 3 21 47 PM 2021-08-06 15_31_08

@ScoutHarris, would you kindly? 

To test:
Repeat the steps described in the issue.  To disable XML-RPC, I used [this plugin](https://wordpress.org/plugins/disable-xml-rpc/).

## Regression Notes
n/a

n/a

n/a

PR submission checklist:


ScoutHarris:
Hey @aerych .

I did the following:
- Disabled XMLRPC (thank you for the plugin tip).
- Deleted a comment.
- The details did dismiss, and the error was shown on the list view.

However, after the comment details is dismissed:
- The loading indicator is shown, not spinning.
- The comment is no longer in the original filter (in my case, All). Nor is it in Trashed (it shouldn't be of course).
- After I re-enabled XMLRPC, the list refreshed and the comment reappeared.

I tried on both iPhone and iPad, the same behavior happens on both.

Simulator Screen Shot - iPhone 12 Pro Max - 2021-08-06 at 15 04 49

aerych:
Howdy @ScoutHarris 👋 

I've updated the patch to restore a deleted comment in the event of an API failure. Hopefully this should make for a better visual experience for the user.  Good call on pointing out that bit of oddness. 

For the stuck spinner, I've poked at it a bit and I'm thinking it might be existing behavior. I can reproduce the glitch in the current beta by disabling XML-RPC, pulling to refresh the comment list, tapping to view a comment detail, then tapping back to the list.  My guesses about this is that either we're not seeing a network error propagated in a timely fashion (maybe AFNetworking is waiting for a timeout?), or we have a case were a failure block is not called that would reset the refresh control. It looks like it can be resolved by calling [self refreshPullToRefresh]; from viewWillAppear so I've added that call.

Ready for another peek!

ScoutHarris (approved):
Nice! LGTM!

:shipit:

aerych:
Thank you!!

PR #16994, opened 2021-08-09
Merge 18.0.0.0 into develop

mokagio:
Code freeze + release toolkit update


fastlane/Pluginfile
```diff
@@ -6,7 +6,7 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 3.2.0'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', '~> 1.3.1'
+gem 'fastlane-plugin-wpmreleasetoolkit', '~> 1.3'
```
mokagio: By removing the constraint on _patch_ version compatibility, the auto-updater in the toolkit is  able to update to the latest version.

As a side note, I think we should update this rule only when the automation uses a feature that is available in a new version, in which case, we should update to `~> <new version major>.<new version minor>`.

mokagio (commented):

fastlane/Pluginfile
loremattei: Hmm... I don't think this is the best way to use the optimistic operator. 

We create a patch version when we want to fix a bug. This means, for example in this case, that `1.3` has a bug which `1.3.1` fixes. By removing the patch number, we basically rely only on `Podfile.lock` when we run `pod install`. 

We have had cases, even recently in production, where this caused regressions because of bad merges of `Podfile.lock` where `git` picked the unpatched version during the merge. This can easily happen and it's pretty difficult to spot during a review (reviewing merges in `Podfile.lock` can be much more difficult that in `Podfile`).

So, since we run `pod install` when we build all our release binaries, I would suggest that we always use the minimum required version of the pod with the optimistic operator. 

So, `~> '1.3'` seems risky in this case, because it could cause `1.3` to be installed if there's an issue with `Podfile.lock`. I would use `~> 1.4` which would pick any future `1.x`, until we have bug fix.  Witch patches, I'd suggest to use the minimum version of the patch that we need.

loremattei (dismissed):

fastlane/Pluginfile
mokagio: Good point @loremattei, thanks for catching that.

While both `~> 1.3.1` and `~> 1.3` would result in `1.3.1` being installed from scratch, there could be scenarios in which `1.3.0` would end up used.

Also `~> 1.3.1` clearly communicates that there was a bug in `1.3.0`.

If we need or want to update to `1.4.0` and therefore need to modify the version requirement, `~> 1.4` works better.

mokagio: P.S. you mentioned `Podfile`s and `pod install`, but this is the Fastlane plugin, which Bundler and `Gemfile.lock` to resolve.

But all the considerations do apply. In fact, I'm pretty sure CocoaPods and Bundler use [the same underlying library](https://github.com/CocoaPods/Molinillo) to do dependency resolution. 👍

mokagio: @loremattei I implemented your suggestion in d2129bc. Could I have another review? Thanks!

loremattei: Yup! You're right! I got confused!
Thanks for implementing the change! 🙇‍♂️

loremattei (approved):

Issue #16995, opened 2021-08-09
Add buttons to recommend the app to friends


dvdchr:
Refs: pctCYC-9r-p2

Update the Me screen and About screen to show a button that enables user to recommend the WordPress app to their friends.

- [x] https://github.com/wordpress-mobile/WordPressKit-iOS/pull/433
- [x] Add feature flag. (Added in #17006.)
- [x] #17002
- [x] #17003
- [x] #16997
- [x] #16996
- [x] Review accessibility
- [x] #17027

Design preview:
<img width="1442" alt="recommend-app" src="https://user-images.githubusercontent.com/1299411/128719159-a25a4018-eae3-4027-a6ae-452b4f194ce9.png">

PR #17005, opened 2021-08-11
Recommend App: Add button with icon cell component

### LINKED ISSUE #17002: Create cell component for the button with icon style
Parent issue: #16995

Create a new table view cell that fulfills the design requirement. Here's the design of the cell:

cell

Some exploratory notes:
- **Option 1:** Reuse `ButtonRow`, which is currently used for login and logout button in the Me screen.
  - The class is an `ImmuTableRow` type, which is constraint for `ImmuTable` – while the About page uses plain table view. We'd need to directly use the table view cell class `WPTableViewCellDefault` and replicate the cell configuration done in `ButtonRow`.
  - `WPTableViewCellDefault` uses plain `UITableViewCell`. This means, the button is actually the default `textLabel` of `UITableViewCell` with center alignment.
  - Another problem with using a default `UITableViewCell` is the icon requirement – the icon needs to be on the leading side of the title label, and the default `imageView` is displayed on the leading edge of the cell.
- **Option 2:** Create a new table view cell with a single `UIButton` element. The `UIButton` has an `imageView` component built-in that displays the image on the leading side of the button and it follows to the tint color.
  - Problem is, `UIButton` absorbs touch event, preventing the cell to trigger `tableView(_ tableView:didSelectRowAt indexPath:)`.
  - There's also the issue with the unconfigurable spacing between the button's image and title. There are some workarounds, but needs more testing for localization compatibility (particularly RTL layout).
  - Potential multiline + accessibility issue.
- **Option 3:** Create a new table view cell with a `UIImageView` and `UILabel` pair, replicating the looks of `UIButton` when it's image is set.
  - Need to ensure that it has the right accessibility traits and label, since it's a label that's presented as a button.
- Lastly, since the contents for the project are fetched asynchronously, the button should be able to show a loading state to indicate an asynchronous activity.

dvdchr:
Refs #16995
Closes #17002

## Summary
This change introduces a new `SingleButtonTableViewCell`, which looks like a single `UIButton` 😅 . There are several approaches explored (see #17002), but I ended up going with a plain `UIImageView` and `UILabel`. I've tried hard to make plain old `UIButton` work for this, but here are the reasons why I decided not to:
- `UIButton` absorbs touch events, which prevents `didSelectRow` from being triggered. 
  - There's a workaround for this by disabling the user interaction for the `UIButton`.
- The feature that will be using this class requires an icon on the leading side of the button. `UIButton` already has this built-in, but the spacing cannot be adjusted.
  - ...But of course, there's always a [workaround](https://noahgilmore.com/blog/uibutton-padding/). This modifies the `titleEdgeInsets` and `contentEdgeInsets` with a combination of positive and negative values, but it can work. Furthermore, it needs some extra adjustments for RTL because `UIEdgeInsets` are still using `left` and `right` values.
- Finally, the multi-line label support. I've found out that there are currently two approaches for this: using a [MultilineButton](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Views/MultilineButton.swift) (which overrides intrinsic size calculation), or pin the button's title label to the button's edges. [WooCommerce uses this approach](https://github.com/woocommerce/woocommerce-ios/blob/66881a554821da1820938cb6f64ead404e44d5b9/WooCommerce/Classes/Extensions/UIButton%2BHelpers.swift#L137-L144). These solutions work, but only if we're not displaying the button's image. I've tried various things, but couldn't get it to work just right.

Anyways, I've spent enough time on this and decided to proceed with `UILabel` and `UIImageView` for now. 😅 . Here's how it looks:

Description | Screenshot
--|--
The design | cell_design
Normal | normal
RTL language | rtl
Large accessibility size | large

## To Test
Since the component is not hooked up anywhere, you can only test this manually by modifying the code. Here's what I did in `AboutViewController`. First, register the cell in the `setupTableView()` method.
```swift
tableView.register(SingleButtonTableViewCell.defaultNib, forCellReuseIdentifier: SingleButtonTableViewCell.defaultReuseID)
```

Second, in the `cellForRow` method, we're gonna sneakily swap out the last cell with the new one:
```Swift
// ...
let row = rows[indexPath.section][indexPath.row]

if indexPath.section == rows.count - 1,
   indexPath.row == rows[indexPath.section].count - 1,
   let buttonCell = tableView.dequeueReusableCell(withIdentifier: SingleButtonTableViewCell.defaultReuseID, for: indexPath) as? SingleButtonTableViewCell {
    buttonCell.title = "Share WordPress with a friend"
    buttonCell.iconImage = .init(systemName: "square.and.arrow.up")
    buttonCell.tintColor = .primary

    return buttonCell
}

// ...
```

Lastly, go to the About screen and verify that the button is displayed correctly.

## Regression Notes
None.

The changes are not reachable, since the component is still unused.

n/a

PR submission checklist:


Podfile
```diff
@@ -47,7 +47,7 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.39.0'
+    pod 'WordPressKit', '~> 4.40.0-beta'
```
aerych, line 50: Was this supposed to be part of this PR?  I'm missing how WordPressKit would be used.

WordPress/Classes/ViewRelated/Cells/SingleButtonTableViewCell.swift
```diff
@@ -0,0 +1,123 @@
...
+        iconImageView.tintColor = tintColor
+    }
+}
+
+// MARK: - Private Helpers
+
+private extension SingleButtonTableViewCell {
+
+    func setupViews() {
+        selectionStyle = .none
```
aerych: Should selectionStyle be .none? It seems like we would still want to show the tap?

aerych (approved):
Howdy @dvdchr 👋 
Looks good!  Thanks for the code snippets in the PR description, they made testing easier. :) 
I just had two questions about the code that might be good to consider.  Otherwise :shipit:

Podfile
aerych: Oh. I see that it's used in https://github.com/wordpress-mobile/WordPress-iOS/pull/17006 and that PR is both branched from and targets this one. Let's just roll with it. :)

WordPress/Classes/ViewRelated/Cells/SingleButtonTableViewCell.swift
dvdchr: You're right! Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17005/commits/81dcbfcebf6b367495598a2f28f85436c25e135b.  🙂

PR #17006, opened 2021-08-11
Recommend App: Add presenter to encapsulate display logic

### LINKED ISSUE #17003: Add component to handle content fetching and presentation logic
Parent issue: #16995

The logic to present fetch the content & present the share sheet may be reused multiple times in different places. Instead of having to write the same logic over and over, it would be convenient to encapsulate this in a single component, so callers can simply use this object if they want to trigger the share app flow.

The share/recommend flow can currently be triggered from the About screen and the Me screen. However, in the future we may want to expand this to contextual entry points (e.g. after posting my first blog post). For more details, see `pctCYC-9r-p2`.

I'm thinking these are the responsibilities that will be handled by the component:
- [x] Request the content from `ShareAppContentServiceRemote`.
- [x] Construct the activity items required for `UIActivityViewController`, format contents based on the proper activity types, and finally present it.
- [x] Show a notice if something goes wrong.
- [x] Maybe cache the content in-memory so subsequent taps don't cause unnecessary roundtrips?
- [ ] Tracks -> will be handled in a separate PR.

### Flow
Screen Shot 2021-08-12 at 17 59 52

dvdchr:
Refs #16995
Closes #17003
Based on #17005 

## Summary
This change introduces a presenter object that encapsulates content fetching and presentation logic for the share app flow, along with a subclass of `UIActivityItemSource` to help format the text contents depending on the activity type. As described in #17003, I think it's worth it to encapsulate this since it's possible to have multiple entry points in the future.

Some notes for this change:
- The presenter exposes a delegate that gets called on every loading state changes. This is useful for the caller to update their display to show e.g. a loading indicator while the content is being fetched.
- Other than formatting needs, a `UIActivityItemSource` subclass is necessary to provide additional information – specifically a subject string provided through `activityViewController:subjectForActivityType`. This information is used when sharing to mail. I'll share some screenshots below on how it looks when it's shared in various apps.
- Lastly, this adds a new feature flag for the recommend button.

## Screenshots

Share Destination | Screenshot
--|--
Messages | share_app_messages
Mail | share_app_mail
Twitter | share_app_twitter
WordPress (Save as Draft) – **Note:** Not sure why the text contents are not carried, but I suspect this needs to be fixed from the WP side. | share_app_wp_draft
Copy (Pasteboard) | share_app_copyToPasteboard
Chrome | share_app_chrome
Slack – Not sure why the text is not present, but this is usually due to the destination's Share Extension not picking up multiple activity items. | share_app_slack
WhatsApp | share_app_whatsapp

### Manually editing the source + compiling to device
Since there are no visual components hooked up to this component, we'll have to wire this up manually in Xcode AND compile it to a device 😅 . 

Here's how I modified the code to test the share functionality:

- In the `MeViewController`, add a private instance variable like so:
  ```Swift
  private let sharePresenter = ShareAppContentPresenter(account: nil)
  ```
- Find the `pushHelp()` method, comment out the code in the return block, and change it to this:
  ```Swift
  return { [unowned self] row in
      defer { self.tableView.deselectSelectedRowWithAnimation(true) }
      guard let indexPath = self.tableView.indexPathForSelectedRow,
            let selectedCell = self.tableView.cellForRow(at: selectedIndexPath) else {
          return
      }
      sharePresenter.present(for: .wordpress, in: self, sourceView: selectedCell)
  }
  ```
- Since there's no loading indicator yet, just wait for one or two seconds and the share sheet should appear shortly after tapping the help button on the Me screen.
- Lastly, try logging in to a self-hosted site (without WP account) and make sure that you can still prompt the share sheet.

### Unit tests
Make sure that the tests are 🟢  !

## Regression Notes
None.

This component is not used yet.

Added some unit tests for the presenter.

PR submission checklist:


WordPress/Classes/ViewRelated/Sharing/ShareAppTextActivityItemSource.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    func activityViewController(_ activityViewController: UIActivityViewController, subjectForActivityType activityType: UIActivity.ActivityType?) -> String {
+        return .defaultSubjectText
+    }
+}
+
+// MARK: - Localized Strings
+
+private extension String {
+    static let defaultSubjectText = NSLocalizedString("WordPress.com Apps - Apps for any screen",
```
dvdchr: @aerych, @mattmiklic are we good to go with this subject text? This text is used as the email subject line when sharing the app via Mail.app.

(Also, worth noting that Gmail.app somehow ignores this value, and takes the text content instead. I'm suspecting it's a bug from their side. 😅 )

mattmiklic: This is fine with me unless @aerych wants to wordsmith it. 😄

aerych: Maybe `WordPress Apps` rather than `WordPress.com Apps`?  If we omit the .com it might sound a tad more inclusive for self-hosted/Jetpack sites.

aerych (approved):
Howdy @dvdchr 👋 
I tested on my iPhone X and everything worked great.  Thanks for showing all the examples in the PR description. It was nice to have something to compare my results to. :) 
Code looks good to me. I had just one suggestion about the subject text. 
Otherwise :shipit: :)

WordPress/Classes/ViewRelated/Sharing/ShareAppContentPresenter.swift
```diff
@@ -0,0 +1,121 @@
...
+            return
+        }
+
+        let activityItems = [
+            ShareAppTextActivityItemSource(message: content.message) as Any,
+            linkURL as Any
+        ]
+
+        let activityViewController = UIActivityViewController(activityItems: activityItems, applicationActivities: nil)
+        viewController.present(activityViewController, animated: true, completion: nil)
```
aerych, line 113: I approved too soon 😭.  I neglected to test on the iPad and went back to do that.  On the iPad the activity controller is presented as a popover.  We'll need to define a source view or source rect for the activityViewController's popoverPresentationViewController or else we'll encounter an exception: 
> Terminating app due to uncaught exception 'NSGenericException', reason: 'UIPopoverPresentationController (<UIPopoverPresentationController: 0x7ff5e70296a0>) should have a non-nil sourceView or barButtonItem set before the presentation occurs.'

Something like `activityViewController.popoverPresentationController?.sourceView = viewController.view ` might be good enough. (Probably needs to be before the call to `.present`.)

aerych (changes requested):
Spotted an issue so I've added another note and marked this request changes since we need to address an exception.  Sorry I missed it in the first round!

WordPress/Classes/ViewRelated/Sharing/ShareAppTextActivityItemSource.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17006/commits/b963748036929e771fd66ee58a740f63515e84ef.

WordPress/Classes/ViewRelated/Sharing/ShareAppContentPresenter.swift
dvdchr: Thank you for noticing this. I completely missed this one! This is addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17006/commits/a9bb473e77a0c7a120ff11fe5475eae7e5d357f6. Tested in iPad simulator and it now works as expected.

aerych (approved):
Heya @dvdchr 👋 
The exception is resolved. Thanks for the changes! Marking approved so as to not block the rest of the work.  

However, I think we might want to see if we can improve the popover presentation on iPad. The solution I suggested was something I hadn't tested, and now that I see it in action I think we might need to make an adjustment.   

Depending on which way I orient the iPad, the popover appears away from where I tapped. The size also varies depending on orientation so sometimes there are more visible options than others.  I'm wondering if sharing the frame of the cell (or maybe just the cell as the view) that was tapped rather than the controller view would help improve the experience.  We'd also need to make the change before the code freeze.  What do you think? 
Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-08-18 at 11 50 02
Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-08-18 at 11 50 05

dvdchr:
Hey @aerych. I wasn't very aware of the design guidelines for iPad before, so thank you for the help and tips on this. Really appreciate it! 🙂

> Depending on which way I orient the iPad, the popover appears away from where I tapped. The size also varies depending on orientation so sometimes there are more visible options than others. I'm wondering if sharing the frame of the cell (or maybe just the cell as the view) that was tapped rather than the controller view would help improve the experience.

Noted. I just realized that since the popover is anchored to the view controller's view, there's not enough space on the side and the system puts the popover on the top side, which makes the content a bit cramped. 

Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17006/commits/6a07b6d63227fdc970c3be61220d54903d9e5fae, I've confirmed that passing the cell as the `sourceView` correctly anchors the popover to the tapped cell, and indeed the share sheet is displayed better. Here's a preview:

Portrait | Landscape
--|--
ipad_portrait | ipad_landscape

> We'd also need to make the change before the code freeze. What do you think?

Since the code review for #17007 and #17020 is blocked by this one, I'll apply the changes on this PR. Then, I'll merge up the changes to both #17007 and #17020, so you can proceed with reviewing both of them. 

I'll aim to merge everything to `develop` tomorrow (provided that you approved them all, of course 😄 ). If there's any new major issue on the PRs, I will address it on a separate PR (since the feature flag is still `false`).

---
P.S.: I've updated the code sample in the description to fill the `sourceView` parameter if you decided to give this another look! 🙂

aerych:
Tested with the changes and they look great!  Thanks @dvdchr!  :shipit: again!

dvdchr:
Thanks @aerych !

PR #17016, opened 2021-08-12
Integrate gutenberg-mobile release 1.58.3

geriux:
## Description
This PR incorporates the 1.58.3 release of gutenberg-mobile.
For more information about this release and testing instructions, please see the related Gutenberg-Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3828

Release Submission Checklist

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -165,7 +165,7 @@ abstract_target 'Apps' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :commit => 'ce5bc935b393a93d31b8352a83608f466fb28b41'
+    gutenberg :tag => 'v1.58.3'
 
     ## Third party libraries
     ## =====================
--- a/Podfile
+++ b/Podfile
@@ -165,7 +165,7 @@ abstract_target 'Apps' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :tag => 'v1.58.2'
+    gutenberg :commit => 'ce5bc935b393a93d31b8352a83608f466fb28b41'
 
     ## Third party libraries
     ## =====================
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -62,7 +62,7 @@ PODS:
     - AppAuth/Core (~> 1.4)
     - GTMSessionFetcher/Core (~> 1.5)
   - GTMSessionFetcher/Core (1.6.1)
-  - Gutenberg (1.58.2):
+  - Gutenberg (1.58.3):
     - React (= 0.64.0)
     - React-CoreModules (= 0.64.0)
     - React-RCTImage (= 0.64.0)
@@ -425,7 +425,7 @@ PODS:
     - React-Core
   - RNSVG (9.13.7-wp):
     - React-Core
-  - RNTAztecView (1.58.2):
+  - RNTAztecView (1.58.3):
     - React-Core
     - WordPress-Aztec-iOS (~> 1.19.4)
   - Sentry (6.2.1):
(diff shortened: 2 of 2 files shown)


fluiddot:
> I have considered if this change warrants user-facing release notes and have added them to RELEASE-NOTES.txt if necessary.

Probably we should update the release notes in this PR and include a section for `17.9.1`, wdyt?

geriux:
> > I have considered if this change warrants user-facing release notes and have added them to RELEASE-NOTES.txt if necessary.
> 
> Probably we should update the release notes in this PR and include a section for `17.9.1`, wdyt?

After some internal discussions, we don't need to update the release notes. 👍

fluiddot (approved):
LGTM 🎊 !

**NOTE:** This PR shouldn't be merged until the Gutenberg Mobile reference points to the `v1.58.3` tag.

mokagio:
@geriux @fluiddot 

> After some internal discussions, we don't need to update the release notes. 👍

Could you still suggest some copy for the release notes? It doesn't have to be in the code, a comment here is fine, I'll take care of pasting it where appropriate.

After a brief look at the PR and the crash on Sentry, the only thing I can come up with is:

> Fixed an occasional crash in the block editor.

Is there something more specific to add? Thanks! 😄 

cc @startuptester

fluiddot:
Yeah, the crash is hard to describe because although we have been able to reproduce it, it's yet not clear the flow that users are following to bump into it.

In the release notes of GB-mobile I added the note:
> Fix crash when pasting OBJECT REPLACEMENT CHARACTER special character.

But it's probably too technical to be used for the app's release notes, so the message you added it's a good one 👍 :
> Fixed an occasional crash in the block editor.

AliSoftware:
> Could you still suggest some copy for the release notes? It doesn't have to be in the code, a comment here is fine, I'll take care of pasting it where appropriate.

@mokagio The issue with hotfixes and release notes is that, given the time-sensitive nature of hotfixes, even if we skip the step of making them go thru Editorial, we still won't have time to make them go thru Mag16 translations. So we'd only have English release notes for hotfixes. This is also why we don't add any in those cases of hotfixes.

Also, by nature again, hotfixes usually happen quite soon after the last release (to fix crashes that happened right after a release, otherwise we'd just wait for the next one) which means we usually want to keep the release notes of the last release (otherwise people getting the hotfix e.g. only 1-2 days after the release wouldn't even had time to see the list of changes).  
So if we were to provide info about what was fixed by the hotfix in the release notes, we'd need to _add_ that item to the existing release notes (i.e. in this case add that "Fixed an occasional crash in the block editor" entry to the existing 17.9 release notes, instead of replacing them completely with just that entry). Which might not be an issue for iOS, but raises an issue with Android, where the character limit for release notes there is pretty small and we're usually already pretty close to that limit with the release notes of the last release (e.g. 17.9) to have enough room to ad new items to it for 17.9.1 at that point. That part might not apply to WPiOS since we don't have such a limit, but overall that was just one more item to add to the rationale about why we don't bother adding release notes to hotfixes after all.

mokagio:
> we usually want to keep the release notes of the last release.

Uhm... Okay, makes sense but I wasn't aware of it 😅 

I would have simply pasted the English copy in all the locales, which is not ideal but felt "accurate" in terms of what the release does.

I'm see the value in using the same release notes as the base release, though. Happy to go with it. Thank you for pointing it out. 😄 🙇‍♂️

AliSoftware:
Yeah using the English copy in all locales would work… if we replaced the release notes of 17.9 with the new notes of 17.9.1. But since I think(†) it makes sense to keep the release notes of 17.9 due to the timing, that means adding release notes for 17.9.1 on top of the ones for 17.9 would not be compatible with this idea of using English copy for all locales… because then that would mean that e.g. Italian release notes would have a mix of Italian copy for 17.9 + the additional English item for 17.9.1…

This is why it's just simpler to not provide release notes for hotfixes and reuse the one from the corresponding release instead

(†) Note that I don't think there's an official rule that we wrote down about this, so that's just me writing out loud what I _think_ the rationale is and why _I_ think it makes sense… but I don't think it's officially documented anywhere…

mokagio:
> This is why it's just simpler to not provide release notes for hotfixes and reuse the one from the corresponding release instead

Indeed 👍

PR #17017, opened 2021-08-12
Integrate gutenberg-mobile release 1.59.1

dcalhoun:
## Description
This PR incorporates the 1.59.1 release of gutenberg-mobile.
For more information about this release and testing instructions, please see the related Gutenberg-Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/3830

Release Submission Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17025, opened 2021-08-13
Merge hotfix 17.9.1 into trunk

mokagio:
Includes:

- Update `mobile-gutenberg` to 15.8.3 to address a new crash (#17016) by @geriux 

Notice that I also added a couple of commits to update the release toolkit to 1.4.0 to use the `skip_deliver` flag when creating the hotfix in order to avoid a tooling crash because the repo doesn't have a `Deliverfile` anymore. That's a fix that already in `develop`, but past the time `17.9` went in code freeze.

Also notice there is no matching merge into `release/18.0`. Rationale for that [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/17017#issuecomment-898204355).


mokagio (commented):
Super straightforward hotfix merge:

- [x] Version bump in `.xcconfig`
- [x] No localizations updates because this is an hotfix and we don't want to wait for them to be ready
- [x] Commits from the PRs that fixed the issue(s) the hotfix addresses

AliSoftware (approved):
Diff and rationale make sense to me :shipit: 

> Also notice there is no matching merge into release/18.0. Rationale for that here.

Discussed in p8Qyks-1p6-p2#comment-516 since, so the corresponding merge was done via https://github.com/wordpress-mobile/WordPress-iOS/pull/17026 (which targets `develop` tho not `release/18.0`, but in that case I think that works too)

PR #17026, opened 2021-08-13
Merge hotfix 17.9.1 into develop

mokagio:
I got conflicts:

  - `Gemfile.lock`
  - `Podfile`
  - `Podfile.lock`
  - `config/Version.internal.xcconfig`
  - `config/Version.public.xcconfig`

I resolved the `.xcconfig` conflicts using `git checkout --theirs`, because, on `develop` we are already on version 18.0 and there is no value in going behind in the version numbering there.

Similarly, I resolved the `Podfile` and `Podfile.lock` conflicts using `--theirs`. It's a bit trickier in this instance, though.

Picking up `develop`'s version of the pods discards the 1.59.1 hotfix, which is not ideal. On the other hand, the alternative would have been losing all the changes that went from 1.59.0 to 1.60.0-alpha.1.

I saw there is already work to integrate the 1.59.1 hotfix into mobile-gutenberg's `develop`, wordpress-mobile/gutenberg-mobile#3837, so I assume the hotfix from 1.59.1 will soon make its way into WordPress iOS's `develop`, too, in the form of a new alpha release for mobile-gutenberg 1.60.0.

Looking the resulting diff, I think my choices were appropriate. @dcalhoun, as the author of the mobile-gutenberg PR I linked, could you confirm my rationale for picking 1.60.0-alpha.1 is solid? Thanks


AliSoftware (approved):
I agree with the rationale on the conflict resolution. :shipit: 

Makes sense to pick develop in those cases, and indeed makes more sense to pick the latest 1.60 alpha for Gutenberg providing the likeliness (and actually need on their end, anyway) of a new 1.60 alpha that will integrate that 1.59.3 fix :+1

dcalhoun:
> Picking up `develop`'s version of the pods discards the 1.59.1 hotfix, which is not ideal. On the other hand, the alternative would have been losing all the changes that went from 1.59.0 to 1.60.0-alpha.1.

> Looking the resulting diff, I think my choices were appropriate. @dcalhoun, as the author of the mobile-gutenberg PR I linked, could you confirm my rationale for picking 1.60.0-alpha.1 is solid? Thanks

@mokagio I concur. I believe your choices are the best approach. Thank you! 🙇🏻 

> I saw there is already work to integrate the 1.59.1 hotfix into mobile-gutenberg's `develop`, [wordpress-mobile/gutenberg-mobile#3837](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3837) [...]

I don't believe the following is material, but I figured I'd provide the nuance and clarity for posterity. 

The "hot fix" for WP 17.9.1 was managed in GB Mobile [1.58.3](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3828). That fix has [already merged](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3832) into gutenberg-mobile's `develop` branch. 

The GB Mobile [1.59.1 PR](https://github.com/wordpress-mobile/gutenberg-mobile/pull/3837) you mentioned is a "beta fix" for the WP 18.0 release. It just so happens that it also includes (via cherry picking) the aforementioned hot fix. 

> [...] I assume the hotfix from 1.59.1 will soon make its way into WordPress iOS's `develop`, too, in the form of a new alpha release for mobile-gutenberg 1.60.0.

You are correct that the next Gutenberg Mobile alpha integrated will land the hot fix into WordPress-iOS' `develop` branch. We generally only integrate alpha releases when we need to make modifications to the bridge methods, so they do not happen on a specific frequency. It will, of course, definitely land whenever the official 1.60.0 release integrates 20 Aug. That said, we could make an intentional alpha sooner if we believe it is helpful.

mokagio:
Thanks for the extra info @dcalhoun 🙌 

> That said, we could make an intentional alpha sooner if we believe it is helpful.

Personally, I don't think there's any rush. As far as I understand it, the crash that 1.58.3 addresses doesn't really affect the main flows of the app, so it shouldn't impact the day to day development and testing. If my hunch is correct, then it makes sense to wait till the schedule integration day and focus on higher ROI work 👍

PR #17029, opened 2021-08-13
Task/16941 - Close account

### LINKED ISSUE #16941: Account deletion
A new [change in App Store Review Guideline](https://developer.apple.com/news/?id=dovxb62h) adds this update:

> 5.1.1(v): Apps supporting account creation must also offer account deletion.

In order to conform to this, we need to implement account deletion on our app.

In Calypso we have the following rule to allow account deletion: https://github.com/Automattic/wp-calypso/blob/b550e845f46f9c93641f54193d73626cb5b669e9/client/me/account-close/main.jsx#L90

The endpoint for account deletion is `v1.1/me/account/close`

nikola-milicevic:
Fixes [#16941](https://github.com/wordpress-mobile/WordPress-iOS/issues/16941)
WPKit: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/435

This PR includes UI + logic for closing user account.

## Note
- We rely on API to [determine](https://github.com/Automattic/wp-calypso/blob/b550e845f46f9c93641f54193d73626cb5b669e9/client/me/account-close/main.jsx#L90) if account deletion is possible

## Tasks
- [x] Implement design
- [x] Implement close account logic
- [x] Error handling in case account deletion is not possible

## To test the happy path:
1. Create a temporary user account that will be deleted
2. Use the newly created account to log in to the app
3. Under **My Site** tab tap on profile button at the top right corner
4. Tap on **Account Settings**
5. Tap on **Close account**
6. Close account alert is displayed
7. Confirm close account by entering your username
7. Tap on **Permanently Close Account**
8. After the account has been closed, user is redirected the the welcome screen

## Happy path flow
https://user-images.githubusercontent.com/12729242/129917471-e974745d-efec-425d-bab1-d747b3c340b8.mp4

## To test the error:
1. Create a temporary user account that will be deleted
2. Create a test site using newly created user
3. Add a yearly Premium plan subscription to test site
4. Follow the steps 2-7 from the happy path
5. Error message is displayed

## Error alert
Screen Shot 2021-08-19 at 10 18 52

## Regression Notes
n/a

n/a

n/a

PR submission checklist:


momo-ozawa:
@nikola-milicevic @osullivanchris 

What do you think of displaying the **Close account** row in the same style as the **Log out** / **Clear ...** rows?
Since **Close account** is a destructive action, it would be nice to make that row visually distinct.

Simulator Screen Shot - iPhone 12 Pro - 2021-08-16 at 09 55 52 | Simulator Screen Shot - iPhone 12 Pro - 2021-08-16 at 09 58 37
-- | --

nikola-milicevic:
> What do you think of displaying the **Close account** row in the same style as the **Log out** / **Clear ...** rows?
> Since **Close account** is a destructive action, it would be nice to make that row visually distinct.

@momo-ozawa Yes, totally agree, even maybe new section (at the bottom of the list).

momo-ozawa (commented):
@nikola-milicevic Closing an account works well!

I wonder if we should add a loading indicator after confirming **Close account** to make the app feel more responsive.


https://user-images.githubusercontent.com/6711616/129547814-6c2ccc85-17ca-4f37-ba21-ca3ba4fa806d.mp4

nikola-milicevic:
> I wonder if we should add a loading indicator after confirming **Close account** to make the app feel more responsive.

Absolutely. This is temporary design, which doesn't include loading indicator while account is being deleted, and some UI / UX changes are needed. I guess @osullivanchris will take a look at that.

osullivanchris:
> Since Close account is a destructive action, it would be nice to make that row visually distinct.

👍  to this. I agree. Let's also break it into a separate section as @nikola-milicevic suggested.

> I wonder if we should add a loading indicator after confirming Close account to make the app feel more responsive.

Looking at that, I agree. For now, you could just put a native spinner on top of the view while it is happening.

I did have a quick look at a flow for this. Its a loose wireframe but I can do more work on it if we prefer this direction. It is a bit more heavy handed than what you've done, so maybe its too much. Let me know what you think. (and cc @leandroalonso )

1. Account delete button in same place as proposed
2. Rather than showing an alert, another full screen modal appears. It shows which account you are deleting, explains what will happen, and tells you how many sites you have (could even have some way to see a list of your sites)
3. After tapping confirm, a loading state displays in place of the buttons (like we have done with Jetpack restore site feature for example)
4. When the site is deleted, a confirmation message is shown. This reassurance would be nice to me as a user. If I just land on the login screen I might wonder hmm was my account definitely deleted, or have I just logged out? When user taps on 'Done' it would bring them to the prologue screen logged out as you have at the end of your flow.


Screenshot 2021-08-16 at 17 43 15

leandroalonso:
@osullivanchris I like the "account deleted" screen to confirm that the account was indeed deleted, but showing a new screen with info, list of sites, etc is a bit too involved for a feature that doesn't actually bring a lot of value. What do you think about doing the same as we do when deleting a site: display an alert stating what will happen and ask the user to confirm by entering their username.

osullivanchris:
@leandroalonso  oh cool, I hadn't gone through that flow before! Noting that we do actually show a screen there with the info, similar to what I was thinking about.

IMG_2039

But if we want to get this out quicker that's fine too. I understand that the user value is very low here. Getting them to type their username seems like a nice safety step here too, I would do that.

We could also use this existing loading state, and then kick them out to the prologue screen as soon as its complete

IMG_2043

leandroalonso:
> We could also use this existing loading state, and then kick them out to the prologue screen as soon as it's complete

We can also display the toast with a message that the account was deleted.

osullivanchris:
Hey again @nikola-milicevic just wanted to close the loop here with a clear proposal! 

I wanted to have a little bit more user confirmation for the action, but I agree with Leandro that we don't want to build anything too complex for this flow. I like his idea of using what we have in site deletion. That way we get an extra bit of certainty from the user, without having to build lots of new views.

What do you think of this? 

- It uses the same mechanism where the user enters their username to make absolutely sure they intend to delete the account. 
- I am then also using the same loading and confirmation states from site deletion so hopefully you can re-use code from there!
- I showed the confirmation toast appearing on logout, but wherever it appears is fine.

delete account flow

nikola-milicevic:
Hey @osullivanchris 👋 Thanks! I think it's clear, with the right amount of information, and precaution step which is great. I would just use the word "close" instead of "delete" account, just to be consistent with the web version, if you agree.

osullivanchris:
> Hey @osullivanchris 👋 Thanks! I think it's clear, with the right amount of information, and precaution step which is great

Nice! Glad its working for you!

> I would just use the word "close" instead of "delete" account, just to be consistent with the web version, if you agree.

Sounds good to me!

WordPress/Classes/ViewRelated/Blog/Site Management/DestructiveAlertHelper.swift
```diff
@@ -0,0 +1,56 @@
+import UIKit
+
+protocol DestructiveAlertHelperLogic {
+    var valueToConfirm: String? { get }
+    var alert: UIAlertController? { get }
+    func makeAlertWithConfirmation(title: String, message: String, valueToConfirm: String, destructiveActionTitle: String, destructiveAction: @escaping () -> Void) -> UIAlertController
+}
+
+class DestructiveAlertHelper: DestructiveAlertHelperLogic {
```
momo-ozawa, line 9: Thank you for making this code reusable!

What do you think defining this logic in a UIAlertController extension (e.g. in `UIAlertController+Helpers`, or a new `UIAlertController+Destructive` etc)?

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -1419,6 +1419,8 @@
 		93C882A31EEB18D700227A59 /* plugin_redirect.html in Resources */ = {isa = PBXBuildFile; fileRef = 93C8829A1EEB18D700227A59 /* plugin_redirect.html */; };
 		93C882A41EEB18D700227A59 /* rsd.xml in Resources */ = {isa = PBXBuildFile; fileRef = 93C8829B1EEB18D700227A59 /* rsd.xml */; };
 		93CD939319099BE70049096E /* authtoken.json in Resources */ = {isa = PBXBuildFile; fileRef = 93CD939219099BE70049096E /* authtoken.json */; };
+		93CDC72126CD342900C8A3A8 /* DestructiveAlertHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 93CDC72026CD342900C8A3A8 /* DestructiveAlertHelper.swift */; };
+		93CDC72226CD342900C8A3A8 /* DestructiveAlertHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 93CDC72026CD342900C8A3A8 /* DestructiveAlertHelper.swift */; };
```
momo-ozawa: Seems like there's a duplicate entry...? 🤔

momo-ozawa (commented):
@nikola-milicevic Looking great, works as described! 🎉 

Something you may already be aware of -- before you merge this PR into `develop`, you'll want to publish a new beta version on WPKit 👍

WordPress/Classes/ViewRelated/Blog/Site Management/DestructiveAlertHelper.swift
nikola-milicevic: Well I think it would be great to have `convenience init` inside an extension instead of `make` method of a helper class, but I think it's not doable, since we need something to encapsulate the value which needs to be confirmed (and pass it to `textFieldDidChange` in order to validate the input). So I think a custom `UIAlertController` subclass and a helper class (which I opted for) would work in this case.

WordPress/WordPress.xcodeproj/project.pbxproj
nikola-milicevic: Yeah, not sure what's going on there. I noticed that, but I thought it's okay. I tried adding any file anywhere, the result is the same. Maybe @leandroalonso could clarify if this is actually okay or not.

WordPress/Classes/ViewRelated/Blog/Site Management/DestructiveAlertHelper.swift
momo-ozawa: > I think it's not doable, since we need something to encapsulate the value which needs to be confirmed (and pass it to textFieldDidChange in order to validate the input). So I think a custom UIAlertController subclass and a helper class (which I opted for) would work in this case.

Ahh ok, I understand now!

nikola-milicevic:
I think the only thing that's missing here now is the error localization in case account closing is not possible. The server returns a `String` message, and currently I simply propagate the `error.localizedDescription` to the `UIAlertController` message. I think the best way to solve such problem is to receive a code from the server which we can use to map to our local localized constants, but not sure if that's possible at the moment.
@momo-ozawa @leandroalonso let me know what you think, and what would be the best way to tackle this.

momo-ozawa (commented):
Error alert is working as described!

> The server returns a String message, and currently I simply propagate the error.localizedDescription to the UIAlertController message. I think the best way to solve such problem is to receive a code from the server which we can use to map to our local localized constants, but not sure if that's possible at the moment.
@momo-ozawa @leandroalonso let me know what you think, and what would be the best way to tackle this.

I think we can map the `error.userInfo[WordPressComRestApi.ErrorKeyErrorCode]` and display the appropriate localized error message.

Here are a couple error codes we could map:
- `unauthorized`
- `atomic-sites`
- `chargebacked-site`
- `active-subscription`

Curious to hear your thoughts too, @leandroalonso  🙇‍♀️

nikola-milicevic:
> I think we can map the `error.userInfo[WordPressComRestApi.ErrorKeyErrorCode]` and display the appropriate localized error message.

@momo-ozawa thanks for the suggestion! 😊 I totally overlooked `WordPressComRestApi.ErrorKeyErrorCode`, I think it's a great fit for this case, so I updated the code.

WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
```diff
@@ -273,11 +274,32 @@ private class AccountSettingsController: SettingsController {
             case .failure(let error):
                 SVProgressHUD.dismiss()
                 DDLogError("Error closing account: \(error.localizedDescription)")
-                self?.showErrorAlert(message: error.localizedDescription)
+                let message = NSLocalizedString(self.generateLocalizedMessage(error), comment: "Error message displayed when unable to close user account.")
```
momo-ozawa: I'm not sure if `Localizable.strings` will pick up the error messages if we pass `generateLocalizedMessage` as a param?

WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
```diff
@@ -228,6 +239,75 @@ private class AccountSettingsController: SettingsController {
...
+        }
+    }
+
+    private func generateLocalizedMessage(_ error: Error) -> String {
+        let userInfo = (error as NSError).userInfo
+        let errorCode = userInfo[WordPressComRestApi.ErrorKeyErrorCode] as? String
+
+        switch errorCode {
+        case "unauthorized":
+            return "You're not authorized to close the account."
```
momo-ozawa: I think you can do something like this instead:

```suggestion
        case "unauthorized":
            return NSLocalizedString("You're not authorized to close the account.", comment: "Error message displayed when unable to close user account due to being unauthorized.")
```

WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
leandroalonso: It won't, we need to use the actual strings inside `NSLocalizedString`.

momo-ozawa (approved):
LGTM!

Issue #16087, opened 2021-03-12
Plugins: horizontal scrolling doesn't work and nothing from the Featured section can be selected


startuptester:
### Expected behavior
I expect to be able to scroll horizontally through the rows and be able to select a plugin from the row of Featured ones

### Actual behavior
**1. Scrolling: I cannot scroll horizontally in any row
2. Featured row: they all seem to be disabled
3. Popular row: It opens a list view when I select one, I don't think that is right**

![IMG_9FF046E363EF-1](https://user-images.githubusercontent.com/66104034/110991929-79cccb00-8332-11eb-9bb7-380dc4ad2d13.jpeg)

### Steps to reproduce the behavior
With a jp connected site, open plugins from My Site
Tap a plugin from the Installed row and note that it opens the Manage screen
Tap a plugin from the Popular row and note that it opens a list view - I expect to see a detail view of that plugin though?
Tap a plugin from the Featured row: nothing happens
Notice that none of the rows scroll horizontally

iPhone 12 Pro, iOS 14.4, WPiOS 16.6-16.9 (I don't think it's version specific)

yaelirub:
@Gio2018 , @frosty this has been broken for a while and I think we should give it a priority. Last relevant changes I see are here : https://github.com/wordpress-mobile/WordPress-iOS/pull/15588

Gio2018:
> @Gio2018 , @frosty this has been broken for a while and I think we should give it a priority. Last relevant changes I see are here : #15588

Thanks for the heads-up @yaelirub , we'll give priority to this issue.

frosty:
FYI I have a fix up for this now: https://github.com/wordpress-mobile/WordPress-iOS/pull/16538

PR #17031, opened 2021-08-13
Fixes a scrolling issue in the plugins VC.

diegoreymendez:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/16087

This time for good.  The culprit was found.

It seems [we were adding an extra container view in a very hackish way](https://github.com/wordpress-mobile/WordPress-iOS/blob/c57edd505c96e20f3f957e8d424a982437f141f2/WordPress/Classes/ViewRelated/Plugins/CollectionViewContainerRow.swift#L140-L148), but were not removing it ever, causing it to intercept all of the gestures that were intended for the collection view below.

## To test:

Test with an iPhone and an iPad.

1. Try loading the plugins view while offline, and make sure the no-results view looks good.
2. Go back online and tap "try again" a few times.  Make sure the plugins show up (it can take a few attempts until WPiOS realizes it's online).
3. Also try installing the App from scratch, as that was the plugins scrolling was broken on the first time the plugins section was opened.

## Regression Notes

There's a change the no-results logic could be broken, but I tested it as part of this PR.  Test the plugins section for a site with no connection and reconnect.  Make sure everything works.

Manually tested them.

None.

PR submission checklist:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 
 18.0
 -----
+* [*] Fixed a bug that would make it impossible to scroll the plugins the first time the plugin section was opened.
 * [*] Resolved an issue where authentication tokens weren't be regenerated when disabled on the server. [#16920]
 * [*] Updated the header text sizes to better support large texts on Choose a Domain and Choose a Design flows. [#16923]
 * [internal] Made a change to how Comment content is displayed. Should be no visible changes, but could cause regressions. [#16933]
--- a/WordPress/Classes/ViewRelated/Plugins/CollectionViewContainerRow.swift
+++ b/WordPress/Classes/ViewRelated/Plugins/CollectionViewContainerRow.swift
@@ -134,24 +134,21 @@ class CollectionViewContainerCell: UITableViewCell {
             oldValue?.removeFromView()
 
             guard let noResultsView = noResultsView else {
-                    return
+                return
             }
 
-            let containerView = UIView(frame: collectionView.frame)
-            containerView.translatesAutoresizingMaskIntoConstraints = false
-
             noResultsView.view.backgroundColor = .clear
-            noResultsView.view.frame = containerView.frame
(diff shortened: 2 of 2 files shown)


diegoreymendez:
Fyi @startuptester - We may want to test this fix.

In case you want to reproduce the original issue, the plugins scrolling was originally broken the first time data was loaded for it.  My mechanism to reproduce was always a fresh install of the app, and head into the plugins for a site.

Gio2018 (approved):
Thank you @diegoreymendez !! :shipit:

mokagio:
@diegoreymendez  this has been bundled as part of [18.0 beta 2 (18.0.0.2)](https://github.com/wordpress-mobile/WordPress-iOS/pull/17047).

Thanks for your work 🙌

startuptester:
Thnks @diegoreymendez and @mokagio! 

I tested this in 18.0.0.2 using iPhone 12 (14.7.1) 
- all rows scroll in portrait mode
- i can install and uninstall
- scrolling works after returning to plugins screen from details screen
- works in dark mode and light mode 
- **only top row scrolls in landscape mode**

startuptester:
Here's a video of the landscape issue
https://cloudup.com/cvcrguUiT_R


Note: NOT a blocker for the 18.0 release since this is still an iterative improvement over before.

mokagio:
> Note: NOT a blocker for the 18.0 release since this is still an iterative improvement over before.

+1

Besides, we have ~24 hours before finalizing the build. Adding more code to the branch would be risky 😄

diegoreymendez:
@startuptester - The reason why it's not  scrolling is because it's showing up to the last available plugin so there's nothing to scroll to.

You can verify this by seeing what the last plugin is for each row in the first part of the video (portrai mode), and then comparing to the last plugin shown for the same row in the second part of the video (landscape mode).

The only "issue" I think would be that we'd like it to bounce as tables do in iOS I guess - an aesthetic issue.  Thoughts?

PR #17024, opened 2021-08-13
Blogging Reminders - Time Selection UI

Gio2018:
Fixes #NA

To test:
Test on iPhone and iPad
- Build/run and go to Site Settings -> Blogging Reminders
- Tap on at least one day in the settings screen and make sure that the time selection button shows up (see screenshot below)
- Tap on the button and make sure that the time selection view shows up (see screenshots below for iPhone and iPad)
- tap on Back and make sure you return correctly to the settings screen

**NOTE: this PR does not actually set the Blogging Reminders time, at this time. It only contains the UI**

Time Selection Button | Time Selection View - iPhone | Time Selection View - iPad
-- | -- | --
 |  | 

## Regression Notes
1. Potential unintended areas of impact
This adds new types and a new element in Blogging Reminders view. The only potential regression could be UI bugs in blogging reminders

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Test Blogging Reminders UI on both iPhone and iPad

3. What automated tests I added (or what prevented me from doing so)
none
PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17033, opened 2021-08-14
Blogging Reminders - Time Selection

Gio2018:
Fixes #NA

To test:
use an account with more than one site
- Before running this branch, build and run another branch (e.g. `develop`) and set some Blogging Reminders to some of the sites (go to Site Settings -> Blogging Reminders, select some days and tap on "Notify Me")

Repeat the following both on a site with pre-existing reminders and on one that did not have reminders
- Build/run this branch then go to Site Settings -> Blogging Reminders
- if you selected a site with previously set reminders, make sure the reminders still show up.
- Select one or more days: this should make the time selection button appear. If no time was selected before, the default time will be 10:00 AM
- tap on the time selection button, you should see a time picker (see #17024 for details on the UI)
- select a time when to schedule a notification (tip: select a time in the near future, e.g. the next minute or two in the same day of the week as today, to avoid waiting too long for the reminder to show up)
- Make sure the title of the time picker screen updates with the selected time
- Tap "Back" and make sure the time in the selection button updates with the time you selected
- Tap "Notify Me" (or "Update, if you changed an existing schedule) and then "All Set" on the epilogue screen (alternatively, just dismiss the epilogue screen)
- Make sure the console logs the following tracking event:

`🔵 Tracked: blogging_reminders_scheduled <blog_type: wpcom, days_of_week_count: 1, selected_time: 12:49>`

- Wait for the notification to show up at the time you selected, and make sure the reminder shows up
- Optionally, go in the tracks live view, wait (about 5 mins) for the above event (`wpios_blogging_reminders_scheduled`) to show up and make sure the `selected_time` property is consistent with what you've set (the time will be standard 24h `HH:mm`)

## Regression Notes
This change refactors a few parts of Blogging Reminders scheduler and changes the format that was previously used for reminders, so it's important to ensure backward compatibility

Install this branch on top of an old one that used the previous scheduling format; tested that notifications are actually arriving at the desired time

none

PR submission checklist:


WordPress/Classes/ViewRelated/Blog/Blogging Reminders/Time Selector/Date+BloggingReminders.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+extension Date {
```
emilylaguna: Since this is used in the `TimeSelectionView` which is decoupled from the Blogging reminders, I'm wondering if this needs to have the `+BloggingReminders` in the name, could we make this more generic? 

Or make it private to blogging reminders so others don't accidentally use it.

emilylaguna (changes requested):
Heya, I tested using the steps in the PR and it worked as expected. 

**Question** Should my atomic site should appear as a `self hosted` site in the event?

`🔵 Tracked: blogging_reminders_screen_shown <blog_type: self_hosted, screen: day_picker>`

**UI issues though:**

- On an iPhone 11 Pro Simulator when I tap on a day and the time shows, the UI shifts and then cuts off the `Tip` title. 

https://user-images.githubusercontent.com/793774/129628466-777cb2b5-afd3-4c14-8600-97fde8d48078.mov

- On iPad: The same issue as above occurs however on first click the text gets set to single line mode.  Also when tapping on the time section, the push has no animation and oddly snaps to the next view. 

https://user-images.githubusercontent.com/793774/129628553-11cba289-ad63-4714-a83a-f5024090bc80.mov

Gio2018:
Thanks for reviewing @emilylaguna !!

>  Question Should my atomic site should appear as a self hosted site in the event?
> 🔵 Tracked: blogging_reminders_screen_shown <blog_type: self_hosted, screen: day_picker>

this is how currently sites are categorized on WPiOS (I figure because atomic are not actually hosted on the shared wp.com), so that's expected

As for the UI Issues, thanks so much for pointing those out! I was planning to have another UI polishing PR after this, just to keep this one focused on the functionality. Does that work?

WordPress/Classes/ViewRelated/Blog/Blogging Reminders/Time Selector/Date+BloggingReminders.swift
Gio2018: That's right, I'll make it more generic

emilylaguna (approved):
This works well for me.

PR #16962, opened 2021-08-02
UC: Add extra top space to Notifications and Comments list

dvdchr:
Refs #16738, https://github.com/wordpress-mobile/WordPress-iOS/pull/16885#issuecomment-883824926

Based on the suggestion from @ScoutHarris and @mattmiklic, this PR adds an extra 10pt of space on the top of the first headers of unified list. Here's how it looks:

  | Top position | Scrolled position
--|--|--
Notifications |  notifications_top | notifications_scrolled
Comments | comments_top | comments_scrolled 

Since the padding only exists at the top most and gone when scrolled, it's not possible to add the padding on the `ListTableHeaderView` – since doing this would also increase the padding when the view is stickied. Therefore, `tableHeaderView` property is utilized for this.

Also, cc-ing @mattmiklic to verify the designs, thank you!

To test:

- Go to Notifications, verify that the top padding is there.
- Go to Comments, verify that the top padding is there.
- Switch off the feature flag for unified list, and go to Notifications. Verify that the top padding does not affect the old design.
- Switch off the feature flag for unified list, and go to Comments. Verify that the top padding does not affect the old design.

## Regression Notes
1. Potential unintended areas of impact
Change may affect the old design.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Made sure that the padding is only active when the feature flag is on. Manually tested.

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17036, opened 2021-08-16
Notifications: Adjust filter bar container background color

dvdchr:
This fixes the background color from the filter tab bar that looks like it's "leaking" to the table view's first header view cell in dark mode. This is because the background of the filter tab bar's superview is set to `.filterBackground`, and in #16962 a 10pt constraint is added on the bottom of the filter bar container to improve the visual aspects.

To fix this, the filter tab bar's superview is set to `.systemBackground` while the unified list feature flag is enabled.

Before | After
--|--
filter_before | filter_after

To test:
- Set the app appearance settings to dark.
- Go to Notifications tab.
- Verify that the filter tab bar is displayed correctly.

## Regression Notes
None

Manually tested in iPhone and iPad.

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift
@@ -596,7 +596,7 @@ private extension NotificationsViewController {
         filterTabBarBottomConstraint.constant = usesUnifiedList ? Constants.filterTabBarBottomSpace : 0
 
         // With the 10pt bottom padding addition, ensure that the extra padding has the same background color
-        // as the table header cell.
+        // as the table header cell. NOTE: Move this line to `setupFilterBar` once the feature flag is removed.
         filterTabBar.superview?.backgroundColor = usesUnifiedList ? .systemBackground : .filterBarBackground
     }
 }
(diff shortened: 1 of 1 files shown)


ScoutHarris (approved):
Interesting. It seems setting `filterTabBar.superview?.backgroundColor` before wasn't even necessary. Well, now it is! :smile: Anywho, the only suggestion I have is maybe moving your code to `setupFilterBar`, where you can simply replace the existing line (`filterTabBar.superview?.backgroundColor = .filterBarBackground`) with your. Seems like a better long term fit.

Otherwise :shipit: !

dvdchr:
Thank you, @ScoutHarris !

> Interesting. It seems setting `filterTabBar.superview?.backgroundColor` before wasn't even necessary. Well, now it is! 😄

Well, actually... it would be even better if we didn't set the `backgroundColor` at all in `setupFilterBar()` 😅 . Removing that line fixes this issue, but I wanted to minimize the impact since this is still feature-flagged.

> moving your code to setupFilterBar, where you can simply replace the existing line

Since this is still feature flagged and the `NotificationsViewController` is the root VC of the tab, I need to react to changes on the feature flag value. Currently, feature flag value for `usesUnifiedList` is [re-fetched](https://github.com/wordpress-mobile/WordPress-iOS/blob/ed0421febd7af4a7604d90974d3dc21b33e15be8/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift#L184-L188) on `viewWillAppear`, and a property observer calls `updateFilterBarConstraints()` [when the flag value changed](https://github.com/wordpress-mobile/WordPress-iOS/blob/ed0421febd7af4a7604d90974d3dc21b33e15be8/WordPress/Classes/ViewRelated/Notifications/Controllers/NotificationsViewController.swift#L106-L117). This way, the background color of the filter bar are properly adjusted when the feature flag is toggled.

I've added a comment (https://github.com/wordpress-mobile/WordPress-iOS/pull/17036/commits/a8f29aae26b19f865240fba64da95e41ecbac0d1) as a future reminder to move this back to `setupFilterBar()` (or better, just delete them altogether 😅 ) once the feature flag is removed. Thanks!

PR #17054, opened 2021-08-19
Add time labels to Blogging Reminders

Gio2018:
Fixes #NA

This PR adds the time in two parts of Blogging Reminders (see screenshots):
- Site settings cell
- Blogging Reminders epilogue screen

To test:

- Build/run and go to site settings
- Set some reminders if you haven't done so already (tap Blogging Reminders then choose some days, choose a time or leave the default, then tap "Notify Me" and then "Done" in the following screen. You can also just dismiss the epilogue screen)
- Make sure the "Blogging Reminders" cell displays frequency and time in the detail label
- Tap on Blogging Reminders select/change days and/or time, then tap "Notify Me" (or "Update" if you have changed an existing schedule) and make sure the epilogue screen shows the days and the time you selected, in the description

Site Settings | Day Picker
-- | --
 | 

## Regression Notes
None, this PR only changes descriptions

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduleFormatter.swift
+++ b/WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduleFormatter.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 struct BloggingRemindersScheduleFormatter {
-    let calendar: Calendar
+    private let calendar: Calendar
 
     init(calendar: Calendar? = nil) {
         self.calendar = calendar ?? {
@@ -11,7 +11,7 @@ struct BloggingRemindersScheduleFormatter {
         }()
     }
 
-    /// Attributed description string of the current schedule for the specified blog.
+    /// Attributed short description string of the current schedule for the specified blog.
     ///
     func shortScheduleDescription(for schedule: BloggingRemindersScheduler.Schedule, time: String? = nil) -> NSAttributedString {
         switch schedule {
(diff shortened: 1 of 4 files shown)


osullivanchris (approved):
Looks good @Gio2018 ! 

Small details:
- as discussed, better not to include the time in the string on the day picker. Can be done as a follow up
- The string is getting tight on site settings. Not sure I have a fix in mind. If it gets truncated, it will be the time rather than the days (more important) that gets cut off, so it could be fine.

Gio2018:
Hey @osullivanchris 

> * as discussed, better not to include the time in the string on the day picker. Can be done as a follow up

already done, this is included in this PR

> * The string is getting tight on site settings. Not sure I have a fix in mind. If it gets truncated, it will be the time rather than the days (more important) that gets cut off, so it could be fine.

I noticed this. The text would scale down up to 50% of its size if needed. I did not go below that cause it seemed too small, but we can tweak the % as needed

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduleFormatter.swift
```diff
@@ -13,44 +13,68 @@ struct BloggingRemindersScheduleFormatter {
...
+            default:
+                return String(format: TextContent.manyRemindersShortDescriptionWithTime, days, time)
+            }
+        }()
+
+        return Self.stringToAttributedString(text)
+    }
+
+    static func shortScheduleDescription(for days: Int) -> NSAttributedString {
+
```
emilylaguna, line 98: What do you think about moving these into an extension?

emilylaguna (approved):
This works well, and looks good to me. I left a non-blocking comment about moving static methods into an extension.

WordPress/Classes/Utility/Blogging Reminders/BloggingRemindersScheduleFormatter.swift
Gio2018: Totally, will do now!

PR #17056, opened 2021-08-19
Site Settings: Improve scroll performance

dvdchr:
Refs `p5T066-2yQ-p2#comment-9613` (& the threads)

This fixes a frame drop issue while accessing and scrolling in the Site Settings page. Upon investigation, this is because the `tableSections` method always returns a new array object, populated with various conditionals and potential expensive operations. The `tableSections` method is called on these methods:

- `numberOfSectionsInTableView:`
- `tableView:cellForRowAtIndexPath:`
- `tableView:didSelectRowAtIndexPath:`
- `tableView:heightForRowAtIndexPath:`
- `tableView:numberOfRowsInSection:`
- `tableView:titleForHeaderInSection:`
- `tableView:viewForFooterInSection:`

The fix was to turn `tableSections` into an `NSArray<NSNumber *> *` instance variable. This makes the current `- (NSArray *)tableSections` method a setter override, and the method body is only invoked when the variable is nil.

Note that the page supports pull to refresh, which calls `refreshData`. To properly repopulate `tableSections`, the variable is reset to `nil` in order to properly repopulate the value.

--- 
cc: @mokagio if this is cutting too close to the submission, just let me know and I'll update this to target 18.1 instead. Thank you! 🙂 

### Scenario 1: The page should no longer be laggy.
- Select a site where you're an Administrator
- Open Site Settings.
- Verify that the page feels buttery smooth ✨ 

### Scenario 2: Data should be refreshed with new changes upon pull to refresh.
- Select a site where you're an Administrator.
- Open Site Settings.
- Open wordpress.com from the web, go to Settings, and change the Site Name and Site Tagline. Tap Save.
- Go back to the app, and perform a pull to refresh on the Site Settings page.
- Verify that the page is updated with the newest changes.

## Regression Notes
1. Potential unintended areas of impact
Table sections having stale values after pull to refresh.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Manually verified that the content is updated on pull to refresh.

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17061, opened 2021-08-20
Merge finalized 18.0 into develop

mokagio:
Notice more translations commits after the version bump. I run `finalize_release` in my afternoon to make sure the build was ready, but waited till my evening to upload to App Store Connect to give a chance to missing translations to come in. Those extra commits are the result of running the lane to download the translations again.

I thought about reverting https://github.com/wordpress-mobile/WordPress-iOS/commit/5f4342d5ae5aca737969f9d0d39f90e07bec8903 to reflect the fact that those translation changes were not part of the build submitted for review but I decided not to because the `18.0` tag will be created against https://github.com/wordpress-mobile/WordPress-iOS/commit/c32e3815bf510afc88f74382702c7a282e3fa303 anyway.


fastlane/metadata/de-DE/subtitle.txt
```diff
@@ -1 +1 @@
-Erstelle eine Website oder ein Blog
\ No newline at end of file
+Erstelle deine neue Website
```
mokagio, line 1: Before 18.1, I'll submit this and the other subtitle changes for review to GlotPress so I won't have to make them manually anymore every time.

mokagio (commented):
- [x] Fastlane  localized `release_notes.txt` updates
- [x] Localized `.strings` updates
- [x] Build number bump in version `.xcconfig` file
- [x] Plus diffs from the PR that landed on the release branch since the last beta

fastlane/jetpack_metadata/it/release_notes.txt
```diff
@@ -0,0 +1,9 @@
+Nella versione 18.0 abbiamo ottimizzati le immagini usate nei Temi, per garantirti download più rapidi e un impatto ridotto sul piano della tua rete dati.
+
+Ora puoi impostare o rimuovere l'immagine in evidenza per un post dalle impostazioni del blocco immagine. E in caso te ne dimenticassi, un nuovo banner in evidenza mostra la selezione corrente.
+
+Anche dal punto di vista del design, i commenti e le notifiche hanno un aspetto più coerente. Nell'ambito della scelta del dominio o del design del sito, le intestazioni sono state ottimizzate per migliorare l'accessibilità sui dispositivi con il testo di grandi dimensioni.
+
+Quando ti confermiamo la pubblicazione di un post tramite un avviso in-app, ora vedrai le e commerciali nel titolo visualizzate correttamente e non in formato HTML.
+
+Last but not least, abbiamo ottimizzato il ripristino da token di accesso invalidati.
```
AliSoftware, line 9: That "Last but not least" untranslated and still in English while the rest is properly in Italian 🤔  Is that intentional, like "the English expression is so commonly known in Italian-speaking countries that it's ok to use it untranslated"? Or more likely a part of the sentence the translator forgot to translate?

Anyway, too late now…

AliSoftware: I'm also wondering, since this last item is about authentication tokens, was it also relevant to Jetpack in the first place or was that an only-WP change?

I think it might be correct to have kept that for both JP+WP because it's about auth tokens not about the login UI (which indeed differs in JP), but still thought it was worth asking – albeit too late.

AliSoftware (approved):
✅  Version bumps OK
✅  Extra fix #17056 in the commit list matches changes in the diff
✅  Translations update ok, despite the note in the PR description (I agree that it's ok to keep the extra commit given the tag will be on the right one)
❔  Left some notes on release notes `it` translations, even if it's likely too late for it now (and not important enough to be a blocker at that stage)

fastlane/jetpack_metadata/it/release_notes.txt
mokagio: > I'm also wondering, since this last item is about authentication tokens, was it also relevant to Jetpack in the first place or was that an only-WP change?
>
> I think it might be correct to have kept that for both JP+WP because it's about auth tokens not about the login UI (which indeed differs in JP), but still thought it was worth asking – albeit too late.

@leandroalonso do you know the answer to this? It's too late know but knowing about this will be helpful for me in the future. Thanks!

mokagio: > That "Last but not least" untranslated and still in English while the rest is properly in Italian 🤔 Is that intentional, like "the English expression is so commonly known in Italian-speaking countries that it's ok to use it untranslated"? Or more likely a part of the sentence the translator forgot to translate?

It's an expression that was known to me when I was back in Italy. But that's not saying much, I had an interest for the English language that was above that of my friends, other than those studying English at university, I guess.

I think most Italians savvy enough to use WordPress on their iPhone would understand this.

leandroalonso: > @leandroalonso do you know the answer to this? It's too late know but knowing about this will be helpful for me in the future. Thanks!

Yes, in this case the token fix would affect JP+WP. :)

PR #17064, opened 2021-08-23
Merge 18.1.0.0 code freeze into develop

mokagio:


mokagio (commented):

Podfile.lock
```diff
@@ -53,7 +53,7 @@ PODS:
   - FSInteractiveMap (0.1.0)
   - Gifu (3.2.0)
   - glog (0.3.5)
-  - GoogleSignIn (6.0.1):
+  - GoogleSignIn (6.0.2):
```
mokagio, line 56: @startuptester notice that there's a Google Sign-In update, but the new version they shipped has changes only in how library authors can use interact with Google Sign-In. There are no changes in the library code itself.

mokagio: This update was due to how [WordPressAuthenticator-iOS declares its GoogleSignIn dependency](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/0437b6c5f2f938b2061e78578ca23717c7bc5804/WordPressAuthenticator.podspec#L41) using the optimistic operator:

```ruby
s.dependency 'GoogleSignIn', '~> 6.0.1'
```

It was not an explicit update. That is, I "had to" do it, I didn't choose to.

AliSoftware: @mokagio I'm not sure why WPAuthenticator uses the optimiztic operator incorrectly there – i.e. not the way SemVer intends it.

The right way should have been using `~> 6.0` with only 2 numbers, because `~>` means "only the last number is allowed to be higher" and in SemVer we want to allow the minor version to increase (but not the major one) hence `~> x.y`, while `~> x.y.z` means only the patch version will be allowed to grow and would prevent us from using GoogleSignIn 6.1 and forward, despite them being non-breaking (assuming GoogleSignIn respects semver)

TL;DR: shouldn't we fix WPAuth's podspec in the first place?

AliSoftware (approved):

Podfile.lock
mokagio: I think we should update authentication, yeah.

It was `~> 4.4` [originally](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/7cbd6269210ee176db2fff34dfd75893a47bb9a5), [then](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/01e1ccf4d574b12157c823826a77a39beb71d1a0) it became `~> 5.0.2`, and [recently](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/commit/ee628af0a330962578a7c3ebe98e51233cebfc8e) it went to `~> 6.0.1`.

I do wonder, though, if it would be best to actually lock it in place and only update intentionally. This is a key library that has given us trouble in the past, I don't like that `pod install` here (may) result in a different version than the one the library was tested against. Currently, the `Pofile.lock` in the library still [points to `6.0.1`](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/0437b6c5f2f938b2061e78578ca23717c7bc5804/Podfile.lock#L16).

AliSoftware: Side note that if we decide to lock the version in place – in fear for GoogleSignIn to not respect semver and break things during patch or minor updates – I think we should then either use `= x.y.z` or `>= x.y.z, < u.v.w` (and also add a `# comment`) rather than using something like `~> x.y.z`.

Because `~> x.y.z` feels like a typo to me (the author using the optimistic operator but not realizing this is not how semver is supposed to work) as opposed to be intentional, while `=` or `>=, <` are clearer in intent if the goal is to explicitly opt out of semver for a dependency.

mokagio: `:nods:` I feel `= x.y.z` would be the option that gives us more control and safety in such a case.

PR #17066, opened 2021-08-23
Local Weekly Roundup

diegoreymendez:
Implements the Weekly Roundup for iOS (with Local Notifications).

WPKit PR: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/438

## Things to do in upcoming PRs:

- Add configuration option to turn these notifications ON / OFF.

### Test how the notifications look like and how they react to taps:

1. Check the debug menu at Me > App Setting > Debug > Weekly Roundup
2. Tap on schedule immediately (the other options can be used too but they don't guarantee a time of delivery).  Keep in mind the static fallback / static notification is currently disabled so it will never show at this time.
3. Background the App, and see that the notifications come up.
4. Tap on any of them.

### Test that the trigger time is right:

1. Place a breakpoint at `BackgroundTasksCoordinator.swift:149`.
2. Run the App.
3. When the BP is hit, `po nextDate` and make sure it's next Monday at 10 am (device time).
4. Try this also using another calendar locale.

### Test tracking:

In your console, filter the messaged by "Tracked:".

1. Check the debug menu at Me > App Setting > Debug > Weekly Roundup
2. Tap "Schedule Immediately".
3. You should see up to a maximum of 5 messages that look like this:

🔵 Tracked: notification_scheduled <notification_type: weekly_roundup, site_id: 105284648>

4. Tap on one of the notifications, the App will open, and you should see a tracked notification like this:

🔵 Tracked: notification_tapped <notification_type: weekly_roundup, site_id: 105284648 >

Make sure you also see these events in the live analytics tool.

## Regression Notes

None.

PR submission checklist:


WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -162,7 +166,7 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
             app.endBackgroundTask(task)
             bgTask = .invalid
         }
-
+/*
```
diegoreymendez: This needs to be restored before pushing the PR.

WordPress/Classes/ViewRelated/Stats/SiteStatsDashboardViewController.swift
```diff
@@ -178,7 +187,14 @@ private extension SiteStatsDashboardViewController {
                                                        animated: false)
             }
 
-            periodTableViewController.selectedDate = periodDate ?? StatsDataHelper.currentDateForSite()
+            if let newDate = periodDate {
```
diegoreymendez, line 190: This should probably be reverted, unless I make it do something different.

diegoreymendez, line 195: Review these changes... I think I will probably roll this back.

This was added to try and not overwrite the selected date that is read from user defaults.

WordPress/Classes/Utility/BuildInformation/FeatureFlagOverrideStore.swift
```diff
@@ -41,10 +44,12 @@ struct FeatureFlagOverrideStore {
 
         if isOverridden(featureFlag) {
             store.removeObject(forKey: key)
+            featureFlag.overrideChanged()
         }
 
         if value != featureFlag.enabled {
             store.set(value, forKey: key)
+            featureFlag.overrideChanged()
```
diegoreymendez, line 52: Implement this or roll the change back.

Podfile
```diff
@@ -47,10 +47,10 @@ def wordpress_ui
 end
 
 def wordpress_kit
-    pod 'WordPressKit', '~> 4.39.0-beta'
+    # pod 'WordPressKit', '~> 4.39.0-beta'
```
diegoreymendez: Roll this back after releaseing a new pod version.

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -233,9 +237,10 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
         // 21-Oct-2017: We are only handling background URLSessions initiated by the share extension so there
         // is no need to inspect the identifier beyond the simple check here.
         if identifier.contains(WPAppGroupName) {
+            /*
             let manager = ShareExtensionSessionManager(appGroup: WPAppGroupName, backgroundSessionIdentifier: identifier)
             manager.backgroundSessionCompletionBlock = completionHandler
-            manager.startBackgroundSession()
+            manager.startBackgroundSession()*/
         }
     }
```
Gio2018: This method does not seem to do anything anymore, can we just remove it?

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -320,7 +325,20 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
         noticePresenter = NoticePresenter()
     }
 
+    private func setupBackgroundTasks() {
+
+    }
```
Gio2018: Can we remove this empty method?

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -320,7 +325,20 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
...
+
     private func setupBackgroundRefresh(_ application: UIApplication) {
+        backgroundTasksCoordinator.scheduleTasks { result in
+            switch result {
+            case .success:
+                break
+            case .failure(let error):
+                DDLogError("Error scheduling background tasks: \(error)")
+            }
+        }
```
Gio2018: nit: since we are only interested in one case, we can also use an `if case ... (let ...)` construct

WordPress/Classes/Utility/BackgroundTasks/BackgroundTasksCoordinator.swift
```diff
@@ -0,0 +1,160 @@
...
+        for task in registeredTasks {
+            schedule(task) { result in
+                switch result {
+                case .success:
+                    break
+                case .failure(let error):
+                    tasksAndErrors[type(of: task).identifier] = error
+                }
+            }
+        }
```
Gio2018: nit: same as above, we could use an if case let if we wanted here

WordPress/Classes/ViewRelated/Developer/WeeklyRoundupDebugScreen.swift
```diff
@@ -0,0 +1,132 @@
...
+
+                    let runDateComponents = calendar.dateComponents([.hour, .minute, .second], from: taskRunDate)
+                    let staticNotificationDateComponents = calendar.dateComponents([.hour, .minute, .second], from: staticNotificationDate)
+
+                    let backgroundTask = WeeklyRoundupBackgroundTask(
+                        runDateComponents: runDateComponents,
+                        staticNotificationDateComponents: staticNotificationDateComponents)
+
+                    WordPressAppDelegate.shared?.backgroundTasksCoordinator.schedule(backgroundTask) { result in
+                        print("Done!")
```
Gio2018: I know it's just a debug screen, but can we use DDLogInfo instead of a standalone print? Or just remove the print?

Gio2018 (commented):
Hey @diegoreymendez this worked good: Notification looked correct, tapping takes to the stats of the right site, `nextDate` looks correct and the tracks show correctly in both the console and the live view. I just left a few code comments that might be worth considering.

Gio2018:
Oh also it looks like some new files are missing the Jetpack target membership, can you please fix that?

WordPress/Classes/ViewRelated/Developer/WeeklyRoundupDebugScreen.swift
diegoreymendez: Yes of course.  I'll check this case to see what's more appropriate.  In this case it's just something I used for debugging and forgot to remove.

WordPress/Classes/Utility/BackgroundTasks/BackgroundTasksCoordinator.swift
diegoreymendez: Changed.

WordPress/Classes/System/WordPressAppDelegate.swift
diegoreymendez: Changed.

WordPress/Classes/System/WordPressAppDelegate.swift
diegoreymendez: Done.

WordPress/Classes/System/WordPressAppDelegate.swift
diegoreymendez: I actually had disabled this because I was having some trouble while testing, but it was unrelated.  I've re-enabled this.

diegoreymendez:
@Gio2018 - Ready for another look.

Gio2018 (approved):
:shipit:

PR #17085, opened 2021-08-25
Add the blog_id to editor session event properties

dcalhoun:
Add the `blog_id` property to editor session analytic events. Field Guide documentation requests that all Tracks events that occur in the context of a site include the `blog_id` property. Relates to 1641-gh-Automattic/tracks-events-validation-status. 

To test: Perform the following steps while (1) logged in to a WP.com account and (2) logged in with WP Admin credentials only.

1. Open the block editor.
1. Tap the three-dot menu in the top right.
1. Tap Switch to HTML Mode.
1. Tap the three-dot menu in the top right.
1. Tap Switch to Visual Mode.
1. Close the editor.
1. Verify the following Tracks events include the correct `blog_id`
   property.<sup>1</sup>
    * `wpios_editor_session_start`
    * `wpios_editor_session_switch_editor`
    * `wpios_editor_session_end`

[1]: The `blog_id` while logged in with WP Admin credentials only will be `0`.

## Regression Notes
  Unintended event property changes could occur.
  Manually reviewed other properties on the event.
  Added unit tests to verify inclusion of `blog_id` for editor session events. 

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
@@ -50,7 +50,7 @@ struct PostEditorAnalyticsSession {
 
     mutating func `switch`(editor: Editor) {
         currentEditor = editor
-        WPAppAnalytics.track(.editorSessionSwitchEditor, withProperties: commonProperties)
+        WPAppAnalytics.track(.editorSessionSwitchEditor, withProperties: commonProperties, withBlogID: blogID)
     }
 
     mutating func forceOutcome(_ newOutcome: Outcome) {
--- a/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
+++ b/WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
@@ -3,6 +3,7 @@ import Foundation
 struct PostEditorAnalyticsSession {
     private let sessionId = UUID().uuidString
     let postType: String
+    let blogID: NSNumber?
     let blogType: String
     let contentType: String
     var started = false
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
```diff
@@ -76,6 +78,7 @@ struct PostEditorAnalyticsSession {
 
 private extension PostEditorAnalyticsSession {
     enum Property {
+        static let blogID = "blog_id"
```
guarani, line 81: I expected to see `blog_id` in Tracks but instead see `blogid` (no underscore). I'm not sure why this is, or if it's expected. The most likely explanation I'm aware of is that it's renamed at some level.
I don't think it's an issue because other events I looked at e.g. `wpios_editor_post_created` also don't use an underscore.

guarani (approved):
I tested and everything is working as expected. I only left one comment but it doesn't feel like a blocker to me.
The tests you added look great too.

- [x] Simple WP.com site
- [x] Atomic WP.com site
- [x] Self-hosted site accessed via its credentials

💯 :shipit:

WordPress/Classes/ViewRelated/Post/PostEditorAnalyticsSession.swift
dcalhoun: @guarani yes, my understanding is that somewhere along the way `blog_id` is renamed to `blogid`. However, I do not know where that rename occurs. I used `blog_id` here because (1) it appears to "work" with a result of `blogid` matching all other Tracks events and (2) other references in this source use this naming convention. 

https://github.com/wordpress-mobile/WordPress-iOS/blob/d68d45bdac8a9792896ddbbf1e3011c74a776443/WordPress/Classes/Utility/Analytics/WPAppAnalytics.m#L15

Given that, I think it makes sense to leave this as-is.

guarani: Sounds good! Marking this resolved since that seems enough assurance that `blogid` is correct.

Issue #17087, opened 2021-08-26
Comments: Create new comment details view


dvdchr:
Ref: pbArwn-2ks-p2

## Summary
Create a new Comment Details view, which will be accessible from the Comments and Notifications page. Note that some components will be addressed separately through different issues, and some functionalities may depend on ongoing development. Here's the design from Zeplin:

![Screen Shot 2021-08-26 at 16 08 28](https://user-images.githubusercontent.com/1299411/130935920-d5e1d4c6-22eb-4085-a46c-b1a3f8b70fc8.png)

Some notes regarding components marked red (going from the top):
- The edit button will be visually included in this issue, but the functionality depends on #17000, and _may_ be wired later on. When tapped, the new edit comment view should be shown in full modal presentation.
  - Since #17046 is merged, the functionality can be included here, but I'll need to make sure that the edit view feature flag is also enabled.
- Same as above, the reply button will be visually included in this issue, but the functionality will be addressed through a separate issue.
- Lastly, the moderation bar will completely be excluded from this and will be addressed through a different issue.

### Site Comments

- [x] Add a feature flag for comment detail. #17088
- [x] Create the initial view using Swift. #17088
- [x] Comment header cell. #17095
- [x] Create a generic cell for the other similar-looking fields (e.g. web address, email address, IP address). #17105 
- [x] Comment reply indicator cell. #17106 
- [x] #17108
- [x] Add "Delete Permanently" button when the Comment is in Trashed state. #17286
- [x] Apply a visual blur effect when the page is scrolled. #17255
- [x] Adjust certain cells' visibility based on account permission. #17147
- [x] https://github.com/wordpress-mobile/WordPressKit-iOS/pull/463
- [x] #17399
- [x] #17408
- [x] Fetch the parent Comment object if `parentID` exists, to get the replied author's name (i.e. "Reply to `Pamela Nguyen`").  #17246
- [x] Show the new details view from My Site > Comments. #17088

#### Finishing touches
- [x] Review accessibility.
- [x] Review Tracks.
- [x] #17466

#### Issues
- [x] Don't show the web address cell if the content is empty (discussion [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/17105#issuecomment-912662400)). Addressed in #17147.
[...]

PR #17095, opened 2021-08-30
Comment Detail: Add header cell

dvdchr:
Refs #17087

This adds the header cell component for the new comment detail. Some notes:

- The header cell is pretty simple, since it's a `UITableViewCell` with `subtitle` style (with properly applied text styles for both text label and detail text label). However, since this cell will be reused again later in Comment Threads, I'll create a new class for now to encapsulate all the styling and the `.subtitle` cell style.
- There are two variants for the header cell content. For this PR, I'll only show the comment's post regardless if the comment is a reply or not.
  - If the current comment is a reply, I'll have to fetch the parent Comment object in order to get the both the replied author's name, and the content. This will be done in a separate PR. I've added a new entry in #17087.

Preview:

Simulator Screen Shot - iPhone 8 - 2021-08-30 at 21 41 57

To test:

- Ensure that the `New Comment Detail` feature flag is enabled.
- Go to My Site > Comments, and tap on any comment. Preferably not the reply ones.
- Verify that the header cell is displayed.
- Tap the header cell, and verify that the post is opened correctly.

## Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
@@ -8,6 +8,10 @@ class CommentDetailViewController: UITableViewController {
 
     private var rows = [RowType]()
 
+    // MARK: Views
+
+    private var headerCell = CommentHeaderTableViewCell()
+
     // MARK: Initialization
 
     @objc required init(comment: Comment) {
@@ -25,6 +29,7 @@ class CommentDetailViewController: UITableViewController {
         super.viewDidLoad()
         configureNavigationBar()
         configureTable()
+        configureRows()
     }
 
     // MARK: Table view data source
@@ -35,6 +40,17 @@ class CommentDetailViewController: UITableViewController {
 
     override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
         return rows.count
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        switch rows[indexPath.row] {
+        case .header:
+            configureHeaderCell()
+            return headerCell
+
+        default:
+            return .init()
+        }
     }
 
 }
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Comments/CommentHeaderTableViewCell.swift
```diff
@@ -0,0 +1,30 @@
+import UIKit
+
+class CommentHeaderTableViewCell: UITableViewCell, Reusable {
+
+    // MARK: Initialization
+
+    required init(reuseIdentifier: String = CommentHeaderTableViewCell.defaultReuseID) {
```
ScoutHarris: I don't see that `reuseIdentifier` needs to be a parameter. `Reusable` has its own `defaultReuseID`, as you're defaulting to. I don't see any reason to allow it to be overridden. Maybe it should just be passed directly to `super.init`?


```
required init() {
    super.init(style: .subtitle, reuseIdentifier: CommentHeaderTableViewCell.defaultReuseID)
    configureStyle()
}
```

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -58,6 +86,46 @@ private extension CommentDetailViewController {
...
+    func navigateToPost() {
+        guard let blog = comment.blog,
+              let siteID = blog.dotComID,
+              blog.supports(.wpComRESTAPI) else {
+            viewPostInWebView()
+            return
+        }
+
+        let readerViewController = ReaderDetailViewController.controllerWithPostID(NSNumber(value: comment.postID), siteID: siteID, isFeed: false)
+        navigationController?.pushFullscreenViewController(readerViewController, animated: true)
```
ScoutHarris, line 113: As I was testing on iPad, I noticed an odd behavior which I think is related to `pushFullscreenViewController`. The `readerViewController` tries to appear full screen, then jumps down to split view. I see this is simply copied behavior from `CommentViewController`, and I did confirm this is an existing issue in the app store version. But maybe we can fix it with this feature? (at some point, not in this PR)

ScoutHarris (approved):
Hey @dvdchr . Just a couple comments, but otherwise 👍 !

:shipit:

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
dvdchr: Yeah, that makes sense. I'll note this down in #17087 and address this in a separate PR. Thanks! 🙂

WordPress/Classes/ViewRelated/Comments/CommentHeaderTableViewCell.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17095/commits/e2c33df7d059541b8aa688fe8e4b98839aa114d6 !

PR #17076, opened 2021-08-24
Fix/16941 - Close account change error flow

nikola-milicevic:
This PR changes alert shown in case user cannot close account due to having atomic site to a more generic one, with a **Contact Support** button.
This change is needed because having an atomic site is something that's considered being an implementation detail user might not necessarily know about.

## To test:
1. Make sure the current user has an atomic site
2. Under **My Site** tab tap on profile button at the top right corner
3. Tap on **Account Settings**
4. Tap on **Close account**
5. Close account alert is displayed
6. Tap on **Contact Support**
7. New alert is displayed for entering contact info
8. Upon entering contact info user can leave a message to support center

## Flow
Atomic site error | Enter contact info dialog | Contact us
--- | --- | ---
Simulator Screen Shot - iPhone 11 - 2021-08-24 at 15 44 03 | IMG_2BE159ACAD98-1 | Simulator Screen Shot - iPhone 11 - 2021-08-24 at 01 42 12

## Regression Notes
1. Potential unintended areas of impact
n/a

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17101, opened 2021-08-31
Fix/16941 - Close account additional logic

nikola-milicevic:
Complements [#17076](https://github.com/wordpress-mobile/WordPress-iOS/pull/17076)

This PR adds additional logic to the close account error flow.
It mimics the [logic found on Calypso front-end](https://github.com/Automattic/wp-calypso/blob/b550e845f46f9c93641f54193d73626cb5b669e9/client/me/account-close/main.jsx#L90), so now user cannot close the account even if it's added as an admin to the the atomic site.

## To test:
1. Make sure the current user is added to someone else's atomic site as an admin
2. Under **My Site** tab tap on profile button at the top right corner
3. Tap on **Account Settings**
4. Tap on **Close account**
5. Close account error alert is displayed

## Screenshot:
Screen Shot 2021-08-31 at 15 33 21

## Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
@@ -125,7 +125,7 @@ private class AccountSettingsController: SettingsController {
 
         let closeAccount = DestructiveButtonRow(
             title: NSLocalizedString("Close Account", comment: "Close account action label"),
-            action: presenter.present(closeAccountAction),
+            action: closeAccountAction,
             accessibilityIdentifier: "closeAccountButtonRow")
 
         return ImmuTable(sections: [
@@ -239,9 +239,8 @@ private class AccountSettingsController: SettingsController {
         }
     }
 
-    private var closeAccountAction: (ImmuTableRow) -> UIAlertController {
+    private var closeAccountAction: (ImmuTableRow) -> Void {
         return { [weak self] _ in
-            return self?.closeAccountAlert ?? UIAlertController()
         }
     }
 
--- a/WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
@@ -244,6 +244,19 @@ private class AccountSettingsController: SettingsController {
         }
     }
 
+    private var hasAtomicSite: Bool {
+        let context = ContextManager.sharedInstance().mainContext
+        let service = AccountService(managedObjectContext: context)
+        guard let account = service.defaultWordPressComAccount() else {
+            return false
+        }
+
+        for blog in account.blogs where blog.isAtomic() {
+            return true
+        }
+        return false
+    }
+
     private var closeAccountAlert: UIAlertController? {
         guard let value = settings?.username else {
             return nil
--- a/WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
@@ -33,6 +33,8 @@ private class AccountSettingsController: SettingsController {
     // MARK: - Initialization
 
     private let accountSettingsService: AccountSettingsService
+    private let accountService: AccountService
+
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
```diff
@@ -239,15 +239,27 @@ private class AccountSettingsController: SettingsController {
...
+                self.showCloseAccountAlert()
+            }
         }
     }
 
-    private var closeAccountAlert: UIAlertController? {
+    private var hasAtomicSite: Bool {
+        let context = ContextManager.sharedInstance().mainContext
+        let service = AccountService(managedObjectContext: context)
+        return service.defaultWordPressComAccount()?.hasAtomicSite() ?? false
```
momo-ozawa: What do you think of using dependency injection for the `AccountService`?

momo-ozawa (commented):
Works as described! 🎉

nikola-milicevic:
I updated the code, thanks for the suggestion!

WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
```diff
@@ -45,24 +47,26 @@ private class AccountSettingsController: SettingsController {
     }
     private let alertHelper = DestructiveAlertHelper()
 
-    init(service: AccountSettingsService) {
-        self.service = service
+    init(accountSettingsService: AccountSettingsService,
+         accountService: AccountService = AccountService(managedObjectContext: ContextManager.sharedInstance().mainContext)) {
+        self.accountSettingsService = accountSettingsService
+        self.accountService = accountService
```
momo-ozawa, line 53: [nit] What do you think of making `AccountService` optional instead, so we can do something like this? We do this in a couple different places, so I thought it might be nice for the sake of consistency 😄 
```suggestion
    init(accountSettingsService: AccountSettingsService,
         accountService: AccountService? = nil) {
        self.accountSettingsService = accountSettingsService
        self.accountService = accountService ?? AccountService(managedObjectContext: ContextManager.sharedInstance().mainContext)
```

momo-ozawa (approved):
LGTM with 1 small comment.

WordPress/Classes/ViewRelated/Me/Account Settings/AccountSettingsViewController.swift
nikola-milicevic: Hmm well I think this is misleading, since this would imply that the `AccountSettingsController` doesn't necessarily rely on `AccountService` (but it does), since it can be optional. And another thing, at the point of use one would expect that passing a `nil` as an argument would result in that property to be `nil`, which would not be the case. Let me know what you think.

momo-ozawa: I think that having a non-optional `AccountService` as a property for `AccountSettingsController` implies that it's a dependency. 

@leandroalonso Do you have any insight / comments about the `xyzService? = nil` [dependency injection pattern ](https://github.com/wordpress-mobile/WordPress-iOS/search?p=1&q=%22Service%3F+%3D+nil%22+path%3A%2FWordPress%2FClasses) we use throughout the repo?

nikola-milicevic: > I think that having a non-optional AccountService as a property for AccountSettingsController implies that it's a dependency.

I agree, but I don't feel having an optional parameter for a non optional property is the right thing.

leandroalonso: While it would be good to follow the pattern, I agree with @nikola-milicevic to make the dependency explicit (ps.: I've done the `nil` a few times myself but looking back it doesn't seem the best decision. Shame! 😝)

I think we can keep it that way (non-nil) and just remember to apply this same pattern to future code.

nikola-milicevic: @momo-ozawa @leandroalonso thanks for the comments! I'm merging it now as is.

Issue #17000, opened 2021-08-09
Comments: add new Edit view


ScoutHarris:
Create a new view to edit a Comment.

- [x] Add feature flag. (#17046)
- [x] Add new view using inset grouped table view style. (#17046)
- [x] Show from:
  - [x] My Site > Comments > Comment details. (Will eventually be shown from the `Edit` nav bar button on the new Comment details view. For now, show from existing `Edit` button.) (#17046)
  - [x] New Comment details navigation bar. (#17088)
- [x] Display Comment data:
  - [x] #17059
  - [x] #17077
- [x] Save changes.
  - [x] #17096
  - [x] #17099
- [x] #17093
- [x] Validation:
  - [x] #17092
  - [x] #17102
  - [x] Web Address - in Account Settings, there is no validation on `Web Address`. I'm assuming because the endpoint seems to accept anything. It does slap a `http://` on something that doesn't have a scheme though, so there's that. 🤷  To avoid any conflicts, I'll follow suit here and not validate the web address.
- [ ] Fix scroll position issues with content field. (https://github.com/wordpress-mobile/WordPress-iOS/pull/17077#pullrequestreview-736981753) 
  - [x] #17110
- [x] Accessibility.
- [x] Tracks.
- [x] Enable feature flag and add release note. (#17111)

From Zeplin:

![Edit Screen](https://user-images.githubusercontent.com/1816888/128779302-e1d47d93-b59a-4008-81af-cb1893d2941d.png)

PR #17102, opened 2021-08-31
Edit Comment: add red border to email field if email is invalid.

ScoutHarris:
Ref: #17000

This shows/hides a red border on the email field depending on the entered email's validity.

To test:
- Enable the `newCommentEdit` feature.
- Go to My Site > Comments > Comment details > Edit.
- Change the email address.
- Verify the red border appears when the email is invalid, and does not when the email in valid. Note that email is not required, so if it is cleared it is valid.

https://user-images.githubusercontent.com/1816888/131565437-8337cb83-cdb1-4566-b4f5-e1628a5d0887.mp4

## Regression Notes
N/A.

N/A.

N/A.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/EditCommentSingleLineCell.swift
+++ b/WordPress/Classes/ViewRelated/Comments/EditCommentSingleLineCell.swift
@@ -41,7 +41,7 @@ class EditCommentSingleLineCell: UITableViewCell, NibReusable {
             return
         }
 
-        contentView.layer.borderColor = UIColor.red.cgColor
+        contentView.layer.borderColor = UIColor.error.cgColor
         contentView.layer.borderWidth = 1.0
         contentView.layer.cornerRadius = 10
     }
--- a/WordPress/Classes/ViewRelated/Comments/EditCommentSingleLineCell.swift
+++ b/WordPress/Classes/ViewRelated/Comments/EditCommentSingleLineCell.swift
@@ -2,7 +2,7 @@ import Foundation
 
 
 protocol EditCommentSingleLineCellDelegate: AnyObject {
-    func fieldUpdated(_ type: TextFieldStyle, updatedText: String?, isValid: Bool)
+    func textUpdatedForCell(_ cell: EditCommentSingleLineCell)
 }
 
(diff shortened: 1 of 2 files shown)


ScoutHarris:
Hey @mattmiklic .

I wanted to run this by you before making it final to make sure it's acceptable.

As we discussed on the [design P2](pbArwn-2ks#comment-3756) , we don't really have existing examples of validation + error messaging on the fly + multiple fields. I looked at how Auth does it as it is the closest example. However, Auth doesn't show errors on the fly. It does validate the email as typing occurs, but the only indicator is the state of the `Continue` button.

https://user-images.githubusercontent.com/1816888/131567206-c005e85e-b896-4456-a4ed-874a86836d9c.mp4

Any errors that are shown are inserted after the button is tapped.

https://user-images.githubusercontent.com/1816888/131566967-1b96f122-bb93-4eb6-8516-98d0b0e25047.mp4

In the edit Comment view, we validate email the same as Auth. That is, `Done` is not enabled unless the email is valid. However, `Done` is also based on changes in other fields so it may not be obvious when email is invalid.

Since the email is validated as the user types, I was trying to avoid any weird UI glitches, jumps, etc. when showing an invalid state. I landed on simply putting a red border around the email field when it is invalid, and remove it when it is valid. What do you think?

https://user-images.githubusercontent.com/1816888/131567109-077c83c6-f0c5-4847-b104-260b45916ca2.mp4

mattmiklic (approved):
I like the red border. It would definitely be weird if an explicit error message showed up while the user's typing, so it's good that this is more subtle than that.

I'm guessing since the Done button is disabled if any of the fields aren't valid, we don't need to worry about showing error messages, right? If that's the case I'm cool with this solution.

ScoutHarris:
> I'm guessing since the Done button is disabled if any of the fields aren't valid, we don't need to worry about showing error messages, right? 

Yep, there is no need to show any error messages. Thank you!

dvdchr (approved):
LGTM! :shipit:

ScoutHarris:
Thanks @dvdchr !

PR #17117, opened 2021-09-03
Fix Reader Discover inconsistent state

leandroalonso:
This PR aims to fix two different issues:

1. Unsaving items from the "Saved" tab
2. Correctly display a post state on "Reader Discover"

### To test

1. Go to "Reader Discover"
2. Save a post
3. Make sure the bookmark icon is now blue
4. Go to "Saved" tab
5. Unsave the post
6. Go back to "Reader Discover"
7. Make sure the icon is not blue
8. Go back to "Saved" and the post shouldn't be there

## Regression Notes
Reader tabs

Tested all the reader tabs, closing and reopening the app

None.

PR submission checklist:


WordPress/Classes/Models/ReaderPost.m
```diff
@@ -505,4 +506,17 @@ - (NSDictionary *)railcarDictionary
...
+- (void) didSave {
+    [super didSave];
+
+    // When saving a post, changes are not reflected in Reader Discover because of the usage of
+    // ReaderCard as the Core Data object. By "faking" this change we make sure Reader Discover
+    // reflects a change on a post.
+    // We can confidently set `topics` to NULL given a card represents a post OR a list of topics.
+    if (self.card.count > 0) {
+        self.card.allObjects[0].topics = NULL;
+        [[ContextManager sharedInstance] saveContext:self.managedObjectContext];
```
leandroalonso, line 516: I personally don't like to have a reference to Core Data inside this model. However, it would be that or a reference to the service. Either feels wrong to me.

Anyway, it's that or creating a lot of code to make `NSFetchedResultsController` to observe changes on a child. 😬

WordPress/Classes/Models/ReaderPost.h
```diff
@@ -50,6 +51,7 @@ extern NSString * const ReaderPostStoredCommentTextKey;
 @property (nonatomic, readonly, strong) NSURL *featuredImageURL;
 @property (nonatomic, strong) NSString *tags;
 @property (nonatomic, strong) ReaderAbstractTopic *topic;
+@property (nonatomic, strong) NSSet<ReaderCard *> *card;
```
aerych, line 54: Are cards really one to many?  From my read of card related code and the changes here it seems like the answer is no and the one to many relationship is something we should sort out.  
To be clear, it would be a change out of scope for this PR.  I'm mostly just puzzled and wondering if I'm missing something that lets a post have multiple cards.

WordPress/Classes/Models/ReaderPost.m
```diff
@@ -505,4 +506,17 @@ - (NSDictionary *)railcarDictionary
     return nil;
 }
 
+- (void) didSave {
+    [super didSave];
+
+    // When saving a post, changes are not reflected in Reader Discover because of the usage of
+    // ReaderCard as the Core Data object. By "faking" this change we make sure Reader Discover
+    // reflects a change on a post.
+    // We can confidently set `topics` to NULL given a card represents a post OR a list of topics.
```
aerych: If I understand correctly: because the card is only showing posts, its okay to make a change to topics since topics should be empty anyway.  Does that sound right?  This jives with my understanding of how ReaderCards are created and how they work but want to confirm.

aerych (approved):
Howdy @leandroalonso 👋 
Tested as described and I can confirm posts and tabs update as expected.  We chatted some about the approach and while I share your misgivings about referencing the core data stack from a model I still think it's a reasonable change given the challenge relationships can pose. 
I had a couple of questions to double check my understanding but otherwise I think this is good to :shipit:  :) 
Cheers sir!

WordPress/Classes/Models/ReaderPost.h
leandroalonso: Very good question indeed and one that I asked myself. They're not. I created this relationship and I think this was an oversight on my side. (SHAME!)

I'll address this in another PR. :)

WordPress/Classes/Models/ReaderPost.m
leandroalonso: Yes!

A card can be only a post OR a topic. Do you have any tips on how to make the comment clear?

aerych: It wasn't really unclear, I just was being extra sure I understood. :) I haven't looked at the card code much.  But since you ask, if I were going to nitpick, maybe something like:  
```
A ReaderCard can have either a post, or a list of topics, but not both. Since this card has a post, we can confidently set `topics` to NULL.
```
But I think what's there is probably fine as is. :)

PR #17120, opened 2021-09-06
Add option to publish a Page from the share extension

### LINKED ISSUE #16084: Add Option to Publish Page from Share Extension
### Expected behavior
When sharing content to the app through the app's share extension, there should be an option to publish a page instead of a post.

### Actual behavior
When sharing content to the app through the app's share extension, the content is always published as a post.

1. Open a text editing app such as Edit, iA Writer, or Bear.
2. Open a note or text document.
3. Open the share sheet.
4. Select the WordPress share extension.
5. Tap "Next"
6. Tap "Publish

ghost:
Fixes #16084

Adds the option to create a Page from the Share Extension in addition to a Post.

| Before  | After |
| -| -|
| Simulator Screen Shot - iPhone 12 - 2021-09-06 at 13 16 29  | Simulator Screen Shot - iPhone 12 - 2021-09-06 at 11 33 09  |

Demo video:

https://user-images.githubusercontent.com/88823662/132219987-d595c6d1-13ad-4807-885f-0e87326c27e7.mov

To test:

- Open an application that can be shared from (a Text Editor, Safari, or Photos)
- Open the Share menu
- Select the Wordpress app
- Edit your Page content
- Tap "Next"
- Change the type to "Page"
- Tap "Publish"

## Regression Notes
- Both the Share extension and the Draft extension share code - possible I may have overlooked something in Drafts?

- Manually tested the Draft Action extension.

- No current tests for Share Extension.

PR submission checklist:


WordPress/WordPressShareExtension/ShareModularViewController.swift
```diff
@@ -389,8 +400,15 @@ extension ShareModularViewController: UITableViewDelegate {
 
 fileprivate extension ShareModularViewController {
     func configureModulesCell(_ cell: UITableViewCell, indexPath: IndexPath) {
-        switch indexPath.section {
-        case ModulesSection.categories.rawValue:
+        let moduleSection = ModulesSection.init(rawValue: indexPath.section) ?? .summary
```
diegoreymendez: The `.init` can be omitted here.

WordPress/WordPressShareExtension/ShareModularViewController.swift
```diff
@@ -445,17 +463,26 @@ fileprivate extension ShareModularViewController {
...
     func isModulesSectionEmpty(_ sectionIndex: Int) -> Bool {
         switch ModulesSection(rawValue: sectionIndex)! {
-        case .categories:
+        case .type:
             return false
+        case .categories:
+            return shareData.postType == .page
         case .tags:
-            return false
+            return shareData.postType == .page
```
diegoreymendez, line 476: I went into the web page editor (Gutenberg) and noticed that while categories are absent, tags can be used.

Would it be possible to enable tags for pages in the share extension as well?

WordPress/Tracks+ShareExtension.swift
```diff
@@ -43,6 +43,15 @@ extension Tracks {
         trackExtensionEvent(.categoriesSelected, properties: properties as [String: AnyObject]?)
     }
 
+    public func trackExtensionPostTypeOpened() {
+        trackExtensionEvent(.postTypeOpened)
+    }
+
+    public func trackExtensionPostTypeSelected(_ postType: String) {
+        let properties = ["post_type_tags": postType]
```
diegoreymendez: Just curious about the "_tags" part of the property name.  IMHO "post_type" could be clearer.

Thoughts?

diegoreymendez (changes requested):
Hey @autoblork,

It seems like when I switch to the "Page" type, and back to "Post", the "Tags" row is no longer tappable.  Tapping on it executes no animation and performs no action.

I've also added some other minor comments for your consideration.

WordPress/WordPressShareExtension/ShareModularViewController.swift
ghost: Yes, I should think so, but I don't see way to add tags when creating a page in the app. When creating a post in-app, you get a prompt to add categories and tags, but not for posts - unless I'm overlooking something? I thought it best to keep consistency with the app experience.

WordPress/Tracks+ShareExtension.swift
ghost: Ah, I meant to bring this up to you. Are these defined anywhere other than in the app (the strings to identify tracking events)? I have just added ones following an existing pattern but I'm unsure if thats enough.

The "_tags" part matches the other track events:

```
 let properties = ["categories_tags": categories]
```
I presume it's to identify the dict as a set of tags pertaining to each element? Or perhaps it's an error on that "categories" one, too, since tags are just:

```
let properties = ["selected_tags": tags]
```

ghost:
> It seems like when I switch to the "Page" type, and back to "Post", the "Tags" row is no longer tappable. Tapping on it executes no animation and performs no action.

Ah, I had overlooked that some of the cells have `isUserInteractionEnabled = false` set, and so now the number of cells can be dynamic, that property always needs to be set in case of cell reuse. The tag cell must be reusing the summary cell, and so it becomes un-interactive. Fixed, good catch!

WordPress/WordPressShareExtension/ShareModularViewController.swift
diegoreymendez: @osullivanchris - Can I ask your thoughts on this?

To summarize:

Should the share extension allow users to add tags to pages, even though the App itself doesn't let you edit the tags for a post?  The web page does let you add tags to a page.

My personal opinion is we can enable it in the share extension, and eventually add it in the Page settings in the App later on.

WordPress/Tracks+ShareExtension.swift
diegoreymendez: > Are these defined anywhere other than in the app (the strings to identify tracking events)?

Not really.  They're captured by our tracking tools, but they're defined by the App.

In the case of categories, I guess it's possible to think of categories as category-tags.  But this is just a theory as I've no recollection of the naming of that property.

IMHO we can safely name this property just:

"post_type"

PS: good points and reflection though

WordPress/WordPressShareExtension/ShareModularViewController.swift
ghost: Sending up Tags in a Page creation request doesn't _just work_, so this might be harder than I imagined. Not to doubt you, @diegoreymendez (😅) but are you certain you can add Tags to a Page? I don't see the option when creating a page, myself (using the web editor on a test site I made), and the [glossary seems to suggest they don't](https://wordpress.org/support/article/post-types/#pages).

diegoreymendez: > are you certain you can add Tags to a Page?

Not at all!  I just know it's possible to add them in the web editor, but I initially posed this as a question in case you found anything that made this undesirable, complicated, or straight out impossible.

Considering your findings I'd say it's ok to just resolve this and go forward without tags.

If necessary, it can be evaluated and implemented separately in the future.

WordPress/WordPressDraftActionExtension/Tracks+DraftAction.swift
```diff
@@ -42,6 +42,15 @@ extension Tracks {
         trackExtensionEvent(.categoriesSelected, properties: properties as [String: AnyObject]?)
     }
 
+    public func trackExtensionPostTypeOpened() {
+        trackExtensionEvent(.postTypeOpened)
+    }
+
+    public func trackExtensionPostTypeSelected(_ postType: String) {
+        let properties = ["post_type_tags": postType]
+        trackExtensionEvent(.postTypeSelected, properties: properties as [String: AnyObject]?)
```
diegoreymendez: Same as above.

WordPress/Tracks+ShareExtension.swift
```diff
@@ -43,6 +43,15 @@ extension Tracks {
         trackExtensionEvent(.categoriesSelected, properties: properties as [String: AnyObject]?)
     }
 
+    public func trackExtensionPostTypeOpened() {
+        trackExtensionEvent(.postTypeOpened)
+    }
+
+    public func trackExtensionPostTypeSelected(_ postType: String) {
+        let properties = ["post_type": postType]
+        trackExtensionEvent(.postTypeSelected, properties: properties as [String: AnyObject]?)
```
diegoreymendez: I think the properties dictionary doesn't need to be cast as an optional.  What's the thinking behind that choice?

WordPress/WordPressShareExtension/AppExtensionsService.swift
```diff
@@ -318,12 +327,18 @@ extension AppExtensionsService {
...
+                remotePost.tags = tags.arrayOfTags()
+            }
 
-        if let remoteCategories = RemotePostCategory.remotePostCategoriesFromString(categories) {
-            remotePost.categories = remoteCategories
+            if let remoteCategories = RemotePostCategory.remotePostCategoriesFromString(categories) {
+                remotePost.categories = remoteCategories
+            }
+        case .page:
+            remotePost.type = "page"
```
diegoreymendez: These 11 lines seem to be a perfect copy of the 11 lines in `saveAndUploadPost(...)`.  Any thoughts on possible improvements?

WordPress/WordPressShareExtension/SharePostTypePickerViewController.swift
```diff
@@ -0,0 +1,195 @@
...
+        let postTypeRow = PostType.allCases.firstIndex(of: postType)!
+        return IndexPath(row: postTypeRow, section: 0)
+    }
+}
+
+// MARK: - Actions
+
+extension SharePostTypePickerViewController {
+    @objc func cancelWasPressed() {
+        _ = navigationController?.popViewController(animated: true)
```
diegoreymendez: There's some space to improve this line of code.

WordPress/WordPressShareExtension/SharePostTypePickerViewController.swift
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    @objc func selectWasPressed() {
+        if originallySelectedPostType != selectedPostType {
+            onValueChanged?(selectedPostType)
+        }
+        _ = navigationController?.popViewController(animated: true)
```
diegoreymendez: Same as above.

WordPress/WordPressShareExtension/ShareModularViewController.swift
```diff
@@ -244,6 +246,22 @@ extension ShareModularViewController {
...
+    func showPostTypePicker() {
+        guard isPublishingPost == false else {
+            return
+        }
+
+        let typePicker = SharePostTypePickerViewController(postType: shareData.postType)
+        typePicker.onValueChanged = { [weak self] postType in
+            self?.tracks.trackExtensionPostTypeSelected(postType.rawValue)
+            self?.shareData.postType = postType
+            self?.refreshModulesTable()
```
diegoreymendez: As none of these lines will run when `self` is `nil` then maybe we can unwrap self on top and bail out immediately.

diegoreymendez (commented):
The feature works as requested.

I've left some comments in the code.

WordPress/Tracks+ShareExtension.swift
ghost: No reason, it's a copy/pasted from the other track methods in the file. You're right, it's unnecessary. I'll clean up the other ones in there, too.

WordPress/WordPressDraftActionExtension/Tracks+DraftAction.swift
ghost: Had my search filter set to the ShareExtension folder only, missing this one. 🤦

WordPress/WordPressShareExtension/SharePostTypePickerViewController.swift
ghost: So with this Picker, I've tried to keep it as close to the other pickers as possible. This line appears in those, too. It does seem like, for this new picker, I should probably not do it like this (others must have been written before `popViewController` was annotated with `@discardableResult`), but what should my scope be for things like this? 

I can see multiple other instances of this, including outside of the Share Extension - should I correct it only in my new file, in all files in the Extension, or across the app?

WordPress/WordPressShareExtension/AppExtensionsService.swift
ghost: Perhaps it makes more sense for these two methods to accept the `ShareData` as a parameter, and then a new `RemotePost` initialiser could be made that accepts it, and we move the shared logic into there. Or, have the methods accept `RemotePost` and move that creation out of both. Either would be a good change, I think; @diegoreymendez would you have a preference?

```
saveAndUploadPost(shareData: ShareData, siteID: Int, [...])
uploadPostWithMedia(shareData: ShareData, siteID: Int, [...])
```
vs
```
saveAndUploadPost(remotePost: RemotePost, [...])
uploadPostWithMedia(remotePost: RemotePost, [...])
```

WordPress/WordPressShareExtension/SharePostTypePickerViewController.swift
diegoreymendez: The scope, as I see it, is to make any added code clean and tight.  I'd definitely avoid making changes throughout the App as it'd be far beyond the scope of this PR.

I agree though that this probably was added like this back when the result wasn't `@discardableResult`, but that's just a guess on my side.

But even then, when copying code from elsewhere my take is that we're signing it, as both the editor and reviewer.

That's not going to save us from having legacy code that's outdated, but at least we can do something for the new code!

Thanks for the explanation!

WordPress/WordPressShareExtension/AppExtensionsService.swift
diegoreymendez: Great observation!  I think both solutions are acceptable, and in fact I would encourage you to pick the one that best suites your preference.

WordPress/WordPressShareExtension/SharePostTypePickerViewController.swift
ghost: Agreed on all points, I'll tidy up what I've added 👍

WordPress/WordPressShareExtension/ShareModularViewController.swift
osullivanchris: > My personal opinion is we can enable it in the share extension, and eventually add it in the Page settings in the App later on.

that seems fine to me.

> but are you certain you can add Tags to a Page?

from a very brief exploration it seems possible for me on WordPress.com sites. For a dotorg site it is not there by default, but I think it can be enabled with a plugin.

WordPress/WordPress.xcodeproj/project.pbxproj
```diff
@@ -2018,6 +2018,8 @@
 		C99B08FC26081AD600CA71EB /* TemplatePreviewViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = C99B08FB26081AD600CA71EB /* TemplatePreviewViewController.swift */; };
 		CB1FD8D826E4BBAA00EDAF06 /* SharePostTypePickerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = CB48172926E0D93D008C2D9B /* SharePostTypePickerViewController.swift */; };
 		CB48172A26E0D93D008C2D9B /* SharePostTypePickerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = CB48172926E0D93D008C2D9B /* SharePostTypePickerViewController.swift */; };
+		CBF6201326E8FB520061A1F8 /* RemotePost+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CBF6201226E8FB520061A1F8 /* RemotePost+Extensions.swift */; };
```
diegoreymendez: I'm adding this here, in lack of a better location to add this comment.

Considering that this file is most likely going to be specifically for the share extension, could be rename this to something more specific, like `RemotePost+ShareExtension.swift` or `RemotePost+ShareData.swift`?

WordPress/WordPressShareExtension/RemotePost+Extensions.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+import WordPressKit
+
+extension RemotePost {
+
+    /// Create a new Remote Post from the data provided by the Share and Action extensions.
+    /// - Parameters:
+    ///   - shareData: The data from which to create the Remote Post
+    ///   - siteID:: Site ID the post will be uploaded to
```
diegoreymendez: Typo: `::`

WordPress/WordPressShareExtension/RemotePost+Extensions.swift
```diff
@@ -0,0 +1,30 @@
...
+    convenience init(shareData: ShareData, siteID: Int) {
+        self.init(siteID: NSNumber(value: siteID),
+                   status: shareData.postStatus.rawValue,
```
diegoreymendez: The indentation seems to be off here.

diegoreymendez (commented):
I think this final change offers a great improvement to our legacy code.  Thumbs up on how you implemented it :)

I've added a few minor comments.

WordPress/WordPress.xcodeproj/project.pbxproj
ghost: I like the 2nd suggestion, more specific 👍

diegoreymendez (approved):
This is good to go.  Can I ask you to resolve the merge conflicts @autoblork ?

Issue #17108, opened 2021-09-02
Comments: Create the new content cell


dvdchr:
Refs #17087

Since the comment content cell is quite large, I'll break down the implementation further so the PR chunks are slightly more manageable. As such, this issue is created to track all the work items for the comment content cell.

Here are some designs from Zeplin:

Description | Screenshot
--|--
Short content | ![Screen Shot 2021-09-02 at 19 02 56](https://user-images.githubusercontent.com/1299411/131840673-7afe31bd-751e-4b4f-a2a4-14366ee10630.png) 
Long content with image | ![Screen Shot 2021-09-02 at 19 16 06](https://user-images.githubusercontent.com/1299411/131841977-049eecd0-1fb6-4bbc-b323-a93ad1ff29f0.png) 
Content with image and blockquote | ![Screen Shot 2021-09-02 at 19 16 30](https://user-images.githubusercontent.com/1299411/131841991-47865285-1b92-4c19-bfa8-abbfd5c132c2.png)

Some notes:

- Content will be displayed using a `WKWebView`.
- Some functionalities of the cell may be disabled, depending on site type and account permissions.

## Todo

- [x] #17109
- [x] #17129
- [x] Handle button taps: 
  - [x] #17269
  - [x] #17273
- [x] #17253
- [x] #17147
- [x] #17254

PR #17129, opened 2021-09-07
Comment Detail: Render content in web view

dvdchr:
Refs #17108, #17087, `aXjNk0P-zpl`

This adds the web view implementation to render the comment content. Some notes:
- Fixed a little bug when fetching the user's avatar URL.
- `richCommentTemplate.html` and `richCommentStyle.css` were added to stylize the contents according to the designs.
- Navigation requests from the web view are temporarily disabled for now. i.e. links look clickable but will do nothing (for now). It'll be implemented in the next PR.
- To handle content resize on orientation change, I ended up reloading the contents (i.e. calling `loadHTMLString` on the web view) again. 
  - I've tried other methods (`reload()`, through javascript's `location.reload()`, re-querying `scrollHeight`, ...) to no avail. The resize only works by calling `loadHTMLString` again. Somewhat expensive, I know. 😅 But it's what works so far and I've yet to find a better way.
- All resources within the comment content (images, video thumbnails) are loaded synchronously for now. For heavy comments (e.g. with >5 images), it may take a couple of seconds for the web view to fully render the content.
  - I'll take a look at loading the images asynchronously in the next PR. My assumption is it'll _at least_ cause the table to recalculate the cell height after each image completes loading, but if the method proves too tricky, I'll defer the async improvement and instead show a loading animation on top of the web view for the much simpler approach.
- Lastly, cell/component visibility based on user permission will be implemented in the upcoming PR.

## To test

- Ensure that the `New Comment Detail` flag is enabled.
- Go to My Site > Comments, and pick any comment.
- Verify that the comment content is displayed correctly (and beautifully ✨). 
  - Note that you may need to jump to multiple sites and hunt for tricky comments that may break the layout or appear strange. If you find them, please let me know! 
- Try changing to landscape orientation, and verify that the content is resized properly.

## Regression Notes
n/a. Feature is still hidden behind a flag.

n/a. Feature is still hidden behind a flag.

n/a. Feature is still hidden behind a flag.

PR submission checklist:


WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -32,36 +35,123 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
...
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        // when screen orientation changed, reload the HTML content to ensure that the content height accounts for the new width.
+        guard let cachedContent = htmlContentCache,
+           traitCollection.verticalSizeClass != previousTraitCollection?.verticalSizeClass
+            || traitCollection.horizontalSizeClass != previousTraitCollection?.horizontalSizeClass else {
+            return
+        }
+
+        webViewHeightConstraint.constant = 1
```
dvdchr: Forgot to add a code comment here, why I needed to reset the constraint before updating it with the new one. Will add this tomorrow.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -168,4 +265,7 @@ private extension String {
                                                         + "%1$d is a placeholder for the number of Likes.")
     static let pluralLikesFormat = NSLocalizedString("%1$d Likes", comment: "Plural button title to Like a comment. "
                                                 + "%1$d is a placeholder for the number of Likes.")
+
+    // pattern that detects empty HTML elements (including HTML comments within).
+    static let emptyElementRegexPattern = "<[a-z]+>(<!-- [a-z\\/:]+ -->)+<\\/[a-z]+>"
```
dvdchr: I did find some parameters embedded in the HTML comments, e.g.
```html
<!-- wp:image {"id":10675,"sizeSlug":"large","linkDestination":"none"} -->
```
which escapes the regex pattern. Will fix this tomorrow!

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -32,36 +35,128 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
...
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        // when screen orientation changed, reload the HTML content to ensure that the content height accounts for the new width.
+        guard let cachedContent = htmlContentCache,
+           traitCollection.verticalSizeClass != previousTraitCollection?.verticalSizeClass
+            || traitCollection.horizontalSizeClass != previousTraitCollection?.horizontalSizeClass else {
+            return
+        }
+
+        webViewHeightConstraint.constant = 1
+        webView.loadHTMLString(cachedContent, baseURL: nil)
```
dvdchr: Should have passed `Self.resourceURL` here so the CSS file is properly loaded. Will fix this tomorrow!

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -32,36 +35,128 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
...
+        let htmlString = String(format: templateFormat, sanitizedContent)
+
+        self.htmlContentCache = htmlString
+        webView.loadHTMLString(htmlString, baseURL: Self.resourceURL)
 
         // TODO: Configure component visibility
     }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        // when screen orientation changed, reload the HTML content to ensure that the content height accounts for the new width.
```
aerych: It looks like this won't be triggered in all situations.  I tried testing with an iPad Pro 9.7 simulator. The comment detail is shown in as the detail of a UISplitViewController.  In this scenario the size will change on rotation but the trait collection does not. 
Maybe this is something that could be done with `viewWillLayoutSubviews` or `viewDidLayoutSubviews`?

WordPress/Resources/HTML/richCommentStyle.css
```diff
@@ -0,0 +1,189 @@
...
+a.mention {
+    text-decoration: none;
+    background-color: var(--mention-background-color);
+    padding: 0 3px;
+    -webkit-border-radius: 3pt;
+}
+
+/** IMAGES **/
+
+.wp-block-image img {
```
aerych: We might need to have a rule for regular `img` tags that aren't part of gutenberg blocks. I tested with a non-gutenberg comment that had an image tag. The portrait image had some issues with layout. The width was constrained to width of the web view but the aspect ratio was lost resulting in the image having it's natural height. 
I tried adding 
```
img {
    margin: 0 auto; /* align center */
    width: auto;
    max-width: 100vw;
    height: auto;
}
```
This seemed to fix the issue with the image tho I'm not sure if it's all necessary or would have other side effects we don't want.

aerych (changes requested):
Heya @dvdchr  👋 
Looking sharp!  I tested in in the iPhone Pro 11 simulator and the iPad Pro 9.7 simulator and things mostly looked good.  I did spot a couple of things and left comments about them. Lemme know what you think.  
Back to you sir!

aerych:
For some light stress testing I created [this comment](https://aerych.wpsandbox.me/2021/07/27/test-image/#comment-19) on a test self-hosted site that includes (I think) all of the default allowed tags for comments, plus images and YouTube embeds (which are not allowed by default).  The comment looks pretty hideous in a browser but it looks rather nice in the app. :) 
Simulator Screen Shot - iPhone 11 Pro - 2021-09-07 at 18 11 51
Simulator Screen Shot - iPhone 11 Pro - 2021-09-07 at 18 11 55
Simulator Screen Shot - iPhone 11 Pro - 2021-09-07 at 18 12 04


I did notice there is some trouble with the the YouTube embed sizing properly.  I'm not sure if that's because of the error screen or something else.  Something we'll want to look into.  Probably we'll have a period of finding lots of one-off quirky things (typical of this sort of change and to be expected) but we'll get 'em sorted and then we'll have some rock solid webview comments. :)

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Ah, good catch! In this case, I think I'll use `viewWillTransition` instead to detect size changes. 🙂 Will address this soon!

WordPress/Resources/HTML/richCommentStyle.css
dvdchr: Good catch again! 🙂 Will fix this in the upcoming commit.

dvdchr:
Appreciate the testing effort! 👍 Thank you.

> Probably we'll have a period of finding lots of one-off quirky things ...

Yeah... Also, considering how many blocks there are, I'm not sure if they would all look great, since the CSS is homegrown and thus, limited. 😆 

Additionally, I've found that embedded tweets are causing problems with height calculation since it asynchronously fetches Twitter's widget.js library and resizes the tweet component. This is rather tricky since I've yet to find a way to block async Javascript without affecting the Javascript code that's used to calculate document height. 

The content is still vertically scrollable, it still doesn't look good since the detail screen now has two vertically scrollable components. Although this only applies to comments with embedded Tweet blocks. I'll take a look at this in the next PR, and will address it if it's not taking up too much time. 

The rough idea is instead of using `evaluateJavascript`, I should use `WKUserContentController` and move the height calculation there. Then, I should be able to hook into Twitter's widget library and listen after the tweet component has finished rendering. Finally, call the height calculation function so it could trigger the table view to resize the cell. 

> I did notice there is some trouble with the YouTube embed sizing properly.

Yeah, for this I figured it's better to restrict everything from the start, and allow elements to scale past `100vw` explicitly (e.g. tables, custom emoji images):
```css
* {
  max-width: 100vw;
}
```
---

**Update**: this is now addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17129/commits/1337847fbee3cd3240a33469a7a30c25b4d643f6.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17129/commits/550a4d2732c9643ee2cd23e60bb09dafdcba6101!

WordPress/Resources/HTML/richCommentStyle.css
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17129/commits/1337847fbee3cd3240a33469a7a30c25b4d643f6.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17129/commits/2c0ae96da04e1f7de9dedbce7851f4be27ca7ccc.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17129/commits/550a4d2732c9643ee2cd23e60bb09dafdcba6101 as part of the refactor to `viewWillTransition` from `traitCollectionDidChange`.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: This line has been removed, so it's no longer needed.

dvdchr:
@aerych this is ready for a second round of review! 🙂

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -68,6 +68,16 @@ class CommentDetailViewController: UITableViewController {
         configureRows()
     }
 
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        super.viewWillTransition(to: size, with: coordinator)
+
+        // when an orientation change is triggered, recalculate the content cell's height.
+        guard let contentRowIndex = rows.firstIndex(where: { $0 == .content }) else {
```
aerych, line 75: Nice.

aerych (approved):
Heya @dvdchr 👋 
Changes look good! Confirming portrait images and youtube embeds are sizing better. :) The styles are preserved when rotating now and sizes are updated. 

Simulator Screen Shot - iPhone 11 Pro - 2021-09-08 at 13 41 17

I noticed if I got very aggressive with rotation I could get into a state where the content height was higher than the actual content. I think it would be pretty hard to do this with an actual device so not a blocker imho.
Simulator Screen Shot - iPhone 11 Pro - 2021-09-08 at 13 43 05

One thing we might add to a list of things to improve as we go is centering embedded content. What do you think? 
Simulator Screen Shot - iPhone 11 Pro - 2021-09-08 at 13 41 56

Nice work sir.  Let's :shipit: !

dvdchr:
Thanks @aerych !

> I noticed if I got very aggressive with rotation I could get into a state where the content height was higher than the actual content. I think it would be pretty hard to do this with an actual device so not a blocker imho.

Good catch, and I agree with you. While this is not something crucial to address, I'll note it down with low priority for future reference. Or at least to keep the issue tracked.

> One thing we might add to a list of things to improve as we go is centering embedded content.

It looks like this is because the reply content is not using blocks. Here's an example HTML content:
```html
Here\'s an example of a YouTube embed:

<iframe width=\"560\" height=\"315\" src=\"https://www.youtube.com/embed/dQw4w9WgXcQ\" title=\"YouTube video player\" frameborder=\"0\" allow=\"accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture\" allowfullscreen></iframe>

Let\'s see how this will be rendered on the mobile side.
```

Unlike blocks, the texts are not wrapped in a paragraph element, thus kinda mucking up with the spacing. I'll try to explore if I can improve the spacing for plain text / non-Gutenberg block comments in the next PR.

PR #17131, opened 2021-09-08
Merge editorialized release notes for 18.2

mokagio:
Please forgive the long commit history. I was curious to try my `msgid` fix, which is merged in the release toolkit `develop` but hasn't been release yet, in another codebase but pointed to `trunk` instead of `develop` and, because `update_appstore_strings` pushes, I didn't have a chance to rebase the mistake away. FWIW, 69e4335 and confirm 7b3ecb5 one more time that the fix is legit.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,22 @@
+GIT
+  remote: git@github.com:wordpress-mobile/release-toolkit
+  revision: 964e732e2996730beb04953a6a0299437ed99c5e
+  branch: trunk
+  specs:
+    fastlane-plugin-wpmreleasetoolkit (1.4.0)
+      activesupport (~> 5)
+      bigdecimal (~> 1.4)
+      chroma (= 0.2.0)
+      diffy (~> 3.3)
+      git (~> 1.3)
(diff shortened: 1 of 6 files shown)


Gemfile.lock
```diff
@@ -245,13 +245,13 @@ GEM
...
     netrc (0.11.0)
-    nokogiri (1.12.3)
+    nokogiri (1.12.4)
       mini_portile2 (~> 2.6.1)
       racc (~> 1.4)
     octokit (4.21.0)
       faraday (>= 0.9)
       sawyer (~> 0.8.0, >= 0.5.3)
-    oj (3.13.1)
+    oj (3.13.4)
```
mokagio, line 254: These gems patch version changes are a byproduct of my dance with the release toolkit version. They snuck in there because I had to go through two versions, so instead of `git revert` two commits I manually rolled back the `Pluginfile` and run `bundle update fastlane-plugin-wpmreleasetoolkit`. 

I don't think they present any issue, so I decided not to spend any extra time to amend them.

oguzkocer (approved):

PR #17145, opened 2021-09-12
Add new TimeZoneSuggester

### LINKED ISSUE #11450: Timezone selection can be confusing if some major cities is not included in the timezone dropdown list
It can be confusing for some users not to be able to find their city in the timezone dropdown list. For example, Mumbai is not listed and Kolkata would need to be selected in that case.

Timezone selection currently follows WordPress Core & Calypso which offers a list of timezones from the PHP zoneinfo database, maintained by PHP. See <a href="https://core.trac.wordpress.org/ticket/3962#comment:67">core.trac/3962#comment:67</a>. Because creating and managing a custom list of timezones may not be something we would want to do, one way to improve this experience would be to match the behavior of a time zone suggester—one has been proposed in [core.trac/10660](https://core.trac.wordpress.org/ticket/10660) and was already implemented in Calypso.

Screenshots for reference:

IMG_2462 &nbsp; IMG_2461
<sup>Settings > Timezone > [Search] in WP Internal 12.1.0.20190408 on iPhone 6S iOS 12.2.</sup>

Screen Shot 2019-04-10 at 3 04 00 PM
<sup>Settings > Timezone in WordPress 5.1.1.</sup>

Screen Shot 2019-04-10 at 3 06 06 PM
<sup>Settings > Timezone in Calypso on WordPress.com as of 2019-04-10.</sup>

h/t @charliescheer for raising the issue.
<sup>(internal reference: p4a5px-2nA-p2)</sup>

sla8c:
Timezone selection from site settings in iOS app is outdated in comparison with WordPress web and Android. This PR brings iOS on par with them by adding a TimeZone Suggester feature.  See https://github.com/wordpress-mobile/WordPress-iOS/issues/11450

Ref #11450 

### To test:
**To test TimeZone Suggester feature:**
1.  Navigate to My Site -> Site Settings -> Time Zone

2.  Make sure the time zone that is suggested is consistent with the timezone of the device / simulator you are testing on

3.  Tap the suggested timezone

4.  Step 3 will navigate you back to Site Settings screen where you should see the Suggested time you have chosen

5.  Tap Time Zone again from Site Settings screen

**To test search results works normally and has no regression:**

6. Search for a timezone (example Casablanca).  TableView results should filter and show Casablanca.

7. Tap Casablanca

8. Step 7 will navigate you back to Site Settings screen where you should see the Suggested time you have chosen (Casablanca)

image

## Regression Notes
TimeZone selector screen was changed - unintended impact could be user would not be able to change their TimeZone

Initial tests are manually tested.  Screenshots above

No automated tests existed previously and ViewController is not in a state that facilitates automated testing

PR submission checklist:


sla8c:
@Gio2018 I've implemented a first version of the TimeZoneSuggester.  Unfortunately I don't have permissions to label this with "DO NOT MERGE" or assign it to you.

Also below are 2x screenshots with different sizing for the TimeZoneSuggester.  One version uses a smaller font like on Android and Web and the version in code uses a larger font version.  In code I have kept the larger font version as I personally think it offers a better "hit area" for users with larger fingers etc and of course I'm happy to change it back if you (or someone) thinks otherwise

image

image

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
```diff
@@ -0,0 +1,62 @@
+import UIKit
```
Gio2018: It might be worth creating a `Time Zone` group under `Tools` since we have a few files that concern this feature now

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
```diff
@@ -0,0 +1,62 @@
...
+
+final class TimeZoneSearchBar: UIView {
```
Gio2018: Given that we should rarely need subclassing, any specific reason why this class is marked `final`?

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
Gio2018: It seems that this `UIView` contains more than just a search bar, so I'm wondering if we should rename it?

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        suggestionLabel.text = Localization.suggestion
+        suggestionButton.addTarget(self, action: #selector(buttonTapped), for: .touchUpInside)
+    }
+
+    @objc private func buttonTapped() {
+        tapped?()
+    }
+
+    func configureSearchWrapperView(searchBar: UISearchBar) {
```
Gio2018: I wonder if we should make the search bar mandatory? You can still create an instance of this type and forget to call this method, which will create it without a search bar?
Any suggestion on how to refactor this type?

Gio2018 (commented):
Hey @sla8c , this seems like a good start!
I left some code comments. Other than that, in the description can you please change the default **Fixes** with a more generic **Ref** label? This is because this is only the first PR that concerns the project and will not address all the tasks described in the issue. Thank you!!

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
sla8c: Sounds good + that's done!

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
sla8c: Sorry if that's confusing; when I was doing going through the code test part of the process, I had seen that in WooCommerce iOS almost everything is marked final where it can be so I had presumed that this was a best practice here!  In terms of other reasoning, there is a small performance benefit also mentioned in Apple's Swift blog [here](https://developer.apple.com/swift/blog/?id=27).  I was trying to be consistent with the codebase / org so maybe I applied a standard from a different codebase incorrectly.  I've removed final 😄

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
sla8c: Done!  I've renamed it to `TimeZoneSearchHeaderView`

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
sla8c: Sounds good!  I've refactored so that the `searchBar` and default `TimeZoneIdentifier` which will be the suggestion is required to be passed into the factory method (`makeFromNib(searchBar: UISearchBar, timezone: String)`)

sla8c:
Hi @Gio2018 I've completed those updates and edited the description too.  Thx!

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
Gio2018: I think the reasoning is pretty solid, so unless we foresee subclassing, I'd say we can re-instate the final

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneSearchHeaderView.swift
```diff
@@ -0,0 +1,54 @@
...
+    var tapped: (() -> Void)?
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+    }
+
+    class func makeFromNib(searchBar: UISearchBar, timezone: String) -> TimeZoneSearchHeaderView {
+        let view = Bundle.main.loadNibNamed(Constants.nibIdentifier,
+                                        owner: self,
+                                        options: nil)?.first as! TimeZoneSearchHeaderView
```
Gio2018: The force unwrap here Is less than ideal. It might be unlikely but if, for any reason, that `.xib` file name changes, you would not have any build errors but the app would crash when users try to change the time zone.
Any ideas on how we can improve this?

Gio2018 (changes requested):
Hey @sla8c , left another comment to consider!

WordPress/Classes/ViewRelated/Tools/Views/TimeZoneSearchBar.swift
sla8c: Done! thx

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneSearchHeaderView.swift
sla8c: @Gio2018 I've made a change so that the `TimeZoneSearchHeaderView.makeFromNib` class function returns an optional `TimeZoneSearchHeaderView` and the ViewController will use the new `TimeZoneSearchHeaderView` component or fallback to the default `SearchController.SearchBar`.

Couple of comments on to give you an idea of other variations I tried as well:

- I also tried the below which is cleaner in code, but unfortunately the double assignment to tableView.tableHeaderView resulted in a poorer user as the SearchBar sometimes didn't appear which obviously doesn't help our users!

image


- I also thought about making the makeFromNib return an empty TimeZoneSearchHeaderView as I see that paradigm has been used in another implementation (s/shot below) however I reasoned that having the UI fallback to the default SearchBar is a better user experience.

image


- Re: the force unwrap I had thought that the standard as I had seen it done in these other TableHeaders so was trying to be consistent but this way is definitely safer! 👍 

```
    class func makeFromNib() -> MyProfileHeaderView {
        return Bundle.main.loadNibNamed("MyProfileHeaderView", owner: self, options: nil)?.first as! MyProfileHeaderView
    }

    @objc class func makeFromNib() -> NoteTableHeaderView {
        return Bundle.main.loadNibNamed("NoteTableHeaderView", owner: self, options: nil)?.first as! NoteTableHeaderView
    }

```

Happy to hear your thoughts 😄

Gio2018 (commented):
The PR worked as expected, please include the requests that I made on Slack and we are good to go!

sla8c:
> The PR worked as expected, please include the requests that I made on Slack and we are good to go!

Hi @Gio2018 Updated PR notes.  I did not update release-notes.txt yet since this PR does not complete the issue - thx!

Gio2018 (approved):
Good job @sla8c !

PR #17147, opened 2021-09-13
Comment Detail: Toggle components visibility based on permission

dvdchr:
Refs #17087

This PR configures the visibility of several components in Comment Detail. 

- The web address view should be hidden if the author URL is empty, or an invalid URL.
- The email and IP address fields are only visible for moderator roles, i.e. `Editor` and `Administrator`.
- In the content cell, the reaction bar (containing Reply and Likes button) is only visible when the site is either self hosted, **OR** the site is either dot com or Atomic and the comment is approved.
  - In short, the reaction bar is only disabled when `the site is dotcom or Atomic` **AND** the comment is not approved **AND** the user viewing the comment doesn't have a moderating role.
- Lastly, the accessory button is only enabled when the comment is approved.

## To Test

Ensure that the `New Comment Detail` feature flag is enabled. 

#### Web address visibility
- Go to My Site > Comments, and select a Comment where the author URL is present.
- Verify that the Web address field is visible.
- Select another comment where the author URL is not present.
- Verify that the Web address field is hidden.

#### Accessory button visibility
- Select any approved comment, and verify that the accessory button is visible.
- Select a spam or pending comment, and verify that the accessory button is hidden.

#### With a dotcom site, using an account that CANNOT moderate the site:
- Select any approved comment.
- Verify that the email address and the IP address field is **hidden**.
- Verify that the Edit button is NOT visible.
- Verify that the Reply and Like button is visible.

#### With a dotcom site, using an account that can moderate the site (Administrator or Editor):
- Select any approved comment.
- Verify that the email address and IP address field is **visible**.
- Verify that the Edit button is visible.
- Verify that the Reply and Like button is visible.

#### With a self-hosted site:
- Select any approved comment.
- Verify that the email address and IP address field is **visible**.
- Verify that the Reply and Likes button is visible.

## Regression Notes
n/a. Feature is hidden behind a feature flag.

n/a. Feature is hidden behind a feature flag.

n/a. Feature is hidden behind a feature flag.

PR submission checklist:


ScoutHarris:
Hey @dvdchr .

The app crashes when saving an edited comment. Cancelling a comment edit works fine.

- On iPad or iPhone.
- `Edit` a comment.
- Made a change.
- Select `Done`.
- The app crashes with: `Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value: file WordPress/CommentContentTableViewCell.swift, line 74`

Screen Shot 2021-09-13 at 2 03 34 PM

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -33,11 +33,14 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
 
     @IBOutlet private weak var webView: WKWebView!
     @IBOutlet private weak var webViewHeightConstraint: NSLayoutConstraint!
+    @IBOutlet weak var webViewBottomConstraint: NSLayoutConstraint!
```
ScoutHarris: Should this be `private`? Since all the other `IBOutlet`s are...

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -93,12 +118,15 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
 
         updateLikeButton(liked: comment.isLiked, numberOfLikes: comment.numberOfLikes())
 
-        // configure comment content
+        // Configure feature availability.
+        isReactionEnabled = !comment.isReadOnly()
+        isCommentLikesEnabled = comment.blog?.supports(.commentLikes) ?? false
```
ScoutHarris: The old comment detail shows Like based on what the blog and the comment support:
`cell.isLikeEnabled = [self.comment.blog supports:BlogFeatureCommentLikes] && self.userCanLikeAndReply;`

I assume the new one should be the same, that is:
`isCommentLikesEnabled = comment.blog?.supports(.commentLikes) ?? !comment.isReadOnly()`

ScoutHarris (changes requested):

ScoutHarris:
Hey @dvdchr .

I noticed that the `Email address` field is displayed when there is no email. I would expect it to be hidden like the `Web address` is when it's empty. AFAIK, it is perfectly valid not to have an email address. You can see this by 1) the Edit Comment view allowing it to be cleared, and 2) the web displays a specific label in this scenario.

Screen Shot 2021-09-13 at 2 40 49 PM


I will admit, I'm not sure how you can reply without an email address, but I do recall @aerych mentioning something about replying anonymously, which results in no email.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
ScoutHarris: What do you think about using a stack view to display the cell contents? 
- When the reaction bar is hidden the stack view will automatically resize, negating the need to manually modify constraints. 
- It should also resolve the crash due to missing constraint.
- I think it will also facilitate showing the new moderation bar. It will be displayed conditionally in this cell (ref pbArwn-2ks#comment-3746), so having a stack view means showing/hiding it should just work like it would for the reaction bar.

ScoutHarris:
This is probably unrelated to this PR, but I just noticed this warning:

Screen Shot 2021-09-22 at 12 12 40 PM

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Ah yep. Actually, the comment button is contained in a view that has its visibility controlled through `isReactionEnabled`. So, when the value of `!comment.isReadOnly()` is `false`, both the like and comment buttons will be hidden regardless of whether they're visible or not.

But I guess this behavior is not very clear just from a single glance unless one also checks out the XIB – and things may change in the future and it might introduce a bug. I'll adjust this soon! Thanks. 🙂

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: > Should this be private? Since all the other IBOutlets are...

Whoops! Good catch. Thanks!

> What do you think about using a stack view to display the cell contents?

Yeah, I can do that. Thanks for the suggestion!

dvdchr:
> I noticed that the Email address field is displayed when there is no email. I would expect it to be hidden like the Web address is when it's empty.
> I do recall @aerych mentioning something about replying anonymously, which results in no email.

Oh, TIL! I'll adjust this soon.

dvdchr:
> This is probably unrelated to this PR, but I just noticed this warning:

Good catch! Since this icon is not available in Gridicons, I'll backport this for iOS 13 by adding this specific icon to our assets.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: This is addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17147/commits/8965bca1460c6d48119fb9d8d3e6b7c27b692fdf. I've tested the Edit Comment flow and confirmed that the crash you mentioned in https://github.com/wordpress-mobile/WordPress-iOS/pull/17147#issuecomment-918537809 no longer happens. Thanks!

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17147/commits/336c31fe0f248044a21061303ad718ee66ca0e4c!

dvdchr:
Hey @ScoutHarris, this should be ready for your n-th round of review. 😅 Thank you for the reviews!! Here's a gist of the update:

- Used stack view to stack the comment content subviews. Also verified that the [crash you've found](https://github.com/wordpress-mobile/WordPress-iOS/pull/17147#issuecomment-918537809) no longer happens after switching this to stack view.
  - Heads up, this will probably cause conflicts on https://github.com/wordpress-mobile/WordPress-iOS/pull/17211! 😬 
- Updated the conditionals for `isCommentLikesEnabled` to be more clear.
- Addressing [your comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/17147#issuecomment-918562797), empty email field is now hidden!
- Backported the `arrowshape.turn.up.backward` SFSymbols to iOS 13.

dvdchr:
It looks like I might've left the Jetpack target unchecked, which may cause the CI jobs to fail. I'll fix this tomorrow!

ScoutHarris (approved):
Looking good!

:shipit:

PR #17148, opened 2021-09-13
Change ReaderPost `card` relationship from "One to many" to "To one"

leandroalonso:
This PR changes the `ReaderPost` `card` relationship from "One to many" to "To one".

### To test

First, install an earlier version of the app (not this branch).

1. Run the older version
2. Go to Discover
3. Save a post

Switch to this branch:

1. Run the app
2. Go to the "Saved" tab
3. Make sure the post is there
4. Interact with the post
5. Unsave it
6. Go to the Discover tab

## Regression Notes
The whole app, it's a Core Data migration. :)

Relied on unit tests migration and manually testing it.

-

PR submission checklist:

### DIFF
--- a/MIGRATIONS.md
+++ b/MIGRATIONS.md
@@ -3,6 +3,12 @@
 This file documents changes in the data model. Please explain any changes to the
 data model as well as any custom migrations.
 
+## WordPress 133
+
+@leandrowalonso 2021-09-13
+
+- `ReaderPost`: change `card` to "To One" relationship
+
 ## WordPress 132
 
 @momo-ozawa 2021-08-19
@@ -56,7 +62,7 @@ data model as well as any custom migrations.
 - `BlockEditorSettings`: added the attribute
     - `rawStyles` (optional, no default, `String`)
     - `rawFeatures` (optional, no default, `String`)
-    
+
 - `BlockEditorSettingElement`: added the attribute
     - `order` (required, 0, `Int`)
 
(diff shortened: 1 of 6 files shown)


aerych (changes requested):
Heya @leandroalonso 👋 
Thanks for wrangling this! 
I gave the patch a spin, and tested as described.  For the test I used the iPhone 11 iOS 14.5 Simulator, first running (a somewhat stale copy of) develop.  I saved three separate Discover feed posts, backgrounded the app, then switched to this branch. 
Something might be missing from the configuration because I see a crash logged with the error : `reason=Can't find mapping model for migration` and copies of the model revisions.  It's a big wall of text so I won't paste it here. 

I'm not seeing where we're handling migrating the many-to relationship to a one-to relationship. From [this doc](https://developer.apple.com/documentation/coredata/using_lightweight_migration) it sounds like a one-to can easily be migrated to a many-to (and no nead for a heavy-weight migration) but the reverse isn't true.  I mention this because I wasn't sure what change we were making in the `xcmappingmodel`, it looks like it's boilerplate.  Am I missing something?  Do we need to add a `NSEntityMigrationPolicy` instance to handle getting `ReaderPost.card.first` from revsion 132 and setting `ReaderPost.card` to it's value in revision 133? 

Side note, should we add a test for this to CoreDataMigrationTests?

leandroalonso:
Thank you so much for taking a look, @aerych.

> Side note, should we add a test for this to CoreDataMigrationTests?

Yeah, absolutely. 

Interesting that I didn't manage to crash the app here, but I had this feeling that something was really missing. I'm gonna fix those. :)

aerych:
> Interesting that I didn't manage to crash the app here, but I had this feeling that something was really missing

Maybe it has something to do with the posts I saved.  These are their links.  Maybe you can spot them in your version of the Discover feed:

https://jeffflesch.com/2021/09/09/a-love-so-thick/
https://danielswearingen.com/2021/09/08/portals-then-now-forever/
https://this-mans-journey.com/2021/09/06/fall-magic-came-early/

leandroalonso:
I'll try to tackle that in a few weeks, given the current conflicts I'll close this PR.

PR #17154, opened 2021-09-14
Filters out a8c p2s from Weekly Roundup candidate sites.

diegoreymendez:
Filters out a8c p2s from Weekly Roundup candidate sites.

@startuptester - This is relevant for testing moving forward.  After this PR we'll no longer get Weekly Roundup notifications in production for A8c p2s.

## Associated PRs:

WPKit: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/445

## To test:

1. Make sure you're on a fresh install - otherwise there may be a delay in the blogs getting synchronized, which is fine, but it may make the testing harder.
2. Go to Me > App Settings > Debug > Weekly Roundup
3. Make sure the toggle "Include a8c p2s" is OFF by default.  This won't be accessible in release builds.
4. Tap "Schedule Immediately".
5. Make sure you get notifications for sites that are not A8c p2s.

Test the same with the toggle ON, and you should get notifications for internal a8c sites too.

For extra security you should try running the same tests by launching the App and changing the date to the following monday after 10 am.  You may need to wait a few hours to see the Weekly Roundup notification.

## Regression Notes

1. Potential unintended areas of impact

Weekly Roundup.

2. What I did to test those areas of impact (or what existing automated tests I relied on)

The testing steps provided above should cover the changes.

3. What automated tests I added (or what prevented me from doing so)

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17177, opened 2021-09-17
Update WordPress Mocks to fix recent UI tests failures

mokagio:
The changes from the pair https://github.com/wordpress-mobile/WordPressKit-iOS/pull/445 https://github.com/wordpress-mobile/WordPress-iOS/pull/17154 added a new property to a model deserialized from _an already existing_ property in the JSON response (at least that's what I make of it by looking at the PRs and Git history see extra details [here](https://github.com/wordpress-mobile/WordPressMocks/pull/35).

Because the property was already there in the live responses, everything worked well when tested manually, but the UI tests failed because the stubbed response they used did not include that key in the JSON. The tests crashed when trying to assign the `nil` value stored in the `organizationId` property serialized from the stub response that didn't include the key.

## Related PRs

- [x] [WordPressMocks PR](https://github.com/wordpress-mobile/WordPressMocks/pull/35) needs to be merged and a new beta shipped before we can update this one to a tagged version and merge. Status Blocked

## How did this happen?

https://github.com/wordpress-mobile/WordPress-iOS/pull/17154 was merged even though Buildkite reported the UI tests failing. This is understandable because we (the Owl team) haven't fully migrated to Buildkite for WordPress iOS yet and the setup has been having a few occasional issues.

Moreover, the UI tests did not run in CircleCI. That's understandable, too, because they are optional and need to be unlocked manually. There's no clear rule of thumb to know when to run them and when it's safe to skip them and not have to wait for them.

The good news is that, once we'll complete the migration to Buildkite, the CI performance will be such that _UI tests will run on every commit_, so this kind of lag in discovering out-of-date tests should disappear.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -365,9 +365,9 @@ end
 ## ===================
 ##
 def wordpress_mocks
-  pod 'WordPressMocks', '~> 0.0.14'
+  # pod 'WordPressMocks', '~> 0.0.14'
   # pod 'WordPressMocks', :git => 'https://github.com/wordpress-mobile/WordPressMocks.git', :commit => ''
-  # pod 'WordPressMocks', :git => 'https://github.com/wordpress-mobile/WordPressMocks.git', :branch => 'task/jetpack-screenshots-dynamic-dates'
+  pod 'WordPressMocks', git: 'https://github.com/wordpress-mobile/WordPressMocks.git', branch: 'add-organization-id-to-me-sites'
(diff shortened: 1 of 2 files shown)


diegoreymendez (approved):
Once the pod is updated feel free to merge!

The reason I didn't notice the UI tests breaking is because I had enabled automerge on these PRs.

mokagio:
I updated the pod reference to the latest stable and run the UI tests on Buildkite and CircleCI: both passed 👍

PR #17178, opened 2021-09-17
3rd HACK week 2021 – Contact Support Self-Service Experiment

mokagio:
This is what I managed to get done during the 3rd HACK week of 2021. The idea is to present users with a self-service decision tree to address the most common questions asked to support which should ideally reduce the load and **get users the help they need faster**.

Something interesting about this PR is that I used the model described in my [microapps architecture post](https://increment.com/mobile/microapps-architecture/). I created a dedicated ContactUs framework which comes with its own microapp and did most of the development in there, enjoying a blazing fast compile time.

---

This is obviously not a finished PR, hence why I opened it as a draft. I don't have the bandwidth for working on this outside of the HACK week, unfortunately. Maybe someone else will pick it up and take over, or use this code as a reference for a new piece of work?

## Regression Notes
N.A.

N.A.

N.A.

---

- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. – _Not sure about this yet, depends whether we decide to go forward with this and merge from this branch_


WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
```diff
@@ -1,10 +1,15 @@
-import UIKit
```
mokagio, line 1: UIKit is implicitly imported in the app. We don't need to add it every time.

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
```diff
@@ -1,10 +1,15 @@
...
 import WordPressAuthenticator
 
 class SupportTableViewController: UITableViewController {
 
     // MARK: - Properties
 
+    lazy var contactSupportSelfServiceController = ContactSupportFlowController(
+        onSupportRequested: { [weak self] in self?.presentZendeskOrLoadForums() },
+        onHelpPageLoaded: { _ in }
+    )
```
mokagio, line 11: I decided to keep the contact us module agnostic of what support SDK we use and of how to log analytics events. It accepts closure to run to delegate those functionalities.

WordPress/Classes/ViewRelated/Support/SupportTableViewController.swift
```diff
@@ -183,24 +188,66 @@ private extension SupportTableViewController {
...
+            identityUpdated: { [weak self] identityUpdated in
+                if identityUpdated {
+                    self?.reloadViewModel()
+                }
+            }
+        )
+    }
+
+    private func loadForums() {
+        guard let url = Constants.forumsURL else { return }
```
mokagio, line 246: As a side note, I think we could remove the overhead of this `guard` by declaring `forumsURL` as a `URL` and force unwrapping its assignment.

https://github.com/wordpress-mobile/WordPress-iOS/blob/2725b02ab24e48b5102ca7db2194272f650e4286/WordPress/Classes/ViewRelated/Me/Account%20Settings/AccountSettingsViewController.swift#L368

The value is a hardcoded URL and the only way it can be `nil` is if a developer makes an error in spelling out that string. It seems like a good use case for force unwrapping.

twstokes:

PR #17210, opened 2021-09-23
Fix empty state layout bug in comments

### LINKED ISSUE #17201: Comments: Loading image is not shown correctly in iPad layout
### Expected behavior
The loading image / view is shown correctly, centered in the detail view controller.

### Actual behavior
The loading image is displayed incorrectly. Looks like a bug on the view's constraints causing the loading view to ignore the sidebar.

ipad_comments_loading-state

- Run the app on an iPad simulator.
- Log in and select any site.
- Tap My Site > Comments.
- The loading image should be shown as described above.

ealeksandrov:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/17201.

## Description

Method `adjustNoResultViewPlacement` gets called few times after viewDidLoad and on first call it takes wrong `UITableView` frame, with negative x,y origin values. This results in initial jumping out of viewport until another refresh in a second.

The fix is simple - adding a check that empty state view already added to hierarchy. This delays frame adjustment to moment when tableView frame is correct.
I've tried instead to use different frames/bounds/etc but it all results in various new animation/layout issues.

## To test:

1. Run the app on an iPad simulator.
2. Tap My Site > Comments.
3. Select Pending/Spam tab to have empty state displayed (if needed).
4. Switch to other section and back to comments.
5. Empty state should appear correctly in detail VC on right side.

iPhone layout checked - unaffected by this change.

## Screenshots

before|after
--|--
Simulator Screen Shot|Simulator Screen Shot

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
@@ -660,6 +660,7 @@ static NSString *RestorableFilterIndexKey = @"restorableFilterIndexKey";
 
 - (void)adjustNoResultViewPlacement
 {
+    // calling this too early results in wrong tableView frame used for initial state
     if(self.noResultsViewController.view.window == nil) {
         return;
     }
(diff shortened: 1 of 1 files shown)


ScoutHarris (approved):
Tested on various devices. Looks good!

:shipit:

ealeksandrov:
Thanks @ScoutHarris!

Issue #17200, opened 2021-09-20
Comments: add new Moderation Bar


ScoutHarris:
Create a new comment moderation bar that contains Pending, Approved, Spam, and Trash actions/states.

- [x] Create view to display bar.
  - [x]  #17211
  - [x] #17212
  - [x] #17235
  - [x] #17242
- [x] Display under comment content in Comment details.
  - [x] #17235
- [x] Show current status.
  - [x] #17244
  - [x] #17289
- [x] Handle actions.
  - [x] #17260
  - [x] #17267
  - [x] #17275
  - [x] #17279
  - [x] #17280
  - [x] #17281
  - [x] #17297
- [x] Snackbars:
  - [x] Show when actions succeed.
  - [x] Show when actions fail.
- [x] Accessibility.
- [x] Tracks.

Design discussion: pbArwn-2ks#comment-3746

From Zeplin:

<img width="344" alt="Screen Shot 2021-09-20 at 4 43 39 PM" src="https://user-images.githubusercontent.com/1816888/134086255-6c9ca068-9485-4dd2-b0d9-543c8c61868d.png">
<img width="340" alt="Screen Shot 2021-09-20 at 4 43 48 PM" src="https://user-images.githubusercontent.com/1816888/134086254-28ca9c36-d123-4ddc-9c43-ae79184e51a4.png">
<img width="341" alt="Screen Shot 2021-09-20 at 4 43 59 PM" src="https://user-images.githubusercontent.com/1816888/134086252-450159a3-0d96-41a6-bc7a-9c5aac179f12.png">
<img width="341" alt="Screen Shot 2021-09-20 at 4 44 06 PM" src="https://user-images.githubusercontent.com/1816888/134086249-6b04d827-aaaf-4e1d-9183-698f72572d82.png">

<img width="901" alt="moderation_dark" src="https://user-images.githubusercontent.com/1816888/135163138-585a37b2-daeb-4e81-a430-d5b246bce925.png">

From pbArwn-2ks#comment-3380:

<img width="442" alt="Screen Shot 2021-10-06 at 5 50 45 PM" src="https://user-images.githubusercontent.com/1816888/136299085-938d767a-688c-43ad-a973-fd13710abcbb.png">

PR #17211, opened 2021-09-23
Comment Moderation Bar: add initial view

ScoutHarris:
Ref: #17200

This adds an initial `CommentModerationBar` view with the `Pending` button. Tapping the button will toggle the UI to indicate selected or not. No action is actually performed yet.

NOTE: The bar is _temporarily_ displayed in `CommentContentTableViewCell`. Since there are [pending](https://github.com/wordpress-mobile/WordPress-iOS/pull/17147) changes to this cell, I will undo the changes here before merging and add it back later. (Hence the `DO NOT MERGE` label.)

To test:
- Go to My Site > Comments > Comment details.
- Verify the bar is displayed in the `CommentContentTableViewCell`, with the `Pending` button in the unselected state.
- Tap the `Pending` button. Verify the UI changes to the selected state.

| Unselected | Selected |
|--------|-------|
| light_unselected | light_selected |
| dark_unselected | dark_selected |

## Regression Notes

PR submission checklist:


ScoutHarris:
Hey @mattmiklic . Since I'll be using this same styling for all the buttons, would you mind giving this a look and see what you think? I wasn't sure how it was supposed to look in dark mode, so I took a stab. Thank you!

Edit: you may want to look at https://github.com/wordpress-mobile/WordPress-iOS/pull/17212 instead/also. It has all the buttons with dividers, so it is closer to the final UI.

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    func configureView() {
+        configureBackground()
+        pendingButton.configureFor(.pending)
+    }
+
+    func configureBackground() {
+        contentView.backgroundColor = .tertiaryFill
+        contentView.layer.cornerRadius = 15
```
ScoutHarris, line 45: I'll move all the style related setting to `WPStyleGuide+CommentDetail` after I finish the bar (i.e. when I get all the styles worked out).

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
```diff
@@ -0,0 +1,132 @@
...
+
+}
+
+// MARK: - Private Extension
+
+private extension CommentModerationBar {
+
+    // MARK: - Configure
+
+    func loadViewFromNib() -> UIView? {
```
dvdchr, line 33: Minor comment: what do you think about making this class conform to `NibLoadable` instead? Then we can remove this function and use the `loadFromNib()` static method instead.

dvdchr (changes requested):
Tested and it looks great! I've found one issue, though. I noticed in RTL layout, the vertical stack button looks off:

Simulator Screen Shot - iPhone 8 - 2021-09-24 at 17 48 14

Since there's already a `flipInsetsForRightToLeftLayoutDirection()` method in the `UIButton` extension, I wonder if we can just call this method on the view depending on the current layout?

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
ScoutHarris: Oh I tried that. 😄 

This is a bit different situation. Because `CommentContentTableViewCell` is rendering `CommentModerationBar` from a xib, the settings are different than they normally would be.

- `CommentModerationBar`'s File Owner has to be set to the class (it is usually not) so `CommentContentTableViewCell` can find it.

Screen Shot 2021-09-24 at 11 27 07 AM

- Because of ☝️ , the View's class cannot be set (it usually is). Doing so causes `init` to be called infinitely because Class is already set above.

Screen Shot 2021-09-24 at 11 27 20 AM


And this is where `NibLoadable` becomes unusable. It expects the View's class to match. In this case it does not, so `NibLoadable` crashes.

`let result = defaultBundle.loadNibNamed(defaultNibName, owner: nil, options: nil)?.first as? Self`

ScoutHarris:
Hey @dvdchr .

Thanks for noticing the RTL issue. Should be better now:

Screen Shot 2021-09-24 at 12 06 24 PM

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
dvdchr: Ah, got it. Thanks for explaining! 👍

dvdchr (approved):
LGTM! 🚢

Issue #16961, opened 2021-08-02
Updating tracking properties in Reader


momo-ozawa:
## Description

Add missing and update ambiguous properties as per the "Required code changes" column in a spreadsheet shared on p9xfpQ-1eg-p2.

This has already been implemented on the Android side by @ParaskP7 

Android Issue: https://github.com/wordpress-mobile/WordPress-Android/issues/14069
Android PR: https://github.com/wordpress-mobile/WordPress-Android/pull/14283

## Notes

Before we start tackling these changes, it would be good to break this task down into smaller subtasks. Doing so will give us a better picture of the scope / size of this issue.

## Tasks

- [x] Update current events with missing properties (excluding source) / remove duplicates / fix triggers for events [#17213](https://github.com/wordpress-mobile/WordPress-iOS/pull/17213)
- [x] Add logic for event properties calculation [#17219](https://github.com/wordpress-mobile/WordPress-iOS/pull/17219)
- [ ] Add source property to all events

ParaskP7:
👋 @momo-ozawa , @nikola-milicevic !

Thanks for the write-up! 🙏 

> Before we start tackling these changes, it would be good to break this task down into smaller subtasks. Doing so will give us a better picture of the scope / size of this issue.

I agree on that to 💯 .

On Android, we didn't do that and the PR ended up being too big. To be honest, some time constraints didn't help with that and thus I decided to create a single monster PR, just so that it can get merged after all, but you don't need to fall into that trap as well.

PS: On Android, we were also having a lot of legacy code that needed refactoring beforehand. Maybe the same applies to iOS, so be careful of that.
PSS: I was trying to add that comment yesterday but GitHub was down... 🤷 😅

PR #17213, opened 2021-09-24
Fix/16961 - Update existing events

nikola-milicevic:
Part of [#16961](https://github.com/wordpress-mobile/WordPress-iOS/issues/16961)

This PR updates Reader tracking events, so now they match proposal found in the spreadsheet shared on p9xfpQ-1eg-p2.

## Notes
- All reader tracking events (excluding `reader_reader_tag_followed` / `reader_reader_tag_unfollowed`) now contain: `blog_id`, `feed_id` and `follow` properties.
- `source` property will be added in a separate PR

## Review notes
- It is best to review this PR by going from commit to commit chronologically, since every commit addresses needed changes for a particular tracking event (including adding needed properties for a event, removing double event, fixing mistyped event, adding new event)

## To test
1. Trigger every event found in the events table found [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14283) (with the generosity of @ParaskP7). Trigger can be found in the Trigger column.
2. Make sure tracking event log in the console includes: `blog_id`, `feed_id` and `follow` properties.

## Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -555,7 +555,7 @@ DEPENDENCIES:
   - WordPressAuthenticator (~> 1.42.0)
   - WordPressKit (~> 4.42.0)
   - WordPressMocks (~> 0.0.15)
-  - WordPressShared (from `../WordPress-iOS-Shared`)
+  - WordPressShared (from `https://github.com/wordpress-mobile/WordPress-iOS-Shared.git`, branch `fix/16961-update-reader-tracking-events`)
   - WordPressUI (~> 1.12.2-beta.1)
   - WPMediaPicker (~> 1.7.2)
   - Yoga (from `https://raw.githubusercontent.com/wordpress-mobile/gutenberg-mobile/v1.62.0/third-party-podspecs/Yoga.podspec.json`)
@@ -711,7 +711,8 @@ EXTERNAL SOURCES:
     :submodules: true
     :tag: v1.62.0
   WordPressShared:
-    :path: "../WordPress-iOS-Shared"
+    :branch: fix/16961-update-reader-tracking-events
+    :git: https://github.com/wordpress-mobile/WordPress-iOS-Shared.git
   Yoga:
     :podspec: https://raw.githubusercontent.com/wordpress-mobile/gutenberg-mobile/v1.62.0/third-party-podspecs/Yoga.podspec.json
 
@@ -727,6 +728,9 @@ CHECKOUT OPTIONS:
     :git: https://github.com/wordpress-mobile/gutenberg-mobile.git
     :submodules: true
     :tag: v1.62.0
+  WordPressShared:
+    :commit: 8f0bcc47c53f24328ba0e66ed6fe0ed2c686bc89
+    :git: https://github.com/wordpress-mobile/WordPress-iOS-Shared.git
 
 SPEC CHECKSUMS:
   1PasswordExtension: f97cc80ae58053c331b2b6dc8843ba7103b33794
@@ -827,6 +831,6 @@ SPEC CHECKSUMS:
   ZendeskSupportSDK: 3a8e508ab1d9dd22dc038df6c694466414e037ba
   ZIPFoundation: e27423c004a5a1410c15933407747374e7c6cb6e
 
-PODFILE CHECKSUM: 39dda0f5fb1a0abc77875bc071864f4f8b82aab3
+PODFILE CHECKSUM: e5974ae80d429b46530e494cd9eb50bc9139d94e
 
 COCOAPODS: 1.10.1
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -555,7 +555,7 @@ DEPENDENCIES:
   - WordPressAuthenticator (~> 1.42.0)
   - WordPressKit (~> 4.42.0)
   - WordPressMocks (~> 0.0.15)
-  - WordPressShared (from `https://github.com/wordpress-mobile/WordPress-iOS-Shared.git`, branch `fix/16961-update-reader-tracking-events`)
+  - WordPressShared (~> 1.16.0)
   - WordPressUI (~> 1.12.2-beta.1)
   - WPMediaPicker (~> 1.7.2)
   - Yoga (from `https://raw.githubusercontent.com/wordpress-mobile/gutenberg-mobile/v1.62.0/third-party-podspecs/Yoga.podspec.json`)
@@ -606,6 +606,7 @@ SPEC REPOS:
     - WordPress-Editor-iOS
     - WordPressKit
     - WordPressMocks
+    - WordPressShared
     - WordPressUI
     - WPMediaPicker
     - wpxmlrpc
@@ -710,9 +711,6 @@ EXTERNAL SOURCES:
     :git: https://github.com/wordpress-mobile/gutenberg-mobile.git
     :submodules: true
     :tag: v1.62.0
-  WordPressShared:
-    :branch: fix/16961-update-reader-tracking-events
-    :git: https://github.com/wordpress-mobile/WordPress-iOS-Shared.git
   Yoga:
     :podspec: https://raw.githubusercontent.com/wordpress-mobile/gutenberg-mobile/v1.62.0/third-party-podspecs/Yoga.podspec.json
 
@@ -728,9 +726,6 @@ CHECKOUT OPTIONS:
     :git: https://github.com/wordpress-mobile/gutenberg-mobile.git
     :submodules: true
     :tag: v1.62.0
-  WordPressShared:
-    :commit: 8f0bcc47c53f24328ba0e66ed6fe0ed2c686bc89
-    :git: https://github.com/wordpress-mobile/WordPress-iOS-Shared.git
 
 SPEC CHECKSUMS:
   1PasswordExtension: f97cc80ae58053c331b2b6dc8843ba7103b33794
@@ -831,6 +826,6 @@ SPEC CHECKSUMS:
   ZendeskSupportSDK: 3a8e508ab1d9dd22dc038df6c694466414e037ba
   ZIPFoundation: e27423c004a5a1410c15933407747374e7c6cb6e
 
-PODFILE CHECKSUM: e5974ae80d429b46530e494cd9eb50bc9139d94e
+PODFILE CHECKSUM: 64119e05ada7945e9dcb2fcbd861a41d6eb4f293
 
 COCOAPODS: 1.10.1
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -11,15 +11,15 @@ PODS:
     - AppAuth/ExternalUserAgent (= 1.4.0)
   - AppAuth/Core (1.4.0)
   - AppAuth/ExternalUserAgent (1.4.0)
-  - AppCenter (4.2.0):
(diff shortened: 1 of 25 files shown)


momo-ozawa:
@nikola-milicevic This PR's milestone is set to 18.4 but the target branch is `epic/16961-update-reader-tracking-properties`. Did you mean to target `develop`?

WordPress/Classes/ViewRelated/Reader/ReaderSaveForLater+Analytics.swift
```diff
@@ -71,14 +76,20 @@ extension ReaderSaveForLaterAction {
 }
 
 extension ReaderStreamViewController {
-    func trackSavedPostNavigation() {
+    func trackSavedPostNavigation(post: ReaderPost) {
+        let blogID = post.siteID ?? 0
+        let feedID = post.feedID ?? 0
+        var properties: [String: Any] = ["blog_id": blogID,
+                                         "feed_id": feedID,
+                                         "foolow": post.isFollowing]
```
momo-ozawa: Typo 👀 
```suggestion
                                         "follow": post.isFollowing]
```

momo-ozawa (commented):
@nikola-milicevic 

~`reader_site_followed` and `reader_site_unfollowed` don't have the `feed_id` included in the event log.~

~According to [this chart](https://github.com/wordpress-mobile/WordPress-Android/pull/14283#issue-835664355), it looks like `feed_id` should be included.~

Nvm -- read your comment!
> All reader tracking events (excluding reader_site_followed / reader_site_unfollowed and reader_reader_tag_followed / reader_reader_tag_unfollowed) now contain: blog_id, feed_id and follow properties.

WordPress/Classes/ViewRelated/Reader/ReaderSaveForLater+Analytics.swift
nikola-milicevic: Oops 😬 I will definitely go for a enum key in a separate PR, but thanks for catching this anyway!

nikola-milicevic:
> @nikola-milicevic This PR's milestone is set to 18.4 but the target branch is `epic/16961-update-reader-tracking-properties`. Did you mean to target `develop`?

@momo-ozawa  Well not sure what's correct here. Here I want to target `epic` definitely, since I will have (at least) two more PR, and then once all is approved / merged into `epic`, I'll merge `epic` into `develop`. How to make the bots happy if I omit the milestone?

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
```diff
@@ -1490,7 +1490,7 @@ + (TracksEventPair *)eventPairForStat:(WPAnalyticsStat)stat
             eventName = @"reader_site_followed";
             break;
         case WPAnalyticsStatReaderSitePreviewed:
-            eventName = @"reader_blog_preview";
+            eventName = @"reader_blog_previewed";
```
leandroalonso, line 1493: Important info: we almost never rename events. The reason is that there's no "merge" in the Analytics tool, so if we do it we end up with two different events and that can possibly affect a lot of calculations, funnels, etc.

In Android the name of the event is `reader_blog_previewed` but AFAIK it has always been like that. It's a shame that both events doesn't have the same name but we should keep `reader_blog_preview`.

leandroalonso, line 1493: Important info: we almost never rename events. The reason is that there's no "merge" in the Analytics tool, so if we do it we end up with two different events and that can possibly affect a lot of calculations, funnels, etc.

In Android the name of the event is `reader_blog_previewed` but AFAIK it has always been like that. It's a shame that both events don't have the same name but we should keep `reader_blog_preview`.

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
```diff
@@ -1506,7 +1506,7 @@ + (TracksEventPair *)eventPairForStat:(WPAnalyticsStat)stat
             eventName = @"reader_tag_loaded";
             break;
         case WPAnalyticsStatReaderTagPreviewed:
-            eventName = @"reader_tag_preview";
+            eventName = @"reader_tag_previewed";
```
leandroalonso: Same as above.

Podfile.lock
```diff
@@ -11,15 +11,15 @@ PODS:
     - AppAuth/ExternalUserAgent (= 1.4.0)
   - AppAuth/Core (1.4.0)
   - AppAuth/ExternalUserAgent (1.4.0)
-  - AppCenter (4.2.0):
-    - AppCenter/Analytics (= 4.2.0)
-    - AppCenter/Crashes (= 4.2.0)
-  - AppCenter/Analytics (4.2.0):
+  - AppCenter (4.3.0):
```
leandroalonso, line 14: Are these `Podfile.lock` changes intentional?

WordPress/Classes/Services/ReaderPostService.h
```diff
@@ -106,6 +106,7 @@ extern NSString * const ReaderPostServiceToggleSiteFollowingState;
  @param failure block called if there is any error. `error` can be any underlying network error.
  */
 - (void)toggleLikedForPost:(ReaderPost *)post
+        fromArticleDetails:(BOOL) details
```
leandroalonso, line 110: Maybe we can rename `details` to `isFromArticleDetails`? Reading the method signature is clear what that means, but in the implementation inside the method `details` is not clear.

leandroalonso (changes requested):
@nikola-milicevic about:

> All reader tracking events (excluding reader_site_followed / reader_site_unfollowed and reader_reader_tag_followed / reader_reader_tag_unfollowed) 

`reader_site_followed` and `reader_site_unfollowed` should have a `feedId` (based on the table from Android's PR).

A few other notes:

* `reader_article_rendered` doesn't seem to exist on iOS. I wonder if it's an event that it's triggered when the article finishes rendering through the webview? Not sure if we should implement that, wdyt?
* `reader_blog_previewed` is duplicated, eg.:

```
2021-09-24 10:39:36:525 WordPress[3506:54946] 🔵 Tracked: reader_blog_previewed <blog_id: 184797733, follow: 0, is_jetpack: 0, post_id: 9569, URL: https://sawidji.com>
2021-09-24 10:39:36:966 WordPress[3506:54946] 🔵 Tracked: reader_blog_previewed <blog_id: 184797733, feed_id: 116888526, follow: 0, site: Sawidji Gallery, source: reader, subscription_count: 8>
```

This is unrelated to this task but it's clearly a bug, we can tackle that on another PR.

leandroalonso:
> Well not sure what's correct here. Here I want to target epic definitely, since I will have (at least) two more PR, and then once all is approved / merged into epic, I'll merge epic into develop. How to make the bots happy if I omit the milestone?

You can add the `Pending` or `Someday` milestones. :)

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
nikola-milicevic: Hmm well then we have `reader_blog_preview` and `reader_blog_previewed` events, since there are two enums defined in the app, called from different places, which are identically named ("-previewed"), but have different values (["preview"](https://github.com/wordpress-mobile/WordPress-iOS/blob/c464da08602a3eab58d3a2987eff72821648147c/WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m#L1492) and ["previewed"](https://github.com/wordpress-mobile/WordPress-iOS/blob/cfdfab82158289015985bd47536a896dc4f40f1a/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift#L289)), so it looked clearly as a typo to me.
Should I revert the change I made?

WordPress/Classes/Services/ReaderPostService.h
nikola-milicevic: Actually I would avoid doing this since I think it is not in line with Objective-C naming conventions. Few examples: [here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html#//apple_ref/doc/uid/20001282-BCIGIJJF), [here](https://google.github.io/styleguide/objcguide.html) and [here](https://stackoverflow.com/questions/15719854/naming-convention-for-local-bool-variables-in-objective-c-and-cocoa). I agree that `details` is not the most clear name, though I would not use `isFromArticleDetails` which is more Swifty. What do you think?

Podfile.lock
nikola-milicevic: I just updated the pods, nothing more. Should I revert this?

leandroalonso: I wouldn't update them if it's not strictly needed for this PR. I know it's just a minor version but we avoid doing so.

Plus, AppCenter is normally handled by the release team.

WordPress/Classes/Services/ReaderPostService.h
leandroalonso: Legibility > conventions IMHO. But this method is not big, so I don't think it's a big problem, you can leave it as it is.

nikola-milicevic:
@leandroalonso 

> `reader_site_followed` and `reader_site_unfollowed` should have a feedId

On it.

> `reader_article_rendered` doesn't seem to exist on iOS.

I'll add it.

> `reader_blog_previewed` is duplicated

Good catch. But how can I reproduce it? Since it didn't happen during my testing.

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
leandroalonso: Ughhhh, you're right. In that case, the ship has sailed, I agree with you, let's leave as it is.

WordPress/Classes/Services/ReaderPostService.h
nikola-milicevic: > Legibility > conventions

I agree, but someone coming from Obj-C might not 😊 I'm leaving it as is for now.

WordPress/Classes/Utility/Analytics/WPAnalyticsTrackerAutomatticTracks.m
nikola-milicevic: I'm reverting this back to `reader_tag_preview`.

leandroalonso:
> Good catch. But how can I reproduce it? Since it didn't happen during my testing.

This was probably fixed in 6fbe41da8eda5ec7a827b8bf5620b2fe02ef98ba, I'll test again once it's ready for another review. :)

nikola-milicevic:
@momo-ozawa @leandroalonso I updated the code, thanks for the comments!

leandroalonso (commented):
@nikola-milicevic all good. The only pending item is to move the new event to the `WPAnalyticsEvent.swift`.

nikola-milicevic:
@leandroalonso Thanks! I update the code. Should be good to go now.

leandroalonso (approved):

PR #17214, opened 2021-09-24
TimeZone selection screen - add contextual labels for GMT Offset timezone, CurrentTime

sla8c:
Timezone selection from site settings in iOS app is outdated in comparison with WordPress web and Android. This PR brings iOS on par with them by adding extra labels into the TableView to show for each WPTimeZone
 * Relative GMT Offset 
 * Current time of each WPTimeZone

Ref #11450

Android [Issue](https://github.com/wordpress-mobile/WordPress-Android/issues/7900) and [PR](https://github.com/wordpress-mobile/WordPress-Android/pull/14198) for reference

### To test:

To test this UI enhancement:

1. Navigate to My Site -> Site Settings -> Time Zone
2. Make sure the time zone that is suggested is consistent with the timezone of the device / simulator you are testing on
3. Check that there's a GMT offset string below the TimeZone primary title and the local time for that TimeZone is shown as a subtitle on the right hand side

### Screenshot of enhancement
image

Additionally I've tested the following with accompanying screenshots:
 * Issue task item "Ensure view(s) are correct on iPhone and iPad" - Portrait and Landscape screenshots
 
image

 
 * Issue task item "Ensure view(s) are correct in both light and dark appearance" - dark appearance screenshot
image

 * Issue task item "Add accessibility support if needed".  2 images:
   * Accessibility with large text and 
   * Accessibility Hint for button
image

image

## Regression Notes
 * TimeZone selector screen was changed - unintended impact could be user would not be able to change their TimeZone
 * TimeZoneSuggestion feature (PR #17145) could no longer work

What I did to test those areas of impact (or what existing automated tests I relied on)
 * I retested all steps in (PR #17145)
   * Navigate to My Site -> Site Settings -> Time Zone
   * Make sure the time zone that is suggested is consistent with the timezone of the device / simulator you are testing on
   * Tap the suggested timezone
   * This will navigate you back to Site Settings screen where you should see the Suggested time you have chosen
   * Tap Time Zone again from Site Settings screen
   * To test search results works normally and has no regression:
   * Search for a timezone (example Casablanca). TableView results should filter and show Casablanca.
   * Tap Casablanca
   * This will navigate you back to Site Settings screen where you should see the Suggested time you have chosen (Casablanca)

No automated tests existed previously and ViewController is not in a state that facilitates automated testing

PR submission checklist:


WordPress/Classes/ViewRelated/Tools/Time Zone/TimeZoneSelectorViewController.swift
```diff
@@ -65,22 +65,79 @@ struct TimeZoneSelectorViewModel: Observable {
         })
     }
 
+    private let timeZoneOffsetFormatter = DateFormatter()
+
+    private let timeAtTimeZoneFormatter = DateFormatter()
+
+    private let currentDate = Date()
```
Gio2018: Splitting the initialization between these lines and the initializer makes the code less clear. Also, since the creation of `DateFormatter` is indeed expensive, why not put the instantiation of them behind the feature flag as well?
How can we change this code to be more clear, and, potentially, also more testable?

WordPress/Classes/ViewRelated/Tools/Time Zone/TimeZoneSelectorViewController.swift
```diff
@@ -65,22 +65,79 @@ struct TimeZoneSelectorViewModel: Observable {
...
+
+    // DateFormatter is expensive https://sarunw.com/posts/how-expensive-is-dateformatter/
+    // initializer to ensure that:
+    // 1. Existing 3 property struct initialization of state, selectedValue, filter work normally
+    // 2. DateFormatters are configured whenever we create instance of TimeZoneSelectorViewModel
+    // 3. DateFormatters are private and their state not mutable from outside callers (thread-safety)
```
Gio2018: I don't think these comments should be here, and structured like this, for the following reasons:
1. this is the place where normally methods documentation should go (and, if so, it should be preceded by /// and not //)
2. DateFormatter is expensive: that's commonly known, no need to add a specific link to a web page that could change at any time
3. "DateFormatters are private and their state not mutable from outside callers (thread-safety)": properties that do not need to be accessed externally should be private regardless, and also, from the [documentation](https://developer.apple.com/documentation/foundation/nsdateformatter): "On iOS 7 and later NSDateFormatter is thread safe"

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneTableViewCell.swift
```diff
@@ -0,0 +1,89 @@
...
+        ])
+
+        contentView.addSubview(stackView)
+        contentView.pinSubviewToAllEdges(stackView, insets: UIEdgeInsets(
+                top: Constants.topPadding,
+                left: Constants.leadingTrailingPadding,
+                bottom: Constants.bottomPadding,
+                right: Constants.leadingTrailingPadding)
+        )
+    }
```
Gio2018: This method is quite long: any chance we can break it up in different methods, for improved legibility?

Gio2018 (changes requested):
Hey @sla8c, I tested and it worked as expected; left some comments in the code for you to consider.

WordPress/Classes/ViewRelated/Tools/Time Zone/TimeZoneSelectorViewController.swift
sla8c: That's a good idea and I've taken your suggestion and run with it by encapsulating that logic in `TimeZoneFormatter` and adding accompanying unit tests. Thanks!

WordPress/Classes/ViewRelated/Tools/Time Zone/TimeZoneSelectorViewController.swift
sla8c: No problem; I thought it was better to explain than not.  Removed!

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneTableViewCell.swift
sla8c: No problem; I've broken up the constraints layout into a new method

WordPress/Classes/ViewRelated/Tools/Time Zone/TimeZoneSelectorViewController.swift
```diff
@@ -65,6 +65,18 @@ struct TimeZoneSelectorViewModel: Observable {
...
     }
 
+    private var timeZoneFormatter: TimeZoneFormatter?
+
+    init(state: State, selectedValue: String?, filter: String?) {
+        self.state = state
+        self.selectedValue = selectedValue
+        self.filter = filter
+
+        if FeatureFlag.timeZoneSuggester.enabled {
```
Gio2018: This works well, thanks for the patience @sla8c !
I was thinking that, we might initialize `TimeZoneFormatter` anyway, and then use the feature flag only to decide what type of row we want to return. This so we don't over-complicate the code with unnecessary `guard let timeZoneFormatter = timeZoneFormatter else`; also, since we are working on a feature branch, if we spot performance issues of some sort, we can still fix them before we merge into `develop`

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneTableViewCell.swift
```diff
@@ -0,0 +1,93 @@
+import UIKit
+
+class TimeZoneTableViewCell: WPTableViewCell {
+
+    let titleLabel = UILabel()
+
+    let leftSubtitle = UILabel()
+
+    let rightSubtitle = UILabel()
```
Gio2018: would it work if we instantiate and setup these UI elements with a closure directly here? It could make the code more legible

Gio2018 (commented):
Hey @sla8c , looking pretty good overall! Left a couple more comments for you to consider

WordPress/Classes/ViewRelated/Tools/Time Zone/TimeZoneSelectorViewController.swift
sla8c: Sounds good @Gio2018  - that's done!

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneTableViewCell.swift
sla8c: Sounds good - that's done as well!

WordPress/Classes/ViewRelated/Tools/Time Zone/Views/TimeZoneTableViewCell.swift
```diff
@@ -0,0 +1,132 @@
...
+        if label.effectiveUserInterfaceLayoutDirection == .leftToRight {
+            // swiftlint:disable:next inverse_text_alignment
+            label.textAlignment = .right
+        } else {
+            // swiftlint:disable:next natural_text_alignment
+            label.textAlignment = .left
+        }
+
+        return label
+    }()
```
Gio2018, line 62: These look good!! One more optional comment here: you don't have to do it, but you could also `lazy var`. This could also help with some code optimization calling a method in there

Gio2018 (approved):

PR #17219, opened 2021-09-26
Fix/16961 - Events properties logic

nikola-milicevic:
Part of [#16961](https://github.com/wordpress-mobile/WordPress-iOS/issues/16961)

This PR replaces the logic for generating reader tracking properties (that is located in many places) with single point for calculation in `ReaderHelpers` class, more specifically `statsProperties(for post: mergeWith existing:)` method. Also, enums are used for properties keys, and only properties with meaningful values are added.

## To test
1. Trigger every event found in the events table found [here](https://github.com/wordpress-mobile/WordPress-Android/pull/14283). Trigger can be found in the Trigger column.
2. Make sure tracking event log in the console includes: `blog_id`, `feed_id` and `follow` properties.

## Regression Notes
n/a

n/a

n/a

PR submission checklist:


WordPress/Classes/Services/NotificationActionsService.swift
```diff
@@ -16,7 +16,7 @@ class NotificationActionsService: LocalCoreDataService {
             return
         }
 
-        siteService.followSite(withID: siteID, site: nil) {
+        siteService.followSite(withID: siteID, topic: nil, post: nil) {
```
leandroalonso: The `topic` and `post` represent a post and/or topic in which the user is following this site?

Maybe a `fromTopic` or `fromPost` would be clear?

WordPress/Classes/Services/ReaderSiteService.h
```diff
@@ -28,25 +30,29 @@ extern NSString * const ReaderSiteServiceErrorDomain;
  Follow a wpcom site by ID.
 
  @param siteID The ID of the site.
- @param site Site that's being followed. Optional parameter.
+ @param topic Topic that's being followed. Optional parameter.
+ @param post Post that's being followed. Optional parameter.
```
leandroalonso: Is this a method to follow site, topics, or posts? I guess it's only sites but the `topic` and `post` params say otherwise.

leandroalonso (commented):
@nikola-milicevic the build is failing for me, as for the code I left some minor suggestions.

WordPress/Classes/ViewRelated/Reader/ReaderHelpers.swift
```diff
@@ -218,26 +218,62 @@ struct ReaderPostMenuButtonTitles {
...
+            feedID = Int(truncating: post.feedID)
+            isFollowing = post.following
+            isJetpack = post.isJetpack
+        } else {
+            return existing ?? [:]
         }
 
-        if let value = value, let key = key {
-            properties[key] = value
+        if blogID != 0 {
```
leandroalonso: I discussed about that briefly with @ParaskP7 and we had prior discussions in which we agreed to add the properties even if they're `0`.

The reason is that having this field makes calculations easier, so you probably don't need to check for the presence or absence of this value and just include it. :)

nikola-milicevic:
> @nikola-milicevic the build is failing for me, as for the code I left some minor suggestions.

@leandroalonso build issues should be fixed now, I updated the code, thanks for suggestions!

WordPress/Classes/Services/ReaderPostService.m
```diff
@@ -321,9 +316,9 @@ - (void)setFollowing:(BOOL)following
 
     ReaderSiteService *siteService = [[ReaderSiteService alloc] initWithManagedObjectContext:self.managedObjectContext];
     if (following) {
-        [siteService followSiteWithID:[siteID integerValue] site:NULL success:successBlock failure:failureBlock];
+        [siteService followSiteWithID:[siteID integerValue] fromTopic:NULL fromPost:NULL success:successBlock failure:failureBlock];
```
momo-ozawa: [Q] Is there a reason why you're using `NULL` instead of `nil`? I thought `NULL` was for C pointers. (https://nshipster.com/nil/)

momo-ozawa (commented):
Builds for me without issues 👍 
Left 1 comment.

WordPress/Classes/Services/ReaderPostService.m
nikola-milicevic: Except that it is not technically incorrect, and the fact that it brings me back in time 😊 no particular reason. `nil` is specific to Objective-C objects, so I agree it is more suitable here. I'll update the code.

momo-ozawa (approved):

PR #17221, opened 2021-09-27
Convert some of the UI tests screen objects from `BaseScreen` to `ScreenObject` – Part 1 of many

mokagio:
This is the first PR of a series that will convert all `BaseScreen` subclasses into `ScreenObject` subclasses.

It might not be super clear by looking at the diff, so here's the structure of a simple `ScreenObject`:

```swift
import ScreenObject // ScreenObject comes from a Swift package, so we need to import it
import XCTest

public class FeaturedImageScreen: ScreenObject {

    // expectedElement comes from the superclass and gets the first expectedElementGetters result
    var removeButton: XCUIElement { expectedElement }

    // The subclass (this one, FeaturedImageScreen) wraps the superclass (ScreenObject)
	// init method so that we can call it with no parameters. All the defaults are set in
    // the context of this file and callers don't need to know about them.
    public init(app: XCUIApplication = XCUIApplication()) throws {
        try super.init(
			// In this simple screen, we only check for one element to verify wether the
            // screen is loaded.
            //
  			// This here is a closure and its type, as defined in the ScreenObject init
			// method signature is (XCUIApplication) -> XCUIElement. That means it
			// takes a XCUIApplication as input and returns an XCUIElement. The $0 is a
            // shorthand to represent the first (0 index) argument passed to the closure.
            //
            // Internally, ScreenObject uses the app instance we pass to in in the next line
            // to call this closure and get the expected element.
            expectedElementGetters: [ { $0.navigationBars.buttons["Remove Featured Image"] } ],
            app: app
        )
    }

	// ScreenObject subclasses in UITestFoundation can expose many methods to interact with
    // the screen they represent.
    public func tapRemoveFeaturedImageButton() {
        removeButton.tap()
        // app is the same value passed at init
        app.sheets.buttons.element(boundBy: 0).tap()
    }
}
```

Most screens use more than one element to verify if they're loaded correctly and for the test interactions. In those cases, we use a two step approach.

First, we define the getter closure as an instance `let`

```swift
let getDoneButton: (XCUIApplication) -> XCUIElement = {
    $0.navigationBars.buttons["doneButton"]
}
```

Then we pass the value in the `expectedElementGetters` array

```swift
init(app: XCUIApplication = XCUIApplication()) throws {
    try super.init(
        expectedElementGetters: [ getDoneButton, getViewButton, ... ],
        app: app
    )
}
```

Finally, we access the element themselves by calling the closure. We don't use `expectedElement` because that only returns the first element.

```swift
var doneButton: XCUIElement { getDoneButton(app) }
var viewButton: XCUIElement { getViewButton(app) }
```    

That's it. I wish I could have found a cleaner way to do this, or one that required less explanation and relied more on the type system. This is the best I could do so far. 😕

---

Notice that I run the UI tests and the all passed.

## Regression Notes
N.A.

N.A.

N.A.

---

### DIFF
--- a/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": null,
-          "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
-          "version": "0.1.0"
+          "revision": "2d65beae47cdcaaf21703ce1b321f382fe0d0730",
+          "version": "0.2.0"
         }
       },
       {
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -1,3 +1,4 @@
+import ScreenObject
 import UIKit
 import UITestsFoundation
 import XCTest
@@ -85,6 +86,19 @@ class JetpackScreenshotGeneration: XCTestCase {
 }
 
 extension BaseScreen {
+    @discardableResult
+    func thenTakeScreenshot(_ index: Int, named title: String) -> Self {
+        let mode = XCUIDevice.inDarkMode ? "dark" : "light"
+        let filename = "\(index)-\(mode)-\(title)"
+
+        snapshot(filename)
+
+        return self
+    }
+}
+
+extension ScreenObject {
+
     @discardableResult
     func thenTakeScreenshot(_ index: Int, named title: String) -> Self {
         let mode = XCUIDevice.inDarkMode ? "dark" : "light"
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -32,7 +32,7 @@ class JetpackScreenshotGeneration: XCTestCase {
         super.tearDown()
     }
 
-    func testGenerateScreenshots() {
+    func testGenerateScreenshots() throws {
 
         // Get My Site screenshot
         let mySite = MySiteScreen()
@@ -41,8 +41,8 @@ class JetpackScreenshotGeneration: XCTestCase {
             .thenTakeScreenshot(1, named: "MySite")
 
         // Get Activity Log screenshot
-        let activityLog = mySite
-            .gotoActivityLog()
+        let activityLog = try mySite
+            .goToActivityLog()
             .thenTakeScreenshot(2, named: "ActivityLog")
 
         if !XCUIDevice.isPad {
--- a/WordPress/UITestsFoundation/Screens/ActivityLogScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/ActivityLogScreen.swift
@@ -1,8 +1,9 @@
+import ScreenObject
 import XCTest
 
-public class ActivityLogScreen: BaseScreen {
+public class ActivityLogScreen: ScreenObject {
 
-    public init() {
-        super.init(element: XCUIApplication().otherElements.firstMatch)
+    public init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(expectedElementGetters: [ { $0.otherElements.firstMatch } ])
     }
 }
--- a/WordPress/UITestsFoundation/Screens/Editor/EditorPostSettings.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/EditorPostSettings.swift
@@ -44,7 +44,8 @@ public class EditorPostSettings: ScreenObject {
 
     public func removeFeatureImage() throws -> EditorPostSettings {
         changeFeaturedImageButton.tap()
-        FeaturedImageScreen()
+
+        try FeaturedImageScreen()
             .tapRemoveFeaturedImageButton()
 
         return try EditorPostSettings()
--- a/WordPress/UITestsFoundation/Screens/Editor/EditorPostSettings.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/EditorPostSettings.swift
@@ -1,22 +1,21 @@
+import ScreenObject
 import XCTest
 
-public class EditorPostSettings: BaseScreen {
-    let settingsTable: XCUIElement
-    let categoriesSection: XCUIElement
-    let tagsSection: XCUIElement
-    let featuredImageButton: XCUIElement
-    var changeFeaturedImageButton: XCUIElement {
-        return settingsTable.cells["CurrentFeaturedImage"]
-    }
+public class EditorPostSettings: ScreenObject {
 
-    init() {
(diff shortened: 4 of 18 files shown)


WordPress.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": null,
-          "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
-          "version": "0.1.0"
+          "revision": "2d65beae47cdcaaf21703ce1b321f382fe0d0730",
+          "version": "0.2.0"
```
mokagio, line 10: Updated `ScreenObject` to the latest version to take advantage of the multiple-elements init method.

WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
```diff
@@ -95,3 +96,16 @@ extension BaseScreen {
...
+    @discardableResult
+    func thenTakeScreenshot(_ index: Int, named title: String) -> Self {
+        let mode = XCUIDevice.inDarkMode ? "dark" : "light"
+        let filename = "\(index)-\(mode)-\(title)"
+
+        snapshot(filename)
+
+        return self
+    }
+}
```
mokagio, line 111: This is a 1:1 copy of the extension on `BaseScreen` above and is a way to combine Fastlane's screenshot logic with a nice DSL in `ScreenObject`.

Eventually, we might want to add some kind of abstraction and make this part of the framework. For the moment, though, the focus is to convert all the screens, so this will do.

WordPress/UITestsFoundation/Screens/ActivityLogScreen.swift
```diff
@@ -1,8 +1,9 @@
+import ScreenObject
 import XCTest
 
-public class ActivityLogScreen: BaseScreen {
+public class ActivityLogScreen: ScreenObject {
 
-    public init() {
-        super.init(element: XCUIApplication().otherElements.firstMatch)
+    public init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(expectedElementGetters: [ { $0.otherElements.firstMatch } ])
```
mokagio, line 7: Worth pointing out that this matches is pretty useless because it will succeed most of the time.

The aim of the current work is not to fix "issues" in the tests, but to move to `ScreenObject` to make it easier to fix those issues later.

WordPress/UITestsFoundation/Screens/Editor/EditorPostSettings.swift
```diff
@@ -74,15 +73,15 @@ public class EditorPostSettings: BaseScreen {
             XCTAssertTrue(imageCount == 0, "Featured image is set but should not be")
         }
 
-        return EditorPostSettings()
+        return try EditorPostSettings()
     }
 
-    // returns void since return screen depends on which editor you're in
+    /// - Note: Returns `Void` because the return screen depends on which editor the user is in.
     public func closePostSettings() {
```
mokagio, line 81: We want the UITestsFoundation framework to be super easy to use for feature developers to write tests with. This little change in how the documentation is formatted means Xcode will show this information during autocompletion or if a user Alt-clicks on the method name.

Screen Shot 2021-09-28 at 3 19 06 pm

WordPress/UITestsFoundation/Screens/Editor/EditorPostSettings.swift
```diff
@@ -74,15 +73,15 @@ public class EditorPostSettings: BaseScreen {
...
         navBackButton.tap()
     }
 
     public static func isLoaded() -> Bool {
-        return XCUIApplication().tables["SettingsTable"].exists
+        return (try? EditorPostSettings().isLoaded) ?? false
```
mokagio, line 86: Rather than duplicating the logic to verify if a certain element exists, generating code that risks getting out of date, we can leverage the `ScreenObject` `init` behavior, which oden't fail the tests automatically if the screen is not loaded.

WordPress/UITestsFoundation/Screens/Editor/EditorPublishEpilogueScreen.swift
```diff
@@ -1,26 +1,42 @@
...
+    var doneButton: XCUIElement { Self.getDoneButton(app) }
+    var viewButton: XCUIElement { Self.getViewButton(app) }
 
-        super.init(element: published)
+    public init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(
+            expectedElementGetters: [
+                Self.getDoneButton,
+                Self.getPublishedLabel,
+                Self.getViewButton
```
mokagio: TODO: Convert these to be instance properties!

WordPress/UITestsFoundation/Screens/Editor/FeaturedImageScreen.swift
```diff
@@ -1,13 +1,15 @@
+import ScreenObject
 import XCTest
 
-public class FeaturedImageScreen: BaseScreen {
+public class FeaturedImageScreen: ScreenObject {
 
-    let removeButton: XCUIElement
+    var removeButton: XCUIElement { expectedElement }
```
mokagio, line 7: TODO: add a code comment about this behavior here

WordPress/UITestsFoundation/Screens/ActivityLogScreen.swift
mokagio: I opened https://github.com/wordpress-mobile/WordPress-iOS/issues/17231 in this regard.

WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
```diff
@@ -31,7 +32,7 @@ class JetpackScreenshotGeneration: XCTestCase {
         super.tearDown()
     }
 
-    func testGenerateScreenshots() {
+    func testGenerateScreenshots() throws {
```
mokagio, line 35: Because the `ScreenObject` `init` `throws`, we now have a sprinkle of `try` calls around the test suite.

That's good. XCTest knows how to fail tests and report errors if a `try` call `throws`.

WordPress/UITestsFoundation/Screens/ActivityLogScreen.swift
pachlava: > Worth pointing out that this matches is pretty useless because it will succeed most of the time.

Indeed 😄

pachlava (approved):
> so here's the structure of a simple ScreenObject:

I wish I started the review from this PR, instead of woocommerce/woocommerce-ios/pull/5091 (which I did because it had 6 changed files less 😅 ). The overview that you provide in the PR description is very helpful! What do you think about making it a part of [README.md](https://github.com/Automattic/ScreenObject/blob/trunk/README.md) from ScreenObject?

I executed the test locally, with no fails 👍

mokagio:
> What do you think about making it a part of README.md from ScreenObject?

That's a great idea, thanks! I don't have time to tackle it just yet and I'd like the code to be used by a few more people to discover rough edges first. I opened https://github.com/Automattic/ScreenObject/issues/10 to remember this.

PR #17234, opened 2021-09-28
Sync comments content only if needed

### LINKED ISSUE #17202: Comments: Paging resets after returning from the comment detail screen
### Expected behavior
The Comments list should preserve the loaded comments.

### Actual behavior
The Comments list got refreshed, and returned to page one. All the loaded contents (from page 2 and so on) are removed.

- Run the app on any device/simulator.
- Select a site that has many comments, such that it has two or more pages when scrolling down the comments list.
- Go to My Site > Comments.
- Scroll down until the fetch request for more content is triggered.
- Select any comment on page two.
- Tap Back.
- Observe that the Comments list is refreshed and the paging is reset.

ealeksandrov:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/17202.

## Description

On each `viewWillAppear` method `syncContent` is called to reload all data, even if there is something already loaded.

The fix is adding a check and call `syncContent` only when it's needed (no loaded content).

## To test:

1. Run the app on an any device/simulator.
2. Select a site that has many comments, such that it has two or more pages when scrolling down the comments list.
3. Tap My Site > Comments.
4. Scroll down until the fetch request for more content is triggered.
5. Select any comment on page two.
6. Tap Back.
7. Check that the Comments list is still at same page and scroll position doesn't jump.

---
PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Comments/CommentsViewController.m
@@ -566,7 +566,7 @@ static NSString *RestorableFilterIndexKey = @"restorableFilterIndexKey";
 
 - (void)refreshAndSyncIfNeeded
 {
-    if (self.blog) {
+    if (self.blog && self.contentIsEmpty) {
         [self.syncHelper syncContent];
     }
 }


dvdchr (approved):
LGTM! :shipit:

#### Tested:
- [x] Scrolled to page two or more on the comments list, and tapped on one Comment. When going back to the list, the paging should be preserved.

#### Additional test scenarios:
- [x] Comments list content snippet should reflect the updated comment data.
- [x] Comments list should correctly remove any comment that's marked as spam or trash from the list.

ealeksandrov:
Thanks for revew David!

PR #17242, opened 2021-09-29
Comment Moderation Bar: move style settings to style guide extension

ScoutHarris:
Ref: #17200

This change does two things:
- Moves style related settings to `WPStyleGuide+CommentDetail`.
- Updates colors for dark mode. 

To test:
- Enable the `newCommentDetail` feature.
- Run the app in light and dark modes.
- Verify all the colors and images match #17200 or in Zeplin.

Examples:
<kbd>light</kbd>

<kbd>dark</kbd>

## Regression Notes
1. Potential unintended areas of impact
N/A. Feature is incomplete and disabled.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N/A. Feature is incomplete and disabled.

3. What automated tests I added (or what prevented me from doing so)
N/A. Feature is incomplete and disabled.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17244, opened 2021-09-29
Comment Moderation Bar: set button for comment's initial status

ScoutHarris:
Ref: #17200
Depends on: #17242

When Comment details is shown, the moderation bar button is selected that corresponds with the Comment status.

To test:
- Go to My Site > Comments.
- Select a comment in each of these filters.
  - Pending
  - Approved
  - Spam
  - Trashed
- Verify the correct moderation bar button is selected.

Example:
| Filtered List | Moderation Button |
|--------|-------|
| filter | button |

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Comment+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Comment+CoreDataClass.swift
@@ -154,4 +154,21 @@ extension Comment: PostContentProvider {
             return "draft"
         }
     }
+
+    static func typeForStatus(_ status: String) -> CommentStatusType? {
+        switch status {
+        case "hold":
+            return .pending
+        case "approve":
+            return .approved
+        case "trash":
+            return .unapproved
+        case "spam":
+            return .spam
+        case "draft":
+            return .draft
+        default:
+            return nil
+        }
+    }
 }
--- a/WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
@@ -6,6 +6,12 @@ class CommentModerationBar: UIView {
 
     // MARK: - Properties
 
+    var comment: Comment? {
+        didSet {
+            setButtonForStatus()
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
```diff
@@ -112,6 +118,26 @@ private extension CommentModerationBar {
...
+
+        switch commentStatusType {
+        case .pending:
+            pendingTapped(pendingButton)
+        case .approved:
+            approvedTapped(approvedButton)
+        case .unapproved:
+            trashTapped(trashButton)
+        case .spam:
+            spamTapped(spamButton)
```
dvdchr: one minor question that I should've raised in the previous PRs 😓 : I wonder if in the `pendingTapped` method and friends we could directly target the corresponding `UIButton` instead? Since I assume it's always guaranteed that `pendingTapped` is always associated with `pendingButton` and so on.

This is all so we could remove the need of parameters for these methods, i.e. `pendingTapped()`, `approvedTapped()`, and so on.

WordPress/Classes/Models/Comment+CoreDataClass.swift
```diff
@@ -154,4 +154,21 @@ extension Comment: PostContentProvider {
             return "draft"
         }
     }
+
+    static func typeForStatus(_ status: String) -> CommentStatusType? {
```
dvdchr, line 158: Nicee. 👍

dvdchr (approved):
Hey @ScoutHarris 👋 . Left a minor comment that should've been raised on the previous PRs 😅. But the rest looks good to me! 🚢 

I just have one clarifying question. I noticed that multiple buttons can be active/selected. Will this be addressed later in a future PR?

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
ScoutHarris: Ah, yea. Nice catch. And done!

PR #17246, opened 2021-09-30
Comment Detail: Modify tap action and title when the comment is a reply

dvdchr:
Refs #17087

This PR differentiates (1) the text displayed on the header row of the new Comment Detail and (2) the action performed when the header row is tapped, based on whether the comment is a reply to another comment (i.e. it has `parentID > 0`), or it's a comment to a post. 

## To test
Make sure the `New Comment Detail` feature flag is turned on.

### Scenario 1: Comment is a reply to another comment

- Go to My Site > Comments.
- Select a comment that's a reply to another comment.
- Verify that the header row shows `Reply to <parent comment's author>`.
- Verify that below the "Reply to..." text, the content snippet of the parent comment is displayed.
- Tap the header row and verify that the comment thread is displayed correctly, highlighting the parent comment.

Extra: Some content snippet contains `\n` characters at the start. Verify that the header row doesn't display blank lines due to the newline character at the start of the content snippet.

### Scenario 2: Comment is a direct comment to a post

- Go to My Site > Comments.
- Select a comment that's a direct comment to a post.
- Verify that the header row shows `Comment on`.
- Verify that below the "Comment on" text, the post title is displayed.
- Tap the header row and verify that the post is displayed correctly.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
@@ -60,7 +60,7 @@ class CommentDetailViewController: UITableViewController {
     }()
 
     private lazy var parentComment: Comment? = {
-        guard comment.hasParentComment,
+        guard comment.hasParentComment(),
               let blog = comment.blog,
               let parentComment = commentService.findComment(withID: NSNumber(value: comment.parentID), in: blog) else {
                   return nil
@@ -142,7 +142,7 @@ class CommentDetailViewController: UITableViewController {
 
         switch rows[indexPath.row] {
         case .header:
-            comment.hasParentComment ? navigateToParentComment() : navigateToPost()
+            comment.hasParentComment() ? navigateToParentComment() : navigateToPost()
 
         case .replyIndicator:
             // TODO: Navigate to the comment reply.
@@ -353,12 +353,6 @@ private extension CommentDetailViewController {
         openWebView(for: authorURL)
     }
 
-}
-
-private extension Comment {
-    var hasParentComment: Bool {
-        return parentID > 0
-    }
 }
 
 // MARK: - Strings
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -48,6 +48,27 @@ class CommentDetailViewController: UITableViewController {
         return cell
     }()
 
+    private lazy var commentService: CommentService = {
+        return .init()
+    }()
```
ScoutHarris: `CommentService` should have an `NSManagedObjectContext`. Since `CommentService` is now used multiple times, maybe this init can use the MOC so it doesn't have to be done elsewhere?


```
private lazy var commentService: CommentService = {
    CommentService(managedObjectContext: ContextManager.shared.mainContext)
}()
```

And then these lines can be removed from `updateComment()`:
```
let context = ContextManager.sharedInstance().mainContext
let commentService = CommentService(managedObjectContext: context)
```

ScoutHarris (approved):
Hey @dvdchr . I had one suggestion. Otherwise, looks good!

:shipit:

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
dvdchr: Whoops, good catch there! I just assumed that creating the service with an empty init would _just_ work. Apparently, it did, because the method that I'm using didn't touch the MOC at all 😅 . But it'll cause trouble when updating the comment since the method did access the `managedObjectContext` property.

This is fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17246/commits/9950653e19a2e03780646076dd521cd064718b98!

dvdchr:
Thanks @ScoutHarris !

PR #17252, opened 2021-10-01
Fix page sharing issues

diegoreymendez:
Fixes an issue with sharing pages to WordPress using the share extension.

Ref: p5T066-2G5-p2#comment-10093

## To test:

1. Start sharing something to WordPress through the Share Extension.
2. Select "Page" as the "Type".
3. Once you hit publish, make sure to quickly send the App you're sharing from to the background.
4. After a bit, open WordPress and make sure the content was shared as a Page.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/PostService.m
+++ b/WordPress/Classes/Services/PostService.m
@@ -128,9 +128,15 @@ const NSUInteger PostServiceDefaultNumberToSync = 40;
                           }
                           if (remotePost) {
                               AbstractPost *post = [self findPostWithID:postID inBlog:blog];
+                              
                               if (!post) {
-                                  post = [self createPostForBlog:blog];
+                                  if ([remotePost.type isEqualToString:PostServiceTypePage]) {
+                                      post = [self createPageForBlog:blog];
+                                  } else {
+                                      post = [self createPostForBlog:blog];
+                                  }
                               }
+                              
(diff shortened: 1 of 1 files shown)


diegoreymendez:
Fyi @startuptester and @renanferrari.

Thanks for the report!

WordPress/Classes/Services/PostService.m
```diff
@@ -128,9 +128,15 @@ - (void)getPostWithID:(NSNumber *)postID
                           }
                           if (remotePost) {
                               AbstractPost *post = [self findPostWithID:postID inBlog:blog];
+                              
                               if (!post) {
-                                  post = [self createPostForBlog:blog];
+                                  if ([remotePost.type isEqualToString:PostServiceTypePage]) {
```
diegoreymendez, line 133: Just a note that there's an identical snippet of code in this class, so we could place this logic inside a method.

That said, because this is a fix for 18.3 which is soon going out, and because I don't want to risk breaking any existing logic, I'm purposedly only changing this part of the code right now.

I'll push another PR unifying both snippets into a single method against `develop`.

frosty: Sounds good, I agree with your thinking!

frosty (approved):
Works for me! :shipit:

diegoreymendez:
Thanks @frosty !

PR #17255, opened 2021-10-01
Comment Detail: Navigation bar styling

dvdchr:
Refs #17087

This PR adds some style ✨ to the navigation bar, according to the designs in Zeplin. I noticed that the look feels similar to Apple's Messages app – where the navigation bar is opaque when positioned on top/scroll edge, but provides visual context for scrolled content behind the navigation bar through a visual blur effect. Here are some previews:

• | Light 🌝 | Dark 🌚 
--|--|--
Top | light_top | dark_top
Scrolled | light_scrolled | dark_scrolled

Additionally, extracted out the assignment of the edit bar button item to its own method, `configureEditButtonItem()` since user permission may change and affect the ability of the user to edit the comment. Moving this to its own method allows it to be called in `refreshData`, so the edit button is hidden/shown correctly based on the newest state.

## To Test

Ensure that the `New Comment Detail` is enabled.

- Go to My Site > Comments.
- Select any comment, preferably one that contains a lot of text (so the view is scrollable). 
- Verify that the navigation bar is opaque when in the top position.
- Scroll the view.
- Verify that the navigation bar shows a visual blur effect.

Ensure that ☝️ behaves as expected in light and dark mode.

## Regression Notes
n/a, the feature is still hidden behind a feature flag.

n/a, the feature is still hidden behind a feature flag.

n/a, the feature is still hidden behind a feature flag.

PR submission checklist:


ScoutHarris:
In an iOS 14.5 simulator, the nav bar has a blur when not scrolled. It looks fine in iOS 15.


| iOS 14.5 | iOS 15.0 |
|--------|-------|
| 14sim | 15sim |

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -181,9 +181,22 @@ private extension CommentDetailViewController {
     }
 
     func configureNavigationBar() {
-        if comment.canModerate {
-            navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .edit, target: self, action: #selector(editButtonTapped))
-        }
+        // apply a Messages-like visual blur effect when content is scrolled.
```
ScoutHarris: Nitpick: the `Messages-like` description will be misleading when Messages changes again. Can we just remove that bit?

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -181,9 +181,22 @@ private extension CommentDetailViewController {
...
+        self.navigationController?.navigationBar.isTranslucent = true
+        self.navigationItem.compactAppearance = appearance
+        self.navigationItem.standardAppearance = appearance
+
+        configureEditButtonItem()
+    }
+
+    func configureEditButtonItem() {
+        let editButtonItem = comment.canModerate ? UIBarButtonItem(barButtonSystemItem: .edit, target: self, action: #selector(editButtonTapped)) : nil
+        navigationItem.rightBarButtonItem = editButtonItem
```
ScoutHarris: Nitpick: This could just be one line.
`navigationItem.rightBarButtonItem = comment.canModerate ? UIBarButtonItem(barButtonSystemItem: .edit, target: self, action: #selector(editButtonTapped)) : nil`

ScoutHarris (changes requested):

dvdchr:
Aw shucks. I knew it! I've been stuck trying to download the iOS 14.5 Simulator. I'll give it another look on Monday. 🙂

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17255/commits/7b5afc4ec613834804af270b21ac95b11f560d98. Thanks!

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
dvdchr: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17255/commits/ffc524b622932fb3972cc20b0158697dac0a89f7. Thank you

dvdchr:
Hey @ScoutHarris! 👋  It's ready for your second round of review. Since the behavior of `scrollEdgeAppearance` is different for iOS 14[^1], I had to listen to scroll events (i.e. `scrollViewDidScroll` and adjust the `UINavigationBarAppearance` accordingly. I've tested this and it works in iOS 13, 14, and 15 now. 🙂 

Oh, also I've updated the visual blur effect style to be `systemThinMaterial` from `systemUltraThinMaterial`, to slightly improve contrast for the navigation bar button items (the back & edit button). Somehow the visual blur looks more "see-through" in dark mode 😅 .

⚠️ NOTE: I'm turning on auto-merge!

[^1]: "When running on apps that use iOS 14 or earlier, this property applies to navigation bars with large titles. In iOS 15, this property applies to all navigation bars" – [source](https://developer.apple.com/documentation/uikit/uinavigationbar/3198027-scrolledgeappearance)

ScoutHarris:
A couple more oddities for ya.

- On iOS15, the nav bar changes color immediately on scroll. But on iOS14, it doesn't change color until the header and part of the first cell is scrolled off.
- Probably related to the first point, but on an iOS14 iPad in dark mode:
  - The nav bar starts out as slightly lighter than black. 
  - Start to scroll. The nav bar turns black.
  - Scroll more. The nav bar is then blurred, which is lighter than the initial color. 
  - The result is it looks like it's flashing light > dark > light.

ScoutHarris (changes requested):
See latest comment.

ScoutHarris:
> NOTE: I'm turning on auto-merge!

Thanks for the warning! 😄

dvdchr:
> On iOS15, the navbar changes color immediately on scroll. But on iOS14, it doesn't change color until the header and part of the first cell are scrolled off.

Hey @ScoutHarris! Thank you for finding these visual bugs. Appreciate it! 🙂 So it's all because the navigation bar is translucent, the content view offset is actually "pulled" to the top, behind the navigation bar and status bar. So instead of checking if the `contentOffset.y` is greater than `0`, the correct value to check is actually `-64.0` – which happens to be the total height of the navbar and the status bar.

...Of course, I didn't hardcode the number! It's dynamically pulled via a custom `topBarHeight` computed property. And with this, it should be ready for your (hopefully) final round of review! 😄

ScoutHarris (approved):
winner

:shipit:

PR #17260, opened 2021-10-04
Comment Moderation Bar: set button states when one is selected

ScoutHarris:
Ref: #17200

When a button on the moderation bar is selected:
- The previously selected button is reset.
- Only one button can be selected at a time.
- Re-selecting the active button does nothing.

To test:
- Go to My Site > Comments.
- Select a comment in each of the filters  `Pending`, `Approved`, `Spam`, `Trashed`.
  - Verify the correct moderation bar button is selected.
- Tap other buttons.
  - Verify the tapped buttons are selected, and the other buttons are not.
  - Verify tapping a selected button does nothing.

https://user-images.githubusercontent.com/1816888/135928290-a5d48992-e7e2-47ec-85cc-a91d548fdd3f.mp4

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
@@ -41,7 +41,7 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
     @IBOutlet private weak var likeButton: UIButton!
 
     // This is public so its delegate can be set directly.
-    @IBOutlet weak var moderationBar: CommentModerationBar!
+    @IBOutlet private(set) weak var moderationBar: CommentModerationBar!
 
     // MARK: Private Properties
 
--- a/WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
@@ -40,7 +40,8 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
     @IBOutlet private weak var replyButton: UIButton!
     @IBOutlet private weak var likeButton: UIButton!
 
-    @IBOutlet private weak var moderationBar: CommentModerationBar!
+    // This is public so its delegate can be set directly.
+    @IBOutlet weak var moderationBar: CommentModerationBar!
 
     // MARK: Private Properties
 
--- a/WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
@@ -10,12 +10,6 @@ class CommentModerationBar: UIView {
 
     // MARK: - Properties
 
-    var comment: Comment? {
-        didSet {
-            setButtonForStatus()
-        }
-    }
-
     @IBOutlet private weak var contentView: UIView!
 
     @IBOutlet private weak var pendingButton: UIButton!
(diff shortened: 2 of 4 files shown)


ScoutHarris:
Hey @dvdchr .  You'll probably notice that the intent will be for `CommentDetailViewController` to handle comment moderation. 

I looked at how Comments and Notifications moderate comments. There really isn't anything common to extract into a shared location. They do different things (messaging, view handling, etc), and they call different `CommentService` methods. So instead I've added a protocol to `CommentModerationBar`. `CommentDetailViewController` will contain methods for handling Comment moderation in its own way. And in the future, `NotificationDetailsViewController` can adopt the protocol and just call its existing comment moderation methods.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -40,7 +40,8 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
     @IBOutlet private weak var replyButton: UIButton!
     @IBOutlet private weak var likeButton: UIButton!
 
-    @IBOutlet private weak var moderationBar: CommentModerationBar!
+    // This is public so its delegate can be set directly.
+    @IBOutlet weak var moderationBar: CommentModerationBar!
```
dvdchr: Should we make this `readonly` for public access, since the caller only needs to assign the `delegate` property?

```suggestion
    @IBOutlet private(set) weak var moderationBar: CommentModerationBar!
```

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
```diff
@@ -30,6 +28,23 @@ class CommentModerationBar: UIView {
     private let iPadPaddingMultiplier: CGFloat = 0.33
     private let iPhonePaddingMultiplier: CGFloat = 0.15
 
+    weak var delegate: CommentModerationBarDelegate?
+
+    var comment: Comment? {
+        didSet {
+            currentStatus = CommentStatusType.typeForStatus(comment?.status)
+        }
+    }
```
dvdchr, line 37: Do we need the Comment object to be passed into this cell? Reading your [comment on this line](https://github.com/wordpress-mobile/WordPress-iOS/pull/17260/files#diff-068cb0f97e51ef29d1181c901df1795395c7808a8302cd41d847fab39a8585cdR212) got me thinking if we could just make the `currentStatus` property public instead, to prevent the Comment object from being mistakenly modified in the future 😄 . What do you think?

dvdchr (approved):
Left some minor comments, and I just noticed something: should we disable multi-touch for the moderation bar view?

```Swift
isExclusiveTouch = true
```

While the state of the moderation bar is still behaving as expected (no two buttons active at the same time), I'd imagine the delegate would still fire two or more network requests in response to the multi-touch.

Anyways, tested the rest and it looks good to me! :shipit:

WordPress/Classes/ViewRelated/Comments/CommentModerationBar.swift
ScoutHarris: 🤔 you're probably right. I passed the Comment object long before realizing it shouldn't be changed here. I'll look at removing it when I add the actual moderation calls, just to make sure.

ScoutHarris:
> I'd imagine the delegate would still fire two or more network requests in response to the multi-touch.

Maybe I'm missing something, but I don't see how this could happen. Each `Tapped` method has a check to prevent that:
`guard !approvedButton.isSelected else {`

I don't see that `isExclusiveTouch` makes a difference, but I'll add it anyway.

Issue #17215, opened 2021-09-24
Empty Stats: Update the Grow Audience card presentation logic


momo-ozawa:
## Description

Update the Grow Audience card presentation logic.

- Make the card dismissable per site (currently dismissing the card will dismiss it for _all_ sites)
- Add a view count threshold for the Grow Audience card display logic.
  - If the view count is less than 30, the Grow Audience card should be displayed.
  - If the view count is greater than or equal to 30, the Grow Audience card should be hidden and the Customize card should be displayed.

Design ref: pbArwn-2Js-p2

Slack ref: p1632153265040000-slack-C02CX9QQYSF

## Tasks
- [x] Make the Grow Audience card site specific [#17223](https://github.com/wordpress-mobile/WordPress-iOS/pull/17223)
- [x] Show the Grow Audience card if view count < 30, hide if view count >= 30 [#17261](https://github.com/wordpress-mobile/WordPress-iOS/pull/17261)

PR #17261, opened 2021-10-04
Empty Stats: Add view count threshold for Grow Audience card display logic

nikola-milicevic:
Part of [#17215](https://github.com/wordpress-mobile/WordPress-iOS/issues/17215)

This PR adds a logic for displaying Grow Audience card if the view count is less than 30.

## To test
1. Delete the app and reinstall it (to delete the cache).
2. Log in
3. Go to **My Site** > **Stats**
4. Make sure the **Grow Audience** card is displayed only if the view count is less than 30
5. Go back, and go to stats again (for the same site)
6. Double check the displayed state of the card is the same as before.

### Note
Because of the data loading mechanism, by checking the newly fetched site, going back, and checking it again, we ensure the consistency between displaying sites from cache and the newly fetched ones.

## Regression Notes
n/a

n/a

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -184,6 +184,7 @@ private extension SiteStatsInsightsTableViewController {
         }
 
         insightsChangeReceipt = viewModel?.onChange { [weak self] in
+            self?.refreshGrowAudienceCardIfNecessary()
             self?.displayEmptyViewIfNecessary()
             self?.refreshTableView()
         }
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -146,8 +146,9 @@ class SiteStatsInsightsTableViewController: UITableViewController, StoryboardLoa
         WPStyleGuide.Stats.configureTable(tableView)
         refreshControl?.addTarget(self, action: #selector(refreshData), for: .valueChanged)
         ImmuTable.registerRows(tableRowTypes(), tableView: tableView)
-        loadInsightsFromUserDefaults()
         initViewModel()
+        loadInsightsFromUserDefaults()
+        updateView()
         tableView.estimatedRowHeight = 500
 
         displayEmptyViewIfNecessary()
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
```diff
@@ -351,6 +359,25 @@ private extension SiteStatsInsightsTableViewController {
         permanentlyDismissInsight(.growAudience, using: key)
     }
 
+    var shouldDisplayGrowAudienceCard: Bool {
+        let threshold = 30
+        let insightsStore = StoreContainer.shared.statsInsights
```
momo-ozawa: What do you think of making `insightsStore` a private let property? We can also inject this `insightsStore` into the `SiteStatsInsightsViewModel`.

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
```diff
@@ -351,6 +359,25 @@ private extension SiteStatsInsightsTableViewController {
         permanentlyDismissInsight(.growAudience, using: key)
     }
 
+    var shouldDisplayGrowAudienceCard: Bool {
+        let threshold = 30
```
momo-ozawa, line 367: Should we have a constant for threshold?

momo-ozawa (commented):
Works as described. 

I wanted to test what happens when go from 29 views to 30 views, then pull down to refresh. However, it's taking a bit of time for the all-time views to get updated server side... so I'll check back in tomorrow.

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
nikola-milicevic: Agreed 😊

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
nikola-milicevic: I think it's better to be explicit (even if something seems obvious now), because that might change in the future and no guesswork about magic numbers is needed.

nikola-milicevic:
@momo-ozawa Thanks for the comments, I updated the code!

momo-ozawa (approved):
Confirmed that the Grow Audience card changes to the Customize card after crossing the 30 view threshold 😄

PR #17266, opened 2021-10-05
Makes the domain searches immediate.

### LINKED ISSUE #17240: Make the domain search trigger immediately.
Site creation has it already.

diegoreymendez:
Fixes #17240 

Notes:

- Live domain searches are debounced with a minimum delay of 0.5 secs.
- The search-in-progress HUD is now shown on top of the results only.  It looked kind of odd when the device keyboard was shown.

## Associated PRs:

WordPress Shared: https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/300

## To test:

1. Go into the Domain Registration flow.
2. In the Domain Suggestions screen start typing.

Make sure searches are triggered 0.5 seconds after the user stops typing.  Make sure the whole search feature works as expected.

## Regression Notes

I've slightly modified `Debouncer.swift` which is used across the app for different purposes.

Outside of that, the edited components are only used for the new Domain Registration flow, so they should not affect existing behaviour.

This class has existing unit tests to verify it's previous behaviour, and those tests are still succeeding.

I've added a test to document and verify the new behaviour that I've added to this class.

PR submission checklist:


Gio2018 (approved):
Worked as expected!

frosty (changes requested):
Changes work as expected, however the HUD gets lower down the screen each time a search is performed:

domains-3

diegoreymendez:
@frosty - Ready for another look.

WordPress/Classes/ViewRelated/NUX/DomainSuggestionsTableViewController.swift
```diff
@@ -133,13 +134,17 @@ class DomainSuggestionsTableViewController: UITableViewController {
             self?.isSearching = false
             self?.noSuggestions = false
             SVProgressHUD.dismiss()
```
frosty, line 136: Nitpick, but we could extract a few of these things into a method we call from within the success and failure blocks (dismissing the hud, removing the background view, ending refreshing.

frosty (approved):
Works for me! One minor suggestion, both otherwise :shipit:

WordPress/Classes/ViewRelated/NUX/DomainSuggestionsTableViewController.swift
diegoreymendez: I'll submit another PR since it turned out to have a lot of space for improvement.

diegoreymendez: New PR with the proposed improvements here: https://github.com/wordpress-mobile/WordPress-iOS/pull/17272

frosty: 👍

PR #17272, opened 2021-10-06
Makes some improvements to DomainSuggestionsTableViewController.

### LINKED ISSUE #17206: Enable the privacy protection flag in the cart, for any domain that supports it

diegoreymendez:
Fixes #17206

Make some improvements to `DomainSuggestionsTableViewController`.  This was created after [some suggestions](https://github.com/wordpress-mobile/WordPress-iOS/pull/17266#discussion_r723380409) by @frosty in a previous review.

This PR also sets the privacy protection on by default, for domains that support it.

## To test:

1. Go to the Domain Purchase flow.
2. Use the Domain Search regularly and make sure the suggestions come up as expected.
3. Go to My Site -> Domains and add a new domain
    a. Search for a domain that does not support privacy
    b. Proceed with the registration and make sure it succeeds
    c. Repeat the step listed in 3, and this time select a domain that support privacy
    d. Proceed with the registration and make sure it succeeds
    e. OPTIONALLY: in both cases, if you have access to the DARC, go there and make sure that the privacy protection is set accordingly
     

For the second test I recommend to insert these lines in `DomainsService.swift:53`, so that the suggestions service will always fail:

```swift
failure(NSError(domain: "testDomain", code: 1, userInfo: [:]))
return
```

1. Go to the Domain Purchase flow.
2. Make sure the no-results view is shown.

## Regression Notes

None as the domains flow is new.

## PR submission checklist:


Gio2018 (changes requested):
Hey @diegoreymendez , tested the three cases and here is the outcome I got:

Test 1: Worked as expected

Test 2: I added the code you suggested in `DomainService`: not sure if I'm doing something wrong, I tested a few times, and all the times I get the no results view, but after random short time, that view disappears and search results re-appear (see video)

<p align=center>

</p>

Test 3: it generally worked. What I noticed in the DARC, with the very few TLDs that don't support privacy protection that I could have access to, is that the domain has `supportsPrivacy` set to `false`, but after the registration the DARC says something like "Privacy protection enabled for free with the TLD". I think this is ok, I just was not able to test a TLD that actually did not have privacy protection at alll

diegoreymendez:
> Test 1: Worked as expected

Cool! 👍 

> Test 2: I added the code you suggested in `DomainService`: not sure if I'm doing something wrong, I tested a few times, and all the times I get the no results view, but after random short time, that view disappears and search results re-appear (see video)

No you're not - it's actually my bad!  I didn't realize that the return statement would still execute the service call.

Instead of `return` try using `return ()`.  I've updated the PR description above to reflect this.
 
> Test 3: it generally worked. What I noticed in the DARC, with the very few TLDs that don't support privacy protection that I could have access to, is that the domain has `supportsPrivacy` set to `false`, but after the registration the DARC says something like "Privacy protection enabled for free with the TLD". I think this is ok, I just was not able to test a TLD that actually did not have privacy protection at alll

Since I didn't commit the code related to the privacy flag, I'm not aware of the details behind its behaviour.

I think if the change is really simple and has no risk it can be acceptable but if there's a chance those changes may cause trouble it may be best to post them separately.

Gio2018 (approved):
:shipit:

diegoreymendez:
Thanks @Gio2018 !

PR #17277, opened 2021-10-07
Temporarily remove aggregate target for credentials codegen in Jetpack

mokagio:
We noticed that Google Login was broken — resulting in the app to crash – in Jetpack.

My diagnosis is that was because the codegen script receiving a `BUILD_SCHEME = WordPress` value, instead of `Jetpack`. The reason, I think, is that the aggregate target has it's own set of build settings and doesn't use the Jetpack ones, despite the build being set with the
aggregate target as a dependency for Jetpack.

To resolve this in the short term, I have rolled back to running the codegen script as a Build Phase Run Script as part of the Jetpack target, removing the dependency on the aggregate target.

There might be better ways to do this in the long run, but for the purpose of fixing this in the code freeze branch or to ship a hotfix, this was the fastest thing I could come up with.

To test: verify that the Jetpack app crashes when trying to Google login on `develop`. Switch to this branch and verify it succeeds.

@yaelirub @leandroalonso let me know whether you think this should be a hotfix or it's fine to put it in the current release. I defaulted to current release to avoid starting an unnecessary hotfix workflow, but we might want to do a hotfix (or maybe not? this issue has possibly been out for a while and we haven't heard of it from users yet).

## Regression Notes
N.A.

See above

None :/

PR submission checklist:

### DIFF
--- a/WordPress/WordPress.xcodeproj/project.pbxproj
+++ b/WordPress/WordPress.xcodeproj/project.pbxproj
@@ -4491,13 +4491,6 @@
 			remoteGlobalIDString = 096A92F526E29FFF00448C68;
 			remoteInfo = GenerateCredentials;
 		};
-		096A92FE26E2A0C000448C68 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = 29B97313FDCFA39411CA2CEA /* Project object */;
-			proxyType = 1;
-			remoteGlobalIDString = 096A92F526E29FFF00448C68;
-			remoteInfo = GenerateCredentials;
-		};
 		3F526C5A2538CF2B0069706C /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = 29B97313FDCFA39411CA2CEA /* Project object */;
@@ -14959,6 +14952,7 @@
 			buildConfigurationList = FABB264D2602FC2C00C8785C /* Build configuration list for PBXNativeTarget "Jetpack" */;
 			buildPhases = (
 				FABB1FA72602FC2C00C8785C /* [CP] Check Pods Manifest.lock */,
+				3F32E4AE270EAF5100A33D51 /* Generate Credentials */,
 				FABB1FA92602FC2C00C8785C /* App Icons: Add Version For Internal Releases */,
 				FABB1FAA2602FC2C00C8785C /* Resources */,
 				FABB20C12602FC2C00C8785C /* Copy Alternate Internal Icons (if needed) */,
@@ -14972,7 +14966,6 @@
 			buildRules = (
 			);
 			dependencies = (
-				096A92FF26E2A0C000448C68 /* PBXTargetDependency */,
 				FABB1F902602FC2C00C8785C /* PBXTargetDependency */,
 			);
 			name = Jetpack;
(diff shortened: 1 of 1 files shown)


yaelirub:
>I defaulted to current release to avoid starting an unnecessary hotfix workflow, but we might want to do a hotfix (or maybe not? this issue has possibly been out for a while and we haven't heard of it from users yet).

We don't have many users in this app, its super hard to find the app in the store and those who tried with Google login probably abandoned so I wouldn't expect to hear from users on something like this. 
I do want us to maintain the same standard of quality that we have in other places so I vote for a hotfix.

mokagio:
I'll start the hot fix workflow soon, then 👍

momo-ozawa (approved):
@mokagio (@leandroalonso)
Works as described! Thank you! 🙇‍♀️

AliSoftware (approved):
If I'm reading the diff properly, you only reverted (from "a dependency to the aggregate" to "build phase") for the Jetpack target, while the WP target still uses the aggregate one, right?

Quite unfortunate to have this behavior and bug indeed; I think the workaround is OK for now (even though I hope we get to find a better solution for this at some point…)

mokagio:
> If I'm reading the diff properly, you only reverted (from "a dependency to the aggregate" to "build phase") for the Jetpack target, while the WP target still uses the aggregate one, right?

Correct

WordPress | Jetpack
--- | ---
image | image

> Quite unfortunate to have this behavior and bug indeed; I think the workaround is OK for now (even though I hope we get to find a better solution for this at some point…)

+1. In the [R.swift discussion](https://github.com/mac-cain13/R.swift/issues/719#issuecomment-936079241) related to the recent [Xcode 13 issue I experienced in the WooCommerce project](https://github.com/woocommerce/woocommerce-ios/pull/5049), some suggest to Apple recommending _Build Rules_ as opposed to _Build Phases_ for code generation. Maybe that's an avenue to explore. Or maybe Xcode has other build environment variables we can reach for to see what's the downstream target that builds a dependency one. 📖 🔍

AliSoftware:
I knew about the possibility to use _Pre-Build Actions_ in the scheme – but I don't like them for multiple reasons, first because it feels hidden but more importantly because if they fail they do so silently and don't prevent the build – but I never considered _Build Rules_ for such use case, that's an interesting approach.

I watched [this part of WWDC'21 video](https://developer.apple.com/videos/play/wwdc2021/10210/?time=379) (as referenced in the other discussion) but my thought is that it seems to mostly apply only if you need to "convert" input files into output files (like the recipe files in the Fruta example), rather than just produce files "out of thin air" or by copying example files or files from secrets, like we do.
So I'm not sure we'd be able to use a Build Rule to generate our credentials in that case, as we wouldn't have a pattern for the "source files to process"…?

AliSoftware:
Also in that WWDC video they don't really say that "Build Phases are not a good option to generate code".
In fact:
 - They [use a Build Phase to run their `package.sh` script which combines multiple input files into a single output file](https://developer.apple.com/videos/play/wwdc2021/10210/?time=616) (something a Build Rule can't do)
 - They suggest to use an Aggregate Target (just like we did) at [14:07](https://developer.apple.com/videos/play/wwdc2021/10210/?time=847)

The issue with the Aggregate Target, at least with the way it's presented in the video, is that its goal is to mutualize actions that would otherwise be run multiple times, once per target. In their example they use an AT because their SmoothieKit frameworks is built twice (once per target) but they want their `package.sh` script to only run once for both architecture.

So in practice it's actually a feature of Aggregate Target to only run once. Exactly like in a Makefile: if WordPress depends on AT and Jetpack depends on AT, but AT has already been run (and its input files haven't been not modified since last run, and its output files still exist) during the build for WP, it won't run again during a subsequent build of WP… or JP. Unless we do a clean build of course.

That being said, the "Based on dependency analysis" checkbox (which I think is quite recent, maybe new in Xcode 12?) is supposed to change that behavior (if you uncheck the box, the help text under it says "Will force script to run in all incremental builds", ie regardless of input, context or output files), so I'd have hoped that unchecking this box would change that whole behavior talked about in the video and make it run for JP even if it had already been run for WP… seems that's not the case, which I wonder should be considered an Xcode bug or a misunderstanding from our part…

PR #17284, opened 2021-10-08
Add tip to the top of domain search for free sites

### LINKED ISSUE #17191: Add the redirection tip at the top of the search page

frosty:
Implements #17191. This PR adds a tip to the top of the domains search screen for free sites:

The text will shrink as needed to a maximum of 3 lines, at which point it will truncate. I tested with very long domains and this seems like a reasonable limit. We need to set a line limit otherwise the text won't shrink, and it can wrap oddly if we don't do that.

The tip should only appear for free sites with a redirect.

**To test**

* Build and run
* Navigate to Domains search for a free site. You should see the tip.
* Navigate to Domains search for a site with a premium plan and an unused domain credit. You should not see the tip.

## Regression Notes

None

N/A

PR submission checklist:


WordPress/Classes/ViewRelated/Domains/Domain registration/RegisterDomainSuggestions/DomainSuggestionsTableViewController.swift
```diff
@@ -262,6 +269,39 @@ extension DomainSuggestionsTableViewController {
...
+            textLabel.adjustsFontSizeToFitWidth = true
+            textLabel.minimumScaleFactor = 0.5
+        }
+
+        let template = NSLocalizedString("Domains purchased on this site will redirect to %@", comment: "Description for the first domain purchased with a free plan.")
+        let formatted = String(format: template, freeSiteAddress)
+        let attributed = NSMutableAttributedString(string: formatted, attributes: [:])
+
+        if let range = formatted.range(of: freeSiteAddress) {
+            attributed.addAttributes([.font: cell.textLabel!.font.bold()], range: NSRange(range, in: formatted))
```
Gio2018: I know this is basically safe, but since we are unwrapping `textLabel` above, why not take advantage of it and use the unwrapped value instead of force unwrapping `textLabel` here? It looks like we could even use a `guard` at the beginning of the method too

Gio2018 (approved):
This worked for me on iPhone 12 Pro Max device and iPad Pro 2nd gen device. Left one code comment.

WordPress/Classes/ViewRelated/Domains/Domain registration/RegisterDomainSuggestions/DomainSuggestionsTableViewController.swift
diegoreymendez: Agree on both counts.

By looking at the rest of the code in this method it looks like it only makes sense if `cell.textLabel` isn't `nil`.

diegoreymendez (approved):
Works as described.

WordPress/Classes/ViewRelated/Domains/Domain registration/RegisterDomainSuggestions/DomainSuggestionsTableViewController.swift
frosty: Whoops, thanks for catching that one – wasn't intentional!

PR #17291, opened 2021-10-08
Use bash cache artifact support

jkmassel:
Fixes slow artifact downloads by implementing our own layer for them.

**To test**
- Verify that the CI runs faster.

**Note:**
This PR will require cutting a new version of https://github.com/Automattic/bash-cache-buildkite-plugin after it's approved, but before it lands

### DIFF
--- a/.buildkite/commands/build-for-testing.sh
+++ b/.buildkite/commands/build-for-testing.sh
@@ -17,4 +17,4 @@ bundle exec fastlane build_for_testing
 
 echo "--- :arrow_up: Upload Build Products"
 tar -cf build-products.tar DerivedData/Build/Products/
-buildkite-agent artifact upload build-products.tar
+upload_artifact build-products.tar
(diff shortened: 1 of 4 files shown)


guarani (approved):
👋 Thanks @jkmassel! The `buildkite/wordpress-ios` job passed in 32m 30s which looks good!
I checked to see how long `Setting up Pods` was taking, but I don't see it anymore under the Build step here. I take it that's because it's getting cached and that's the fix here in [this companion PR](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/7), but I just wanted to be sure.

It certainly looks like this PR fixes the slow builds, but I wonder why the slow builds were not consistently slow before. For example, I noticed a recent WPiOS PR took only 35m 26s (see build [here](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/2927#49385e8a-e485-45c9-940b-6fc7d76ac462)), so it looks like even without the fix, builds occasionally run fast.

jkmassel:
Yeah, so the issue is actually [here](https://github.com/buildkite/agent/blob/4d96c145cf6fff43f924f618cd70062f4b3eb7eb/agent/s3_downloader.go#L59-L64) – this approach is a great trick, but it also means that the download speed can be wildly variable outside of AWS. The new approach should have far more consistent timings, because it's serving the resources via AWS' internal network

PR #17305, opened 2021-10-13
Update keywords for English locales

mokagio:
Updates the English keywords as suggested by @jrushfinn.

Nothing to test. Once the automation to submit version 18.4 will run, these will be uploaded to Ap Store Connect.

### DIFF
--- a/fastlane/metadata/en-AU/keywords.txt
+++ b/fastlane/metadata/en-AU/keywords.txt
@@ -1,1 +1,1 @@
-social,network,notes,jetpack,photos,writing,geotagging,media,blog,wordpress,website,blogging,design
+writing,blog,blogging,blogger,website,web,design,maker,online,store,business,make,site,build,create
--- a/fastlane/metadata/en-CA/keywords.txt
+++ b/fastlane/metadata/en-CA/keywords.txt
@@ -1,1 +1,1 @@
-social,notes,jetpack,writing,geotagging,media,blog,website,blogging,journal
+writing,blog,blogging,blogger,website,web,design,maker,online,store,business,make,site,build,create
--- a/fastlane/metadata/en-GB/keywords.txt
+++ b/fastlane/metadata/en-GB/keywords.txt
@@ -1,1 +1,1 @@
-social,notes,jetpack,writing,geotagging,media,blog,website,blogging,journal
+writing,blog,blogging,blogger,website,web,design,maker,online,store,business,make,site,build,create
--- a/fastlane/metadata/en-US/keywords.txt
+++ b/fastlane/metadata/en-US/keywords.txt
@@ -1,1 +1,1 @@
-social,notes,jetpack,writing,geotagging,media,blog,website,blogging,journal
+writing,blog,blogging,blogger,website,web,design,maker,online,store,business,make,site,build,create


mokagio (commented):
I do wonder if our automation will override these by downloading "new" versions from GlotPress 🤔 

I'll have to keep an eye out for it during submission.

AliSoftware:
> I do wonder if our automation will override these by downloading "new" versions from GlotPress 🤔 
> 
> 
> 
> I'll have to keep an eye out for it during submission.

I am pretty sure they will.

You should update the copies on GlotPress directly instead.

AliSoftware (commented):
@mokagio 

1. Is the goal to update the originals — and then have translators translate the new keywords originals in GlotPress and have these changes in the keywords applied to all locales ultimately? (If so, which I think is what would make sense, then you need to change the `.po` file)
2. Or is the goal to not change the originals (which serve as the basis for translations) and thus keep the old keywords in other locales, but instead only update the keywords for English locales and not the others? In that case you need to make the changes in GlotPress as it will otherwise be overwritten during release.

Personally I'd go with (1), and optionally once the PR is merged and the new `.po` imported in GlotPress, provide the translations for `en-AU`, `en-GB` and `en-CA` in GlotPress yourself (since you already know those as per your `txt` changes here). Then let translators take care of the other locales.

mokagio:
Thanks @AliSoftware.

The end goal is to have those locales use the new keywords in the App Store listing. I don't know what's the plan for the other locales.

I've updated GlotPress manually as you suggested, but I'm not sure if that's the solution we want to adopt in the long run.

From what I understand, we want to do more experiment and fine tuning of the keywords, starting with English.

I think in the long run the kind of flow we have now, where the English list is the source of truth on which all other locales base their translations won't work anymore. I mean, it would still work, but someone looking at the keywords in GlotPress without knowing the experimentation intent might be tempted to update the translations 🤔 At that point, it might be best to keep the source of truth closer to the codebase, namely with the localized `.txt` files and stop downloading localized translations.

mokagio:
Closing with the rationale described in https://github.com/wordpress-mobile/WordPress-iOS/pull/17327#issue-758942249.

But fear not, **ASC does have these new keywords**:

image

AliSoftware:
I agree that if, for locale-aware ASO purposes, the keywords are gonna end up being different to optimize for each locale, like I actually suggested to Jacob in pcbrnV-3zU-p2#comment-4656, we might want to stop making them go thru GlotPress, and instead just have the `fastlane/metadata/<locale>/keywords.txt` files be the new source of truth.

That being said, since the current focus is only on English locales and to my knowledge we never really tried to optimize for other locales just yet (it's the next step, but up until now the keywords for other locales have always been the verbatim translation as the ones in English since the beginning of the app existence), I'm not sure there will really be an impact in having the original list (aka English) be updated, uploaded in GlotPress, and still verbatim-translated into other locales.

Sure we weren't asked to change the keywords in the other locales because we haven't started the ASO study for those other locales yet, but we haven't been asked not to change them either, I mean at that point it's not like we had proof that the old keywords in other locales are gonna perform better or worse in those locales than the new one, as we've only started the study for English. Once we start saying "ok now that we've locked the English keywords to optimize for English locales, let's start the ASO study for other locales" then yeah we'll need to disconnect them from GlotPress to not be verbatim-translated anymore.
But until then, I feel like it won't be a big deal to continue verbatim-translate the English ones even during that experiment, so updating the originals (in the `.po`) should still be a good idea, otherwise as you said we risk having an English translator see that the English translation (new keywords) doesn't match the original (old keywords) and risk updating them. And we can unplug the `keywords` from GlotPress and remove them from the `.po` once we start the other-locales-ASO study later.

---

btw, I ended up finding the internal reference for the original request so X-posting it here for posterity: pcbrnV-3HK-p2

Issue #17307, opened 2021-10-13
Empty Stats: Event tracking for nudges


nikola-milicevic:
## Description

All nudges should have appropriate tracking events.
Reference: p1633948738005600-slack-C02CX9QQYSF

## Tasks
- Add event tracking for all the nudges
  - [x] Publicize
  - [x] Blogging reminders
  - [x] Reader discover [#17352](https://github.com/wordpress-mobile/WordPress-iOS/pull/17352)
- Validate events (PCYsg-ktp-p2)
  - [x] Publicize
  - [x] Blogging reminders
  - [x] Reader discover

PR #17313, opened 2021-10-13
Empty Stats: Add event tracking for Publicize / Blogging Reminders

momo-ozawa:
Part of #17307

## Description

This PR adds event tracking for the following events:
 
```
// Stats - Empty Stats nudges

stats_publicize_nudge_shown
stats_publicize_nudge_tapped
stats_publicize_nudge_dismissed
stats_publicize_nudge_completed
stats_blogging_reminders_nudge_shown
stats_blogging_reminders_nudge_tapped
stats_blogging_reminders_nudge_dismissed
stats_blogging_reminders_nudge_completed

// Stats - Customize card

stats_customize_insights_shown
```

## To test

To review this PR, please do the following:
1. Make sure events fire correctly by following the trigger in the table below
2. Validate the events (PCYsg-ktp-p2)

> What does it mean for an event to be valid?
> 
> An event can be marked valid when you’ve checked that it:
> 
>     Fires when it’s supposed to.
>     Fires exactly once each time it’s supposed to fire.
>     Has a blogid set if the event takes place in a site context.  If it’s not in a site context (e.g. a landing page or the wpcom homepage), you can omit a blogid.
>     Has the event properties that you intend it to have, and that are being attached to the event by your code.
> 
> Who should review it?
> 
> Similarly to code reviews, someone on your team or working in the same area as you is probably best to review your event.

## Event table

Event | Trigger | blog_id?
-- | -- | --
stats_publicize_nudge_shown | Publicize card is shown upon visiting the Stats Insights screen | ✅ 
stats_publicize_nudge_tapped  | Tapping on the `Enable post sharing` button on the Publicize card | ✅ 
stats_publicize_nudge_dismissed | Tapping on the `Dismiss` button on the Publicize card | ✅ 
stats_publicize_nudge_completed | Successfully setting up at least one connection via Publicize | ✅ 
stats_blogging_reminders_nudge_shown | Blogging Reminders card is shown upon visiting the Stats Insights screen | ✅ 
stats_blogging_reminders_nudge_tapped | Tapping on the `Set up blogging reminders` button on the Blogging Reminders card | ✅ 
stats_blogging_reminders_nudge_dismissed | Tapping on the `Dismiss` button on the Blogging Reminders card | ✅ 
stats_blogging_reminders_nudge_completed | Successfully setting up Blogging Reminders via the Blogging Reminders flow | ✅ 
stats_customize_insights_shown | Customize Insights card is shown upon visiting the Stats Insights screen | ❌ 

## Regression Notes
N/A

Manually tested and made sure events fire properly with the correct triggers / properties

N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -565,6 +565,8 @@ extension SiteStatsInsightsTableViewController: BloggingRemindersFlowDelegate {
     func didSetUpBloggingReminders() {
         viewModel?.markEmptyStatsNudgeAsCompleted()
         refreshTableView()
+
+        trackNudgeEvent(.statsBloggingRemindersNudgeCompleted)
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -211,7 +211,7 @@ private extension SiteStatsInsightsTableViewController {
     }
 
     func loadPinnedCards() {
-        let viewsCount = insightsStore.getAllTimeStats()?.viewsCount
+        let viewsCount = insightsStore.getAllTimeStats()?.viewsCount ?? 0
         switch pinnedItemStore?.itemToDisplay(for: viewsCount) {
         case .none:
             insightsToShow = insightsToShow.filter { $0 != .growAudience || $0 != .customize }
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -466,6 +466,8 @@ extension SiteStatsInsightsTableViewController: SiteStatsInsightsDelegate {
         present(navigationController, animated: true)
 
         applyTableUpdates()
+
+        trackNudgeEvent(.statsPublicizeNudgeTapped)
     }
 
     func growAudienceBloggingRemindersButtonTapped() {
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -622,3 +622,17 @@ extension SiteStatsInsightsTableViewController: NoResultsViewHost {
         static let manageInsightsButtonTitle = NSLocalizedString("Add stats card", comment: "Button title displayed when the user has removed all Insights from display.")
     }
 }
+
+// MARK: - Tracks Support
+
+private extension SiteStatsInsightsTableViewController {
+
+    func trackNudgeEvent(_ event: WPAnalyticsEvent) {
+        if let blogId = SiteStatsInformation.sharedInstance.siteID,
+           let blog = Blog.lookup(withID: blogId, in: mainContext) {
+            WPAnalytics.track(event, properties: [:], blog: blog)
+        } else {
+            WPAnalytics.track(event)
+        }
+    }
+}
--- a/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
+++ b/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
@@ -554,6 +554,8 @@ extension SiteStatsInsightsTableViewController: SharingViewControllerDelegate {
     func didChangePublicizeServices() {
         viewModel?.markEmptyStatsNudgeAsCompleted()
         refreshTableView()
+
+        trackNudgeEvent(.statsPublicizeNudgeCompleted)
     }
 }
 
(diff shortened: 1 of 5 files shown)


momo-ozawa:
## Notes
- I'll add the `nudge_dismissed` events after Nikola's fix presentation logic PR lands. The plan is to track the events in the individual dismiss nudge methods for `SiteStatsInsightsDelegate` -> ✅

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
```diff
@@ -211,18 +211,20 @@ private extension SiteStatsInsightsTableViewController {
...
     func loadPinnedCards() {
-        let viewsCount = insightsStore.getAllTimeStats()?.viewsCount ?? 0
+        let viewsCount = insightsStore.getAllTimeStats()?.viewsCount
         switch pinnedItemStore?.itemToDisplay(for: viewsCount) {
         case .none:
             insightsToShow = insightsToShow.filter { $0 != .growAudience || $0 != .customize }
         case .some(let item):
             switch item {
-            case is GrowAudienceCell.HintType where !insightsToShow.contains(.growAudience):
+            case let hintType as GrowAudienceCell.HintType where !insightsToShow.contains(.growAudience):
```
nikola-milicevic, line 220: 👌

momo-ozawa:
Reverted b4fb243 because it was causing the presentation logic to break.
Added a workaround in c74d496 to make sure the `shown` events are fired only once.

nikola-milicevic (commented):
Great work!
Since you've already cleaned up some of my code 😄 could you just please get rid of nil coalescing operator [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/b4fb243fe419db5cf8b771d11b14d2799123b244/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift#L245)?
And `case let` syntax would be [great](https://github.com/wordpress-mobile/WordPress-iOS/blob/b4fb243fe419db5cf8b771d11b14d2799123b244/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsPinnedItemStore.swift#L58) here as well! (so I don't have to do [this](https://github.com/wordpress-mobile/WordPress-iOS/blob/b4fb243fe419db5cf8b771d11b14d2799123b244/WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsPinnedItemStore.swift#L59))

I have just two concerns:
- `stats_blogging_reminders_nudge_shown` wasn't triggered for me 😬, I will check again (few times)
- `didChangePublicizeServices` method of `SharingViewControllerDelegate` could fire even if there are no changes (but that's not the scope of this PR - just pointing out here)

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
```diff
@@ -211,18 +211,30 @@ private extension SiteStatsInsightsTableViewController {
...
+                // Work around to make sure nudge shown is tracked only once
+                if viewsCount != nil {
+                    trackNudgeShown(for: hintType)
+                }
+
             case InsightType.customize where !insightsToShow.contains(.customize):
                 insightsToShow = insightsToShow.filter { $0 != .growAudience }
                 insightsToShow.insert(.customize, at: 0)
+
+                // Work around to make sure customize insights shown is tracked only once
```
leandroalonso, line 233: Is this to prevent the event from being fired multiple times?

if that's the case, this is something we should not be worried about, given that normally we look at "unique" events.

leandroalonso (changes requested):
@momo-ozawa it worked just fine for all the described events, except for the blogging reminders nudge appearing. No events are fired at all.

At least here, inside `loadPinnedCards` the execution is hitting the `break`. Let me know if you need more info!

nikola-milicevic (approved):
> stats_blogging_reminders_nudge_shown wasn't triggered for me 😬, I will check again (few times)

Checked again, few times, all good 😊
Not sure what happened.

leandroalonso:
> stats_blogging_reminders_nudge_shown wasn't triggered for me 😬, I will check again (few times)

I have the same issue 😬

leandroalonso:
FYI:

Screen Shot 2021-10-15 at 09 16 52

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
momo-ozawa: Actually, this is a workaround to make sure the _correct_ event will be fired _once_.

If you go to a site with more than 30 views (i.e. customize should be showing) and comment out the `if viewsCount != nilstatement`, you'll see that the `publicize_nudge_shown` event is fired right before the correct event, `customize_insights_shown`. 

I tried doing b4fb243 in an attempt to fix this behavior, but that actually ended up breaking the presentation logic. (after dismissing a card, the next card shown would be the incorrect card, etc.) Maybe @nikola-milicevic has some insight on how to address this issue without breaking the presentation logic?

momo-ozawa:
> @momo-ozawa it worked just fine for all the described events, except for the blogging reminders nudge appearing. No events are fired at all.
> 
> At least here, inside `loadPinnedCards` the execution is hitting the `break`. Let me know if you need more info!

🤯 Hmmmm. Looking into this. So weird -- in your screenshot `item` evaluates to `.bloggingReminders`, which should be caught by the `case let hintType as GrowAudienceCell.HintType...` statement. Hmmmmm.

WordPress/Classes/ViewRelated/Stats/Insights/SiteStatsInsightsTableViewController.swift
leandroalonso: Oh ok! If that's the case then don't worry. Just wanted to make sure.

momo-ozawa:
@leandroalonso @nikola-milicevic 
Ready for a re-review! The mysterious bug where `blogging_reminders_nudge_shown` wasn't being fired has been fixed! 
94d4b9e

leandroalonso (approved):
@momo-ozawa :shipit: 

I also validated all the events.

PR #17302, opened 2021-10-12
Integrate gutenberg-mobile release 1.63.1

dcalhoun:
## Description
This PR incorporates the 1.63.1 release of gutenberg-mobile.
For more information about this release and testing instructions, please see the related Gutenberg-Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/4109

Release Submission Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17318, opened 2021-10-14
Merge 18.4.0.2 into `develop`

mokagio:
Includes:

- Gutenberg 1.63.1 ([#17302](https://github.com/wordpress-mobile/WordPress-iOS/pull/17302)) by @dcalhoun 

Conflicts on `Podfile` and `Podfile.lock`, I solved them by keeping 1.63.1, the version on the release branch. On `develop`, we a 1.64.0 pre-release but it doesn't yet contain those fixes. They'll come in when a new pre-release or a stable version of 1.64.0 is cut, as they are already in Gutenberg mobile's develop: wordpress-mobile/gutenberg-mobile#4113.

Nothing to test, if CI Shows green, we're good.

## Regression Notes
N.A.

N.A.

N.A.

---


AliSoftware:
> Conflicts on Podfile and Podfile.lock, I solved them by keeping 1.63.1, the version on the release branch. On develop, we a 1.64.0 pre-release but it doesn't yet contain those fixes. 

@mokagio I usually solve it the opposite way, keeping what's in `develop` to resolve the conflict. Because if you make `1.63.1` land in `develop`, any new feature worked on in `develop` which depends on what was introduced in `1.64.0-alpha*` (a very likely scenario to exist, otherwise why would they have landed the alpha in develop already if no new feature needed it?) will break if you make develop revert to `1.63.1`.

The team usually integrate a new version in `develop` very soon after fixing things in `release` — so that develop ultimately include the same fix — so imho better not break `develop`, and then wait for the fix to be duplicated there soon, rather than introducing a regression in `develop` via this PR landing.

Would love to hear @dcalhoun confirm (or deny) this though.

Podfile
```diff
@@ -166,7 +166,7 @@ abstract_target 'Apps' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :tag => 'v1.64.0-alpha2'
+    gutenberg :tag => 'v1.63.1'
```
AliSoftware, line 169: This feels like the wrong conflict resolution to me (see previous comment)

AliSoftware (changes requested):

Podfile
dcalhoun: Agreed, the correct conflict resolution would be the newer tag within `develop` — `1.64.0-alpha2`.

dcalhoun:
> Because if you make `1.63.1` land in `develop`, any new feature worked on in `develop` which depends on what was introduced in `1.64.0-alpha*` (a very likely scenario to exist, otherwise why would they have landed the alpha in develop already if no new feature needed it?) will break if you make develop revert to `1.63.1`.

This is correct. The very existence of an alpha tag in the host WordPress app repos means _something_ in Gutenberg is dependent upon the changes in the alpha release. Reverting from an alpha to a lower release will likely break the Gutenberg `develop` branch for local development against the WordPress host apps. 

> The team usually integrate a new version in `develop` very soon after fixing things in `release` — so that develop ultimately include the same fix — so imho better not break `develop`, and then wait for the fix to be duplicated there soon, rather than introducing a regression in `develop` via this PR landing.

This is somewhat true. For better or worse, we only land alpha releases into the WordPress host apps when Gutenberg _depends_ upon the changes the release contains. For the fix in `1.63.1`, it was a fix for something completely contained in Gutenberg. So, it is unlikely we will land it into the WordPress apps `develop` until (A) the next release (`1.64.0`) or (B) an alpha release for an unrelated feature (`1.65.0-alpha.X`). 

If this practices causes confusion or trouble, we could discuss immediately landing _any_ change, but we avoid doing so because it is currently a fair amount of manual work to land changes in both WordPress host apps. 😩

mokagio:
Lol and that's why I shouldn't open merge PR in the evening 🤦‍♂️

PR #17324, opened 2021-10-14
Updated Domains success screen

frosty:
This PR implements the success screen for the domains purchase flow. When rotating to landscape on a phone I opted to remove the illustration. When text gets very big, the illustration may be hidden (for 'accessibility' type sizes) and the view should scroll.

cc @kyleaparker

| Standard | XXL | Accessibility XXXL |
|---|---|---|
| Simulator Screen Shot - iPhone 12 mini - 2021-10-14 at 16 15 00 | Simulator Screen Shot - iPhone 12 mini - 2021-10-14 at 15 02 20 | Simulator Screen Shot - iPhone 12 mini - 2021-10-14 at 15 02 33 |
| Simulator Screen Shot - iPhone 12 mini - 2021-10-14 at 16 15 03 | Simulator Screen Shot - iPhone 12 mini - 2021-10-14 at 15 02 23 | Simulator Screen Shot - iPhone 12 mini - 2021-10-14 at 15 02 38 |

## To Test

- Build and run
- Try both of the following:
    - Redeem a domain credit to use the native flow, and check you see this screen at the end
    - Add another domain to use the web flow and check you see this screen. I was unable to test this as I'm having issues again with the webview auto dismissing. Looking at the code, I _think_ it should be displayed though.
- Check with different dynamic font sizes

## Regression Notes

- This affects the existing domain flows, but this screen is an improvement over what we had already. Changes are limited to this new screen.

N/A

PR submission checklist:


WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
```diff
@@ -1,12 +1,15 @@
 import UIKit
+import WordPressAuthenticator
```
frosty: I'm not a fan of using Authenticator here for the NUXButtonViewController, but we've already used it elsewhere in the domains flow so I've done it here for consistency of UI and ease of implementation. We should pull this out into the main app soon.

mokagio:
I'm going to bump this to the next release because we'll be code freezing 18.5 today and it seems like this needs a bit more touches.

From the looks of it, it seems like this might be something we want in 18.5, though? If this cannot wait two weeks and it's important that it makes it into this release, let me know and we'll organize a new beta once ready. 👍 😃

WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
```diff
@@ -20,50 +23,168 @@ class DomainCreditRedemptionSuccessViewController: UIViewController {
...
+            stackViewContainer.topAnchor.constraint(equalTo: scrollView.topAnchor),
+            stackViewContainer.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor),
+            stackViewContainer.heightAnchor.constraint(greaterThanOrEqualTo: scrollView.heightAnchor),
+
+            buttonContainer.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            buttonContainer.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            buttonContainer.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            buttonContainer.heightAnchor.constraint(greaterThanOrEqualToConstant: Metrics.buttonControllerMinHeight)
+        ])
     }
```
Gio2018: Any chance we can break this `viewDidLoad` into smaller methods?

WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
Gio2018: Can we use a `FancyButton` instead, similarly to what we are doing in `WhatIsNewView.swift` for the `continueButton`?

Gio2018 (changes requested):
Hey @frosty , the test cases worked as expected (including the web view), but I noticed that if the feature flag is off, it still shows the new success screen, and the same happens with the domain prices (cc @diegoreymendez ).
Other than that, I left a couple of code comments.

WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
diegoreymendez: +1

WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
```diff
@@ -20,50 +23,168 @@ class DomainCreditRedemptionSuccessViewController: UIViewController {
...
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .vertical
+        stackView.spacing = Metrics.stackViewSpacing
+        stackView.alignment = .fill
+
+        // Image
+
+        let illustration = UIImageView(image: UIImage(named: "domains-success"))
+        illustration.contentMode = .scaleAspectFit
+        self.illustration = illustration
```
diegoreymendez: We could initialize the `illustration` property lazily where it's declared instead of here since the initialization doesn't really take any external parameter.

If we did that the `illustration` property would no longer be optional and this method (`viewDidLoad`) would be simplified a bit.

kyleaparker:
Looking good @frosty !

I think we can remove the shadow from behind the button area, we generally use a separator line elsewhere in the app. 

I'm not sure if you have worked on the dark mode version yet, if not this is what I had in mind: 
1  Overlay_ Purchase complete

It's using the secondary system background colour and light blue for the primary button. 

Let me know what you think!

diegoreymendez (commented):
Adding some comments on the code.

WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
frosty: 👍

WordPress/Classes/ViewRelated/Domains/Domain credit/DomainCreditRedemptionSuccessViewController.swift
frosty: 👍

frosty:
Thanks folks, I think I've addressed all the feedback now. I've improved the appearance in dark mode too.

@Gio2018 The only thing I haven't done is use a feature flag – as this is just a visual improvement over the previous screen, I didn't feel that it really needed feature flagging, but if you think otherwise I can duplicate the previous view controller and selectively display the 'correct' screen according to the feature flag.

frosty:
@kyleaparker How's this?

|   |   |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-10-20 at 09 43 10 | Simulator Screen Shot - iPhone 12 Pro - 2021-10-20 at 10 04 43 |

kyleaparker:
Looks great!

Gio2018 (approved):

PR #17327, opened 2021-10-15
Merge finalized release 18.4 into `develop`

mokagio:
This includes the changes from beta 2, 18.3.0.2, as well. I opened that yesterday, but made a silly mistake in the conflicts resolution (thanks again @AliSoftware and @dcalhoun for your watchful eyes). Since I knew I'd be finalizing the build today, I decided to batch them rather than pushing a new PR and stacking this one on top of that.

You'll also notice that there is no diff for the `Podfile*` files regarding Gutenberg, as we want to keep the version in `develop` because it's the latest stable and incorporates the `1.63.1` hotfix as well.

Finally, I decided not to bring in the changes from #17305, even though I used them to upload the desired keywords to ASC. I updated the `en-*` translations in GlotPress earlier today, but I guess no one saw my suggested changes yet. The Fastlane files will update based on those eventually, so we should be good. Another reason for not updating the source of truth on which the `.po` is generated here is that this keyword work is an experiment and we don't want to kick off a full round of keywords localization based on this first iteration. 

I wouldn't be surprised if we'll end up changing the keywords process in the future to give us the ability to experiment with them faster. _Stay tuned_ 🍿


config/Version.public.xcconfig
```diff
@@ -1,4 +1,4 @@
 VERSION_SHORT=18.4
 
 // Public long version example: VERSION_LONG=9.9.0.0
-VERSION_LONG=18.4.0.1
+VERSION_LONG=18.4.0.3
```
mokagio, line 4: As I mentioned in the description, this PR also includes `18.4.0.2`, which is why there is a non unary version bump in this diff.

mokagio (commented):
- [x] Fastlane  localized `release_notes.txt` updates
- [x] Localized `.strings` updates
- [x] Build number bump in version `.xcconfig` file

AliSoftware (approved):
✅  `release_notes.txt` updated
✅  `description.txt` updated (for `nl-NL`) 🤷 
✅  `.strings` updated
✅  version bump in `.xcconfig`
✅  Gutenberg version bump in `Podfile` not part of the merge back to `develop` to not introduce regression

I'm 🤷  on the `nl-NL` description being updated – I mean, I don't think the original for the app description has been changed in a while, so why suddenly have an updated description of this? But I'm guessing it's just a `nl-NL`-speaking translator who decided to do some refinements, so sounds good to me.

fastlane/metadata/nl-NL/description.txt
```diff
@@ -1,10 +1,10 @@
...
 
 Met WordPress voor iOS heb je publicatie mogelijkheden in de palm van je hand. Schrijf een spontane haiku vanaf je bank. Maak een foto tijdens je lunchpauze en upload deze. Reageer op de laatste opmerkingen of houd je statistieken in de gaten om te zien uit welke landen de bezoekers van vandaag komen.
 
-WordPress voor iOS is een opensourceproject; dat betekent dat ook jij kunt helpen bij de ontwikkeling ervan. Ga naar https://apps.wordpress.com/contribute/ voor meer informatie.
+WordPress voor iOS is een opensourceproject, dat betekent dat ook jij kunt helpen bij de ontwikkeling ervan. Ga naar https://apps.wordpress.com/contribute/ voor meer informatie.
 
-WordPress voor iOS ondersteunt WordPress.com en zelf-gehoste WordPress.org-sites die op WordPress 4.0 of nieuwer worden uitgevoerd.
+WordPress voor iOS ondersteunt WordPress.com en zelf gehoste WordPress.org sites die op WordPress 4.0 of nieuwer worden uitgevoerd.
 
 Hulp nodig bij de app? Bezoek de forums op https://ios.forums.wordpress.org/ of tweet ons @WordPressiOS.
```
AliSoftware, line 10: 💅

PR #17333, opened 2021-10-18
Reader Comment: Add feature flag for following via notifications

dvdchr:
Refs `pctCYC-9x-p2`

Included in this PR:
- Added the feature flag for Follow Conversation via Notifications.
- Implemented logic in FollowCommentService for toggling the post subscription's notification settings.
- Added the new "Follow" button on the navigation bar. The new button is not functional yet, but it should correctly represent the subscription state (i.e. "Follow" button title when the user hasn't subscribed, or the bell icon otherwise).

## To Test

- To cut your testing time, scout for two posts where you have one post subscribed and the other unsubscribed.
- (with the feature flag disabled,) verify that the current "Follow conversation by email" buttons work as expected.
- When you have the posts ready, enable the "Follow Conversation via Notifications" feature flag.

#### Verifying the subscribed state
- Go to the Reader section, and select one of the subscribed posts. 
- Verify that the "Followed conversation by email" button in the post header is no longer there.
- Verify that the navigation bar shows a button on the right with a bell icon.

#### Verifying the unsubscribed state
- Go to the Reader section, and select one of the unsubscribed posts.
- Verify that the "Follow conversation by email" button in the post header is no longer there.
- Verify that the navigation bar shows a button with the title "Follow".

## Regression Notes
While the feature is still hidden behind a flag, the conditional implementation may affect the current Follow Conversation buttons.

Manually tested the feature, ensured there are no side effects from the feature flag implementation.

Added tests for the FollowCommentService.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/FollowCommentsService.swift
+++ b/WordPress/Classes/Services/FollowCommentsService.swift
@@ -104,10 +104,15 @@ class FollowCommentsService: NSObject {
                                           success: @escaping () -> Void,
                                           failure: @escaping (Error?) -> Void) {
         remote.updateNotificationSettingsForPost(with: postID, siteID: siteID, receiveNotifications: isNotificationsEnabled) { [weak self] in
-            guard let self = self else { return }
+            guard let self = self else {
+                failure(nil)
+                return
+            }
+
             self.post.receivesCommentNotifications = isNotificationsEnabled
             ContextManager.sharedInstance().saveContextAndWait(self.context)
             success()
+
         } failure: { error in
             DDLogError("Error updating notification settings for followed conversation: \(String(describing: error))")
             failure(error)
--- a/WordPress/Classes/Services/FollowCommentsService.swift
+++ b/WordPress/Classes/Services/FollowCommentsService.swift
@@ -94,6 +94,26 @@ class FollowCommentsService: NSObject {
         }
     }
 
+    /// Toggles the notification setting for a specified post.
+    ///
+    /// - Parameters:
+    ///   - isNotificationsEnabled: Determines whether the user should receive notifications for new comments on the specified post.
+    ///   - success: Block called after the operation completes successfully.
+    ///   - failure: Block called when the operation fails.
+    @objc func toggleNotificationSettings(_ isNotificationsEnabled: Bool,
+                                          success: @escaping () -> Void,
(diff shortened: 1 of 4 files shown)


WordPress/Classes/Services/FollowCommentsService.swift
```diff
@@ -94,6 +94,26 @@ class FollowCommentsService: NSObject {
...
+    ///
+    /// - Parameters:
+    ///   - isNotificationsEnabled: Determines whether the user should receive notifications for new comments on the specified post.
+    ///   - success: Block called after the operation completes successfully.
+    ///   - failure: Block called when the operation fails.
+    @objc func toggleNotificationSettings(_ isNotificationsEnabled: Bool,
+                                          success: @escaping () -> Void,
+                                          failure: @escaping (Error?) -> Void) {
+        remote.updateNotificationSettingsForPost(with: postID, siteID: siteID, receiveNotifications: isNotificationsEnabled) { [weak self] in
+            guard let self = self else { return }
```
ScoutHarris: Per our style guide, the `return` should be on the next line. 
```
            guard let self = self else {
                return
            }
```

ScoutHarris: Also, I'm curious about the "silent failure" here. On the off chance there is no `self`, then nothing in the success block is called. Would it be good to call `failure` inside the `guard`'s `else` block?

ScoutHarris:
> Added the new "Follow" button on the navigation bar. The new button is not functional yet, but it should correctly represent the subscription state (i.e. "Follow" button title when the user hasn't subscribed, or the bell icon otherwise).

It seems the new button is half functional. :smile: When I tap `Follow` it successfully follows the conversation. However, it does not automatically change to the bell icon. I don't know if this is in your future plans, but since it can be added later regardless, I won't hold up this PR for it.

ScoutHarris (approved):
Some minor notes, but otherwise 👍 .

:shipit:

WordPress/Classes/Services/FollowCommentsService.swift
dvdchr: Good catch! This has been addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17333/commits/8c1f6d80d00606040f2df0caf8ea46c8145b4300.

PR #17336, opened 2021-10-19
Add domain suggestion product discounts.

diegoreymendez:
Shows product pricing and discount information for domains.

## Known limitations:

The currency formatting is inconsistent.  This will be addressed separately.

## Associated PRs:

WPKit: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/455

### Claim your free domain:

Make sure that when you claim the free domain for a site, the price text still looks as it should.

### Add a domain:

Test adding a domain to a site that already spent it's domain credit.  If you search for a .blog domain you should see a discount:

## Regression Notes

None.

PR submission checklist:


WordPress/Classes/Services/DomainsService.swift
```diff
@@ -29,27 +49,61 @@ struct DomainsService {
...
+                                         domainSuggestionType: DomainsServiceRemote.DomainSuggestionType? = nil,
+                                         success: @escaping ([FullyQuotedDomainSuggestion]) -> Void,
+                                         failure: @escaping (Error) -> Void) {
+
+        productsRemote.getProducts { result in
+            switch result {
+            case .failure(let error):
+                failure(error)
+            case .success(let products):
+                getDomainSuggestions(query: query, segmentID: segmentID, quantity: quantity, domainSuggestionType: domainSuggestionType, success: { domainSuggestions in
```
Gio2018, line 81: since we don't use `products` in `getDomainSuggestions`, maybe we can optimize this code using a `DispatchGroup`, or `Combine` perhaps?

Gio2018 (approved):
This worked as expected. Just one code comment

WordPress/Classes/Services/DomainsService.swift
diegoreymendez: That was my initial intention but I found some issues that made me get stuck down the rabbit hole, and I turned back to avoid getting stuck in it.

For Combine, the caller needs to retain the subscription until the results are published, which doesn't mix great with how we use our service instances today (they're usually short lived).  I wasn't sure how we want to do it, but I'm open to trying it out.

I think a great tool would be to use `async` and `await` but those are available in iOS 15 and later.

PR #17339, opened 2021-10-19
Reader Comments: Implement functionality for the new Follow button

dvdchr:
Refs `pctCYC-9x-p2`

This adds the implementation for the new Follow button, mainly the flow that introduces the user to enable notifications through the little snackbar views. Some notes on the changes:

- Added a method on `UIViewController+Notice` so that I could show a notice with an action handler from Objective-C. I know this is slightly unrelated to the PR, but since it's a minor change and required by this PR, I figured to just include it together.
- The bottom sheet implementation will be introduced in the next PR. So it's expected that tapping the bell button does nothing for now.

Designs:

Screen Shot 2021-10-20 at 02 53 26

## To Test

Prepare some posts that are unsubscribed, and ensure that the `Follow Conversation via Notifications` flag is enabled.

- Go to Reader, and select any unsubscribed post.
- Tap on the Follow button.
- Verify that the notice appears, and displays the content correctly with the "Enable" action button.
- Verify that the Follow button now turns into a bell icon.
- Tap on the "Enable" button.
- Verify that the notice appears, and displays the content correctly with the "Undo" action button.
- Tap on the "Undo" button.
- Verify that the notice appears informing that the Undo process has been completed.

## Regression Notes
1. Potential unintended areas of impact
The feature is still hidden behind a flag, but the current follow functionality may be impacted. Additionally, this introduced some additional methods to `UIViewController+Notice` and _could_ disrupt the existing usages of `displayNotice(title: String, message: String?)` due to it being slightly modified.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Manually tested and ensured that both implementations work as expected.

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17340, opened 2021-10-19
Reader Comments: Implement notification settings sheet

dvdchr:
Refs `pctCYC-9x-p2`
Depends on #17339

This PR adds the bottom sheet UI, along with the functionality. With this, the Follow Conversation via Notifications should be fully functional. Here are a couple of notes:

- It was a bit tricky to get the view controller to work with the `BottomSheetViewController`, as the height calculation doesn't seem to match properly. To solve this, I had to do a manual calculation – but thankfully nothing is hardcoded. Let me know if you think there's a better approach! 🙂 
- There's a bit of back-and-forth between the ObjC and the Swift version of `ReaderCommentsViewController` since the `BottomSheetViewController` could only be used from Swift (without changing the bottom sheet component).
- The feature flag is still disabled. It will be enabled in the next PR, along with some Tracks to track data on notification enable/disable actions.

Designs:

Screen Shot 2021-10-20 at 04 45 05

## To Test

Make sure that the `Follow Conversation via Notifications` flag is enabled.

The best way to test this is by using the [WordPress developer console](https://developer.wordpress.com/docs/api/console/), or some kind of API client (e.g. Postman) side by side with the app. The idea is to make changes on the iOS app, and then verify the values returned from `GET v1.1/sites/$site/posts/$post_id/subscribers/mine`:
```json
{
  "i_subscribe": false,
  "receives_notifications": false,
}
```

- Go to Reader, and select an unsubscribed post.
- Tap on the Follow button.
- Verify that the API returns `{ "i_subscribe": true, "receives_notifications": false }`.
- Tap on the "Enable" button. 
- Verify that the API returns `{ "i_subscribe": true, "receives_notifications": true }`.
- Tap the "Undo" button.
- Verify that the API now returns `{ "i_subscribe": true, "receives_notifications": false }`.
- Tap on the bell icon.
- Tap on the notification switch button, to enable notifications.
- Verify that the API returns `{ "i_subscribe": true, "receives_notifications": true }`.
- Now, turn the switch off.
- Verify that the API returns `{ "i_subscribe": true, "receives_notifications": false }`.
- Tap on the "Unfollow conversation" button.
- Verify that the API now returns `{ "i_subscribe": false, "receives_notifications": false }`.
- Verify that the bell icon button now turns into "Follow".

## Regression Notes
n/a. The feature is still hidden behind a flag.

n/a. The feature is still hidden behind a flag.

n/a. The feature is still hidden behind a flag.

PR submission checklist:


WordPress/Classes/ViewRelated/Reader/ReaderCommentsNotificationSheetViewController.swift
```diff
@@ -0,0 +1,224 @@
...
+
+        view.layoutIfNeeded()
+        return .contentHeight(containerStackView.frame.height
+                                + BottomSheetViewController.Constants.additionalContentTopMargin
+                                + view.safeAreaInsets.bottom
+                                + Constants.contentInset)
+    }
+
+    func handleDismiss() {
+        // flush out the queued Notices before unlocking. Otherwise, all the Notices would be shown consecutively until the queue is empty.
```
ScoutHarris: > all the Notices would be shown consecutively until the queue is empty

Is this a bad thing? I don't think we prevent that anywhere else. Probably because it could appear as if actions didn't actually do anything.

ScoutHarris:
Some issues with appearance:

- The text size isn't dynamic.

text_size

- The popover doesn't look quite right on a iPad. There probably could be more top margin and the height could be reduced. Because it's actually a popover and not a sheet, the sizing gets funky. You may have to look at other implementations to see how they handle iPad. IDK if this is helpful, but an example is `UserProfileSheetViewController:viewDidLayoutSubviews` and `collapsedHeight`.

ipad_sheet

ScoutHarris:
On an iPad, the snackbars don't appear as expected.

On a subscribed post:
- Unfollow conversation via sheet.
- Did not get snackbar.
- Follow conversation. 
- Did not get snackbar.

On an unsubscribed post:
- Follow conversation.
- Got snackbar. 
- Select `Enable`.
- Got snackbar.
- Select `Undo`.
- Got snackbar.
- Unfollow conversation.
- Did not get snackbar.
- Follow conversation again.
- Did not get snackbar.

The snackbar seems just broken at this point. I tried all of these, but never saw the snackbar again.
- Exiting and re-viewed the comments, un/followed again.
- Un/followed other conversations. 
- Disabled the feature, and un/followed with the old follow button.

ScoutHarris (changes requested):

WordPress/Classes/ViewRelated/Reader/ReaderCommentsNotificationSheetViewController.swift
dvdchr: When I was playing around with toggling the switch, the notices would stack up. Since it's a queue, the oldest, un-shown notice will be displayed yet, which _might_ be confusing to users. e.g., assume that the state is `disabled` and these actions are performed:

- toggle -> `enabled`
- toggle -> `disabled`
- dismiss the sheet.

Then the first notice that appears is "In-app notifications enabled". Of course, if the notice is dismissed, it'll show the next in queue until the last one is displayed. Hence why I've thought of emptying the notices on dismiss.

...Or so I thought. I ended up not flushing the notices since I actually misread the documentation for `NoticeAction.empty`: 
```Swift
/// Removes all Notices except the current one.
case empty
```

I thought it removes all except the last one in the queue 😅 , but the exception turned out to be the one that's currently displayed. When tapping on the unfollow the notice still appears though, because it's processed on the `dismiss` method's completion block.

Anyways, I've now removed this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17340/commits/c74b5ba2149735ef1fd65e0cbcf92d1826f8e0cf !

dvdchr:
👋  @ScoutHarris! Thanks for catching all these issues. I was too focused on getting the sizing right on iPhone, and it broke the iPad. 🤦 

> The text size isn't dynamic.

This is fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17340/commits/37d8191545af0d9a6778a129858c8d8cc118f16c.

> The popover doesn't look quite right on a iPad. There probably could be more top margin and the height could be reduced. Because it's actually a popover and not a sheet, the sizing gets funky. You may have to look at other implementations to see how they handle iPad. IDK if this is helpful, but an example is `UserProfileSheetViewController:viewDidLayoutSubviews` and `collapsedHeight`.
>
> On an iPad, the snackbars don't appear as expected.

I've fixed this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17340/commits/42da651ba2795a5f10d8f3501d392ed7ef8a30f0. For the snack bar not appearing on iPad, it was because programmatic dismiss won't call `presentationControllerDidDismiss(_ presentationController: UIPresentationController)`, so I had to call `handleDismiss()` myself. 

Ready for your second – and hopefully last – round of review! 🙂

WordPress/Classes/ViewRelated/Reader/ReaderCommentsNotificationSheetViewController.swift
```diff
@@ -0,0 +1,289 @@
...
+    var verticalPadding: CGFloat {
+        return Constants.contentInsets.top
+            + Constants.contentInsets.bottom
+            + additionalVerticalPadding
+    }
+
+    /// Calculates the default top margin from the `BottomSheetViewController`, plus the bottom safe area inset.
+    /// The 5pt is for an extra bottom padding on iPad, to make it look better.
+    var additionalVerticalPadding: CGFloat {
+        WPDeviceIdentification.isiPad() ? 5 : BottomSheetViewController.Constants.additionalContentTopMargin + view.safeAreaInsets.bottom
```
ScoutHarris: Can this `5` be added to `Constants` with all the other iPad-y values?

ScoutHarris (changes requested):
Hey @dvdchr . The iPad popover looks much better! Although I did notice a couple more issues.

- When the popover is being shown and in-app notifications is toggled, the snackbar message doesn't appear until the popover is dismissed. I expected it immediately like any other message. It does the same on iPhone, but I guess it makes sense here as it would cover the sheet?
- With large text, when in-app notifications is toggled, the button is cut off.  (In this example the text size is XXXL.)

https://user-images.githubusercontent.com/1816888/138759077-f25c6352-cc4f-47ca-9bcd-31bbb99323c0.mp4

WordPress/Classes/ViewRelated/Reader/ReaderCommentsNotificationSheetViewController.swift
dvdchr: Resolved in https://github.com/wordpress-mobile/WordPress-iOS/pull/17340/commits/d96d5056356c4b5da80edb74828508b837345434.

dvdchr:
> When the popover is being shown and in-app notifications is toggled, the snackbar message doesn't appear until the popover is dismissed. I expected it immediately like any other message. It does the same on iPhone, but I guess it makes sense here as it would cover the sheet?
>
> With large text, when in-app notifications is toggled, the button is cut off. (In this example the text size is XXXL.)

These are all fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17340/commits/5d9f5eab0f37c9cfaab07258665b8b45f91fed17. The popover should now resize itself on content changes, and the notices should now be shown on iPad, but not on iPhone while the sheet is on screen.

Ready for your review again, thank you! 🙂

ScoutHarris (approved):
Well, I can't find anything else to complain about, so I guess we should :shipit: . 😄

dvdchr:
Thanks @ScoutHarris !

PR #17341, opened 2021-10-20
Integrate gutenberg-mobile release 1.64.1

geriux:
## Description
This PR incorporates the 1.64.1 release of gutenberg-mobile.
For more information about this release and testing instructions, please see the related Gutenberg-Mobile PR: https://github.com/wordpress-mobile/gutenberg-mobile/pull/4135

### Change 1
- https://github.com/WordPress/gutenberg/pull/35721

### Change 2
- https://github.com/WordPress/gutenberg/pull/35792

---

Release Submission Checklist

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17346, opened 2021-10-21
Merge editorialized release notes and 18.5.0.3 into `develop`

mokagio:


WordPress/Resources/AppStoreStrings.po
```diff
@@ -34,21 +34,15 @@ msgid ""
 "View the Privacy Notice for California Users at https://automattic.com/privacy/#california-consumer-privacy-act-ccpa.\n"
 msgstr ""
 
-#. translators: Keywords used in the App Store search engine to find the app.
-#. Delimit with a comma between each keyword. Limit to 100 characters including spaces and commas.
-msgctxt "app_store_keywords"
-msgid "social,notes,jetpack,writing,geotagging,media,blog,website,blogging,journal"
-msgstr ""
```
mokagio, line 41: We'll no longer translate keywords so that we can experiment with ASO using localized Fastlane files (`keywords.txt`) as the source of truth and not risk the GlotPress automation overriding them.

mokagio (commented):
LOL at the format of our release notes:

- _Something something_ easier...
- _Something something_ now...
- _Something something_ help...
- _`<closing adverb>`_, we...

image

fastlane/Fastfile
```diff
@@ -184,7 +184,6 @@ platform :ios do
       whats_new: File.join(PROJECT_ROOT_FOLDER, 'WordPress', 'Resources', 'release_notes.txt'),
       app_store_subtitle: File.join(source_metadata_folder, 'subtitle.txt'),
       app_store_desc: File.join(source_metadata_folder, 'description.txt'),
-      app_store_keywords: File.join(source_metadata_folder, 'keywords.txt'),
```
AliSoftware, line 187: You might want to remove the similar line from `fastlane/Jetpack-Fastfile` too. Except if that was intentional and we still want the keywords to be translated for Jetpack, and only disable them for WordPress because that's what the ASO study is focused on?

AliSoftware (commented):
✅  `release_notes.txt` updated for WP and JP
✅  `.po` file updated for WP and JP
⚠️   Also contains changes from the new beta (Gutenberg bugfix) in `Podfile`, but not the related version bumps in `xcconfig` 🤔 

LGTM overall, but:
 - Left a question to check if it was intentional that the Jetpack keywords were still translated (while the WordPress ones are not anymore). I think it's intentional, but just wanted to flag to be sure.
 - The fact that this PR contains the update in `Podfile`/`Podfile.lock` about `gutenberg 1.64.1` but not the app version bumps in `xcconfig` makes me think you haven't done the related beta yet?
 
Which means it would be odd to merge this PR as is with half the changes for the beta (G but not xcconfig) in it. I think it makes sense to wait for you to do the beta and have the version bump included in this PR before merging it all (hence why I'm just commenting not approving)

AliSoftware:
The Buildkite failure on `io-tests-ipad` seems a bit odd:
```
📦 Restoring cache for wordpress-mobile/wordpress-ios/ from a8c-repo-mirrors | 0s
-- | --
  | Using Git Mirror Server at http://10.254.3.14:41362
  | % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
  | Dload  Upload   Total   Spent    Left  Speed
  | 0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  | 🚨 Error: The plugin git-s3-cache pre-checkout hook exited with status 1
```

I'm hoping this was just a glitch or network issue, so I just restarted it in Buildkite, let's hope and see…

mokagio:
Looks like the Buildkite fail was just a network issue after all 😄 

image

mokagio:
> Left a question to check if it was intentional that the Jetpack keywords were still translated (while the WordPress ones are not anymore). I think it's intentional, but just wanted to flag to be sure.

Confirming that was intentional. Since no work has been done on Jetpack and, as far as I know, no work is planned either, I left the Jetpack keywords setup through GlotPress.

> The fact that this PR contains the update in `Podfile`/`Podfile.lock` about gutenberg 1.64.1 but not the app version bumps in `xcconfig` makes me think you haven't done the related beta yet?

That must be a byproduct of this PR being straight from the `release/18.5` and #17341 having gone in after I opened this but before it was merged:

image

I try to use a merge branch only when necessary (i.e. when I know there would be conflicts trying to merge into `develop`) to keep the Git history a bit tidier. But that does come with this kind of issue when there's activity on the release branch ⚖️ 

I'm going to start a new beta for that and the localization fixes soon.

fastlane/Fastfile
AliSoftware: Follow-up: that was intentional, so marking conversation as resolved.

Podfile
```diff
@@ -166,7 +166,7 @@ abstract_target 'Apps' do
     ## Gutenberg (React Native)
     ## =====================
     ##
-    gutenberg :tag => 'v1.64.0'
+    gutenberg :tag => 'v1.64.1'
```
AliSoftware, line 169: I'm surprised that `develop` is not yet on an `v1.65.0-alpha` in WPiOS (like it was when I did my beta for WPAndroid) – which is why this hasn't triggered a merge conflict and why this bump can be merged as is in `develop`

Flagging @antonis @geriux just to check about the plan for the new `1.65.0-alpha*` in WPiOS (ie either make it be cut from `1.64.1`, or, more likely, include that fix from `1.64.1` in the new `1.65.0-alpha*`, as you already did that for WPAndroid. Given the increase in Gutenberg versions in WPiOS's `develop` differed here from WPAndroid where we already had a new alpha in-between, just wanted to be sure.

I think it's all good because the upcoming `1.65.0-alpha2` you are working on, and likely plan to merge in WPiOS's `develop` (just like what you started in WPAndroid), is already supposed to include that same fix that was in `1.64.1` anyway (it's the same Gutenberg as WPAndroid after all)… but it doesn't hurt to ask confirmation and better safe than sorry 😄

AliSoftware (approved):
New commits added:
 - ✅ Version bump in `xcconfig` for new beta
 - ✅  Update of `.strings` file with latest GlotPress translations

Gutenberg version didn't generate a conflict because at the time of this writing, `develop` is still pointing to `1.64.0` (and not a new `1.65.0-alpha*` which would have been the case if a new feature or fix for Gutenberg had been necessary for a feature worked on in `develop` in the meantime), so should be all good to merge as is!

Podfile
geriux: >  I'm surprised that develop is not yet on an v1.65.0-alpha in WPiOS (like it was when I did my beta for WPAndroid) – which is why this hasn't triggered a merge conflict and why this bump can be merged as is in develop.

Usually, alpha versions are merged when there are native changes that affect Gutenberg mobile, sometimes is for both platforms but in this case, it was only for Android because of new changes to `react-native-aztec` and S3.

It's all good because next week we will cut `1.65.0` to be merged in develop before the next release cut.

AliSoftware: Thanks for the confirmation. All good then 👍

mokagio: Thanks folks!

PR #17348, opened 2021-10-21
Convert UI tests screens to be `ScreenObject` subclasses – Part 2 of many

mokagio:
Continues the work started in #17221. Refer to that PR for extra context on the structure of a `ScreenObject` subclass.

## Regression Notes
N.A.

N.A.

N.A.

---

### DIFF
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -63,8 +63,8 @@ class JetpackScreenshotGeneration: XCTestCase {
         }
 
         // Get Backup screenshot
-        let jetpackBackup = mySite
-            .gotoJetpackBackup()
+        let jetpackBackup = try mySite
+            .goToJetpackBackup()
 
         let jetpackBackupOptions = try jetpackBackup
             .goToBackupOptions()
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -50,8 +50,8 @@ class JetpackScreenshotGeneration: XCTestCase {
         }
 
         // Get Scan screenshot
-        let jetpackScan = mySite
-            .gotoJetpackScan()
+        let jetpackScan = try mySite
+            .goToJetpackScan()
 
         sleep(scanWaitTime)
 
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -66,7 +66,7 @@ class JetpackScreenshotGeneration: XCTestCase {
         let jetpackBackup = mySite
             .gotoJetpackBackup()
 
-        let jetpackBackupOptions = jetpackBackup
+        let jetpackBackupOptions = try jetpackBackup
             .goToBackupOptions()
             .thenTakeScreenshot(4, named: "JetpackBackup")
 
--- a/WordPress/UITestsFoundation/BaseScreen.swift
+++ b/WordPress/UITestsFoundation/BaseScreen.swift
@@ -90,13 +90,13 @@ extension BaseScreen {
     }
 
     @discardableResult
-    public func dismissNotificationAlertIfNeeded(_ action: FancyAlertComponent.Action = .cancel) -> Self {
+    public func dismissNotificationAlertIfNeeded(_ action: FancyAlertComponent.Action = .cancel) throws -> Self {
         if FancyAlertComponent.isLoaded() {
             switch action {
             case .accept:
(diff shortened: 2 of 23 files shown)


WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
```diff
@@ -50,8 +50,8 @@ class JetpackScreenshotGeneration: XCTestCase {
         }
 
         // Get Scan screenshot
-        let jetpackScan = mySite
-            .gotoJetpackScan()
+        let jetpackScan = try mySite
+            .goToJetpackScan()
```
mokagio, line 54: Notice the camel case fix in the method: `goto` 👉 `goTo`. I've encountered a bunch of methods with `goto` and tackled them as part of the migration for the screen they belonged to.

WordPress/UITestsFoundation/BaseScreen.swift
```diff
@@ -90,13 +90,13 @@ extension BaseScreen {
...
-    public func dismissNotificationAlertIfNeeded(_ action: FancyAlertComponent.Action = .cancel) -> Self {
+    public func dismissNotificationAlertIfNeeded(_ action: FancyAlertComponent.Action = .cancel) throws -> Self {
         if FancyAlertComponent.isLoaded() {
             switch action {
             case .accept:
-                FancyAlertComponent().acceptAlert()
+                try FancyAlertComponent().acceptAlert()
             case .cancel:
-                FancyAlertComponent().cancelAlert()
+                try FancyAlertComponent().cancelAlert()
```
mokagio, line 99: I considered creating the `FancyAlertComponent` instance before the `switch` so that I could "reuse it" between cases. But, only one case runs at runtime, so there wouldn't have been any real benefit. Besides, the compiler might be doing other optimizations under the hood, too.

mokagio: Now that I've written this comment, though, I noticed that I could have done something like:

```swift
guard let alert = try? FancyAlertComponent() else { return self }

switch action {
case .accept: alert.acceptAlert()
...
```

🤔 I might do this in a followup...

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
```diff
@@ -17,10 +17,10 @@ class SupportScreenTests: XCTestCase {
 
     // Test Support Section Loads
     // From Prologue > continue, tap "help" and make sure Support Screen loads
-    func testSupportScreenLoads() {
-        let supportScreen = PrologueScreen().selectContinue().selectHelp()
+    func testSupportScreenLoads() throws {
+        let supportScreen = try PrologueScreen().selectContinue().selectHelp()
 
-        XCTAssert(supportScreen.isLoaded())
+        XCTAssert(supportScreen.isLoaded)
```
mokagio, line 23: Reminder: `ScreenObject` has this as a computed variable instead of a function. 

I did this to match the Swift APIs, e.g. `Array` `isEmpty`.

mokagio:
@pachlava I run out of time in my day to see this PR through in CI.

It's possible the tests will fail due to some time outs. I think it's a mix of an issues that were there already my changes revealed and stuff I genuinely broke by changing some of the assumptions on which element is required to be available at some point in time. In my [long running WIP branch](https://github.com/wordpress-mobile/WordPress-iOS/tree/convert-to-screen-objects-p3), I have a few fixes that I might apply if that happens:

- c77c053f4487480e81f32574cdd25692ac290697
- d0d582448635031ec4250547e70d47aedac9a655
- 85fb3a16ecf6063f031a23912ee3846f5a1e3a1e
- 8ea9c703597be80ee428fe9aa24537dbc98cee31

WordPress/UITestsFoundation/Screens/Login/Unified/GetStartedScreen.swift
```diff
@@ -1,26 +1,40 @@
...
-private struct ElementStringIDs {
-    static let navBar = "WordPress.GetStartedView"
-    static let emailTextField = "Email address"
-    static let continueButton = "Get Started Email Continue Button"
-    static let helpButton = "authenticator-help-button"
-}
+public class GetStartedScreen: ScreenObject {
+
+    private let navBarGetter: (XCUIApplication) -> XCUIElement = {
+        $0.navigationBars["Get Started"]
```
mokagio, line 7: Note that this was "WordPress.GetStartedView" before. I'm not sure how the tests passed, because this version can't find this view. The string also don't appear in this repo or in the authenticator pod. 🤔 Maybe the element was created but never accessed?

pachlava: I think this is the case.

pachlava (approved):
I've executed the UI tests locally (passed), and also took a closer look into each changed file in Xcode, if it was about more than addition of `throws` / `try` - and noticed no issues. With auto-merge enabled, giving it an approval now 🚀. Thank you, Gio!

PR #17351, opened 2021-10-21
Comment Moderation: when Next is tapped on the snackbar, show the next comment.

ScoutHarris:
Ref: #17290 

When `Next` is tapped on the comment moderation confirmation snackbar, the comment details view is updated with the next comment in the current filter. If the currently displayed comment is the last one, `Next` is not displayed on the snackbar.

To test:
- Enable the `newCommentDetail` feature.
- Go to My Site > Comments.
- Select a comment and moderate it. Stay on the details view.
  - Tap `Next` on the snackbar.
  - Verify the comment details view is updated with the next comment in the list.
- Select the last comment in a list and moderate it. Stay on the details view.
  - Verify the confirmation snackbar does not have a `Next` button.
 
NOTE: The comment content field height is not updated correctly when the next comment has less content than the prior one. The same problem occurs when editing a comment to have less content. As that is not related to this change, I'll leave that to be addressed later.

## Regression Notes
1. Potential unintended areas of impact
N/A. Feature is incomplete.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N/A. Feature is incomplete.

3. What automated tests I added (or what prevented me from doing so)
N/A. Feature is incomplete.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #17290, opened 2021-10-08
Comment moderation: `Next` button on confirmation snackbar


ScoutHarris:
From design discussion wp.me/pbArwn-2ks#comment-3346:

Tapping Next in the snackbar always advances to the next item in the list you opened the detail view from.

Tasks:
- [x] #17345
- [x] #17351
- [x] #17356
- [x] #17364

![next](https://user-images.githubusercontent.com/1816888/137962337-6d58649d-4418-4572-9053-db11ea8de5e1.png)

PR #17356, opened 2021-10-22
Comment Moderation snackbar: add `>` after the Next button

ScoutHarris:
Ref: #17290 
Depends on: #17351

This adds a `>` image after the `Next` button on the comment moderation confirmation snackbar.

NOTE: The tappable area is still just the size of the `Next` button (this is legacy), so it's kind of small. I'll see about making it bigger and including the arrow in a follow up.

To test:
- Go to My Site > Comments.
- Select a comment that is not last in a list.
- Moderate it and stay on the details view.
- Verify `>` appears after the `Next` button.

next_action

- Select the last comment in a list.
- Moderate it and stay on the details view.
- Verify `Next >` does not appear. 

no_action

- Do some other action in the app that presents a snackbar with an action button. (The new `followConversationViaNotifications` is a good example.)
- Verify the `>` does not appear.
 
other_action

## Regression Notes
N/A. Feature is incomplete.

N/A. Feature is incomplete.

N/A. Feature is incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/UIViewController+Notice.swift
+++ b/WordPress/Classes/Extensions/UIViewController+Notice.swift
@@ -1,13 +1,25 @@
 import Foundation
 import WordPressFlux
 
-@objc extension UIViewController {
+extension UIViewController {
     @objc func displayNotice(title: String, message: String? = nil) {
         displayActionableNotice(title: title, message: message)
     }
 
-    @objc func displayActionableNotice(title: String, message: String? = nil, actionTitle: String? = nil, actionHandler: ((Bool) -> Void)? = nil) {
-        let notice = Notice(title: title, message: message, actionTitle: actionTitle, actionHandler: actionHandler)
+    @objc func displayActionableNotice(title: String,
+                                       message: String? = nil,
+                                       actionTitle: String? = nil,
+                                       actionHandler: ((Bool) -> Void)? = nil) {
+        displayActionableNotice(title: title, message: message, style: NormalNoticeStyle(), actionTitle: actionTitle, actionHandler: actionHandler)
+    }
+
+    // NoticeStyle is Swift only, so this method is needed to set it.
+    func displayActionableNotice(title: String,
+                                 message: String? = nil,
+                                 style: NoticeStyle = NormalNoticeStyle(),
+                                 actionTitle: String? = nil,
+                                 actionHandler: ((Bool) -> Void)? = nil) {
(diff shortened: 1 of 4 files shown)


dvdchr (approved):
> NOTE: The tappable area is still just the size of the Next button (this is legacy), so it's kind of small. I'll see about making it bigger and including the arrow in a follow up.

👍 . Also unrelated, but I noticed that in large content sizes, the labels are truncated: 

Screen Shot 2021-10-26 at 00 26 03

Perhaps we need @mattmiklic 's thoughts on how to best present this in large sizes.

Anyways, LGTM!

mattmiklic:
> Perhaps we need @mattmiklic 's thoughts on how to best present this in large sizes.

Thinking back to the call we had with Apple's accessibility team earlier in the year (internal ref: pbArwn-2d9-p2) the best option would be to make each option full-width, so the options are shown in a vertical rather than horizontal list. I can mock that up if it'd be helpful, but what do y'all think about the technical feasibility of that approach?

ScoutHarris:
Hey @mattmiklic .

> I can mock that up if it'd be helpful

Yes please! 

> what do y'all think about the technical feasibility of that approach?

Depends on ☝️ of course, but yep it should be doable. The buttons are already in a horizontal stack view. So flipping it to vertical should be fairly straightforward.

Thank you!

(And thanks to @dvdchr for noticing!)

mattmiklic:
Screen Shot 2021-11-03 at 5 36 46 PM

Here's a visual example of what I was thinking. This shows the screen content at text size AX1, the smallest of the "[larger accessibility type sizes](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/#dynamic-type-sizes)". I'm not sure about how we'd decide when to switch to the vertical layout; ideally before text gets truncated obviously, but that will differ by language. I'm open to suggestions on the logic that would control the switch.

I've added the vertical moderation control to the Components section in Zeplin.

ScoutHarris:
Thank you @mattmiklic. This is very helpful! (And looks quite nice to boot.)

> I'm not sure about how we'd decide when to switch to the vertical layout

Typically we just check if the text size is accessibility (specifically `traitCollection.preferredContentSizeCategory.isAccessibilityCategory`). I would think that would trigger for _any_ accessibility text size, including AX1. But I'll verify when I implement the changes.

mattmiklic:
That makes sense! Sounds good to me.

PR #17359, opened 2021-10-25
Convert UI tests screens to be `ScreenObject` subclasses – Part 3 of many

mokagio:
See https://github.com/wordpress-mobile/WordPress-iOS/pull/17221 and https://github.com/wordpress-mobile/WordPress-iOS/pull/17348.

## Regression Notes
N.A.

N.A.

N.A.

---

### DIFF
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -77,7 +77,7 @@ class JetpackScreenshotGeneration: XCTestCase {
         }
 
         // Get Stats screenshot
-        let statsScreen = mySite.gotoStatsScreen()
+        let statsScreen = try mySite.goToStatsScreen()
         statsScreen
             .dismissCustomizeInsightsNotice()
             .switchTo(mode: .months)
--- a/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
+++ b/WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
@@ -35,7 +35,7 @@ class JetpackScreenshotGeneration: XCTestCase {
     func testGenerateScreenshots() throws {
 
         // Get My Site screenshot
-        let mySite = MySiteScreen()
+        let mySite = try MySiteScreen()
             .showSiteSwitcher()
             .switchToSite(withTitle: "yourjetpack.blog")
             .thenTakeScreenshot(1, named: "MySite")
--- a/WordPress/UITestsFoundation/Globals.swift
+++ b/WordPress/UITestsFoundation/Globals.swift
@@ -13,4 +13,27 @@ extension ScreenObject {
     public func pop() {
         navBackButton.tap()
     }
+
+    public func openMagicLink() {
+        XCTContext.runActivity(named: "Open magic link in Safari") { activity in
+            let safari = XCUIApplication(bundleIdentifier: "com.apple.mobilesafari")
+            safari.launch()
+
+            // Select the URL bar when Safari opens
+            let urlBar = safari.textFields["URL"]
+            if !urlBar.waitForExistence(timeout: 5) {
+                safari.buttons["URL"].tap()
+            }
+
+            // Follow the magic link
+            var magicLinkComponents = URLComponents(url: WireMock.URL(), resolvingAgainstBaseURL: false)!
+            magicLinkComponents.path = "/magic-link"
+            magicLinkComponents.queryItems = [URLQueryItem(name: "scheme", value: "wpdebug")]
+
+            urlBar.typeText("\(magicLinkComponents.url!.absoluteString)\n")
+
+            // Accept the prompt to open the deep link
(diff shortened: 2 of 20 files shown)


WordPress/UITestsFoundation/Globals.swift
```diff
@@ -13,4 +13,27 @@ extension ScreenObject {
...
+            var magicLinkComponents = URLComponents(url: WireMock.URL(), resolvingAgainstBaseURL: false)!
+            magicLinkComponents.path = "/magic-link"
+            magicLinkComponents.queryItems = [URLQueryItem(name: "scheme", value: "wpdebug")]
+
+            urlBar.typeText("\(magicLinkComponents.url!.absoluteString)\n")
+
+            // Accept the prompt to open the deep link
+            safari.scrollViews.element(boundBy: 0).buttons.element(boundBy: 1).tap()
+        }
+    }
```
mokagio, line 38: This is duplicated from the `BaseScreen` version. Once we'll have no more `BaseScreen` in the UI tests, it will become the only one in the codebase. It's also something we might want to move to the authentication framework.

WordPress/UITestsFoundation/Screens/Login/WelcomeScreen.swift
```diff
@@ -7,15 +8,21 @@ private struct ElementStringIDs {
     static let signupButton = "Prologue Signup Button"
 }
```
mokagio: Note to self: remove this unused `struct` before merging.

WordPress/WordPressUITests/Tests/MainNavigationTests.swift
```diff
@@ -8,8 +8,8 @@ class MainNavigationTests: XCTestCase {
         setUpTestSuite()
 
         try LoginFlow.login(siteUrl: WPUITestCredentials.testWPcomSiteAddress, email: WPUITestCredentials.testWPcomUserEmail, password: WPUITestCredentials.testWPcomPassword)
-        mySiteScreen = TabNavComponent()
-         .gotoMySiteScreen()
+        mySiteScreen = try TabNavComponent()
+            .goToMySiteScreen()
```
mokagio, line 12: Just an indentation fix

WordPress/JetpackScreenshotGeneration/JetpackScreenshotGeneration.swift
```diff
@@ -77,7 +77,7 @@ class JetpackScreenshotGeneration: XCTestCase {
         }
 
         // Get Stats screenshot
-        let statsScreen = mySite.gotoStatsScreen()
+        let statsScreen = try mySite.goToStatsScreen()
```
mokagio, line 80: There'll be a bunch more of this changes because of the `goto` to `goTo` rename. Sorry for the noise.

WordPress/UITestsFoundation/Screens/Login/WelcomeScreen.swift
```diff
@@ -31,6 +33,6 @@ public class WelcomeScreen: BaseScreen {
     }
 
     static func isLoaded() -> Bool {
-        return XCUIApplication().buttons[ElementStringIDs.loginButton].exists
+        (try? WelcomeScreen().isLoaded) ?? false
     }
```
pachlava, line 37: I know this is the same in previous PRs, but for some reason it caught my eye only now. 

To be honest, I spent some time not understanding the reason for having an `isLoaded()` method for some of the screens, when there's an `isLoaded` variable in the `ScreenObject`. After a bit of processing, I realized that this is a shorter way to calculate that property in the runtime, without the need to handle potential error every time.

pachlava (approved):
I executed the UI tests and screenshot generation locally, and both passed. Thanks for doing the code clean-up besides the 'main' task 👍

WordPress/UITestsFoundation/Screens/Login/WelcomeScreen.swift
mokagio: Correct.

To be fair, the `static` approach made sense only in the `BaseScreen` context, where a calling `init` on a screen not displayed would fail the test. With `ScreenObject`, it's less useful. I'd like to remove as many of these as possible.

For example, the following doesn't make much sense anymore:

https://github.com/wordpress-mobile/WordPress-iOS/blob/b6d9650a894838f5879ee3d852fda32310171ae7/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift#L141-L147

We could replace it with something like:

```swift
guard let alert = FancyAlertComponent() else {
    publishNowButton.tap() 
    return
}

alert.acceptAlert()
```

On the other hand, this does. Since we don't access any of those screen directly, it's handy to call a method on the type:

https://github.com/wordpress-mobile/WordPress-iOS/blob/3675ad4ed9bc44dd374a2b2b10c1fdf6efb587d1/WordPress/WordPressUITests/Flows/EditorFlow.swift#L6-L8

pachlava: Thanks for explaining it, Gio!

PR #17338, opened 2021-10-19
Blog url creation: Add check for optional strings.

ScoutHarris:
Fixes #17330 

This adds check for the two optional strings that are used to create a url string. This should prevent the crash on #17330.

To test:
I couldn't repro the crash, but per the Sentry logs it looks like it can occur when attempting to create a wp.com site or add a self-hosted site. Thus:
- Log in with an account that has no sites.
- Select `Add new site`.
- Select either `Create WordPress.com site` or `Add self-hosted site`.
- Verify the app doesn't crash.

## Regression Notes
1. Potential unintended areas of impact
Should be none.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Tested options for `Create WordPress.com site` and `Add self-hosted site`.

3. What automated tests I added (or what prevented me from doing so)
N/A.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17365, opened 2021-10-26
Merge 18.4.1 hotfix into release 18.5

mokagio:
Includes the crash fix from #17338.

As usual, attempting the merge generated conflicts on the version `.xcconfig` files, which I resolved by keep the version from the `release/18.5` branch, as that is the latest and current.


mokagio (commented):
The only file in the diff is the one from the fix PR, as expected.

AliSoftware (approved):
Commit list and conflict resolution make sense (even though it confused me at first to see b32bc7e2b57671250a8ae2444a41732ed26ddbb2 and 39672f791da672b56066c9588b3c967b7de1f97d being the same fix applied twice, but then realized that one is the cherry-pick and the other is the merge commit… so it all checks out 👍)

:shipit:

AliSoftware:
CI failure on `buildkite/wordpress-ios/io-tests-iphone` check was the usual `client_loop: send disconnect: Broken pipe`, so just restarted it.

(We apparently can't enable auto-merge on PRs targeting branches other than `develop` or `trunk`, so we'll have to wait for CI to go green to merge manually afterwards.)

PR #17363, opened 2021-10-25
Reader Comments: Follow conversation via notifications

dvdchr:
Refs: `pctCYC-9x-p2`
Depends on #17369

This change enables the "Follow Conversation via Notifications" feature flag for the public. Additionally, this PR also addresses https://github.com/wordpress-mobile/WordPress-iOS/pull/17339#discussion_r732240314 regarding localization strings placements.

For now, the strings are encapsulated in an enum in Swift, and exposed to the Obj-C side. Since the feature flag is not yet removed, it only returns the titles for enabling/disabling notifications. Once the feature flag is removed, I'll add the titles for following/unfollowing conversations.

## To test

- Go to Reader, and select any unsubscribed post.
- Verify that the "Follow" button should be visible on the navigation bar.
- Tap the Follow button.
- Verify that the snackbar with the "Enable" action button is shown.
- Tap on the "Enable" button.
- Verify that a feedback notice with an "Undo" action button is shown.
- Tap on the "Undo" action button.
- Verify that a feedback notice is shown.
- Verify that the Follow button has turned into a bell icon button in the navigation bar.
- Go to Reader, and select any subscribed post.
- Tap on the bell icon button.
- Verify that the bottom sheet is shown.
- Tap on the switch.
- Dismiss the bottom sheet.
- Verify that a feedback notification is shown, and correctly represents the desired state.
- Tap the bell icon button again.
- Tap on the Unfollow Conversation button.
- Verify that a feedback notification is shown.
- Verify that the bell icon button has now turned back into "Follow".

## Related PRs:
- #17369
- #17340 
- #17339 
- #17333
- #17322
- https://github.com/wordpress-mobile/WordPressKit-iOS/pull/452

## Regression Notes
1. Potential unintended areas of impact
When updating from the previous version, the "follow conversation" state may be incorrectly represented (e.g. off when it's supposed to be on). Otherwise, the feature itself is pretty isolated so it should not impact other features.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Manually tested the feature.

3. What automated tests I added (or what prevented me from doing so)
Added unit tests for the service changes related to the follow conversation feature.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17369, opened 2021-10-26
Reader Comments: Track when notification setting is toggled

dvdchr:
Refs https://github.com/wordpress-mobile/WordPress-iOS/pull/17363#issuecomment-951362139
Depends on #17340

This adds a new event, `reader_toggle_comment_notifications`, which tracks whenever the comments notification setting is toggled. In addition, these properties are added:

- `blog_id`: The site ID of the blog that the user enabled the notifications on.
- `notifications_enabled`: `true` if the notifications is enabled, or `false` otherwise.

## To test:

Make sure the "Follow Conversation via Notifications" flag is enabled. You should see `🔵 Tracked: reader_toggle_comment_notifications` when:

- Tapping "Follow" > "Enable" notice action.
- Tapping the "Undo" notice action.
- Toggling the switch for the in-app notifications sheet.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/FollowCommentsService.swift
+++ b/WordPress/Classes/Services/FollowCommentsService.swift
@@ -103,6 +103,9 @@ class FollowCommentsService: NSObject {
     @objc func toggleNotificationSettings(_ isNotificationsEnabled: Bool,
                                           success: @escaping () -> Void,
                                           failure: @escaping (Error?) -> Void) {
+        WPAnalytics.trackReader(.readerToggleCommentNotifications,
+                                properties: [WPAppAnalyticsKeyBlogID: self.siteID, AnalyticsKeys.enabled: isNotificationsEnabled])
+
         remote.updateNotificationSettingsForPost(with: postID, siteID: siteID, receiveNotifications: isNotificationsEnabled) { [weak self] in
             guard let self = self else {
                 failure(nil)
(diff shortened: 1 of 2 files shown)


WordPress/Classes/Services/FollowCommentsService.swift
```diff
@@ -113,6 +113,9 @@ class FollowCommentsService: NSObject {
             ContextManager.sharedInstance().saveContextAndWait(self.context)
             success()
 
+            WPAnalytics.trackReader(.readerToggleCommentNotifications,
+                                    properties: [WPAppAnalyticsKeyBlogID: self.siteID, AnalyticsKeys.enabled: isNotificationsEnabled])
+
```
ScoutHarris: This will only log the event if it's successful. We usually track users' intent, so we don't usually log events only on success. Given that, this should probably be before the `remote` call so it's always logged.

ScoutHarris (approved):
Just one suggestion. Otherwise :shipit: !

WordPress/Classes/Services/FollowCommentsService.swift
dvdchr: I see... That makes sense. This has been addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17369/commits/c40d2fc4d643228c94105f286d1b30a8e7323741. Thank you!

PR #17379, opened 2021-10-27
iOS 15: Fix user profile sheet being shown fullscreen

### LINKED ISSUE #17378: iOS 15: User profile sheet is fullscreen
### Expected behavior
The user profile sheet should be collapsed initially.

### Actual behavior
The user profile sheet is full screen initially.

Go to either:
- Notifications > Likes.
- Reader > Post > Likes.

Select a liker. Notice the user profile is fullscreen instead of collapsed.

| iOS15 | iOS14 |
|--------|-------|
| ios15 | ios14 |

ScoutHarris:
Fixes #17378

This fixes the issue where, on iOS15 iPhone, the user profile sheet was presented fullscreen instead of collapsed.

NOTE: Making this change in `develop` didn't fix the problem. This needs other iOS15 fixes in the `issue\xcode_13` branch. Thus, this PR targets that branch.

To test:
On an iOS15 iPhone:
- Go to either:
  - Notifications > Likes.
  - Reader > Post > Likes.
- Select a liker.
- Verify the sheet is collapsed initially.

| Before | After |
|--------|-------|
| before | after |

## Regression Notes
Prior iOS versions could render incorrectly.

Tested on iOS13 and iOS14 to verify no change there.

N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/User Profile Sheet/UserProfileSheetViewController.swift
+++ b/WordPress/Classes/ViewRelated/User Profile Sheet/UserProfileSheetViewController.swift
@@ -45,7 +45,6 @@ class UserProfileSheetViewController: UITableViewController {
         size.height += bottomPadding
 
         preferredContentSize = size
-        presentedVC?.presentedView?.layoutIfNeeded()
     }
 }
 
@@ -171,6 +170,7 @@ private extension UserProfileSheetViewController {
     func configureTable() {
         tableView.backgroundColor = .basicBackground
         tableView.separatorStyle = .none
+        tableView.isScrollEnabled = false
     }
 
     func registerTableCells() {


leandroalonso (approved):
@ScoutHarris works as described!

Small suggestion: it was an intentional decision to allow the user to resize the bottom sheet?

If we do this:

```
extension UserProfileSheetViewController: DrawerPresentable {

    var allowsUserTransition: Bool {
        false
    }
```

This doesn't allow the user to increase the bottom sheet, which IMHO is a nice touch. However, it's completely up to you!

dvdchr (approved):
👋  @ScoutHarris. Tested and it works well! 

I've spotted a minor visual issue when the sheet is displayed in devices without a home indicator (e.g. iPhone 8). It looks like the initial sheet is shown with an extra bottom spacing, but it seemed to be sized correctly after I maximized the sheet and minimized it back to the collapsed state. Here's a preview:

https://user-images.githubusercontent.com/1299411/139274668-12be838e-7f3c-4070-b96f-547629d2d1ab.mp4

I'm thinking if it's due to the usage of `Constants.iPhoneBottomPadding`, and using `view.safeAreaInsets.bottom` (along with a reduced constant value) could work better in this case?

ScoutHarris:
Hey guys. 

> it was an intentional decision to allow the user to resize the bottom sheet?

@leandroalonso It wasn't a conscious decision either way. I didn't even think about it actually. I suppose there's no reason for the user to resize the sheet. Good point! I'll fix that separately as it is not iOS15 related and should be fixed in `develop`.

> I've spotted a minor visual issue when the sheet is displayed in devices without a home indicator 

@dvdchr Nice catch! This does look iOS15 related, but I'll fix it separately as well since it's a different issue.

Thanks for the feedback!

PR #17399, opened 2021-11-01
Comments: Add reply ID to the object schema

dvdchr:
Refs #17087

This PR modifies the Comments model to add `replyID`. Some notes:

- This property will be used later to store the comment ID of the user's reply to this comment. For example:
  ```
  - Logged in with account X
  - comment 1 by Y
    - comment 2 by Z
    - comment 3 by X
  ```
  The `replyID` of comment 1 will be `3` (as in comment 3), since comment 3 is the current user's reply to comment 1.
- The property is initialized with the default value `0`. However, each time the user enters the Comment Detail page, a fetch will be performed to check if the `replyID` needs to be updated.  There are some scenarios:
  -  From `0 to <something>`: The most obvious example is the current user submitted a reply to an unreplied comment. This would be the most common scenario.
  - From `<something> to 0`: The current user's reply could be deleted. In this case, the parent comment needs to update its `replyID` since it's no longer relevant.
  - From `<something> to <something else>`: The current user posted a reply to an already replied post. In this case, the parent comment should update its `replyID` to the most recent one. Conversely, considering the previous comment deletion behavior, the user's most recent reply could be deleted and thus updating the parent comment's `replyID` to an older reply.

Considering the scenarios above, the most sensible place to update is where the reply indicator is shown (i.e. the upcoming Comment Detail page). Each time the Comment Detail page is loaded, perform a [fetch](https://github.com/wordpress-mobile/WordPressKit-iOS/pull/463) to check if the user has an existing reply to the comment – then show/hide the reply indicator accordingly.

## To Test

Ensure that the app doesn't crash:

- My Site > Comments list
- Comment Details
- Comment Thread
- Posting a new comment
- Moderating comments
- Deleting a comment

## Regression Notes
Should be none. The property is not used anywhere, it has a default value, and is marked optional.

Tested various screens that use the Comment object, and ensured that nothing went wrong.

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Comment+CoreDataProperties.swift
+++ b/WordPress/Classes/Models/Comment+CoreDataProperties.swift
@@ -25,6 +25,7 @@ extension Comment {
     @NSManaged public var parentID: Int32
     @NSManaged public var depth: Int16
     @NSManaged public var hierarchy: String
+    @NSManaged public var replyID: Int32
 
     /*
      // Hierarchy is a string representation of a comments ancestors. Each ancestor's
--- a/WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
+++ b/WordPress/Classes/WordPress.xcdatamodeld/.xccurrentversion
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>WordPress 135.xcdatamodel</string>
+	<string>WordPress 136.xcdatamodel</string>
 </dict>
 </plist>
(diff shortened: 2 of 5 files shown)


ScoutHarris (changes requested):
Hey @dvdchr .

I think the wrong model version got updated.

When I look at 136:
- Comment does not have `replyID`.
- The Current model version is 135.

When I look at 135:
- Comment does have `replyID`.
- The Current model version is 136.

dvdchr:
> I think the wrong model version got updated.

🤦 . Sorry about that! I've addressed this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17399/commits/86f8ff7cb4806bfe3f7f2dae9b079427df7979d7. The property is now added in version 136 instead, and the current model version is now updated to 136.

ScoutHarris (approved):
Tested with WP and self-hosted. 👍 !

:shipit:

PR #17408, opened 2021-11-02
Comments: Add method to fetch the latest reply comment ID

dvdchr:
Refs #17087 
Depends on #17399, https://github.com/wordpress-mobile/WordPressKit-iOS/pull/463

As titled, this PR extends Comment Service so it could fetch the latest reply ID from a specified comment ID. In the method parameters, I've added `accountService` to make it testable. As for the `AccountService` usage, the method needs to fetch the current user's dotcom ID to filter returned comments. 

## To test

Since there are no visual changes, make sure all the tests pass.

## Regression Notes
1. Potential unintended areas of impact
n/a. Feature is unreleased.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a. Feature is unreleased.

3. What automated tests I added (or what prevented me from doing so)
Added unit tests for the `CommentService`.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17409, opened 2021-11-02
Comment Details: Implement reply indicator functionality

dvdchr:
Refs #17087 
Depends on #17408 

This PR implements the reply indicator functionality for the new comment details screen (finally! 😅 ). Some notes:
- When navigating to the new comment details, the reply ID will be fetched on every `viewDidLoad`. This means, `Comment`'s `replyID` is not guaranteed to be relevant, but it will _eventually_ be correct once it's used in `CommentDetailViewController`. 
- While the `replyID` is now available in the `Comment` model, I'll be using the local `replyID` variable in `CommentDetailViewController` as preparation for the later integration into Notifications, since we'll be getting the reply ID directly from the `Notification` data. The variable is still currently populated with `Comment`'s `replyID`, though.
-  As for the property observer on `replyID`, I tried to ensure that only the reply indicator cell is updated (either inserted or removed) in the table view in order to prevent the content cell from reloading its web content twice. Also, for reference: https://github.com/wordpress-mobile/WordPress-iOS/pull/17399

## To test

Ensure that the `New Comment Details` feature flag is enabled.

#### Verifying the default state
- Go to My Sites > Comments, and select a comment that you haven't replied to yet.
- Verify that the reply indicator cell is not shown.
- Go back to My Sites > Comments, and select a comment that you have replied.
- Verify that the reply indicator cell is shown.
- Tap the reply indicator cell.
- Verify that you're navigated into the comment thread, highlighting your reply comment.

#### Verifying reply indicator state changes
- Go to My Sites > Comments, and select one of your reply comments.
- Set your reply comment's moderation status to Pending.
- Go back to My Sites > Comments, and select the parent comment.
- Verify that the reply indicator is initially shown, but is hidden shortly after.
- Go back to your reply comment again.
- Set your reply comment's moderation status to Approved.
- Go back to the parent comment again.
- Verify that the reply indicator is initially not shown, but is shown shortly after.

## Regression Notes

PR submission checklist:


WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -20,6 +20,38 @@ class CommentDetailViewController: UITableViewController {
         return navigationController?.visibleViewController == self
     }
 
+    private var replyID: Int32 {
+        didSet {
+            // toggle reply indicator cell visibility as needed.
+            // to prevent the content cell from reloading its web content, only the specific index path is targeted.
+            switch (oldValue, replyID) {
+            case (.zero, 1...Int32.max):
```
ScoutHarris: I find these `switch` `case` statement confusing and hard to read. I figured it out after thinking about it, but honestly IMO I shouldn't have to think that hard about what each `case` does. 😄 Can these be clearer / more readable?

ScoutHarris (approved):
Hey @dvdchr . I just had one request. Otherwise, it seems to work fine. 🏅 Nice!

:shipit:

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
dvdchr: Thanks for the input! I've tried to improve this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17409/commits/bfb2b792e0d7b62fcdc579c8f6e7a108594865cc by using the traditional if/else. Hope this is a bit clearer now or otherwise please let me know! 🙂

ScoutHarris: Thank you!!

Issue #17412, opened 2021-11-02
Comment Details: add Reply functionality


ScoutHarris:
On the [new](https://github.com/wordpress-mobile/WordPress-iOS/issues/17087) Comment Details view, add Reply functionality. Update the Reply view UI per design.

New Comment Details Reply:
- [x] #17414
- [x] #17428
- [x] #17432
- [x] #17433

Shared Reply views:
- [x] #17443
- [x] #17445
- [x] #17447
- [x] #17448

From Zeplin:
<img width="1049" alt="Screen Shot 2021-11-09 at 5 58 52 PM" src="https://user-images.githubusercontent.com/1816888/141030223-6db35487-30a5-4084-9f26-308d6642bf6c.png">

PR #17414, opened 2021-11-03
Convert CommentDetailViewController to a UIViewController

ScoutHarris:
Ref: #17412 
Depends on: #17409 (to avoid massive merge conflicts)

This changes `CommentDetailViewController` from a `UITableViewController` to a `UIViewController`. 

This is to facilitate showing `ReplyTextView` and `SuggestionsTableView` when the `Reply` button is tapped. In a future PR, these views to be added to the new `containerStackView` when needed.

To test:
- Enable `newCommentDetail`.
- Go to My Site > Comments > Comment details.
- Verify the details appears and functions correctly.
- On a split view and non split view device, rotate and verify the view is displayed correctly.

## Regression Notes
N/A. Feature is flagged.

N/A. Feature is flagged.

N/A. Feature is flagged.

PR submission checklist:


WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -187,7 +190,8 @@ class CommentDetailViewController: UITableViewController {
         self.isLastInList = isLastInList
         self.managedObjectContext = managedObjectContext
         self.replyID = comment.replyID
-        super.init(style: .plain)
+        super.init(nibName: nil, bundle: nil)
+        configureView()
```
dvdchr: Not a major issue, but calling this method in `init` causes the `viewDidLoad` to be prematurely called (since `configureView` calls the `view` property which triggers `viewDidLoad`) – this caused the `navigationController` to be nil while configuring the navigation bar, thus causing `translucent` property to stay false.

This caused a minor visual bug where the content won't scroll below the navigation bar (since it's not translucent). What do you think about moving this to `viewDidLoad`? I've tried it, and it seems to fix the issue. 🙂

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -791,3 +729,84 @@ private extension CommentDetailViewController {
...
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
+        switch rows[indexPath.row] {
+        case .header:
+            comment.hasParentComment() ? navigateToParentComment() : navigateToPost()
+
+        case .replyIndicator:
+            // TODO: Navigate to the comment reply.
+            break
```
dvdchr:
```suggestion
            navigateToReplyComment()
```

I'm not sure why this is reverted 🤔 , but since this PR is based on #17409, this should navigate to the reply comment properly.

dvdchr (changes requested):
Tested and things work fine albeit with some minor 🐛 issues! I've left some comments for them, but since one issue impacts functionality, I'll request changes for now.

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
ScoutHarris: Well you're assuming I didn't do something dumb. Which clearly I did. 😆  Thank you!

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
ScoutHarris: Nice catch! Will do!

ScoutHarris:
Hey @dvdchr . Thanks for catching those issues! They should be fixed now. Ready for another review please.

dvdchr (approved):
LGTM! :shipit:

ScoutHarris:
Thanks @dvdchr !

PR #17417, opened 2021-11-03
Merge editorialized release notes for 18.6 into `develop`

mokagio:
This dedicated PR will ensure we send the copy out for translation ASAP.


WordPress/Resources/release_notes.txt
```diff
@@ -1,11 +1,8 @@
...
-* [*] Block editor: Fix Embed Block loading glitch with resolver resolution approach [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4146]
-* [*] Block editor: Fixed an issue where the Help screens may not respect an iOS device's notch. [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4110]
-* [**] Block editor: Block inserter indicates newly available block types [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4047]
-* [*] Block editor: Add support for the Mark HTML tag [https://github.com/wordpress-mobile/gutenberg-mobile/pull/4162]
-* [*] Stats Insights: HTML tags no longer display in post titles. [#17380]
-
+- We've made it easier to see when there are new blocks available for use on your site
+- You can now highlight text using the <mark> HTML tag, as well as change the text and background color of the Search block button. Shiny!
+- We've hidden HTML tags in post titles when you're checking out your stats.
+- We've improved Reader Discovery – showing you new things to read, while lending a helping hand to sites with low traffic. Win-win.
```
mokagio, line 4: Notice this is not part of the Jetpack notes because there's not Reader there.

WordPress/Resources/release_notes.txt
```diff
@@ -1,11 +1,8 @@
...
+- If you're part of a comment thread, you can keep track of it through in-app notifications, not just your email inbox.
+- Speaking of comments, we heard that some might be missing when viewing comments from the Reader. Don't worry – we found and fixed the problem.
```
mokagio, line 6: This is not part of the Jetpack notes either because there's not Reader there.

mokagio (commented):
- [x] `.txt` and associated `.po` updated

oguzkocer (approved):

WordPress/Jetpack/Resources/AppStoreStrings.po
```diff
@@ -38,15 +38,14 @@ msgctxt "app_store_keywords"
...
 msgid ""
-"This update makes it even easier to embed external content on your sites! Introducing the new dedicated embed blocks for Facebook, Instagram, Loom, and Smartframe. Plus, the editor automatically detects embeddable URLs you paste into an empty paragraph. Alongside these new features, we fixed an issue where some embedded URLs could not be edited and adjusted the preview cut-off so that you’ll always see the full embedded content.\n"
-"\n"
-"Speaking of blocks, you can now customize the text and background color in the Pullquote and Preformatted blocks.\n"
-"\n"
-"Need help reaching more readers? The new Publicize and Blogging Reminder nudges will help you stick to a consistent publishing schedule.\n"
-"\n"
-"Finally, we improved support for larger text sizes when choosing a homepage or page layout, fixed the site comments list not refreshing on certain occasions, and polished the pre-publishing bottom sheet header.\n"
+"- We've made it easier to see when there are new blocks available for use on your site\n"
+"- You can now highlight text using the <mark> HTML tag, as well as change the text and background color of the Search block button. Shiny!\n"
```
AliSoftware, line 44: @mokagio Thought: you might want to double-check at submission time that the use of unescaped `<mark>` didn't generate any bug at any point in our tooling pipeline. I actually have no idea if there is a risk that somewhere in the tooling this gets interpreted as a tag and maybe even removed, or if we already handle this properly without any issue?

I'm hoping this won't cause any issue in practice, but I figured it would be worth checking on ASC (after those will be translated and you'll run `deliver`) that the `<mark>` is still there (and without any odd escaping sequence erroneously introduced along the way)

mokagio: Good point. Will do that.

PR #17426, opened 2021-11-04
Onboarding Improvements: Quickstart prompt UI

### LINKED ISSUE #17386: Onboarding Improvements: Quick Start prompt
## Description

Update the Quick Start prompt UI.

Ref: pbArwn-32R-p2

Screen Shot 2021-10-28 at 16 42 09

## Tasks
- [x] Update UI [#17426](https://github.com/wordpress-mobile/WordPress-iOS/pull/17426)
- [x] CTA: Show me around -> dismiss then present Quick Start checklist [#17426](https://github.com/wordpress-mobile/WordPress-iOS/pull/17426)
- [x] CTA: No thanks -> dismiss and prevent Quick Start prompt from showing in the future for that blog [#17426](https://github.com/wordpress-mobile/WordPress-iOS/pull/17426)
- [x] Fine tune design [per feedback ](https://github.com/wordpress-mobile/WordPress-iOS/pull/17449#pullrequestreview-804802387) from @osullivanchris [#17485](https://github.com/wordpress-mobile/WordPress-iOS/pull/17485)

momo-ozawa:
Fixes #17386 

Ref: pbArwn-32R-p2

## Description
- Updated the Quick Start prompt UI
- Added the CTA: Show me around -> dismiss then present Quick Start checklist
- Added the CTA: No thanks -> dismiss and prevent Quick Start prompt from showing in the future for that blog

## Notes
- There are some obsolete files we should be able to delete, once this PR gets approved. I'll make a separate PR.
- UI tests will be addressed in a different PR

### CTA: Show me around

1. Log out then log in
2. Select a site from the Epilogue
3. ✅ The Quick Start prompt with the updated UI is displayed
4. Select Show me around
5. ✅ The prompt is dismissed, then after a short delay the Quick Start Customize checklist is displayed
6. Tap the X button
7. ✅ The checklist is dismissed, and the site selected is the is the site you chose in step 2

<details>
 <summary>Click here for video</summary>

https://user-images.githubusercontent.com/6711616/140362558-1fecdd08-18ac-4d95-b092-bc3efc7335db.mp4

</details>

### CTA: No thanks

1. Log out then log in
2. Select a site from the Epilogue
3. ✅ The Quick Start prompt with the updated UI is displayed
4. Select No thanks
5. ✅ The prompt is dismissed, and the site selected is the is the site you chose in step 2
6. Repeat steps 1 and 2
7. ✅ Notice that this time around the Quick Start prompt was not displayed

<details>
 <summary>Click here for video</summary>

https://user-images.githubusercontent.com/6711616/140362717-b2888c9d-5e92-48c2-9466-4f79ae9ee105.mp4

</details>

## Regression Notes
Quick Start

Tested the above steps

N/A

## PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -101,8 +101,7 @@ final class QuickStartPromptViewController: UIViewController {
     // MARK: - IBAction
 
     @IBAction private func showMeAroundButtonTapped(_ sender: Any) {
-        onDismiss?(blog)
-        navigationController?.dismiss(animated: true, completion: { [weak self] in
+        dismissPrompt { [weak self] in
 
             DispatchQueue.main.asyncAfter(deadline: .now() + Constants.quickStartDelay) {
                 guard let self = self else {
@@ -110,12 +109,16 @@ final class QuickStartPromptViewController: UIViewController {
                 }
                 QuickStartTourGuide.shared.setup(for: self.blog)
             }
-        })
+        }
     }
 
     @IBAction private func noThanksButtonTapped(_ sender: Any) {
+        dismissPrompt()
+    }
+
+    private func dismissPrompt(completion: (() -> Void)? = nil) {
         onDismiss?(blog)
-        navigationController?.dismiss(animated: true)
+        navigationController?.dismiss(animated: true, completion: completion)
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -101,9 +101,16 @@ final class QuickStartPromptViewController: UIViewController {
     // MARK: - IBAction
 
     @IBAction private func showMeAroundButtonTapped(_ sender: Any) {
-        let controller = QuickStartChecklistViewController(blog: blog, type: .customize)
-        let navigationController = UINavigationController(rootViewController: controller)
-        present(navigationController, animated: true)
+        onDismiss?(blog)
+        navigationController?.dismiss(animated: true, completion: { [weak self] in
+
+            DispatchQueue.main.asyncAfter(deadline: .now() + Constants.quickStartDelay) {
+                guard let self = self else {
+                    return
+                }
+                QuickStartTourGuide.shared.setup(for: self.blog)
+            }
+        })
     }
 
     @IBAction private func noThanksButtonTapped(_ sender: Any) {
@@ -119,5 +126,9 @@ extension QuickStartPromptViewController {
         static let promptDescription = NSLocalizedString("Learn the basics with a quick walk through.", comment: "Description for a prompt asking if users want to try out the quick start checklist.")
         static let showMeAroundButtonTitle = NSLocalizedString("Show me around", comment: "Button title. When tapped, the quick start checklist will be shown.")
         static let noThanksButtonTitle = NSLocalizedString("No thanks", comment: "Button title. When tapped, the quick start checklist will not be shown, and the prompt will be dismissed.")
+    }
+
+    private enum Constants {
+        static let quickStartDelay: DispatchTimeInterval = DispatchTimeInterval.milliseconds(1000)
     }
 }
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -21,6 +21,10 @@ final class QuickStartPromptViewController: UIViewController {
 
     private let blog: Blog
 
+    /// Closure to be executed upon dismissal.
+    ///
+    var onDismiss: ((Blog) -> Void)?
+
     // MARK: - Init
 
     init(blog: Blog) {
@@ -97,11 +101,14 @@ final class QuickStartPromptViewController: UIViewController {
     // MARK: - IBAction
 
     @IBAction private func showMeAroundButtonTapped(_ sender: Any) {
-        dismiss(animated: true)
+        let controller = QuickStartChecklistViewController(blog: blog, type: .customize)
+        let navigationController = UINavigationController(rootViewController: controller)
+        present(navigationController, animated: true)
     }
 
     @IBAction private func noThanksButtonTapped(_ sender: Any) {
-        dismiss(animated: true)
+        onDismiss?(blog)
+        navigationController?.dismiss(animated: true)
     }
 }
 
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -113,6 +113,7 @@ final class QuickStartPromptViewController: UIViewController {
     }
 
     @IBAction private func noThanksButtonTapped(_ sender: Any) {
+        UserDefaults.standard.setQuickStartWasDismissed(true, for: blog)
         dismissPrompt()
     }
 
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -1,0 +1,116 @@
+import UIKit
+
+final class QuickStartPromptViewController: UIViewController {
+
+    // MARK: - IBOutlets
+
+    /// Site info
+    @IBOutlet private weak var siteIconView: UIImageView!
+    @IBOutlet private weak var siteTitleLabel: UILabel!
+    @IBOutlet private weak var siteDescriptionLabel: UILabel!
+
+    /// Prompt info
+    @IBOutlet private weak var promptTitleLabel: UILabel!
+    @IBOutlet private weak var promptDescriptionLabel: UILabel!
+
+    /// Buttons
+    @IBOutlet private weak var showMeAroundButton: FancyButton!
+    @IBOutlet private weak var noThanksButton: FancyButton!
+
+    // MARK: - Properties
+
+    private let blog: Blog
+
+    // MARK: - Init
+
+    init(blog: Blog) {
+        self.blog = blog
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Lifecycle
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        applyStyles()
+        setup()
+    }
+
+    // MARK: - Styling
+
+    private func applyStyles() {
+        siteTitleLabel.numberOfLines = 0
+        siteTitleLabel.font = WPStyleGuide.fontForTextStyle(.subheadline)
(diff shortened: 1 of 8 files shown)


momo-ozawa:
@osullivanchris 👋 

I added a one second delay before showing the Quick Start Customize checklist, per our convo. 
p1636025155003200/1636024315.001400-slack-C027K4MNPGQ

I added some videos up above, let me know what you think.

Installable build can be downloaded from [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/17426#issuecomment-961168661), once it's ready 😄

osullivanchris (commented):
Really nice stuff @momo-ozawa 

- the transition from 'show me around' to the My Site feels totally fine in my hand actually. The fade is good. So I'm happy to go with this.
- This is both pedantic and subjective, but maybe the delay is a little too long. Maybe 0.5sec would be enough to get a glimpse, but be quickly moved on to the tasks? Another option would be not to autoprompt the tasks list but seems fine to keep it as its existing behaviour.
- Small UI detail while I'm here. I think maybe there should be less space between 'Choose your own site" and "you can switch sites any time". Also, maybe these text styles should be exactly the same as the title and url of the sites below. Currently they both look 1 size smaller than their counter parts in the title and url of the sites in the rows below. WDYT?

momo-ozawa:
> the transition from 'show me around' to the My Site feels totally fine in my hand actually. The fade is good. So I'm happy to go with this.

🥳 

> This is both pedantic and subjective, but maybe the delay is a little too long. Maybe 0.5sec would be enough to get a glimpse, but be quickly moved on to the tasks? Another option would be not to autoprompt the tasks list but seems fine to keep it as its existing behaviour.

I debated between 0.5 secs and 1 sec, actually! Here's what 0.5 secs would look like 

Uploading Simulator Screen Recording - iPhone 12 Pro - 2021-11-04 at 16.47.16.mp4…

> Small UI detail while I'm here. I think maybe there should be less space between 'Choose your own site" and "you can switch sites any time". Also, maybe these text styles should be exactly the same as the title and url of the sites below. Currently they both look 1 size smaller than their counter parts in the title and url of the sites in the rows below. WDYT?

Added as a subtask for the Epilogue screen, I'll update these details in another PR 💪

momo-ozawa:
@osullivanchris Thanks for the feedback!

> the transition from 'show me around' to the My Site feels totally fine in my hand actually. The fade is good. So I'm happy to go with this.

🥳 

> This is both pedantic and subjective, but maybe the delay is a little too long. Maybe 0.5sec would be enough to get a glimpse, but be quickly moved on to the tasks? Another option would be not to autoprompt the tasks list but seems fine to keep it as its existing behaviour.

I debated between 0.5 secs and 1 sec, actually! Here's what 0.5 secs would look like 


https://user-images.githubusercontent.com/6711616/140383276-e76a8f94-f460-42e9-885e-45d4e9c4532a.mp4


> Small UI detail while I'm here. I think maybe there should be less space between 'Choose your own site" and "you can switch sites any time". Also, maybe these text styles should be exactly the same as the title and url of the sites below. Currently they both look 1 size smaller than their counter parts in the title and url of the sites in the rows below. WDYT?

Added as a subtask for the Epilogue screen, I'll update these details in another PR 💪

osullivanchris:
> I debated between 0.5 secs and 1 sec, actually! Here's what 0.5 secs would look like

I think I prefer 0.5. Interested to get another pair of eyes now that we've bounced back and forth a couple times. Maybe @leandroalonso 

> Added as a subtask for the Epilogue screen, I'll update these details in another PR 💪

sounds good, thanks!!

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+    @IBOutlet private weak var showMeAroundButton: FancyButton!
+    @IBOutlet private weak var noThanksButton: FancyButton!
+
+    // MARK: - Properties
+
+    private let blog: Blog
+
+    /// Closure to be executed upon dismissal.
+    ///
+    var onDismiss: ((Blog) -> Void)?
```
leandroalonso, line 27: I checked on Instruments and it seems that `QuickStartPromptViewController` is not being deallocated after its dismissal.

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    private enum Strings {
+        static let promptTitle = NSLocalizedString("Want a little help managing this site with the app?", comment: "Title for a prompt asking if users want to try out the quick start checklist.")
+        static let promptDescription = NSLocalizedString("Learn the basics with a quick walk through.", comment: "Description for a prompt asking if users want to try out the quick start checklist.")
+        static let showMeAroundButtonTitle = NSLocalizedString("Show me around", comment: "Button title. When tapped, the quick start checklist will be shown.")
+        static let noThanksButtonTitle = NSLocalizedString("No thanks", comment: "Button title. When tapped, the quick start checklist will not be shown, and the prompt will be dismissed.")
+    }
+
+    private enum Constants {
+        static let quickStartDelay: DispatchTimeInterval = DispatchTimeInterval.milliseconds(1000)
```
leandroalonso: [nit]

```suggestion
        static let quickStartDelay: DispatchTimeInterval = .milliseconds(1000)
```

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+}
+
+extension UserDefaults {
+
+    func quickStartWasDismissed(for blog: Blog) -> Bool {
```
leandroalonso: I have mixed feelings about extending `UserDefaults` to account for quick start stuff.

Do you think it would be too complicated to create a new entity responsible for that and inject where is needed?

leandroalonso (commented):
@momo-ozawa On iPad we're not following the `readableContentGuide`:



@osullivanchris I think I also prefer `0.5` too.

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
momo-ozawa: @leandroalonso 👋 

> Do you think it would be too complicated to create a new entity responsible for that and inject where is needed?

Do you mean like create a QuickStartDefaultsManager / Helper that handles everything? I feel that may be a bit overengineered, but I'm curious why you have mixed feelings about extending `UserDefaults` -- maybe I'm overlooking something -- can you elaborate?

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
momo-ozawa: Fixed! d3c5d75

momo-ozawa:
> On iPad we're not following the readableContentGuide

Fixed! 88695bd

Simulator Screen Shot - iPad Air (4th generation) - 2021-11-05 at 11 38 29

momo-ozawa:
Hi @diegoreymendez 👋 
Tagging you for a review since Leandro is AFK for the week, and GH suggested you as a reviewer.
Lmk if you have any q's about this feature. Thanks in advance! 🙇‍♀️

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
diegoreymendez: I agree with @leandroalonso here.

The issue with accessing singletons (like `UserDefaults`) directly throughout the code is that the code becomes much harder to unit test because [the method calls are not deterministic](https://en.wikipedia.org/wiki/Deterministic_algorithm), requiring you to mock the singleton (`UserDefaults.standard`) which isn't great and sometimes can bring several other complications if running several tests in parallel using the same mocked singleton.

An alternative to this can be dependency injection, but most of the times you don't really need a new class.

As an example, if I'm implementing `SomeClass` and I need it to use a singleton, I try to use a pattern like this:

```swift
    class SomeClass {
        private let someKey = "someKey"
        private let userDefaults: UserDefaults

        init(userDefaults: UserDefaults = .standard) {
            self.userDefaults = userDefaults
        }

        // Mark - User Defaults Storage

        func getSomeDate() -> Date? {
            userDefaults.object(forKey: birthday) as? Date
        }

        func setSomeDate(_ date: Date) {
            userDefaults.set(date, forKey: someKey)
        }
    }
```

This means that when I'm unit-testing this class I can do:

```swift
let someObj = SomeClass(userDefaults: myMockUserDefaults)

// ...test away
```

DISCLAIMER: I'm being a bit purist here and I've failed to do this myself more than once.  I still think it's worth it to try and use this approach.

diegoreymendez: The issue with accessing singletons (like `UserDefaults`) directly throughout the code is that the code becomes much harder to unit test.

An alternative to this can be a dedicated class with dependency injection.

And It's not that different from what you already have, so adjusting should be easy / cheap.

As an example:

```swift
class QuickStartDefaultsManager {
    private let userDefaults: UserDefaults

    init(userDefaults: UserDefaults = .standard) {
        self.userDefaults = userDefaults
    }

    // Mark - User Defaults Storage
    func quickStartWasDismissed(for blog: Blog) -> Bool {
        guard let key = quickStartWasDismissedKey(for: blog) else {
            return false
        }
        return bool(forKey: key)
    }

    func setQuickStartWasDismissed(_ value: Bool, for blog: Blog) {
        guard let key = quickStartWasDismissedKey(for: blog) else {
            return
        }
        set(value, forKey: key)
    }

    private func quickStartWasDismissedKey(for blog: Blog) -> String? {
        guard let siteID = blog.dotComID?.intValue else {
            return nil
        }
        return "QuickStartWasDismissed-\(siteID)"
    }
}
```

This means that we can unit-test this class by mocking the user defaults:

```swift
let quickStartDefaultsManager = QuickStartDefaultsManager(userDefaults: myMockUserDefaults)

// ...test away
```

diegoreymendez (commented):
I added some comments for the code.

Additionally I have a question: is it expected that the prompt is only shown once per site?  Logging in again and selecting the same site won't show the prompt.  Only reinstalling the App made it show again.

I imagine this is expected but since the PR description doesn't mention this, I wanted to double check.

momo-ozawa:
Hi @diegoreymendez 👋 
Thanks for reviewing. Ready for a re-review!

> Additionally I have a question: is it expected that the prompt is only shown once per site? Logging in again and selecting the same site won't show the prompt. Only reinstalling the App made it show again.

If the user taps the No thanks CTA, then the prompt won't be shown again for that site, unless they do a fresh install of  WPiOS.

If the user taps the Show me around CTA, then the prompt will be shown again next time they log in.

Ref: pbArwn-2rX-p2

> If the user says no, don’t show it again for that site, save locally. If they delete the app, its erased and on a fresh install they would see the prompt again.

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
momo-ozawa: Fixed! 

fe54651
94ba763 
c7c3c26

diegoreymendez (approved):
Code looks great, works as described, and I really like that you took the opportunity to add unit tests!

Good work!

PR #17428, opened 2021-11-04
Comment Details: show reply view when Reply button tapped.

ScoutHarris:
Ref: #17412 

When the `Reply` button is tapped, the Reply view is shown and becomes first responder (i.e. the keyboard appears).

Notes:
- Replies are not created yet.
- UI updates are not done yet.

To test:
- Enable `newCommentDetail`.
- Go to My Site > Comments > comment details.
- Tap `Reply`.
- Verify the reply view and the keyboard is shown.
- Tap `Reply` again. Verify nothing happens.
- Enter some text with returns.
- Verify the view grows to about 4 lines, then starts scrolling.

| init | reply | entered_text |
|--------|-------|-------|

## Regression Notes
N/A. Feature is incomplete.

N/A. Feature is incomplete.

N/A. Feature is incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
@@ -12,6 +12,10 @@ class CommentDetailViewController: UIViewController {
     private let containerStackView = UIStackView()
     private let tableView = UITableView(frame: .zero, style: .plain)
 
+    private var replyTextView: ReplyTextView?
+    private var keyboardManager: KeyboardDismissHelper?
+    private var dismissKeyboardTapGesture = UITapGestureRecognizer()
+
     @objc weak var delegate: CommentDetailsDelegate?
     private var comment: Comment
     private var isLastInList: Bool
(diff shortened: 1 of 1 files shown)


dvdchr (approved):
LGTM! :shipit: 

I've just a minor comment on the reply text view appearance animation since I _think_ it looked a little odd. I wonder if it would look better if it appeared from the bottom instead?

https://user-images.githubusercontent.com/1299411/140735594-e35c886d-d922-43af-8413-cf2609deab9b.mov

Anyways, this is just a minor subjective suggestion of mine – so I'll just leave it up to you. 🙂

ScoutHarris:
> reply text view appearance animation

Ugh. I thought I fixed that. Clearly I didn't notice it still jumped. I'll look at it separately. Thanks!

ScoutHarris:
Hey @dvdchr .

> I wonder if it would look better if it appeared from the bottom instead?

The reply view isn't added with animation. It's added to the stack view as hidden in `viewDidLoad`. When `Reply` is tapped, it is then simply unhidden. 

I think the problem is more to do with the `KeyboardDismissHelper` used to show/hide the keyboard. You're example is the keyboard attempting to show as the Reply view becomes first responder, but is then immediately dismissed because (I'm assuming) you have it disabled in the simulator. A more real world scenario is, if you enable the software keyboard, then tap to dismiss it, you'll see that gray area again as the keyboard slides down. Turns out, this isn't new. In the old details view, the same thing happens. It's just not as noticeable because the background is white, and somewhere here I have a gray background. I'm not sure how much trouble it will be to fix the root cause, but I can at least try to make it look better.

PR #17432, opened 2021-11-05
Comment Reply: show mention suggestions

ScoutHarris:
Ref: #17412 
Depends on: #17428

When @ is entered in the Reply field, the mention suggestions list is now shown.

Notes:
- The suggestions list is cut off in the fullscreen view. This is legacy, so I'll look at it separately.
- Replies are not created yet.
- UI updates are not done yet.

To test:
Sorry for the lack of screenshots, but there's just too much PII to block out. However, you can compare this with replying in the old comment details. It should look and function the same.

- Enable `newCommentDetail`.
- Go to My Site > Comments > comment details.
- Tap `Reply`.
- On a site that does not support @ mentions:
  - Hack tip: I couldn't find a site that didn't support @ mentions. So you can return `false` in `shouldShowSuggestions` in `CommentDetailViewController` to verify this scenario.
  - Enter @ in the reply field.
  - Verify nothing is displayed above the reply field.
- On a site that does support @ mentions:
  - Enter @ in the reply field.
  - Verify a list of users is displayed above the reply field.
  - Tap the up arrow to enter full screen.
  - Enter @ in the reply field.
  - Verify a list of users is displayed below the text.

## Regression Notes
N/A. Feature incomplete.

N/A. Feature incomplete.

N/A. Feature incomplete.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
@@ -948,7 +948,7 @@ extension CommentDetailViewController: SuggestionsTableViewDelegate {
 
     func suggestionsTableView(_ suggestionsTableView: SuggestionsTableView, didSelectSuggestion suggestion: String?, forSearchText text: String) {
         replyTextView?.replaceTextAtCaret(text as NSString?, withText: suggestion)
-        suggestionsTableView.showSuggestions(forWord: String())
+        suggestionsTableView.hideSuggestions()
     }
 
 }
--- a/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
@@ -12,7 +12,9 @@ class CommentDetailViewController: UIViewController {
     private let containerStackView = UIStackView()
     private let tableView = UITableView(frame: .zero, style: .plain)
 
+    // Reply properties
     private var replyTextView: ReplyTextView?
+    private var suggestionsTableView: SuggestionsTableView?
     private var keyboardManager: KeyboardDismissHelper?
     private var dismissKeyboardTapGesture = UITapGestureRecognizer()
 
@@ -25,6 +27,10 @@ class CommentDetailViewController: UIViewController {
 
     private var viewIsVisible: Bool {
         return navigationController?.visibleViewController == self
+    }
+
+    private var siteID: NSNumber? {
+        return comment.blog?.dotComID
     }
 
     private var replyID: Int32 {
@@ -205,6 +211,7 @@ class CommentDetailViewController: UIViewController {
         configureView()
         configureReplyView()
         setupKeyboardManager()
+        configureSuggestionsView()
         configureNavigationBar()
         configureTable()
         configureRows()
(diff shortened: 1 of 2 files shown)


ScoutHarris:
> The suggestions list is cut off in the fullscreen view. This is legacy, so I'll look at it separately.

Turns out this isn't a bug per se. 
- The number of rows in the suggestion table is hard coded (and is intended to be). In this case, it's set to 3.
- In fullscreen the suggestions could appear above or below the caret, meaning it may not have room for more rows depending on the location. (Yes, I tried. It broke. 😄 ) 

While it looks strange, it is functioning as intended. The 3 row table is scrollable, so it's not prohibitive, just annoying. To avoid a major refactor, I'll leave it as-is. In addition, `SuggestionsTableView` is used by the Gutenberg editor, so any changes could impact that.

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -884,4 +892,63 @@ private extension CommentDetailViewController {
         // TODO: create reply
     }
 
+    func configureSuggestionsView() {
+        guard shouldShowSuggestions,
+              let siteID = siteID,
+        let replyTextView = replyTextView else {
```
dvdchr: minor nitpick: maybe add one or two indentation levels here so this line aligns with `let siteID` above it?

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
```diff
@@ -884,4 +892,63 @@ private extension CommentDetailViewController {
...
+
+}
+
+// MARK: - SuggestionsTableViewDelegate
+
+extension CommentDetailViewController: SuggestionsTableViewDelegate {
+
+    func suggestionsTableView(_ suggestionsTableView: SuggestionsTableView, didSelectSuggestion suggestion: String?, forSearchText text: String) {
+        replyTextView?.replaceTextAtCaret(text as NSString?, withText: suggestion)
+        suggestionsTableView.showSuggestions(forWord: String())
```
dvdchr: What do you think about calling `hideSuggestions()` here instead? At first, I assumed that calling `showSuggestions` with an empty string means showing a list of all mentionable users, but turns out it's just to dismiss the suggestion view. But then I checked that `hideSuggestions()` is actually doing the same:
```Objective-C
- (void)hideSuggestions
{
    [self showSuggestionsForWord:@""];
}
```

So maybe we could use `hideSuggestions()` instead for clarity? 😄

dvdchr (approved):
Tested and it worked as expected. Left a few minor comments, but overall LGTM! :shipit:

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
ScoutHarris: I swear I did ctrl+I several times and it didn't change! And of course it does _now_.  I'll fix it.

WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
ScoutHarris: Ha! That's what I get for copy + paste + assuming it was right to start with. Can do! Thanks.

PR #17440, opened 2021-11-09
Comment Detail: Show contents in a web view for self-hosted sites

dvdchr:
Refs #17087 

This fixes an issue where tapping the comment header on the new Comment Detail does nothing when the comment detail is displaying a comment that is a reply to another comment (e.g. a reply comment). The reason is mainly that because the `siteID` is never `nil`. Instead, it returns `0` for self-hosted non-Jetpack sites since the value is stored with an `Int32` type in the data model.

The fix adds a check for the comment's blog if it supports REST API, and instead of falling back to `navigateToPost()`, it should now show a web view displaying the parent comment.

## To Test

- Log in to a self-hosted site where you have a reply comment somewhere on the site.
- Enable the `newCommentDetail` feature flag.
- Go to My Site > Comments, and select any reply comment.
- Tap on the comment header.
- Verify that the web view should now be shown, displaying the content of the parent comment.

## Regression Notes
n/a. Feature is still unreleased.

n/a. Feature is still unreleased.

n/a. Feature is still unreleased.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
+++ b/WordPress/Classes/ViewRelated/Comments/CommentDetailViewController.swift
@@ -489,10 +489,13 @@ private extension CommentDetailViewController {
     // Shows the comment thread with the parent comment highlighted.
     func navigateToParentComment() {
         guard let parentComment = parentComment,
-              let siteID = siteID else {
-                  navigateToPost()
-                  return
-              }
+              let siteID = siteID,
+              let blog = comment.blog,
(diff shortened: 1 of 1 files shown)


dvdchr:
I also noticed that while being logged in as the site administrator, the moderation bar is not showing for any comments. while in the old comment detail, the moderation bar is shown.

EDIT: I'll address this issue later in a separate PR. 🙂

ScoutHarris (approved):
LGTM!

:shipit:

PR #17436, opened 2021-11-08
Re-enable EditorGutenbergTests temporarily skipped because of missing Blogging Reminders support

tiagomar:
Editor Gutenberg UI tests [were skipped by this PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/16798) as they were failing since the introduction of the new blogging reminders feature [as described in this issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/16797).

This PR adds support for the blogging reminders alert dismissal and re-enables the skipped tests.

Commits
[Re-enables Gutenberg UI tests by adding support for blogging reminders alert dismissal](https://github.com/wordpress-mobile/WordPress-iOS/pull/17436/commits/d4c602c36db5959ae41b19eb0aefbbc98e230433)
[Adds support for blogging reminders dismissal on iPads](https://github.com/wordpress-mobile/WordPress-iOS/pull/17436/commits/ac2e7c27055c2122635256ec19d95b359e7b6719)

## To test
EditorGutenbergTests tests must pass on CI.

Testing "WordPressUITests" scheme locally must pass.

## Regression Notes
1. Potential unintended areas of impact
None. The changes would only impact  EditorGutenbergTests.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
N/A

3. What automated tests I added (or what prevented me from doing so)
N/A

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes. **N/A**
- [x] I have considered adding accessibility improvements for my changes. **N/A**
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. **N/A**

PR #17455, opened 2021-11-11
Replay "Re-enable `EditorGutenberTests` (#17436) on top of `ScreenObject` updates

mokagio:
See [this comment](https://github.com/wordpress-mobile/WordPress-iOS/pull/17436#issuecomment-966698816).

~~**Update:** The green CI we're seeing is a _false positive_! See https://github.com/wordpress-mobile/WordPress-iOS/pull/17456.~~ The Xcode 13 UI tests CI fix has now been merged.

### DIFF
--- a/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
@@ -107,7 +107,14 @@ public class BlockEditorScreen: ScreenObject {
 
     public func publish() throws -> EditorNoticeComponent {
         let publishButton = app.buttons["Publish"]
-        publishButton.tap()
+        let publishNowButton = app.buttons["Publish Now"]
+        var tries = 0
+        // This loop to check for Publish Now Button is an attempt to confirm that the publishButton.tap() call took effect.
+        // The tests would fail sometimes in the pipeline with no apparent reason.
+        repeat {
+            publishButton.tap()
+            tries += 1
+        } while !publishNowButton.waitForIsHittable(timeout: 3) && tries <= 3
         try confirmPublish()
 
         return try EditorNoticeComponent(withNotice: "Post published", andAction: "View")
--- a/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
@@ -150,6 +150,17 @@ public class BlockEditorScreen: ScreenObject {
         } else {
             let publishNowButton = app.buttons["Publish Now"]
             publishNowButton.tap()
+            dismissBloggingRemindersAlertIfNeeded()
+        }
+    }
+
+    public func dismissBloggingRemindersAlertIfNeeded() {
+        let bloggingRemindersAlertIsLoaded = app.buttons["Set reminders"].waitForExistence(timeout: 3)
+
+        if bloggingRemindersAlertIsLoaded {
+            let dismissBloggingRemindersAlertButton = app.buttons.element(boundBy: 0)
+
+            dismissBloggingRemindersAlertButton.tap()
         }
     }
 
--- a/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
@@ -130,8 +130,10 @@ public class BlockEditorScreen: ScreenObject {
     }
 
     private func addBlock(_ blockLabel: String) {
+        let blockButton = XCUIApplication().buttons[blockLabel]
         addBlockButton.tap()
-        XCUIApplication().buttons[blockLabel].tap()
+        guard blockButton.waitForIsHittable(timeout: 3) else { return }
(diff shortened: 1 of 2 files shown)


mokagio:
@tiagomar good news, sort of. As you can see by inspecting the CI logs, this branch (which doesn't include your `repeat` fix) is getting some flakyness issue.

Interestingly, there's also a [fail on iPhone in CircleCI](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-iOS/25596/workflows/41abd508-44bb-44a6-8f11-e2ec7b2d1d34/jobs/60319?invite=true#step-107-382) due to an error I hadn't seen before here:

```
Failed to get matching snapshot: No matches found for Elements matching predicate '"Image block" IN identifiers' from input {(
```

If you don't get to it first, I'll cherry pick the `repeat` commit on this branch tomorrow and see how CI behaves with that.

tiagomar:
Hey @mokagio, I think this is good news =P

Now we know that the issue is not device/simulator specific, not element specific and not test case specific. Maybe performance or interruptions...

I won't just cherry pick the `repeat` commit because it will need a couple more lines after the `ScreenObject` work. I'll make a new commit.

tiagomar:
The good news this time is that, after running the tests a few times, I haven't seen the previous `Publish Now` button issue. 🎉

The not so good news is that the `Image block` issue has nothing to do with the previous one. Also, it happens intermittently for both iPhone and iPad. I couldn't reproduce it locally, as usual, so I added a wait for the button is hittable before tapping it to check how it goes in pipeline.The failure rate is close to 50%.

Improving reliability on these tests could be a project on its own. =P

WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -123,8 +130,10 @@ public class BlockEditorScreen: ScreenObject {
     }
 
     private func addBlock(_ blockLabel: String) {
+        let blockButton = XCUIApplication().buttons[blockLabel]
```
mokagio:
```suggestion
        let blockButton = app.buttons[blockLabel]
```

`ScreenObject` as an [`app`](https://github.com/Automattic/ScreenObject/blob/2d65beae47cdcaaf21703ce1b321f382fe0d0730/Sources/ScreenObject/ScreenObject.swift#L21-L23) property which exposes its [required `XCUIApplication` init parameter](https://github.com/Automattic/ScreenObject/blob/2d65beae47cdcaaf21703ce1b321f382fe0d0730/Sources/ScreenObject/ScreenObject.swift#L41-L45). 

99% of the time, `app` and `XCUIApplication()` are going to be the same. The real advantage is in not having to re-instantiate the object.

WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -123,8 +130,10 @@ public class BlockEditorScreen: ScreenObject {
...
         addBlockButton.tap()
-        XCUIApplication().buttons[blockLabel].tap()
+        guard blockButton.waitForIsHittable(timeout: 3) else { return }
```
mokagio: Would we want the test to fail at this point? That is, we're waiting for the button to be hittable to avoid a failure, but if it is not hittable, then the app is not in the expected state and we should fail the test as soon as possible.

```suggestion
        XCTAssertTrue(blockButton.waitForIsHittable(timeout: 3))
```

Or am I missing something?

tiagomar: Yeah, you're right.

tiagomar:
Hey @mokagio, what can we do with this PR? I didn't want to approve my own changes but I can do it if say it is good to go.😅

jostnes (approved):
Looks good. Tested this locally a few times, and the `EditorGutenbergTests` set is green consistently 🎉  

Nothing to change for this PR but one observation, I see that timeout value is being set independently now. Not too sure how timeout works in framework, yet but we could check if we set a default timeout for `waitFor...` so we don't have to do that every time.

tiagomar:
Yeah, we should definitely set a default timeout and maybe a couple (long, short) more. We would probably like to keep them in [XCUITestHelpers](https://github.com/Automattic/XCUITestHelpers/).

mokagio:
Sorry @tiagomar for not following up on this PR in a timely manner. Good call on merging it, cc @jostnes  👍 

Regarding the timeouts: 

> I see that timeout value is being set independently now. Not too sure how timeout works in framework, yet but we could check if we set a default timeout for waitFor... so we don't have to do that every time.

I'm not sure whether `XCTest` allows for setting a default global timeout. But we can definitely do that in all our custom methods at the XCUITestHelpers and ScreenObject packages level.

I like the idea of having different timeout levels, to account for different response-times in the UI. Something like

```swift
enum Timeouts {
    static let short: TimeInterval = 1
    static let `default`: TimeInterval = 3
    static let long: TimeInterval = 10
}
```

Note: I like to use `enum` + `static let` in these instances so that it's clear that the type only holds constants (you cannot call `Timeouts()`) while also avoiding the awkward `.rawValue` call we'd have using the cases of an `enum Timeout: TimeInterval`.

PR #17233, opened 2021-09-28
iOS 15: Change Buildkite image to Xcode 13

leandroalonso:
This PR changes the Buildkit image to use Xcode 13. This should allow the build to stop failing.

Notice that this change is targeting a feature branch, not `develop`.

### To test:

1. Make sure Buildkite build is green

## Regression Notes
1. Potential unintended areas of impact
n/a

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a

3. What automated tests I added (or what prevented me from doing so)
n/a

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17456, opened 2021-11-12
Fix UI tests not running since Xcode 13 migration

mokagio:
When we worked on [migrating to Xcode 13](https://github.com/wordpress-mobile/WordPress-iOS/pull/17233), we encountered weird looking failures running the tests via Fastlane. @leandroalonso did some [great investigation](https://github.com/wordpress-mobile/WordPress-iOS/pull/17233#issuecomment-954897098) and worked around them by introducing a filter for the tests to run (see the linked comments for more details).

What we didn't realize at the time was that those filters resulted in _every UI test being skipped_, as you can see by how fast the UI tests run and by the message they printed in the logs:

Screen Shot 2021-11-12 at 3 42 13 pm

_UI tests running in ~2 minutes when usually they run in ~20 minutes_

Screen Shot 2021-11-12 at 3 42 28 pm

_"then number of tests (0)"_

This PR address the issue by following up on [my observation](https://github.com/wordpress-mobile/WordPress-iOS/pull/17233#issuecomment-958648497) that the reason for the weird test failures was in `multi_scan` a third-party `xcodebuild test` overlay that we introduced in the past to have a form of test repetition in CI.

Xcode 13 introduced a Test Repetition Mode configuration for test plans and this PR uses it to remove `multi_scan` in favor of Fastlane's default `scan`.

To test this look at the CI behavior. As you can see in my [comment below](https://github.com/wordpress-mobile/WordPress-iOS/pull/17233#issuecomment-958648497), without even trying, we already had a chance to verify the retry on failure behavior 🙃 

image
_You can see that the UI tests run in ~20 minutes each, as expected_

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -304,7 +304,7 @@ workflows:
           requires: [ "Build Tests" ]
       # Always run UI tests on develop and release branches
       - UI Tests:
-          name: UI Tests (iPhone 11)
+          name: UI Tests (iPhone)
           <<: *iphone_test_device
           post-to-slack: true
           requires: [ "Build Tests" ]
@@ -315,7 +315,7 @@ workflows:
                 - /^release.*/
                 - /^gutenberg\/integrate_release_.*/
       - UI Tests:
-          name: UI Tests (iPad Air 4th generation)
+          name: UI Tests (iPad)
           <<: *ipad_test_device
           post-to-slack: true
           requires: [ "Build Tests" ]
@@ -338,11 +338,11 @@ workflows:
                 - /^release.*/
                 - /^gutenberg\/integrate_release_.*/
       - UI Tests:
-          name: UI Tests (iPhone 11)
+          name: UI Tests (iPhone)
           <<: *iphone_test_device
           requires: [ "Optional Tests" ]
       - UI Tests:
-          name: UI Tests (iPad Air 4th generation)
+          name: UI Tests (iPad)
           <<: *ipad_test_device
           requires: [ "Optional Tests" ]
   Installable Build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,7 @@ jobs:
           command: bundle --path vendor/bundle
       - run:
           name: Run Unit Tests
-          command: bundle exec fastlane test_without_building name:WordPressUnitTests try_count:3 device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
+          command: bundle exec fastlane test_without_building name:WordPressUnitTests device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
       - ios/save-xcodebuild-artifacts:
           result-bundle-path: build/results
   UI Tests:
@@ -133,7 +133,7 @@ jobs:
           background: true
       - run:
           name: Run UI Tests
-          command: bundle exec fastlane test_without_building name:WordPressUITests try_count:3 device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
+          command: bundle exec fastlane test_without_building name:WordPressUITests device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
       - ios/save-xcodebuild-artifacts:
           result-bundle-path: build/results
       - when:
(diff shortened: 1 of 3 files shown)


mokagio:
Bad news: We got [a test failure](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/3450#8087d0f2-8cc4-42e2-a3b3-d1280ed49571/574-886) due to some flakiness that I hoped Xcode 13 had resolved.

Good news: The failure clearly shows `xcodebuild` honoring the Test Repetition setting and retrying on failure for 3 times<sup>*</sup> 😄

Screen Shot 2021-11-12 at 2 17 53 pm

---


`*`: 3 times is a fixed number that we can't change

mokagio:
[I force pushed](https://github.com/wordpress-mobile/WordPress-iOS/compare/35eb718487dbaa0262a5d165572399dffc5c97bb..afded64823031bb849ce84b788a9b34650a058e8) to move these changes on top of `develop` and get them merged ASAP. We'll look at the failing tests in appropriate branches.

.circleci/config.yml
```diff
@@ -304,7 +304,7 @@ workflows:
           requires: [ "Build Tests" ]
       # Always run UI tests on develop and release branches
       - UI Tests:
-          name: UI Tests (iPhone 11)
+          name: UI Tests (iPhone)
```
mokagio, line 307: Removed the device model information, so we don't have to keep updating it. 

I think there's value in knowing the tests are on iPhone or iPad, but given we don't run on multiple iPhone (or iPad) Simulators, knowing the exact device is not necessary.

This is also the pattern already adopted in the Buildkite pipeline: https://github.com/wordpress-mobile/WordPress-iOS/blob/ab81d4d401199b43da1af966596c734434ca4392/.buildkite/pipeline.yml#L47

WordPress/WordPressUITests/WordPressUITests.xctestplan
```diff
@@ -15,7 +15,8 @@
       "containerPath" : "container:WordPress.xcodeproj",
       "identifier" : "1D6058900D05DD3D006BFB54",
       "name" : "WordPress"
-    }
+    },
+    "testRepetitionMode" : "retryOnFailure"
```
mokagio, line 19: There are four modes. Here's my thoughts on when to use them:

1. _None_. Do not repeat the tests
2. _Until Failure_. Runs the tests until one fails, up to three times. Useful to discover flaky tests.
3. _Retry on Failure_. Will retry the tests that failed up to three times. Helpful to work around flaky tests that you haven't had the time to fix yet.
4. _Up Until Maximum Repetition_. Will run the tests three times (the number doesn't seem configurable yet) and report the success ratio. Useful to gauge the stability of your test suite.

I think "Retry on Failure" is the right option for us. In the future, it might be cool to explore using Until Failure on a dedicated pipeline running on a schedule to hammer the suite and discover flakes.

.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           command: bundle --path vendor/bundle
       - run:
           name: Run Unit Tests
-          command: bundle exec fastlane test_without_building name:WordPressUnitTests try_count:3 device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
```
mokagio, line 104: Now that the retry is configured in the Test Plan file and handled by `xcodebuild`, we don't need to manually specify one to pass down to the custom action `multi_scan`.

.circleci/config.yml
AliSoftware: I wonder if there is a risk for some UI tests to pass on a given iPhone model but fail on another 🤔 I'm especially thinking about UI tests written such that they scroll the screen a certain amount hoping to put an element into view before tapping it, but with different screen sizes the amount would be different etc.

I'm sure for those kind of cases written that way we could rewrite the test to be more generic (eg "scroll in a while loop until the element is visible") so that they are more flexible. But just thought I'd ponder about this out loud anyway in case you had thoughts about it.

AliSoftware (approved):

.circleci/config.yml
mokagio: Yes. I think there's a non trivial chance of this happening. At this point in time, though, we are basically running tests always on the same limited set of devices (e.g. @pachlava and I have been using the iPhone 12 Simulator for Simplenote because it's proven to be the most stable when he started writing tests).

I'd love to get to the point when we have a code coverage that's robust enough on one device that we can start thinking about testing on different ones. I'd be in particularly keen to run the tests on the smallest screen, biggest screen, and most common device on our user base.

Also, yes we ought to write tests in a "generic" way. And hopefully XCUITestHelpers will help us with that and with keeping in check issues across different devices.

_One day..._ 😄 

(I've been using devices as a synonym for Simulators)

PR #17457, opened 2021-11-12
Support custom UIToolbar appearances in iOS 15

### LINKED ISSUE #17454: Post preview buttons are indistinguishable from content
### Expected behavior

The "Publish" button on the post preview screen should be easily distinguishable from the post content.

### Actual behavior

The "Publish" button on the post preview screen is easily lost in the post content.

Screen Shot 2021-11-11 at 17 53 08 (1)

1. Open a new post
2. Tap the ellipsis three-dot button
3. Tap Preview
4. Notice that the "Publish" button lacks a surrounding foreground to distinguish it from the post content underneath.

### Additional info

This issue is **not** present on the current WPiOS 18.6 beta. The bug likely only appears on builds using Xcode 13, while to my knowledge WPiOS 18.6 was built using Xcode 12.5.1 (although I'm not 100% sure of this). It seems like Xcode 13, which uses the iOS 15 base SDK, causes the bug.

twstokes:
Fixes #17454

This PR aims to resolve a [toolbar](https://developer.apple.com/design/human-interface-guidelines/ios/bars/toolbars/) styling issue on iOS 15 when the project is compiled with Xcode 13+.

### To test:

**On an iOS 15 device running the project compiled by Xcode 13:**
1. On a site with published posts, go to My Site
2. Tap Posts
3. Tap View
4. Observe that the toolbar at the bottom has an opaque background that's appropriate for light and dark mode

<div>
Post preview in light mode

Post preview in dark mode
</div>

## Regression Notes
- iOS 13 and iOS 14 views with toolbars (e.g. viewing a Post or Page preview)
- Any other views that include toolbars in the app [[See Apple HIG](https://developer.apple.com/design/human-interface-guidelines/ios/bars/toolbars/) for examples of toolbars]

- Compared the toolbars on iOS 14 and 15 when viewing a published post
- Inspected other toolbars, such as the one shown when cropping an image inside a post

None, as this is a visual change.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
@@ -108,6 +108,16 @@ extension WPStyleGuide {
                                                    for: .disabled)
 
     }
+
+    class func configureToolbarAppearance() {
+        if #available(iOS 15.0, *) {
+            let appearance = UIToolbarAppearance()
+            appearance.configureWithDefaultBackground()
+
+            UIToolbar.appearance().standardAppearance = appearance
+            UIToolbar.appearance().scrollEdgeAppearance = appearance
+        }
+    }
 }
 
 
--- a/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
+++ b/WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
@@ -110,11 +110,12 @@ extension WPStyleGuide {
     }
 
     class func configureToolbarAppearance() {
-        if #available(iOS 15.0, *) {
-            let appearance = UIToolbarAppearance()
-            appearance.configureWithDefaultBackground()
+        let appearance = UIToolbarAppearance()
+        appearance.configureWithDefaultBackground()
 
-            UIToolbar.appearance().standardAppearance = appearance
+        UIToolbar.appearance().standardAppearance = appearance
+
+        if #available(iOS 15.0, *) {
             UIToolbar.appearance().scrollEdgeAppearance = appearance
         }
     }
--- a/WordPress/Classes/Utility/WebKitViewController.swift
+++ b/WordPress/Classes/Utility/WebKitViewController.swift
@@ -319,14 +319,15 @@ class WebKitViewController: UIViewController, WebKitAuthenticatable {
             return
         }
 
+        let appearance = UIToolbarAppearance()
+        appearance.configureWithDefaultBackground()
+        appearance.backgroundColor = UIColor(light: .white, dark: .appBarBackground)
+        appearance.backgroundEffect = UIBlurEffect(style: .systemThinMaterial)
+
+        toolBar.standardAppearance = appearance
+
         if #available(iOS 15.0, *) {
-            let appearance = UIToolbarAppearance()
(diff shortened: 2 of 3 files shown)


WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
```diff
@@ -108,6 +108,16 @@ extension WPStyleGuide {
                                                    for: .disabled)
 
     }
+
+    class func configureToolbarAppearance() {
+        if #available(iOS 15.0, *) {
```
twstokes, line 113: At this point I'm still unsure why on iOS 15 the toolbar is entirely translucent unless we explicitly set the appearance somewhere. I added this as a fallback for all toolbars, in case this happens in other areas.

WordPress/Classes/Utility/WebKitViewController.swift
```diff
@@ -318,7 +318,17 @@ class WebKitViewController: UIViewController, WebKitAuthenticatable {
         guard let toolBar = navigationController?.toolbar else {
             return
         }
-        toolBar.barTintColor = UIColor(light: .white, dark: .appBarBackground)
+
+        if #available(iOS 15.0, *) {
+            let appearance = UIToolbarAppearance()
+            appearance.configureWithDefaultBackground()
+            appearance.backgroundColor = UIColor(light: .white, dark: .appBarBackground)
```
twstokes, line 325: As far as I can tell there's no "tint color" concept so while the background of the toolbar will match iOS 14, it loses the slight blur + transparency effect.

WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
```diff
@@ -108,6 +108,17 @@ extension WPStyleGuide {
                                                    for: .disabled)
 
     }
+
+    class func configureToolbarAppearance() {
```
twstokes, line 112: I added this as a global fallback default, in case other areas were affected and we hadn't caught them yet. I don't have a great argument for keeping this, but it felt safe.

WordPress/Classes/Utility/WebKitViewController.swift
```diff
@@ -318,7 +318,17 @@ class WebKitViewController: UIViewController, WebKitAuthenticatable {
         guard let toolBar = navigationController?.toolbar else {
             return
         }
-        toolBar.barTintColor = UIColor(light: .white, dark: .appBarBackground)
+
+        let appearance = UIToolbarAppearance()
```
twstokes, line 322: Instead of adding an exception for iOS 15 I went ahead and applied these stylings to all iOS versions for consistency. This does mean that iOS 14's "tinting" of the toolbar with a blur effect is gone - it's entirely opaque now.

WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift
guarani: I agree this is OK to call globally. I also checked other places in the app for regressions and found none.

guarani (approved):
I tested on iOS 15 and iOS 13 and can't reproduce the bug anymore 👍 Also checked other parts of the app such as Reader which use toolbars and all look good there. Also checked dark mode and no issues.

Thanks for handling this, @twstokes!

twstokes:
Thank you for the review @guarani!

Issue #17405, opened 2021-11-02
[iOS] About Screen v1: Add logo cell


frosty:
Add cell showing logos for all Automattic apps. Ideally dynamic based on device motion.

<img width="403" alt="Screenshot 2021-11-02 at 11 08 38" src="https://user-images.githubusercontent.com/4780/139835765-1f49bbd3-7a83-4bf9-bbfb-5cf4acb0c234.png">

PR #17477, opened 2021-11-15
Unified About: Add device motion and haptics

frosty:
Refs #17405. This PR re-introduces device motion to the about screen and also adds haptic feedback on iPhones (thanks for the suggestion @annchichi!)

https://user-images.githubusercontent.com/4780/141813384-e01f090b-4d2d-4e6f-a6ea-053566aaa93d.mov

(the above video doesn't seem to be working for me in Firefox but is fine in Safari)

I finally found a way to lock the orientation of this screen on iPhone! I don't feel like it's too much of a hack, and it's limited to only this screen so it shouldn't affect anything else. As such, I've re-added device motion support to the app bubbles. Now as you rotate your device they should track the motion of the device. You'll also feel some haptic feedback on iPhone devices as they collide.

On iPad the orientation isn't locked, so the screen will auto-rotate. However, as the device is larger this doesn't feel as bad as it did on iPhones (you only had to tilt them a little before the screen would flip).

**Known Issues**

On iPhone, if you're viewing the Me screen in landscape and tap to view the About screen, it initially displays in landscape and the balls won't move in the correct direction of gravity. Once you rotate to portrait everything will be correct. I would hope that this is a quite unlikely to happen and it'll likely resolve itself as the user rotates their device anyway.

**To test:**

* Build and run on physical devices. Test both iPhone and iPad.
* Navigate to the About screen and try tilting and rotating the device. Check that the balls move in the direction you expect.
* Double check that a few other screens continue to rotate as you'd expect – I think essentially most screens should rotate except for the login flow.

## Regression Notes

* Rotation in other areas of the app, but I think my changes are just limited to this one specific screen due to the extra protocol I added.

I did a smoke test of rotating various areas of the app and everything seems to work as it should.

None

PR submission checklist:

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,3 +1,5 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
+BUNDLE_RETRY: "3"
+BUNDLE_JOBS: "3"
 BUNDLE_WITHOUT: "screenshots"
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,5 +1,3 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
-BUNDLE_RETRY: "3"
-BUNDLE_JOBS: "3"
 BUNDLE_WITHOUT: "screenshots"
--- a/WordPress/Classes/System/WordPressAppDelegate.swift
+++ b/WordPress/Classes/System/WordPressAppDelegate.swift
@@ -254,6 +254,17 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
+    // Note that this method only appears to be called for iPhone devices, not iPad.
+    // This allows individual view controllers to cancel rotation if they need to.
+    func application(_ application: UIApplication, supportedInterfaceOrientationsFor window: UIWindow?) -> UIInterfaceOrientationMask {
+        if let vc = window?.topmostPresentedViewController,
+           vc is OrientationLimited {
+            return vc.supportedInterfaceOrientations
+        }
+
+        return application.supportedInterfaceOrientations(for: window)
+    }
+
     // MARK: - Setup
 
     func runStartupSequence(with launchOptions: [UIApplication.LaunchOptionsKey: Any] = [:]) {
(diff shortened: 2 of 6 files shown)


.bundle/config
```diff
@@ -1,3 +1,5 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
+BUNDLE_RETRY: "3"
+BUNDLE_JOBS: "3"
```
Gio2018: are these related to this specific PR?

WordPress/Classes/System/WordPressAppDelegate.swift
```diff
@@ -254,6 +254,17 @@ class WordPressAppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
+    // Note that this method only appears to be called for iPhone devices, not iPad.
+    // This allows individual view controllers to cancel rotation if they need to.
```
Gio2018, line 258: nice!

Gio2018 (approved):
Just one question, but for the rest :shipit: !

.bundle/config
frosty: No, I missed that this got committed. I'm not sure why it changed exactly (guessing a different bundler version) so I'll revert it.

annchichi:
Great work, team! 👏👏👏. 

@frosty 

**iPhone**
- 1: It feels the gap is a bit larger. Preferably, it's 16pt. Currently, it's like it's around 34pt. 
- 2: I notice there is an arrow there. Is it possible to remove the arrow so that the move from bubbles can feel more natural? 😺 

**iPad**
- 3: It feels like the bubbles stop at the red lines. But perhaps you have already noticed. 

That's all from me. I Will add if I find more things. Thanks! 


Screen Shot 2021-11-17 at 10 24 48 PM
Screen Shot 2021-11-17 at 10 30 50 PM

frosty:
Thanks for the feedback, @annchichi! 

1. This is just the standard grouped tableview section spacing. I can adjust it if we need, but I wonder if it would be best to stick with the default?
2. Totally, good catch, this shouldn't be here. Will get it removed!
3. This is actually on purpose. Originally, the logos went all the way to the edges of the cell, but on a wider view like the iPad it'd result in a single row of icons and it didn't look great to me:

image

But I can revert to this if you'd prefer :)

annchichi:
Hey hey @frosty, 
1. hmm I think you are right. Let's stick with the default for now. It feels like a good reference for this one. 👍 
3. Regarding this one, I feel like it's natural that the bubbles become one-line where they can roll around or stack if the iPad is at a certain angle. The current one will make me feel like if it's a bug or they are 'blocked' by an invisible wall. Let's give it a try, how does that sound? 😃 

Thanks!

Issue #17421, opened 2021-11-04
[iOS] About Screen v1: implement "Share with Friends" navigation


diegoreymendez:
Share with Friends should replicate the new "Share WordPress with a friend" functionality from the Me screen

frosty:
Implemented by #17481

PR #17481, opened 2021-11-16
Unified About: Add 'Share' navigation

frosty:
Implements #17421. This PR adds the Share action to the unified about screen. I've used the same share presenter from the Me view controller and passed it into the About screen.

I've also made some tweaks to the `AboutItem` struct, and we'll likely need to change things a bit more for future actions. As it as declared as a static property, it wasn't possible to use the sharePresenter, so I've moved that logic out of the AboutItems and into the view controller itself. This way, the about items are simply about the presentation in the table rather than the logic.

**To test**

* Build and run
* Navigate to the new about screen
* Tap Share and check the share sheet appears

## Regression Notes

None

N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Me/App Settings/About/UnifiedAboutViewController.swift
+++ b/WordPress/Classes/ViewRelated/Me/App Settings/About/UnifiedAboutViewController.swift
@@ -9,15 +9,13 @@ struct AboutItem {
     let cellStyle: AboutItemCellStyle
     let accessoryType: UITableViewCell.AccessoryType
     let hidesSeparator: Bool
-    let action: (() -> Void)?
 
-    init(title: String, subtitle: String? = nil, cellStyle: AboutItemCellStyle = .default, accessoryType: UITableViewCell.AccessoryType = .disclosureIndicator, hidesSeparator: Bool = false, action: (() -> Void)? = nil) {
+    init(title: String, subtitle: String? = nil, cellStyle: AboutItemCellStyle = .default, accessoryType: UITableViewCell.AccessoryType = .disclosureIndicator, hidesSeparator: Bool = false) {
         self.title = title
         self.subtitle = subtitle
         self.cellStyle = cellStyle
         self.accessoryType = accessoryType
         self.hidesSeparator = hidesSeparator
-        self.action = action
     }
 
     func makeCell() -> UITableViewCell {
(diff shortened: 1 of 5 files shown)


WordPress/Classes/ViewRelated/Me/App Settings/About/UnifiedAboutViewController.swift
```diff
@@ -64,26 +62,52 @@ struct AboutItem {
...
+        case rateUs
+        case share
+        case socialTwitter
+        case socialFacebook
+        case socialInstagram
+        case legalAndMore
+        case automatticFamily
+        case appLogos
+        case workWithUs
+    }
```
diegoreymendez: I'm placing this here but it's actually a broader topic. 

## Some context:

One of the things that I find really bad about some of the libraries we have, is that although they unify a lot of logic, they do so in a very rigid way that forces out to continuously modify those libraries for each App.

This process is usually longer than just editing code within the App, because it forces us to continuously release new library versions, and integrate them into the App, even when the edit is necessary for just one App.

I'd argue in most cases it'd also be avoidable if the library design was slightly different.

One of the things I'd like us to do with this library, is to try and completely avoid this process.  Make it so that the library only needs modifications when it makes sense for all our Apps, or when there's a bug.

## My suggestion:

Which brings me to an improvement suggestion.

Why not go towards making our `sections` property customisable by having each `AboutItem` completely define what each row does (look and behaviour)?

This would involve:

1. Not having this fixed enum (`ItemIdentifier`) but rather having the `AboutItem` define each row.
2. Also not having `itemDetails`, since each `AboutItem` would have its own definitions.
3. Not having methods like `presentShareSheet(from:)` in `UnifiedAboutViewController` but instead have the action defined also in `AboutItem`.
4. Probably have a set of default `AboutItem`s to show in `UnifiedAboutViewController`, but allowing Apps to replace them or add new ones.

diegoreymendez (commented):
The feature works.

I added some suggestions in the code review.

I think this was probably not introduced by your PR but I noticed while testing that the link that I copied wasn't right... when pasted I got:

"Hey! Here is a link to download the WordPress app. I'm really enjoying it and thought you might too.".

WordPress/Classes/ViewRelated/Me/App Settings/About/UnifiedAboutViewController.swift
frosty: Thanks for the helpful suggestions, @diegoreymendez! I've pushed up some changes to take another attempt at this. One main difference from the list at the end of your comment is there's currently not a default list of AboutItems, but we could look at adding that if we wanted to. However, I feel they would be of limited utility as we need the host app to tell us what each item should do when tapped.

Let me know what you think!

diegoreymendez (approved):

WordPress/Classes/ViewRelated/Me/App Settings/About/UnifiedAboutViewController.swift
diegoreymendez: I agree on not having a default list of about items, because I don't think it's clear there's a need to.

Issue #17475, opened 2021-11-15
Update Reader Comments with the new design


dvdchr:
Refs `pbArwn-2ks-p2`

Reuse the existing components and logic in `CommentDetailViewController` for the new designs in `ReaderCommentsViewController`, particularly the header cell, content cell, and the moderation logic.

Designs from Zeplin:

![comment-threads-4](https://user-images.githubusercontent.com/1299411/141793520-e7483a42-a281-4065-b4bc-76f272156168.png)

## Tasks 🎯 

- [x] Add feature flag for the updates on Reader Comments. Addressed in #17480
- [x] Comment header cell
  - [x] Update the header cell so it can show/hide disclosure indicator. Addressed in #17480
  - [x] #17487
  - [x] #17517
- [x] #17476
- [x] Show an overflow dropdown menu for administrator roles.
  - [x] Implement the overflow dropdown menu. Addressed in #17601.
  - [x] #17622
  - [x] Present a share sheet when the Share menu is tapped. Addressed in #17601.
  - [x] #17623
  - [x] #17629
- [x] Review accessibility.
- [x] #17658
- [x] #17659

PR #17487, opened 2021-11-17
Reader Comments: Display the new comment header cell

dvdchr:
Refs #17475 

This PR implements the new comment header view in Reader Comments, with some notes:

- When the `newCommentThread` flag is on, `ReaderPostHeaderView` is no longer used. Instead, the new comment header cell will be displayed as a table section header. Ideally, this should be displayed as a regular table view cell in a separate section – but it's a bit complicated due to the dependency on `WPTableViewHandler`.
  - ⚠️  I just realized that the header is sticky due to it being a table section header 😅 . The only way that I can think of is to use the `tableHeaderView` – but it's tricky to make it work since it's still frame-based. Also, if this header cell is going to be removed later on, maybe this is okay for now? Anyways, I'll make the changes in a separate PR if needed.
- Impacts to the follow conversation logic. Some of the states are stored and read from the `ReaderPostHeaderView`, which will not be initialized when the `newCommentThread` flag is turned on. To preserve the original behavior, I've added a `subscribedToPost` property which replaces any calls to `postHeaderView.isSubscribedToPost` while maintaining behavior when the flag is turned off.
- Since the view controller now has two active flags: `followConversationViaNotifications` and `newCommentThread`, I've added support for when the comment thread flag is disabled, and when both flags are disabled. However, I did not handle the case where `followConversationViaNotifications` is turned OFF while `newCommentThread` is ON. Here are some screenshots:

Description | Screenshot
--|--
Both flags on | threads_both_on
`newCommentThread` disabled | threads_follow_on
Both flags disabled | threads_legacy

#### With the `newCommentThread` flag enabled:

- Go to Reader > Reader Comments.
- 🔍 Verify that the new comment header is displayed without disclosure indicator and selection style.
- 🔍 Verify that the Follow Conversation flow works as expected (roughly following the testing steps described in #17363).

#### With the `newCommentThread` flag disabled:

- Go to Reader > Reader Comments.
- 🔍 Verify that the previous comment header is displayed.
- 🔍 Verify that the Follow Conversation flow works as expected (roughly following the testing steps described in #17363).

#### With both `newCommentThread` and `followConversationViaNotifications` flag disabled:

- Go to Reader > Reader Comments.
- 🔍 Verify that the previous comment header is displayed.
- 🔍 Verify that the "Follow conversation via email" button is shown, and works when tapped.

## Regression Notes
Follow Conversation flow might be impacted.

Manually tested to ensure that the flow works as expected (both with the flag ON and OFF).

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
@@ -1,4 +1,5 @@
 import Foundation
+import UIKit
 
 @objc public extension ReaderCommentsViewController {
     func shouldShowSuggestions(for siteID: NSNumber?) -> Bool {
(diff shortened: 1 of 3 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
```diff
@@ -33,4 +34,31 @@ import Foundation
...
+        guard let post = post else {
+            return .init()
+        }
+
+        let cell = CommentHeaderTableViewCell()
+        cell.backgroundColor = .systemBackground
+        cell.configure(for: .thread, subtitle: post.titleForDisplay(), showsDisclosureIndicator: false)
+
+        // the table view does not render separators for the section header views, so we need to create one.
+        let separatorView = UIView()
```
ScoutHarris: Instead of adding a separator view here, I wonder if we can take advantage of the `UIView` extension that does the same thing?

`cell.contentView.addBottomBorder(withColor: .separator, leadingMargin: tableView.separatorInset.left)`

I only checked one device, but it looks the same to me.

ScoutHarris (approved):
I have one suggestion, but otherwise 👍 .

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
dvdchr: TIL! Updated in https://github.com/wordpress-mobile/WordPress-iOS/pull/17487/commits/ccf2e6151b6d25e0686aabdeda53f972156da19a. Thank you!

PR #17504, opened 2021-11-18
Add UI test to check that contact us can be loaded during login flow

tiagomar:
One of the identified critical flows to be automated was the ability to load the Contact Support messaging area during the Login. This PR adds a test to cover the scenario.

To test:
SupportScreenTests > testContactUsCanBeLoadedDuringLogin() must pass on CI.

## Regression Notes
None. The changes would only impact SupportScreenTests

N/A

N/A

PR submission checklist:

- [ x ] I have completed the Regression Notes.
- [ x ] I have considered adding unit tests for my changes. **N/A**
- [ x ] I have considered adding accessibility improvements for my changes. **N/A**
- [ x ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. **N/A**


WordPress/UITestsFoundation/Screens/Me/SupportScreen.swift
```diff
@@ -27,6 +27,21 @@ public class SupportScreen: ScreenObject {
...
     }
 
+    public func contactSupport() throws -> ContactUsScreen {
+        app.cells["contact-support-button"].tap()
+        addContactInformationIfNeeded()
+        return try ContactUsScreen()
+    }
+
+    private func addContactInformationIfNeeded() {
+        let emailTextField = app.textFields["Email"]
```
jostnes: Would it be possible to group this locator here for reusability? https://github.com/wordpress-mobile/WordPress-iOS/blob/32983737f021caf2957c07f6c302aabe8fe2abbd/WordPress/UITestsFoundation/Screens/Me/SupportScreen.swift#L12

jostnes (commented):
Tested the new test locally and it works fine, one small comment around where we keep locators. 

I'll refrain from approving for now, to let a more experienced team member review in case I'm missing something obvious 😄

WordPress/UITestsFoundation/Screens/Me/SupportScreen.swift
tiagomar: Good question! I'll just share my thoughts here, please don't take it as resistance for changing.

This element is in a dialog that kicks in when you tap `Contact Support` or `My Tickets` without having an email set in `Contact Email`.

image

My initial thought was that this element is not actually part of the SupportScreen and, as we are using Screen Objects model, having that element along with other SupportScreen elements might be confusing. I also thought it might be too much to have a specific ScreenObject for this dialog.

Let me know if it makes sense to you or I'll just change it. =)

jostnes: Fair enough 😄 

I would have grouped it together as part of the `SupportScreen` if it is an element that is only going to be displayed when on that screen. That said, I don't believe we have agreed to the style and standards of writing tests yet, so should be good to keep as it (your explanation does make sense) and we can talk about this later so that expectations are more aligned.

tiagomar: > I would have grouped it together as part of the SupportScreen if it is an element that is only going to be displayed when on that screen.

Yeah, it makes total sense to me.

> I don't believe we have agreed to the style and standards of writing tests yet, so should be good to keep as it 

I don't know if there was any agreement in the past but the two apps I've worked so far have completely different standards. We definitely need to approach it holistically.

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
```diff
@@ -25,4 +25,19 @@ class SupportScreenTests: XCTestCase {
...
     }
+
+    func testContactUsCanBeLoadedDuringLogin() throws {
+        let contactUsScreen = try PrologueScreen()
+            .selectContinue()
+            .selectHelp()
+            .contactSupport()
+
+        XCTAssert(contactUsScreen.isLoaded)
+        XCTAssert(contactUsScreen.canSendMessage())
```
pachlava: `canSendMessage()` checks if the`Send` button is enabled when there's text in the input. This means that if `Send` button in always enabled, no matter if there's text or not, the test will still pass. What do you think of additionally checking that `Send` button is disabled before text is entered there? (and maybe that it gets disabled again, onec text is deleted)?

pachlava: It's subjective, but if there was a dedicated `assertContactUsScreenLoaded` method, returning `ContactUsScreen`, and also having `canSendMessage` separated into smaller methods, it would be possible to fully use the advantage(?) of method chaining and have a test looking something like this:

```suggestion
    func testContactUsCanBeLoadedDuringLogin() throws {
        PrologueScreen()
            .selectContinue()
            .selectHelp()
            .contactSupport()
            .assertContactUsScreenLoaded()
            .enterText("A")
            .assertCanSendMessage()
```

(just a thought, it just about cosmetic change)

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
```diff
@@ -25,4 +25,19 @@ class SupportScreenTests: XCTestCase {
...
+
+        //Dismiss because tearDown() can't handle modals currently
+        try contactUsScreen.dismiss().dismiss()
+        //This extra step is actually handled by tearDown(), but adding it here cuts the execution time in half
+        try GetStartedScreen().goBackToPrologue()
```
pachlava: I tried moving this part to `tearDownWithError()`, and it actually worked for me (iPhone 13 / iOS 15.0 / Xcode 13.1). 

The difference in execution time was 3-5 seconds, and I suspect this is because we need to create another instance of `ContactUsScreen()` when in `tearDownWithError()`, which takes some time.

WordPress/UITestsFoundation/Screens/Me/ContactUsScreen.swift
```diff
@@ -0,0 +1,52 @@
...
+                attachButtonGetter,
+            ],
+            app: app
+        )
+    }
+
+    public func canSendMessage() -> Bool {
+        app.typeText("A")
+        let isSendButtonEnabled = sendButton.isEnabled
+        app.typeText(String(repeating: XCUIKeyboardKey.delete.rawValue, count: 1))
```
pachlava: Since the length of entered text is known, it's possible to avoid using a hardcoded count value, and use the string length instead.

pachlava (changes requested):
I've executed the tests multiple times locally, on both iPhone 12 and 13 (because I wanted to measure the performance), and it passed in all cases 👍 (the latest news is that iPhone 13 + iOS 15.0 is faster than iPhone 12 + iOS 14.5, which contradicts what I said one hour earlier on the call 😅).

Left a few notes, but I consider only one (about asserting the `Send` button state also in the case when no text is present) important enough to postpone merging. Once it's done, I think it's good to go.

WordPress/WordPressUITests/Tests/SupportScreenTests.swift
tiagomar: Yeah, it was actually cutting time compared to not calling it at all. As I removed a disabled no longer necessary test I could move this part to `tearDownWithError()` which definitely makes more sense.

WordPress/UITestsFoundation/Screens/Me/ContactUsScreen.swift
tiagomar: Yep. However, by addressing previous comments I ended up removing that code and just discarding the message.

WordPress/UITestsFoundation/Screens/Me/SupportScreen.swift
tiagomar: @jostnes as I did address the locators as you suggested. Thank you for pointing that out. =)

WordPress/UITestsFoundation/Screens/Me/ContactUsScreen.swift
```diff
@@ -0,0 +1,70 @@
...
+            app: app
+        )
+    }
+
+    public func assertCanNotSendEmptyMessage() -> ContactUsScreen {
+        XCTAssert(!sendButton.isEnabled)
+        return self
+    }
+
+    public func assertCanSendMessage() {
```
pachlava: For the consistency, what do you think of making this return `ContactUsScreen` too?

pachlava (approved):
Thank you for the follow-up @tiagomar! The test passed, and I also intentionally swapped the assertions to see test fail - and it did 👍 

I left one note, really a nitpicky one. PR is good to go as it is 🚀 Thank for working on this!

WordPress/UITestsFoundation/Screens/Me/ContactUsScreen.swift
tiagomar: Sure! Done in [6f21fa2](https://github.com/wordpress-mobile/WordPress-iOS/pull/17504/commits/6f21fa27936dc20abfe3da01d44285d3e421ec05).

PR #17519, opened 2021-11-22
[Tooling] L10n Modernization: Step 1

AliSoftware:
Part of project ref paaHJt-2Ib-p2 & paaHJt-2J8-p2 – which aims to modernize our L10n tooling for Widgets, and migrate the `Scripts/` in the app repos into Fastlane actions in our release-toolkit.

## What this does

This points to a commit on the release-toolkit which deprecated the `ios_localize_project` in favor of the newly introduced `ios_generate_strings_file_from_code` action. See wordpress-mobile/release-toolkit#309.

This PR introduces a new `generate_strings_file_for_glotpress` lane (which calls the new `ios_generate_strings_file_from_code` action) to do what the old `Scripts/localize.py` did – i.e. invoke `genstrings` on the code to generate `.strings` file – with a couple of differences along the way to clean things up:

 - The new `ios_generate_strings_file_from_code` action only generates one `.strings` file per invocation, while the `localize.py` script generated one for the app's `en.lproj` (the one imported by GlotPress) but also a `Base.lproj/Localizable.strings` file for the TodayWidget + ShareExtension. This is for now the responsibility of the lane in the `Fastfile` to invoke the action 3 times. This helps being explicit and less obscure, and the last 2 will soon not be necessary anymore anyway – see inline comment below for details.
 - The new action does not commit nor push to the repository. Indeed we decided in a discussion in Slack (ref: p1637266084156700-slack-C02KLTL3MKM) that it would be better for the Fastfiles to decide when to commit and push, rather than making our actions do it behind our backs. Having the `Fastfiles` do the commit + push will add some lines to the `Fastfiles` but we feel it's better to be explicit and have it that way.

### Create a baseline to compare your results with

 - Checkout this branch just before the commit removing the `Scripts/localize.py`
 - Run `bundle exec fastlane run ios_localize_project`
 - Observe that it tells you that the action is deprecated, in favor of the new `ios_generate_strings_file_from_code` action

_Note: After that point you can if you want update back to the HEAD of this PR's branch (including the commit removing the `Scripts/localize.py`)._

### Test the new action does not differ from the baseline

 - Validate that the diff is empty and that it didn't generate any change in the `.strings` files compared to the ones generated by `ios_localize_project` / `Scripts/localize.py`
 - Edit the codebase to alter one of the `NSLocalizedString` call. One great test is to modify (or add a new) one so that it has an English copy different from its key, using `NSLocalizedString("some key", value: "English copy here", comment: "Hint for translators here")`
 - Re-run `bundle exec fastlane generate_strings_file_for_glotpress`
 - Validate that the change you made in the codebase was taken into account and generated the corresponding diff in the generated `.strings` file.

---

#### 1. Potential unintended areas of impact

The generation of `.strings` file during code freeze

#### 2. What I did to test those areas of impact (or what existing automated tests I relied on)

I ran `./Scripts/localize.py` (old script), then my new lane (new action), and compared the outputs.
That allowed me to identify 2 keys for which the English copy was different from the key, which I fixed in the codebase (rather than via tooling) as expected per the new plan / process

#### 3. What automated tests I added (or what prevented me from doing so)

 -

PR submission checklist:


Scripts/localize.py
```diff
@@ -1,158 +0,0 @@
-#!/usr/bin/env python
```
AliSoftware, line 1: One down, 3 to go 😅

WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift
```diff
@@ -286,7 +286,7 @@ class AppSettingsViewController: UITableViewController {
                 NSUserActivity.deleteAllSavedUserActivities {}
             }
 
-            let notice = Notice(title: NSLocalizedString("Siri Reset Confirmation", comment: "Notice displayed to the user after clearing the Siri activity donations."), feedbackType: .success)
+            let notice = Notice(title: NSLocalizedString("Siri Reset Confirmation", value: "Successfully cleared Siri Shortcut Suggestions", comment: "Notice displayed to the user after clearing the Siri activity donations."), feedbackType: .success)
```
AliSoftware, line 289: This is the official and recommended way to provide English copies that differ from their key (as opposed to having the developers modify the `.strings` file manually).

Obviously once the whole project is done I'll publish a P2 post explaining the new official process, including how to handle those "key≠English copy" cases (tho full support for this will require the future migration of `Script/fix-translations` to handle the case too, that will be added later in the plan).

fastlane/Fastfile
```diff
@@ -138,9 +138,11 @@ platform :ios do
   desc 'Creates a new release branch from the current develop'
   lane :complete_code_freeze do |options|
     ios_completecodefreeze_prechecks(options)
-    ios_localize_project
-    version = ios_get_app_version
-    trigger_beta_build(branch_to_build: "release/#{version}")
+    generate_strings_file_for_glotpress
+
+    UI.confirm('Ready to push changes to remote and trigger the beta build?') unless ENV['RELEASE_TOOLKIT_SKIP_PUSH_CONFIRM']
+    push_to_git_remote
```
AliSoftware, line 144: As discussed in Slack (p1637266084156700-slack-C02KLTL3MKM), I've now moved the responsibility for doing the `git push` to the Fastfile, with an optional confirmation, as opposed to making the action do the push behind our back.

We initially mentioned creating a dedicated action to do the commit and push stuff, but after discovering that those actions existed built-in in `fastlane`, I decided to go with the built-in ones like here and [below](https://github.com/wordpress-mobile/WordPress-iOS/pull/17519/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R432).

fastlane/Fastfile
```diff
@@ -399,6 +401,37 @@ platform :ios do
...
+      exclude: ['*Vendor*', 'WordPress/WordPressTest/I18n.swift', 'WordPress/WordPressStatsWidgets/Views/Localization/LocalizedStringKey+extension.swift', '*/Secrets.swift'],
+      output_dir: en_lproj
+    )
+
+    # @TODO: Remove once we migrated the codebase of the widgets to use the app bundle for .strings
+    today_widget_base = 'WordPress/WordPressTodayWidget/Base.lproj'
+    ios_generate_strings_file_from_code(paths: ['WordPress/WordPressTodayWidget'], output_dir: today_widget_base)
+    share_ext_base = 'WordPress/WordPressShareExtension/Base.lproj'
+    ios_generate_strings_file_from_code(paths: ['WordPress/WordPressShareExtension'], output_dir: share_ext_base)
+    # END TODO
```
AliSoftware, line 430: This whole set of calls will very soon disappear – as soon as I implement Step 2 of paaHJt-2J8-p2 – which I plan to tackle right after those PRs finishing up Step 1.

But I figured because I'm not sure if Step 2 will be ready and merged in time before next code-freeze, I decided to keep that logic for now – so that the `Scripts/extract-framework-translations.swift` script (which is called before each new beta and before the final release) can still rely on those as it needs them to work (see 4️⃣  in paaHJt-2KK-p2).

Once I've migrated the whole codebase of the Widgets to make them look up their strings in the App bundle (rather than their own) – aka Step 2 planned later this week – and validated that this new way works as expected (ie widgets can indeed look up strings from a _parent_ bundle), these lines to generate the 2 `Base.lproj` (and the logic in 4️⃣ `extract-framework-translations.swift` which rely on them) won't be needed anymore and those will be removed.

Gemfile.lock
```diff
@@ -1,3 +1,22 @@
+GIT
+  remote: git@github.com:wordpress-mobile/release-toolkit
+  revision: 664f5f3e654abc0eb67b7b4650bb560731041e6f
```
AliSoftware: This is the latest commit after the merge of https://github.com/wordpress-mobile/release-toolkit/pull/309.

I don't plan to create an official new release of the toolkit until I finish all the work at least on the Widget L10n project (ref: paaHJt-2J8-p2): doesn't seem worth it until we have everything in place and cleaned up.

mokagio: I'm fine with this.

One other option, maybe a bit tidier, could be to ship a few betas. That does come, though, with the overhead of preparing them and doing the GitHub dances, which might add friction to the progress on this project, so 🤷 .

WordPress/Classes/ViewRelated/Me/App Settings/AppSettingsViewController.swift
mokagio: I had never seen the `value` parameter in action. Here's [a link to the docs](https://developer.apple.com/documentation/foundation/1418095-nslocalizedstring) for mine and other reviewer's reference.

fastlane/Fastfile
mokagio: I like your approach in splitting out the work here and leaving the automation in a working state. One never know what might happen that would delay work on this, so it's great to have everything working, even though it's in a transition state. 👍

fastlane/Pluginfile
```diff
@@ -6,9 +6,9 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 3.2.0'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', '~> 2.1'
+# gem 'fastlane-plugin-wpmreleasetoolkit', '~> 2.1'
 # This comment avoids typing to switch to a development version for testing.
-# gem 'fastlane-plugin-wpmreleasetoolkit', git: 'git@github.com:wordpress-mobile/release-toolkit', branch: ''
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'git@github.com:wordpress-mobile/release-toolkit', branch: 'develop'
```
mokagio, line 11: I was going to suggest locking this to the desired commit to avoid unintentional updates, but I don't think anyone runs `bundle update` here other than us infra folks and we know what we're doing (usually 😅🙃) so I think this is okay like this.

.bundle/config
```diff
@@ -1,3 +1,4 @@
 ---
 BUNDLE_PATH: "vendor/bundle"
+BUNDLE_JOBS: "16"
```
mokagio, line 3: If I recall correctly, we had a conversation about a similar diff in the past, it that right?

I haven't looked it up, but if I recall correctly we concluded that "yeah it should be fine, let's see how it goes?" 🤔 

I tried updating the `Pluginfile` and running `bundle install` with Bundler 2.2.31 (like in this PR) on my M1 MacMini and this didn't happen 🤔 🤔 🤔 🤷‍♂️

mokagio (approved):
Changes LGTM 👍 

I noticed something unexpected, but it's not related with the changes here: If `genstrings` ~~fails~~ logs an error, the action does not.

**Update** Since posting this, I realized that `genstrings` doesn't fail even though it prints that `error:` message. So I guess this is even less of an issue, or not at all one. Regardless, it's a topic for a conversation outside of this PR's scope.

I accidentally left out a `,` in my test edit (I wish I could say I did that on purpose...) to verify that the action modifies the output and I got this

Screen Shot 2021-11-23 at 1 53 35 pm

_Notice also the green arrow in my terminal, which I get only if the return code from the previous command was 0_

Realistically, this should never happen IRL because the code would not build and we'd notice that first. Still, I was surprised to see an error in the output and the action completed sucessfully.

The rest of the expected behaviors were all good. E.g.:

Screen Shot 2021-11-23 at 1 49 48 pm

Screen Shot 2021-11-23 at 1 47 14 pm

.bundle/config
AliSoftware: Yeah I still have no idea why on some machines (like on my Mac) it always changes this to 16 while on other machines it does not… 🤷

AliSoftware:
> I noticed something unexpected, but it's not related with the changes here: If `genstrings` ~fails~ logs an error, the action does not.

That's a very good point, I hadn't realized that.

Indeed since `genstrings` still exit with code `0` one could consider this expected that the action also doesn't error.
Another option would be to catch this at call site – since the action returns an Array of lines output by genstrings, so one could do `UI.user_error!(…) if output.any? { |line| line.include?('error') }` in the `Fastfile` if we want to catch those.

Alternatively, we could consider giving the option at the action level (i.e. provide a `fail_on_error:` ConfigItem to the action), which could be even nicer. I'll see to it.

> Realistically, this should never happen IRL because the code would not build and we'd notice that first.

This is not entirely true. Such error can also happen if one writes valid code, but uses a variable instead of a string literal as parameter. since `genstrings` can't / don't resolve variables dynamically. So this would also produce an error from `genstrings`:

```swift
let key = "Foo"
NSLocalizedString(key, comment: "genstrings is only able to parse string literals")
```

So maybe we need those to error and fail the lane after all…

AliSoftware:
Now that https://github.com/wordpress-mobile/release-toolkit/pull/311#issuecomment-976737450 is done, I've [updated the toolkit's sha1 in `Gemfile.lock` to point to the latest `release-toolkit`'s `develop`](https://github.com/wordpress-mobile/WordPress-iOS/pull/17519/commits/dc76204c9f8059376cbad609f0a954b91c4b271b) so that it contains that "fail on genstrings error" logic.

I've tested the new logic in this branch, by (arbitrarily) editing `Notification+Interface.swift` in the WPiOS codebase to use an intermediate variable for a key instead of a literal:

```swift
var description: String {
            switch self {
            case .Months:
                let monthOldKey = "Older than a Month"
                return NSLocalizedString(monthOldKey, comment: "Notifications Months Section Header")
…
```
And can see the error when running the lane (and no commit done)
```
$ fl generate_strings_file_for_glotpress
…
[19:16:28]: fastlane finished with errors

[!] genstrings: error: bad entry in file WordPress/Classes/Extensions/Notifications/Notification+Interface.swift (line = 63): Argument is not a literal string.
```
Which is exactly what we want, in order to catch such issues if they were to arise in the code. So all should be good 👍 

I'm going to merge this PR as soon as CI goes green in order to start working on Step 2 on top of these changes 🙂

mokagio:
> Such error [`genstrings: error: ..`] can also happen if one writes valid code, but uses a variable instead of a string literal as parameter. since `genstrings` can't / don't resolve variables dynamically. 

Oh, I hadn't thought of that. 

Despite what I said yesterday regarding this being `genstrings` behavior and it being okay for us to run along, I'm very happy you managed to implement error handling in a straightforward way at the plugin level. Being stricter in what we allow definitely helps "at scale" with many people contributing to the code. 👍 🎉

mokagio:
@AliSoftware, since you wrote

> I'm going to merge this PR as soon as CI goes green in order to start working on Step 2 on top of these changes 🙂

and I see all checks green, I'm going to merge 😄

AliSoftware:
Ahah yeah I've left that tab open waiting for the CI and kinda forgot about it (on second thought I should just have enabled auto-merge… but got distracted). And then you started your day way sooner than I thought and just beat me to it 😅

PR #17520, opened 2021-11-22
Onboarding Improvements: Add skip button to Epilogue if there are no sites yet

### LINKED ISSUE #17506: Onboarding Improvements: Hide "Choose a site" and the "OR" divider on the Epilogue screen when there are no sites
Thanks @pachlava for finding this!

Ref: p1637317113044600-slack-C027K4MNPGQ

### Expected behavior
If a user has no sites yet, the following should be hidden in the Epilogue:
-  `Choose a site to open` header
-  `----- OR -----` divider

### Actual behavior
If a user has no sites yet, the following are still shown in the Epilogue:
-  `Choose a site to open` header
-  `----- OR -----` divider

1. Log in to an account with no sites
2. Make sure you see the Post Signup Interstitial screen (see below for an example)

3. Log out and in again
4. Notice the `Choose a site to open` header and `----- OR -----` divider are shown even thought it doesn't really makes sense given the context

momo-ozawa:
Fixes #17506 

## Description
This PR adds a skip button to the Epilogue screen if there are no sites yet 

Slack ref: p1637330082057200/1637317113.044600-slack-C027K4MNPGQ

Before | After
-- | --
 | 

## How to test

1. Log out / log in to a site w no sites
2. Make sure you see the Post Signup Interstitial screen (see below for an example)

3. ✅ The skip button should be showing
4. Tap the skip button
5. ✅ The Epilogue should be dismissed

Case | JP iPhone | WP iPhone | WP iPad
-- | -- | -- | --
No sites |   |  |  
1-3 sites |  |  |  
4+ sites |   |  |  

## Regression Notes
- iPad readable content width
- Epilogue for 1-3 sites
- Epilogue for 4+ sites
- Jetpack app (Create a site should NOT be showing in the Epilogue ever)

- Tested the above steps manually

- N/A

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/NUX/LoginEpilogue.storyboard
+++ b/WordPress/Classes/ViewRelated/NUX/LoginEpilogue.storyboard
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="19162" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="FBE-8U-liw">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="19455" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="FBE-8U-liw">
     <device id="retina4_7" orientation="portrait" appearance="light"/>
     <dependencies>
         <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="19144"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="19454"/>
         <capability name="Safe area layout guides" minToolsVersion="9.0"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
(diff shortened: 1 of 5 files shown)


momo-ozawa:
@osullivanchris 👋 
Things to note about the new "no sites" variation:
- The skip button is a primary button
- The `--OR--` divider is hidden
- The button panel background color matches the screen background (I did this because there's no scrollable content)

@pachlava (@startuptester) 
Thanks for catching this!

leandroalonso (approved):
@momo-ozawa JFYI the screen that appears after a fresh install is this one:



Then, if I tap "Not right now" and logout/login I'm able to see this new flow.

Maybe we missed that but I wonder that it would have required less code if we just displayed that existent screen instead of adapting the current flow.

startuptester:
Tested on iPhone 13 Pro, iOS v15.1.1: 
- light/ dark 
- largest and smallest dynamic text 
- no sites, 1 sites, 30+ sites 
- choose skip, choose create, choose help
- choose create and dismiss
- delete only site, logout and back in (the first time I did this, I got the screen @leandroalonso mentioned ⬆️, but on all following logins, I got the new flow) 
- rotation (rotation to landscape does not happen in any screen of the flow)

What was not tested: (LMK if nobody else is able to test these and I will) 
- iPad
- JP 
- translations, LTR language 
- VoiceOver
- new account creation

momo-ozawa:
@leandroalonso @startuptester Thanks for the review!

> JFYI the screen that appears after a fresh install is this one:

That is expected! I forgot to mention this -- I updated the testing instructions to clarify the flow. Currently in `develop` we show the Post Signup Interstitial screen on a fresh install if there are no sites. Once a user has seen that screen, they will see a "No sites after first login" Epilogue variation on subsequent logins.

https://github.com/wordpress-mobile/WordPress-iOS/blob/a3eae1d75066b28c215180cafd12e4947c74e556/WordPress/Classes/ViewRelated/NUX/Post%20Signup%20Interstitial/PostSignUpInterstitialCoordinator.swift#L35-L44

> Maybe we missed that but I wonder that it would have required less code if we just displayed that existent screen instead of adapting the current flow.

Flow-wise, things wouldn't change since we already have a Epilogue screen layout variation for "No sites after first login". I thought about re-using the Post Signup Interstitial screen but wasn't sure if we'd want to change the flow logic.

No sites first log in | No sites after first login (Current develop) | No sites after first login (This PR)
-- | -- | --
 |  | 

Also. About Done vs. Skip, some feedback from @osullivanchris 

> I'd probably say 'skip' rather than done if paired with the 'Create new site' button. I'd be thinking "done what?"

Ref: p1637329996056800/1637317113.044600-slack-C027K4MNPGQ

pachlava (approved):
Since I was selected as a reviewer, I'm approving the PR based on the manual testing 🙂 Please keep in mind that I'm not commenting on the code itself.

I tested this on a real iPhone XR and on iPad Pro Simulator, and found no issues. Also tried VoiceOver. Giving my :shipit: . Thanks for working on this @momo-ozawa !

Along with testing, a few questions/considerations came up, that are probably well outside the scope, but I'd ask them here, because this PR provides a good overview of various screens.

1. Is there a certain reason for not having a `Skip` button also for the case when user already has sites? It's allowed in the very first screen (`Not right now`), and in the case when user has no sites (`Skip`) - why not have it in the third kind of screen?

2. I really liked the screen layout for the case when the user has 4+ sites, and `Create a new site` button is stuck to the screen bottom and feels like a solid chunk of UI. Just my thought: if this layout will also be used for "1-3 sites" case, there will be fewer cases (code) to maintain, with no real loss of usability.

3. It looks to me that `Welcome to WordPress` screen is very close in functionality to the `You don't have any sites yet` case. It's only 'advantage' is a bit more text and a 'Welcome' message. Same as in point above, it looks like merging these two screens (e.g. by moving more text to the `You don't have any sites yet`) will make the UX more logical, and will also mean less code to maintain.

I think that generalizing the UI and not having several screens that are close in their purpose will make it easier both for users and for engineers.

cc @osullivanchris

momo-ozawa:
Thanks for your reviews :)
Merging this for the time being. 

RE: reusing the Post Signup Interstitial screen for no sites
We can address this as needed in a different PR, after getting input from Chris.

PR #17532, opened 2021-11-23
Reader Comments: Display the new content cell

dvdchr:
Refs #17476
Depends on #17517

This PR shows the new content cell in Reader Comments when `newCommentThread` feature flag is enabled. Some other notes:

- Added a public boolean property `hidesModerationBar` to always hide the moderation bar when this is set to `true`. In the comment threads, the moderation bar is never shown even if the current user has moderating capabilities. Note that moderating actions will be available through the overflow dropdown menu later on.
- There are some noticeable performance and cell sizing issues when the flag is enabled, particularly for posts with a lot of comments _and_ content-rich comments. I'll address these later on after I've covered all the features for the new comment threads.

#### With the `newCommentThread` flag is enabled:

- Go to Reader > comment threads.
- 🔍  Verify that the new comment cell is displayed. 

#### With the `newCommentThread` flag disabled:

- Make sure that the account has moderating capabilities on the site.
- Go to My Site > Comments and select any comment.
- 🔍  Verify that the moderation bar is shown and functional.

## Regression Notes
1. Potential unintended areas of impact
n/a. The feature is hidden behind a flag.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a. The feature is hidden behind a flag.

3. What automated tests I added (or what prevented me from doing so)
n/a. The feature is hidden behind a flag.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #17476, opened 2021-11-15
Reader Comments: Reuse the new comment content cell


dvdchr:
Refs #17475, `pbArwn-2ks-p2#comment-4358`, and `pbArwn-38X-p2`

Reuse the content cell from `CommentDetailViewController`, reuse logic where possible, and customize as needed. Here's a list of expected changes to the content cell:

Description | Design
--|--
Author badge | <img width="357" alt="Screen Shot 2021-11-16 at 18 12 55" src="https://user-images.githubusercontent.com/1299411/141975350-d82a7f9f-9cb6-412c-89fc-d97044d6daab.png">
Highlighted comment[^1] | <img width="376" alt="Screen Shot 2021-11-16 at 18 14 32" src="https://user-images.githubusercontent.com/1299411/141975576-3abbe18f-df47-48df-b144-deb7871b64b9.png">
Top-level only separators | <img width="373" alt="Screen Shot 2021-11-16 at 18 16 11" src="https://user-images.githubusercontent.com/1299411/141975758-afdc03d1-28eb-4fad-9188-7dde0cdc7f5c.png">
Pending comments | <img width="370" alt="Screen Shot 2021-11-16 at 18 15 09" src="https://user-images.githubusercontent.com/1299411/141975630-56453de2-f7d0-4c70-9d9a-cdd7e0189345.png">
Reply button active state | <img width="299" alt="Screen Shot 2021-11-16 at 18 17 37" src="https://user-images.githubusercontent.com/1299411/141975968-ca78e40b-1251-49d6-a2a7-35b94cf598a4.png">

## Tasks

- [x] #17532
- [x] #17535
- [x] #17557
- [x] #17549
- [x] #17551
- [x] Parse comment author's ID from API: 
  - [x] https://github.com/wordpress-mobile/WordPressKit-iOS/pull/469
  - [x] #17566
- [x] Show author badge:
  - [x] Update WordPressKit to populate `authorID` in `RemoteReaderPost`. Addressed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/469.
  - [x] Parse `authorID` in `ReaderPost`. Addressed in #17566
  - [x] #17589
- [x] Implement logic for when the Reply and Like buttons are tapped. Addressed in #17600.
- [x] Handle when links in the comment body are tapped. Addressed in #17600.
- [x] Update the Reply button state when the reply text view is "targeting" the comment. Addressed in #17656.
[...]

PR #17535, opened 2021-11-23
Reader Comments: Update reaction bar logic

dvdchr:
Refs #17476
Depends on #17532

This PR updates the reaction bar logic to also consider Reader scenarios where the `comment` might not have a `Blog` reference. Some other notes:

- Updated the reaction bar to use a horizontal stack view instead. This is due to a case where the Like button could be shown while the Reply button is hidden – and I figured using a stack view would simplify things.
- Added convenient methods `canReply()` and `canLike()` on `Comment`, since the logic is a bit different for Reader, Site comments, and self-hosted.

Please note that Comment Likes can be toggled, but the API does not provide that info yet. Due to this, the Like button is still shown even when the feature is toggled off. This is legacy behavior, but I'll maintain this behavior for now.

## To Test

All of the test scenarios here require `newCommentThread` flag to be enabled. Here's a quick reminder on how to toggle post comments: Go to WordPress.com dashboard > Posts > Edit on a post > Discussion > Check/uncheck "Allow comments" > Update.

#### With comments enabled 🟢 :

- Navigate to Reader, and go to any comment thread. 
- 🔍 Verify that the Reply and Like button is shown.
- Go to My Site > Comments, and select any comment.
- 🔍 Verify that the Reply and Like button is shown.

#### With comments disabled 🔴 :

- Navigate to Reader, and go to any comment thread. 
- 🔍 Verify that the Reply button is hidden, but the Like button is shown.
- Go to My Site > Comments, and select any comment.
- 🔍 Verify that the Reply and Like button is shown.
  - ⚠️ Note that the Reply button should be hidden here, but non-Reader comments currently do not provide comment status information, unlike `ReaderPost` (i.e. `commentsOpen`). While this is a bug, I think this is a legacy issue unrelated to this PR. I'll log this into a separate issue later.

### Self-hosted site

(Note that the comment status is not relevant in self-hosted sites, as it currently only affects Reader comments due to the bug I've described above.)

- Make sure that you're logged out from your WP.com account, and then log in to your self-hosted site.
- Navigate to Reader, and go to any comment thread.
- 🔍 Verify that both the Reply and Like button is hidden.
- Navigate to My Site > Comments, and select any comment.
- 🔍 Verify that the Reply button is shown, and the Like button is hidden.

## Regression Notes
This could impact Site Comments, since the `CommentContentTableViewCell` is modified.

Manually tested Site Comments to ensure that it is not impacted.

n/a

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Comment+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Comment+CoreDataClass.swift
@@ -85,17 +85,18 @@ public class Comment: NSManagedObject {
         return !isReadOnly()
     }
 
+    // NOTE: Comment Likes could be disabled, but the API doesn't have that info yet. Let's update this once it's available.
     func canLike() -> Bool {
-        if let readerPost = post as? ReaderPost {
-            return readerPost.isLikesEnabled
+        if let _ = post as? ReaderPost {
+            return ReaderHelpers.isLoggedIn()
         }
 
         guard let blog = blog else {
-            // Likes feature is disabled for self-hosted sites.
+            // Disable likes feature for self-hosted sites.
             return false
         }
 
-        return blog.supports(.commentLikes)
+        return !isReadOnly() && blog.supports(.commentLikes)
     }
 }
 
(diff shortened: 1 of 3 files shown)


ScoutHarris (approved):
Looking good! 

:shipit:

PR #17536, opened 2021-11-23
Add UI tests to check the ability to scroll reader content and interact with the items

tiagomar:
One of the identified critical flows to be automated was to open the Reader section, scroll content, open and close selections. This PR adds tests to cover the scenario.

To test:
ReaderTests must pass on CI.

## Regression Notes
None. The changes would only impact ReaderTests.

N/A

N/A

PR submission checklist:

- [x] I have considered adding unit tests for my changes. **N/A**
- [x] I have considered adding accessibility improvements for my changes. **N/A**
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. **N/A**

### DIFF
--- a/WordPress/UITestsFoundation/Screens/ReaderScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/ReaderScreen.swift
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
         )
     }
 
+    public func openTheLastPostInApp() {
+        getTheLastPost().tap()
+    }
+
+    public func openTheLastPostInSafari() {
+        getTheLastPost().buttons["More"].tap()
+        app.buttons["Visit"].tap()
+    }
+
+    public func getTheLastPost() -> XCUIElement {
+        let postPosition = app.cells.count - 1
+        let post = app.cells.element(boundBy: postPosition)
+
+        scrollDownUntilElementIsHittable(element: post)
+        return post
+    }
+
+    private func scrollDownUntilElementIsHittable(element: XCUIElement) {
+        var loopCount = 0
+        while !element.waitForIsHittable(timeout: 3) && loopCount < 10 {
+            loopCount += 1
+            app.swipeUp(velocity: .fast)
+            print(loopCount)
+        }
+    }
+
+    public func postContentEquals(expected: String) -> Bool {
+        let equalsPostContent = NSPredicate(format: "label == %@", expected)
+        let isPostContentEqual = app.staticTexts.element(matching: equalsPostContent).waitForIsHittable(timeout: 3)
+
+        return isPostContentEqual
+    }
+
+    public func dismissPost() {
+        let backButton = app.buttons["Back"]
+        let dismissButton = app.buttons["Dismiss"]
+
+        if dismissButton.isHittable {
+            dismissButton.tap()
+        } else if backButton.isHittable {
+            backButton.tap()
+        }
+    }
+
     public static func isLoaded() -> Bool {
         (try? ReaderScreen().isLoaded) ?? false
     }
--- a/WordPress/UITestsFoundation/Screens/ReaderScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/ReaderScreen.swift
@@ -20,19 +20,17 @@ public class ReaderScreen: ScreenObject {
         )
     }
 
-    public func openTheLastPostInApp() {
-        getTheLastPost().tap()
+    public func openLastPost() {
+        getLastPost().tap()
     }
 
-    public func openTheLastPostInSafari() {
-        getTheLastPost().buttons["More"].tap()
+    public func openLastPostInSafari() {
+        getLastPost().buttons["More"].tap()
         app.buttons["Visit"].tap()
     }
 
-    public func getTheLastPost() -> XCUIElement {
-        let postPosition = app.cells.count - 1
-        let post = app.cells.element(boundBy: postPosition)
-
+    public func getLastPost() -> XCUIElement {
+        guard let post = app.cells.lastMatch else { fatalError("ReaderScreen: No posts loaded") }
         scrollDownUntilElementIsHittable(element: post)
         return post
     }
@@ -42,11 +40,10 @@ public class ReaderScreen: ScreenObject {
         while !element.waitForIsHittable(timeout: 3) && loopCount < 10 {
             loopCount += 1
             app.swipeUp(velocity: .fast)
-            print(loopCount)
         }
     }
 
-    public func postContentEquals(expected: String) -> Bool {
+    public func postContentEquals(_ expected: String) -> Bool {
         let equalsPostContent = NSPredicate(format: "label == %@", expected)
         let isPostContentEqual = app.staticTexts.element(matching: equalsPostContent).waitForIsHittable(timeout: 3)
 
--- a/WordPress/UITestsFoundation/Screens/ReaderScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/ReaderScreen.swift
@@ -54,11 +54,8 @@ public class ReaderScreen: ScreenObject {
         let backButton = app.buttons["Back"]
         let dismissButton = app.buttons["Dismiss"]
 
-        if dismissButton.isHittable {
-            dismissButton.tap()
-        } else if backButton.isHittable {
-            backButton.tap()
-        }
+        if dismissButton.isHittable { dismissButton.tap() }
+        if backButton.isHittable { backButton.tap() }
     }
 
     public static func isLoaded() -> Bool {
--- a/WordPress/WordPressUITests/Tests/ReaderTests.swift
+++ b/WordPress/WordPressUITests/Tests/ReaderTests.swift
@@ -1,0 +1,36 @@
+import UITestsFoundation
+import XCTest
+
+class ReaderTests: XCTestCase {
+    private var readerScreen: ReaderScreen!
+
+    override func setUpWithError() throws {
+        setUpTestSuite()
+
+        _ = try LoginFlow.loginIfNeeded(siteUrl: WPUITestCredentials.testWPcomSiteAddress, email: WPUITestCredentials.testWPcomUserEmail, password: WPUITestCredentials.testWPcomPassword)
+        readerScreen = try EditorFlow
+            .goToMySiteScreen()
+            .tabBar.goToReaderScreen()
+    }
+
+    override func tearDownWithError() throws {
+        takeScreenshotOfFailedTest()
+        if readerScreen != nil && !TabNavComponent.isVisible() {
+            readerScreen.dismissPost()
+        }
+        try LoginFlow.logoutIfNeeded()
+        try super.tearDownWithError()
+    }
+
+    let expectedPostContent = "Aenean vehicula nunc in sapien rutrum, nec vehicula enim iaculis. Aenean vehicula nunc in sapien rutrum, nec vehicula enim iaculis. Proin dictum non ligula aliquam varius. Nam ornare accumsan ante, sollicitudin bibendum erat bibendum nec. Aenean vehicula nunc in sapien rutrum, nec vehicula enim iaculis."
+
+    func testViewPostInApp() {
+        readerScreen.openTheLastPostInApp()
+        XCTAssert(readerScreen.postContentEquals(expected: expectedPostContent))
+    }
+
+    func testViewPostInSafari() {
+        readerScreen.openTheLastPostInSafari()
+        XCTAssert(readerScreen.postContentEquals(expected: expectedPostContent))
+    }
+}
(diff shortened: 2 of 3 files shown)


WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
         )
     }
 
+    public func openTheLastPostInApp() {
```
mokagio: Nitpick: Function names don't usually have articles in them. I don't see it spelled out clearly in the [Swift API Guidelines](https://www.swift.org/documentation/api-design-guidelines/#naming) and, thinking about it, I never really read this anywhere, but, based on all the code I've read over time, I think my assertion is accurate.

```suggestion
    public func openLastPostInApp() {
```

Also, I think it's safe to omit "in app" from the name. I'm guessing you did it to differentiate it from the method below that uses Safari. But, IMHO, if nothing else is specified it's safe to assume the action will occur in the app.

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
...
+    }
+
+    public func openTheLastPostInSafari() {
+        getTheLastPost().buttons["More"].tap()
+        app.buttons["Visit"].tap()
+    }
+
+    public func getTheLastPost() -> XCUIElement {
+        let postPosition = app.cells.count - 1
+        let post = app.cells.element(boundBy: postPosition)
```
mokagio: Out of curiosity, did you try the `lastMatch` helper here? It _should_ return the same value.

https://github.com/wordpress-mobile/WordPress-iOS/blob/01ccb6968fc68bbb3c470ba050b26a8b127ca90f/WordPress/UITestsFoundation/XCUIElementQuery%2BUtils.swift#L6-L8

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
...
+        scrollDownUntilElementIsHittable(element: post)
+        return post
+    }
+
+    private func scrollDownUntilElementIsHittable(element: XCUIElement) {
+        var loopCount = 0
+        while !element.waitForIsHittable(timeout: 3) && loopCount < 10 {
+            loopCount += 1
+            app.swipeUp(velocity: .fast)
+            print(loopCount)
```
mokagio: This looks like a debug statement. Did you intend to leave it in here? If not:

```suggestion
```

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
...
+
+    public func dismissPost() {
+        let backButton = app.buttons["Back"]
+        let dismissButton = app.buttons["Dismiss"]
+
+        if dismissButton.isHittable {
+            dismissButton.tap()
+        } else if backButton.isHittable {
+            backButton.tap()
+        }
```
mokagio: What would happen in the tests if neither is hittable?

WordPress/WordPressUITests/Tests/ReaderTests.swift
```diff
@@ -0,0 +1,36 @@
...
+            .tabBar.goToReaderScreen()
+    }
+
+    override func tearDownWithError() throws {
+        takeScreenshotOfFailedTest()
+        if readerScreen != nil && !TabNavComponent.isVisible() {
+            readerScreen.dismissPost()
+        }
+        try LoginFlow.logoutIfNeeded()
+        try super.tearDownWithError()
```
mokagio, line 22: It should not be necessary to call `super` here. Did you add this because of some error in the tests?

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
...
+    private func scrollDownUntilElementIsHittable(element: XCUIElement) {
+        var loopCount = 0
+        while !element.waitForIsHittable(timeout: 3) && loopCount < 10 {
+            loopCount += 1
+            app.swipeUp(velocity: .fast)
+            print(loopCount)
+        }
+    }
+
+    public func postContentEquals(expected: String) -> Bool {
```
mokagio: Another naming nitpick. Following the ["Omit needless words" guideline](https://www.swift.org/documentation/api-design-guidelines/#naming)

```suggestion
    public func postContentEquals(_ content: String) -> Bool {
```

or if you prefer:

```suggestion
    public func postContentEquals(_ expected: String) -> Bool {
```

Screen Shot 2021-11-24 at 4 02 06 pm

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,50 @@ public class ReaderScreen: ScreenObject {
...
+
+    public func getTheLastPost() -> XCUIElement {
+        let postPosition = app.cells.count - 1
+        let post = app.cells.element(boundBy: postPosition)
+
+        scrollDownUntilElementIsHittable(element: post)
+        return post
+    }
+
+    private func scrollDownUntilElementIsHittable(element: XCUIElement) {
```
mokagio, line 38: This is a nice method. Probably better than the [`scrollIntoView(within:, threshold:)`](https://github.com/wordpress-mobile/WordPress-iOS/blob/01ccb6968fc68bbb3c470ba050b26a8b127ca90f/WordPress/UITestsFoundation/XCUIElement%2BScroll.swift) we already have, because it doesn't rely on already knowing the scrollable container element.

I might steal it in the future 😄

mokagio (commented):
CI is happy, and the tests passed locally, too 🎉 

I left a few questions and suggestions. Once those are addressed, this will be ready to merge as far as I'm concerned.

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
tiagomar: Yeah, the idea was really to differentiate it. I was actually a bit uncomfortable with "InApp" because the "InSafari" is somewhat still _in app_. =P

I'll get rid of the articles. =)

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
tiagomar: I haven't tried it, probably because I started writing a more generic function to get the Nth post but then I realized I was doing premature optimization and also affecting the test readability. I'll use `lastMatch`. =)

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
tiagomar: 🤦 Yeah, don't tell my mother.

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
tiagomar: The `dismissPost()` is called only in `tearDownWithError()`. None of them being hittable will only happen if the tests fail prematurely even before opening the post and `tearDownWithError()` would continue its flow normally.

WordPress/WordPressUITests/Tests/ReaderTests.swift
tiagomar: No, I have just followed what all the other tests were doing. I did try removing the `super` now but it went into a recursive call loop. 

It runs just fine when removing the whole line but I would be more comfortable removing it if I knew why it was added in first place.

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
tiagomar: Yeah, way better!

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
pachlava: I assume (I didn't test this) that this method is designed to scroll and wait for hittability of the last element only, because swiping up with fast velocity will probably will make it not suitable for the ones in the middle; and because it's called by`getLastPost()`.

If this is so, what do you think of renaming the method to reflect this?

WordPress/WordPressUITests/Tests/ReaderTests.swift
```diff
@@ -0,0 +1,36 @@
...
+
+        _ = try LoginFlow.loginIfNeeded(siteUrl: WPUITestCredentials.testWPcomSiteAddress, email: WPUITestCredentials.testWPcomUserEmail, password: WPUITestCredentials.testWPcomPassword)
+        readerScreen = try EditorFlow
+            .goToMySiteScreen()
+            .tabBar.goToReaderScreen()
+    }
+
+    override func tearDownWithError() throws {
+        takeScreenshotOfFailedTest()
+        if readerScreen != nil && !TabNavComponent.isVisible() {
```
pachlava, line 18: Just thinking aloud: as long as `dismissPost()` is used only here, it's fine to have this 'fuse' here, but once it will be used in the other places too (like inside the tests), I think this check can be moved to the `dismissPost()` method itself.

WordPress/WordPressUITests/Tests/ReaderTests.swift
```diff
@@ -0,0 +1,36 @@
+import UITestsFoundation
+import XCTest
+
+class ReaderTests: XCTestCase {
+    private var readerScreen: ReaderScreen!
+
+    override func setUpWithError() throws {
+        setUpTestSuite()
+
+        _ = try LoginFlow.loginIfNeeded(siteUrl: WPUITestCredentials.testWPcomSiteAddress, email: WPUITestCredentials.testWPcomUserEmail, password: WPUITestCredentials.testWPcomPassword)
```
pachlava, line 10: Maybe helpful information: it's possible to get rid of the `_ = ` part by appending a `@discardable` before the respective `loginIfNeeded`.

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
mokagio: Yeah, no stress. I was just curious.

There are good utilities in our [XCUITestHelpers](https://github.com/Automattic/XCUITestHelpers/) library, but one needs to know about them first, and we don't have an easy way to discover them. I guess it's just like with any other library: We get familiar with them by using them and we can all help each other out by sharing suggestions 😄 

By the way: +1 for avoiding premature optimization. At the same time, if you stumble upon something that's a bit over-engineered for the current use case but looks general and something that could be useful in other tests feel free to implement it that way and extracting it into XCUITestHelpers as a followup 👍

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
mokagio: 👍

WordPress/WordPressUITests/Tests/ReaderTests.swift
mokagio: Ouch! My suggestions was not clear, sorry 😳 

"to call `super`" should have been "to call `super.tearDownWithError()` or "to call the super implementation". 

If you updated to 

```diff
-         try super.tearDownWithError()
+        try tearDownWithError()
```

it will indeed got into an infinite loop. ♻️ 

> No, I have just followed what all the other tests were doing.

Fair enough. Those methods are (should be?) empty in `XCTestCase` so it doesn't hurt to call them. It simply adds unnecessary code to the test.

Feel free to ignore this. I plan to audit the test suite for this kind of stuff during my HACK week. 👍 https://github.com/wordpress-mobile/WordPress-iOS/issues/17569

mokagio (approved):
Thank you for following up @tiagomar. :shipit:

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
```diff
@@ -20,6 +20,47 @@ public class ReaderScreen: ScreenObject {
...
+
+    public func dismissPost() {
+        let backButton = app.buttons["Back"]
+        let dismissButton = app.buttons["Dismiss"]
+
+        if dismissButton.isHittable {
+            dismissButton.tap()
+        } else if backButton.isHittable {
+            backButton.tap()
+        }
```
pachlava: Currently, the method assumes that the reader post will be either opened in a usual way (and then `Back` button should be used), or that it will be opened via `More>Visit` from reader list view. But there's also a third option: if it's opened in Safari from the post view (by clicking the globe button). In that case, it firstly needed to tap `Dismiss` and then `Back`. I assume that changing `else if` to just `if` would allow to also support this scenario.

pachlava (approved):
I started from executing the tests on iPhone 12, and they passed. Nice job @tiagomar, the code looks solid to me 👍. 

I left some notes, but none of them are blockers. Started review earlier today, but actually finished after Gio already approved the PR. Two approvals are better than one 😄

WordPress/UITestsFoundation/Screens/ReaderScreen.swift
tiagomar: Good point!

WordPress/WordPressUITests/Tests/ReaderTests.swift
tiagomar: Good to know. =D

WordPress/WordPressUITests/Tests/ReaderTests.swift
tiagomar: > If this is so, what do you think of renaming the method to reflect this?

The idea was to have it more generic but you are right about the `.fast`. It wouldn't make much difference in this case though, because the posts are mocked and we have only 3 of them.

Issue #17511, opened 2021-11-19
Reader Post Details: add Comments section


ScoutHarris:
Add a `Comments` section to Reader post details view that shows the first two comments and a `Show All Comments` button.

- [x] #17505
- [x] https://github.com/wordpress-mobile/WordPressKit-iOS/pull/468
- [x] #17515
- [x] #17524
- [x]  #17546
- [x] #17548
- [x] #17560
- [x] View all button: set different colors for border and label.
  - [x] #17598
  - [x] #17602
- [x] Show first comment.
  - [x] #17603
  - [x]  #17613
- [x] #17604
- [x] #17615
- [x] #17631
- [x] #17637 
- [x] #17638
- [x] Accessibility.
- [x] #17650

NOTE: `Follow Conversation` functionality will be added with #17632.

---
**From Zeplin:**

![Comments Under Post_Teaser](https://user-images.githubusercontent.com/1816888/143933825-24e28660-acb8-4d56-a126-0d98c25ab68a.png)

---
**Example empty view from Android:**

![no_comments](https://user-images.githubusercontent.com/1816888/144472888-0867a4c9-f6fe-4360-9167-c6d967735c3c.png)

PR #17546, opened 2021-11-23
Reader Post Details: fetch 2 top level comments

ScoutHarris:
Ref: #17511 

This fetches the post's comments and provides them to the comment table delegate.
- `syncHierarchicalComments` fetches the comments and updates Core Data.
- `topLevelComments` gets the cached comments.
- `updateComments` passes the comments and total comment count to `ReaderDetailCommentsTableViewDelegate`.

The comments are not displayed yet, but the number of rows in the empty table should correspond to the number of comments, plus one for the button row. (i.e. there will be a minimum of 1 row, and a maximum of 3)

To test:
- Enable the `postDetailsComments` feature.
- Go to Reader and select a post without comments.
  - Verify an empty table is shown with one row (0 comment rows + button row).
- Select a post with more than 2 top level comments.
  - Verify an empty table is shown with three rows (2 comment rows + button row).

## Regression Notes
N/A. WIP.

N/A. WIP.

N/A. WIP.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/CommentService.h
+++ b/WordPress/Classes/Services/CommentService.h
@@ -103,9 +103,12 @@ extern NSUInteger const WPTopLevelHierarchicalCommentsPerPage;
 
 // Sync a list of comments sorted by hierarchy, restricted by the specified number of _top level_ comments.
 - (void)syncHierarchicalCommentsForPost:(ReaderPost *)post
-                 numberTopLevelComments:(NSUInteger)number
+                 topLevelComments:(NSUInteger)number
                                 success:(void (^)(BOOL hasMore, NSNumber *totalComments))success
                                 failure:(void (^)(NSError *error))failure;
+
+// Get the specified number of top level comments for the specified post.
+- (NSArray *)topLevelComments:(NSUInteger)number forPost:(ReaderPost *)post;
 
 // Counts and returns the number of full pages of hierarchcial comments synced for a post.
 // A partial set does not count toward the total number of pages. 
--- a/WordPress/Classes/Services/CommentService.h
+++ b/WordPress/Classes/Services/CommentService.h
@@ -102,12 +102,16 @@ extern NSUInteger const WPTopLevelHierarchicalCommentsPerPage;
                                 failure:(void (^)(NSError *error))failure;
 
 // Sync a list of comments sorted by hierarchy, restricted by the specified number of _top level_ comments.
+// This method is intended to get a small number of comments.
+// Therefore it is restricted to page 1 only.
 - (void)syncHierarchicalCommentsForPost:(ReaderPost *)post
-                 topLevelComments:(NSUInteger)number
+                       topLevelComments:(NSUInteger)number
                                 success:(void (^)(BOOL hasMore, NSNumber *totalComments))success
                                 failure:(void (^)(NSError *error))failure;
 
 // Get the specified number of top level comments for the specified post.
+// This method is intended to get a small number of comments.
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Services/CommentService.h
```diff
@@ -103,10 +103,13 @@ extern NSUInteger const WPTopLevelHierarchicalCommentsPerPage;
 
 // Sync a list of comments sorted by hierarchy, restricted by the specified number of _top level_ comments.
 - (void)syncHierarchicalCommentsForPost:(ReaderPost *)post
-                 numberTopLevelComments:(NSUInteger)number
+                 topLevelComments:(NSUInteger)number
```
dvdchr: minor nitpick here but I think this needs one more tab for alignment 😄

WordPress/Classes/Services/CommentService.m
```diff
@@ -1195,6 +1195,13 @@ - (NSArray *)topLevelCommentsForPage:(NSUInteger)page forPost:(ReaderPost *)post
     return fetchedObjects;
 }
 
+- (NSArray *)topLevelComments:(NSUInteger)number forPost:(ReaderPost *)post
+{
+    NSArray *comments = [self topLevelCommentsForPage:1 forPost:post];
```
dvdchr, line 1200: I know this is out of scope and we're only going to pass in `2` here for the comments below post; But since this is a public method, the implementation could be misleading as it only returns comments from page 1. 

Should the method also fetch comments from page 2 (or more) in case if there are comments on the next pages and the `number` is not yet fulfilled?

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
```diff
@@ -188,8 +194,23 @@ class ReaderDetailCoordinator {
     /// Fetch Comments for the current post.
     ///
     func fetchComments(for post: ReaderPost) {
-        // TODO: fetch comments.
-        view?.updateComments()
+        commentService.syncHierarchicalComments(for: post,
+                                                   topLevelComments: commentsDisplayed,
+                                                   success: { [weak self] _, totalComments in
```
dvdchr: Nothing to be fixed here, just wanted to point out yet another little Xcode 13 bug. The IDE recognizes `for` as a for-loop syntax instead of a method parameter, and it misaligns the other parameters on the next lines 😅 .

dvdchr (approved):
Left some minor comments, and the rest is LGTM! :shipit:

WordPress/Classes/Services/CommentService.m
ScoutHarris: Since there are existing methods to get more comments with correct pagination, I purposefully restricted the new ones to page 1 since they are intended to get a minimal number of comments. But you make a good point - it is not clear. So instead of expanding them to allow multiple pages, I added comments noting that they are intentionally restricted to page 1.

WordPress/Classes/ViewRelated/Reader/Detail/ReaderDetailCoordinator.swift
ScoutHarris: > The IDE recognizes for as a for-loop syntax

Well of course it did. 😆 Thanks!

WordPress/Classes/Services/CommentService.h
ScoutHarris: Thank you!

PR #17554, opened 2021-11-24
Crash when going into a like-notification details screen.

### LINKED ISSUE #17529: Crash when taping on a like notification in Notifications tab
### Expected behavior

 - When tapping on a like notification in the Notifications tab, I expect to see the list of people who liked my post

### Actual behavior

 - After briefly showing the previous list of likers that the app had in cache, the app crashes without showing the updated list containing the latest likers

 - Go on the Notifications Tab
 - Tap on a "like" notifications that you already opened previously (†)
 - See the next screen, with the list of likers from the last time you opened that like notification, show up briefly, without the latest likers yet
 - After about 1 second on that screen (while it's apparently trying to update that list of latest likers), the app crashes

(†) This will only crash on specific "like" notifications. Given the log below, it likely only happens if a user who has their preferred blog having `user: nil` field (for some reason) is part of the list of likers.

### Logs

I found this in the logs which is **very likely** to be the culprit of the crash:

```
2021-11-23 10:54:01:834 🔵 Tracked: notifications_notification_details_opened <notification_type: like>
2021-11-23 10:54:03:933 Unresolved Core Data save error: Error Domain=NSCocoaErrorDomain Code=1570 "user est une valeur obligatoire." UserInfo={NSValidationErrorObject=<LikeUserPreferredBlog: 0x283a094a0> (entity: LikeUserPreferredBlog; id: 0xa74f17ddc8ce9958 <x-coredata://B2589825-657F-4C25-A3B2-DB4E71C15E45/LikeUserPreferredBlog/p2155>; data: {
    blogID = **redacted**;
    blogName = "**redacted**";
    blogUrl = "**redacted**";
    iconUrl = "";
    user = nil;
}), NSLocalizedDescription=user est une valeur obligatoire., NSValidationErrorKey=user, NSValidationErrorValue=null}
2021-11-23 10:54:03:933 Generating exception with reason:
Non-optional property with a nil value on LikeUserPreferredBlog.user
```

_Note: I've redacted the user info manually; for un-redacted data and video, see p1637670932064900-slack-C012H19SZQ8. Also note that the error's LocalizedDescription is in French simply because my phone is in French._

iPhone 11 Pro, iOS 15.1, WPiOS 18.7.0.0

diegoreymendez:
Fixes #17529 

There was an issue in how we were updating a CoreData relationship, in which we weren't properly deleting old stale managed objects, leaving them in a state that would cause a CoreData saving error.

This PR also adds unit tests to try and ensure we won't see a regression with this issue.

## Special Thanks

Special thanks to @jleandroperez for the great debugging session that resulted in us finding the actual problem.
Thanks @AliSoftware for all the help and great feedback when putting together this fix.

## To test:

- Checkout commit `e1db2d13ef8b3987a2f98408d6655e631fbea085`, which only contains the newly added unit tests.  Run them to verify that the second test crashes reproducing the issue we're seeing in production.
- Check out the latest commit from this branch, and run the unit tests again to ensure they succeed this time.

## Regression Notes

Notifications > Like details.  I'd expect this fix to resolve the issues and not introduce any new problem, but we have to test this thoroughly.

Added unit tests to ensure we made progress without breaking what was working.

I added some unit tests to ensure updates to the "likes" data works fine.

PR submission checklist:


jleandroperez:
Hey thanks for inviting me to the party @diegoreymendez !! Missed working with you sir!!

WordPress/WordPressTest/LikeUserHelperTests.swift
```diff
@@ -0,0 +1,74 @@
...
+        // Then we remove the preferred blog from the remote user, so we can save it again and make sure
+        // the preferred blog deletion works fine.
+        remoteUser.preferredBlog = nil
+        let updatedLikeUser = LikeUserHelper.createOrUpdateFrom(remoteUser: remoteUser, context: context)
+        XCTAssertNotNil(updatedLikeUser)
+        XCTAssertTrue(existingPreferredBlog?.isDeleted ?? false)
+
+        TestContextManager.sharedInstance().save(context) {
+            completionExpectation.fulfill()
+        }
```
AliSoftware: As suggested in Slack: it would be better for this Unit Test to properly fail, as opposed to crashing the test runner (as per the `[NSException raise]` that got triggered on failure from `ContextManager+ErrorHandling.swift:69`).

One way to do this would be to have the `TestContextManager` override `handleSaveError(_:in:)` to catch the errors – instead of raising an exception (that would make the app or test completely crash).
Another (maybe easier and better?) way would be to have your test catch the raised exception using `XCTAssertThrows`.

AliSoftware (changes requested):
Thanks for the fix and for adding unit tests for this!

I've left a suggestion to avoid making the test crash (instead of fail) on failure, but once this is addressed we should be good to go!

WordPress/WordPressTest/LikeUserHelperTests.swift
diegoreymendez: Addressed.

diegoreymendez:
Thanks for the feedback @AliSoftware , this is ready for another look.

WordPress/WordPressTest/LikeUserHelperTests.swift
```diff
@@ -28,46 +29,49 @@ class LikeUserHelperTests: XCTestCase {
 
     func testNewLikeUserWithPreferredBlog() {
         let completionExpectation = expectation(description: "We expect the context to save successfully")
-        let context = TestContextManager.sharedInstance().mainContext
```
AliSoftware: Probably something for a separate PR to not overload this one and side-track the work, but we should keep in mind that it would be nice to completely get rid of the `TestContextManager.sharedInstance` function altogether, so that there would be no risk for anyone to be tempted to use it (and instead be forced to create one brand new instance per test like you did here)

WordPress/WordPressTest/LikeUserHelperTests.swift
```diff
@@ -28,46 +29,49 @@ class LikeUserHelperTests: XCTestCase {
...
         let updatedLikeUser = LikeUserHelper.createOrUpdateFrom(remoteUser: remoteUser, context: context)
         XCTAssertNotNil(updatedLikeUser)
-        XCTAssertTrue(existingPreferredBlog?.isDeleted ?? false)
+        XCTAssertTrue(existingPreferredBlog.isDeleted)
 
-        TestContextManager.sharedInstance().save(context) {
+        XCTAssertNoThrow(contextManager.save(context, withCompletionBlock: {
             completionExpectation.fulfill()
-        }
+        }))
```
AliSoftware: I tested this by `git revert`-ing commit `7dca8d7fe4c8270e5c5f3a9cefc221f381b2d137` (which applies the fix for the issue), then running the `LikeUserHelperTests` test suite… and it seems I still got the test runner to crash (and not fail)… 😒 

```
*** Terminating app due to uncaught exception 'Unresolved Core Data save error (Main Context)', reason: 'Non-optional property with a nil value on LikeUserPreferredBlog.user'
```

So seems like `XCTAssertNoThrow` is not enough here to catch that `NSException`? (Maybe because the throw is happening async on a separate thread? 🤔 )

AliSoftware (changes requested):

WordPress/WordPressTest/LikeUserHelperTests.swift
diegoreymendez: Completely agree, but at the same time I also agree it's beyond scope for this specific PR.

WordPress/WordPressTest/LikeUserHelperTests.swift
diegoreymendez: Rolled back the changes to the initial version, as discussed.

WordPress/WordPressTest/LikeUserHelperTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        // Then we remove the preferred blog from the remote user, so we can save it again and make sure
+        // the preferred blog deletion works fine.
+        remoteUser.preferredBlog = nil
+        let updatedLikeUser = LikeUserHelper.createOrUpdateFrom(remoteUser: remoteUser, context: context)
+        XCTAssertNotNil(updatedLikeUser)
+        XCTAssertTrue(existingPreferredBlog.isDeleted)
+
+        XCTAssertNoThrow(contextManager.save(context, withCompletionBlock: {
+            completionExpectation.fulfill()
+        }))
```
AliSoftware: For reference, I also tried the following in an attempt to catch the ObjC `NSException`:

```swift
XCTAssertNoThrow(
    try WPException.objcTry {
        contextManager.save(context) {
            completionExpectation.fulfill()
        }
        self.waitForExpectations(timeout: 50.1)
    }
)
```

But even when wrapping the call inside the `WPException` wrapper (which converts raised ObjC `NSException` into thrown `Swift.Error`) and ensuring exceptions raised from `waitForExpectation` would also be caught (since that's the actual line on which the test runner crashes), it didn't seem enough to be able to catch the `NSException`

So after all I cave and say let's call it a day, revert this, add a comment, and let the test crash on failure for now. And keep the fix about "making our unit tests failing instead of crashing on CoreData save failure" for another PR, ince if seems like it would be more complicated fix than initially thought anyway.

```suggestion
        // Note: If the save fails, the current code will `@throw` an ObjC-based `NSException` and the test
        // runner will thus **crash** (as opposed to having the test properly report a failure.
        // Sadly, `XCTAssertNoThrow` is not enough to catch this (as it only catches `Swift.Error` `throw`n by Swift
        // or `(NSError**)` inout-returned by ObjC, not interupt-based `[NSException raise]`), and wrapping all
        // this within an `WPException.objcTry` to convert the raised `NSException` into a thrown `Swift.Error` does
        // not seem enough either – likely because the exception is thrown from a different thread?).
        // TODO: Fix this (maybe by making the `TestContextManager` not raise an `NSException` in the first place, as
        // opposed to the regular `ContextManager` used by the app?) to make the unit test fail properly instead of crashing.
        contextManager.save(context, withCompletionBlock: {
            completionExpectation.fulfill()
        }))
```

AliSoftware (approved):
I'd suggest to add a `// TODO` comment on the test mentioning that we know it crashes rather than fail (maybe not as lengthy as the one in my suggestion 😅  feel free to come up with a shorter comment of your own) and should fix that later.

Otherwise LGTM so I'll approve the PR so you can merge ASAP after that.

WordPress/WordPressTest/LikeUserHelperTests.swift
AliSoftware: Ahah seems you commented here at the same time I added my review+comment and we crossed the streams 😅

PR #17579, opened 2021-11-26
Enable new About screen for Jetpack

frosty:
This PR enables the new About screen for Jetpack. I've updated the screen to use the existing `AppConstants`, and added a few extras so that they can share an About Screen configuration.

I'm not sure if the `currentAppIconName` code really belongs in the AppConstants file or not, I could see a case for it being moved elsewhere. Let me know if you think it should go somewhere else!

**To test**

* Build and run Jetpack
* Navigate to Me > About Jetpack for iOS
* Check the screen looks good and that each item navigates where you'd expect. Rate Us, Share, Twitter, and Blog should all take you to Jetpack-specific content.
* Also Build and run WordPress and double check the About screen there still reflects WordPress and not Jetpack.

## Regression Notes

None really, unless the WordPress about screen accidentally now referenced Jetpack somewhere.

I retested each item in the WordPress About screen.

None

PR submission checklist:


WordPress/Classes/Utility/App Configuration/AppConstants.swift
```diff
@@ -20,6 +22,10 @@ import WordPressAuthenticator
     static let pushNotificationAppId = "org.wordpress.appstore"
     #endif
     #endif
+
+    static var currentAppIconName: String {
```
leandroalonso: I think this would be a better addition in `AppStyleGuide`: https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Jetpack/AppStyleGuide.swift#L29

leandroalonso (approved):

WordPress/Classes/Utility/App Configuration/AppConstants.swift
frosty: Thanks! I pushed some code to make it easier to just get the current app icon from the `AppIcon` struct / helper, and I extracted that file from the AppIconViewController as it's now used in more than one place. Most of the code in there is just as it was, I just added a new method to fetch the current icon filename as a string and also some comments.

leandroalonso (approved):

PR #17581, opened 2021-11-27
Create Installable Builds with Buildkite

jkmassel:
Moves installable builds from CircleCI to Buildkite. 

Uses the release toolkit to post the installable build instructions instead of writing out a file for Peril – apart from that, it works the same way.

To test:
- Note that all the tasks on this PR pass.
- Download the IPA files from the links.
- Verify that the builds show up in App Center.
- Extra ~paranoia~ credit: try installing the App Center builds to ensure they work.

## Regression Notes
Could theoretically break installable builds.

Checked that they were actually posted to App Center.

n/a

PR submission checklist:


jkmassel:
This is a test

jkmassel:
This is a test

jkmassel:
You can test the <strong>WordPress</strong> changes on this Pull Request by downloading it from AppCenter <a href='https://install.appcenter.ms/orgs/automattic/apps/WPiOS-One-Offs/'>here</a> with build number: <code>pr17581-d8ea670</code>. IPA is available <a href='https://appcenter-filemanagement-distrib4ede6f06e.azureedge.net/ff8a8a45-cee4-4cc5-8f2d-125fa4a8163d/WordPress%20Alpha.ipa?sv=2019-02-02&sr=c&sig=XHDeDJNw7FliL9QUPPAib%2Fpxbf3WEQTvIugANqd61qA%3D&se=2022-01-28T23%3A26%3A22Z&sp=r'>here</a>. If you need access to this, you can ask a maintainer to add you.

jkmassel:
You can test the <strong>Jetpack</strong> changes on this Pull Request by downloading it from AppCenter <a href='https://install.appcenter.ms/orgs/automattic/apps/jetpack-installable-builds/'>here</a> with build number: <code>pr17581-d8ea670</code>. IPA is available <a href='https://appcenter-filemanagement-distrib5ede6f06e.azureedge.net/33d9af07-c504-4e43-8e63-b2107e7a6cf9/Jetpack%20Alpha.ipa?sv=2019-02-02&sr=c&sig=CFEPJutOoZ%2BL541KYUdM0%2BpMzYzvDMuQZHNB3OWZi%2F8%3D&se=2022-01-28T23%3A23%3A45Z&sp=r'>here</a>. If you need access to this, you can ask a maintainer to add you.

jkmassel:
Test

jkmassel:
Test

jkmassel:
Test

fastlane/Jetpack-Fastfile
```diff
@@ -17,12 +17,21 @@ APP_IDENTIFIER = 'com.automattic.jetpack'
...
     jetpack_alpha_code_signing
+    require 'git'
 
     # Get the current build version, and update it if needed
     version_config_path = File.join(PROJECT_ROOT_FOLDER, "config", "Version.internal.xcconfig")
     versions = Xcodeproj::Config.new(File.new(version_config_path)).to_hash
 
-    build_number = Time.new.strftime("%Y-%m-%d-%H")
+    long_hash = Git.open(PROJECT_ROOT_FOLDER).revparse('HEAD')
+    short_hash = `git rev-parse --short #{long_hash}`.strip!
```
AliSoftware: We have a helper method in release-toolkit which does something similar. Worth exposing as an action so it can be access via our Fastfiles? 🤔

.buildkite/commands/installable-build-jetpack.sh
```diff
@@ -0,0 +1,17 @@
...
+install_gems
+
+echo "--- :cocoapods: Setting up Pods"
+install_cocoapods
+
+echo "--- :closed_lock_with_key: Installing Secrets"
+bundle exec fastlane run configure_apply
+
+echo "--- :hammer_and_wrench: Building"
+bundle exec fastlane build_and_upload_jetpack_installable_build build_number:$BUILDKITE_PULL_REQUEST
```
AliSoftware: Given the similarities between the 2 scripts (WP vs JP) I wonder if it wouldn't be better to provide the target as argument?

 - Either by making a common lane `build_and_upload_installable_build app:{wordpress|jetpack} build_number:<value>` (which could then internally in the `Fastfile` switch on the parameter and call the appropriate original lane?) – and then pass down that `app:` argument from the pipeline to the `.sh` to the fastlane call
 - Or by keeping individual lanes and make the switch at the `.sh` level to call the appropriate lane (which would make a smaller change and allow to detect typos earlier)

This could even be something as simple as:
```suggestion
APP="$1"
bundle exec fastlane "build_and_upload_${APP}_installable_build" build_number:$BUILDKITE_PULL_REQUEST
```

That would avoid the duplication (and double-maintenance) for the 2 otherwise similar cases 🙂

fastlane/Jetpack-Fastfile
jkmassel: I thought about this, but [the current implementation](https://github.com/wordpress-mobile/release-toolkit/blob/68f3356b1578935e8368f59ac6c32bbe14327bb3/lib/fastlane/plugin/wpmreleasetoolkit/helper/git_helper.rb#L129-L136) returns a full hash, and the `ruby-git` project doesn't expose a nice interface for this.

Doesn't mean we couldn't fix it later, but for now I'm fine with this as a one-off? 🤷

.buildkite/commands/installable-build-jetpack.sh
jkmassel: The repetition here exists for all of the other commands too, which was why I figured this was ok. It'd be possible to extract it all our into a shared `install-dependencies.sh` file, but in the past I've found we've run into a lot of weirdness when we try to over-DRY things 😅

.buildkite/commands/build-for-testing.sh
```diff
@@ -5,7 +5,6 @@ echo "--- :rubygems: Fixing Ruby Setup"
 gem install bundler
 
 echo "--- :rubygems: Setting up Gems"
-restore_cache "$(hash_file .ruby-version)-$(hash_file Gemfile.lock)"
```
jkmassel, line 8: This was redundant

.buildkite/pipeline.yml
```diff
@@ -12,10 +12,26 @@ common_params:
...
+    command: ".buildkite/commands/installable-build-jetpack.sh"
+    env: *common_env
+    plugins: *common_plugins
+    if: build.pull_request.id != null
+
   #################
   # Build the app
   #################
-  - label: ":pipeline: Build"
+  - label: ":pipeline: Build for Testing"
```
mokagio: 👍

.circleci/config.yml
```diff
@@ -345,45 +290,9 @@ workflows:
...
-  Jetpack Nightly:
-    triggers:
-       - schedule:
-           cron: "0 0 * * *"
-           filters:
-             branches:
-               only:
-                 - develop
-    jobs:
-      - Jetpack Installable Build
```
mokagio: I checked the Buildkite schedules tab and there is no equivalent for this.

Has it been deemed unnecessary?

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 GIT
   remote: git@github.com:wordpress-mobile/release-toolkit
-  revision: eccf5a273276e976cd42c66590af087d1654037f
-  branch: develop
+  revision: 83e4d5f036f1871ce6f8e9411550d4040cebddca
+  branch: add/github-comment-action
```
mokagio: Asking because the PR is "Ready for review", not the "Draft": do you plan on shipping a new release toolkit build before merging this?

IMHO it would be appropriate, but I'm okay if it doesn't happen.

fastlane/Fastfile
```diff
@@ -580,23 +580,27 @@ platform :ios do
       notify_testers: false
     )
 
-    # Install SentryCLI prior to trying to upload dSYMs
-    sh('curl -sL https://sentry.io/get-cli/ | bash')
-
```
mokagio, line 602: The fact that CI passes is proof enough that this is unnecessary now, but for reference, this is unnecessary because our CI agents [are provisioned with it already](https://github.com/Automattic/buildkite-ci/blob/29b21405254993f92783c45875bb475a22afcc8f/src/macos-vms/manifests/xcode-13.json#L5).

fastlane/Fastfile
```diff
@@ -580,23 +580,27 @@ platform :ios do
...
     sentry_upload_dsym(
       auth_token: get_required_env('SENTRY_AUTH_TOKEN'),
       org_slug: 'a8c',
       project_slug: 'wordpress-ios',
       dsym_path: lane_context[SharedValues::DSYM_OUTPUT_PATH]
     )
 
+    return if ENV['BUILDKITE_PULL_REQUEST'].nil?
```
mokagio, line 604: Have you considered moving this check to the start of the lane? Is there a use case in which we would run all the above not on a pull request?

fastlane/Jetpack-Fastfile
mokagio: > Doesn't mean we couldn't fix it later, but for now I'm fine with this as a one-off? 🤷

👍 worth adding an issue / note somewhere? It's the kind of thing that would be nice to have, but easy enough to roll out manually. Meaning that the next time someone needs this, unless they know about / remember this conversation, they're likely to add yet another manual call.

mokagio (commented):
Looking good. Left just one question about a possible early return.

fastlane/Jetpack-Fastfile
AliSoftware: I mean, better than that, we could have used the built-in action already existing in fastlane for that apparently…

```
$ fastlane actions | grep git
…
| last_git_commit                             | Return last git commit hash, abbreviated commit hash, commit      | ngutman          |
…

$ fastlane action last_git_commit
…

+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|                                                                              last_git_commit Return Value                                                                              |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Returns the following dict: {commit_hash: "commit hash", abbreviated_commit_hash: "abbreviated commit hash" author: "Author", author_email: "author email", message: "commit message"} |
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

AliSoftware:
```suggestion
    short_hash = last_git_commit[:abbreviated_commit_hash]
```

.circleci/config.yml
AliSoftware: I'll let Jeremy reply but my guess is, since we'll now create Installable Builds for all PRs without waiting for confirmation (i.e. we won't have the "Hold" job anymore), this indeed kinda makes those Nightly Builds irrelevant anymore?

Gemfile.lock
AliSoftware: Just chiming in to say that since I likely won't get back to edit the release-toolkit (to port more actions to it as part of my L10n project) _that_ soon, as I'm still working on the client apps's widgets and extensions, I'm ok with us doing a minor version of the release-toolkit for this if deemed appropriate indeed. I'll soon need to work on a couple of new actions (for the Siri Intents and all) but that can totally go in a separate release-toolkit minor version after we do a new release for this.

That is, as long as @mokagio you confirm that you fixing your local env on https://github.com/wordpress-mobile/release-toolkit/issues/314#issuecomment-982860291 and https://github.com/wordpress-mobile/release-toolkit/issues/315#issuecomment-982884562 fixed those issues you had and that those are not blockers to making a new version which would include localization changes too.

PS: I'll make a quick PR to isolate my change that fixes #314 so it can be merged easier (rather than waiting for my PR on next widget to land) – EDIT: done in https://github.com/wordpress-mobile/WordPress-iOS/pull/17627

fastlane/Fastfile
loremattei: Is this impacting our confidence in being able to run a release build locally in an emergency, if we needed to?

.circleci/config.yml
jkmassel: Yep, this is the case – we can also trivially re-add it with a separate pipeline file, but I'd like to see if it's actually needed :)

fastlane/Fastfile
jkmassel: Good question – I'd be fine to do something like an `UI.user_error('Sentry not installed') unless sh('command -v sentry-cli')` in a `before` hook to preserve running locally in an uninitialized environment, but IMHO having a `curl $ANYTHING | bash` is generally a bad idea (though if you `blame` this file, I think I'm the one who put this here... 😅 )

fastlane/Fastfile
jkmassel: > Is there a use case in which we would run all the above not on a pull request

I don't know of one, but I also can't entirely rule it out (for instance, @startuptester had requested the ability to create an installable build of `develop` on-demand from a webhook, which I _think_ could call this action and "just work")

Gemfile.lock
jkmassel: I'm following @AliSoftware's lead on this, so I'm fine to continue to target `develop` (I'll update this reference) as long as he needs to, or move to a tag if one is created.

mokagio: I followed up on the `dirname` issue [confirming](https://github.com/wordpress-mobile/release-toolkit/issues/315#issuecomment-987123633) it was due to messed up state on my end. Still unclear about the `--tags` one, but https://github.com/wordpress-mobile/WordPress-iOS/pull/17627 will prevent it from happening during a code freeze again (whatever the tags-related cause is), so we're good 👍

fastlane/Fastfile
mokagio: 👍

fastlane/Fastfile
mokagio: I like the idea of erroring out if a pre-required external tool is not available, with a suggestion on how to install it, but without doing it for the user.

If we attempted to install it, we'd take on board the responsibility of keeping the install code working and up to date.

AliSoftware:
Could we make it so that the CI check name appears with a less ugly name on GitHub checks? (Ideally all while still keeping the 🛠️ emoji in Buildkite logs?)

image

fastlane/Fastfile
jkmassel: Addressed in 19e1106b1a12de17987349c42f92f577a65404cb

jkmassel:
> Could we make it so that the CI check name appears with a less ugly name on GitHub checks? (Ideally all while still keeping the 🛠️ emoji in Buildkite logs?)

Yep addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17581/commits/19e1106b1a12de17987349c42f92f577a65404cb

Gemfile.lock
jkmassel: Addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17581/commits/138b43f7a3b8a43d164533fefc12e950288afe44

AliSoftware:
> > Could we make it so that the CI check name appears with a less ugly name on GitHub checks? (Ideally all while still keeping the 🛠️ emoji in Buildkite logs?)
> 
> Yep addressed in [19e1106](https://github.com/wordpress-mobile/WordPress-iOS/commit/19e1106b1a12de17987349c42f92f577a65404cb)

You meant e0173f1 😛  but thanks anyway 🙇

AliSoftware:
> Ohhh so that's where those `pipeline-` prefixes that I've been seeing in many other steps's names before! 👍

FWIW I was about to submit a bug report to Buildkite, but then noticed that other people beat me to it, so logging this here for reference:
 - https://forum.buildkite.community/t/github-status-check-names-and-emojis/686 ➡️  Don't hesitate to upvote this one for visibility 😉 
 - https://github.com/buildkite/agent/issues/374 ➡️  Unfortunately closed 3+ years ago as part of their spring cleaning and never addressed 😞 

I wonder if we could push this back into their radar by pinging them on our Slack Connection with them? 🤔 

[EDIT] Note that the Forum answer hints at the fact that they were working on this back in 2020. I took a look [at the docs](https://buildkite.com/docs/pipelines/command-step) again and I saw that there's a `key` attribute (in addition to `label`) for command steps. Could it be used to solve this then, providing a concise `key` name and hoping they now use that (and not `label` as the base for the GitHub status check name?

jkmassel:
> [EDIT] Note that the Forum answer hints at the fact that they were working on this back in 2020. I took a look [at the docs](https://buildkite.com/docs/pipelines/command-step) again and I saw that there's a `key` attribute (in addition to `label`) for command steps. Could it be used to solve this then, providing a concise `key` name and hoping they now use that (and not `label` as the base for the GitHub status check name?

So [this works](https://github.com/wordpress-mobile/WordPress-iOS/pull/17581/files#diff-a3991ebf1475eb82acab13946ffc1eca02b43917f6d5bec3898f45c8b0b9bd53R24-R26), but unfortunately it's not particularly concise.

jkmassel:
@AliSoftware I've used the `github_commit_status` key for the moment to set prettier job names. I figure we can iterate on this in the future?

AliSoftware:
> @AliSoftware I've used the `github_commit_status` key for the moment to set prettier job names. I figure we can iterate on this in the future?

That looks perfect 👍 Note I see that the 22 checks on this PR still include both the old and new names, but I'm assuming it's because those checks ran with the old name before you made the change and that's why they still appear here, and wouldn't in subsequent PRs once this is merged, right?

(And sorry, I didn't intent to derail the original intent of this PR that much with that aside about CI check names 😅 )

jkmassel:
> I'm assuming it's because those checks ran with the old name before you made the change and that's why they still appear here, and wouldn't in subsequent PRs once this is merged, right?

This is my assumption as well 🙂

fastlane/Jetpack-Fastfile
```diff
@@ -17,12 +17,20 @@ APP_IDENTIFIER = 'com.automattic.jetpack'
   desc "Builds and uploads a Jetpack installable build"
   lane :build_and_upload_jetpack_installable_build do | options |
     jetpack_alpha_code_signing
+    require 'git'
```
AliSoftware: This os obsolete now that we rely on the `last_git_commit` action

fastlane/Jetpack-Fastfile
```diff
@@ -54,15 +62,27 @@ APP_IDENTIFIER = 'com.automattic.jetpack'
       notify_testers: false
     )
 
-    # Install SentryCLI prior to trying to upload dSYMs
-    sh("curl -sL https://sentry.io/get-cli/ | bash") unless sh("which sentry-cli")
-
```
AliSoftware, line 59: No call to `ensure_sentry_installed` here to replace this deleted code?

fastlane/Jetpack-Fastfile
```diff
@@ -17,12 +17,20 @@ APP_IDENTIFIER = 'com.automattic.jetpack'
...
     versions = Xcodeproj::Config.new(File.new(version_config_path)).to_hash
 
-    build_number = Time.new.strftime("%Y-%m-%d-%H")
+    short_hash = last_git_commit[:abbreviated_commit_hash]
+
+    build_number = [
+      short_hash,
+      '-',
+      Time.new.strftime("%Y%m%d-%H%M")
+    ].join
```
AliSoftware: Any reason why the build_number for Jetpack is forced to be this `<sha1-Ymd-HM>` (nice) but Wordpress [does not seem to follow that same logic](https://github.com/wordpress-mobile/WordPress-iOS/pull/17581/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688L548-L558) in the main `Fastfile`?

Why not make WordPress follow the same version format?

.buildkite/commands/installable-build-jetpack.sh
AliSoftware: > ```
> bundle exec fastlane build_and_upload_jetpack_installable_build build_number:$BUILDKITE_PULL_REQUEST
> ```

According to the `Jetpack-Fastfile` the `build_number` parameter is totally unused in the case of Jetpack, as [instead the sha1 and datetime are used as build number](https://github.com/wordpress-mobile/WordPress-iOS/pull/17581/files#diff-289f1885ca764ad16f0b4ad827f30af5649ab86283df3ebd2504ac25260072f1R26-R32).

We should definitively uniformize this so that both WP and JP use the same build-number format, especially when they're both related to the same build and commit, so we won't compare apple to oranges.

I like the idea of having the `BUILDKITE_PULL_REQUEST` as part of the build number if it's present, though including the sha1 even if it comes from a PR seems relevant to differentiate Installable Builds from same PR but generated before/after additional commits being pushed to the PR. I don't really see a strong point in having the date and time in the build number tho.

---

I'd suggest to use a similar format to what is used with Android, namely `prNNN-sha1` for Installable Builds made from PRs, and `trunk-sha1` for the ones one might build from a branch (e.g. if an EW manually triggers a build on the current `trunk` branch to test it, or if we do nightly builds, etc?).

Something like this probably (not tested):
```suggestion
if [[ "$BUILDKITE_PULL_REQUEST" == "false" ]]; then
  BUILD_NUMBER="${BUILDKITE_BRANCH}-${BUILDKITE_COMMIT:0:7}"
else
  BUILD_NUMBER="pr${BUILDKITE_PULL_REQUEST}-${BUILDKITE_COMMIT:0:7}"
fi
bundle exec fastlane build_and_upload_jetpack_installable_build build_number:"$BUILD_NUMBER"
```
Then you'd need to ensure that the logic in the `Jetpack-Fastfile` for JP is the same as the one in the `Fastfile` for WP.

AliSoftware (changes requested):
- Inconsistencies in WP vs JP, both in build number used and logic to derive it.
 - Misssing check for sentry
 - Obsolete `require`

mokagio:
@jkmassel  I'm going to bump this to the next release because we'll be code freezing 19.0 today and it seems like this needs a bit more touches.

mokagio:
I'm going to bump this to the next release because we'll be code freezing 19.1 today and it seems like this needs a bit more touches.

.buildkite/commands/installable-build-jetpack.sh
jkmassel: Addressed in de0c32a29bda7ba57fbd8c1397d9a34a8d007989 – I like the format, but I moved the logic into Ruby – this makes it a bit easier to maintain IMHO, and also drops the requirement for build numbers to be specified on the command line – the lane has enough information in any case to decide what it should be called.

If, in the future, we need the ability to override it, I'd be fine to add the option back in.

fastlane/Jetpack-Fastfile
jkmassel: Nice one – fixed in 210e9c78d31e206fb241490d2ac27c6f2f5dd7ff

fastlane/Jetpack-Fastfile
jkmassel: Removed in de0c32a29bda7ba57fbd8c1397d9a34a8d007989

fastlane/Jetpack-Fastfile
jkmassel: Aligned as part of de0c32a29bda7ba57fbd8c1397d9a34a8d007989

fastlane/Fastfile
```diff
@@ -967,6 +971,28 @@ def read_version_from_config
...
+
+# This function is Buildkite-specific
+def generate_installable_build_number
+
+  if ENV['BUILDKITE']
+    commit = ENV['BUILDKITE_COMMIT'][0,7]
+    branch = ENV['BUILDKITE_BRANCH']
+    pr_num = ENV['BUILDKITE_PULL_REQUEST']
+
+    return pr_num == 'false' ? "#{branch}-#{commit}" : "pr#{pr_num}-#{commit}"
```
AliSoftware, line 986: Way more readable than the previous way 👍 👏

fastlane/Fastfile
```diff
@@ -597,23 +594,27 @@ platform :ios do
...
     install_url = 'https://install.appcenter.ms/orgs/automattic/apps/WPiOS-One-Offs/'
 
-    # Create a comment.json file so that Peril to comment with the build details, if this is running on CI
-    comment_body = "You can test the changes on this Pull Request by downloading it from AppCenter [here](#{install_url}) with build number: #{build_number}. IPA is available [here](#{download_url}). If you need access to this, you can ask a maintainer to add you."
-    File.write('comment.json', { body: comment_body }.to_json)
+    comment_body = "You can test the <strong>WordPress</strong> changes on this Pull Request by downloading it from AppCenter <a href='#{install_url}'>here</a> with build number: <code>#{build_number}</code>. IPA is available <a href='#{download_url}'>here</a>. If you need access to this, you can ask a maintainer to add you."
+
+    comment_on_pr(
+      project: 'wordpress-mobile/wordpress-ios',
+      pr_number: Integer(ENV['BUILDKITE_PULL_REQUEST']),
```
AliSoftware, line 614: nit: never seen the use of `Integer(string)` in Ruby before, TIL, as I believe the idiomatic way to do that usually is to use `string.to_i` instead 🙃

AliSoftware (approved):
Looks all great now 👍 :shipit: 

_PS: Just saw that my last review ended up being individual comments instead of a proper PR review submission, sorry 😅  (did it from my phone which is probably what misled me 😅 )_

mokagio:
🎉

PR #17584, opened 2021-11-29
Update WPMediaPicker from version 1.7.2 to 1.8.0-beta.1

mokagio:
This is a routine update prompted by a step in our release process checklist.

@jkmassel @joshheald @frosty you folks are the last to have interacted with the library. Jeremy and Josh [on Woo](https://github.com/woocommerce/woocommerce-ios/pull/5138) and [the lib itself](https://github.com/wordpress-mobile/MediaPicker-iOS/compare/1.7.2...1.7.3) and Frosty here, 0b3c392bfc. Can you tell me if there's a reason why this was never update, and no one raised an issue about it?

Two possibilities:

- I missed it since its release, or like in the previous release cycle, simply run out of time for this
- There's no reason to upgrade, so no dev did it

If it's the latter, then feel free to close this PR.

I also noticed @SergioEstevao [bumped](https://github.com/wordpress-mobile/MediaPicker-iOS/blob/56371391da9fc7ea1350b9808af8b4d2f67ecbfd/WPMediaPicker.podspec#L5) the library to 1.8.0-beta.1 (although, there's no tag for it, yet). So let me know if it would be better to use that beta here instead.

## Regression Notes
The media picking workflow, I guess? The version has been out for a while and used in Woo, so it should be safe-ish to adopt here, too.

N.A.

N.A.

---

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -207,7 +207,7 @@ abstract_target 'Apps' do
 
     pod 'NSURL+IDN', '~> 0.4'
 
-    pod 'WPMediaPicker', '~> 1.7.2'
+    pod 'WPMediaPicker', '~> 1.8.0-beta.1'
     #pod 'WPMediaPicker', :git => 'https://github.com/wordpress-mobile/MediaPicker-iOS.git', :tag => '1.7.0'
     ## while PR is in review:
     # pod 'WPMediaPicker', :git => 'https://github.com/wordpress-mobile/MediaPicker-iOS.git', :branch => ''
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -464,7 +464,7 @@ PODS:
     - CocoaLumberjack (~> 3.4)
     - FormatterKit/TimeIntervalFormatter (~> 1.8)
   - WordPressUI (1.12.2)
-  - WPMediaPicker (1.7.2)
+  - WPMediaPicker (1.7.3)
   - wpxmlrpc (0.9.0)
   - Yoga (1.14.0)
   - ZendeskCommonUISDK (6.1.1)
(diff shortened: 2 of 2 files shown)


joshheald:
> @jkmassel @joshheald @frosty you folks are the last to have interacted with the library. Jeremy and Josh [on Woo](https://github.com/woocommerce/woocommerce-ios/pull/5138) and [the lib itself](https://github.com/wordpress-mobile/MediaPicker-iOS/compare/1.7.2...1.7.3) and Frosty here, [0b3c392](https://github.com/wordpress-mobile/WordPress-iOS/commit/0b3c392bfca44726f05e4cf45dc9de0774ec30ea). Can you tell me if there's a reason why this was never update, and no one raised an issue about it?

The update I made to the library was to avoid a crash when `accessibilityLabel` was requested for media which we show just as files. That label is based on `filename` on the asset, but that's an optional part of the asset protocol and previously, there was no check to see whether it's implemented before calling `filename`, which would result in a crash when using VoiceOver and other assistive technologies (and possibly even if you're not using them, I can't remember for sure.)

I checked the WPiOS app, and it wasn't susceptible to this crash because it implements `filename` on its asset subclass, so I couldn't see a pressing need to update. Should I have raised an issue regardless here, do you think? I don't remember seeing anything in the checklist for that, but perhaps the checklist was written from the perspective of someone who's working on the WPiOS app finding an issue, fixing it, and so updating the library version in WPiOS was implied...

frosty:
Sergio and I opted for 1.8.0 for his set of changes as they bumped the minimum supported version of iOS, so we felt it warranted a bigger version bump than just a bugfix release. I'm not sure why this was replaced by 1.7.3, however?

mokagio:
@joshheald 

>  I don't remember seeing anything in the checklist for that, but perhaps the checklist was written from the perspective of someone who's working on the WPiOS app finding an issue, fixing it, and so updating the library version in WPiOS was implied...

Nah, that's okay. I think it should be up to the WP iOS devs and us release train conductors to monitor and update the libraries as they change, if necessary. I mean, it would not have been _wrong_ for you to open a PR to update it here, too, since the change was safe, but we can't expect library authors to updated _every_ downstream consumer. It doesn't scale.

@frosty 

> Sergio and I opted for 1.8.0 for his set of changes as they bumped the minimum supported version of iOS, so we felt it warranted a bigger version bump than just a bugfix release. I'm not sure why this was replaced by 1.7.3, however?

🤔 I went for 1.7.3 because that's what the `pod update` command resolved.

I think both "issues" are due to this repo not being part of the list we (I?) check a part of the code freeze. I'll make sure to add it from now on. 😅

mokagio:
@frosty @SergioEstevao @joshheald I bumped the library to 1.8.0-beta.1, can you have another look please? 🙏 🙇‍♂️

frosty (approved):

PR #17566, opened 2021-11-25
Comments: Add author ID field.

dvdchr:
Refs #17476 
Depends on https://github.com/wordpress-mobile/WordPressKit-iOS/pull/469 👈 Please review this first, thank you!

This PR adds an `authorID` field, which represents the comment author's dotcom user ID. For context, this field will be used later on to show an "Author" badge in the comment thread if the comment is written by the post author (see #17476 for more details). Some methods in `CommentService` are slightly adjusted to ensure that the `authorID` is properly parsed to/from `RemoteComment`.

NOTE: I'll update the Podfile to point to WordPressKit `4.44.0-beta.2` once https://github.com/wordpress-mobile/WordPressKit-iOS/pull/469 is merged.

## To test

Since the field is not used yet, we'll need to ensure that this change does not introduce any side effects. Here are a couple of test cases that I've used:

- Go to My Site > Comments, and scroll through the list. Verify that the comments are displayed correctly.
- Tap one of the comments. Verify that the comment detail lists the comment correctly.
- Submit a reply to the comment. Verify that a reply is successfully submitted. 
- Go to Notifications, and select the comments filter. Verify that the comments are displayed correctly.
- Go to Reader, and tap on the comment icon. Verify that the thread displays the comments correctly.
- Go back to Reader, and this time try to submit a comment on the post. Verify that the comment is submitted successfully.

## Regression Notes
1. Potential unintended areas of impact
Should be none. The new field is not used anywhere yet.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
Manually smoke-tested  the comments flow.

3. What automated tests I added (or what prevented me from doing so)
Modified `CommentServiceRemoteREST` tests in WordPressKit to account for `authorID`.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17589, opened 2021-11-29
Reader Comments: Show the author badge for comments from author

dvdchr:
Refs #17476 
Depends on #17566

As titled, this shows the author badge when the comment is from the post author's. The author badge is a view that's added beside the name label in the content cell, and configured to show when the public property `badgeTitle` is not nil. Here's a preview:

Simulator Screen Shot - iPhone 12 - 2021-11-29 at 21 20 21

Some notes:
- There's an issue with atomic sites causing the author badge to not appear sometimes. When fetching reader comments from an atomic site, the author ID could sometimes be 0, causing it to never match with the author ID from the post.
- In Zeplin, the author badge uses SFCompactText font. This is not possible to do in iOS since it's reserved for watchOS, so instead, I'm using the default system font.

## To test

- Enable `newCommentThread` flag.
- Go to Reader, and go to any comment threads. Ideally non-atomic sites, but there's no easy way to know that unfortunately 😅 It's probably an atomic site when the author comments are not showing the badge. In that case, please try again with another post.
- Verify that the author badge is visible for author comments. Also, the badge should be hidden for the other comments.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Models/Comment+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Comment+CoreDataClass.swift
@@ -108,8 +108,7 @@ public class Comment: NSManagedObject {
     }
 
     func isFromPostAuthor() -> Bool {
-        guard let post = post,
-              let postAuthorID = post.authorID?.int32Value,
+        guard let postAuthorID = post?.authorID?.int32Value,
               postAuthorID > 0,
               authorID > 0 else {
                   return false
--- a/WordPress/Classes/Models/Comment+CoreDataClass.swift
+++ b/WordPress/Classes/Models/Comment+CoreDataClass.swift
@@ -106,6 +106,17 @@ public class Comment: NSManagedObject {
     @objc func isTopLevelComment() -> Bool {
         return depth == 0
     }
+
+    func isFromPostAuthor() -> Bool {
+        guard let post = post,
+              let postAuthorID = post.authorID?.int32Value,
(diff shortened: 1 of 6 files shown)


WordPress/Classes/Models/Comment+CoreDataClass.swift
```diff
@@ -106,6 +106,17 @@ public class Comment: NSManagedObject {
     @objc func isTopLevelComment() -> Bool {
         return depth == 0
     }
+
+    func isFromPostAuthor() -> Bool {
+        guard let post = post,
+              let postAuthorID = post.authorID?.int32Value,
```
ScoutHarris: Nitpick - this could just be `guard let postAuthorID = post?.authorID?.int32Value`.

ScoutHarris (approved):
One minor comment, but otherwise LGTM!

:shipit:

WordPress/Classes/Models/Comment+CoreDataClass.swift
dvdchr: Thank you! I initially planned to check other properties (e.g., author name) if the `authorID` does not exist or has invalid values but ended up returning `false` instead to prevent false positives. https://github.com/wordpress-mobile/WordPress-iOS/pull/17589/commits/ce1d2c7c7083cdf55b54bf8bef19f002845cef5f should address this issue.

PR #17564, opened 2021-11-25
Onboarding Improvements: Epilogue (select site / create a new site) + Quick Start prompt

momo-ozawa:
Fixes #17383

## Description

This PR merges the feature branch for the Onboarding Improvements project into `develop`

Ref: pbArwn-35A-p2

## How to test

Follow testing directions from:
- #17411
- #17449
- #17484

Case | JP iPhone | WP iPhone | WP iPad
-- | -- | -- | --
No sites |   |  |  
1-3 sites |  |  |  
4+ sites |   |  |  

## Regression Notes
1. Potential unintended areas of impact
- Login Epilogue flow
- Quick Start
- Site creation

2. What I did to test those areas of impact (or what existing automated tests I relied on)
- Tested the flows above manually

3. What automated tests I added (or what prevented me from doing so)
- Added UI tests #17561

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17592, opened 2021-11-29
Onboarding Improvements: only show quick start prompt when adding a self-hosted site

leandroalonso:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/pull/17564#issuecomment-979252223

This PR changes the Epilogue flow for when adding a self-hosted site.

### To test

Scenario: self-hosted

1. Do a clean install
2. Add a self-hosted site
3. Make sure that the prompt ("Want a little help...") is shown and the epilogue to select a site is *not shown*

Scenario: WP.com

1. Login with your .com account
2. Make sure that the epilogue appears
3. Tap a site and make sure the QuickStart prompt appears and behaves correctly

Scenario: creating a new site

1. Tap the site selector
2. Tap the "+" icon
3. Create a new site
4. When prompted for a little help tap "No Thanks"
5. Quick start should not start
6. Repeat the same steps but tapping "Show me around"
7. Quick start should start

**Important**: if the user taps "No thanks" the prompt shouldn't appear if they remove and then re-add the site.

## Regression Notes
Epilogue

Manual testing

-

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -24,6 +24,9 @@ final class QuickStartPromptViewController: UIViewController {
 
     /// Closure to be executed upon dismissal.
     ///
+    /// - Parameters:
+    ///   - Blog: the blog for which the prompt was dismissed
+    ///   - Bool: `true` if Quick Start should start, otherwise `false`
     var onDismiss: ((Blog, Bool) -> Void)?
 
     // MARK: - Init
--- a/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
+++ b/WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
@@ -24,7 +24,7 @@ final class QuickStartPromptViewController: UIViewController {
 
     /// Closure to be executed upon dismissal.
     ///
-    var onDismiss: ((Blog) -> Void)?
+    var onDismiss: ((Blog, Bool) -> Void)?
 
     // MARK: - Init
 
@@ -108,7 +108,7 @@ final class QuickStartPromptViewController: UIViewController {
     // MARK: - IBAction
 
     @IBAction private func showMeAroundButtonTapped(_ sender: Any) {
-        onDismiss?(blog)
+        onDismiss?(blog, true)
         dismiss(animated: true)
 
         WPAnalytics.track(.quickStartRequestAlertButtonTapped, withProperties: ["type": "positive"])
@@ -116,7 +116,7 @@ final class QuickStartPromptViewController: UIViewController {
 
     @IBAction private func noThanksButtonTapped(_ sender: Any) {
         quickStartSettings.setPromptWasDismissed(true, for: blog)
-        onDismiss?(blog)
+        onDismiss?(blog, false)
         dismiss(animated: true)
 
         WPAnalytics.track(.quickStartRequestAlertButtonTapped, withProperties: ["type": "neutral"])
--- a/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
+++ b/WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
@@ -526,7 +526,7 @@ private extension WordPressAuthenticationManager {
 
         // Otherwise, show the My Site screen for the specified blog and after a short delay,
         // trigger the Quick Start tour
-        self.windowManager.dismissFullscreenSignIn(blogToShow: blog, completion: {
+        self.windowManager.showAppUI(for: blog, completion: {
             QuickStartTourGuide.shared.setupWithDelay(for: blog)
         })
     }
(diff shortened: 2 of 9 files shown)


leandroalonso:
@osullivanchris I decided to show the Quick Start prompt instead of taking the user directly to My Site. If they tap "No thanks" then this won't be displayed if they re-add the site in the future, hope that's ok. :)

momo-ozawa (commented):
Thanks for wrangling this 🙏 

When I add a self-hosted site then tap "Show me around" on the Quick Start prompt, the quick start tour never starts / the quick start menu items don't show up 

https://user-images.githubusercontent.com/6711616/143916009-ef1840f5-7b4d-4854-81af-2b5d477c0f20.mp4

I think one way to solve this is to dismiss the My Sites screen upon successfully adding a self-hosted site to go back to the My Site screen before showing the Quick Start prompt.

This will also make the behavior between creating a wp.com site vs adding a self-hosted site more consistent. Right now, when a user successfully creates a wp.com site and taps the "Done" button, they are brought back to the My Site screen and the Quick Start prompt is shown as needed.

On a separate note, I think you'll need to update the `testAddSelfHostedSiteAfterWPcomLogin()` UI test.

leandroalonso:
@momo-ozawa thanks for all that info!

In addition to that I think I found an issue:

1. Create a new site
2. Tap "No thanks"
3. Quick start will start anyway

I'll try to fix that as part of this PR.

osullivanchris:
> @osullivanchris I decided to show the Quick Start prompt instead of taking the user directly to My Site. If they tap "No thanks" then this won't be displayed if they re-add the site in the future, hope that's ok. :)

Yeah that seems ok to me! I don't think we offered Quick Start to self-hosted in the past so this is new?

I was just testing around on the current build adding a self-hosted site, and after adding it I was still in the site picker. So maybe that's connected to the issue @momo-ozawa pointed out.

leandroalonso:
@momo-ozawa functionality-wise the PR is ready for review.

However, I'm having trouble with the UI tests. The thing is that right now for self-hosted sites the QuickStart prompt might appear (or not, if it has already been dismissed). Our tests relied on the epilogue *always* appearing, so this is taking longer than I expected.

I think I can still benefit from your review though, if you don't mind.

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
```diff
@@ -24,7 +24,7 @@ final class QuickStartPromptViewController: UIViewController {
 
     /// Closure to be executed upon dismissal.
     ///
-    var onDismiss: ((Blog) -> Void)?
+    var onDismiss: ((Blog, Bool) -> Void)?
```
momo-ozawa, line 30: Thanks for fixing this closure! Can you add a comment to clarify what the params mean?

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
```diff
@@ -519,6 +492,46 @@ extension WordPressAuthenticationManager: WordPressAuthenticatorDelegate {
...
+    func presentQuickStartPrompt(for blog: Blog, in navigationController: UINavigationController, onDismiss: ((Blog, Bool) -> Void)?) {
+        // If the quick start prompt has already been dismissed,
+        // then show the My Site screen for the specified blog
+        guard !quickStartSettings.promptWasDismissed(for: blog) else {
+
+            if self.windowManager.isShowingFullscreenSignIn {
+                self.windowManager.dismissFullscreenSignIn(blogToShow: blog)
+            } else {
+                navigationController.dismiss(animated: true)
+            }
```
momo-ozawa, line 511: I _think_ we can get rid of the if-else here, now that we're explicitly handling adding self-hosted sites.

```suggestion
            self.windowManager.dismissFullscreenSignIn(blogToShow: blog)
```

momo-ozawa (approved):
Works as described.

I tested the flows below for both No thanks / Show me around
- Logging into wp.com and selecting a site from the Epilogue screen
- Logging into wp.com and creating a new site from the Epilogue screen
- Adding a self-hosted site from the My Sites screen
- Creating a new site from the My Sites screen

leandroalonso:
@momo-ozawa the UI tests are ready for review. :)

I would recommend running them on your machine, since the CI is going through some instability but you can see that they're succeeding here: https://buildkite.com/wordpress-mobile/wordpress-ios/builds/3913#eb976da2-c779-4303-90ff-4d0cc2676a12

WordPress/Classes/ViewRelated/NUX/WordPressAuthenticationManager.swift
leandroalonso: We can't! There's one single flow that will break if we do this change:

1. Login into an account with no sites
2. Tap "Add a self-hosted site"
3. Fill the fields
4. On the end, you'll get stuck on the last screen (it will never be dismissed)

WordPress/Classes/ViewRelated/Blog/QuickStartPromptViewController.swift
leandroalonso: Done in c329da329787912562596f06045f6c17d1130b65

leandroalonso:
Closing this one in favor of https://github.com/wordpress-mobile/WordPress-iOS/pull/17611 (targeting `18.8`)

PR #17593, opened 2021-11-29
Add UI test to verify the site stats are being loaded properly

tiagomar:
One of the identified critical flows to be automated was verifying that Stats screen load correctly for a site with established traffic. This PR adds tests to cover the scenario.

### To test
StatsTests must pass on CI.

## Regression Notes
None. The changes would only impact StatsTests.

N/A

N/A

PR submission checklist:

- [x] I have considered adding unit tests for my changes. **N/A**
- [x] I have considered adding accessibility improvements for my changes. **N/A**
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. **N/A**


WordPress/UITestsFoundation/Screens/StatsScreen.swift
```diff
@@ -16,6 +23,38 @@ public class StatsScreen: ScreenObject {
         )
     }
 
+    public func verifyStatsAreLoaded(_ stats: [String]) -> Bool {
```
pachlava: Just a naming nitpick (applicable to all four newly created methods): articles in methods names are generally not necessary. It's [not against the rules](https://github.com/wordpress-mobile/WordPress-iOS/pull/17536#discussion_r755699987) at the same time. 🙂 

```suggestion
    public func verifyStatsLoaded(_ stats: [String]) -> Bool {
```

(I must admit, I just checked my Simplenote tests, and they also contain articles in some cases 😅).

pachlava (commented):
Hey @tiagomar !

I started with local execution (iPhone13 + iOS 15 / iPhone 12 + iOS 14.5) and saw some flakiness, so far caused by two reasons:

1. In some cases the `Years` stats failed to load, and still showed the Stats from `Insights`

Screenshot 2021-11-30 at 16 46 46

I'd say, for me, this happens maybe every third or fourth execution. I could not reproduce this manually 🤷 

Possibly, this can be solved by a repetitive tap, or by adding a pull-to-refresh once the tab is clicked.

2. Even though there's a `dismissCustomizeInsightsNotice` call in setUp, and this method uses a correct and actual Dismiss button ID, it failed to work on one occasion when I was executing on iPhone 12. Might be related to the absence of timeout required for the method to wait for button presence. Again, I'm not sure exactly 🤷 

On a general note, it should be discussed which level of detail is required for the Stats tests to be considered finished 🤔 I would say that the current test lays a good foundation for future work on stats, which is an area of the app that could easily deserve a separate project thread, if we would go for near-to-complete coverage. If so, things to keep in mind for future:
- Also checking the other tabs (Days, Weeks, Months), which would mean updating some mocks.
- So far, the tests check only a part of data shown on screen (e.g. not checking `Authors` and `Posts and Pages` in `Years` stats).
- Our source of test data is mocks. Because of this, we don't actually need to duplicate all the stats numbers in arrays in Xcode project. Instead, we can read from the mocks, and use the data from the mocks as our source of truth. Something like this is [done](https://github.com/woocommerce/woocommerce-android/blob/develop/WooCommerce/src/androidTest/kotlin/com/woocommerce/android/screenshots/util/MocksReader.kt) in Woo Android.

That being said, I'm not giving my explicit approval now only because I'm worried about the flakiness: did you experience it locally before, or on the CI?

WordPress/UITestsFoundation/Screens/StatsScreen.swift
```diff
@@ -4,10 +4,17 @@ import XCTest
 public class StatsScreen: ScreenObject {
 
     public enum Mode: String {
+        case insights
         case months
         case years
     }
 
+    private enum Timeouts {
+        static let short: TimeInterval = 1
```
pachlava, line 13: I like the idea of having timeouts for several cases 👍 
However, is a one-second timeout really reliable and enough?

WordPress/WordPressUITests/Tests/StatsTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    let yearsChartBars: [String] = [
+        "Views,  2019: 9148",
+        "Visitors,  2019: 4216",
+        "Views,  2018: 1215",
+        "Visitors,  2018: 632",
+        "Views,  2017: 788",
+        "Visitors,  2017: 465"
+    ]
+
+    func testStatsLoadProperly() {
```
pachlava: What do you think of having this tests separated in two, with more specific naming: `testInsightsStatsLoadProperly` and `testYearsStatsLoadProperly`?

By doing so, we can have two things:
- This will underline that so far only these two tabs are covered
- In case if, say, `Insights` test fails, the `Years` will still be executed

WordPress/UITestsFoundation/Screens/StatsScreen.swift
tiagomar: I stole it from @mokagio without even sending him a note. =P

I just wanted to try it out, so it really wasn't a thoughtful decision on those numbers but I saw the `3 ` seconds all over the place so it felt like a `default` and enough time to load elements depending on additional API calls, so I looked at `1` second timeout for elements/screens that are not dependent on API calls, like pop-up validation for an empty required field or settings screen. However, as the tests are taking a life to run and this 2 seconds wouldn't make much difference, we may end up never using the `1` second and always sticking to the `default` _just to be safe_.

tiagomar:
Thanks for taking time to review, @pachlava !

I'll take a look into item `2.` but my first guess would be that it is related to the item `1.`.

About `1.`, yes, I've seen it a lot locally and already tried double tap and pull-to-refresh without success.
What I observed so far:

- It _only_ happens after running the tests a few times (sometimes it happens right from the second try)
- It is symmetrical, some times it will stay 'locked' `Years` tab, not loading `Insights`.
- It _won't_ happen if you uninstall the app between tries
- It _won't_ happen if you switch emulators between tries
- When it happens it won't change the screen if you manually tap the other tabs
- To recover it manually I have to alternate multiple times between `tap any tab` and `pull-to-refresh`. I couldn't get to a combination that always work.
- I still don't have a real iOS device to test, but I don think it would reproduce.

My guess is that it could be an `Application/Emulator + Mocks` issue. From what I could notice it seems unlikely that it kicks in on a fresh run, that's what makes me think that it wouldn't be seen in the pipelines (assuming that we have ephemeral environments), but the only way to know it is actually have them running. So I'll be re-running then in the pipeline to see what the success rate looks like and post it here.

About the which Stats should we be checking, you're right, we should discuss what the strategy should be for both iOS and Android apps. The current approach came after a conversation with @thehenrybyrd, based on the experience with issues in that screen and how it is tested manually. I'll save details for the proper discussion with the team on the overall strategy. But again, I agree with you, this test is more like a foundation.

WordPress/UITestsFoundation/Screens/StatsScreen.swift
tiagomar: Oops, I did it again. ®️ 
Would it be the same for assertion methods like `assertStatsAreLoaded()`?  🤔

tiagomar:
> In some cases the Years stats failed to load, and still showed the Stats from Insights

I'll be updating here the `StatsTests` success rate in the pipeline: 100% (10/10). _Each run being iPhone AND iPad._

_Notes:_ 
_1. One of the runs failed in iPhone because of a Log in test. All tests passed in a second run._
_2. One run failed for iPad because of an Editor test: `✗ testBasicPostPublish, failed: caught error: "timedOut"` [not a good error message]_

mokagio:
@tiagomar @pachlava, sorry for not having looked at this yet. I got side tracked with other things and then went down a rabbit hole with what looked like a flaky test https://github.com/wordpress-mobile/WordPress-iOS/pull/17607

WordPress/UITestsFoundation/Screens/StatsScreen.swift
pachlava: > Would it be the same for assertion methods like assertStatsAreLoaded()?

I tend to say yes, because from a code style guide POV (which doesn't really cover this point), it should be the same for all cases.

pachlava (approved):
Thanks for the extra information about flakiness, and taking your time to check the stability on CI. Just executed the now separated test(s) - it went well locally. As long as it's stable on CI, I'd say it's good to go :shipit: Thank you, @tiagomar !

tiagomar:
Thank you, @pachlava!
StatsTests passed 10 out of 10 runs in pipeline.

PR #17600, opened 2021-11-30
Reader Comments: Handle content cell action buttons

dvdchr:
Refs #17476

This adds functionality for the Reply and Like buttons when tapped. Additionally, when tapping links in the comment body, the destination URL should be presented in a web view. Some notes:

- When tapping the Reply button, it's not yet highlighted (as per the design). I'm confirming the exact colors just to be sure since the screen doesn't exist yet in Zeplin. Ref: `pbArwn-2ks#comment-4426`.
- Note that in the current implementation, like/unlike action is tracked after the network request succeeds. I'm planning to change this behavior in future PRs, such that the action is tracked regardless of the request result (similar to the comment detail in site comments). I'll also take a look later if an additional `commentID` parameter is needed, since currently only `siteID` and `postID` are tracked.
- As for the accessory button, I'll do it in a separate PR.

#### Reply functionality

- Ensure that the `newCommentThread` flag is enabled.
- Go to Reader > select any comment thread.
- Tap a Reply button on any comment.
- 🔍 Verify that the reply text view placeholder is updated with the comment author's name.
- Send the reply.
- 🔍 Verify that a new reply comment is added in the correct position.

#### Like functionality

- Tap the Like button on an unliked comment.
- 🔍 Verify that the Like button image is now highlighted and shows the correct number of likes.
- Go back to Reader, and enter the comment thread again.
- 🔍 Verify that the Like button for the previous comment is highlighted and still shows the correct number.
- Tap the Like button again to unlike.
- 🔍 Verify that the Like button image is now returned to its default state, and shows the correct number of likes.
- Go back to Reader, and enter the comment thread again.
- 🔍 Verify that the Like button for the previous comment is not highlighted and still shows the correct number.
- Switch the device to airplane mode (or turn off your wifi if using a Simulator).
- Tap on the Like button.
- 🔍 Verify that the app shows an alert, and the Like button remains unchanged.

#### Content links functionality

- (Don't forget to turn your connection on again! 😄 )
- Select any comment that contains a link, and tap the link.
- 🔍 Verify that a web view is presented, showing the contents of the link.

## Regression Notes
1. Potential unintended areas of impact
n/a. The feature is still hidden behind a flag.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a. The feature is still hidden behind a flag.

3. What automated tests I added (or what prevented me from doing so)
n/a. The feature is still hidden behind a flag.

PR submission checklist:

- [x] I have completed the Regression Notes.
[...]

PR #17601, opened 2021-11-30
Reader Comments: Overflow dropdown menu

dvdchr:
Refs #17475 
Depends on #17600

This PR adds the overflow dropdown menu when the content cell's accessory button is tapped while using an account with moderation roles. Some notes:

- Please note that showing a context menu (i.e. `UIMenu`) on tap is only supported in iOS 14 and above. This can be achieved by setting the `showsMenuAsPrimaryAction` to `true` and assigning a `UIMenu` to the `menu` property of `UIButton`. Currently, tapping on the ellipsis button does nothing in iOS 13. I'll need to come up with a fallback implementation for iOS 13, and will address this in a separate PR.
- Most of the tap actions are not yet implemented (except the Share button). This will be addressed in a separate PR.

Simulator Screen Shot - iPhone 12 - 2021-11-30 at 23 55 15

## To Test

Ensure that the `newCommentThread` flag is enabled.

#### With a moderator account:

- Go to Reader > any comment thread (from a site that you can moderate).
- Tap on the ellipsis button on any comment.
- 🔍 Verify that the context menu is correctly shown and anchored to the accessory view, with all the correct menu items.
- Tap on the Share menu.
- 🔍 Verify that a share sheet is presented, containing the comment's URL.

Note: As noted above, in iOS 13.0, tapping on the ellipsis button will do nothing for now.

#### With a non-moderator account:

- Go to Reader > any comment thread.
- Tap on the Share button on any comment.
- 🔍 Verify that a share sheet is presented, containing the comment's URL.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
+++ b/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
@@ -1083,6 +1083,33 @@ static NSString *CommentContentCellIdentifier = @"CommentContentTableViewCell";
     }];
 }
 
+- (void)presentMenuForComment:(Comment *)comment fromView:(UIView *)sourceView
+{
+    // create context menu.
+
+    UIAction *unapproveAction = [UIAction actionWithTitle:NSLocalizedString(@"Unapprove", @"Unapproves a comment")
+                                                    image:[UIImage systemImageNamed:@"x.circle"]
+                                               identifier:nil
+                                                  handler:^(__kindof UIAction * _Nonnull action) {
+        // TODO: Unapprove comment.
+    }];
+
+    UIAction *spamAction = [UIAction actionWithTitle:NSLocalizedString(@"Spam", @"Marks comment as spam")
+                                               image:[UIImage systemImageNamed:@"exclamationmark.octagon"]
+                                          identifier:nil
+                                             handler:^(__kindof UIAction * _Nonnull action) {
+        // TODO: Spam comment.
+    }];
+
+    UIAction *trashAction = [UIAction actionWithTitle:NSLocalizedString(@"Trash", @"Trashes the comment")
+                                                image:[UIImage systemImageNamed:@"trash"]
+                                           identifier:nil
+                                              handler:^(__kindof UIAction * _Nonnull action) {
+        // TODO: Trash comment.
+    }];
+
+}
+
 #pragma mark - Sync methods
 
 - (void)syncHelper:(WPContentSyncHelper *)syncHelper syncContentWithUserInteraction:(BOOL)userInteraction success:(void (^)(BOOL))success failure:(void (^)(NSError *))failure
@@ -1185,6 +1212,14 @@ static NSString *CommentContentCellIdentifier = @"CommentContentTableViewCell";
 
         // configure button actions.
         __weak __typeof(self) weakSelf = self;
+
+        cell.accessoryButtonAction = ^(UIView * _Nonnull sourceView) {
+            if ([comment allowsModeration]) {
+                // TODO: Show menu in iOS 13.
+            } else {
+                [self shareComment:comment sourceView:sourceView];
+            }
+        };
 
         cell.replyButtonAction = ^{
             [weakSelf didTapReplyAtIndexPath:indexPath];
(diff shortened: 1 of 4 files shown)


WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
```diff
@@ -89,4 +129,48 @@ private extension ReaderCommentsViewController {
...
+        case share
+
+        var title: String {
+            switch self {
+            case .unapprove:
+                return NSLocalizedString("Unapprove", comment: "Unapproves a comment")
+            case .spam:
+                return NSLocalizedString("Spam", comment: "Marks comment as spam")
+            case .trash:
+                return NSLocalizedString("Trash", comment: "Trashes the comment")
```
ScoutHarris: In Zeplin, these titles are:
- `Mark as Spam`
- `Mark as Trash`

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
```diff
@@ -1083,6 +1083,33 @@ - (void)didTapLikeForComment:(Comment *)comment atIndexPath:(NSIndexPath *)index
     }];
 }
 
+- (void)presentMenuForComment:(Comment *)comment fromView:(UIView *)sourceView
```
ScoutHarris: It appears this entire method isn't used yet, but it's causing several warnings since the `UIAction`s aren't used. Can this be removed for now to avoid unnecessary warnings in `develop`?

ScoutHarris:
When viewing the new threaded comments, I see a slew of these in the logs:
`[UILog] Called -[UIContextMenuInteraction updateVisibleMenuWithBlock:] while no context menu is visible. This won't do anything.`

I don't see them with the old threaded comments.

When the menu is shown, I see some of these as well:
`[UICollectionViewRecursion] cv == 0x7f7de89e2c00 Disabling recursion trigger logging`

I honestly don't know what they mean 😄 , but can they be resolved?

ScoutHarris (changes requested):
I left a few comments, but my request for changes is mainly due to the warnings produced by `presentMenuForComment`.

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
dvdchr: Thank you for noticing this! I don't know why I missed this 🤦 . This is fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17601/commits/4d621be16b989d3890b4581d540a62e4b094570e.

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m
dvdchr: That method is reserved to display the menu for iOS 13, but I've decided to split it into a different PR. Of course, I forgot to clean it up! 😬 Thanks for noticing this. Fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17601/commits/56fe10ea9a694a5cf2e70cebb96150fa7e4455a5.

dvdchr:
> `[UILog] Called -[UIContextMenuInteraction updateVisibleMenuWithBlock:] while no context menu is visible. This won't do anything.`

It looks like this is an issue with Xcode 13 targeting iOS 15. I've tried creating a standalone project that shows a menu, and the warning is still there. This doesn't appear when building for iOS 14.5, though!

> `[UICollectionViewRecursion] cv == 0x7f7de89e2c00 Disabling recursion trigger logging`

After searching around, this also seems to be [an issue specific to Xcode 13](https://www.hackingwithswift.com/forums/ios/disabling-recursion-trigger-logging/10036). This only appears when building for iOS 15.

Ready for another round of review, thank you! 🙂

WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.swift
ScoutHarris: Well it's a good thing you checked because I got it wrong too! 😆

ScoutHarris:
Thanks for looking into those warnings!

ScoutHarris (approved):
👍 !

:shipit:

PR #17607, opened 2021-12-01
Tweak UI tests to attempt addressing inconsistent failures

mokagio:
I've seen quite a few UI tests failures in CI over the past couple of weeks, e.g. [this](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/3872) and [this](https://app.circleci.com/pipelines/github/wordpress-mobile/WordPress-iOS/26094/workflows/54c60844-a868-4955-b857-f5c3ffdbc48e/jobs/61817).

Based on the CI logs, I tracked down some of them to the editor block picker being still on screen after a test failure and hijacking the rest of the tests.

Screen Shot 2021-12-01 at 9 25 49 am
 
These changes are an attempt to make the tests robust against this kind of failure.

I developed them by forcing a failure in the `addBlock(_:)` method:

```diff
@@ -131,7 +131,7 @@ public class BlockEditorScreen: ScreenObject {
     private func addBlock(_ blockLabel: String) {
         addBlockButton.tap()
         let blockButton = app.buttons[blockLabel]
-        XCTAssertTrue(blockButton.waitForIsHittable(timeout: 3))
+        XCTAssertFalse(blockButton.waitForIsHittable(timeout: 3))
         blockButton.tap()
     }
```

### DIFF
--- a/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
@@ -130,8 +130,8 @@ public class BlockEditorScreen: ScreenObject {
     }
 
     private func addBlock(_ blockLabel: String) {
-        let blockButton = app.buttons[blockLabel]
         addBlockButton.tap()
+        let blockButton = app.buttons[blockLabel]
         XCTAssertTrue(blockButton.waitForIsHittable(timeout: 3))
         blockButton.tap()
     }
--- a/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
+++ b/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
@@ -91,12 +91,11 @@ public class BlockEditorScreen: ScreenObject {
             editorCloseButton.tap()
 
             XCTContext.runActivity(named: "Discard any local changes") { (activity) in
-                let notSavedState = app.staticTexts["You have unsaved changes."]
-                if notSavedState.exists {
-                    Logger.log(message: "Discarding unsaved changes", event: .v)
-                    let discardButton = app.buttons["Discard"] // Uses a localized string
-                    discardButton.tap()
-                }
+                guard app.staticTexts["You have unsaved changes."].waitForIsHittable(timeout: 3) else { return }
+
+                Logger.log(message: "Discarding unsaved changes", event: .v)
+                let discardButton = app.buttons["Discard"] // Uses a localized string
+                discardButton.tap()
             }
 
             let editorNavBar = app.navigationBars["Gutenberg Editor Navigation Bar"]
@@ -134,6 +133,19 @@ public class BlockEditorScreen: ScreenObject {
         let blockButton = app.buttons[blockLabel]
         XCTAssertTrue(blockButton.waitForIsHittable(timeout: 3))
         blockButton.tap()
+    }
+
+    /// Some tests might fail during the block picking flow. In such cases, we need to dismiss the
+    /// block picker itself before being able to interact with the rest of the app again.
+    public func dismissBlocksPickerIfNeeded() {
+        // Determine whether the block picker is on screen using the visibility of the add block
+        // button as a proxy
+        guard addBlockButton.isFullyVisibleOnScreen == false else { return }
+
+        // Dismiss the block picker by swiping down
+        app.swipeDown()
+
+        XCTAssertTrue(addBlockButton.waitForIsHittable(timeout: 3))
     }
 
     /*
--- a/WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
+++ b/WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
@@ -25,8 +25,8 @@ class EditorGutenbergTests: XCTestCase {
 
     func testTextPostPublish() throws {
 
-        let title = getRandomPhrase()
-        let content = getRandomContent()
+        let title = "Text post title"
+        let content = "Text post content"
         try editorScreen
             .dismissNotificationAlertIfNeeded(.accept)
             .enterTextInTitle(text: title)
@@ -39,8 +39,8 @@ class EditorGutenbergTests: XCTestCase {
 
     func testBasicPostPublish() throws {
 
-        let title = getRandomPhrase()
-        let content = getRandomContent()
+        let title = "Rich post title"
+        let content = "Some text, and more text"
         let category = getCategory()
         let tag = getTag()
         try editorScreen
--- a/WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
+++ b/WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
@@ -15,7 +15,9 @@ class EditorGutenbergTests: XCTestCase {
 
     override func tearDownWithError() throws {
         takeScreenshotOfFailedTest()
+
         if editorScreen != nil && !TabNavComponent.isVisible() {
+            editorScreen.dismissBlocksPickerIfNeeded()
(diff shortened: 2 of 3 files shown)


WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -91,12 +91,11 @@ public class BlockEditorScreen: ScreenObject {
...
-                let notSavedState = app.staticTexts["You have unsaved changes."]
-                if notSavedState.exists {
-                    Logger.log(message: "Discarding unsaved changes", event: .v)
-                    let discardButton = app.buttons["Discard"] // Uses a localized string
-                    discardButton.tap()
-                }
+                guard app.staticTexts["You have unsaved changes."].waitForIsHittable(timeout: 3) else { return }
+
+                Logger.log(message: "Discarding unsaved changes", event: .v)
+                let discardButton = app.buttons["Discard"] // Uses a localized string
```
mokagio, line 97: I'm not sure whether that comment is accurate. I decided to leave it there because I didn't have the time to investigate.

WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -91,12 +91,11 @@ public class BlockEditorScreen: ScreenObject {
             editorCloseButton.tap()
 
             XCTContext.runActivity(named: "Discard any local changes") { (activity) in
-                let notSavedState = app.staticTexts["You have unsaved changes."]
-                if notSavedState.exists {
-                    Logger.log(message: "Discarding unsaved changes", event: .v)
-                    let discardButton = app.buttons["Discard"] // Uses a localized string
-                    discardButton.tap()
-                }
+                guard app.staticTexts["You have unsaved changes."].waitForIsHittable(timeout: 3) else { return }
```
mokagio, line 94: In my tests, this code didn't run once, despite me seeing the sheet in the Simulator. I replaced the `.exists` check with a `waitFor...` to account for possible display time lags.

WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift
```diff
@@ -130,12 +129,25 @@ public class BlockEditorScreen: ScreenObject {
     }
 
     private func addBlock(_ blockLabel: String) {
-        let blockButton = app.buttons[blockLabel]
         addBlockButton.tap()
+        let blockButton = app.buttons[blockLabel]
```
mokagio, line 133: I don't think this will make much of a difference because we do call the `waitForIsHittable(_:)` method on the object. Still, it makes the test flow a bit clearer, IMHO.

First we tap the add block button, then we look for the block button for the given label

WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
```diff
@@ -15,7 +15,9 @@ class EditorGutenbergTests: XCTestCase {
 
     override func tearDownWithError() throws {
         takeScreenshotOfFailedTest()
+
         if editorScreen != nil && !TabNavComponent.isVisible() {
+            editorScreen.dismissBlocksPickerIfNeeded()
             EditorFlow.returnToMainEditorScreen()
             editorScreen.closeEditor()
```
mokagio, line 22: I thought about moving the `dismissBlockPickerIfNeeded() call in the `closeEditor()` method, but it wouldn't have worked because we need to call `EditorFlow.returnToMainEditorScreen()` first. `EditorFlow` exists the UI tests target but not in the UITestsFoundation framework, so I decided to leave it as it is.

WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
```diff
@@ -25,8 +27,8 @@ class EditorGutenbergTests: XCTestCase {
 
     func testTextPostPublish() throws {
 
-        let title = getRandomPhrase()
-        let content = getRandomContent()
+        let title = "Text post title"
+        let content = "Text post content"
```
mokagio, line 31: Random values in tests are great to catch weird behavior and input edge cases that the test author might not have considered at the time of writing.

But, one needs a way to repeat the test with the same input in order to reproduce the issue. Otherwise, we risk getting flaky tests with no way of knowing why.

Without a configurable seed from which to generate random values, random input can be dangerous for the consistency of the test suite.

Another reason I'm making this change is because the Simulator takes a long time to type such long inputs one character at a time. _That's the reason why I ended up looking at this code in the first place._

WordPress/WordPressUITests/Utils/XCTest+Extensions.swift
```diff
@@ -24,10 +24,10 @@ extension XCTestCase {
...
 
     public func takeScreenshotOfFailedTest() {
-        if let failureCount = testRun?.failureCount, failureCount > 0 {
-            XCTContext.runActivity(named: "Take a screenshot at the end of a failed test") { (activity) in
-                add(XCTAttachment(screenshot: XCUIApplication().windows.firstMatch.screenshot()))
-            }
+        guard let failuresCount = testRun?.failureCount, failuresCount > 0 else { return }
+
+        XCTContext.runActivity(named: "Take a screenshot at the end of a failed test") { activity in
+            add(XCTAttachment(screenshot: XCUIApplication().windows.firstMatch.screenshot()))
```
mokagio, line 30: Same code, just using `guard` instead of `if` to avoid nested indentations.

mokagio (commented):
Now I only need to wait and see what CI has to say about these changes...

mokagio:
😔 I got this failure [in Buildkite](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/3906#75686c2d-dc7b-4e06-b511-799b75368e20):

```
[22:47:50]: ▸ WordPressUITests.EditorGutenbergTests
[22:47:50]: ▸   testBasicPostPublish, Failed to get matching snapshot: No matches found for Elements matching predicate '"View" IN identifiers' from input {(
[22:47:50]: ▸   <unknown>:0
```

🤔 not too sure where in the flow this was exactly, maybe here:

https://github.com/wordpress-mobile/WordPress-iOS/blob/72ad4801c567ffff6dfa25b39539bbf330db4f77/WordPress/UITestsFoundation/Screens/Editor/BlockEditorScreen.swift#L120

On the other hand, that was the only failure I got, and it's different from the ones this PR is trying to address 😄

WordPress/WordPressUITests/Tests/EditorGutenbergTests.swift
tiagomar: Yes, good point about the reproducibility.

pachlava: > Without a configurable seed from which to generate random values, random input can be dangerous for the consistency of the test suite.

This is a good point, thanks for the extra context!

> because the Simulator takes a long time to type such long inputs one character at a time

So far I'm using iPhone 12 with iOS 14.5 as a workaround, but actually typing is the only thing it does faster. From what I saw recently, the execution of tests that do not involve typing is noticeably faster on iPhone 13.

pachlava (approved):
Even though I never saw such kind of fail locally, the changes make sense to me and should definitely improve the stability. Firstly, I executed all tests as they are (on iPhone 13) and they passed, and then with intentional fail injected as you mentioned in the description. And the dismissal of block picker worked 👍 Thank you!

tiagomar (approved):
I tested with iPhone 13 and iPad 4th gen and it works as expected. I haven’t reproduced the issue naturally but forcing it to fail without `dismissBlocksPickerIfNeeded()` had definitely put the execution into an unrecoverable state, not allowing the other tests to run. This piece of code will surely enhance the stability.

PR #17622, opened 2021-12-03
Reader Comments: Add fallback implementation for overflow menu on iOS 13

dvdchr:
Refs #17476 

This PR adds `MenuSheetViewController`, which is used to mimic the appearance of `UIContextMenu` in iOS 13. The main difference is that this can be fully customizable – can be triggered on tap, or shown programmatically – unlike `UIContextMenu` which can only be shown on long press.

The component is displayed using a `UIPopoverPresentationController` with a "forced" setting on iPhone such that it always shows the menu with `.popover` presentation style (instead of `.formSheet`). 

Here's a preview:

• | `UIMenu`, iOS 14+ | Fallback implementation, iOS 13
--|--|--
Light scheme | menu_light | fallback_light
Dark scheme | menu_dark | fallback_dark

There are slight differences visually (positioning, caret, visual effects, shadows, etc.), but I'm thinking this is good enough to be implemented in a short amount of time.

## To Test

Ensure that the `newCommentThread` feature flag is enabled.

#### With iOS 14+:

- Go to any comment thread on a site that you can moderate. (For example, My Site > Comments > tap any comment reply > tap on the header.)
- Tap on the ellipsis button.
- 🔍  Verify that the `UIMenu` is shown.
- Tap on the Share menu.
- 🔍  Verify that the share sheet is presented.

#### With iOS 13:

(Tip: if you don't have iOS 13 simulator downloaded, you can temporarily disable the code in ReaderCommentsViewController.swift, particularly the if #available lines.)

- Go to any comment thread on a site that you can moderate. (For example, My Site > Comments > tap any comment reply > tap on the header.)
- Tap on the ellipsis button.
- 🔍  Verify that the menu is shown.
- Tap on the Share menu.
- 🔍  Verify that the share sheet is presented.

## Regression Notes
n/a. Feature is unreleased.

n/a. Feature is unreleased.

n/a. Feature is unreleased.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Reader/MenuSheetViewController.swift
+++ b/WordPress/Classes/ViewRelated/Reader/MenuSheetViewController.swift
@@ -1,0 +1,97 @@
+import UIKit
+import WordPressUI
+
+/// Provides a fallback implementation for showing `UIMenu` in iOS 13. Instead of showing a floating menu, this embeds
+/// the menu in a table view through `BottomSheetViewController`. Note that to simplify things, nested elements will be
+/// displayed as if `UIMenuOptions.displayInline` is applied.
+///
+/// In iOS 13, `UIMenu` can only appear through long press gesture. There is no way to make it appear programmatically
+/// or through different gestures. However, in iOS 14 menus can be configured to appear on tap events. Refer to
+/// `showsMenuAsPrimaryAction` for more details.
+///
+/// TODO: Remove this component (and its usage) in favor of `UIMenu` when the minimum version is bumped to iOS 14.
+///
+class MenuSheetViewController: UITableViewController {
+
+    struct MenuItem {
+        let title: String
+        let image: UIImage?
+        let handler: () -> Void
+    }
+
+    private let itemSource: [[MenuItem]]
+
+    // MARK: Lifecycle
+
+    required init(items: [[MenuItem]]) {
+        self.itemSource = items
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        configureTable()
+    }
+}
+
+// MARK: - Table View
+
+extension MenuSheetViewController {
+
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        return itemSource.count
+    }
+
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        guard section < itemSource.count else {
+            return 0
+        }
+        return itemSource[section].count
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let item = itemSource[indexPath.section][indexPath.row]
+        let cell = tableView.dequeueReusableCell(withIdentifier: Constants.cellIdentifier, for: indexPath)
+
+        cell.textLabel?.setText(item.title)
+        cell.accessoryView = UIImageView(image: item.image?.withTintColor(.text))
+        return cell
+    }
+
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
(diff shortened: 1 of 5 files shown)


ScoutHarris:
Hey @dvdchr . Got a few issues for ya.

**iOS13:**

Menu disappears on rotation:
- In portrait, show the menu on a comment close to the bottom of the list.
- Rotate to landscape.
- Notice the menu is not shown (I'm assuming because the comment isn't).
- The menu is shown again when rotating back to portrait.
- Rotate a few more times. Notice the app becomes unresponsive. It doesn't appear to crash, it just stops responding.

| missing_portrait | missing_land |
|--------|-------|


Menu is misplaced on rotation:
- In portrait, show the menu on a comment in the middle of the list.
- Rotate to landscape.
- Notice the menu is not inline with the comment button.

| misplaced_port | misplaced_land |
|--------|-------|

Menu UI:
- There appears to be an extra row on top of the menu.

extra_row


**iOS14.5:**

Menu disappears on rotation:
- In portrait, show the menu on any comment.
- Rotate to landscape.
- Notice the menu is not shown.
- Rotating back to portrait doesn't make it reappear as it does in iOS13.

ScoutHarris (changes requested):
Hey @dvdchr . I know this is a WIP, but I'm requesting changes primarily due to the app becoming unresponsive in iOS13.

dvdchr:
Thanks for the details and screenshots @ScoutHarris ! 

> iOS 13:
> 
> Menu disappears on rotation:

When fixing this, I noticed that it will be a totally different approach with iOS 14.5 since it uses a `UIMenu`. However, after searching around, it looks like the system intentionally dismisses menus on orientation change – even when it changes from landscape to landscape. I've tried this by going to Apple's Messages > conversation screen in landscape mode. While in split view, I tapped on the ellipsis menu and it also shows a `UIMenu`. 

Therefore, I figured it's best for the fallback implementation to follow the `UIMenu` behavior instead. This has been addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17622/commits/d7ba1ff1e9fc6e10bceedc28c3d0506bb2e6f862. Note that this should also address these comments:

- iOS 13: Menu is misplaced on rotation -> Menu will now be dismissed on rotation.
- iOS 14.5: Menu disappears on rotation -> This is expected behavior.

> Menu UI:
>
> There appears to be an extra row on top of the menu.

Thank you for noticing this! I've initially fixed this, but somehow it got removed in [this commit](https://github.com/wordpress-mobile/WordPress-iOS/pull/17622/commits/ec95d71186ddb873f2e4d8768ceea04ebf6fe7c8#diff-ef2b4c45e4793d7105ec1bd0ec1835763e183a7fef4b0fc56eaf7328109a9d7cL114). I've re-added this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17622/commits/1eb9d2434e70e03c3e0f6a1227835b5d656a8282. 🙂 

> Rotate a few more times. Notice the app becomes unresponsive. It doesn't appear to crash, it just stops responding.

Now, this is an interesting one. After investigating this, it's actually unrelated to the menu and it also happens in iOS 14+. The  problem is, the `ReaderCommentsViewController` has a constant, `EstimatedCommentRowHeight`, which is used in the `estimatedHeightForRow` method. When scrolling through the new comment threads when it has a lot of text per comments, the `prepareForReuse` and `configureCell` would get stuck in an infinite loop and this caused the app to be unresponsive 😬 .

In  https://github.com/wordpress-mobile/WordPress-iOS/pull/17622/commits/5887a6106ce9f689d601d132734041ebc38683aa, I've updated the `estimatedHeightForRow` to use automatic dimension addresses this issue. The thread still jumps around as you scroll, but I think this is a lot better than turning unresponsive.

Ready for another round of review!

ScoutHarris:
> iOS 13: Menu is misplaced on rotation -> Menu will now be dismissed on rotation.
> iOS 14.5: Menu disappears on rotation -> This is expected behavior.

Sounds good. ➕ 1️⃣ for consistency. 😄 

> The thread still jumps around as you scroll, but I think this is a lot better than turning unresponsive.

Agreed!

ScoutHarris (approved):

dvdchr:
Thanks @ScoutHarris !

PR #17628, opened 2021-12-06
Test PR to verify `isFullyVisibleOnScreen` removal

mokagio:


WordPress/UITestsFoundation/XCUIElement+Scroll.swift
```diff
@@ -11,12 +12,12 @@ extension XCUIElement {
     public func scrollIntoView(within scrollView: XCUIElement, threshold: Int = 1000) {
         var iteration = 0
 
-        while !isFullyVisibleOnScreen && iteration < threshold {
+        while isFullyVisibleOnScreen() && iteration < threshold {
```
tiagomar: I know this is the test PR, but was this logic change intentional? `!isFullyVisibleOnScreen` -> `isFullyVisibleOnScreen()`

I would expect some tests failing because of that, but I remember seeing cases where `isFullyVisibleOnScreen` returned `true` for an element that was partially visible. 🤔

mokagio: Great catch @tiagomar! This is purely a mistake on my end.

This actually made me smile because I left a comment earlier on `!foo` vs `foo == false` and how `!` can be missed, which is _exactly_ the error I made here. Case in point, I guess 🙃 

I will fix it anyway, because I plan to cherrypick this commit in the PR that will update XCUITestHelper to a the new version once HACK week is done 👍

mokagio:
Closing in favor of #17691

PR #17623, opened 2021-12-03
Reader Comments: Implement edit action on the overflow menu

dvdchr:
Refs #17476 
Depends on #17622

As titled, the Edit button should now be functional. 

I've made a small change to `CommentService`. Comments fetched from the reader endpoint don't have their `blog` property assigned, and this caused the `uploadComment:` method to fail silently due to an empty `comment.blog`. To solve this, I added `remoteForComment:` which can now create `CommentServiceRemote` for both Site comments and Reader comments.

If the user can update a comment from the comment thread, that means they have a moderator role assigned to their account. And that means they should have the blog/site synchronized locally before opening the comment thread.

## To test

- Ensure that `newCommentThread` feature flag is enabled.
- Go to any comment thread on a site that you can moderate. (For example, My Site > Comments > tap any comment reply > tap on the header.)
- Tap on the ellipsis button, and tap on the Edit menu.
- 🔍 Verify that the Edit Comment screen is shown.
- Edit the comment, and tap Save.
- 🔍 Verify that the edited comment in the comment thread is updated with the new value.
- Go back to My Site > Comments.
- 🔍 Verify that the comment is also updated in this list.

## Regression Notes
1. Potential unintended areas of impact
n/a. Feature is hidden behind a flag.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a. Feature is hidden behind a flag.

3. What automated tests I added (or what prevented me from doing so)
n/a. Feature is hidden behind a flag.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17629, opened 2021-12-06
Reader Comments: Implement moderation actions

dvdchr:
Refs #17476
Depends on #17623

This PR adds functionality to the moderation menu items. 
- After a comment is moderated, it should be removed from the table view after the request succeeds. 
- Note that this only hides the moderated comment. The correct behavior is that all descendants of non-approved comments should be hidden, but the API request still returns approved descendants which could cause data inconsistency. Therefore I'll hold off on this implementation until we can also align with the backend side.
- I've also updated the `ReaderPost`'s comment count (decrementing it by 1) when a comment is moderated. 
  - Note that I've decided not to implement the other way around – i.e. incrementing the ReaderPost's comment count when the comment is approved through My Site > Comments. This is because I feel like it's less noticeable since the action is performed from different tabs, and the comment count will eventually get updated if the user opens the post or comment thread.

## To test

- Ensure that the `newCommentThread` flag is enabled.
- Navigate to Reader > any comment thread that you can moderate.
- Tap on the ellipsis button, and select "Unapprove". 
- Verify that the comment is removed from the table view, and a notice is shown.
- Go back to Reader.
- Verify that the comment count in the Reader list is decremented by one.
- Repeat the steps above for "Mark as spam", and "Move to trash" actions.

## Regression Notes

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Services/CommentService.m
+++ b/WordPress/Classes/Services/CommentService.m
@@ -913,7 +913,7 @@ static NSTimeInterval const CommentsRefreshTimeoutInSeconds = 60 * 5; // 5 minut
 
     comment.status = currentStatus;
     [[ContextManager sharedInstance] saveContext:self.managedObjectContext];
-    id <CommentServiceRemote> remote = [self remoteForBlog:comment.blog];
+    id <CommentServiceRemote> remote = [self remoteForComment:comment];
     RemoteComment *remoteComment = [self remoteCommentWithComment:comment];
     NSManagedObjectID *commentID = comment.objectID;
     [remote moderateComment:remoteComment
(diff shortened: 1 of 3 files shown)


ScoutHarris:
I noticed one small issue that probably pre-dates your changes. 
- In the threaded view, moderate a comment. It disappears as expected.
- Select the My Site tab (i.e. don't navigate back via the nav bar). 
- Go to Comments and approve the comment (or any comment in that thread I suppose).
- Select the Reader tab. 
- Since the last Reader view you looked at was threaded comments, it is still displayed. But the newly approved comment is not.

It would be nice to refresh the view in `viewWillAppear` instead of `viewDidLoad`. I didn't fully test it, but a quick test shows it might just be a matter of moving the `[self refreshAndSync]` call.

ScoutHarris (approved):
I left one suggestion for consideration. Otherwise 👍 !

:shipit:

dvdchr:
> It would be nice to refresh the view in viewWillAppear instead of viewDidLoad. I didn't fully test it, but a quick test shows it might just be a matter of moving the [self refreshAndSync] call.

Thanks for the suggestion! Also, this way the comment count for the post will be correct since we're syncing once the user comes back to the thread. This is addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17629/commits/84490b6445da7038cacf89de30fb95131811206a. 🙂

PR #17634, opened 2021-12-07
Reader Comments: Refactor content cell to have configurable render method

dvdchr:
Refs #17476

This PR refactors `CommentContentTableViewCell` so it could render its content in several different ways. 

- The web view in the cell is now replaced with an empty UIView. Depending on the rendering method, a WKWebView might be added to the UIView – or it could be a UITextView later on.
- No new method has been added just yet, so this should still render the web view as before.
- As this is a non-user facing change, please ensure that things look the same as before.

#### Site Comments

- Go to My Site > Comments.
- Select any comment.
- 🔍  Verify that the comment cell is displayed correctly and is not impacted by this change.

#### Comment Threads

- Ensure that the `newCommentThread` is enabled.
- Go to Reader > any comment threads.
- 🔍 Verify that the comment threads are rendered with the web view method (same as before), and not impacted by this change.

## Regression Notes
1. Potential unintended areas of impact
n/a. This feature is still hidden behind a flag.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
n/a. This feature is still hidden behind a flag.

3. What automated tests I added (or what prevented me from doing so)
n/a. This feature is still hidden behind a flag.

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17635, opened 2021-12-07
Reader Comments: Implement comment rendering via WPRichContent

dvdchr:
Refs #17476
Depends on #17634

As titled, this PR reintroduces the previous content rendering method through `WPRichContentView` into the comment thread.  Note that all pre-existing issues with the `WPRichContentView` (e.g. enlarged emojis, extra white spaces, partially clipped text) are not fixed here.

Here's a preview:
Simulator Screen Shot - iPhone 12 - 2021-12-07 at 22 09 36

---

I also fixed a minor visual issue regarding table view separators. Previously the separators were hidden – but now since we are showing a separator, the separators between empty cells show while the loading view is shown. This PR fixes that issue.

Before | After
--|--
before | after

## To test

- Ensure that the `newCommentThread` flag is enabled.
- Go to Reader > any comment thread.
- 🔍 Verify that the comment body is displayed with the old appearance, but contained in the new comment cell.
- 🔍 Tap on the overflow menu and verify that the menu items work.
- 🔍 Verify that the reply button works.
- 🔍 Verify that the like button works.
- Find a comment that contains a link in it.
- 🔍  Tap the link and verify that a modal window is presented and it loads the tapped URL.
- Find a comment that contains an image (or some).
- 🔍  Tap on the image and verify that a modal window is presented and it loads the image.

#### Verifying the table separator issue

- Go to Notifications, and tap any comment reply notification.
- Tap the notification header.
- 🔍 Verify that no table separators are shown while the loading view is visible. 

## Regression Notes

PR submission checklist:


WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -170,7 +179,7 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
     /// - Parameters:
     ///   - comment: The `Comment` object to display.
     ///   - onContentLoaded: Callback to be called once the content has been loaded. Provides the new content height as parameter.
-    func configure(with comment: Comment, onContentLoaded: ((CGFloat) -> Void)?) {
+    func configure(with comment: Comment, renderMethod: RenderMethod = .web, onContentLoaded: ((CGFloat) -> Void)?) {
```
ScoutHarris, line 187: It'd be nice to add `renderMethod` to the params list in the method description above.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
```diff
@@ -214,13 +223,24 @@ class CommentContentTableViewCell: UITableViewCell, NibReusable {
         isAccessoryButtonEnabled = false
     }
 
+    @objc func ensureRichContentTextViewLayout() {
+        guard let renderer = renderer,
```
ScoutHarris: This throws a warning:
`Immutable value 'renderer' was never used; consider replacing with '_' or removing it`

ScoutHarris:
It looks like the layout isn't sized quite right sometimes.

Moderation in post threaded comments:
- In the threaded view, moderate a comment. It disappears as expected.
- Select the My Site tab (i.e. don't navigate back via the nav bar). 
- Go to Comments and approve the comment.
- Select the Reader tab. 
- The comment does reappear in the threaded view, but sometimes the UI isn't quite right. I couldn't repro it 100% of the time, but I did see it a couple of times.

Screen Shot 2021-12-07 at 1 11 37 PM


Notification Comments:
- On a mention comment notification, tap the header to show the threaded comments.
- Sometimes the reaction bar is cut off or duplicated:


Screen Shot 2021-12-07 at 1 50 15 PM

It seems to only happen when the comments are initially loaded because after the first viewing the UI is fine. But it is 100% reproducible when the comments are reloaded.

ScoutHarris:
I can repro this with the old threaded comments, so it pre-exists your changes. But it would be nice to fix this at some point. Sometimes the loading view still appears after the comments are loaded:

Simulator Screen Shot - iPhone 12 Pro Max - 2021-12-07 at 13 32 29

Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-12-07 at 13 16 54

ScoutHarris (changes requested):

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Thank you! I forgot about this. This is now fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17635/commits/54d99efd3249066902638b9bc55b9df240a85f94.

WordPress/Classes/ViewRelated/Comments/CommentContentTableViewCell.swift
dvdchr: Whoops! I missed this one. It's now addressed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17635/commits/54d99efd3249066902638b9bc55b9df240a85f94. Thanks!

dvdchr:
Thank you for the details! 👍 

> Sometimes the loading view still appears after the comments are loaded:

This should be fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17635/commits/a7b7fcd8c3fdb1d5f15050b4a65eaf163fa5f5b9. With this, I've also addressed the wrong NRV placement in iPad. Here's a preview:

• | Before | After
--|--|--
Portrait | ipad_portrait_before | ipad_portrait_after
Landscape | ipad_landscape_before | ipad_landscape_after

> Sometimes the reaction bar is cut off or duplicated:

Yeah, it looks like this is a bug that has been haunting our threads 😅 . I've investigated this and filed an issue here: https://github.com/wordpress-mobile/WordPress-iOS/issues/17653. I've also attempted to address this in https://github.com/wordpress-mobile/WordPress-iOS/pull/17635/commits/7d2fad2ffbc3ac84bce8b7b1b22ae30ec6b1c615, since we're going to turn this on for the public anyways this upcoming release and I'd want to prevent side effects when we have to turn off the flag.

Here's a preview:

• | Before | After
--|--|--
Portrait | 17635_ipad_portrait_before | 17635_ipad_portrait_after
Landscape | 17635_ipad_landscape_before |  17635_ipad_landscape_after

As for the case with moderation and the Reader tab, I've been unable to reproduce it so far.  If this still appears, I'll give it a go again next week. Ready for your next round of review, please! 🙂

WordPress/Classes/ViewRelated/Views/WPRichText/WPRichContentView.swift
```diff
@@ -178,7 +178,6 @@ private extension WPRichContentView {
             "em, i { font:-apple-system-body; font-family: 'Noto Serif'; font-weight: normal; font-style: italic; line-height:1.6; } " +
             "a { color: #\(style.linkColorHex); text-decoration: none; } " +
             "a:active { color: #\(style.linkColorActiveHex); } " +
-            "img.emoji { vertical-align: baseline; max-width: 1rem; } " +
```
dvdchr, line 181: This was an experiment that I forgot to clean up. This line doesn't do anything to the `WPRichContentView`. 😅

ScoutHarris:
Well, I see the NRV is misplaced on iPad:

<kbd>ipad</kbd>

It could be fixed separately if you like since it's not really related to this PR.

ScoutHarris (approved):

dvdchr:
> Well, I see the NRV is misplaced on iPad:

Thank you! I was only testing this on iOS 14.5 and couldn't figure out why it isn't happening on my end. Turns out, it happens on iOS 15 – and guess what: iOS 15 decides to call `traitCollectionDidChange:` on push navigation. The `ReaderCommentsViewController` happens to call `[self refreshTableViewAndNoResultsView]` on this method, which is the root cause of the misalignment. Life of an iOS engineer, indeed. 😄 

It should be fixed in https://github.com/wordpress-mobile/WordPress-iOS/pull/17635/commits/eea5e108442aac01e425992a7ef3db35ddaed21f, and I'll merge this ahead for now. If there are any further issues, I'll address them through a separate PR. 🙂

PR #17116, opened 2021-09-03
Implement weekly roundup toggle

diegoreymendez:
Implements the Weekly Roundup notification toggle, per-site.
Enabled the Weekly Roundup feature flag.
Fixes a bug that we had in Weekly Roundup that could cause an issue with the chosen sites for the notifications.

### First, check that the notification works as usual:

1. Use the debug menu for Weekly Roundup to toggle an immediate notification.
2. Take not of the sites you see the notifications for.
3. If you feel like it, repeat the notification and make sure the sites and numbers stay the same as before.

### Then check that you can disable the notifications for a site:

1. Using the list from the previous test, go to the Notifications tab > Gear > Push Notifications and disable Weekly Roundup for one of the top 5 sites.
2. Schedule the immediate notification again.
3. When the notifications come up, check that the site you disabled notifications for has been replaced by another site.

### Extra:

If you want, you can check that if the Weekly Roundup feature flag is OFF, that the toggle is not shown in notifications.

## Regression Notes

1. Potential unintended areas of impact

This PR only affects Weekly Roundup which is a new feature.

2. What I did to test those areas of impact (or what existing automated tests I relied on)
3. What automated tests I added (or what prevented me from doing so)

PR submission checklist:

- [x] I have completed the Regression Notes.
- [x] I have considered adding unit tests for my changes.
- [x] I have considered adding accessibility improvements for my changes.
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #17640, opened 2021-12-08
Use derived context for WeeklyRoundup background task

### LINKED ISSUE #17639: NSGenericException: Collection __NSCFSet was mutated while being enumerated
Sentry: https://sentry.io/share/issue/68ce8485d27d458bbe59d24ab180c8c8/
Discussion: p1638850640127900-slack-C012H19SZQ8

```
NSGenericException

*** Collection <__NSCFSet: 0x281eb7b10> was mutated while being enumerated.
```

## Findings

The error message indicates that the managed object context is being accessed from a different thread while another operation was running in its original thread. When the crash happened, these are the most common operations running on the main thread:
- `-[AccountService accountWithUUID:]` (Classes/Services/AccountService.m:286)
  - Called by `WPTabBarController.makeApi`
- `-[ReaderTopicService allTopics]` (Classes/Services/ReaderTopicService.m:1111)
- `PlanService.allPlanFeatures` (PlanService.swift:97)

Based on the stack trace, it could be related to this part of the code: https://github.com/wordpress-mobile/WordPress-iOS/blob/1e7bcd213b970b44c67d9d0a7970d498c92cbb9f/WordPress/Classes/Utility/BackgroundTasks/WeeklyRoundupBackgroundTask.swift#L163-L167

- The `NotificationSettingsService` is instantiated with a `mainContext`, which in turn creates an `AccountService` with the same context – leading to the crash at `-[AccountService accountUUID:]`. 
- However, I did notice that the `WeeklyRoundupBackgroundTask` holds an `NSManagedObjectContext` property that's also initialized with a `mainContext` and it's being used everywhere else within the class. Strangely enough, `filterWeeklyRoundupEnabledSites > NotificationSettingsService > AccountService` is the only place where the crash happened.
  - Note that there's another Core Data fetch in the `getAllSites()` method but that seemed to work just fine despite having a `mainContext`. https://github.com/wordpress-mobile/WordPress-iOS/blob/1e7bcd213b970b44c67d9d0a7970d498c92cbb9f/WordPress/Classes/Utility/BackgroundTasks/WeeklyRoundupBackgroundTask.swift#L194-L199 The only difference with `AccountService` is that it creates an `NSFetchRequest` directly instead of through an `NSFetchedResultsController`. I'm not sure if there are any quirky differences between the two, though.

## Repro Steps

Unfortunately, I wasn't able to reproduce this. I tried following the feature testing steps described in #17066 and #17116, but since the crash is intermittent, this could only happen under specific circumstances.

Based on the crash information, it happens when there's a Core Data activity on the main thread (maybe the app fetching some stuff from the Core Data to display) while the background task is running.

dvdchr:
Fixes #17639 

This attempts to fix a crash due to the managed object context being accessed from the wrong thread. The `WeeklyRoundupDataProvider` is now initialized with a derived context so all of its operations won't be running on the main context and hopefully prevents overlap with any operations running in the main queue.

To test:

Refer to the steps described in #17066 and #17116, and make sure everything works.

## Regression Notes
The Weekly Roundup feature might not work as expected.

Manually tested the scenarios described in #17066 and #17116.

n/a.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/Utility/BackgroundTasks/WeeklyRoundupBackgroundTask.swift
+++ b/WordPress/Classes/Utility/BackgroundTasks/WeeklyRoundupBackgroundTask.swift
@@ -163,7 +163,7 @@ class WeeklyRoundupDataProvider {
     /// Filters the sites that have the Weekly Roundup notification setting enabled.
     ///
     private func filterWeeklyRoundupEnabledSites(_ sites: [Blog], result: @escaping (Result<[Blog], Error>) -> Void) {
-        let noteService = NotificationSettingsService(managedObjectContext: ContextManager.sharedInstance().mainContext)
+        let noteService = NotificationSettingsService(managedObjectContext: context)
 
         noteService.getAllSettings { settings in
             let weeklyRoundupEnabledSites = sites.filter { site in
@@ -377,7 +377,7 @@ class WeeklyRoundupBackgroundTask: BackgroundTask {
             }
         }
 
-        let dataProvider = WeeklyRoundupDataProvider(context: ContextManager.shared.mainContext, onError: onError)
+        let dataProvider = WeeklyRoundupDataProvider(context: ContextManager.shared.newDerivedContext(), onError: onError)
         var siteStats: [Blog: StatsSummaryData]? = nil
 
         let requestData = BlockOperation {


mokagio:
@dvdchr how do you feel about targeting the release branch for this?

The metrics for the crash look encouraging in that the impact is limited, but, because of the December release pause, 18.9 is almost a month away from reaching the users.

Screen Shot 2021-12-08 at 10 28 33 am

dvdchr:
> how do you feel about targeting the release branch for this?
> 
> The metrics for the crash look encouraging in that the impact is limited, but, because of the December release pause, 18.9 is almost a month away from reaching the users.

@mokagio Yeah, that makes sense! As long as @diegoreymendez feels confident with the fix. Thank you for the heads up! I'll create a new PR based on `release/18.8`, cherry-picking the commit. 🙂

dvdchr:
Closing in favor of #17642.

PR #17656, opened 2021-12-09
Reader Comments: Highlight the cell when it's focused or navigated to

dvdchr:
Refs #17476, pbArwn-2ks-p2#comment-4438
Depends on #17635

As titled, this implements the new highlighted style as discussed in `pbArwn-2ks-p2#comment-4438`. A couple of notes on the behavior:

- There are two components that can be highlighted: the background, and the Reply button.
  - When navigating to a specific comment (i.e. from Notification comments), only the background is highlighted.
  - When tapping on the Reply button, both the background and the Reply button are highlighted.
- With a cell already highlighted, tapping anywhere on the screen un-highlights the cell, hides the keyboard, and updates the Reply text view to target the post.
- When a cell is already highlighted, tapping the Reply button on a different cell will un-highlight the first one, and highlights the new one.
- When tapping on an already highlighted Reply button, it will un-highlight the cell. The Reply text view should be updated to reply to the post.

I also noticed some legacy issues that have preexisted and decided not to tackle them for now:

1. With a cell highlighted and the keyboard is shown, tapping anywhere on the table view (except interactable elements like the Reply button, etc.) should un-highlight the cell and hide the keyboard. However, tapping on a `UITextView` will hide the keyboard, but will NOT un-highlight the cell. 
    - Notice that the Reply text view placeholder is still "Reply to <author>". This is also true with the previous implementation. I'm not sure why it behaved like this, but I think this is not a major issue and I can take a look next week if this needs to be fixed in 18.9.
2. After submitting a reply, there's this "makeshift" comment that's inserted into the table, and then updated after the upload process completes:
  Simulator Screen Shot - iPhone 12 - 2021-12-09 at 19 25 33
  This is because a local comment is created in the `CommentService`, and saved before hitting the remote, and this informs the results controller which in turn updates the table view. I'm holding off from fixing this for now as it could go quite deep into the hierarchy logic and the comments service... but I can create an issue later for this one.

## To test

- Enable the `newCommentThread` feature flag.
- Go to Notifications, and select any comment notification.
- Tap on the header to go into the comment threads.
- 🔍 Verify that the comment background is highlighted, and the Reply button is NOT highlighted.
- 🔍 Verify that the Reply text view placeholder shows "Reply to post".
- Tap a Reply button on a different row.
- 🔍 Verify that the previously highlighted cell is now unhighlighted.
- 🔍 Verify that on the selected row, the background is highlighted and the Reply button is highlighted.
- 🔍 Verify that the Reply text view placeholder is updated to show the new author.
- Try sending a reply.
- 🔍 Verify that a new comment reply is created, and the hierarchy is correct.
- Tap on the highlighted Reply button.
- 🔍 Verify that the row is un-highlighted: background is NOT highlighted, Reply button is NOT highlighted.
- 🔍 Verify that the Reply text view shows "Reply to post".
- Try sending a reply.
- 🔍 Verify that a new post reply is created.

## Regression Notes
N/A. Feature is hidden behind a flag.

N/A. Feature is hidden behind a flag.

N/A. Feature is hidden behind a flag.

PR submission checklist:


ScoutHarris:
I've tried replying via:
- Notifications > header > threaded comments
- My Site > Comments > header > threaded comments
- On iPhone and iPad. 

All do the same thing. That is, after a reply is sent, the view is reloaded and a much of it ends up blank. It happens every time on iPhone. On iPad it seems to only happen when I reply to a comment towards the bottom of the view.

Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-12-09 at 11 58 57


Due to PII, I'll DM you videos so you can see it in action.

ScoutHarris (changes requested):
See comment about blank view.

dvdchr:
👋 @ScoutHarris . Thank you for noticing these issues!

> All do the same thing. That is, after a reply is sent, the view is reloaded and a much of it ends up blank. It happens every time on iPhone. On iPad it seems to only happen when I reply to a comment towards the bottom of the view.

So this is an issue that I think is particular to the Simulator running iOS 15, since I couldn't reproduce this on a device; and this is reproducible in the `release/18.8` branch. I've investigated it further and found a workaround for it – and since this is going to be released pretty soon, I also included the fix in this PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17656/commits/be6a9893870ac30d1e0a7a99285946cc9262671e.

So my assumption is that there could be some sort of race condition between `NSFetchedResultsController` and calling `[tableView reloadData]` manually – since we reload the table right when the context is saving/persisting the data. Since we don't use `saveContextAndWait`, this probably caused the race condition where the data could be in an invalid state when the tableView reloads (which in turn causes the results controller to execute a query on Core Data). Plus, we are also [managing the table view updates manually](https://github.com/wordpress-mobile/WordPress-iOS/blob/3c7f13f59c0a9c1480ed72e42cd064fa23175233/WordPress/Classes/ViewRelated/Reader/ReaderCommentsViewController.m#L402) in Reader Comments, so that might've contributed to the problem.

Delaying `[tableView reloadData]` to the next run loop seems to have fixed this issue... But this is probably just a workaround since the problem I described above is just my assumption.

---

Additionally, I tried to [fix the NRV misalignment in iOS 15](https://github.com/wordpress-mobile/WordPress-iOS/commit/eea5e108442aac01e425992a7ef3db35ddaed21f) through my previous PR, but I've decided to undo the fix and try to address it in a separate PR.

I mentioned that in iOS 15, somehow the `traitCollectionDidChange:` gets called everytime a view controller is pushed – and turns out this only happens when we override Appearance setting (My Site > Me > App Settings > Appearance) AND the overridden value is different with the system scheme. This only happens in iOS 15. So again, I'll try to look for another way to fix this in a separate PR.

Now, I think it's ready for a second look!

ScoutHarris (approved):
The highlighting is really nice! 🥇 

:shipit:

PR #17664, opened 2021-12-10
Workaround for back button not being recognized by XCUITest

tiagomar:
## Context
Looking into last month's UI Tests pipeline failures for `develop` and `release` branches , the most common error message (12 out of 19) really coming from the tests look like the one below and once it happens all the following tests do fail the same way.

> Failed to synthesize event: Failed to scroll to visible (by AX action) Button, {{2.0, 762.0}, {126.0, 48.0}}, identifier: 'mySitesTabButton', label: 'My Site', Selected, error: Error kAXErrorCannotComplete performing AXAction kAXScrollToVisibleAction on element AX element pid: 3066, elementOrHash.elementID: 140453558983040.361. (Underlying Error: Error kAXErrorCannotComplete performing AXAction kAXScrollToVisibleAction on element AX element pid: 3066, elementOrHash.elementID: 140453558983040.361)

Sometimes, for some unknown reason, the Back Button in the navigation bar is not recognized by XCUITest as an element making `navBackButton.tap()` actually tap what would be the second button in the navigation bar, Help button in some cases, breaking `tearDown` and `setUp` for all the tests.

#### Happy case
image
>   ↪︎Find: Descendants matching type NavigationBar
    Output: {
      NavigationBar, {{0.0, 47.0}, {390.0, 44.0}}, identifier: 'Log In'
    }
    ↪︎Find: Element at index 0
      Output: {
        NavigationBar, {{0.0, 47.0}, {390.0, 44.0}}, identifier: 'Log In'
      }
      ↪︎Find: Descendants matching type Button
        Output: {
          **Button, {{0.0, 47.0}, {44.0, 44.0}}, label: 'Get Started'
          Button, {{331.0, 47.0}, {38.0, 44.0}}, identifier: 'authenticator-help-button', label: 'Help'**
        }
        ↪︎Find: Element at index 0
          Output: {
            **Button, {{0.0, 47.0}, {44.0, 44.0}}, label: 'Get Started'**
          }

#### Issue
https://user-images.githubusercontent.com/42008628/145590305-481e9417-e78a-4869-b5b1-c2068cb48074.mov

>   ↪︎Find: Descendants matching type NavigationBar
    Output: {
      NavigationBar, {{0.0, 47.0}, {390.0, 44.0}}, identifier: 'Log In'
    }
    ↪︎Find: Element at index 0
      Output: {
        NavigationBar, {{0.0, 47.0}, {390.0, 44.0}}, identifier: 'Log In'
      }
      ↪︎Find: Descendants matching type Button
        Output: {
          **Button, {{331.0, 47.0}, {38.0, 44.0}}, identifier: 'authenticator-help-button', label: 'Help'**
        }
        ↪︎Find: Element at index 0
          Output: {
            **Button, {{331.0, 47.0}, {38.0, 44.0}}, identifier: 'authenticator-help-button', label: 'Help'**
          }

## Objective
This PR adds a swipe back gesture as redundancy for the back button. 

## Testing
If CircleCI and Buildkite pass, it should be ok.

## Regression Notes
N/A

N/A

N/A

PR submission checklist:

### DIFF
--- a/WordPress/UITestsFoundation/Globals.swift
+++ b/WordPress/UITestsFoundation/Globals.swift
@@ -2,16 +2,16 @@ import UIKit
 import XCTest
 import ScreenObject
 
-// TODO: This should maybe go in an `XCUIApplication` extension? Also, should it be computed rather
-// than stored as a reference? 🤔
-public let navBackButton = XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0)
+// TODO: This should maybe go in an `XCUIApplication` extension?
+public var navBackButton: XCUIElement { XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0) }
 
 // Sometimes the Back Button in Navigation Bar is not recognized by XCUITest as an element.
 // This method identifies when it happens and uses a swipe back gesture instead of tapping the button.
 public func navigateBack() {
-    if XCUIApplication().navigationBars.element(boundBy: 0).buttons.count == 1 {
-        let leftEdge = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
-        let center = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
+    let app = XCUIApplication()
+    if app.navigationBars.element(boundBy: 0).buttons.count == 1 {
+        let leftEdge = app.coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
+        let center = app.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
 
         leftEdge.press(forDuration: 0.01, thenDragTo: center)
     } else { navBackButton.tap() }
--- a/WordPress/UITestsFoundation/Globals.swift
+++ b/WordPress/UITestsFoundation/Globals.swift
@@ -6,6 +6,17 @@ import ScreenObject
 // than stored as a reference? 🤔
 public let navBackButton = XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0)
 
+// Sometimes the Back Button in Navigation Bar is not recognized by XCUITest as an element.
+// This method identifies when it happens and uses a swipe back gesture instead of tapping the button.
+public func navigateBack() {
+    if XCUIApplication().navigationBars.element(boundBy: 0).buttons.count == 1 {
+        let leftEdge = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
+        let center = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
+
+        leftEdge.press(forDuration: 0.01, thenDragTo: center)
+    } else { navBackButton.tap() }
+}
+
 extension ScreenObject {
 
     // TODO: This was implemented on the original `BaseScreen` and is here just as a copy-paste for the transition.
--- a/WordPress/UITestsFoundation/Globals.swift
+++ b/WordPress/UITestsFoundation/Globals.swift
@@ -5,16 +5,23 @@ import ScreenObject
 // TODO: This should maybe go in an `XCUIApplication` extension?
 public var navBackButton: XCUIElement { XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0) }
 
+// This list has all the navBarButton labels currently covered by UI tests and must be updated when adding new ones.
+public let navBackButtonLabels = ["Post Settings", "Back", "Get Started"]
+
 // Sometimes the Back Button in Navigation Bar is not recognized by XCUITest as an element.
 // This method identifies when it happens and uses a swipe back gesture instead of tapping the button.
 public func navigateBack() {
     let app = XCUIApplication()
-    if app.navigationBars.element(boundBy: 0).buttons.count == 1 {
+    let isBackButonAvailableInNavigationBar = navBackButtonLabels.contains(navBackButton.label)
+
+    if isBackButonAvailableInNavigationBar {
+        navBackButton.tap()
+    } else {
         let leftEdge = app.coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
         let center = app.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
 
         leftEdge.press(forDuration: 0.01, thenDragTo: center)
-    } else { navBackButton.tap() }
+    }
 }
 
 extension ScreenObject {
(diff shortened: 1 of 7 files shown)


tiagomar:
Hey @mokagio , @pachlava  and @jostnes !
The tests have already passed Buildkite, still running in Circle CI, but I would to get early review from you sine I'll be running the pipelines several times anyways. =)

WordPress/UITestsFoundation/Globals.swift
```diff
@@ -6,6 +6,17 @@ import ScreenObject
...
 
+// Sometimes the Back Button in Navigation Bar is not recognized by XCUITest as an element.
+// This method identifies when it happens and uses a swipe back gesture instead of tapping the button.
+public func navigateBack() {
+    if XCUIApplication().navigationBars.element(boundBy: 0).buttons.count == 1 {
+        let leftEdge = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
+        let center = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
+
+        leftEdge.press(forDuration: 0.01, thenDragTo: center)
+    } else { navBackButton.tap() }
```
mokagio: I wonder if that's got something to do with my hypothesis that storing elements in `let` instead of computing them at runtime may result in them not being found? 🤔 

See comment in `navBackButton` above: https://github.com/wordpress-mobile/WordPress-iOS/blob/2fc79960c7febaeb1d84051b99928b588cac7bf9/WordPress/UITestsFoundation/Globals.swift#L5-L7

If that was the case, then we could try to compute `navBackButton` instead, which would keep the code simpler:

```swift
public var navBackButton: XCUIElement { XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0) }
```

WordPress/UITestsFoundation/Globals.swift
```diff
@@ -6,6 +6,17 @@ import ScreenObject
 // than stored as a reference? 🤔
 public let navBackButton = XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0)
 
+// Sometimes the Back Button in Navigation Bar is not recognized by XCUITest as an element.
+// This method identifies when it happens and uses a swipe back gesture instead of tapping the button.
+public func navigateBack() {
+    if XCUIApplication().navigationBars.element(boundBy: 0).buttons.count == 1 {
```
mokagio: I had to re-read this code and look at the screenshot to understand why `== 1` was there. At first, I thought it should have been a `.isEmpty` because the comment mentions the button not being there.

How could we make it clearer that we expect there to be 2 buttons? Also, is that a safe assumption to make? 🤔

WordPress/UITestsFoundation/Globals.swift
```diff
@@ -6,6 +6,17 @@ import ScreenObject
...
+        let leftEdge = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
+        let center = XCUIApplication().coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
```
mokagio: We could avoid running through the `XCUIApplication` instantiation flow by encapsulating it in a `let` at the start:

```suggestion
    let app = XCUIApplication()
    if app.navigationBars.element(boundBy: 0).buttons.count == 1 {
        let leftEdge = app.coordinate(withNormalizedOffset: CGVector(dx: 0, dy: 0.5))
        let center = app.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5))
```

mokagio (commented):
Looking good so far, keen to see what hammering CI says. 🤞 

Left a couple of comment suggestions.

WordPress/UITestsFoundation/Globals.swift
tiagomar: This is the tricky part, making it generic enough since navigation bars will look different in many screens. Back buFor the screenshot in the description, yes, 2 buttons were expected. However there are cases were only one is expected (screenshot below) and for those the `swipe back` will be used, but the `tap` would work as well. 

image

This was not on purpose but it works like using a safer route (swipe) when the risk is high (there's only 1 button). I thought about keeping only the `swipe back` but for some unknown reason, the second `swipe back` in a row would never work. Documentation is not XCUITests' super power and I gave up on this at some point. 😓

WordPress/UITestsFoundation/Globals.swift
tiagomar: Added to the next commit.

WordPress/UITestsFoundation/Globals.swift
tiagomar: I can change it and give it a try but for this specific issue I'm pretty sure that test code is not the problem. Accessibility Inspector is not able to identify the button when the issue kicks in and the issue will be gone if you just leave the page and get back to it. That's why I believe it is on the Simulator side. @jostnes also mentioned seeing this happening in previous projects.

jostnes (approved):
Looks good to me! Tested the login set 3 times locally and didn't experience any failures. 

I can see the concern with using number of buttons as the condition to check, but I think it should be okay for now if this reduces the flakiness. 

Also I'm not sure if XCUITest has the ability to do a check if button _not_ exist. If it does, we can also attempt to use that.

tiagomar:
Good point, @jostnes!
XCUITests does have the ability to check if a button exists or not. I initially moved away from this because the button's identifier vary depending on what screen they'll be going back to, making it harder to generalize. However, after reading your comment, and @mokagio's as well, I'll take some time to run all the tests checking every time the navigation back button is tapped looking for similarities in order to have a more deterministic way to decide when to use the `swipe back`. My current guess is that in most cases the buttons will be from the type `back button`.

tiagomar:
I confirmed my guess, all the navigation bar's back buttons had that `Type: back button` (see picture below) property in Accessibility Inspector but XCUITest seems not to be able to read it (I'd love to be wrong). 😓  



I ran all the tests and grabbed all the labels currently in use for `navBackButton` and used this list to validate if the button is properly loaded. This is not the best approach, but it's what we can do with what XCUITest gives us and still much better than counting the buttons.

WordPress/UITestsFoundation/Globals.swift
```diff
@@ -2,9 +2,27 @@ import UIKit
...
 import ScreenObject
 
-// TODO: This should maybe go in an `XCUIApplication` extension? Also, should it be computed rather
-// than stored as a reference? 🤔
-public let navBackButton = XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0)
+// TODO: This should maybe go in an `XCUIApplication` extension?
+public var navBackButton: XCUIElement { XCUIApplication().navigationBars.element(boundBy: 0).buttons.element(boundBy: 0) }
+
+// This list has all the navBarButton labels currently covered by UI tests and must be updated when adding new ones.
+public let navBackButtonLabels = ["Post Settings", "Back", "Get Started"]
```
tiagomar, line 9: Made it public thinking on leveraging it in `tearDowns`.

PR #17670, opened 2021-12-11
Filter media types when presenting the Media Picker for Mobile Gutenberg

twstokes:
Fixes:
- https://github.com/wordpress-mobile/gutenberg-mobile/issues/4371

### Images
1. Ensure at least one image and one video exist in the Photos app and that the WordPress app has appropriate access (configure access to Photos in iOS Settings -> WordPress). It's simplest to allow access to All Photos.
2. Start a new Post and add an Image block
3. Tap "Add Image".
4. Tap "Choose from device".
5. Thumbnails of videos should not be shown
6. Videos should so a count of 0

### Videos
1. Ensure at least one image and one video exist in the Photos app and that the WordPress app has appropriate access (configure access to Photos in iOS Settings -> WordPress). It's simplest to allow access to All Photos.
2. Start a new Post and add an Video block
3. Tap "Add Video".
4. Tap "Choose from device".
5. Thumbnails of images should not be shown

### Before
| Before | After |
| ------ | ----- |
| <video src="https://user-images.githubusercontent.com/2092798/145693360-7cc002a9-6518-4d68-affe-f8a4ce7a3dbe.MP4" /> | <video src="https://user-images.githubusercontent.com/2092798/145693367-3454e8d8-22ba-4f46-9027-f7f654e52757.MP4"> |

## Regression Notes
Any blocks that allow uploading media from the device such as the Image, Gallery, Video, and Media & Text blocks.

I manually tested this on a device.

None.

PR submission checklist:

### DIFF
--- a/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
+++ b/WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
@@ -1,7 +1,9 @@
 import Foundation
 import CoreServices
-import WPMediaPicker
+import UIKit
 import Photos
+import WordPressShared
+import WPMediaPicker
 import Gutenberg
 
 public typealias GutenbergMediaPickerHelperCallback = ([WPMediaAsset]?) -> Void
@@ -39,18 +41,6 @@ class GutenbergMediaPickerHelper: NSObject {
         self.post = post
     }
 
-    fileprivate func defaultMediaPickerOptions() -> WPMediaPickerOptions {
-        let options = WPMediaPickerOptions()
-        options.showMostRecentFirst = true
-        options.filter = [.image]
-        options.allowCaptureOfMedia = false
-        options.showSearchBar = true
-        options.badgedUTTypes = [String(kUTTypeGIF)]
-        options.allowMultipleSelection = false
-        options.preferredStatusBarStyle = WPStyleGuide.preferredStatusBarStyle
-        return options
-    }
-
     func presentMediaPickerFullScreen(animated: Bool,
                                       filter: WPMediaType,
                                       dataSourceType: MediaPickerDataSourceType = .device,
@@ -59,10 +49,7 @@ class GutenbergMediaPickerHelper: NSObject {
 
         didPickMediaCallback = callback
 
-        let mediaPickerOptions = defaultMediaPickerOptions()
-        mediaPickerOptions.filter = filter
-        mediaPickerOptions.allowMultipleSelection = allowMultipleSelection
-
+        let mediaPickerOptions = WPMediaPickerOptions.withDefaults(filter: filter, allowMultipleSelection: allowMultipleSelection)
         let picker = WPNavigationMediaPickerViewController(options: mediaPickerOptions)
         navigationPicker = picker
         switch dataSourceType {
@@ -93,7 +80,7 @@ class GutenbergMediaPickerHelper: NSObject {
 
     private lazy var cameraPicker: WPMediaPickerViewController = {
         let cameraPicker = WPMediaPickerViewController()
-        cameraPicker.options = defaultMediaPickerOptions()
+        cameraPicker.options = WPMediaPickerOptions.withDefaults()
         cameraPicker.mediaPickerDelegate = self
         cameraPicker.dataSource = WPPHAssetDataSource.sharedInstance()
         return cameraPicker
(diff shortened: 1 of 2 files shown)


WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
```diff
@@ -31,7 +31,14 @@ class GutenbergMediaPickerHelper: NSObject {
...
 
-    fileprivate lazy var mediaPickerOptions: WPMediaPickerOptions = {
+    var didPickMediaCallback: GutenbergMediaPickerHelperCallback?
+
+    init(context: UIViewController, post: AbstractPost) {
+        self.context = context
+        self.post = post
+    }
+
+    fileprivate func defaultMediaPickerOptions() -> WPMediaPickerOptions {
```
twstokes: I changed this to a function so that we'd get a fresh `WPMediaPickerOptions` each time rather than mutating a property.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
```diff
@@ -58,7 +58,11 @@ class GutenbergMediaPickerHelper: NSObject {
 
         didPickMediaCallback = callback
 
-        let picker = WPNavigationMediaPickerViewController()
+        let mediaPickerOptions = defaultMediaPickerOptions()
+        mediaPickerOptions.filter = filter
+        mediaPickerOptions.allowMultipleSelection = allowMultipleSelection
+
+        let picker = WPNavigationMediaPickerViewController(options: mediaPickerOptions)
```
ttahmouch, line 53: **Disclaimer:** I don't `Swift` often. Most of my historical knowledge is in `Objective-C`.

If we know that `filter` and `allowMultipleSelection` are more frequently use fields on `WPMediaPickerOptions` instances, then should we consider just passing them directly to the `defaultMediaPickerOptions` method with default parameters? 

e.g.,
```
fileprivate func defaultMediaPickerOptions(filter: WPMediaType = [.image], allowMultipleSelection: Bool = false) -> WPMediaPickerOptions {
...
let picker = WPNavigationMediaPickerViewController(options: defaultMediaPickerOptions(filter: filter, allowMultipleSelection: allowMultipleSelection))
```

This is just a preference thing as I find I tend to write code more in a declarative style with as few state mutations as possible.

**Aside:** It seems like modern `Swift` still requires the parameters to be named even if the variables passed as arguments to the method match the parameter names identically. I'm not positive though.

twstokes: > declarative style with as few state mutations as possible.

That's a good call @ttahmouch. I'll work to touch this up. 👍

WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
```diff
@@ -1,7 +1,9 @@
 import Foundation
 import CoreServices
-import WPMediaPicker
+import UIKit
```
twstokes, line 3: Xcode added the imports automatically, and then I tried to order them by starting with core libraries.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
twstokes: This has been changed to a fileprivate extension containing a static function. Let me know your thoughts!

twstokes: > Aside: It seems like modern Swift still requires the parameters to be named even if the variables passed as arguments to the method match the parameter names identically. I'm not positive though.

This is correct AFAIK.

ttahmouch: The changes read much better to me now. Thanks for addressing, @twstokes . ❤️

The only thing I'm not especially fond of it the use of a static method as that may just make it more difficult to stub if we ever need to mock the options instance for test purposes. I dont think it would be practical to mock with a protocol interface if its static but I may be completely wrong. It also probably doesnt matter too much. I'm not sure if we'd ever need to stage instances for testing.

On second thought, we'd probably be mocking the picker if anything, and I realized its a class extension.

ttahmouch (approved):
I feel weird approving the changes because I'm not super familiar with the code, but if you'd like to also seek a second opinion I totally understand.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
twstokes: Thanks @ttahmouch - I follow somewhat, but would probably need a concrete example to fully understand. I went with the `static` type method to tie closer to `WPMediaPickerOptions`, but certainly the `fileprivate` visibility would make testing a challenge. Outside of protocols, it may be worthy to consider a `class` type method which would be overridable.

If your proposed `fileprivate func defaultMediaPickerOptions` version is a fix to this problem, could you help me understand better?

Unfortunately none of this is currently being tested as it's riddled with `fileprivate` keywords and types that don't belong to the test target. It would definitely be good to get this in better shape down the road.

twstokes:
> I feel weird approving the changes because I'm not super familiar with the code, but if you'd like to also seek a second opinion I totally understand.

No worries! Thanks for taking the time and providing your feedback. I can also get a glance from an iOS dev.

WordPress/Classes/ViewRelated/Gutenberg/GutenbergMediaPickerHelper.swift
ttahmouch: Ive been reviewing from my phone so it may be tough to show an easy example at the moment, but if I remember Ill try to send one later.

Ive noticed that the test coverage is a pretty common occurrence across the board.

ttahmouch: Having the options creator method be a method of something that could be injected as a field via a property of the other class would make it easily mockable, e.g., builder pattern or similar. Ideally though, if we were mocking and injecting anything it would probably be a protocol interface wrapped around the Picker; not the Options.

Sorry, phone usage is hard for reviews. :D

twstokes: Ah, I see! Also, props for doing this all on your phone 😆. Yeah, there'd have to be some refactoring all around to make this better with tests.

frosty (approved):
Looks good to me, and fixes the issue!

PR #17636, opened 2021-12-07
[Tooling] Migrate ShareExtension to use strings from app bundle

AliSoftware:
📝  This is part of project ref paaHJt-2Ib-p2 & paaHJt-2J8-p2
➕  This builds on top of https://github.com/wordpress-mobile/WordPress-iOS/pull/17630 and the `tooling/l10n-step2/ios13-widgets` branch.

## What's in this PR

Same stuff as what I did with the iOS14-style Widget and the 3 iOS13-style widgets in previous PRs:
1. Added `AppLocalizedStrings.swift` to the `WordPressShareExtension` target. Also had to add it to the `WordPressDraftActionExtension` target, because files which I migrated to `AppLocalizedString` in step 2 happen to also be part of that target too, so I needed it to make that target continue to compile too
2. Migrated all the usages of `NSLocalizedString` in any of the files that are part of that `WordPressShareExtension`'s "Compile Sources" build phase, to use `AppLocalizedString` instead, so that it looks up the strings in the app bundle(†)
3. Checked(‡) that all the keys and values in the `*.lproj/Localizable.strings` files that were in  the `WordPressShareExtension` target so fare were also present in the corresponding `WordPress/Resources/*.lproj/Localizable.strings`, with the same value, to ensure it was safe to delete them
4. Delete all the `WordPress/WordPressShareExtension/*.lproj/Localizable.strings` files from the disk and the `ShareExtension` target in Xcode

#### Note: Scripts I used to help me with steps 2 & 3

<details><summary>(†) Script to check that the keys in <tt>.strings</tt> files matched</summary>

This script iterated over each `*.lproj/Localizable.strings` of the ShareExtension target, and for each checks that all the keys and values present in that file were also present and with the same translation in the corresponding `.strings` file from the app bundle.

```ruby
#!/usr/bin/env ruby

require 'json'

### Checks that any key that was in the `*.lproj/Localizable.strings` files of `targets`
### are also present in the main app's `Resources/*.lproj/Localizable.strings`
### and with the same translation.

def dict(plist_path)
  JSON.parse(`plutil -convert json -o - #{plist_path}`)
end

targets = ['WordPressShareExtension', 'WordPressTodayWidget', 'WordPressThisWeekWidget', 'WordPressAllTimeWidget']

targets.each do |target|
  puts "== #{target} =="
  Dir["WordPress/#{target}/*.lproj/Localizable.strings"].each do |tlf|
    lproj = File.basename(File.dirname(tlf))
[...]

PR #17672, opened 2021-12-14
Merge 18.9 code freeze into develop

jkmassel:
Merges the `18.9` code freeze into `develop`, including:
- Version number bumps
- Release toolkit update
- Fastlane + Fastlane Plugin updates
- Production Pods
- Translation pipeline fixes


fastlane/Fastfile
```diff
@@ -418,19 +418,12 @@ platform :ios do
     en_lproj = 'WordPress/Resources/en.lproj'
     ios_generate_strings_file_from_code(
       paths: ['WordPress/', 'Pods/WordPress*/', 'Pods/WPMediaPicker/', 'WordPressShared/WordPressShared/', 'Pods/Gutenberg/'],
-      exclude: ['*Vendor*', 'WordPress/WordPressTest/I18n.swift', 'WordPress/WordPressStatsWidgets/Views/Localization/AppLocalizedString.swift'],
+      exclude: ['*Vendor*', 'WordPress/WordPressTest/I18n.swift', 'WordPress/WordPressStatsWidgets/Views/Localization/AppLocalizedString.swift', 'WordPress/Classes/Utility/AppLocalizedString.swift'],
```
jkmassel: @AliSoftware – I needed to make this change to get this method to work – LMK if that looks correct? I think we could potentially also use a wildcard here, but I kept it simple to be safe.

fastlane/Fastfile
```diff
@@ -418,19 +418,12 @@ platform :ios do
...
       routines: ['AppLocalizedString'],
       output_dir: en_lproj
     )
 
-    # @TODO: Remove once we migrated the codebase of the widgets to use the app bundle for .strings
```
jkmassel, line 426: @AliSoftware – I needed to remove these in order to complete the code freeze – I believe these were resolve as part of #17630 and #17636 , but LMK if anything doesn't look right?

fastlane/Fastfile
AliSoftware: The file was actually moved during https://github.com/wordpress-mobile/WordPress-iOS/pull/17630, from its previous location (in `WordPress/WordPressStatsWidgets/Views/Localization/AppLocalizedString.swift`) to the new one you added; so you should just replace the old one with the new actually:
```suggestion
      exclude: ['*Vendor*', 'WordPress/WordPressTest/I18n.swift', 'WordPress/Classes/Utility/AppLocalizedString.swift'],
```

(I should have done that in https://github.com/wordpress-mobile/WordPress-iOS/pull/17630 alongside the file move but totally missed it back then)

fastlane/Fastfile
AliSoftware: That's correct and I should have removed those as part of those PRs, thanks for taking care of this.

There might still be an issue waiting to happen during `new_beta_release` and `finalize_release` when those will call the scripts downloading the translations from GlotPress and those scripts (which are still yet to be migrated) might still be trying to re-dispatch the keys to those widget targets based on those (now deleted) `Base.lproj` keys. Until that side of the migration of our scripts is over we might need to [patch the `extract-framework-translations.swift` scripts](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/Scripts/extract-framework-translations.swift#L10-L13) (or even stop calling that script altogether by removing [these lines](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/Scripts/update-translations.rb#L166-L168)) to stop doing that for those targets until I finalize the related work.

oguzkocer (approved):

PR #17676, opened 2021-12-15
Cleanup Xcode project and unused files

AliSoftware:
This PR does a couple of cleanup (which was needed as a preliminary to my PR to come on SwiftLint)

It consists mainly of deletion of unused/obsolete source files, `pbxproj` cleanup, and fastlane/CP update.

## Xcode project (`.pbxproj`) cleanup

 - Disable "Automatic Code Signing" on the `WordPressScreenshotGeneration`, `JetpackScreenshotGeneration` and `WordPressUITests` targets, so that it does not try to register their bundleID to our Enterprise account (and fail due to permissions for most users) — b549315f2e76028bb16c62d1308e04f54c56849c
 - Remove duplicate references to `Pods_WordPressScreenshotGeneration` (found by Xcode itself and its "Validate Project" warnings) — 8f679875e8aa53d39bda0d0e1893ccc80e0d0d5e
 - Remove reference to non-existing file `WebViewAuthenticator.swift` (file was not in any target and did not exist anymore in the repo) — 2870c4a5b74ef48f76d3204f04adf71626080210
 - Update the `LastUpgradeVersion` field of shared `xcscheme` (done by Xcode itself while running "Validate Project") — 9a31461f444150a877ce2d0572b023fbaed6073f

## Tools Update

 - `bundle update fastlane`
 - `bundle update cocoapods`
 - Then ran `bundle exec pod install` to update the `xcodeproj` appropriately (and moved a file around and back in the project to make Xcode re-save the project file by itself to be sure)

Relevant commit: eff5d73597466f11e5fc3dcd3b80811d39419cc6

## Remove Obsolete Files on Disk

I found _a lot_ of files that were still on disk and in git while they were not used anymore in the project.

<details><summary>I used the below script to detect them</summary>
I wrote and run the following script to compare .h/.m/.mm/.swift files that were in the project's File References and in any of the Compile Source build phases of any of the targets, and compare those with the files on disk.

```ruby
#!/usr/bin/env ruby

require 'xcodeproj'

def only_source_files(list)
  list.select { |path| %w[m mm swift].any? { |ext| File.extname(path) == ".#{ext}" }}
end

def list(array)
  array.map { |x| " - #{x}" }.join("\n")
end

project = Xcodeproj::Project.open('WordPress/WordPress.xcodeproj')

project_filerefs = project.files.map(&:real_path).map(&:to_s)

build_phases = project.targets.map do |t|
[...]

PR #17677, opened 2021-12-15
[Tooling] Add Swiftlint rule for NSLocalizedString -> AppLocalizedString

AliSoftware:
This work is part of parent project paaHJt-2Ib-p2 & subproject paaHJt-2J8-p2
Especially see comment paaHJt-2J8-p2#comment-5650

~~ℹ️  This PR is built on top of https://github.com/wordpress-mobile/WordPress-iOS/pull/17676 (target branch: `cleanup/project-files`) because some of the obsolete files that got deleted had violations; so I chose to delete those obsolete files as part of https://github.com/wordpress-mobile/WordPress-iOS/pull/17676 before fixing the remaining, valid files to use `AppLocalizedString`.~~ Update: #17676 has been merged, base is now the standard main branch.

## What it does

 - Add a `swiftlint` rule to enforce the use of `AppLocalizedString` instead of `NSLocalizedString` everywhere except in Tests and except in `.m` files (because top-level functions in Swift are not exposed to ObjC…)
 - Replace all occurrences of `NSLocalizedString` with `AppLocalizedString` in every `*.swift` files except tests
 - Fix one of the Unit Test which was relying on the locale the test was run on to be English
 - Fix the SwiftLint aggregate target / build phase to be compatible with `rbenv`

## RFC: Is it the right approach?

TBH I'm not sure this is the right approach for multiple reasons:

 - This is a bit unusual/unconventional to have to use `AppLocalizedString` and not `NSLocalizaedString` in every place – even the ones only used in the app
 - This enforcement will differ from other app codebases like WooCommerce, because WCiOS does not currently have app extensions or widgets, so there's no reason to enforce a similar rule on WCiOS. Which means we'd have different rules / conventions for localized string in WPiOS vs WCiOS

:one: IMHO, it would make way more sense to extract any source code that is common and used in both the app and the widget(s) and app extensions into one or more separate framework target (e.g WPStats.framework), then make the app and widget target both depend on it. That would then allow us to focus the SwiftLint rule about `AppLocalizedString` only on this target's folder — instead of enforcing it everywhere.

But that ☝️  would be a change out of scope of my current project, which is why I didn't go that route.

:two: Instead, this Draft PR is meant to be used as a point of discussion and see the impact that it could have if we still go with the "Enforce `AppLocalizedString` everywhere" solution instead in the meantime, its impacts and its Pros and Cons.

3️⃣  Another, middle-ground solution would be to keep the status quo (`NSLocalizedString` in most places, no separate fmk yet) and just specify the paths to the shared files explicitly in the `.swiftlint.yml`; but that would not be very scalable if more files are added to the app extension / widget targets…

## PR Status: Ready to Review / RFC, but still Draft / Do Not Merge

The PR is ready for review, but it was created more as a way to see the impact it would have on the codebase and start the discussion mentioned above.
Besides, even if we decide to go with that solution, I'd still want to write a P2 post to officially announce such a change before landing it.

## To test

 - Select the "SwiftLint" scheme in Xcode and build it (⌘B), and check that it passes without error nor warning. (Alternatively, you can run `bundle exec rake lint` from the Terminal.)
 - Introduce an intentional violation by replacing one of the `AppLocalizedString` instance in the code with `NSLocalizedString`, re-run the SwiftLint scheme, and check that it complains about the violation. Then revert your change.
 - Ensure that the project builds for both the WordPress and Jetpack schemes
 - Ensure the Unit Tests still build and pass
 - Smoke-test the app in a locale other than English to ensure that it still uses the translations for that non-English locale


AliSoftware:
Mmmmh this seems wrong, Hound [should have run swiftlint](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/.hound.yml#L6-L7) and found the violation I left intentionally in 8b4861b67d37197553cbfae213866b4ec59cc47b… (and which is found during `rake lint`) 🤔
Any ideas anyone?

image

AliSoftware:
> Mmmmh this seems wrong, Hound [should have run swiftlint](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/.hound.yml#L6-L7) and found the violation I left intentionally in [8b4861b](https://github.com/wordpress-mobile/WordPress-iOS/commit/8b4861b67d37197553cbfae213866b4ec59cc47b)… (and which is found during `rake lint`)

🤔 💭 :bulb:  Given Hound is defined globally for the repo, maybe Hound uses the `.hound.yml` — and the `.swiftlint.yml` file it references — from… the default branch (aka `trunk` or `develop`), instead of using the one from the current PR? Which would be not-great from them, but the best explanation I can think of…

AliSoftware:
I just [reverted](https://github.com/wordpress-mobile/WordPress-iOS/pull/17677/commits/28f18377cff38d0235db7822f27c82ff191ab5e6) the commit which [previously added intentional violation](https://github.com/wordpress-mobile/WordPress-iOS/pull/17677/commits/8b4861b67d37197553cbfae213866b4ec59cc47b), given that Hound won't detect it until the `.swiftlint.yml` lands in `develop` anyway.

AliSoftware:
A summary of Pros/Cons I can think of with the different options:

<table>
<tr><th>Solution</th><th> 👍  Pros</th><th> 👎  Cons</th></tr>
<tr>
  <td>:one: Extract common code in a framework target, and restrict the <tt>swiftlint</tt> rule to only scan the app extensions + this frameworks</td>
  <td><ul>
    <li>Cleaner architecture</li>
    <li>Does not enforce <tt>AppLocalizedString</tt> where it's not needed</li>
    <li>Avoid double standard with WCiOS</li>
  </ul></td><td><ul>
    <li>Will require larger work, outside of the scope of my current project</li>
  </ul></td>
</tr><tr>
  <td>:two: Enforce <tt>AppLocalizedString</tt> everywhere (this PR)</td>
  <td><ul>
    <li>Easier rule to write in <tt>.swiftlint</tt></li>
    <li>Easier to document, and to reason about for developers</li>
  </ul></td><td><ul>
    <li>Not consistent with WCiOS where this likely won't be enforced (yet)</li>
    <li>Requires a change of habit from developers everywhere even if the requirement is only for files used in AppEx in practice</li>
  </ul></td>
</tr><tr>
  <td>:three: Hardcode path to files in <tt>.swiftlint.yml</tt></td>
  <td><ul>
    <li>Limits the enforcement only to where strictly necessary</li>
    <li>Does not require any large migration nor any big architecture rewrite</li>
  </ul></td><td><ul>
    <li>Not scalable, since adding a new file to one of the Widget targets might not end up being included</li>
    <li>After giving it a try (see comment below), the RegEx would end up way too convoluted, not realistic in practice</li>
  </ul></td>
</tr><tr>
  <td>:three:🅱️  Hardcode path to files in <tt>.swiftlint.yml</tt> + add CI check for sync'ing the file list</td>
  <td><ul>
    <li>Limits the enforcement only to where strictly necessary</li>
    <li>Does not require any large migration nor any big architecture rewrite</li>
    <li>A script parsing the Xcodeproj to check the files in those app extension targets, and comparing it with the ones listed in the <tt>.swiftlint.yml</tt>, would warn us if the rule needs to be updated</li>
  </ul></td><td><ul>
    <li>Feels a bit hack-y?</li>
    <li>After giving it a try (see comment below), the RegEx would end up way too convoluted, not realistic in practice</li>
  </ul></td>
</tr>
</table>

AliSoftware:
So, I tried to implement 3️⃣  / 3️⃣ 🅱️  to compare it with this PR, but it turns out there are really a lot of files from everywhere in the filesystem included in those app extension targets… so that won't be easy.

This script below lists any files that is part of at least one app extension target… but for which even a regex as broad as `%r{WordPress/(WordPress.*Extension|WordPress.*Widgets?|WordPressIntents|Classes/ViewRelated/Stats)/.*\.(swift|m)}` for the `included:` field of of the rule in `.swiftlint.yml` would not be enough match.

<details><summary>Script to print all source files that are in any App Extension target but not caught by my initial regex</summary>

```ruby
#!/usr/bin/env ruby

require 'xcodeproj'
require 'yaml'

root = Pathname.new('..').realpath
project = Xcodeproj::Project.open(root + 'WordPress' + 'WordPress.xcodeproj')
extension_targets = project.targets
  .select { |t| t.is_a?(Xcodeproj::Project::Object::PBXNativeTarget) && t.extension_target_type? }

def included?(p)
  re = %r{WordPress/(WordPress.*Extension|WordPress.*Widgets?|WordPressIntents|Classes/ViewRelated/Stats)/.*\.(swift|m)}
  re.match?(p.to_s)
end

def source_file?(f)
  ['.m', '.swift'].any? { |ext| File.extname(f.to_s) == ext }
end

missing_files_count = {}
extension_targets.each do |t|
  build_phase = t.build_phases.find { |p| p.is_a?(Xcodeproj::Project::Object::PBXSourcesBuildPhase) }
  next if build_phase.nil?
  paths = build_phase.files_references
    .map { |p| p.real_path.relative_path_from(root).to_s }
    .select { |f| source_file?(f) }
    .sort
  target_violations = paths.reject { |path| included?(path) }
  target_violations.each { |f| missing_files_count[f] = (missing_files_count[f] || 0) + 1 }
end

puts missing_files_count.map { |f,n| "#{f} (#{n})" }.join("\n")
```
</details>

When running the above script, it still lists a lot (50+) of files that are part ofan App Extension targets but still wouldn't be caught by the regex above. Showing how much spread those files included in AppEx targets are, and suggesting that our RegEx would really need to become quite convoluted if we wanted it to catch all those files to be linted for `AppLocalizedString` usage.

<details><summary>All the 50+ source files included in AppEx still not caught by the regex we'd use for the swiftlint rule</summary>

```
WordPress/Classes/Extensions/Colors and Styles/MurielColor.swift (6)
WordPress/Classes/Extensions/Colors and Styles/UIColor+MurielColors.swift (6)
WordPress/Classes/Extensions/Colors and Styles/UIColor+WordPressColors.swift (6)
WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+ApplicationStyles.swift (2)
WordPress/Classes/Extensions/Colors and Styles/WPStyleGuide+Gridicon.swift (2)
WordPress/Classes/Extensions/NSAttributedStringKey+Conversion.swift (2)
WordPress/Classes/Extensions/String+Ranges.swift (2)
WordPress/Classes/Extensions/String+RegEx.swift (2)
WordPress/Classes/Extensions/UIApplication+mainWindow.swift (2)
WordPress/Classes/Services/KeyValueDatabase.swift (8)
WordPress/Classes/Services/ShareExtensionService.swift (2)
WordPress/Classes/System/Constants.m (9)
WordPress/Classes/Utility/App Configuration/AppConfiguration.swift (2)
WordPress/Classes/Utility/App Configuration/AppStyleGuide.swift (6)
WordPress/Classes/Utility/AppLocalizedString.swift (6)
WordPress/Classes/Utility/BuildInformation/BuildConfiguration.swift (7)
WordPress/Classes/Utility/BuildInformation/FeatureFlag.swift (7)
WordPress/Classes/Utility/BuildInformation/FeatureFlagOverrideStore.swift (8)
WordPress/Classes/Utility/Logging/CocoaLumberjack.swift (7)
WordPress/Classes/Utility/SFHFKeychainUtils.m (9)
WordPress/Classes/ViewRelated/Aztec/Extensions/FormatBarItemProviders.swift (2)
WordPress/Classes/ViewRelated/Aztec/Extensions/Header+WordPress.swift (2)
WordPress/Classes/ViewRelated/Aztec/Extensions/TextList+WordPress.swift (2)
WordPress/Classes/ViewRelated/Aztec/Processors/ImgUploadProcessor.swift (2)
WordPress/Classes/ViewRelated/Cells/WPReusableTableViewCells.swift (2)
WordPress/Classes/ViewRelated/Post/Scheduling/LightNavigationController.swift (2)
WordPress/Classes/ViewRelated/Tools/TableViewKeyboardObserver.swift (2)
WordPress/Classes/ViewRelated/Views/NoResultsViewController.swift (2)
WordPress/Classes/ViewRelated/Views/WPAnimatedBox.m (2)
WordPress/Shared/SharePost.swift (2)
WordPress/Tracks+ShareExtension.swift (1)
WordPress/Classes/ViewRelated/Views/CircularImageView.swift (1)
WordPress/Classes/Extensions/NSAttributedString+Helpers.swift (1)
WordPress/Classes/Extensions/UIImage+Assets.swift (1)
WordPress/Classes/Utility/FormattableContent/Actions/DefaultFormattableContentAction.swift (1)
WordPress/Classes/Utility/FormattableContent/Actions/FormattableContentAction.swift (1)
WordPress/Classes/Utility/FormattableContent/Actions/FormattableContentActionCommand.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableContent.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableContentFactory.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableContentGroup.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableContentRange.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableContentStyles.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableMediaContent.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableRangesFactory.swift (1)
WordPress/Classes/Utility/FormattableContent/FormattableTextContent.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/FormattableCommentContent.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/FormattableUserContent.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/Notifiable.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/NotificationContentFactory.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/NotificationTextContent.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/Ranges/FormattableNoticonRange.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/Ranges/NotificationCommentRange.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/Ranges/NotificationContentRange.swift (1)
WordPress/Classes/ViewRelated/Notifications/FormattableContent/Ranges/NotificationContentRangeFactory.swift (1)
```
</details>

---

Note that also suggests that solution 1️⃣ , which would consist of grouping all those files (used by the app extension targets and the main app) into a dedicated framework… would likely be quite tricky and complex too, given all the different places those files are spread into.

mokagio:
**TL;DR** All in all, I think the pros of adopting and enforcing `AppLocalizedString` everywhere (:two:) outweigh the cons. Doing so also makes the code internally consistent, which is beneficial (no "when do I use one vs the other" kind of questions). If we think that adding a lint for this will be useful, we should go for it.

---

I think, in the long run, extracting the files in dedicated frameworks ( 1️⃣ ) is the way to go. As you mention, this is outside the scope of the localization tooling project and I also agree it would take a non trivial amount of time, based on my experience with the extracting classes in the UI testing frameworks.

You already showed that hardcoding paths is possible but hard to maintain.

In you opinion, would it be possible to make the RegEx more straightforward by doing a bit of file shuffling and tidying up? Or, is it just that there's so much going on in this project that any kind of folder structure convention is bound to "leak" files?

Pending that, it seems like enforcing `AppLocalizedString` (2️⃣) is the only actionable option within our constraints. So it's a matter of weighing up the pros of having the `AppLocalizedString` usage ensured where we need it against the mental overhead and break from convention of not using `NSLocalizedString`.

I'm warming up to the idea of using a non-standard approach, i.e. `AppLocalizedString` everywhere because at least it's consistent. Once you internalize that `AppLocalizedString` is the norm in this codebase, you don't have to ask yourself the question "is this a file in which I should use `NSLocalizedString` or the custom one?" every time you write a string.

It's true that breaking away from conventions is a bit tricky. It will certainly add a bit of overhead to onboarding and might be surprising for external folks at first. But the SwiftLint integration will go a long way to help people getting used to this.

The only argument I can come up against going for it is that, if we all agree building dedicated framework would be beneficial beyond just reducing the lint complexity, maybe we should wait for that to happen instead of introducing this change. This argument doesn't seem too strong because building more framework is not something in the timeline (?), so we should shuffle things around for it.

One final question is whether we should have this kind of check at all. I think the answer is yes. The whole point of introducing `AppLocalizedString` was ensuring we have a more robust way to localize extensions. Automating the check is a more efficient way to ensure the usage occurs, and therefore the localization, than to rely on every developer to know about it and hoping no issue escapes code review.

AliSoftware:
@mokagio I tend to agree, even if it still pains me a bit to had a difference between WP and WC standards. Maybe we should enforce the same rule in WCiOS even if in practice it is not needed, also because at some point they plan to add widgets to WCiOS (ref: p1639663766173700-slack-C6H8C3G23) so this will likely become needed in the future anyway.

Once think worth adding is a suggestion @jkmassel made, of making it a Peril/Danger rule instead of SwiftLint, which would detect any introduction of a new `NSLocalizedString` call in any PR, and post an (inline?) comment suggesting to double-check if that is correct or should be a `AppLocalizedString` instead. That would mitigate the "use `AppLocalizedString` everywhere even if it's only really needed in 10% of cases" thing, but would also easily fall into the cracks of reviewers inattention. Besides, even if one call does not need to be an `AppLocalizedString` today (because the file is only used in the app target(s)), it could easily suddenly become necessary later if that file gets added to an appex target in a future PR…

AliSoftware:
TL;DR: I think we'll end up opting for 2️⃣  anyway (`AppLocalizedString` everywhere) as it seems our only way forward (except the "doing nothing solution" which would kinda defeat the whole purpose of having done this whole work).

But we should wait until all the rest of the work on the L10n project is done and we're ready to post a P2 with all the new instructions and conventions documented officially before merging this change still, so that it doesn't come out of the blue to developers.

AliSoftware:
Oh, just thought of a new option 4️⃣ :
 - Write lane that would compute the list all the files in the `Xcodeproj` that belongs to at least one appex (like the scripts I already wrote in one of my previous comments), then `grep` for `NSLocalizedString` in each file found, then issue a warning for each violation used
   - In the context of running that lane locally (via a Script Build Phase on build?), emit that violation as a `warning:` in Xcode
   - In the context of running that lane on CI as part of a PR, we'd use [our new `comment_on_pr` fastlane action](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb) to write the violation in the PR comment (ideally an inline comment attached to the line each violation was found, though we'd need to add support for that to our new action but doesn't seem too complex)

Basically it would do something very similar to what swiftlint would do, except by writing that lane ourselves instead of writing it as a SwiftLint rule, we'll be able to test for Xcode target membership and apply the rule only to the relevant files.

WDYT?

mokagio:
@AliSoftware I'm going to bump this to the next release because we'll be code freezing 19.0 today and this is still a draft.

I haven't had a chance to read and reply to your comments, by the way, as I just came back from a week AFK and I'm starting with the code freeze as the most time critical task.

mokagio:
I'm going to bump this to the next release because we'll be code freezing 19.1 today and this is still a draft.

I will followup on the conversation here this week...

mokagio:
@AliSoftware late comment on your RFC, sorry about it. I like [option 4](https://github.com/wordpress-mobile/WordPress-iOS/pull/17677#issuecomment-1004037861) for the reasons you listed:

> by writing that lane ourselves instead of writing it as a SwiftLint rule, we'll be able to test for Xcode target membership and apply the rule only to the relevant files.

On top of it, as much as I like SwiftLint and I don't see it going anywhere (I know Apple introduced [swift-format](https://github.com/apple/swift-format) and that it has [a `lint` subcommand](https://github.com/apple/swift-format#linting), as far as I've seen, it's not as powerful and hasn't been adopted much yet?), by keeping the logic at the release toolkit level we avoid spreading critical-ish parts of our automation across different tools. Given the implementation cost seems comparable, that seems another point in favor of option 4.

mokagio:
I'm going to bump this to the next release because we'll be code freezing 19.2 today and this is still a draft.

mokagio:
I bumped this to the next release because we'll be code freezing 19.4 today and this is still a draft.

AliSoftware:
> Oh, just thought of a new option 4️⃣ :
>
> [compute the list all the files in the `xcodeproj` that belongs to at least one appex, grep for `NSLocalizedString` in each file found, then issue a warning for each violation used]

I ended up implementing that solution in #18427, with just some slight variations:
 - Implemented the linter as a `Scripts/BuildPhase/*.rb` script (as opposed to a lane in fastlane as initially implied by past comment above)
 - Made the linter generate an error instead of a warning. After all, it _is_ an error, as it would make the string never appear translated at runtime in languages other than English when displayed in the context of those app extension targets
 - Called the linter as a Script Build Phase of _each_ app extension target (so that if one builds an extension target in isolation, they would also see the error even if not building the whole app or scheme other than the main one)
 - Didn't do any special treatment for Xcode vs CI, as if the build fails in Xcode due to one of the linter script build phase failing, CI will fail as well after all. We don't report linter errors as PR comments for other build errors/warnings, so no reason to give those special treatments either.

I'm closing this Draft PR in favor of #18427.

PR #17683, opened 2021-12-17
[Tooling] Cleanup files scanned for NSLocalizedString

AliSoftware:
### What

 - Updates the list of excluded files during the `generate_strings_file_for_glotpress` lane so that it excludes Unit Tests and uses more generic glob patterns
 - Fixes the `I18n.swift` file which was referencing `NSLocalizedString` while it shouldn't use it when it's a use case like here not intended to be extracted by `genstrings`.
 - Fixes a unit test using `NSLocalizedString` on a string format that was not localized in the tested app code anyway.
 - Updates the `release-toolkit` to `2.3.0` now that it's out (because my `bundle install` was complaining about `develop` when I tried to test those changes otherwise)

### To Test

 - Create a test branch from this one
 - In the commit that was created, observe that the `WordPress/Resources/en.lproj/Localizable.strings` file has a couple of keys or comments removed
 - For each removed comment or key, search all the occurrences of the corresponding key in the source code and confirm that the instances that were removed were only present in the Unit Tests.
    - For example, for the `"%@ comments"` key, the `Number of Comments` translator hint got removed from the `.strings` file. but we can confirm that [the only instance with that comment for that key is in the `ReaderPostCardCellTests.swift`](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressTest/ReaderPostCardCellTests.swift#L164) and was thus only used for unit tests.
    - Meanwhile, the only 2 other instances are both used in the app code — [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Blog/Site%20Settings/DiscussionSettingsViewController.swift#L529) and [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/ViewRelated/Reader/ReaderPostCardCell.swift#L850) — and the comments for those two are still present on that key in the generated `.strings` file.
   - Likewise, for the `"%@, %@"` key which got completely removed (comment, key and translation) from the `.strings` file, we can confirm that [the only call sites using that key were all in `PostTests.swift`](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/WordPressTest/PostTests.swift#L324-L343) and nowhere used in the app itself, which is why it's safe to remove them from translation. In fact, I've updated that test in this PR because in the corresponding app code this test is testing, [the `%@, %@` format string itself isn't translated (but its individual components are)](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Models/Post.swift#L281-L287).
 - Delete your test branch


fastlane/Fastfile
```diff
@@ -418,7 +418,7 @@ platform :ios do
     en_lproj = 'WordPress/Resources/en.lproj'
     ios_generate_strings_file_from_code(
       paths: ['WordPress/', 'Pods/WordPress*/', 'Pods/WPMediaPicker/', 'WordPressShared/WordPressShared/', 'Pods/Gutenberg/'],
-      exclude: ['*Vendor*', 'WordPress/WordPressTest/I18n.swift', 'WordPress/Classes/Utility/AppLocalizedString.swift'],
+      exclude: ['*Vendor*', 'WordPress/WordPressTest/**', '**/AppLocalizedString.swift'],
```
AliSoftware, line 421: Excluding all the test files. After all, we don't want to pick up any `NSLocalizedString` (and their `comment:`) used in tests for the strings we send to translators.

WordPress/WordPressTest/PostTests.swift
```diff
@@ -323,11 +323,11 @@ class PostTests: XCTestCase {
 
         revision.status = .pending
         let pendingStatusDisplay = "\(Post.title(for: .pending))"
-        XCTAssertEqual(revision.statusForDisplay(), String(format: NSLocalizedString("%@, %@", comment: ""), pendingStatusDisplay, local))
```
AliSoftware, line 326: This `"%@, %@"` format [is not itself translated in the `Post.swift` file and class](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Models/Post.swift#L284) that this tests (instead the 2 arguments are `NSLocalizedString`'d separately); so no reason to use `NSLocalizedString` for the format in the test itself.

mokagio (approved):
I didn't check _all_ removed comments and keys, just a handful, but they all behaved as expected 👍 

For the record, here's the diff of the commit generated on my end

<details>
<summary>Click to expand</summary>

```diff
commit ee757baf83cbb776b0aeba148c64b398048840fa
Author: Gio Lodi <giovanni.lodi42@gmail.com>
Date:   Mon Dec 20 16:08:07 2021 +0100

    Update strings for localization

diff --git a/WordPress/Resources/en.lproj/Localizable.strings b/WordPress/Resources/en.lproj/Localizable.strings
index e1bfb623d3..efe098d02a 100644
--- a/WordPress/Resources/en.lproj/Localizable.strings
+++ b/WordPress/Resources/en.lproj/Localizable.strings
@@ -64,7 +64,6 @@
 "%@ comment" = "%@ comment";
 
 /* Accessibility label for comments button (plural)
-   Number of Comments
    Number of Comments per Page */
 "%@ comments" = "%@ comments";
 
@@ -117,9 +116,6 @@
 /* Difference label for Period Overview stat, indicating change from previous period. Ex: +99.9K (5%) */
 "%@%@ (%@%%)" = "%1$@%2$@ (%3$@%%)";
 
-/* No comment provided by engineer. */
-"%@, %@" = "%1$@, %2$@";
-
 /* Accessibility label for a post in the post list. The parameters are the title, and date respectively. For example, \"Let it Go, 1 hour ago.\" */
 "%@, %@." = "%1$@, %2$@.";
 
@@ -241,8 +237,7 @@
 /* translators: Missing block alert title. %s: The localized block name */
 "'%s' is not fully-supported" = "'%s' is not fully-supported";
 
-/* (no title)
-   Lets a user know that a local draft does not have a title. */
+/* Lets a user know that a local draft does not have a title. */
 "(no title)" = "(no title)";
 
 /* Empty Post Title */
@@ -2829,8 +2824,7 @@
 /* Message displayed when approving a comment fails. */
 "Error approving comment." = "Error approving comment.";
 
-/* No comment (test)
-   This secondary message is displayed if a user encounters a general error. */
+/* This secondary message is displayed if a user encounters a general error. */
 "Error communicating with the server, please try again" = "Error communicating with the server, please try again";
 
 /* There was an error deactivating a plugin, placeholder is the plugin name */
@@ -4124,7 +4118,6 @@
 "Local" = "Local";
 
 /* A status label for a post that only exists on the user's iOS device, and has not yet been published to their blog.
-   Local
    Title of the Local Changes Badge */
 "Local changes" = "Local changes";
 
@@ -4179,8 +4172,7 @@
 /* LogOut confirmation text, whenever there are no local changes */
 "Log out of WordPress?" = "Log out of WordPress?";
 
-/* Error message shown when a user is trying to login.
-   Login Request Expired */
+/* Login Request Expired */
 "Login Request Expired" = "Login Request Expired";
 
 /* No comment provided by engineer. */
@@ -4391,7 +4383,6 @@
    Accessibility label for the More button on Reader's post details
    Action button to display more available options
    Label for the more post button. Tapping displays an action sheet with post options.
-   More
    Spoken accessibility label */
 "More" = "More";
 
@@ -4683,8 +4674,7 @@
 /* Title displayed when the user has removed all Insights from display. */
 "No insights added yet" = "No insights added yet";
 
-/* No comment (test)
-   This primary error message is displayed if a user encounters an issue with network reachability. */
+/* This primary error message is displayed if a user encounters an issue with network reachability. */
 "No internet connection" = "No internet connection";
 
 /* Error message shown when the user is browsing Notifications without an internet connection.
@@ -5194,7 +5184,6 @@
 "Pending" = "Pending";
 
 /* Name for the status of a post pending review.
-   Pending review
    Title of the Pending Review Badge */
 "Pending review" = "Pending review";
 
@@ -5676,7 +5665,6 @@
 
 /* Name for the status of a published post.
    Period Stats 'Published' header
-   Published
    Title of the published filter. This filter shows a list of posts that the user has published.
    Title of the published header in search list. */
 "Published" = "Published";
@@ -6120,8 +6108,7 @@
 /* Text displayed in the share extension's summary view that describes the save draft post action. */
 "Save draft post on:" = "Save draft post on:";
 
-/* Accessibility label for the 'Save Post' button.
-   Save post */
+/* Accessibility label for the 'Save Post' button. */
 "Save post" = "Save post";
 
 /* Title of alert informing users about the Reader Save for Later feature. */
@@ -6180,7 +6167,6 @@
 "Schedule Now" = "Schedule Now";
 
 /* Name for the status of a scheduled post
-   Scheduled
    Title of the scheduled filter. This filter shows a list of posts that are scheduled to be published at a future date. */
 "Scheduled" = "Scheduled";
 
@@ -7349,7 +7335,6 @@
 "There is an issue connecting to %@. Reconnect to continue publicizing." = "There is an issue connecting to %@. Reconnect to continue publicizing.";
 
 /* Displayed during Site Creation, when searching for Verticals and the server returns an error.
-   No comment (test)
    The Jetpack view title used when an error occurred
    This primary message message is displayed if a user encounters a general error. */
 "There was a problem" = "There was a problem";
@@ -7696,8 +7681,7 @@
 
 /* Name for the status of a trashed post
    Title of the trashed filter. This filter shows posts that have been moved to the trash bin.
-   Title of trashed Comments filter.
-   Trashed */
+   Title of trashed Comments filter. */
 "Trashed" = "Trashed";
 
 /* Connect when the SSL certificate is invalid */
@@ -8195,8 +8179,7 @@
 /* Message shown when a verification email was re-sent succesfully */
 "Verification email sent, check your inbox." = "Verification email sent, check your inbox.";
 
-/* Push Authentication Alert Title
-   Title of a prompt. A user must verify their login attempt. */
+/* Push Authentication Alert Title */
 "Verify Log In" = "Verify Log In";
 
 /* Notice displayed after domain credit redemption success. */
```

</details>

---

I'm going to solve the conflict currently on `Gemfile.lock` and then merge to avoid this getting stale during your/ours end of year AFKs.

